PR #4, opened 2016-08-28
allow users to attach local files

ChristophWurst:
client-side TODO
- [x] upload files asynchronously
- [x] add uploaded files to list of attachments
- [ ] show loading and/or progress indicator
- [ ] lock 'Send'/'Reply' button until all local attachments have been uploaded successfully
- [ ] handle errors (remove faulty attachment from list)
- [ ] add special property to attachment so that server can distinguish between local and cloud attachments 

server-side TODO
- [x] add attachment store for asynchronously uploaded files
- [ ] move uploaded attachments to temp dir (use core's temp manager)
- [ ] limit uploaded file size to a sane size (e.g. a few MBs)
- [ ] distinguish between local and cloud attachments when sending a message and load local attachments from the attachment store


mar1u5:
What is the filesize limit for cloud attachments? I think it should be the same :grin:

jancborchardt:
@ChristophWurst postpone to 0.7? :)

ChristophWurst:
Development stalled, so I'm closing this for now. We can always restore the branch/commits if necessary.

jancborchardt:
I moved the todos to an issue so we don’t forget about this feature ;) https://github.com/nextcloud/mail/issues/251

PR #5, opened 2016-08-28
Tipsy tooltip

ChristophWurst:
Finally get rid of all those errors :)

Please review @nextcloud/mail


jancborchardt:
cc @irgendwie we’ll continue here :)

PR #12, opened 2016-08-29
remove localStorage clearing part cause of move to sessionStorage

jancborchardt:
Is this correct @ChristophWurst? I assumed the section is not needed anymore but not sure.


ChristophWurst:
:+1: yep, correct

PR #27, opened 2016-09-02
enable server's app code checker and fix remaining issues

ChristophWurst:
This ensures we don't use any private and/or deprecated APIs any more :rocket: 

cc @nextcloud/mail @nickvergessen


nickvergessen:
:+1:

PR #3, opened 2016-08-28
Create default account automatically

ChristophWurst:
TODO:
- [x] save login password to session if config is available
- [x] create a default account if it does not exist
- [x] update the account password if it changed
- [ ] add unit tests
- [ ] find someone who tests this with LDAP
- [ ] document how to configure the default account

PR #28, opened 2016-09-02
create a dynamic default account

ChristophWurst:
Alternative approach to https://github.com/nextcloud/mail/pull/3

@LukasReschke this one is for you :-)


appinfo/application.php
```diff
@@ -43,8 +49,25 @@ public function __construct(array $urlParams = []) {
 		$container->registerParameter("appName", "mail");
 		$container->registerParameter("userFolder", $container->getServer()->getUserFolder($user));
 		$container->registerParameter("testSmtp", $testSmtp);
-		$container->registerParameter("referrer", isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : null);
+		$container->registerParameter("referrer", isset($_SERVER['HTTP_REFERER']) ? : null);
```
ChristophWurst: - [ ] TODO: undo

lib/db/mailaccountmapper.php
```diff
@@ -62,6 +62,16 @@ public function findByUserId($userId) {
 		return $this->findEntities($sql, $params);
 	}
 
+	public function findByEmail($email, $userId) {
```
ChristophWurst: - [ ] TODO: undo

ChristophWurst:
cc @pierreozoux since you've been waiting for this feature

lib/service/defaultaccount/defaultaccountmanager.php
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	public function getDefaultAccount() {
+		$config = $this->getConfig();
+		if (is_null($config)) {
+			return null;
+		}
+
+		$user = $this->userSession->getUser();
+		$this->logger->info('building default account for user ' . $user->getUID());
+		$password = $this->crypto->encrypt($this->session->get('mail_default_account_password'));
```
ChristophWurst: - [ ] TODO: check if password is set (admins could set the config *after* their users logged in)

ChristophWurst:
Sample config
```
  'app.mail.accounts.default' => [
    'email' => '%USERID%@yourdomain.at'',
    'imapHost' => 'imap.yourserver.com',
    'imapPort' => 993,
    'imapUser' => '%USERID%@yourdomain.at',
    'imapSslMode' => 'ssl',
    'smtpHost' => 'smtp.yourserver.com',
    'smtpPort' => 486,
    'smtpUser' => '%USERID%@yourdomain.at'',
    'smtpSslMode' => 'none',
  ],
```

ChristophWurst:
seems to work. @LukasReschke @jancborchardt @pierreozoux @nextcloud/mail review please :-)

pierreozoux:
Hum... Not sure if related to this specific PR, but, I followed instructions from the README to "compile" this PR to test it. at the following step, I get an error:

```
www-data@43e47166ee64:~/html/apps/mail$ make optimize-js
npm install --production
./node_modules/bower/bin/bower install
./node_modules/requirejs/bin/r.js -o build/build.js

/var/www/html/apps/mail/node_modules/requirejs/bin/r.js:22358
        function addFile(file, fileUrl) {
        ^^^^^^^^
SyntaxError: In strict mode code, functions can only be declared at top level or immediately within another function.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
Makefile:38: recipe for target 'optimize-js' failed
make: *** [optimize-js] Error 8
```

Can you tell me what I can do to fix it? thanks!

```
www-data@43e47166ee64:~/html/apps/mail$ node -v
v0.10.29
www-data@43e47166ee64:~/html/apps/mail$ npm -v
1.4.21
```

ChristophWurst:
Strange. Update to a recent version of node/npm maybe? :see_no_evil: 

I have node v6.5.0 and npm 3.10.6. 

NPM can be updated with ``npm install -g npm``

pierreozoux:
Yes, it is working now, the issue in the README is this line: https://github.com/nextcloud/mail#developer-setup-info

should we point to https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions instead?

ChristophWurst:
> should we point to https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions instead?

Sounds good. If we point to https://nodejs.org/en/download/package-manager/ it's distribution independent

Could you please create a simple PR to fix that? :-)

pierreozoux:
Ok, again, not sure if related to this PR, but I get the following js error:

```
Error: An "el" #app-navigation #app-navigation-accounts must exist in DOM
```

And the lest side bar is empty, no folders are displayed.
https://cloud.pierre-o.fr/index.php/s/GZkjaf9a0H2nD7B

ChristophWurst:
@pierreozoux that looks unrelated, as I did not touch any of that client-side code. Does building of the master branch work for you?

ChristophWurst:
> /var/www/html/apps/mail/node_modules/requirejs/bin/r.js:22358

Hm, that's actually a 3rd-party file. I've never seen something like that and the build also succeeds on travis. Maybe a environment-specific error?

pierreozoux:
Ok, I used node v6, and it is now working \o/

ChristophWurst:
@LukasReschke @jancborchardt mind giving this another test run? :-)

jancborchardt:
Yo @LukasReschke can you enable this for us please? ;)

ChristophWurst:
Be aware that once we have the remember-me functionality back, this way of storing the login password will break. Then the account will magically disappear as soon as the user is logged in again via the remember-me cookie :hankey:

jancborchardt:
Wait, what? :D So this will break anyway?

ChristophWurst:
> Wait, what? :D So this will break anyway?

well, not as long as remember-me is broken :rocket:

ChristophWurst:
> Error: An "el" #app-navigation #app-navigation-accounts must exist in DOM

@Gomez has the same problem on a production system. @pierreozoux which server/core did you use for testing?

pierreozoux:
The way to solve it I guess was to use node v6.
I use server v10.0

pierreozoux:
No I told you, I'm fine now (I think it is the upgrade to node v6 that fixed it)

ChristophWurst:
Actually, it might be a problem with NC9 that we cannot reproduce on NC10.

jancborchardt:
What’s the verdict here now @ChristophWurst @pierreozoux @LukasReschke? Merge, kill, or different approach? :)

ChristophWurst:
I just have to finish https://github.com/nextcloud/server/pull/2044 then we can finally have this feature with NC11 🚀

lib/AppInfo/Application.php
```diff
@@ -30,6 +31,10 @@ class Application extends App {
 	public function __construct(array $urlParams = []) {
 		parent::__construct('mail', $urlParams);
 
+		$this->initializeAppContainer();
+	}
+
+	public function initializeAppContainer() {
```
Quix0r: Really `public`? I don't see it anywhere else in your PR.

ChristophWurst:
Oh, almost forgot to finish this. Now that  nextcloud/server#2044 has been merged I can continue this PR 🚀

ChristophWurst:
rebased and resolved conflicts

ChristophWurst:
While I theoretically could already finish this PR, we'd make the app depend on NC12+. Therefore, I'll wait until it has been released, because then we'll probably support NC12-13 (we're at NC11-12 ATM).

pierreozoux:
Does it mean that if I have NC12, this PR would work? I can't wait to try :)

ChristophWurst:
@pierreozoux exactly :-) Just needs a bit of polishing, but this is more or less ready 🚀

rakshazi:
I wanna try it, too! :)

jancborchardt:
@pierreozoux @rakshazi please go ahead and test it with current Nextcloud server master. :)

lib/AppInfo/Application.php
ChristophWurst: Yup, makes sense. Fixed.

ChristophWurst:
All manual tests succeeded, CI is happy and code coverage increased. I'd say we're ready for a final review 🚀

jancborchardt:
@LukasReschke @pierreozoux @rakshazi please go ahead and test it. :)

lib/Service/AccountService.php
```diff
@@ -43,12 +45,17 @@ class AccountService {
...
 
+	/** @var Manager */
+	private $defaultAccountManager;
+
 	/**
 	 * @param MailAccountMapper $mapper
 	 */
-	public function __construct(MailAccountMapper $mapper, IL10N $l10n) {
+	public function __construct(MailAccountMapper $mapper, IL10N $l10n,
+		Manager $defaultAccountManager) {
```
lcalaresu: Are the `@param` important?
They were not consistent on that file before the PR anyway, so maybe it is not important... but I'm asking coz I am tempted to change some headers on a file I'm changing on another PR :yum:

ChristophWurst: My bad, of course they should be in sync with the actual parameters of the method :-)

ChristophWurst: Although one could argue that the `@param` annotation has no added value when the parameter is type hinted anyway 😉

lcalaresu:
I finally got it working with the `%EMAIL%` in the config file. 😵 
It's easier to catch by reading the code 😝 

Actually, my test user had a different email than what was supposed to be (from the config file perspective). Do not know if it's clear... 

Here is the story of John (my user):

- He put `john@doe.com` in his profile because he prefer using his personal email.

- The administrator of his nextcloud instance wrote the following configuration file:
```
'app.mail.accounts.default' => [
    'email' => '%USERID%@domain.tld',
    'imapHost' => 'imap.domain.tld',
    'imapPort' => 993,
    'imapUser' => '%EMAIL%',
    'imapSslMode' => 'ssl',
    'smtpHost' => 'smtp.domain.tld',
    'smtpPort' => 465,
    'smtpUser' => '%EMAIL%',
    'smtpSslMode' => 'ssl',
],
```
- When he goes to the mail app, the mailbox `john@domain.tld` has been created, but the app is trying to connect to the server using his personal email address and his nextcloud password... John (actually, that was me) is confused: he cannot access his professional mailbox, and cannot even delete it...

Actually, I thought the `%EMAIL%` keyword was just to avoid repeating the email written just above in the config file.

But that let me think: is it safe to rely on the user email? (that could be changed by him at any time by himself).
If the admin chooses the host, the port and the ssl mode, shouldn't he be certain the email will not change over time?

ChristophWurst:
> I finally got it working with the %EMAIL% in the config file. 😵
> It's easier to catch by reading the code 😝

Yeah, there should be more documentation on how to use this new feature.

Regarding the email issue: good catch! I didn't think about that scenario. However, it's the responsibility of the admin to set the default account settings to something that actually makes sense and works. On most instances the %USERID% replacement is all that is actually needed, because you can easily build credentials for a single domain.
Therefore we have two possible paths to take: first, remove this %EMAIL% replacement feature right away. second, leave it for complex setups. I could make sense to use a user-provided mail address, for example on instances where users are connected to the same user back-end, but with different email domains.

lcalaresu:
@ChristophWurst Yeah, that could be convenient for that admin indeed. 
I am just afraid end users would have the power to (silently) break their mail app.

Actually, I don't use LDAP or others on NC. But in that case, can the end user modify his email or is it "forced" by the admin? 
If it cannot be changed, I would say that is indeed not really a problem if the "limitation" is documented for those not using domain users (who will be less likely to use that feature anyway I think).

ChristophWurst:
> Actually, I don't use LDAP or others on NC. But in that case, can the end user modify his email or is it "forced" by the admin?

I don't know.

ChristophWurst:
> If it cannot be changed, I would say that is indeed not really a problem if the "limitation" is documented for those not using domain users (who will be less likely to use that feature anyway I think).

IMO having a centralized user back-end that is both used by your Nextcloud and IMAP server is the only realistic use case for this feature 😉

lcalaresu:
> IMO having a centralized user back-end that is both used by your Nextcloud and IMAP server is the only realistic use case for this feature 😉

I agree. 

Concerning the "Delete account", do you think it could be possible to disable or hide it easily?
Nothing really important here since the user cannot delete the account anyway... But if the button does not work, maybe it could be seen as a bug for some people.  Maybe in another PR...

Everything else worked well for me, it works as expected. 👍 👍

ChristophWurst:
> But if the button does not work, maybe it could be seen as a bug for some people. Maybe in another PR...

Ah yes, we should hide it ofc.

ChristophWurst:
Fixed 🚀

ChristophWurst:
Thanks a bunch for testing and reviewing 😀

lcalaresu:
I did the easy part 😉

fsedarkalex:
Just one question...

first of all it seems working but what should happen if the email password differs from the NC-Password?
Currently endless loading occurs. Do you see any chance to temporarily disable the mailbox (until next login) or even better ask for the password?

ChristophWurst:
> first of all it seems working but what should happen if the email password differs from the NC-Password?
> Currently endless loading occurs. Do you see any chance to temporarily disable the mailbox (until next login) or even better ask for the password?

Yes, a bit of failure tolerance is needed regardless of an automatically created account because any IMAP account can be temporarily unavailable. Luckily, the code in #326 already comes with a bit of tolerance when one of the accounts fails to load. You still see a warning, but other accounts are usable.

sim6:
Please, could you release a version with this changes?

ChristophWurst:
@sim6 sure, as soon as we've got into a more or less stable state again. Currently there are some issues, e.g. with the background synchronization that make your browser freeze in certain circumstances. #326 tackles some of the issues. Any help (bugfixing, testing, issue reporting, reviewing) is very much appreciated.

Extarys:
Since I'm using 3 different domain name, I would like the login to be the email address of the account.
Instead of `    'email' => '%USERID%@domain.tld',` => ` 'email' => '%EMAIL%',`, But it doesn't work. :\

ChristophWurst:
Hi,

please file a new ticket if you have found bugs and do not comment on old and closed pull requests. Thanks!

PR #33, opened 2016-09-03
fix client-side account order

ChristophWurst:
fixes the following bug
* configure first acount
* configure second account

Then the account order was: account1, unified account, account2

Now it is: unified account, account1, account2

cc @nextcloud/mail


tahaalibra:
Working Ok :+1:

jancborchardt:
Nice :+1:

ChristophWurst:
@jancborchardt could you please add @tahaalibra to the NC org?

jancborchardt:
Done, also added him to the Mail team :)

PR #34, opened 2016-09-03
show a loading spinner while loading folders

ChristophWurst:
fixes the bug where no loading spinner was shown while switching to a folder that has not been loaded yet and therefore no cached version is shown. That was especially annoying on mobile browsers.

@jancborchardt  I think you noticed that too

cc @nextcloud/mail

### DIFF
--- a/js/templates/loading.html
+++ b/js/templates/loading.html
@@ -3,4 +3,6 @@
 	<a class="icon-loading"></a>
 	<h2>{{{ hint }}}</h2>
 </div>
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@jancborchardt could you please test this fix? Without it we are currently lacking many loading spinners (e.g. loading the message content shows a blank element)

jancborchardt:
Nice, works again :+1: :)

PR #39, opened 2016-09-04
use button for the 'New message' button and fix its styling

### LINKED ISSUE #38: Make Contacts & Mail Buttons look better and more the same
https://github.com/owncloud/contacts/pull/493 moved the "Add contact"-button to the sidebar - which is nice. Unfortunately - it is a bit different to the Mail-apps one 😉 

- [x] I think the solution of the mail-app has a better "+" because it is a bit bigger. (Done with https://github.com/owncloud/contacts/pull/511)
- [ ] The contacts button has a better button, because it fits better in the sidebar.

Maybe we can take the positive aspects from every button-style 😜 

Mail:
mail-png

Contacts:
contacts-png

cc @jancborchardt @ChristophWurst @skjnldsv

ChristophWurst:
fixes #38

Before
bildschirmfoto von 2016-09-04 18-16-26
After
bildschirmfoto von 2016-09-04 18-16-33

@mar1u5 @jancborchardt @Gomez review please

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -391,7 +391,6 @@
 
 /* editor */
 #mail-new-message-fixed {
-	width: 241px !important;
 	z-index: 111; /* navigation menu is 110 ;-) */
 	background-color: rgba(255, 255, 255, .75);
 	padding: 10px;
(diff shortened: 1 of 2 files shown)


MariusBluem:
👍

PR #47, opened 2016-09-06
enable server-side cache by default

ChristophWurst:
I have had it enabled in production for ages and have never seen any issues with it. Therefore I'd like to enable it for everyone, by default :rocket:

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,13 +110,6 @@ You can enable IMAP and SMTP backend logging. A horde_imap.log for IMAP and hord
 'app.mail.smtplog.enabled' => true
 ```
 
-### Server-side caching
-Mailbox messages and accounts can be cached on the Nextcloud server to reduce mail server load:
-This requires a valid memcache to be configured
-```php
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@jancborchardt @Gomez @irgendwie @pierreozoux @nextcloud/mail did you guys use the cache?

Gomez:
Yes, works here without problem. 

We should add at least docu for debugging. Caching needs to be disabled if you run with debug enabled.

ChristophWurst:
> We should add at least docu for debugging. Caching needs to be disabled if you run with debug enabled.

Well, devs can still use that option to disable it and they'll find that in the code if they know what they are doing. I don't think we need to document that.

irgendwie:
Wasn't aware of that feature, going to use it from now on :wink:
I'll report issues, if there are any!

Gomez:
Ok.
LGTM

zeugmatis:
Just enabled it here too, working great thus far.

PR #49, opened 2016-09-06
show text too when a loading spinner is used

ChristophWurst:
@jancborchardt how about providing feedback about *what* we are loading?

As a user, I'd like to know what the software does. This makes the application state more clear. However, this would be different to other Nextcloud UIs (e.g. files and other community apps)…

### DIFF
--- a/js/controller/foldercontroller.js
+++ b/js/controller/foldercontroller.js
@@ -125,7 +125,9 @@ define(function(require) {
 	 */
 	function showFolder(account, folder) {
 		Radio.ui.trigger('search:set', '');
-		Radio.ui.trigger('content:loading');
+		Radio.ui.trigger('content:loading', t('mail', 'Loading {folder}', {
+			folder: folder.get('name')
+		}));
 		loadFolderMessages(account, folder, false);
 
 		// Save current folder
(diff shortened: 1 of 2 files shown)


jancborchardt:
Nice! The one problem I see with it is when mail subjects are shown. Because they wrap, or are very long, etc.

Instead of showing them underneath the loading spinner, we should show the mail heading already on the top because we have all the info (subject for sure, sender & recipient?) and the spinner below centered.

jancborchardt:
Also thought about this again and not sure if it’s necessary showing the text in the case of loading folders. I would only do the enhancement of showing the mail header with subject already when it’s loading.

ChristophWurst:
> Also thought about this again and not sure if it’s necessary showing the text in the case of loading folders

I thought about it again, too. On mobile, you have only one spinner while it loads the folder. Unlike on desktop, you can't see which folder is active and therefore which folder is being loaded. Therefore I'm in favor of showing `loading xxx` instead of only the spinner.

For the message spinner I agree, we can show the detailed view already and insert the body later. However, that needs bigger changes. Should I remove it from this PR?

ChristophWurst:
Removed the text when loading the message, now it only shows 'Loading Inbox'

irgendwie:
LGTM :+1:

Issue #50, opened 2016-09-07
Error while loading messages


MZellhofer:
### Steps to reproduce
1. create account
2. open account (opens automatically)

### Expected behaviour
Load messages of inbox folder

### Actual behaviour
Shows spinner and error "Error while loading messages". Since the app can load all the inbox folders, I would say that auth is okay. Mailserver is on the same host - I tried localhost as well as full domain names-

### Mail app

**Mail app version:** 0.5.3

**Mailserver or service:** imap

**Transport security - IMAP:** SSL

**Transport security - SMTP:** STARTTLS

**Number of accounts:** 1

### Server configuration
**Operating system:** Ubuntu

**Web server:** Apache 2.2.14

**Database:** MySQL 5.1.73

**PHP version:** 5.5.19

**Version:** 10.0.0

**Updated from an older version or fresh install:** fresh install

#### Client configuration
**Browser:** Chrome

**Operating system:** Windows 10

##### Web server error log
```
[Wed Sep 07 14:28:53 2016] [crit] [client MY_IP_ADDRESS] (13)Permission denied: PATH_TO_NEXTCLOUD/nextcloud/data/.htaccess pcfg_openfile: unable to check htaccess file, ensure it is readable
```

##### Server log (data/nextcloud.log)
<details>
```
{"reqId":"V9AFGX8AAQEAABSjhA4AABAv","remoteAddr":"MY_IP_ADDRESS","app":"mail","message":"Setting up manually configured account","level":1,"time":"2016-09-07T12:16:25+00:00","method":"POST","url":"\/index.php\/apps\/mail\/accounts","user":"myuser"}
{"reqId":"V9AFGX8AAQEAABSjhA4AABAv","remoteAddr":"MY_IP_ADDRESS","app":"mail","message":"Connecting to account myuser@mydomain.tld","level":0,"time":"2016-09-07T12:16:25+00:00","method":"POST","url":"\/index.php\/apps\/mail\/accounts","user":"myuser"}
{"reqId":"V9AFGX8AAQEAABSjhA4AABAv","remoteAddr":"MY_IP_ADDRESS","app":"mail","message":"account created 8","level":0,"time":"2016-09-07T12:16:25+00:00","method":"POST","url":"\/index.php\/apps\/mail\/accounts","user":"myuser"}
[...]

PR #51, opened 2016-09-07
fallback for incorrectly advertised SORT capability

MZellhofer:
See https://github.com/nextcloud/mail/issues/50: Mailserver might advertise SORT capability but internally reject SORT. In this case try again without SORT.

### DIFF
--- a/lib/mailbox.php
+++ b/lib/mailbox.php
@@ -11,6 +11,7 @@
  * @author Thomas I <thomas@oatr.be>
  * @author Thomas Mueller <thomas.mueller@tmit.eu>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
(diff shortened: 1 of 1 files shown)


ChristophWurst:
thanks for fixing this so quickly @MZellhofer :rocket: 

@zeugmatis could you please test this patch with your setup, too? Thanks.

ChristophWurst:
@MZellhofer is your git configured correctly? Apparently GitHub shows different avatars for your user and for your commits

ChristophWurst:
Oh, you commited with a different email address

MZellhofer:
Ah damn - different Notebook tonight. Do you want me to change the email address in the commit?

zeugmatis:
Tested and confirmed to fix here!  👍

ChristophWurst:
:+1: thanks!

PR #54, opened 2016-09-10
Making IMAP and SMTP timeouts configurable

llamallama:
I was testing mail with GoDaddy as the mail host. Its SMTP server doesn't respond until long after the 2 seconds currently hard-coded into Nextcloud mail. This was resulted in Nextcloud mail timing out on the SMTP setup phase.

I decided rather than changing a hard-coded value that I should make it configurable instead.

### DIFF
--- a/lib/account.php
+++ b/lib/account.php
@@ -139,7 +139,7 @@ class Account implements IAccount {
 				'hostspec' => $host,
 				'port' => $port,
 				'secure' => $ssl_mode,
-				'timeout' => 20,
+				'timeout' => $this->config->getSystemValue('app.mail.imap.timeout', 20),
 			];
 			if ($this->config->getSystemValue('app.mail.imaplog.enabled', false)) {
 				$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde_imap.log';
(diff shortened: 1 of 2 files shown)


lib/account.php
```diff
@@ -139,7 +139,7 @@ public function getImapConnection() {
 				'hostspec' => $host,
 				'port' => $port,
 				'secure' => $ssl_mode,
-				'timeout' => 20,
+				'timeout' => $this->config->getSystemValue('app.mail.imap.timeout', 20),
```
ChristophWurst: it might make sense to cast to int here, like `'timeout' => (int) $this->config->getSystemValue('app.mail.imap.timeout', 20)`

ChristophWurst:
Hi @llamallama,

thanks a lot for your patch! Although it might be a edge case, I'm okay with this new configurable timeout. The only problem I see is that once it' set to a higher value, connections to *any* server will get a higher timeout, but I guess that's okay

ChristophWurst:
Btw, you commited with an email address that is not associated with your GitHub account. Was that inteded? Do you want to fix that?

llamallama:
Good call on casting to int first. I will do that and update the request. Thanks for catching the email address. I added it to my account.

llamallama:
Regarding it being an edge case, that may be so but that will be true for anyone that has their mail in GoDaddy which is still a lot of people.

llamallama:
OK. Cast to int is done.

ChristophWurst:
:+1: patch looks good. Thank you very much

llamallama:
Thanks for accepting it.

Issue #56, opened 2016-09-13
Replace loading-view with error-view if the content can not be loaded


ChristophWurst:
Currently we show an ugly yellow notification on the top of the page whenever an error occurs while loading a view's content, e.g. loading the list of messages or the message body.

Instead, I'd like to try using something similar to the "No search result" or "No messages in this folder" view: replacing the loading view with an "An error occured while xxx" or "Could not load xxx" view and get rid of the yellow notifications.

We could even go one step further and put a "Retry" button on that view, where the user can re-trigger loading the content. I think that would be nice UI and UX-wise.

@nextcloud/designers @nextcloud/mail @jancborchardt @MariusBluem let me know what you think :-)

ChristophWurst:
Another argument for it: notifications vanish after 7 seconds, while the error-view would stay in place until the user explicitly acts again (e.g. clicking on another folder)

te-online:
Also, in the same strike, I would love to see the auto-reload when an error happened redesigned. Maybe it should try to fetch the page in the background or do some other testing of the connection. And if that's not possible it should display a retry button. With the current solution, we auto-remove the precious loaded content from the screen when the user is offline for longer than 5 seconds.

ChristophWurst:
@te-online good catch. That annoys me too, but AFAIK that is triggered by the server or the notifications app. There's an issue somewhere…

ChristophWurst:
here it is: https://github.com/nextcloud/server/issues/992

te-online:
Yeah, sorry, that is a system wide behaviour I guess...

PR #57, opened 2016-09-13
show error-view instead of yellow notification on content loading errors

ChristophWurst:
implements part 1 of https://github.com/nextcloud/mail/issues/56 – the error view.

Examples:
bildschirmfoto von 2016-09-13 19-30-49
bildschirmfoto von 2016-09-13 19-30-03

@te-online @skjnldsv @jancborchardt how about a nice error-logo above the error message to be consistent with the *no messages in this folder* and *no search results* view?

### DIFF
--- a/js/util/errormessagefactory.js
+++ b/js/util/errormessagefactory.js
@@ -1,0 +1,79 @@
+/**
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+define(function(require) {
+	'use strict';
+
+	var smileys = [
+		':-(',
+		':-/',
+		':-\\',
+		':-|',
+		':\'-(',
+		':\'-/',
+		':\'-\\',
+		':\'-|'
+	];
+
+	function getRandomSmiley() {
+		return smileys[Math.floor(Math.random() * smileys.length)]
+	}
+
+	/**
+	 * @param {Folder} folder
+	 * @returns {string}
+	 */
+	function getRandomFolderErrorMessage(folder) {
+		var folderName = folder.get('name');
+		var rawTexts = [
+			t('mail', 'Could not load {tag}{name}{endtag}', {
+				name: folderName
+			}),
+			t('mail', 'We couldn’t load {tag}{name}{endtag}', {
+				name: folderName
+			}),
+			t('mail', 'There was a problem loading {tag}{name}{endtag}', {
+				name: folderName
+			})
+		];
+		var texts = _.map(rawTexts, function(text) {
+			return text.replace('{tag}', '<strong>').replace('{endtag}', '</strong>');
+		});
+		var text = texts[Math.floor(Math.random() * texts.length)]
+		return text + ' ' + getRandomSmiley();
+	}
+
+	/**
+	 * @returns {string}
+	 */
+	function getRandomMessageErrorMessage() {
+		var texts = [
+			t('mail', 'We couldn’t load your message'),
+			t('mail', 'Unable to load the desired message'),
+			t('mail', 'There was a problem loading the message')
+		];
+		var text = texts[Math.floor(Math.random() * texts.length)]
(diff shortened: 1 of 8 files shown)


ChristophWurst:
Maybe we can also find a better and shorter wording for "Error while loading the selected message" :wink:

skjnldsv:
"Unable to load the desired message" ? :)
Nice idea btw! 👍

jancborchardt:
For the individual folders, I would use the folder-relevant icon for the error icon. For individual mails or when there is no specific icon I’d say let’s use the mail-envelope-icon.

The messages could be a bit improved to this to be less sterile and also a bit fun and helpful. First line is h2 in emptycontent, and second line is the `<p>`

> We couldn’t load the **Inbox** :( (the **Sent mails**, the **Trash** etc.)
> Try reloading the page

> **Vacation planning** isn’t loading :\
> Click on another folder and come back later, maybe?

> There was a problem loading the message :(
> Maybe you deleted it on your phone?

Yes, with the :( smiley. Maybe even cycle randomly between :( and :\ and :/ and ;( – that makes the error message at least a bit fun. ;)

ChristophWurst:
@jancborchardt something like this?
bildschirmfoto von 2016-09-14 21-09-46

We'll need a generic icon for non-special-use folders. Could could make one, please?

ChristophWurst:
> For individual mails or when there is no specific icon I’d say let’s use the mail-envelope-icon.

Ah, didn't read that before…

ChristophWurst:
oh, we have an 'icon-folder' too. Hmm…
bildschirmfoto von 2016-09-14 21-13-26

te-online:
But give the smileys noses, please! 😃
________________________________
Von: Jan-Christoph Borchardt<mailto:notifications@github.com>
Gesendet: ‎14.‎09.‎2016 19:40
An: nextcloud/mail<mailto:mail@noreply.github.com>
Cc: Thomas Ebert<mailto:thomas.ebert@te-online.net>; Mention<mailto:mention@noreply.github.com>
Betreff: Re: [nextcloud/mail] show error-view instead of yellow notification on content loading errors (#57)


For the individual folders, I would use the folder-relevant icon for the error icon. For individual mails or when there is no specific icon I’d say let’s use the mail-envelope-icon.

The messages could be a bit improved to this to be less sterile and also a bit fun and helpful. First line is h2 in emptycontent, and second line is the <p>

We couldn’t load the Inbox :( (the Sent mails, the Trash etc.)
Try reloading the page

Vacation planning isn’t loading :\
Click on another folder and come back later, maybe?

There was a problem loading the message :(
Maybe you deleted it on your phone?

Yes, with the :( smiley. Maybe even cycle randomly between :( and :\ and :/ and ;( – that makes the error message at least a bit fun. ;)

jancborchardt:
@ChristophWurst, yeah, basically like that! :) Is it not possible to bold the foldername instead of using quotes? That would look waaay better.

Also let’s please use the mail icon instead of the folder icon. We do not use the folder icon anywhere in the mail app – it belongs to Files. In fact other mail apps like Gmail use the concept of »Tags« rather than folders so let’s use a generic one.

And below that sentence, put a `<p>` with the second-line text I showed above. Or do you think it’s too much?

ChristophWurst:
> Is it not possible to bold the foldername instead of using quotes? That would look waaay better.

tried that with search already but failed. Maybe someone can enhance that later

ChristophWurst:
@LukasReschke ^

js/util/errormessagefactory.js
```diff
@@ -0,0 +1,82 @@
...
+			}),
+			t('mail', 'We couldn’t load {tag}{name}{endtag}', {
+				name: folderName
+			}),
+			t('mail', 'There was a problem loading {tag}{name}{endtag}', {
+				name: folderName
+			})
+		];
+		var texts = _.map(rawTexts, function(text) {
+			return text.replace('{tag}', '<strong>').replace('{endtag}', '</strong>');
```
skjnldsv: Are you trying to bold the folder name here?

ChristophWurst: yes

skjnldsv: see https://github.com/nextcloud/mail/pull/57#issuecomment-247383309 :)
If the strong is correctly outputted on the html and still isn't displayed bold, this could be because the strong css value "font-weight" has been reset. If so just use a span and set it in css. Should be enough! ;)

jancborchardt: Riight, we need to add in our CSS:
`strong { font-weight: 600; }`

@skjnldsv great to see you back here by the way! :)

jancborchardt: Fix at https://github.com/nextcloud/server/pull/1423

skjnldsv: Glad to see you both! :)

js/views/navigation-accounts.js
```diff
@@ -49,7 +49,9 @@ define(function(require) {
 				});
 			});
 
-			folder.set('active', true);
```
ChristophWurst: small bugfix: if the first folder fails loading, `folder` was undefined here and therefore an error was shown in the js console.

jancborchardt:
NICE, looks very good ;D :+1:

PR #58, opened 2016-09-13
Re-add message on the client if it cannot be deleted

ChristophWurst:
To prevent inconstencies on the client, we should re-add a deleted message if it cannot be deleted.

Example: you're disconnected from your nc server temporarily. With this PR you can delete the message later. On master, the message disappeared from the list, but will come back with the next reload.

TODO
- [x] reset folder properties (e.g. total, unseen counter)

### DIFF
--- a/js/service/messageservice.js
+++ b/js/service/messageservice.js
@@ -384,6 +384,9 @@ define(function(require) {
 				defer.resolve();
 			},
 			error: function() {
+				// Add the message to the collection again
+				folder.addMessage(message);
+
 				defer.reject();
 			}
 		});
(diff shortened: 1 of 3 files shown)


ChristophWurst:
@nextcloud/mail review please

ChristophWurst:
> Good! Does it cover every type of deleting error or just an ajax failure?
> 
> If the imap server refuse the deletion for example?

Well, it covers any ajax error, regardless of the reason and HTTP status code. Do IMAP servers refuse message deletion?

skjnldsv:
Well, that's a good question. I guess it could?
Never occurred to me. Just a random thought ;)

ChristophWurst:
> Never occurred to me. Just a random thought ;)

hehe to me neither, so let's ignore that case for now 🙈

PR #60, opened 2016-09-14
Ignore js dependencies and tests from scrutinizer

ChristophWurst:


ChristophWurst:
> Test Coverage has increased to 52% (+1%). 

@nextcloud/mail easy review :wink:

tahaalibra:
:+1:  can we also exclude node_modules?

ChristophWurst:
> can we also exclude node_modules?

good catch, but I *think* scrutinizer doesn't install npm dependencies anyway, so that shouldn't be a problem

PR #61, opened 2016-09-14
We do not have nightly builds at the moment

ChristophWurst:
We should either remove or fix this.

Generally, I would prefer to have a solution that is maintainable by nc mail devs.

cc @enoch85 @nextcloud/mail

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,16 +76,6 @@ make install-composer-deps
 make optimize-js
 ```
 
-### Nightly builds
-
-Instead of setting everything up manually, you can just [download the nightly builds](https://en0ch.se/index.php/s/3v5J3XXeQCCp73p) instead. These builds are updated every 12 hours - 00:00 (12:00 AM) and 12:00 (12:00 PM) UTC+0200, and are pre-configured with all the needed dependencies.
-
-1. Download
-2. Extract the tar archive to 'path-to-nextcloud/apps'
-3. Navigate to »Apps«, choose the category »Productivity«, find the Mail app and enable it.
-
-The nightly builds are provided by [Daniel Hansson](https://github.com/enoch85)
-
 ### Resetting the app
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Okay, I played a bit with Travis' API and we can easily trigger a "special" build by a cronjob script.

```bash
body='{
"request": {
  "message": "Nightly build",
  "branch":"master",
  "config": {
    "env": {
      "matrix": ["DB=sqlite CORE_BRANCH=master TEST_JS=FALSE PHP_COVERAGE=FALSE PACKAGE=TRUE NIGHTLY=TRUE"]
    },
    "php": ["7"],
    "matrix": {
        "fast_finish": false
    }
  }
}}'

curl -s -X POST \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Travis-API-Version: 3" \
  -H "Authorization: token xxxxxxx" \
  -d "$body" \
  https://api.travis-ci.org/repo/Nextcloud%2Fmail/requests
```

All we have to do is upload the artifact somewhere. Do you have a FTP server that we could use for that @Gomez? Let's hack something in the next week ;-)

I'd prefer such a solution over the one we had in the past because then everybody can see the nightly build results on travis-ci.org -> https://travis-ci.org/nextcloud/mail/builds/159929194 :rocket: 

The next step would be to also submit the nightly to the shiny new Nextcloud app store.

ChristophWurst:
Relevant app store docu: https://nextcloudappstore.readthedocs.io/en/latest/restapi.html#publish-a-new-app-release – that one-liner can be added to the travis file. Should be doable easily

irgendwie:
Do you want to fix the new nightly builds in a different PR or here?
Anyway the Changes to the README are Ok! :+1:

ChristophWurst:
another PR. We'll first have to find a server we can upload the zip to.

enoch85:
missed this :/

PR #63, opened 2016-09-14
remove misleading version eye badges

ChristophWurst:
remove version eye badges showing oc mail's outdated dependencies

At some point we should properly set this up again and add badges for this project.


jancborchardt:
Good :+1:

PR #76, opened 2016-09-19
Add JS Coverage

Gomez:


ChristophWurst:
As discussed, this has been superseded by https://github.com/nextcloud/mail/pull/78. Let's only print it to travis. We can always add a fancy GitHub integration later

PR #79, opened 2016-09-19
fix ProxyController's response

ChristophWurst:
IResponse cannot be converted to string automatically, therefore we need to get the
body before we create the response object.

This fixes the bug where images of HTML messages do not load.

cc @Gomez @jancborchardt @eppfel @hechi this *should* be the last blocking issue for 0.6 :rocket:


Gomez:
LGTM

PR #82, opened 2016-09-19
enable coveralls coverage reports

ChristophWurst:


ChristophWurst:
comments have been disabled

ChristophWurst:
the coverage level seems a tad too high, but we'll figure out what's causing that.

cc @Gomez @georgehrke @nextcloud/mail

Gomez:
:+1:

PR #83, opened 2016-09-20
[WIP] horde sync WIP

ChristophWurst:
:construction: WIP :construction: 

Some experiments with horde sync to speed up checking for new messages and background checks.


ChristophWurst:
experiment over 👷

PR #85, opened 2016-09-20
Version bump 0.6.0

ChristophWurst:
cc @Gomez @jancborchardt @tahaalibra @juliushaertl please take a look

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,12 +100,10 @@ script:
 
 matrix:
   include:
-    - php: 7.0
+    - php: 5.4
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -100,12 +100,10 @@ script:
 
 matrix:
   include:
-    - php: 7.0
+    - php: 5.4
       env: "DB=mysql TEST_JS=TRUE"
```
juliushaertl: We need CORE_BRANCH=stable10 here as well, as 5.4 was dropped on master

PR #88, opened 2016-09-20
Test calendarview

Gomez:


js/tests/views_calendarview_spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+	describe('CalendarView', function () {
+
+		var calendarview;
+
+		beforeEach(function () {
+			calendarview = new CalendarView({});
+		});
+
+		describe('Rendering', function () {
```
ChristophWurst: do we really need the second `describe` block? I guess one *describe* and one *it* should be enough :wink:

ChristophWurst:
test coverage decreases as more modules are loaded, but that's okay I guess

js/tests/views_calendarview_spec.js
Gomez: Ok, removed

ChristophWurst:
:+1:

PR #93, opened 2016-09-20
rework account-folders relation

ChristophWurst:
follow-up PR of #9 to keep things consistent and tested.

cc @Gomez @tahaalibra


Gomez:
UI tested and no error appeared. 

:+1:

PR #90, opened 2016-09-20
Moving mails

ChristophWurst:
TODO:
- [ ] unified inbox hacks
- [x] special folders hack (e.g. favourites search folder)
- [x] visual drop feedback
- [x] add drag/drop scopes to prevent cross-account moves
- [ ] message removal animation
- [ ] subfolders are broken

fixes https://github.com/nextcloud/mail/issues/24 :money_with_wings: :moneybag: :dollar: :dancer:

PR #97, opened 2016-09-21
Json serializable

ChristophWurst:
small cleanup needed for a server-side change for #90 (i need to add the account ID and folder ID to the message json somehow)

@Gomez @hechi review please ;-)


Gomez:
:+1: Quick test

PR #98, opened 2016-09-21
Allow navigationg from one account settings page to another

ChristophWurst:
cc @hechi, good catch ;-)


ChristophWurst:
Steps to reproduce:
1. open settings of one account
2. try to open settings of another account

Before the first settings page was still shown, now we switch correctly

hechi:
:) thx :+1:  and thx for fixing it

ChristophWurst:
:+1:

PR #107, opened 2016-09-22
width changed for composer fields

RucajDenisa:
It looks better instead of being too streched.


jancborchardt:
Yeah, looks much better! :) Down to a readable width rather than so big. :+1:

ChristophWurst:
nice, thanks a lot @RucajDenisa :rocket:

PR #116, opened 2016-09-23
We must also drop the oc_mail_aliases table.

matiasdelellis:
Document was lacking since the feature was implemented.


ChristophWurst:
:+1: thank you very much

PR #118, opened 2016-09-23
Fix wrong method being using to add message to a folder

ChristophWurst:
This bug prevent the background checker from adding new incoming messages to the folder's message list.

cc @nextcloud/mail @hechi


ChristophWurst:
For reference, here's the method that *does* exist: https://github.com/nextcloud/mail/blob/679773116e4e67051dd2de77abb8e296457c60d5/js/models/folder.js#L41

skjnldsv:
So was that a typo or planned?

Anyhow: 👍 :)

ChristophWurst:
I guess I didn't test carefully enough when I touched that code the last time 🙈

Issue #109, opened 2016-09-23
0.6-RC1 bugs


ChristophWurst:
This is a meta-issue to keep track of the testing results of the first release candidate (https://github.com/nextcloud/mail/releases/tag/v0.6.0-RC1).

ref https://help.nextcloud.com/t/mail-0-6-rc1-ready-for-testing/3503/1

ChristophWurst:
@plastikschnitzer could you please post your test results here? Thanks :-)

ChristophWurst:
also @nextcloud/mail please update to the RC and give it a test run

ChristophWurst:
(unconfirmed) autoconfig has issues with some providers. cc @Gomez @matiasdelellis

ChristophWurst:
Minor, won't fix in this release:

* Navigation from account settings to new message composer is impossible #10

pierreozoux:
Now we are speaking about autoconfig, never managed to make it working (It is working in Thunderbird though).

plastikschnitzer:
Dear all,
I am following and using this app since version 0.2 and I really love to see the progress because I think cloud is not just about files, its also about mails – many informations are and will always be exchanged by mail and mails are just everywhere (like it or not ;-) ) – to have it right in the cloud is great, I really like to contribute here.

I tested Mail 0.6.0 on NextCloud 10 with multiple mailboxes set up. I use Kubuntu 16.04.1 with Firefox 48 to access the Cloud. 

I am commenting right below the release notes issue list:

> - Alias support
>   [#1523](https://github.com/owncloud/mail/pull/1523) @tahaalibra


Setting up mail alias works, also sending email with this alias. (great feature!!!)
If you add an alias and want to click on "new message" is not possible. All mailbox links below are working. (I think @ChristophWurst mentioned it above)

> - New incoming messages are prefetched
>   [#1631](https://github.com/owncloud/mail/pull/1631) @ChristophWurst

Seems to work, but not always on my side. I can not reconstruct it to 100%, sometimes mails appear immediately by switching in between accounts or folders, sometimes they are loaded again. Problem: If I delete a message, it seems to remain in cache sometimes and this will result in an error message, if it is the first message in a list of mails.

> - Custom app folder support
>   [#1627](https://github.com/owncloud/mail/pull/1627) @juliushaertl

Not tested.

> - Improved search
>   [#1609](https://github.com/owncloud/mail/pull/1609) @ChristophWurst

[...]

PR #120, opened 2016-09-26
download attachments in new tab to prevent canceled connections

ChristophWurst:
fixes https://github.com/nextcloud/mail/issues/109#issuecomment-249362308

cc @plastikschnitzer @nextcloud/mail


skjnldsv:
👍

PR #126, opened 2016-09-27
Test 'app' module to load all modules for correct code coverage

ChristophWurst:
Although only a few lines have changes, I expect this PR to have a huge effect on code coverage. Since the module `mail` can be seen as the root of all js modules, we basically load *all* modules the application uses. Therefore, the js code coverage should finally show the real percentage of tested code.

We can later extend this test case to have a full integration test of the app start. Currently I mocked the initialization code and discarded the emitted events to keep things simple.

cc @Gomez @nextcloud/mail


build/build.js
```diff
@@ -14,7 +14,6 @@
 	name: 'app',
 	out: '../js/mail.min.js',
 	insertRequire: [
-		'app',
-		'notification'
```
ChristophWurst: cc @skjnldsv: I fixed another occurrence of that bloody module. I knew it was loaded twice, but I couldn't find it when I fix the other one…

js/init.js
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+define(function(require) {
+	var App = require('app');
+
+	$(function() {
+		// Start app when the page is ready
+		console.log('starting Mail…');
```
ChristophWurst: before someone nitpicks this as debug code – I've added in on purpose because it could be handy to debug js errors in the future if someone reports a weird client error. We'd then know whether the app tried to start or not.

ChristophWurst:
Hm. No coverage reports 28 instead of 15 files… Not sure if we actually caught all modules 🤔

Still, it's an improvement…

build/build.js
skjnldsv: God dammit! You again Mr notification! I was expecting you!

js/init.js
skjnldsv: Good call.

jancborchardt: Nitpickingly corrected the writing style to »Starting Mail …« ;D

ChristophWurst: really? ;-)

skjnldsv: Ahahahah

ChristophWurst:
@jancborchardt @skjnldsv any issues detected? Can we merge this PR?

skjnldsv:
Let me check a moment.

jancborchardt:
Looks all good :+1:

jancborchardt:
I guess coveralls decreasing is separte

skjnldsv:
Haaaan, @ChristophWurst is doing all the hard work and you get the pleasure to merge! :O

jancborchardt:
@skjnldsv :DD you can also press the button if you like. ;) It only seems that the plugin doesn’t register :+1: when it was written inside of a review comment (which is good I guess).

PR #134, opened 2016-10-03
do not unnecessarily re-render views

ChristophWurst:
This improves performance and is necessary for #81.

@nextcloud/mail @jancborchardt please test this patch. As far as I tested only two views needed adjustment, all other either don't have any model changes that need propagation or are collection views, which are automatically rendered by Marionette already.

I also removed the `TemplateCache` *hack* because we always create a handlebars template manually now.


Gomez:
:+1:

PR #135, opened 2016-10-06
fix buggy wheel/scroll event handling that caused unexpected reloads

ChristophWurst:
fixes https://github.com/nextcloud/mail/issues/109#issuecomment-249789045

Before the list of messages was sometimes refreshed although because the scroll/wheel events where interpreted wrongly. This should finally fix it.

Tested with FF and Chromium.

cc @plastikschnitzer @jancborchardt @Gomez @irgendwie @skjnldsv

### DIFF
--- a/js/views/messagesview.js
+++ b/js/views/messagesview.js
@@ -41,7 +41,8 @@ define(function(require) {
 		loadingMore: false,
 		reloaded: false,
 		events: {
-			wheel: 'onScroll'
+			DOMMouseScroll: 'onWheel',
+			mousewheel: 'onWheel'
 		},
 		collectionEvents: {
 			update: 'render'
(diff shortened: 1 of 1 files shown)


js/views/messagesview.js
```diff
@@ -194,6 +195,15 @@ define(function(require) {
 				return;
 			}
 		},
+		onWheel: function(event) {
+			if (event.originalEvent.wheelDelta && event.originalEvent.wheelDelta > 0) {
+				// Scrolling up in non-FF browsers
+				this.onScroll();
+			} else if (event.originalEvent.detail && event.originalEvent.detail < 0) {
+				// Scrolling up in FF
+				this.onScroll();
```
skjnldsv: Is this supposed to be the same lines (201 & 204)?
Because I feel like whatever if we got, the result is always `this.onScroll();` So I don't see why this is needed! :)

skjnldsv: Okay, I misread, makes sense!

skjnldsv:
👍

skjnldsv:
@ChristophWurst merging faster than his shadow!

PR #137, opened 2016-10-06
Add avatars and placeholders to the recipient autocomplete widget

ChristophWurst:
cc @jancborchardt @eppfel since we talked about that at the conf

would be great if a CSS ninja could fix the design a bit (e.g. some padding, fix vertical alignment and remove ugly background color/image). @skjnldsv maybe 😉 

bildschirmfoto von 2016-10-06 22-09-32

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -470,6 +470,9 @@ input.bcc {
 	border-top: none;
 	border-radius: 0;
 }
+.mail-recipient-autocomplete .avatar {
+	display: inline-block;
+}
 textarea.message-body {
 	margin: 0;
 	border-top: none;
--- a/css/mail.css
+++ b/css/mail.css
@@ -389,6 +389,16 @@
 	right: -10px;
 }
 
+/* autocomplete list */
+.ui-autocomplete .ui-menu-item a.mail-recipient-autocomplete {
+	display: flex;
+	align-content: center;
+	align-items: center;
+}
+.ui-autocomplete .ui-menu-item a.mail-recipient-autocomplete span {
+	margin-left: 10px;
+}
+
 /* editor */
 #mail-new-message-fixed {
 	z-index: 111; /* navigation menu is 110 ;-) */
@@ -469,9 +479,6 @@ input.bcc {
 	margin: 0;
 	border-top: none;
 	border-radius: 0;
-}
-.mail-recipient-autocomplete .avatar {
-	display: inline-block;
 }
 textarea.message-body {
 	margin: 0;
--- a/js/views/composerview.js
+++ b/js/views/composerview.js
@@ -488,7 +488,7 @@ define(function(require) {
 						$row.append($placeholder);
 					}
 
-					$row.append($('<span>').html(item.value));
+					$row.append($('<span>').text(item.value));
 
 					$item.append($row);
 					$item.appendTo($ul);
--- a/js/views/composerview.js
+++ b/js/views/composerview.js
@@ -473,14 +473,15 @@ define(function(require) {
 						this.value = terms.join(', ');
 						return false;
 					}
-				}).data('ui-autocomplete')._renderItem = function ($ul, item) {
+				}).data('ui-autocomplete')._renderItem = function($ul, item) {
 					var $item = $('<li/>');
 					var $row = $('<a/>');
 
 					$row.addClass('mail-recipient-autocomplete');
 
-					if (prevUID === item['id']) {
-						var $placeholder = $('<div/>');
+					var $placeholder;
+					if (prevUID === item.id) {
+						$placeholder = $('<div/>');
 						$placeholder.addClass('avatar');
 						$row.append($placeholder);
 					} else if (item.photo && item.photo !== null) {
@@ -491,13 +492,13 @@ define(function(require) {
 						$avatar.attr('src', item.photo);
 						$row.append($avatar);
 					} else {
-						var $placeholder = $('<div/>');
+						$placeholder = $('<div/>');
 						$placeholder.imageplaceholder(item.value);
 						$placeholder.addClass('avatar');
 						$row.append($placeholder);
 					}
 
-					prevUID = item['id'];
+					prevUID = item.id;
 
 					$row.append($('<span>').text(item.value));
 
--- a/js/views/composerview.js
+++ b/js/views/composerview.js
@@ -479,7 +479,7 @@ define(function(require) {
 
 					$row.addClass('mail-recipient-autocomplete');
 
-					if (prevUID == item['id']) {
+					if (prevUID === item['id']) {
 						var $placeholder = $('<div/>');
 						$placeholder.addClass('avatar');
 						$row.append($placeholder);
--- a/js/views/composerview.js
+++ b/js/views/composerview.js
@@ -468,7 +468,32 @@ define(function(require) {
 						this.value = terms.join(', ');
 						return false;
 					}
-				});
+				}).data('ui-autocomplete')._renderItem = function ($ul, item) {
+					var $item = $('<li/>');
+					var $row = $('<a/>');
+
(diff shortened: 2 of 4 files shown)


js/views/composerview.js
```diff
@@ -468,7 +468,32 @@ define(function(require) {
...
+				}).data('ui-autocomplete')._renderItem = function ($ul, item) {
+					var $item = $('<li/>');
+					var $row = $('<a/>');
+
+					$row.addClass('mail-recipient-autocomplete');
+
+					if (item.photo && item.photo !== null) {
+						var $avatar = $('<img/>');
+						$avatar.addClass('avatar');
+						$avatar.height('32px');
```
ChristophWurst: we should move that to mail.css too

MariusBluem:
> remove ugly background color/image

This would be also nice for the server in general 😁 I think the grey is a relict from very legacy versions 😜 @jancborchardt

skjnldsv:
@MariusBluem It would be great to remove all background in general! I'm working on a night theme to prevent my eyes from dying (the white of nextcloud/owncloud is terrible! 😭) and there is so many white/grey background superposed this is becoming a nightmare to edit... 😞

ChristophWurst:
> I'm working on a night theme to prevent my eyes from dying

Try http://jonls.dk/redshift/ 😎

skjnldsv:
> Try http://jonls.dk/redshift/ 😎
I already am! But it's not enough. I'm usually forced to use my screen in negative! ^^

skjnldsv:
@ChristophWurst Strange, avatars aren't displaying on my nc. Is it taking the data from contacts? I have avatars though! 😕

ChristophWurst:
@skjnldsv do avatars work for you in the message list? There's currently a bug where it does not work on some setups. Cannot reproduce locally unfortunately.

What exactly do you see? Nothing? Yes, it takes your contacts images.

MariusBluem:
> @MariusBluem It would be great to remove all background in general! I'm working on a night theme to prevent my eyes from dying (the white of nextcloud/owncloud is terrible! 😭) and there is so many white/grey background superposed this is becoming a nightmare to edit... 😞

I like this idea from a accessibility point of view 💪 Great!

ChristophWurst:
> I see a list without images.

Could you inspect that row? Do you see any of the HTML that I've added with js?.

Did you enable debug mode *or* run the Makefile to minify the js files? Maybe you're using scripts from another branch

jancborchardt:
Very nice @ChristophWurst! :) I’ll take it on if no one else wants to.

Some technical things:
- [ ] The apostrophes, be it single- or double-quotes, need to be removed. Otherwise the avatar placeholder displays them. And they are also just unnecessary.
- [ ] For contacts which have multiple email addresses, the email should be shown in half-grey behind or below the name. Otherwise they are impossible to distinguish.

skjnldsv:
DOn't worry Christoph, the error was mine. Wasn't compiling properly.
Here you go: 
screen

ChristophWurst:
> The apostrophes, be it single- or double-quotes, need to be removed. Otherwise the avatar placeholder displays them. And they are also just unnecessary.

IMO this is a separate issue of either contacts or collected addresses being stored wrongly. That should be fixed properly on the server side. We should refactor that anyways because we're now showing duplicates too.

css/mail.css
```diff
@@ -389,6 +389,16 @@
 	right: -10px;
 }
 
+/* autocomplete list */
+.ui-autocomplete .ui-menu-item a.mail-recipient-autocomplete {
+	display: flex;
```
ChristophWurst: nice!

ChristophWurst:
> The apostrophes, be it single- or double-quotes, need to be removed. Otherwise the avatar placeholder displays them. And they are also just unnecessary.
> 
> IMO this is a separate issue of either contacts or collected addresses being stored wrongly. That should be fixed properly on the server side. We should refactor that anyways because we're now showing duplicates too.

Just checked: the ``oc_mail_collected_addresses`` contains mail address information in different formats. Most are stored as ``Jane Doe <jane@doe.de>``, but some like ``"Jane Doe" <jane@doe.de>"``. This could be fixed by a repair step.

jancborchardt:
@ChristophWurst yeah, it sounds like a separate issue but actually in combination with this improvement it then looks strange with the placeholder avatars. ;) Like, you only see `"` in circles, not characters.

skjnldsv:
Thanks :)
Btw, when multiple mail, I only got one displayed! :/

ChristophWurst:
> For contacts which have multiple email addresses, the email should be shown in half-grey behind or below the name. Otherwise they are impossible to distinguish.

Brought back email addresses (there was an error in my code), but not yet half-grey. I'll first add the migration step to fix existing collected addresses and then we can do that visual enhancements for both contacts and collected addresses 🚀

ChristophWurst:
> The apostrophes, be it single- or double-quotes, need to be removed. Otherwise the avatar placeholder displays them. And they are also just unnecessary.

As mentioned above, I'll submit a PR to fix that soon. A simple migration step should be able to sanitize the existing data and split it into recipient name and address (I'll add a new column to the database).

@jancborchardt @skjnldsv @Gomez please re-review this PR! Thanks

skjnldsv:
LGTM 👍 

But I think we also should:
- Only display one avatar for the same contac
- Display the mail address and not only the name

ChristophWurst:
> Only display one avatar for the same contac

Might be tricky since we now rely on the jQuery autocomplete to render the list.

> Display the mail address and not only the name

I fixed that with my last commit.

ChristophWurst:
- [ ] TODO: fix unit tests

```
There was 1 failure:

1) ContactsIntegrationTest::testGetMatchingRecipient

--- Expected

+++ Actual

@@ @@

 Array (

     0 => Array (

         'id' => 1

         'label' => '"Jonathan Frakes" <jonathan@f...s.com>'

         'value' => '"Jonathan Frakes" <jonathan@f...s.com>'

+        'photo' => null

     )

     1 => Array (

         'id' => 2

         'label' => '"John Doe" <john@doe.info>'

         'value' => '"John Doe" <john@doe.info>'

+        'photo' => null

     )

     2 => Array (

         'id' => 2

         'label' => '"John Doe" <doe@john.info>'

         'value' => '"John Doe" <doe@john.info>'

+        'photo' => null

     )

 )
```

js/views/composerview.js
```diff
@@ -474,7 +479,11 @@ define(function(require) {
 
 					$row.addClass('mail-recipient-autocomplete');
 
-					if (item.photo && item.photo !== null) {
+					if (prevUID == item['id']) {
```
ChristophWurst: use === instead please

js/views/composerview.js
```diff
@@ -488,6 +497,8 @@ define(function(require) {
 						$row.append($placeholder);
 					}
 
+					prevUID = item['id'];
```
skjnldsv: Well, this is a requirement from the contact app and the vcard guidelines.
So I guess it should be? 
https://github.com/owncloud/contacts/blob/master/js/services/contact_service.js#L87

js/views/composerview.js
skjnldsv: 👍

jancborchardt:
@ChristophWurst did you see https://github.com/nextcloud/server/pull/1704 ;)

ChristophWurst:
> @ChristophWurst did you see nextcloud/server#1704 ;)

yup :-)

ChristophWurst:
> The apostrophes, be it single- or double-quotes, need to be removed. Otherwise the avatar placeholder displays them. And they are also just unnecessary.

Should be fixed by #151

ChristophWurst:
👍 for @skjnldsv's part. Nice collaboration 😉

skjnldsv:
Thanks! As always, a pleasure! ;)

PR #139, opened 2016-10-07
Use the "live-update" mechanism for timestamps

### LINKED ISSUE #21: Update relative timestamps to give a live feeling
https://github.com/nextcloud/server/pull/1215

I made the thing very general, so apps can use it aswell. Requirements:
* Object has class `live-relative-timestamp`
* Milliseconds timestamp is in attribute `data-timestamp`
* Then the `text()` of the object is updated every 30 seconds.

See https://github.com/nextcloud/activity/pull/38 for another example

@ChristophWurst

nickvergessen:
Fix #21

### DIFF
--- a/js/templates/message-list-item.html
+++ b/js/templates/message-list-item.html
@@ -30,7 +30,8 @@
 			{{subject}}
 		</div>
 		<div class="date">
-				<span class="modified"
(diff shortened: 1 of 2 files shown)


ChristophWurst:
very nice 👍 😍

PR #140, opened 2016-10-07
Be lazy when fetching UserFolder

rullzer:
Since on a lot of requests the routes are loaded (and thus the
application is initialized) it is good to not be greedy. Now we only
fetch the user folder when one of the controllers is actually called.

Me and @LukasReschke found this while profiling some requests on a production instance.

CC: @ChristophWurst

### DIFF
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -39,9 +39,11 @@ class Application extends App {
 			return $c->getServer()->getSession();
 		});
 
-		$user = $container->query("UserId");
 		$container->registerParameter("appName", "mail");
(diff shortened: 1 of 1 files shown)


LukasReschke:
Saves 39 SQL queries on every loaded page. So quite a lot 👍 

See https://blackfire.io/profiles/compare/f0b2df48-2672-4775-a4f7-6fbddb52d315/graph for the diff

cc @ChristophWurst

ChristophWurst:
👍 wow, thanks a lot guys!

PR #141, opened 2016-10-08
Validate info.xml against the new app store's xml schema

ChristophWurst:
Next step: register nightlies at apps.nextcloud.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
     - mysql-client-core-5.6
     - mysql-client-5.6
     - php5-pgsql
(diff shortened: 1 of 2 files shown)


ChristophWurst:
cc @Gomez @BernhardPosselt @skjnldsv

skjnldsv:
Seems good to me! :)
And with tests!! 👍

ChristophWurst:
Thanks @skjnldsv.

I could also still install/uninstall the app and the info about authors and the license is shown correctly on Nextcloud's app page.

PR #148, opened 2016-10-12
Exclude '.lgtm' from tar command.

irgendwie:
Removes '.lgtm' from the appstore tar.
The file was introduced by #7


ChristophWurst:
:+1:

PR #151, opened 2016-10-14
Clean up collected addresses and split into address and display name

ChristophWurst:
Follow up PR for #137 to clean up existing mail addresses


ChristophWurst:
one issue left:
- [ ] insert full address when choosing a recipient and not only it's mail address, e.g. `"Franz Fuchs" <franz@fuchs.at>`

ChristophWurst:
Alright, seems to work as expected. @jancborchardt @Gomez @skjnldsv please give it a test run and let me know it something explodes.

Note that we might still not have the desired display format (e.g. I'd like to see the email address of auto-complete recipients, but the " at the beginning should not be taken for the avatar generation). However, I made this PR solely to clean up our database structure and it's data. Once this PR is in, we can improve the UI part in follow-up PRs.

skjnldsv:
Works fine!
We should just increase slightly the height of the dropdown for the contacts when sending a message, because it makes the element scroll when hovering the third element.

ChristophWurst:
- [ ] TODO: fix tests

jancborchardt:
@skjnldsv why does the dropdown still look so oldstyle? Do you have current master, or something older?

skjnldsv:
@jancborchardt ah right! I'm not up-to-date on this test server. Wait a sec! 😄

ChristophWurst:
tests and one small bug fixed. please test and review @nextcloud/mail :-)

jancborchardt:
@ChristophWurst what does this exactly fix? I still get people with the `""` quotes around and avatars with quotes, see screenshot in https://github.com/nextcloud/server/pull/2194

ChristophWurst:
> @ChristophWurst what does this exactly fix? I still get people with the "" quotes around and avatars with quotes, see screenshot in nextcloud/server#2194

Strange. Are you sure you were on this branch, you ran the app update and you've either enabled debug mode or generated the minified js?

I just tested it again. This is what I saw in the DB *before* the app update:
bildschirmfoto von 2016-12-03 16-45-39
(note the mix of label and email address)

Then I ran the update:
bildschirmfoto von 2016-12-03 16-45-45

Another look at the DB confirms that the entries have been cleaned up:
bildschirmfoto von 2016-12-03 16-46-15

On the UI I can auto-complete one of the collected entries:
bildschirmfoto von 2016-12-03 16-48-15

When I click it the correct label and e-mail address is inserted in the desired format the SMTP server will understand when sending the message:
bildschirmfoto von 2016-12-03 16-52-32

@jancborchardt could you please try to reproduce the steps above? Note that this PR is only meant to fix collected addresses. Auto-completed contact images are unrelated and can be fixed in another PR 😉

jancborchardt:
Ah ok, so this doesn’t fix addresses already in Contacts? Cause this is what I see:
capture du 2016-12-05 13-09-39

jancborchardt:
But let’s get this in and fix the other stuff going forward maybe? Too many pull requests open :)

PR #157, opened 2016-10-15
move model specs/tests into their own directory

ChristophWurst:
With a growing number of spec files it's a good idea to organize them into directories.


Gomez:
:+1:

PR #108, opened 2016-09-22
Increased the space between label from and to

RucajDenisa:

PR #162, opened 2016-10-19
improve styles of message composer and reply

jancborchardt:
Before & after:
capture du 2016-10-19 14-31-28
- more whitespace
- proper sizes of inputs and buttons for clickability
- less lines and boxy-look, more fitting to Nextcloud
- bonus fixes to the attachment list style
- also works nicely on mobile and for replies
- width change which was introduced by @RucajDenisa is still there, as the Message textarea is restricted in width :)

Please review @nextcloud/mail @nextcloud/designers

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -20,6 +20,10 @@
 
 .mail-account {
 	width: 100%;
+	height: 44px;
+	padding-left: 30px;
+	background-color: transparent;
+	border: none;
 }
 
 
@@ -398,6 +402,12 @@
 .ui-autocomplete .ui-menu-item a.mail-recipient-autocomplete span {
 	margin-left: 10px;
 }
+.ui-autocomplete {
+	padding-left: 16px !important;
+	border-radius: 0 !important;
+	border-left: none;
+	border-right: none;
+}
 
 /* editor */
 #mail-new-message-fixed {
@@ -419,9 +429,8 @@
 }
 
 .message-composer {
-	margin: 10px;
+	margin: 0;
 	z-index: 100;
-	max-width: 700px;
 }
 #reply-composer .message-composer {
 	margin: 0;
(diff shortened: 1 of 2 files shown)


skjnldsv:
👍 yes!

ChristophWurst:
why is "Subject" displayed so big? Can you harmonize that with the other labels please?

ChristophWurst:
this replaces https://github.com/nextcloud/mail/pull/108

jancborchardt:
@ChristophWurst Subject is so big intentionally because it’s the title of the whole thing. It’s symmetric to how we show it when viewing an email.

skjnldsv:
I like it this way :)

MariusBluem:
I like this too 🙈👍🏼

ChristophWurst:
1. Why did you change the button's size so much? Doesn't that differ from all the other buttons in NC? I think if we change that, we should change it consistently for all apps.
2. Can we style the address drop down too? At least on FF it looks bad with the default browser styling :-/

ChristophWurst:
Apart from those two nitpicks it looks really nice 💃

jancborchardt:
> 1. Why did you change the button's size so much? Doesn't that differ from all the other buttons in NC? I think if we change that, we should change it consistently for all apps.

So that they have the proper minimum 44px size to be a proper target. :) I agree we should change it everywhere eventually, but it will break a lot of stuff so we have to do it step by step.

> 2. Can we style the address drop down too? At least on FF it looks bad with the default browser styling :-/

According to @Henni this is not possible so people can not mess with the accessibility of the browser and such. This is also the reason select2 exists.

So TL;DR let’s merge? ;) We can address anything else in follow-ups.

ChristophWurst:
okidoke

elioqoshi:
Love this! Small but effective improvement

skjnldsv:
Yeah, @jancborchardt is on fire lately! 🔥 🔥 🔥

jancborchardt:
An awesome community is quite motivating. ;)

Issue #152, opened 2016-10-14
Settingspage not loaded if authentication failed


hechi:
### Steps to reproduce
1. change settings to not valid once (example change sslMode)
2. save changes in DB or late with #132
3. reload page
### Expected behaviour

The Settingspage tells me that it doesn't work, but i am able to change settings.
### Actual behaviour

Spinner appears and nothing happens.
### Mail app

**Mail app version:** git branch issue_16_editAccountSettings, but i guess its in the current master

**Mailserver or service:** own

**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 

**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)

**Number of accounts:** 1

**Mail app version build date:** my branch
### Server configuration

**Operating system**: Debian

**Web server:** Nginx 

**Database:** sqlite

**PHP version:** 7

**Version:** current git

**Updated from an older version or fresh install:** fresh

**List of activated apps:** mail, files

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder
```

<details>
[...]

PR #163, opened 2016-10-19
made account settings editable

hechi:


hechi:
Hey @ChristophWurst and @Gomez here is my PR for the account settings :)

Gomez:
Thanks for your improvement! There are some jscs errors, check the travis link.

I opened the settings page and change my password to a wrong one. After a refresh and some kind of timeout my account got shown but i cant access the settings page. Can we make this work? 

Is the main usecase? Something on the mailserver changed and the user wants to update the settings in the mail app.

hechi:
Yeah, i found this in https://github.com/nextcloud/mail/issues/152 i can try to find it and then throw one of the fancy error messages :)
i will fix travis.

Gomez:
We should show the mail account on the left side and make it possible to edit the settings and save them. 
But lets merge this and mark #152 as a blocker for v0.6.1?

hechi:
ok sounds great :)

hechi:
@Gomez i need the approval if i should merge it ^^ or you merge ^^

Gomez:
:+1: 
One more :+1:  from @ChristophWurst?

ChristophWurst:
First of all: thanks a lot for implementing this. I gave it a quick test without looking at the code (tested from a user's perspective) and there seem to be a few issues:

1. The settings page does not look appealing. We wanted to clean up the alias section already, but since nobody really did it it still looks odd to me. Adding more stuff to it makes the pages's layout even worse. How about we use a tab-view for the settings page and structure all the settings nicely into selectable sub-pages? Something like 'general', 'aliases', 'signatures' and so on. cc @jancborchardt 
2. There is no visual feedback on whether the form submit does anything. First, I thought it wouldn't even work. There should be both feedback *while* updating the settings and *after* they have been saved or if there's an error.
3. I am able to enter wrong settings and the app persists them. This means we "allow" users to get the app into an unusable state. We should prevent that as far as possible and always check whether the newly added settings actually allow connecting to the IMAP/SMTP server.

ChristophWurst:
@hechi would it be possible to do the refactoring *before* actually re-using the form? Code-wise this is a lot of duplication at the moment. It should be possible to extract the existing setup form into something like `AccountSettingFormView` with little effort. That view should only contain the view logic to display the html form and handle the ssl input updates. The callback for submitting could be passed as constructor parameter. That way we can use it for both views: setup & edit.

hechi:
Hey,

to 1.) yes i guess this would be nice an we can do it in a extra Issue
to 2.) o.O it should be a little feedback like "saving..." -> "saved" on the submit button
to 3.) probably it could be that a password is already wrong. so there is no working setting. I would want to save nonfunctional settings but it would be nice if there is a feedback that it worked or not.

hechi:
@ChristophWurst ok if this would be better then i refactor first

jancborchardt:
@ChristophWurst @hechi how should we proceed here by the way? :)

Yamakasi:
When can this be in ? The app is kinda "useless" at the moment for people migrating and such.

hechi:
Hey, sry i had some stuff to do for my final exam, but now its christmastime and i am at the congress. Its my plan to do it there. I hope i am done before new year.

Yamakasi:
Is this commit working now, I tried the master branch but it's still the same issue.

hechi:
hey @Yamakasi  its not in master because there are merge conflicts and i wanted to refactor the implementation. For 6 hours i tried to setup my dev environment but got stuck, so neither i was no able to fix the merge conflicts nor to refactor this stuff :( so if any one want to take the ticket please do it.

Yamakasi:
@hechi Hi, sorry I saw your other reply by now, thanks for the update. Let's hope we can use it soon!

jancborchardt:
@hechi where do you have issues with the setup? :) @jospoortvliet and @blizzz should be able to help you at the Berlin Nextcloud meetup ;)

css/mail.css
```diff
@@ -28,14 +28,18 @@
 
 
 /* SETUP */
-#setup {
```
Quix0r: Uh, oh. Also another PR is around for same line. This may later cause conflicts on merging. The PR is here: https://github.com/nextcloud/mail/pull/298

ChristophWurst: > Conflicting files
> js/service/accountservice.js
> js/views/accountsettings.js
> lib/account.php
> lib/controller/accountscontroller.php
> lib/db/mailaccountmapper.php
> lib/service/iaccount.php
> lib/service/unifiedaccount.php

this PR is conflicting already 🙈

Quix0r: Yeah. Yes, it does.

Yamakasi:
Guys what is the showstopper here ?

hechi:
Hey @Yamakasi like i wrote above, its my time. So if someone wants to implement/refactor it. plz go :)

Yamakasi:
I'm lost here, I need to find time, but how is this usable for anyone ?

ChristophWurst:
https://github.com/nextcloud/mail/pull/298 has been merged. Now this can be finished. Since this is conflicting a lot with the other changes (possibly due to the duplicated code in this PR) it makes sense to freshly branch of from master, take the relevant code from this PR and create a new one. Thanks

PR #164, opened 2016-10-20
move mail address select triangle back next to address

jancborchardt:
As requested by @irgendwie ;D

The downward-pointing triangle of the »from« selector was a bit strange, so the select is simply now as wide as the value to have the triangle directly next to it.

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -19,7 +19,7 @@
 }
 
 .mail-account {
-	width: 100%;
+	max-width: 100%;
 	height: 44px;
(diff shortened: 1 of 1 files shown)


jancborchardt:
Yoyo @irgendwie please check it out ;D

jancborchardt:
@RucajDenisa do you want to give this a quick review? :) It’s just a quick CSS adjustment.

ChristophWurst:
👍 looks good

PR #169, opened 2016-10-25
Make work with nonces

LukasReschke:
Required for https://github.com/nextcloud/server/pull/1871

cc @ChristophWurst As discussed.


ChristophWurst:
👍 works nicely, although there's still a security error on FF:
bildschirmfoto von 2016-11-01 21-25-49
@LukasReschke anything we should fix?

PR #173, opened 2016-11-01
NotFoundException does not exist

ChristophWurst:
Imagine a world where we have unit tests that would actually have detected that faulty code …

### DIFF
--- a/lib/Service/AliasesService.php
+++ b/lib/Service/AliasesService.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @author Tahaa Karim <tahaalibra@gmail.com>
  *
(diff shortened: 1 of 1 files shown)


jancborchardt:
cc @Gomez @irgendwie for review maybe? :)

irgendwie:
Apart from the removed try {...} catch {...} (and some use) Statements, nothing changed - LGTM I guess.
No exception handling "needed anymore", or is it done otherwise now?

ChristophWurst:
I just removed the dead code. We don't have error handling there at the moment :-/

Btw this code is also lacking unit tests. #yolo

PR #174, opened 2016-11-01
fix tests

ChristophWurst:
* use PSR-4
* consistently use phpunit's testcase as base class
* small doc block fixes

cc @nickvergessen @Gomez


tests/Controller/AccountsControllerTest.php
```diff
@@ -19,13 +19,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>
  *
  */
+
```
nickvergessen: class is missing namespace

tests/Controller/AliasesControllerTest.php
```diff
@@ -19,14 +19,10 @@
  *
  */
```
nickvergessen: class is missing namespace

nickvergessen:
👍

PR #177, opened 2016-11-03
also show Sent folder by default, fix #176

### LINKED ISSUE #176: Also show »Sent« folder in the reduced list of folders?
This is the only folder I find myself looking through often and opening the »Show al folders« expander for. Maybe we should show it by default as well in the reduced list of folders?

cc @nextcloud/mail what do you think?

jancborchardt:
Fix #176, please review @skjnldsv @pierreozoux @stefan-niedermann :)

And thank you @ChristophWurst for the pointer!


skjnldsv:
👍

PR #180, opened 2016-11-03
use correct internal server version

ChristophWurst:
Otherwise the app installation fails with

```
occ app:enable mail                                                                                                     

                                                                                                                                 
  [Exception]                                                                                                                    
  App "Mail" cannot be installed because the following dependencies are not fulfilled: Server version 10 or higher is required.  
```

I suspect PR https://github.com/nextcloud/server/pull/1940 caused this incompatibility. cc @LukasReschke


LukasReschke:
:+1:

PR #181, opened 2016-11-04
Update nl.js

megaxlr:
Various changes to nl.js


ChristophWurst:
Hi @megaxlr,

thanks a lot for your contribution. Translations are managed via Transifex, so please submit them on https://www.transifex.com/nextcloud/.
(We have a bot that imports the translation files regularly. Any manual change to the repository would be overridden.)

PR #188, opened 2016-11-07
correctly filter FTP links in mails

ChristophWurst:
* detect FTP urls, so that the proxy controller is used to prevent leaking the request token
* block images served via FTP

### DIFF
--- a/lib/Controller/ProxyController.php
+++ b/lib/Controller/ProxyController.php
@@ -79,7 +79,9 @@ class ProxyController extends Controller {
 	public function redirect($src) {
 		$authorizedRedirect = false;
 
-		if (strpos($src, 'http://') !== 0 && strpos($src, 'https://') !== 0) {
+		if (strpos($src, 'http://') !== 0
+			&& strpos($src, 'https://') !== 0
+			&& strpos($src, 'ftp://') !== 0) {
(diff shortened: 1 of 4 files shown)


ChristophWurst:
@LukasReschke review please :-)

irgendwie:
- :+1: blocks images from ftp servers
- when unblocking, it seems to correctly use the proxy:
http://localhost:9000/index.php/apps/mail/proxy?src=ftp%3A%2F%2F{url}&requesttoken={token}
- Can't test any further: "Message: cURL error 1: Protocol "ftp" not supported or disabled in libcurl" ;-)

ChristophWurst:
> Can't test any further: "Message: cURL error 1: Protocol "ftp" not supported or disabled in libcurl" ;-)

Where did that pop up?

> Edit: Should it detect sftp too?

Nope, sftp should be blocked generally.

irgendwie:
> Where did that pop up?

Image can't be loaded **in my setup?**, this happens when I copy the image link (the proxy url) and directly try to open it in the browser
=> Unrelated I suppose

ChristophWurst:
For security reasons it's important to block ftp resources by default. But please open an issue so we can addresss that error in the proxy controller too. Thanks :-)

PR #215, opened 2016-12-02
automatically sign and package the app

ChristophWurst:
This rsyncs the files into a directory that can then be used to sign the app's
files. Moreover, the created tar.gz is then signed and the certificate is
printed to the command line. All steps that require the private key are only
run if the key is available. Therefore peole can also use this to build
their own version :-)

Only tested without a key because I don't have one for Mail yet cc @Gomez :-)


Gomez:
:+1: Nice! Travis makes trouble with the db?

ChristophWurst:
Strange, will take a look later.

PR #216, opened 2016-12-02
enable mysql service on travis

ChristophWurst:
*Should* fix travis failures on mysql like in #215


ChristophWurst:
💥 💥 💥 🤔

ChristophWurst:
works again.

PR #219, opened 2016-12-03
Remove quotes from contact integration repient label

ChristophWurst:
Fixes https://github.com/nextcloud/mail/pull/151#issuecomment-261396530 (only suggested recipients from the address book)

Now the proper avatar is shown instead of ":
bildschirmfoto von 2016-12-03 16-59-32

cc @jancborchardt @skjnldsv :-)

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -65,11 +65,10 @@ class ContactsIntegration {
 
 			// loop through all email addresses of this contact
 			foreach ($email as $e) {
-				$displayName = "\"$fn\" <$e>";
 				$receivers[] = [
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Btw. we should have a consensus on the label format. It should be identical for contacts and collected addresses. I think we should show the label (e.g. name) *and* the email address for each entry.

MariusBluem:
Does this also applies on Hofer Mails or only on VDB's 🤔😜 @ChristophWurst

MariusBluem:
👍

skjnldsv:
👍

ChristophWurst:
only VdB :-P

Issue #182, opened 2016-11-04
Only use »make« for compiling


jancborchardt:
Instead of the two-step process:
> make install-composer-deps
> make optimize-js

Can we make it so that simply running »make« does everything? It seems like that’s even already the case and we just need to change the readme?

ChristophWurst:
well, once we decided that 'make' should build the app store package. And that's what it still does. We should try to find a common procedure across all apps (e.g. calendar, contacts and news).

ChristophWurst:
I'm not a big fan of the Makefile anyway because it abstracts the underlying technology away, although that is widely known and accepted in other web projects (composer, bower, npm).

jancborchardt:
One of our main goals is to make it easy for people to contribute. Building the app store package is something only the maintainer does, and not often. So it could be »make app-store« or so.

`make` should ideally:
- install all the dependencies needed for building (this node and npm stuff especially which is a drag …)
- carry out all the commands
- if any command fails because of cached files or whatnot, clean them up

cc @Henni @irgendwie @georgehrke @raimund-schluessler @v1r0x @oparoz for other apps.

ChristophWurst:
For newbies it might be too complicated to use the tools directly, but anybody who has experience with web development knows how to deal with npm and bower. The latter are probably confused more when they find the Makefile because that is very uncommon for this kind of software project (they are common for C/C++ and the like though).
But I see your point. If some1 volunteers to create this magic Makefile, I'm happy to review their PR 😉

Quix0r:
GNUSocial uses a Makefile to build e.g. language files (gettext).

ChristophWurst:
Well, translations is about all the Makefile does. https://git.gnu.io/gnu/gnu-social/blob/master/Makefile

Quix0r:
Yupp. :-)

jancborchardt:
> anybody who has experience with web development knows how to deal with npm and bower

Sorry but this is not entirely true. They are all commands you need to remember, and there are varying degrees of »experience with web development«. And we also want to attract designers, and people who primarily work with HTML and CSS. They don’t necessarily know how to grunt build and npm install and all this stuff.

oparoz:
I don't think the problem is the make file. From the moment you need bower, npm, etc. It's game over for beginners. They need to be able to edit JS files and see the results right away.

ChristophWurst:
Ok. So does anybody have a good solution for this?

[...]

PR #220, opened 2016-12-04
include bower dependencies in git

ChristophWurst:
Makes people's life easier as discussed in https://github.com/nextcloud/mail/issues/182.

I will add npm deps too once this has been merged.

cc @jancborchardt @nextcloud/mail


jancborchardt:
Works nicely :+1:

jancborchardt:
Also cc @Henni & @irgendwie from Contacts. This is a thought we should have for Contacts as well since we had problems getting it set up for contributors in the past.

skjnldsv:
Agreed, but I don't think this is why no one work on contacts 😄

jancborchardt:
It can definitely be a factor. At an event I helped someone set up a Contacts dev environment and it was really frustrating with all the dependencies. We ended up needing to copy my vendor files etc. over.

PR #222, opened 2016-12-05
remove horizontal scrolling from left navigation

jancborchardt:
The left navigation had a glitch where it could be scrolled horizontally. Caused by negative right margin of the 3-dot-menu for accounts.

Please review @nextcloud/designers @nextcloud/mail


skjnldsv:
👍

PR #224, opened 2016-12-05
fix and sort rsync excludes

ChristophWurst:
Excludes are now sorted alphabetically.

Two new entries:
* nbproject (I use NetBeans)
* coverage (generated if you run js unit tests locally)


ChristophWurst:
effects only packaging and I used this to release 0.6.1 already -> merging now 🙈

PR #231, opened 2016-12-08
Update changelog for 0.6.2

ChristophWurst:
Everything in 0.6.1 was moved to 0.6.2 because we had to release a NC11 compatible version and I chose it to be 0.6.1.
Therefore, I now increased the version number for master and added all relevant changes that were made in this milestone. If no blocking issues are found in the next few days (I'm using the nightly build to verify in production), I will push it to the app store. Again, we did not release anything for three months, although we've learned from the past that smaller and more regular releases are easier to handle. I'd like to get back to that strategy and release more often now.

cc @Gomez @jancborchardt @nextcloud/mail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## 0.6.2 – 2016-12-10
+## 0.6.2 – 2016-12-12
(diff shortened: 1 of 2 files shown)


ChristophWurst:
#232 should be the last one. Then we're ready. @nextcloud/mail

PR #232, opened 2016-12-08
use detachChildView instead of reset

ChristophWurst:
This caused a js error and stalled user interface when creating the first account.


ChristophWurst:
ref http://marionettejs.com/docs/master/marionette.view.html#detaching-a-child-view

PR #248, opened 2016-12-16
Limit NC support to 11-12, php to 5.6-7.1

ChristophWurst:


cmtonkinson:
When is the next release scheduled that will include this patch? I'm unable to use this app on a fresh Docker install of NextCloud 11:latest

MariusBluem:
PHP dependencies seems to be not longer needed. They are equal to the server's one.

cmtonkinson:
Apologies - I was "researching while mobile" (should be a chargeable offense) and didn't notice this was in milestone 0.6.3.

ChristophWurst:
Hey,

we do not follow a strict release plan, hence I cannot tell you when we will release the next version of this app. All I can say is that we're actively working on finishing a release, but some bugs are yet to resolve before we can do so. As always, any help is very much appreciated - feel free to help us with coding, testing, translation or documentation 😉

cmtonkinson:
Understood, thanks @ChristophWurst.

ChristophWurst:
The latest version is compatible with 7.0, 7.1 and 7.2: https://github.com/nextcloud/mail/blob/v0.11.0/appinfo/info.xml#L24

Issue #234, opened 2016-12-13
App not loading when installing from app store


juliushaertl:
### Steps to reproduce
1. Install mail 0.6.2 from the Nc app store
2. Open the mail app

### Expected behaviour
It should load

### Actual behaviour
It doesn't load

##### Browser log

![2016-12-13-133506_1916x754_scrot](https://cloud.githubusercontent.com/assets/3404133/21140668/ab619b10-c139-11e6-8791-dd2c3631949d.png)

juliushaertl:
vendor/.../build/ directories are not existent in the released packages.

ChristophWurst:
Could the markdown editor somehow interfere with our js files? Apparently something breaks requirejs.

> vendor/.../build/ directories are not existent in the released packages.

Which file/directory is needed?

ChristophWurst:
Oh, do you have debug mode enabled? Indeed some files are missing, but it should not affect the bundled mail.min.js file as it includes everything.

juliushaertl:
@ChristophWurst Ah thanks. It works fine after disabling files_markdown and debug mode. 

> Which file/directory is needed?

When debug mode is enabled it tries to fetch js/vendor/backbone.radio/build.backbone.radio.js and js/vendor/ical.js/build/ical.min.js which do not exist.

The following errors appear with disabled debug mode and enabled files_markdown app:

![2016-12-13-141038_728x107_scrot](https://cloud.githubusercontent.com/assets/3404133/21141609/6a0cce46-c13e-11e6-9416-aa642815dd77.png)
![2016-12-13-141044_672x146_scrot](https://cloud.githubusercontent.com/assets/3404133/21141614/6d162c5e-c13e-11e6-944f-f649dd4cc2b2.png)

ChristophWurst:
> The following errors appear with disabled debug mode and enabled files_markdown app:

I'd say that's a bug of the files_markdown app. Its js should not be loaded on other app's pages.

> When debug mode is enabled it tries to fetch js/vendor/backbone.radio/build.backbone.radio.js and js/vendor/ical.js/build/ical.min.js which do not exist.

I messed up the rsync command in the Makefile, apparently excludes are not relative paths but patterns. I wanted to exclude the build directory of the app, but that also excluded *any* build dir in the directory tree.

eppfel:
Jep, had exactly the same. Deactivated `debug` now.

plastikschnitzer:
[...]

PR #253, opened 2016-12-18
fix rsync excludes to include 'build' directory of dependencies

ChristophWurst:
fixes https://github.com/nextcloud/mail/issues/234

@nickvergessen thanks for the hint

I compared the packaged directory with the one from master and got the following results:
```
Nur in artifacts/mail/js/vendor/backbone.babysitter/src: build.
Nur in artifacts/mail/js/vendor/backbone.radio: build.
Nur in artifacts/mail/js/vendor/backbone.wreqr/src: build.
Nur in artifacts/mail/js/vendor/ical.js: build.
```
So the missing build directories are included in the archive now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Makefile for building the project
 
 app_name=mail
-project_dir=$(CURDIR)/../$(app_name)
+project_dir=$(CURDIR)
 build_dir=$(CURDIR)/build/artifacts
 appstore_dir=$(build_dir)/appstore
 source_dir=$(build_dir)/source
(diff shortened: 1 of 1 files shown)


nickvergessen:
That s why the current upload doesnt load the list i guess?

:+1:

ChristophWurst:
> That s why the current upload doesnt load the list i guess?

List? If you mean the 'build' directories missing in the last release, then yes.

nickvergessen:
What I meant is that I get "Beim Laden von Posteingang ist ein Problem aufgetreten :-(" on my inbox folder. Others seem to work fine

ChristophWurst:
please file an issue :-)

PR #263, opened 2016-12-27
line up attachment button with send button

jancborchardt:
Open the »New message« window, write a long message (so the main view scrolls) and you’ll notice the bottom buttons not being aligned:
capture du 2016-12-17 03-50-04

This is fixed now:
capture du 2016-12-17 03-49-50

Please review @nextcloud/designers ;)

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -424,6 +424,7 @@
 
 .message-composer {
 	margin: 0;
+	margin-bottom: 10px; /* line up with the send button */
 	z-index: 100;
 }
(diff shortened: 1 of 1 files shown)


skjnldsv:
👍 
But shouldn't we put the "add attachment" into a floating mode like the send button? :)

jancborchardt:
@skjnldsv I wouldn’t, because then the reply action gets overpowered. :) Because the floating reply button is basically a shortcut to the whole reply area.

PR #270, opened 2017-01-03
Use ES6 promise in accountservice

ChristophWurst:
Let's use the ES6 standardized promises instead of jQuery's. I'll create small, reviewable PRs so that we catch possible regressions 🚀 🤖 🕺

### DIFF
--- a/js/controller/accountcontroller.js
+++ b/js/controller/accountcontroller.js
@@ -31,12 +31,10 @@ define(function(require) {
 	 */
 	function loadAccounts() {
 		var defer = $.Deferred();
-		var fetchingAccounts = Radio.account.request('entities');
-
 		// Do not show sidebar content until everything has been loaded
 		Radio.ui.trigger('sidebar:loading');
 
-		$.when(fetchingAccounts).done(function(accounts) {
+		Radio.account.request('entities').then(function(accounts) {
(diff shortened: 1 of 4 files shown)


ChristophWurst:
It's easier to review with https://github.com/nextcloud/mail/pull/270/files?w=1

jancborchardt:
@nickvergessen @skjnldsv can you review this? :)

skjnldsv:
Oups sorry! Didn't saw that one! 🙈 
I have issue with the inbox number not being updated when I delete some mails before they are fully loaded and when removing a message, the message stays on the right panel.

I don't think this is related to this however :)

nickvergessen:
@jancborchardt asking me for a js review is like asking you for a php review ;)

jancborchardt:
@nickvergessen :D I only @-mentioned you because Christoph apparently set @skjnldsv and you as reviewers. ;)

skjnldsv:
Mwahaha! 👿 @jancborchardt

nickvergessen:
> And just doing that doesn’t notify those people.

It does, I have more then 30 mails with:

jancborchardt:
Ahh, good to know! :)

nickvergessen:
Assigning also sends an email these days

PR #271, opened 2017-01-04
Refactor server-side folder management implementation

ChristophWurst:
I finally found the time to study the server-side implementation and refactor it into a testable, extensible form. This is needed for the planned features like threading. Before I start with those I want the code to be tested almost 100% because testing IMAP manually is a major PITA. This PR shows that moving the existing code a bit around can lead to a much better code quality. Actually, I could even remove some of the code in place because it's just not needed or can be done more efficiently.

* better OOP architecture
* dependency injection everywhere
* abstraction where it made sense
* SRP applied as much as possible
* added a bunch of tests for the new code
* REST API compatible to old implementation
* new code is not optimized
* old code was not removed

TODO:
- [x] 100% code coverage (actually only a few code paths missing)
- [x] ~~check why unified inbox flagged folder is not shown any more~~ well, we never had that feature 🙈 
- [ ] file ticket about the js bug that occurs when accounts are added
- [x] re-enable search mailbox (starred messages filter)


jancborchardt:
Aaawesome! :) Let us know as soon as it’s reviewable! cc @nextcloud/mail

ChristophWurst:
I have to do some more refactoring before I can finish this one unfortunately, otherwise I'd have to add those ugly unified inbox hacks to the new shiny code too 😢 Should be easy though 😉

ChristophWurst:
@nextcloud/mail this should be ready for a review 🚀 

Let me know if something breaks 😉

ChristophWurst:
Open todo (though that shouldn't break anything when added back): caching.

ChristophWurst:
Please help with testing/reviewing this PR @nextcloud/mail @Quix0r

skjnldsv:
That is a long pr! Care to told us how to review? :p

ChristophWurst:
> That is a long pr! Care to told us how to review? :p

Yup, sorry 'bout that. Well, just look at the diff 😉 I pretty much re-wrote the old implementation and added a bunch of tests. Unfortunately, I cannot tell you how this could be reviewed easily 😟

skjnldsv:
Sooo, read the js and process it into your head? 😆

ChristophWurst:
> Sooo, read the js and process it into your head? 😆

Sure you can do that hehe

Most of the changes were made to the server implementation though.

So I basically rewrote the code that handles folder management, which is mainly only about listing folders of an IMAP account. All we do after we got the list is translation, sorting, guessing the folder role and building the folder tree from the source list.

To have the most flexibility, maintainability, testability and readability I tried to split the code into logical parts:
* a high level service to use in the controller -> a new folder manager
* a class that translates IMAP into the OOP world -> a folder mapper
* a class to handle translation -> a translator class

The rest of the rather big diff is caused by tests because I almost hit 100% coverage on the new classes (only one little, uncritical if-statement is not covered).

Thus you can start at the controller, look what it calls (spoiler: a single method on the manager) and review the delegated functionality in there. Everything should be self-explanatory from the class/method names, or at least that was my intention 😉 

Let me know if there's anything else you need to understand the changes :-)

Quix0r:
Hmm, `/js/models/account.js` has lost some coverage: https://coveralls.io/builds/10080949/source?filename=js%2Fmodels%2Faccount.js

jancborchardt:
Been using it today and can’t find any issues so far. :) Any blockers or things to add back in this specific PR?

ChristophWurst:
@jancborchardt cool, thanks for testing! I think this is ready for merging 😉

ChristophWurst:
About the js code coverage: it's not really reliable. It goes up and down. We're not at 33% coverage anyway, the number seems to be wrong :-|

jancborchardt:
Perfect! Looking forward to the other unified folders. ;) What can I test next?

skjnldsv:
Damn, sorry @ChristophWurst I forgot this one! 😢 
Thanks @jancborchardt

Quix0r:
@ChristophWurst TODO 3 is still open in your list!

js/models/account.js
```diff
@@ -74,6 +75,17 @@ define(function(require) {
 			}
 			return undefined;
 		},
+		getSpecialFolder: function() {
+			if (!this.folders) {
+				return undefined;
+			}
+			return _.find(this.folders, function(folder) {
+				// TODO: handle special folders in subfolder properly
```
Quix0r: Maybe use `@TODO` so documentation tools can pick it up? It is very common though.

lib/Contracts/IMailManager.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+namespace OCA\Mail\Contracts;
+
+use OCA\Mail\Account;
+
+interface IMailManager {
+
+	public function getFolders(Account $account);
```
Quix0r: Methods declared in interfaces are always public, no need to have the then redundant `public` statement around.

lib/Controller/FoldersController.php
```diff
@@ -23,65 +23,57 @@
...
-	 * @param string $appName
+	 * @param type $appName
 	 * @param IRequest $request
 	 * @param AccountService $accountService
-	 * @param $UserId
+	 * @param string $UserId
+	 * @param IMailManager $mailManager
 	 */
-	public function __construct($appName, IRequest $request, AccountService $accountService, $UserId) {
+	public function __construct($appName, IRequest $request,
```
Quix0r: Really split this into more than one line? If that is the coding-convention here, okay.

tests/Controller/FoldersControllerTest.php
```diff
@@ -22,70 +22,78 @@
...
 	}
 
-	/**
-	 * @dataProvider folderDataProvider
-	 */
-	public function testIndex($data) {
-		$account = $this->getMockBuilder('OCA\Mail\Account')
-			->disableOriginalConstructor()
-			->getMock();
+	public function testIndex() {
```
Quix0r: Please don't kill doc tags, they produce nice documentation (e.g. generated by `doxygen`).

tests/FolderTest.php
```diff
@@ -0,0 +1,161 @@
+<?php
+
+/**
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
```
Quix0r: Have you (the team) already made a decision for `@copyright` tag? Maybe give it to FSFE so they can handle any law issues for you. Maybe on the developer team (in general) to express that people develop code not for themselves but for the project?

ChristophWurst:
@Quix0r will file an issue right away 🚀

tests/FolderTest.php
```diff
@@ -0,0 +1,161 @@
...
+		$this->assertTrue($this->folder->isSearchable());
+	}
+
+	public function testIsNotSearchable() {
+		$this->mockFolder(['\noselect']);
+
+		$this->assertFalse($this->folder->isSearchable());
+	}
+
+	public function testJsonSerialize() {
```
Quix0r: Lots of code is missing documentation (what does it do, what is the method `foo()`and `bar()` for?)

tests/data/folders_german.json
```diff
@@ -1,81 +0,0 @@
-{
```
Quix0r: Got rid of an old file? Good. :+1:

js/models/account.js
ChristophWurst: makes sense, will use it next time.

Quix0r: Then maybe apply this globally, e.g. search for `// TODO` and replace it with `// @TODO`. And same with multi-line comments.

tests/FolderTest.php
ChristophWurst: cc @schiessle 

What do you think is best for this project?

tests/FolderTest.php
ChristophWurst: That's true, many segments of our code are either only documented poorly or not documented at all. We can work on that ofc :-)

tests/Controller/FoldersControllerTest.php
ChristophWurst: What do you mean by *doc tags*?

lib/Contracts/IMailManager.php
jancborchardt: Seems like good practice to always be specific. :)

tests/Controller/FoldersControllerTest.php
Quix0r:
````
/**
 * Some documented method, does foo things and need bar for its purpose. You have to enter a better description than this
 *
 * @param string $bar Some bar string
 * @return void
 */
public function doFoo (string $bar) {
    // Do foo things with bar
[...]
}
````
That is a so called doc-tag. :-) It will produce nice documentation in such tools (see `Doxygen`, it can process more than just C/C++).

PR #273, opened 2017-01-06
Change to full path for nightly upload

Gomez:
The tar -C "path" changes the dir before the upload. This adds a full path to curl to upload the tar.gz.


Gomez:
No, tricky with travis. It should not break anything :-)

PR #279, opened 2017-01-12
migrate alias code to es6 promises

ChristophWurst:
Migrates more code to the es6 standard promises and removes dead code


ChristophWurst:
> All software dependencies are fine. You are awesome!

Thanks, VersionEye! 😉

PR #282, opened 2017-01-12
decouple and recactor account deletion for es6-promises

ChristophWurst:
Following our usual architecture, I've moved the ajax call from the view to the account service. Additionally, I've migrated this call to use an es6 promise.

### DIFF
--- a/js/service/accountservice.js
+++ b/js/service/accountservice.js
@@ -1,3 +1,5 @@
+/* global Promise */
+
 /**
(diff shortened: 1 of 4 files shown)


ChristophWurst:
pushed another commit to fix a failing test

skjnldsv:
I indeed didn't checked the tests.

ChristophWurst:
fixed

PR #286, opened 2017-01-15
Messageservice es6 promise

ChristophWurst:
This is yet another cleanup to use es6 promises everywhere. Only 4 are left 🥂 

I really like how the code gets a bit cleaner 🤠


ChristophWurst:
the diff at https://github.com/nextcloud/mail/pull/286/files?w=1 is probably easier to review

ChristophWurst:
it's also easier to review each commit separately as I've tried to migrate one method per commit.

ChristophWurst:
🕺 😉

PR #298, opened 2017-01-28
show nice loading view while the account is being created

ChristophWurst:
This refactors the account form into a separate and hopefully reusable sub-view for https://github.com/nextcloud/mail/pull/163. cc @hechi 😉 

While at it, I give this a little overhaul. After a user submit their account data, a nice and smooth loading view is shown (including a status/loading message) instead of the form being greyed out. @jancborchardt not sure if a smiley is appropriate there too, but we could show one 😜 

There's one thing that still bothers me: in case on an setup error, we show an ugly yellow notification on top. @jancborchardt @skjnldsv @eppfel suggestions on what we could show instead? We could
* show something like "Unable to create your account" and a "Retry" button that shows the form again
* show something like "Unable to create your account" and automatically show the form after a few seconds

TODO:
- [x] fill the form (or reuse the old subview) on failure so that users don't have to type in their data again
- [x] some js tests would be nice


jancborchardt:
> show something like "Unable to create your account" and automatically show the form after a few seconds

That sounds best. :) The extra retry button would be just a step people have to do manually.

ChristophWurst:
Okay, but for how long should we show this? Too shortly means people won't be able to read it and if it takes too long it's kinda annoying with multiple retries.

js/templates/account-form.html
```diff
@@ -0,0 +1,112 @@
...
+				       placeholder="{{ t 'SMTP Password' }}" value=""
+				       required />
+			</p>
+		</div>
+
+		<input type="submit"
+		       class="primary"
+		       value="{{ t 'Connect' }}"/>
+	</fieldset>
+</form>
```
Quix0r: Misses a new-line character! This can confuse/crash old editors. Most todays modern editors (including CoolEdit with Midnight Commander) allow warning the user if such line is missing.

js/templates/setup.html
```diff
@@ -1,113 +1 @@
...
-		</div>
-
-		<div id="connect-loading" class="icon-loading"></div>
-		<input type="submit"
-		       class="primary"
-		       value="{{ t 'Connect' }}"/>
-	</fieldset>
-</form>
\ No newline at end of file
+<div class="setup-content container"></div>
```
Quix0r: New-line character missing.

js/views/appview.js
```diff
@@ -24,7 +24,7 @@ define(function(require) {
 	var ErrorView = require('views/errorview');
 	var LoadingView = require('views/loadingview');
 	var NavigationView = require('views/navigation');
-	var SetupView = require('views/setup');
+	var SetupView = require('views/setupview');
```
Quix0r: Nice one to have naming constant.

js/service/accountservice.js
```diff
@@ -23,6 +23,10 @@ define(function(require) {
 
 	function createAccount(config) {
 		return new Promise(function(resolve, reject) {
+			setTimeout(function() {
```
ChristophWurst: @jancborchardt if you remove this and the return statement, the setup should work again.

Quix0r:
One question, have my mentioned things being addressed? :-) Else, nice work!

ChristophWurst:
> One question, have my mentioned things being addressed? :-) Else, nice work!

Not yet, but I will definitely work on those. Thanks a lot for your thorough code review :-)

js/templates/account-form.html
ChristophWurst: done :)

ChristophWurst:
Now everything seems to work nicely, time to review @jancborchardt @lcalaresu @Quix0r 🚀

ChristophWurst:
Tested extensively, automated tests were added too and CI checks are happy -> I'll merge this now.

jancborchardt:
Nice! :)

Issue #243, opened 2016-12-14
Inbox Content is empty


flonowa:
### Steps to reproduce
1. Open the app and go in the Inbox Folder
### Expected behaviour
Tell us what should happen
See the Content of my Mails
### Actual behaviour
Tell us what happens instead
When you go on all Mails in the Inbox the Content is empty and a circle spin but this bug is not in all folders. The sent Folder are ok!
### Mail app

**Mail app version:** (see apps admin page)
0.6.2
**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
Domainfactory
**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 
none SSL auto
**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)
none SSL auto
**Number of accounts:**
2
**Mail app version build date:** (only if you are using a Nightly Build)

0.6.2
### Server configuration
**Operating system**:
Ubuntu 16.04
**Web server:**
Nginx
**Database:**
MariaDB
**PHP version:**
7
**Version:** (see admin page)

**Updated from an older version or fresh install:**
fresh install
**List of activated apps:**
collaboraonline
```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder

"system": {
        "instanceid": "ocf1lv2u949m",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "cloud.mydomain.net"
        ],
        "datadirectory": "\/var\/www\/nextcloud_data",
        "overwrite.cli.url": "http:\/\/cloud.mydomain.net\/nextcloud",
        "dbtype": "mysql",
        "version": "9.1.2.2",
        "dbname": "nextcloud_db",
        "dbhost": "localhost",
        "dbport": "",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
        "installed": true,
        "memcache.local": "\\OC\\Memcache\\APCu",
        "theme": "",
        "loglevel": 0,
        "maintenance": false,
        "mail_smtpmode": "smtp",
        "mail_smtpsecure": "ssl",
        "mail_from_address": "flo",
        "mail_smtpauthtype": "LOGIN",
        "mail_smtpauth": 1,
[...]

PR #309, opened 2017-02-17
Remove the broken //IGNORE option

ChappIO:
In the latest versions of iconv, the //IGNORE options is broken.

Instead of filtering out all invalid characters, it will silently fail and return an empty string. The result of this is that the user cannot see the body of the email.

This has already been reported in issue #243

### DIFF
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -594,8 +594,10 @@ class IMAPMessage implements IMessage, JsonSerializable {
 		$data = $fetch->getBodyPart($partNo);
 
 		$p->setContents($data);
-		
-		return mb_convert_encoding($p->getContents(), 'UTF-8', $p->getCharset());
+		$data = $p->getContents();
+
+		$data = mb_convert_encoding($data, 'UTF-8', $p->getCharset());
+		return $data;
 	}
 
 	public function getContent() {
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -596,7 +596,7 @@ class IMAPMessage implements IMessage, JsonSerializable {
 		$p->setContents($data);
 		$data = $p->getContents();
 
-		$data = iconv($p->getCharset(), 'utf-8//IGNORE', $data);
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Hey @ChappIO,

thank you for this bugfix. I will try to test this ASAP ans see if this could break something. IIRC I once touched that code/line too because of a bug, but I cannot remember whether I removed/wanted to remove/added the //IGNORE part.

ChristophWurst:
Found it! https://github.com/owncloudarchive/mail/pull/697 - @LukasReschke @irgendwie could you help with this? 😉

ChappIO:
Ah I see the tests now fail on UTF-8 input? So I guess that error should be handled some way?

This issue with iconv is documented here: http://php.net/manual/ru/function.iconv.php

orrd101@gmail.com mentions a workaround:

```
  ini_set('mbstring.substitute_character', "none"); 
  $text= mb_convert_encoding($text, 'UTF-8', 'UTF-8'); 
```

I am not proficient enough in PHP to know whether or not this has the same (and desired) behavior.

ChappIO:
Any updates on this issue?

ChristophWurst:
> I am not proficient enough in PHP to know whether or not this has the same (and desired) behavior.

Me neither. How about we adjust the code and tests and see what happens? :-)

ChappIO:
So I updated it to use `mb_convert_encoding` instead of `iconv`. Tests seem to pass now, however coveralls things I reduced code coverage for some reason.

Documentation: http://php.net/manual/en/function.mb-convert-encoding.php

jancborchardt:
@rullzer @icewind1991 @nickvergessen could you as PHP pros look at this? :) Would be awesome

jancborchardt:
@icewind1991 @ChristophWurst so this is good to merge? :)

zeugmatis:
Thanks, please continue - !  

Mark

—

jancborchardt:
@zeugmatis if you like, please review the other pull requests too: https://github.com/nextcloud/mail/pulls – any contribution counts and is _very_ welcome! :tada: :)

PR #323, opened 2017-02-27
add Retry button to error views, ref #314

### LINKED ISSUE #314: »Reload« button on error views
All the error views »Couldn’t load message« or »Couldn’t load folder« should have a »Reload« button below. :)

@ChristophWurst let me know which JS it should trigger and I can try to put it in. :)

jancborchardt:
Fix #314 

@nextcloud/mail can one of you hook the button up so it actually works? :)


ChristophWurst:
I will for now close this PR as it's only a very little piece of the actually required changes. Most of it requires some proper retry-handling in our client-side code. This patch only adds a simple button to the template.

PR #324, opened 2017-02-27
prevent stacking of Aw, Wg, Fwd just like Re

jancborchardt:
We already prevent Re: stacking (repeating the same reply prefix in the subject), so this does the same for very common other prefixes.

Please review @nextcloud/mail

### DIFF
--- a/lib/Model/ReplyMessage.php
+++ b/lib/Model/ReplyMessage.php
@@ -26,6 +26,14 @@ class ReplyMessage extends Message {
 		// prevent 'Re: Re:' stacking
 		if (strcasecmp(substr($subject, 0, 4), 'Re: ') === 0) {
 			parent::setSubject($subject);
+		} else if (strcasecmp(substr($subject, 0, 4), 'Aw: ') === 0) {
+			parent::setSubject($subject);
+		} else if (strcasecmp(substr($subject, 0, 4), 'Wg: ') === 0) {
+			parent::setSubject($subject);
+		} else if (strcasecmp(substr($subject, 0, 4), 'Fw: ') === 0) {
+			parent::setSubject($subject);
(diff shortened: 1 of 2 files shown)


avanc:
It would be helpful to have the list of key words configurable.

ChristophWurst:
One does not simply not fix the unit tests when touching code :-P

```
1) OCA\Mail\Tests\Model\ReplyMessageTest::testSubjectReStacking

Failed asserting that two strings are identical.

--- Expected

+++ Actual

@@ @@

-Re: test message

+Re: Re: test message

/home/travis/build/nextcloud/core/apps/mail/tests/Model/ReplyMessageTest.php:48

2) OCA\Mail\Tests\Model\ReplyMessageTest::testSubjectReCaseStacking

Failed asserting that two strings are identical.

--- Expected

+++ Actual

@@ @@

-RE: test message

+Re: RE: test message
```

jancborchardt:
Fixed the code and tests.

@avanc making it configurable would offload the work to every single user. That’s not cool. :) Instead we should just put this whole list of the translations of `Re:` and `Fwd:` into an array to check: https://en.wikipedia.org/wiki/List_of_email_subject_abbreviations#Abbreviations_in_other_languages – do you want to maybe contribute this as a pull request? :)

avanc:
@jancborchardt I was thinking about a global configuration in e.g. config.php. But you are right, it is not core related. Your idea with an array to check sounds good. Currently, I don't have the development environment set up, but I'll come back to it later.

Gomez:
@avanc We merged the current state. Would be cool if you improve it further! Ping us if you need help.

jancborchardt:
Also opened a follow-up issue for this: https://github.com/nextcloud/mail/issues/346 – @avanc you are very welcome to contribute that as soon as you have time. :) If you need any help don’t hesitate to ask.

PR #344, opened 2017-03-27
add text to email account menu

jancborchardt:
Check out the 3-dot menu of the mail account :) now has text next to the icons like all other menus in Nextcloud.
capture du 2017-03-27 20-42-08

Please review @nextcloud/mail @nextcloud/designers

### DIFF
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -10,11 +10,21 @@
 	</ul>
 </div>
 
-<div class="app-navigation-entry-menu">
-	<ul>
-		<li><button class="icon-rename svg" title="{{ t 'Settings' }}"></button></li>
-		<li><button class="icon-delete svg" title="{{ t 'Delete account' }}"></button></li>
-	</ul>
+<div class="app-navigation-entry-menu popovermenu bubble menu">
+    <ul>
+        <li>
+            <a href="#" class="menuitem action action-settings permanent" data-action="Settings">
+                <span class="icon icon-rename"></span>
+                <span>{{ t 'Settings' }}</span>
+            </a>
+        </li>
+        <li>
+            <a href="#" class="menuitem action action-delete permanent" data-action="Delete">
+                <span class="icon icon-delete"></span>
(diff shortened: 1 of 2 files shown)


jancborchardt:
Added a screenshot above, now easier to review. :)

jancborchardt:
Easy review, please check this out @nextcloud/mail :)

js/views/accountview.js
```diff
@@ -41,9 +41,9 @@ define(function(require) {
 		},
 		ui: {
 			email: '.mail-account-email',
-			menu: 'div.app-navigation-entry-menu',
-			deleteButton: 'button[class^="icon-delete"]',
-			settingsButton: 'button[class^="icon-rename"]'
+			menu: '.app-navigation-entry-menu',
+			settingsButton: '.action-settings',
```
ChristophWurst: doesn't this change mean that clicking the text has no effect? Only the icon will have a click handler registered now.

js/templates/account.html
```diff
@@ -10,17 +10,29 @@ <h2 class="mail-account-email" title="{{emailAddress}}">{{emailAddress}}</h2>
...
+            <a href="#" class="menuitem action action-delete permanent" data-action="Delete">
+                <span class="icon icon-delete"></span>
+                <span>{{ t 'Delete account' }}</span>
+            </a>
+        </li>
+    </ul>
 </div>
 {{/if}}
 
+
```
ChristophWurst: unnecessary line breaks

ChristophWurst:
@jancborchardt ping 😉

js/views/accountview.js
jancborchardt: Nope, because .action-settings is the whole a element which spans the full width. ;)

js/templates/account.html
jancborchardt: Removed :)

jancborchardt:
Addressed your concerns and rebased to master. :)

Issue #251, opened 2016-12-17
Possibility to attach local files [$50]


jancborchardt:
(Carrying over the tasks from https://github.com/nextcloud/mail/pull/4 so we don’t forget it ;)

client-side TODO
- [ ] upload files asynchronously
- [ ] add uploaded files to list of attachments
- [ ] show loading and/or progress indicator
- [ ] lock 'Send'/'Reply' button until all local attachments have been uploaded successfully
- [ ] handle errors (remove faulty attachment from list)
- [ ] add special property to attachment so that server can distinguish between local and cloud attachments 

server-side TODO
- [ ] add attachment store for asynchronously uploaded files
- [ ] move uploaded attachments to temp dir (use core's temp manager)
- [ ] limit uploaded file size to a sane size (e.g. a few MBs)
- [ ] distinguish between local and cloud attachments when sending a message and load local attachments from the attachment store

cc @nextcloud/mail

rakshazi:
Hello!
Any updates here?

jancborchardt:
@rakshazi unfortunately not yet, no. Since you are a developer it would be awesome if you can contribute a part of this via pull request – any contribution is welcome and much appreciated! :) We need more people working on the Mail app.

ChristophWurst:
@lcalaresu FYI put a bounty on this long awaited feature 🤑

PR #383, opened 2017-06-03
Issue 251 possibility to attach local files

lcalaresu:
For issue #251.
I took the code from commit 01fcff9 (that has been developed for the former #4) and started experimenting.
Will need to discuss some points to continue and add some tasks.

_Tasks from the issue :_ 

client-side TODO
- [x] add uploaded files to list of attachments
- [x] upload files asynchronously
- [x] show loading and/or progress indicator
- [x] lock 'Send'/'Reply' button until all local attachments have been uploaded successfully
- [x] handle errors (remove faulty attachment from list)
- [x] add special property to attachment so that server can distinguish between local and cloud attachments 

server-side TODO
- [x] add attachment store for asynchronously uploaded files
- [x] move uploaded attachments to temp dir (use core's temp manager)
- [ ] limit uploaded file size to a sane size (e.g. a few MBs)
- [x] distinguish between local and cloud attachments when sending a message and load local attachments from the attachment store


lcalaresu:
OK, right now it doesn't do much but I am not sure where to go exactly.

In terms of design I am not sure it is acceptable. When we upload a local attachment, the filename and the upload progress bar is blue. It both become red if an error occurred, and goes back to black if everything went well (like the attachments from Files).  
screen shot 2017-06-05 at 11 35 55 am

Concerning the file storage, it goes to a Nextcloud temporary folder (using `OCP\ITempManager`), but I am not sure it is a very good idea because it seems those temporary files might be deleted after a while. Am I wrong?
It could be a problem if the user starts writing an email with local attachments but finishes it later (using drafts)... I am afraid the files might be lost.
Maybe using AppData or something might be a better choice?

ChristophWurst:
> In terms of design I am not sure it is acceptable. When we upload a local attachment, the filename and the upload progress bar is blue. It both become red if an error occurred. The progress bar disappear and the filename turns black if everything went well (like the attachments from Files).

cc @jancborchardt 

> Maybe using AppData or something might be a better choice?

Yes, that sounds like a better approach!

> local attachments but finishes it later (using drafts)... I am afraid the files might be lost.

IMO that's a tricky situation. I guess ideally we should attach the file to the draft message then. Of course, the implementation will be much more complex then :-|

Thanks a lot for working on this, much appreciated :)

js/models/localattachment.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @author Luc Calaresu <dev@calaresu.com>
+ *
+ * ownCloud - Mail
```
ChristophWurst: Nextcloud *cough* :-P

js/service/attachmentservice.js
```diff
@@ -66,4 +71,81 @@ define(function(require) {
...
+			data: fd,
+			processData: false,
+			contentType: false,
+		}).done(function(data) {
+			defer.resolve(localAttachmentModel, data.id);
+		}).fail(function(err) {
+			defer.reject(localAttachmentModel);
+		});
+
+		return defer.promise();
```
ChristophWurst: in meantime we've moved to native ES6 promises. Those jQuery deferreds should be replaced hence :)

js/models/localattachment.js
ChristophWurst: same applies for the other new files

ChristophWurst:
> limit uploaded file size to a sane size (e.g. a few MBs)

Tough question: what's a sane limit? Technically, it will depend on the mail server or provider, but attachments shouldn't exceed a few MBs IMO.

lcalaresu:
> Tough question: what's a sane limit?

Oh yeah, that was another question I forgot to mention!
I think many big email providers (Gmail, Yahoo, etc.) make it 25MB, but I've seen some others limiting attachments at 10MB (even on reception!) @_@

js/models/localattachment.js
lcalaresu: Oups hahaha, thanks for that! I double checked the headers, but it seems I saw what I wanted to see :stuck_out_tongue:

lcalaresu:
I've just written some Jasmine tests for this new development, it's now fully covered.
It made me realised that I used the jquery-ui progressbar for the attachment uploads... but jquery-ui is not provided as a dependency of the mail app. The code I wrote work because the library is provided by nextcloud server 😄 
However, the Jasmine tests do not want to run if I don't add JQuery-ui (custom with only the progressbar) into the vendor directory. 
I do not know if it is OK to add that library? Maybe another library of choice for the progress bar?

jancborchardt:
This is so awesome @lcalaresu! :)

The design is cool, very good! One thing I’m not sure is possible but I always liked for attachments are filetype icons or even better file previews. That would be even better to show as confirmation that it’s attached, as improvement on just the name being black. What do you think?

lcalaresu:
@jancborchardt I'm glad it's ok ;)
Concerning filetype icons or file previews, I think that could be really great indeed!! 
For that, even the existing 'attachment from Files' should be modified to be consistent, right ?

In order to keep this change simpler, I think it would be better to open a new improvement 'issue' about that.
In fact, I'm trying to touch the current 'attachment from files' as little as possible, don't want to break something in there 😝

lib/Service/Attachment/UploadedFile.php
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * @param array $fileData
+	 */
+	public function __construct($fileData) {
+		$this->fileData = $fileData;
+		// move the php tmp file to a nextcloud temporary folder
+		$tmp_folder = \OC::$server->getTempManager()->getTemporaryFolder();
+		$this->oc_tmp_path = $tmp_folder.$fileData['name'];
+		move_uploaded_file($fileData['tmp_name'], $this->oc_tmp_path);
```
ChristophWurst: @rullzer is it save to assume the temp folder being an a local fs? Otherwise this won't work I guess.

jancborchardt:
Sure, let's keep that separate. :)

lib/Service/Attachment/UploadedFile.php
lcalaresu: @ChristophWurst that part does not work on my server anyway, I'm going to delete it.
I wanted to use the AppData class to store the attachments, but I got a bit lost in the nextcloud/server code :roll_eyes: 
Do you know if there is a document for that class? 
Or maybe just an obvious example in another nc app...

ChristophWurst: AFAIK the scss caching stores compiled css into the app data directory.

ChristophWurst: You can find some related code here https://github.com/nextcloud/server/search?q=AppData ;-)

ChristophWurst: Let me know if you need help with this PR. I'd be happy to help finishing this :)

lcalaresu: Actually my hard drive died last week, I'm in the middle of reinstalling everything @_@
So, any help on this PR is welcome :smile: 

I think the JS part (code + tests) is almost done, except for the file size limitation.

It is almost only about PHP now. I think:
- move the uploaded file somewhere in filesystem,
- a big "if" to distinguish local vs nc attachments when sending the email,
- maybe check on the PHP source code: PHP is not my cup of tea, I might have done smtg wrong :smiley_cat:

js/models/localattachment.js
```diff
@@ -0,0 +1,33 @@
...
+	'use strict';
+
+	var Attachment = require('models/attachment');
+
+	var LocalAttachment = Attachment.extend({
+		defaults: {
+			progress: 0,
+			uploadStatus: 0  /* 0=pending, 1=ongoing, 2=error, 3=success */
+		},
+		initialize: function() {
```
ChristophWurst: unused. remove this and then the parent constructor will be called automatically for you.

js/require_config.js
```diff
@@ -28,7 +28,8 @@
 			ical: 'vendor/ical.js/build/ical.min',
 			marionette: 'vendor/backbone.marionette/lib/backbone.marionette',
 			underscore: 'vendor/underscore/underscore',
-			text: 'vendor/text/text'
+			text: 'vendor/text/text',
+			'jquery-ui': 'vendor/jquery-ui/ui/minified/jquery-ui.custom.min'
```
ChristophWurst: those libs are actually sorted alphabetically, hence this line should be moved up.

js/service/attachmentservice.js
```diff
@@ -66,4 +71,88 @@ define(function(require) {
...
+							progressCallback.bind(localAttachmentModel),
+							false);
+					}
+					return customXhr;
+				},
+				data: fd,
+				processData: false,
+				contentType: false
+			})
+			.done(function(data) {
```
ChristophWurst: just use `return $.ajax(…);` inside the promise construction function and it will be wrapped automatically ;-)

js/service/attachmentservice.js
```diff
@@ -66,4 +71,88 @@ define(function(require) {
...
+				resolve(data.id);
```
ChristophWurst: you could then just chain the promise with `.then(function(data) { return data.id; })`

js/service/attachmentservice.js
```diff
@@ -66,4 +71,88 @@ define(function(require) {
...
+				reject();
+			});;
+		});
+	}
+
+	/**
+	 * This method is called when a local attachment upload should be aborted.
+	 * If there is no upload ongoing, this method has no effect.
+	 *
+	 * @param {function} localAttachmentModel
```
ChristophWurst: isn't it an `Attachment`?

js/service/attachmentservice.js
```diff
@@ -66,4 +71,88 @@ define(function(require) {
...
+	 */
+	function abortLocalAttachment(localAttachmentModel) {
+		var uploadRequest = localAttachmentModel.get('uploadRequest');
+		if (uploadRequest && uploadRequest.readyState < 4) {
+			uploadRequest.abort();
```
ChristophWurst: Would make sense to unset the `uploadRequest` property here if we don't need it anymore.

ChristophWurst: nvm, I guess that's what happens on L149

js/service/attachmentservice.js
```diff
@@ -66,4 +71,88 @@ define(function(require) {
...
+	 * successfully finished. The server returned the db attachment id.
+	 *
+	 * @param {function} localAttachmentModel
+	 * @param {number} filedId
+	 */
+	function uploadLocalAttachmentFinished(localAttachmentModel, fileId) {
+		if (fileId === undefined || localAttachmentModel.get('progress') < 1) {
+			localAttachmentModel.set('uploadStatus', 2);  // error
+		}
+		else {
```
ChristophWurst: move else keyword to previous line please, like `} else {`

js/service/attachmentservice.js
```diff
@@ -66,4 +71,88 @@ define(function(require) {
...
+			uploadRequest.abort();
+		}
+		localAttachmentModel.collection.remove(localAttachmentModel);
+	}
+
+	/**
+	 * This method is called when a local attachment upload has
+	 * successfully finished. The server returned the db attachment id.
+	 *
+	 * @param {function} localAttachmentModel
```
ChristophWurst: same here, it's a model and not a function

js/tests/views/attachment_spec.js
```diff
@@ -0,0 +1,124 @@
...
+			it ('should exist', function () {
+				expect(view).toBeDefined();
+			});
+
+			it ('should have a progress bar set at 0', function () {
+				// jquery-ui progressbar is initialised
+				expect(view.ui.attachmentName.attr('role')).toBe('progressbar');
+				// value of progressbar is set to 0
+				expect(view.ui.attachmentName.attr('aria-valuenow')).toBe('0');
+    	});
```
ChristophWurst: indentation is off

js/tests/views/attachment_spec.js
```diff
@@ -0,0 +1,124 @@
...
+
+			it ('should allow attachment removal', function () {
+				/* We just test that the 'upload:abort' event has been sent */
+				/* The deletion action itself is tested on the attachmentService tests */
+				spyOn(Radio.attachment, 'request');
+				view.removeAttachment();
+				expect(Radio.attachment.request).toHaveBeenCalledWith('upload:abort', model);
+    	});
```
ChristophWurst: same

ChristophWurst: and the ones below too ;-)

lib/Service/Attachment/UploadedFile.php
ChristophWurst: Let's continue in https://github.com/nextcloud/mail/pull/408. I've resolved the conflicts, cleaned-up the code, added app data support and wrote a bunch of tests.

Issue #387, opened 2017-06-11
Issues running phpunit tests


lcalaresu:
I experienced the same problem as on nextcloud/server#4440.
Running phpunit tests on my machine threw lots of `__construct() must implement interface OCP\IDBConnection`. 

I tried to follow exactly the same steps as the travis build, but it's impossible for me to make it work.

Thanks to georgehrke on irc, I got the correction quickly. It seems as simple as calling the parent setUp() on 2 test classes... However, I do not know why it seems to happen to some people only (same on nextcloud server).

### Steps to reproduce
1. Just run `phpunit --configuration phpunit.xml`

### Expected behaviour
Tests should run fine

### Actual behaviour
Tell us what happens instead

### Mail app

**Mail app version:** master (as of 2017-06-11)

**Mailserver or service:** christophwurst/owncloud-mail-test-docker

### Server configuration
**Operating system**: Debian Jessie

**Web server**: Nginx

**Database:** SQLite

**PHP version:** 7.0/7.1

##### PhpUnit log
<details>
Repeated 60 times:

```
TypeError: Argument 1 passed to OC\BackgroundJob\JobList::__construct() must implement interface OCP\IDBConnection, null given, called in /home/luc/dev/core/lib/private/Server.php on line 508
```
</details>

ChristophWurst:
Fixed by #388 :)

PR #388, opened 2017-06-11
Added missing parent::setUp() that made the PHPUnit tests to fail

lcalaresu:
For nextcloud/mail#387

Signed-off-by: Luc Calaresu <dev@calaresu.com>


ChristophWurst:
btw @lcalaresu, since you're a member of this repo you may also push your branches here directly instead of your private fork :)

lcalaresu:
@ChristophWurst OK will do! :yum:

PR #389, opened 2017-06-12
Install phpunit via composer

ChristophWurst:
This way we don't have to rely on a globally installed phpunit, especially since some projects now use 6.x whereas this one doesn't due to incompatibilities.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,20 +43,12 @@ cache:
 before_install:
   - php --info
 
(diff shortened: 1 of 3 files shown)


ChristophWurst:
CI is happy and so am I. Merging without a review because this is jus a minor change and unlikely to break stuff.

lcalaresu:
Well, thank you for that one, it sure helps!

Issue #320, opened 2017-02-23
JS error when adding a second account


ChristophWurst:
While testing https://github.com/nextcloud/mail/pull/271 I noticed that sometimes a js error is thrown after adding a second account. Once I reloaded the page, everything worked like a charm, so I suspect it's just the transition that has some issues.

jancborchardt:
Same here:

> Error while loading the accounts.
> TypeError: Cannot read property 'first' of undefined

Then I end in this state:
![capture du 2017-03-29 17-28-21](https://cloud.githubusercontent.com/assets/925062/24462782/3404ea80-14a5-11e7-8240-f4fb49414c83.png)

ChristophWurst:
Fix is in https://github.com/nextcloud/mail/pull/390

PR #390, opened 2017-06-12
Fix adding a new account

ChristophWurst:
Apparently Backbone.Collection.fetch resolves its promise with
the raw data returned by the ajax call. Since the consuming code
relied on getting proper Account models, it failed when a new
account was added. This fix 'converts' the promise so that it
resolves with a collection of Account models.

Found while working on https://github.com/nextcloud/mail/pull/298.

Fixes https://github.com/nextcloud/mail/issues/320

### DIFF
--- a/js/service/accountservice.js
+++ b/js/service/accountservice.js
@@ -64,7 +64,17 @@ define(function(require) {
 			return Promise.resolve(accounts);
 		}
 
-		return Promise.resolve(accounts.fetch());
+		return new Promise(function(resolve, reject) {
+			accounts.fetch({
+				success: function() {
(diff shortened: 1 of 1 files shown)


lcalaresu:
Do not know why the `VersionEye` check is not happy recently though...

ChristophWurst:
> Do not know why the VersionEye check is not happy recently though...

Because life would be just too easy if there weren't random CI failures ;-)

I think their service doesn't like composer packages that aren't served via packagist. Hence all Horde packages lack a proper license. I'll ignore it for now. If this problem doesn't magically disappear, I'll remove the check.

Thanks for your review :)

PR #326, opened 2017-02-28
Refactor folder sync

ChristophWurst:
~~Partially~~ mostly implements #318
Fixes https://github.com/nextcloud/mail/issues/244, #318, https://github.com/nextcloud/mail/issues/335, https://github.com/nextcloud/mail/issues/290

- Sync
  - [x] Use horde sync to get new/changed/vanished messages
  - [x] Implement message mapper (reuse old code if possible)
  - [x] Implement sync mechanism in the client
  - [x] Implement sync for the unified inbox
  - [x] Update unified inbox if individual ones change
  - [x] Implement sync for search mailboxes – should work now *fingers crossed*
  - [x] Implement pagination for search mailboxes
  - [x] test search
- Pagination
  - [x] Implement date-based cursor-based pagination on the server
  - [x] Implement new pagination on the client
  - [x] Implement pagination for the unified inbox

- Necessary follow-ups
  - Kill background sync. It's extremely buggy in its current form. If the new horde sync approach is fast enough, I'm planning to just let the current folder view refresh its content once a minute. If the folder is a specific inbox, we refresh the unified one. This should cover everything.

PR #397, opened 2017-06-19
Fix Folder::jsonSerialize warning due to undefined index

ChristophWurst:
This makes the serialization a bit more error tolerant as in status
values not being set does not result in PHP errors anymore but fall
back to sensible default values.

I cannot reproduce https://github.com/nextcloud/mail/pull/326#issuecomment-308420686 but this should fix it.


lcalaresu:
@ChristophWurst 
I cannot see two of the three errors anymore :smiley: 
I still have an 'undefined index: unseen ' error on line 161 of the same file though.

Actually, this error shows every time I refresh the whole "/apps/mail" page.

ChristophWurst:
@lcalaresu fixed :)

PR #402, opened 2017-06-26
Do not load/show 'noSelect' folders

ChristophWurst:
Once there was some logic in place which showed an error message, but let's just not show those folders.

Ref https://github.com/nextcloud/mail/pull/326#issuecomment-310191688

### DIFF
--- a/css/mail.css
+++ b/css/mail.css
@@ -30,6 +30,9 @@
 	background-color: rgb(240, 240, 240);
 }
 
+.folder.no-select {
+	cursor: not-allowed;
+}
 
 /* SETUP */
 #setup {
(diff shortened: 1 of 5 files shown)


lcalaresu:
@ChristophWurst  Sorry for the dumb question but... what are 'noSelect' folders? 
 I don't have access to my Gmail account right now, but maybe I can review that PR with another mail account...

ChristophWurst:
> what are 'noSelect' folders?

Good question. From https://www.ietf.org/rfc/rfc2060.txt

> If these levels of hierarchy are
      not also selectable mailboxes, they are returned with the
      \Noselect mailbox name attribute (see the description of the LIST
      response for more details).

So those are mailboxes containing other ones, but cannot be used like one. I've only seen that with GMail so far I think.

ChristophWurst:
@lcalaresu thanks a lot for testing and reviewing!

Issue #392, opened 2017-06-14
Memcache breaks default user


fsedarkalex:
### Steps to reproduce
1. use current nightly with default mailbox config
2. enable a memcache (tried memcached and APCu)
3. Inbox of default mailbox is no longer loading (very long unsuccessful loading)
4. disable all memcaches ->works again

### Expected behaviour
Load Mailbox as usual

### Actual behaviour
loading for ages then error opening mailbox (Fehler beim Laden des ausgewählten Kontos.)

### Mail app

**Mail app version:** (see apps admin page)
Nightly 14.06.2017

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
own mail service (dovecot)

**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 
STARTTLS

**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)
STARTTLS

**Number of accounts:**
1 or 2 (default is broken and tried a configured one - that works)

**Mail app version build date:** (only if you are using a Nightly Build)
14.06.2017

### Server configuration
**Operating system**:
Debian Jessie

**Web server:**
Apache 2

**Database:**
MariaDB

**PHP version:**
7.1

**Version:** (see admin page)
12

**Updated from an older version or fresh install:**
11.x

**List of activated apps:**
many... don't think this is relevant

**The content of config/config.php:**

```
<?php
$CONFIG = array (
  'instanceid' => '',
  'passwordsalt' => '',
  'secret' => '',
  'trusted_domains' => 
  array (
    0 => '',
  ),
  'datadirectory' => '/owncloud/storage/',
  'overwrite.cli.url' => '',
  'htaccess.RewriteBase' => '/',
  'check_for_working_wellknown_setup' => false,
  'check_for_working_webdav' => false,
  'check_for_working_htaccess' => false,
  'dbtype' => 'mysql',
  'version' => '12.0.0.29',
  'dbname' => '',
  'dbhost' => '',
  'dbtableprefix' => 'oc_',
  'dbuser' => '',
  'dbpassword' => '',
  'logtimezone' => 'UTC',
  'installed' => true,
  'mail_from_address' => '',
  'mail_smtpmode' => 'php',
  'mail_domain' => '',
  'ldapIgnoreNamingRules' => false,
  'upgrade.disable-web' => false,
  'maintenance' => false,
  'loglevel' => 2,
[...]

PR #403, opened 2017-06-26
Fix missing use-statment for the Cache class

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/392

cc @fsedarkalex


ChristophWurst:
> Is it "fixed" on the refactoring branch? Coz I couldn't reproduce earlier on that branch...

nope, just here

PR #412, opened 2017-07-11
Update ES6 Promises, Handlebars and RequireJS

ChristophWurst:
This updates some of our js libs to their latest version. Did a quick test and it seems everything works like before: promises are resolved, views are rendered and js files are loaded.


ChristophWurst:
Thanks!

PR #426, opened 2017-08-05
Add IMAP integration test framework

ChristophWurst:
Add a trait-based testing framework that automatically resets
the IMAP testing account before a test case is run to prevent
side effects and dependencies across different tests.
A simple self test was added to ensure the functionality works
as expected.
A new super class for integration tests was added which should
be used now instead of phpunit's class. This will allow to write
tests with less boilerplate code, as we can for example check
the usage of the IMAP test trait and reset accounts automatically.
This was inspired by the Laravel testing capabilities.

### DIFF
--- a/tests/Integration/Framework/ImapTest.php
+++ b/tests/Integration/Framework/ImapTest.php
@@ -1,0 +1,181 @@
+<?php
+
+/**
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * Mail
+ *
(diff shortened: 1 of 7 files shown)


ChristophWurst:
From https://scrutinizer-ci.com/g/nextcloud/mail/inspections/1030e0d2-991c-4181-8d8e-1181fdeb92ac

> Test Coverage has increased to 60% (+3%). 

🤖 🚀 🕺

ChristophWurst:
Only test files affected and CI passes -> merging.

PR #428, opened 2017-08-07
Refactor and test connectivity test classes

ChristophWurst:
* Use composition instead of inheritance
* Unit test classes
* Mock connections to external servers


ChristophWurst:
> Test Coverage has increased to 62% (+2%). 

Mission accomplished -> merging.

PR #429, opened 2017-08-07
Add more unit tests for the autoconfig logic

ChristophWurst:
As part of my long-time goal to reach a higher testing coverage, I took a closer look at https://scrutinizer-ci.com/g/nextcloud/mail/code-structure/master/code-coverage/lib/Service/AutoConfig/ and tested the simple classes. Some of the tests aren't 100% pure unit tests as they connect to external resources, but still better than no unit tests.


ChristophWurst:
> Test Coverage has increased to 64% (+3%). 

No functional changes, only test files were added and CI passes -> merging without review.

PR #430, opened 2017-08-07
Ignore template coverage as we don't test them

ChristophWurst:
This should increase the coverage percentage a bit.


ChristophWurst:
Looks like Scrutinizer doesn't care about the updated config

> No notable changes
>
>There were no notable changes in the changeset which was inspected.

-> Merging

PR #431, opened 2017-08-07
Fix test base class and namespace

ChristophWurst:
Just some minor cleanups of our test code.


ChristophWurst:
No functional changes and CI passes -> merging.

PR #432, opened 2017-08-07
Fix IMAP tests

ChristophWurst:
IMAP integration tests failed because the 'Drafts' mailbox
is not part of the automatically created ones, hence it was
removed now.

Hopefully, this fixes failing CI runs in https://github.com/nextcloud/mail/pull/326 🙏


ChristophWurst:
If anyone wonders how to reproduce the bug: reset your local IMAP docker and only run the SelfTest. It will fail as 'Drafts' has not been created yet. I think it's a byproduct of running other tests.

ChristophWurst:
CI happy -> merging.

PR #437, opened 2017-08-14
Use a common base class for all tests

ChristophWurst:
This will easy the migration to a newer phpunit (the class is now namespaced) and furthermore we don't rely on Nextcloud server code.


ChristophWurst:
> Much easier to understand! 👍

Thanks for your review. Much appreciated 😁

PR #438, opened 2017-08-14
Add integration tests for outgoing messages with attachments

ChristophWurst:
Yet another (simple) test to verify message transmission works correctly, also when attaching both local and cloud files.
This also adds another convenience trait to create a test user during (integration) tests.


ChristophWurst:
All tests passed an coverage remained the same -> good to merge

Issue #9, opened 2016-08-29
Issues with drafts [$65]


jancborchardt:
There are a bunch of things problematic with drafts, and we should fix them. Cause drafts are a pretty helpful feature to save your stuff and sync it.

Sorted by severity:
- [x] **For each email, a bunch of drafts are often saved and the previous one is not deleted**
- [x] When an email is sent, the drafts of that email are not deleted
- [x] The subject isn’t saved
- [x] Special characters like umlauts and `>` is escaped which is horrible for quoted text

For later:
- Drafts of replies are not shown automatically in the reply box when opening the mail again → moved to https://github.com/nextcloud/mail/issues/459
- Attachments are not saved → moved to https://github.com/nextcloud/mail/issues/458

koreywithak:
I use the Mail app with my Gmail account. I notice a few additional issues:
1. When clicking on the [Gmail] subfolder and then Drafts, it will not load the draft e-mails. It does, however, load other subfolders, like Sent.
2. When working on a new message or reply, it does save the draft into Drafts (although I can't view it in with the Nextcloud Mail app, due to above issue). However, it ends up saving multiple copies over the course of e-mail composition.

keunes:
I created an IMAP account with my hoster, and in their webmail (Roundcube) if I save a draft without a recipient, then Nextcloud Mail will say: "Something's wrong with loading your drafts :/" (manual translation). After adding a recipient in my hoster's Roundcube, Nextcloud Mail _does_ load my drafts properly.

jrullo:
Is this solved? I have close to 200 draft emails in my mail app on nextcloud instance. I've replied to less than 10 mails that I can recall while using the mail app. I have no idea why so many drafts would be saved. I usually just read the mail and delete the ones I don't need.

jancborchardt:
@ChristophWurst @irgendwie @Gomez can anyone of you look into this? :) It’s one of the most annoying bugs I encounter daily. And it actually punishes the people who use the mail app a lot. :\

lcalaresu:
Actually, no drafts are even saved on my environment using the master branch (it's working with the last stable version). The console gives me the message:

> Uncaught (in promise) TypeError: Cannot read property 'draftUID' of undefined

It seems "`this`" has been used inside an anonymous function in `js/views/composerview.js` (lines 397 & 400) ; maybe use `_this` instead?

Does anyone reproduce this? It seems that these changes in the composerview are a bit old... Or maybe nobody uses the drafts 😄 
[...]

PR #439, opened 2017-08-14
Remove wrong abstraction in IAccount and use impl instead

ChristophWurst:
This removes a minor flaw in our OOP design. Those two methods are only implemented for a specific account, hence they shouldn't be part of the interface. In fact, the mail transmission logic even assumed to get an instance of `Account` and the impl in `UnifiedAccount` would throw an exception anyway.

Found while digging into the drafts code for https://github.com/nextcloud/mail/issues/9.

### DIFF
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -321,6 +321,7 @@ class AccountsController extends Controller {
 
 		// create transport and save message
 		try {
+			/* @var $account Account */
 			$newUID = $account->saveDraft($message, $uid);
 		} catch (Horde_Exception $ex) {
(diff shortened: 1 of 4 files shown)


ChristophWurst:
FYI the `Account` class is by far the most complex class we have in our code base. It's tightly coupled to several other classes and does not leverage the advantages of IoC/DI and hence can only be tested partly (just look at the constructor and see how dependencies are located statically). Still it is one of the most used classes spread over the full code base. As a long time goal we should try to break it into functional blocks (something like the mail transmission service that I created).

ChristophWurst:
Minor change, CI is happy and no feedback within a week -> merging.

PR #445, opened 2017-08-21
Use another IMAP docker image that allows more connections

ChristophWurst:
Since CI builds of #326 failed and I could reproduce this locally,
it turned out that running single tests or tests of a single directory
was no problem. However, executing all tests at once led to failing
IMAP connections. Apparently an increased number of allowed IMAP
connections fixes the problem (locally).


ChristophWurst:
As expected, CI builds pass as before. Will merge this now to make some progress with #326.

PR #462, opened 2017-08-24
Remove some unneeded containers from the HTML

jancborchardt:
A little bit cleanup which should not interfere with the Backbone/Marionette stuff. Please check @ChristophWurst @Gomez @lcalaresu


css/mail.css
```diff
@@ -148,10 +148,8 @@
 	z-index: 1;
 }
 
-#app-navigation > ul {
+#app-navigation #app-navigation-accounts {
```
ChristophWurst: I guess one id is enough ;)

jancborchardt: Fixed! :)

PR #464, opened 2017-08-24
Better html mails

Gomez:
Update HTMLPurifier, add CSSTidy. Add styles tag back on html mails


lib/Service/Html.php
```diff
@@ -129,11 +133,22 @@ public function sanitizeHtmlMailBody($mailBody, array $messageParameters, Closur
...
 
 		$result = $purifier->purify($mailBody);
 		// eat xml parse errors within HTMLPurifier
 		libxml_clear_errors();
+
+		// Add back the style tag
+		$styles = $purifier->context->get('StyleBlocks');
+		if ($styles) {
+			$result = implode("\n", array(
+				'<style type="text/css">',
```
ChristophWurst: @LukasReschke anything scary about this? Is it safe? :)

Gomez: AFAIK it runs through CSSTidy, should be good :)

Gomez: There is a warning: http://htmlpurifier.org/live/configdoc/plain.html at "Filter.ExtractStyleBlocks". @LukasReschke imagecrash attack?

ChristophWurst: according to @LukasReschke this change is okay as long as we're iframing the message.

Issue #463, opened 2017-08-24
Further avatar sources [$50]


jancborchardt:
Most often my inbox looks like this:
![screenshot from 2017-08-24 18-11-46](https://user-images.githubusercontent.com/925062/29676231-c3792fca-88f7-11e7-9738-1b4a2b425226.png)![screenshot from 2017-08-24 18-11-32](https://user-images.githubusercontent.com/925062/29676233-c399381a-88f7-11e7-9d3e-81f99871a997.png)![screenshot from 2017-08-24 18-11-19](https://user-images.githubusercontent.com/925062/29676234-c39b373c-88f7-11e7-8a53-dee13a4ab1db.png)
And so on, you get the point.

Our flow looks like this:
1. Check for avatars from the Contacts app (if installed and if a picture is present for the contact)
2. Show a placeholder

The problem is that many of the emails we get are not necessarily from people in our contacts. We should improve it like this:
1. Check for avatars from the Contacts app (if installed and if a picture is present for the contact)
2. Check for avatars from federated Nextcloud servers
3. Check https://en.gravatar.com (yes this might be a privacy concern, but we could offer a setting to deactivate this)
4. Check the domain for a favicon we can use, which is good especially for notifications from Twitter, Github, Facebook, Meetup etc. (this could also have a setting to deactivate since it pings other services)
5. Show a placeholder

Any other ideas?

jakobsack:
Current offline WIP:

- [x] Contacts app
- [ ] Federated servers
- [x] Gravatar
- [ ] Favicon

jancborchardt:
@jakobsack aaaawesome! :)

ChristophWurst:
Fixed by #469

jancborchardt:
Everyday usage looks soooo much better now! Thanks @jakobsack! :)

Issue #236, opened 2016-12-14
More sources to get avatars from [$15]


jancborchardt:
Oftentimes the avatars in the inbox look like this:
![capture du 2016-12-14 02-11-14](https://cloud.githubusercontent.com/assets/925062/21165944/bc048644-c1a2-11e6-937b-a3294f6489b0.png) ![capture du 2016-12-14 02-10-47](https://cloud.githubusercontent.com/assets/925062/21165945/bc1c720e-c1a2-11e6-8567-cb2e6e7f69f5.png) ![capture du 2016-12-14 02-10-21](https://cloud.githubusercontent.com/assets/925062/21165946/bc1fce4a-c1a2-11e6-9c4e-f648d8fe64cf.png)

That’s really sad but unavoidable. You’re not going to have all of these mails in your Contacts, especially not from notifications of Twitter or Meetup or Facebook or Wikimedia, DHL, Amazon, etc blabla.

So how about when there’s no avatar we try to get something from another source?
1. Contacts app / federation (for people you know) cc @schiessle 
2. Gravatar (if people use that)
3. Favicon of the domain (especially good for notifications)
4. More …?

cc @BernhardPosselt @nextcloud/news since you fetch the favicons of sites in the News app.

jancborchardt:
Getting the favicon of the domain as a fallback at least for popular services like Twitter notifications, Amazon mails, Github notifications, Instagram etc. would already make it a lot nicer. :)

ChristophWurst:
@jancborchardt dude, again you filed a duplicate of your own issue #463 🙈 

But well, now @jakobsack gets $50+$15 🚀

ChristophWurst:
Fixed by #469

PR #469, opened 2017-08-25
Fetch avatars from different sources

jakobsack:
This implements parts of #463 and https://github.com/nextcloud/mail/issues/236
When you open mail/avatars?email=test@example.com you get a JSON response 
{ url: $url, source: $source }.
If no image has been found $url is null.


js/models/messagecollection.js
```diff
@@ -21,6 +21,20 @@ define(function(require) {
 		model: Message,
 		comparator: function(message) {
 			return message.get('dateInt') * -1;
+		},
+		initialize: function() {
+			this.on('add', function(message) {
```
ChristophWurst: in regards to consistent coding style and testability I'd suggest to move this anonymous callback to a method of this class and just have the wiring in the constructor ;)

js/models/messagecollection.js
```diff
@@ -21,6 +21,20 @@ define(function(require) {
...
+				var url = OC.generateUrl('apps/mail/avatars?email={email}', {
+					email: message.get('fromEmail')
+				});
+
+				Promise.resolve($.ajax(url)).then(function(avatar) {
+					if (avatar.source != 'none'){
+						message.set('senderImage', avatar.url);
+					}
+				});
```
ChristophWurst: please add a `.catch(console.error.bind(this))` at the end so that errors are logged to the console

js/models/messagecollection.js
```diff
@@ -21,6 +21,20 @@ define(function(require) {
...
 			return message.get('dateInt') * -1;
+		},
+		initialize: function() {
+			this.on('add', function(message) {
+				var url = OC.generateUrl('apps/mail/avatars?email={email}', {
+					email: message.get('fromEmail')
+				});
+
+				Promise.resolve($.ajax(url)).then(function(avatar) {
+					if (avatar.source != 'none'){
```
ChristophWurst: use `!==` instead

js/models/messagecollection.js
```diff
@@ -21,6 +21,20 @@ define(function(require) {
 		model: Message,
 		comparator: function(message) {
 			return message.get('dateInt') * -1;
+		},
+		initialize: function() {
+			this.on('add', function(message) {
+				var url = OC.generateUrl('apps/mail/avatars?email={email}', {
```
ChristophWurst: would be great to have this logic in a service of its own because then we can test this more easily ;)

lib/Controller/AvatarsController.php
```diff
@@ -0,0 +1,68 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Mail\Controller;
+
+use Exception;
```
ChristophWurst: unused?

lib/Controller/AvatarsController.php
```diff
@@ -0,0 +1,68 @@
...
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 *
+	 * @param string $email
+	 * @return Response
+	 */
+	public function show($email) {
+		// close the session to allow parallel downloads
+		$this->session->close();
```
ChristophWurst: I don't think this is necessary if you don't annotate the controller with `@UseSession`. cc @LukasReschke

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,149 @@
...
+		// No avatar found. Return 204 (no content)
+		return [
+			'email' => $email,
+			'source' => 'none',
+			'url' => null
+		];
+	}
+}
+
+__halt_compiler();
```
ChristophWurst: 😮

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,149 @@
...
+use Exception;
+use OCA\Mail\Service\ContactsIntegration;
+use OCP\IUserManager;
+use OCP\IAvatarManager;
+use OCP\IURLGenerator;
+use OCP\Http\Client\IClientService;
+use Gravatar\Gravatar;
+use Mpclarkson\IconScraper\Scraper;
+
+class AvatarService {
```
ChristophWurst: I'd really like to have this class tested. In its current form it's a bit difficult to do so. May I suggest to refactor this class into three classes for the individual avatar sources and just combine their results in this class? I can also take care of this, just let me know :)

lib/Service/AvatarService.php
```diff
@@ -0,0 +1,149 @@
...
+		$gravatar = new Gravatar(['size' => 128], true);
+		$gravatarUrl = $gravatar->avatar($email, ['d' => 404]);
+
+		$client = $this->clientService->newClient();
+		$foundGravatar = true;
+		try {
+			$response = $client->get($gravatarUrl);
+		}
+		catch (\Exception $exception) {
+			$foundGravatar = false;
```
ChristophWurst: just `return null;` here and remove the `$foundAvatar` stuff ;)

jakobsack:
The avatars are now cached in `data/$user/mail/avatars`. The images are loaded right from the beginning if possible.

ChristophWurst:
@jakobsack what's your current status? The title and the tags are a bit contradictory :-P

Btw, there seem to be conflicts with the composer deps.

jakobsack:
Sorry, I bricked my laptop last Friday. I'll try to add the last adjustments during this week.

ChristophWurst:
> Sorry, I bricked my laptop last Friday. I'll try to add the last adjustments during this week.

Take your time :)

jancborchardt:
@jakobsack any update here or any way we can help? :)

jakobsack:
@ChristophWurst @Gomez Can you test the latest changes please? And tell me why composer won't work on travis? On my machine everything works fine :-/

jakobsack:
I don't know why the MySQL version fails but it seems not to be related to my changes.

Gomez:
Great work, i really like it! 
Needs a rebase & small jslint fix.

We should discuss if we make it opt-in as it use gravatar and fetches data from external server.

I would do opt-in. Cloud we prefill the cache with the major domains (like gmail, web.de)?

ChristophWurst:
> ChristophWurst self-assigned this 8 days ago
> @ChristophWurst ChristophWurst added to SELECTED in Christoph's Tasks 8 days ago

My plan was/is to work on this as soon as I've finished the release of v0.7.4.

lib/AppInfo/Application.php
```diff
@@ -57,9 +58,13 @@ private function initializeAppContainer() {
 			$user = $container->query("UserId");
 			return $container->getServer()->getUserFolder($user);
 		});
+
 		$container->registerParameter("testSmtp", $testSmtp);
 		$container->registerParameter("referrer", isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : null);
 		$container->registerParameter("hostname", Util::getServerHostName());
-	}
 
+		$container->registerService('avatarStorage', function($c) {
```
ChristophWurst: - [ ] TODO: not necessary

lib/Http/AvatarDownloadResponse.php
```diff
@@ -0,0 +1,55 @@
...
+	 * contains the passed string
+	 * Additionally the response will be cacheable by browsers. Since the content is
+	 * generally not sensitive content (e.g. Logos in mails) this should not be a problem.
+	 * @param IFile $content the content that should be written into the file
+	 */
+	public function __construct($content){
+		parent::__construct('avatar', 'application/octet-stream');
+		$this->content = $content;
+
+		$expires = new \DateTime('now + 2 hours');
```
ChristophWurst: - [ ] TODO: use `$this->cacheFor(…)` instead

ChristophWurst:
The server-side implementation *should* ™️ be finished now. Let's see if CI passes, then I'll continue with the client-side. Hopefully I'm able wo wrap it up today.

ChristophWurst:
We might have a performance issue with this. To know whether we've got an avatar for a given sender can take some time. Obviously, contacts in the address book can be looked up quickly, but searching gravatar and especially scraping the favicon is slow. And we have to do it for each message that's being loaded.
Imagine a setup with three accounts. On page load, 3 * 20 = 60 messages are being loaded.

If we check on the server-side, it takes noticeable time to fetch the list of messages.

If we let the client check, it means that 60 requests hit the server, which also takes some time, plus it blocks any further requests due to http's pipelining and thus prevents the message content to load quickly as its request is sent *after* the ones for the avatars have been dispatched.

Of course, we do a lot of caching and thus subsequent request will be faster, but still this is a issue I wouldn't ignore.

For now, I'll try to finish the implementation without solving this problem. Just so that we know how well it generally works – e.g. how many domains will it actually find an avatar?

Later, but before integrating this into master, we have to find a solution for the perf issue.

ChristophWurst:
So this now works, to some extend. At least these issues are left:

- [ ] Pretty bad performance - see comment above
- [ ] Probably security issues due to blindly piping external resources into the browser withouth checking the content/MIME

ChristophWurst:
Alright, IMO we've reached a first working version that would be ready to integrate. To improve performance this relies on two levels of caching: HTTP and the internal cache factory. That means that we trust the browser to store avatar HTTP responses to speed up page loading. If the app is used on multiple devices, the internal cache will prevent the avatar detection logic to run multiple times as it caches the result. To have the latter one working, a working memcache is required, but Nextcloud will warn admins anyway if none is configured and thus we can hope that most installations will have one: https://docs.nextcloud.com/server/12/admin_manual/configuration_server/caching_configuration.html.

Following the famous advice *premature optimization is the root of all evil* I'm gonna stop here and do any improvements in follow-up PRs, but only if we even see a demand for it.

Still, some ideas for that are: have a avatar service on the client-side that aggregates avatar loading and de-duplicates the requests. This is especially useful where you have many messages from the same sender and thus the current logic would send multiple requests.

ChristophWurst:
> Still, some ideas for that are: have a avatar service on the client-side that aggregates avatar loading and de-duplicates the requests. This is especially useful where you have many messages from the same sender and thus the current logic would send multiple requests.

Plus, we should check whether `senderImage` is already set and do nothing in that case.

ChristophWurst:
FYI: setting the yolo-no-review-but-merging-anyway-timeout to 2 weeks 😉

ChristophWurst:
> Still, some ideas for that are: have a avatar service on the client-side that aggregates avatar loading and de-duplicates the requests. This is especially useful where you have many messages from the same sender and thus the current logic would send multiple requests.

And we could add a background job that pre-poluates the cache 😉

jancborchardt:
@ChristophWurst will start reviewing now, meanwhile Fyi there are conflicting files. (And the package-lock.json should be in .gitignore maybe?)

ChristophWurst:
> meanwhile Fyi there are conflicting files.

Thanks. Will take care of it.

> And the package-lock.json should be in .gitignore maybe

Nope, that makes sure everyone installs the exact same version of a dependency and additionally also speeds up the installation of npm packages: https://docs.npmjs.com/files/package-lock.json

jancborchardt:
So I did `make install-composer-deps`, but then on the next it fails like this:
```
jan@Rechenknecht:~/nextcloud/apps/mail$ make optimize-js 
npm install --production
npm WARN grunt-jscs@1.8.0 requires a peer of grunt@~0.4.2 but none is installed. You must install peer dependencies yourself.

up to date in 2.901s
./node_modules/webpack/bin/webpack.js --config js/webpack.prod.config.js
module.js:515
    throw err;
    ^

Error: Cannot find module 'json-stable-stringify'
    at Function.Module._resolveFilename (module.js:513:15)
    at Function.Module._load (module.js:463:25)
    at Module.require (module.js:556:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/jan/nextcloud/apps/mail/node_modules/ajv/lib/compile/util.js:19:20)
    at Module._compile (module.js:612:30)
    at Object.Module._extensions..js (module.js:623:10)
    at Module.load (module.js:531:32)
    at tryModuleLoad (module.js:494:12)
    at Function.Module._load (module.js:486:3)
Makefile:39: recipe for target 'optimize-js' failed
make: *** [optimize-js] Error 1
```

ChristophWurst:
> but then on the next it fails like this:

See https://github.com/nextcloud/mail/pull/615

jancborchardt:
Doesn’t seem to work for me, even with the fixed optimize-js:
screenshot from 2017-11-16 16-40-03

jancborchardt:
After switching folders now it seemed to load. :D

ChristophWurst:
Whoops now I was too happy about the approved review that I forgot to squash the commits 🙈

ChristophWurst:
@jakobsack don't forget to claim your bounties for https://github.com/nextcloud/mail/issues/463 and https://github.com/nextcloud/mail/issues/236 😉

Titan-C:
How is this configuration activated? I have nextcloud 13 & contacs 2.1.2. I load my contacts but avatars are not fetched.

PR #472, opened 2017-08-27
Mail, Node.js, set up, ↵

comradekingu:
Node.js as per https://nodejs.org/static/documents/trademark-policy.pdf#page=3
↵ might be an issue for blind people, (?), but consistent with ← and →
https://portknox.net are inconsistent with their use of big versal k in PortKnox, i reckon if they do it at all, they meant to do it all the time.  ┐('～')┌


Gomez:
Thanks @comradekingu, looks good. 
As i am the Portknox guy, i would perfer Portknox (small k).

comradekingu:
@Gomez As you wish.

PR #473, opened 2017-08-28
Add an integration test for saving a draft

ChristophWurst:
Little testing code, but covers a lot of critical application code. We really should have integration tests for stuff like this :)


ChristophWurst:
> Test Coverage has increased to 66% (+1%). 

🚀 🤖 🕺

lib/Service/MailTransmission.php
```diff
@@ -103,7 +103,7 @@ public function sendMessage($userId, NewMessageData $messageData, RepliedMessage
 	/**
 	 * @param NewMessageData $message
 	 * @param type $draftUID
```
nickvergessen: *hust*

PR #476, opened 2017-08-28
Fix encoding of drafts messages

ChristophWurst:
Apparently the horde libs do not produce a valid MIME unless you
actually use the 'send' method. Since they fortunatelly have a
null transport with an empty impl, we can use this to have the same
encoding magic going on as we do have in place for actually sending
messages via SMTP. Now drafts stored to IMAP are encoded correctly.

Fixes https://github.com/nextcloud/mail/issues/9


ChristophWurst:
Merged #473 into this to have CI test the new code too.

PR #483, opened 2017-08-29
Add integration test for replacing an existing draft

ChristophWurst:
Yet another little integration test to cover more of our features. I even found a little issue that caused wrongly cached mailboxes when we created a new drafts/sent folder. Now the cache is cleared and the current list of mailboxes is fetched from the server.


ChristophWurst:
>Test Coverage increased to 66% (+1%). 

🚀 🤖 🕺

PR #487, opened 2017-08-30
Allow navigation from account settings to composer, fixes #110

### LINKED ISSUE #110: Navigation from account settings to new message composer is impossible [$10 awarded]
Since the current code expects that a folder view is shown, nobody listens to the `composer:show` event when the account settings are currently shown.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37983751-navigation-from-account-settings-to-new-message-composer-is-impossible?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kingjan1999:
I added a listener for the composer:show event in js/views/accountsettings.js, which triggers the navigation to the folder view and then invokes the composer:show event again.
Note: I introduced an extra flag (`triggerComposer`) to achieve this behavior. I saw `showComposer` which sound like it would do basically do the same, but unfortunately I wasn't able to find any code making use of this flag. Thus I was unsure if I could reuse it here and introduced a new parameter to stay safe.

Feel free to leave comments, especially regarding code style, as the contributing guidelines aren't very extensive.


ChristophWurst:
Haven't tested it yet but code looks clean and changes makes sense. Thanks a lot :)

kingjan1999:
Hey,

i updated the branch with the current master in both pull requests. However this fix works for me [at least in Firefox](https://streamable.com/fhdp4) and chrome. If I find time later this evening I will investigate this behavior further.

kingjan1999:
Hey,

I can't implement the navigation between keyboard shortcuts & account setup and the composer view, because the account data (folderId, accountId) aren't passed to the views. We'd have to pass the data to the views, but this requires us to choose from one of the accounts. This problem doesn't affect the account settings view, because it's account specific.   

You can preview, what sth. like this would look like in 604657427c79b554ecd5b9942a7e5801ec0655d9 . I decided to use the first account for the composer view. What's missing here is some kind of error routine, if no account exists (yet).

ChristophWurst:
> because the account data (folderId, accountId) aren't passed to the views.

No problem. Just use the first account and its first folder. That's *good enough* IMO. 

> What's missing here is some kind of error routine, if no account exists (yet).

In that case the *new message* button should be hidden anyway. I'm pretty sure we already do that.

kingjan1999:
Ok, I merged the above mentioned commit into this PR, it should work now. If you want, I can update this PR to the current master and squash all commits into one

js/routecontroller.js
```diff
@@ -167,15 +167,23 @@ define(function(require) {
 		 */
 		showSetup: function() {
 			this._navigate('setup');
+			var account;
+			if(this.accounts && this.accounts.models) {
+				account = this.accounts.models[0]; // select first account
```
ChristophWurst: `this.accounts.first()`

js/routecontroller.js
```diff
@@ -167,15 +167,23 @@ define(function(require) {
...
 			Radio.ui.trigger('navigation:hide');
-			Radio.ui.trigger('setup:show');
+			Radio.ui.trigger('setup:show', account);
 			return Promise.resolve();
 		},
 		showKeyboardShortcuts: function() {
 			this._navigate('shortcuts');
+			var account;
+			if(this.accounts && this.accounts.models) {
+				account = this.accounts.models[0]; // select first account
```
ChristophWurst: `.first()`

js/views/appview.js
```diff
@@ -153,19 +153,22 @@ define(function(require) {
 			OC.Notification.showTemporary(message);
 			$('#mail_message').removeClass('icon-loading');
 		},
-		showSetup: function() {
+		showSetup: function(account) {
```
ChristophWurst: 👎 this is way too hacky. why should we pass an account object when we show the setup screen? This is independent of any (existing) account.

js/views/appview.js
```diff
@@ -153,19 +153,22 @@ define(function(require) {
...
 				config: {
 					accountName: $('#user-displayname').text(),
 					emailAddress: $('#user-email').text()
-				}
+				},
+				account: account
 			}));
 		},
-		showKeyboardShortcuts: function() {
+		showKeyboardShortcuts: function(account, folder) {
```
ChristophWurst: 👎 same as above. keyboard shortcuts are account neutral/independent.

ChristophWurst:
@kingjan1999 I will try to specify the requirements of the original tickets in more detail. It seems it wasn't as clear as I thought it would be.

kingjan1999:
Feel free to do that, unfortunately, I don't have time to take care of this at the moment, so I will close this PR for now.

PR #484, opened 2017-08-29
Add integration tests for replies and reply subjects

ChristophWurst:
Aaaaand another integration test.

PR #488, opened 2017-08-31
Use right folder & messageId when replying, should (partially) fix #482

### LINKED ISSUE #482: Replying to an email doesn’t send the subject anymore [$50]
### Steps to reproduce
1. Reply to an email
2. Check the sent folder

### Expected behaviour
The subject of the emails should be visible

### Actual behaviour
It’s not there. ;)

Additionally, when you then refresh in the sent folder, all the recipients names are replaced with your name.

cc @nextcloud/mail

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/48758937-replying-to-an-email-doesn-t-send-the-subject-anymore?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kingjan1999:
It seems like the folderId and messageId were determined in a wrong way, which eventually lead to RepliedMessageData.php returning false for isReply().

This should fix the main issue of #482:  
replyfix
except for:
> Additionally, when you then refresh in the sent folder, all the recipients names are replaced with your name.

Which is a separate issue IMHO.


ChristophWurst:
> Additionally, when you then refresh in the sent folder, all the recipients names are replaced with your name.
> 
> Which is a separate issue IMHO.

Yes, it is.

ChristophWurst:
@kingjan1999 thanks a lot for the fix. The changes make sense. Before I approve it, I'd like @jancborchardt to verify the fix because he has experienced this annoying bug a lot :)

ChristophWurst:
@jancborchardt this basically confirms my suspicion that the bug was hiding somewhere on the client side because I couldn't make the reply logic fail while working on #484, which only covered the server-side code.

Gomez:
I can confirm this fixes #482. 

Great @kingjan1999! Thanks for your help.

As this is a bad regression, we should do a release after merging. What you mean @ChristophWurst?

ChristophWurst:
Sure, that was my plan too ;)

jancborchardt:
Cool, looking forward to more contributions @kingjan1999! :)

PR #494, opened 2017-09-04
Fix error propagation in case messages cannot be loaded

ChristophWurst:
For some reason messages cannot be loaded in my dev instance at the moment, so I discovered that we do not propagate this error correctly and the app gets stuck at the loading view, which is kinda misleading. Now a beautiful error view is displayed if one of the accounts fails to load 😁

### DIFF
--- a/js/service/messageservice.js
+++ b/js/service/messageservice.js
@@ -57,17 +57,20 @@ define(function(require) {
 			folderId: folder.get('id')
 		});
 
-		return Promise.resolve($.ajax(url, {
-			data: {
-				filter: options.filter
-			},
-			error: function(error, status) {
-				if (status !== 'abort') {
-					console.error('error loading messages', error);
-					throw new Error(error);
+		return new Promise(function(resolve, reject) {
+			$.ajax(url, {
+				data: {
+					filter: options.filter
+				},
+				success: resolve,
+				error: function(error, status) {
+					if (status !== 'abort') {
+						console.error('error loading messages', error, status);
+						reject(error);
+					}
 				}
-			}
-		})).then(function(messages) {
+			});
+		}).then(function(messages) {
 			var isSearching = options.filter !== '';
 			var collection = folder.messages;
 


Gomez:
Not sure how to test. 
I have added a new mail account then changed password. After a timeout i got this error message "Error while loading the selected account."

ChristophWurst:
> Not sure how to test.
> I have added a new mail account then changed password. After a timeout i got this error message "Error while loading the selected account."

It's hard to reproduce. Try adding a `throw new \Exception()` to `MessagesController.php`. I only had this error because I didn't update my composer deps and so my installation was missing the new html purify css thingy.

ChristophWurst:
> Not sure how to test.
> I have added a new mail account then changed password. After a timeout i got this error message "Error while loading the selected account."

Try this patch

```patch
 lib/Controller/MessagesController.php | 1 +
 1 file changed, 1 insertion(+)

diff --git a/lib/Controller/MessagesController.php b/lib/Controller/MessagesController.php
index 6cbad691..eeaeb929 100755
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -125,6 +125,7 @@ class MessagesController extends Controller {
 	 * @return JSONResponse
 	 */
 	public function index($accountId, $folderId, $cursor = null, $filter=null, $ids=null) {
+		throw new \Exception();
 		if (!is_null($ids)) {
 			$ids = explode(',', $ids);
 ```

jancborchardt:
There’s a scrollbar in the emptycontent view:

screenshot from 2017-09-06 01-01-23

Removing the emptycontent ID from the inner div and instead adding it as a class to the outer div fixes it maybe? 
screenshot from 2017-09-06 01-01-54

ChristophWurst:
> There’s a scrollbar in the emptycontent view

Please file a new issue for this. This is unrelated to this PR and I'm not sure how to solve it (as mentioned at the conf, Marionette often needs more HTML elements that you'd usually use due to its way of wrapping views and regions).

PR #496, opened 2017-09-05
Release v0.7.2

ChristophWurst:
[*Release early, release often* ](https://en.wikipedia.org/wiki/Release_early,_release_often) 🚀 🤖 🐛 

Three important bugfixes have been merged, so let's get this out to our users.

RC-build: [mail.tar.gz](https://github.com/nextcloud/mail/files/1276318/mail.tar.gz)

cc @nextcloud/mail @plastikschnitzer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-## 0.7.1 – 2017-08-31
+## 0.7.2 – 2017-09-05
+### Fixed
+- Subject of replied messages
+- Navigation to next message in unified inbox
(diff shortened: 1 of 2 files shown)


Gomez:
Few minutes to late:
https://github.com/nextcloud/mail/pull/497

Time to test & include?

ChristophWurst:
> Few minutes to late:
> #497
> 
> Time to test & include?

Let's do that with the next release. I also thought about not including any js sources in the tarball as we don't really need them for production use ;)

ChristophWurst:
#498 #500 #502 will be part of 0.7.1 too as they just got merged into master

PR #497, opened 2017-09-05
Quick minify config for webpack

Gomez:
From build.js  2.6 MB to build.js  328 kB 

Ugly copy of webpack config for production.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ install-npm-deps-dev:
 	npm install --deps
 
 optimize-js: install-npm-deps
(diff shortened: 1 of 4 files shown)


ChristophWurst:
> Not a big fan of this duplication TBH.

Still an awesome enhancement, especially for page loads on mobile devices. Thanks a lot for digging into this 🚀

Gomez:
Yeah there is a way. This is just a quick hack. Will improve in a later step.

ChristophWurst:
This seems to be inactive and thus I'm closing this PR now.

Gomez:
:P Will work on it now. Hackweek ftw

PR #498, opened 2017-09-05
Add earlier loading feedback

ChristophWurst:
In terms of UX, I like to know what the app is doing, thus
IMO it makes sense to tell the user we're loading their accounts.

From a technical perspective, I know that the app's js has started
executing. This is useful for debugging where you didn't know
whether the page was still loading or the app was already started.

If you have any suggestions for a better wording, let me know :)

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -102,7 +102,7 @@ define(function(require) {
 		this.view = new AppView();
 		Cache.init();
 
-		Radio.ui.trigger('content:loading');
+		Radio.ui.trigger('content:loading', t('mail', 'Loading accounts …'));
 
 		this.registerProtocolHandler();
 		this.requestNotificationPermissions();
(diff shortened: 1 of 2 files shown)


jancborchardt:
Maybe simplify wording to "Loading accounts …" - we rarely use "your".

ChristophWurst:
> Maybe simplify wording to "Loading accounts …" - we rarely use "your".

Done. And also fixed one failing test. Please re-review, thanks :)

jancborchardt:
There’s only two things:

- A small jump – initially the vertically centered loader icon appears, then the one with the text which is slightly higher. Can we directly show the one with text?
- The left navigation still has the extra spinner. Time to remove this one? Mobile starts without the sidebar anyway so I don’t think it’s necessary.

ChristophWurst:
1) no, becaus it takes some time for the js to kick in. The first spinner is hard-coded into the HTML on page load, which re replace ASAP. If we replaced that with the text we wouldnt see when the app starts and you're back to where we are without this path.
2) surely doable somehow but requires more changes, more testing of different routing scenarious an thus out of scope for this little enhancement PR.

Issue #406, opened 2017-06-30
Background-fetching and showing some kind of notification icon [$15]


fsedarkalex:
This is a feature request.

I would love to see a feature (should be configurable by the user (configurable default in config.php) or at least through config.php) that adds a small red dot or similar visible indicator that makes a user notice that there are new mails in his inbox even when mail app is not running.
This indicator should show up when new mail is received until the user opens the mail app once, not every time when there are unread mails. It is about new items in Mailapp not unread mails.

If an indicator is not possible in that way what about a notification? I guess notifications can also be canceled when user enters mail app?

I hope you know what I am talking about... :laughing: 

Oh! The notification could look like this:
**You got X new E-Mails**
_Subject of E-Mail 1_
_Subject of E-Mail 2_
_Subject of E-Mail 3_

Email 1 is the latest email, 2 the second latest and so on... Only showing E-Mails that have not been noticed yet (noticed means seen - not read)

ChristophWurst:
Actually we had this functionality once, but now that I'm thinking about it, I've not seen the red dot icon for a long time. IIRC it's a bit of a fragile feature as we're overwriting the icon that is set by the Nextcloud server component. @jancborchardt @juliushaertl or @nickvergessen could know more about it, I think we theme the icon now.

In any case, the background sync which provides the basis for a notification icon is kinda broken right now. I'll work on it as soon as #326 has been merged.

juliushaertl:
@ChristophWurst Yes, the favicon will be generated depending on the app and theming color.

But there are some js libraries around to modify favicons on the fly. Those might be helpful for that:
- http://lab.ejci.net/favico.js/
- http://blog.tommoor.com/tinycon/

PR #499, opened 2017-09-05
Fix the notification favicon

ChristophWurst:
The code to our special notification icon with the red dot was
somehow broken. This triggers the favicon change even again and
also fixes the outdated jQuery selector for the favicon DOM element.

Fixes https://github.com/nextcloud/mail/issues/406

cc @fsedarkalex


skjnldsv:
Me neither.
@ChristophWurst, we're supposed to get the same number on the icon as the number in the page title?

ChristophWurst:
Retested and works for me: 
bildschirmfoto von 2017-09-06 09-53-39

Note that you have to change tabs / application to get an updated notification favicon. We only show it if the window is not focused.

ChristophWurst:
Okay, I found a little flaw. The favicon was not updated if one didn't allow desktop notifications. Now it updates it regardless of the notification permissions :)

jancborchardt:
Hm, still doesn’t work for me in neither Firefox nor Chromium.

How about you @fsedarkalex @lcalaresu @Gomez?

ChristophWurst:
> Hm, still doesn’t work for me in neither Firefox nor Chromium.
> 
> How about you @fsedarkalex @lcalaresu @Gomez?

Please provide the *exact* steps you took to test this. Did you run `make optimize-js`? Did you refresh your browser window? ;)

ChristophWurst:
> Please provide the exact steps you took to test this. Did you run make optimize-js? Did you refresh your browser window? ;)

And most importantly: did you receive any new emails while the browser tab was not in focus? Did you see a desktop notification?

skjnldsv:
@ChristophWurst the mails are not refreshed if not focused, that's why I can't test it!
No requests in the network console either. :/

jancborchardt:
@ChristophWurst I followed all these steps. I’m running on this branch, ran the make optimize-js, sent a mail to myself, changed to another tab, and I’m also getting the notifications. But no red dot. I tested in Firefox and Chromium and also in private tabs.

skjnldsv:
ToFix: go back to the old favicon (no red dot) on window focus?

ChristophWurst:
> ToFix: go back to the old favicon (no red dot) on window focus?

Should be good now, please re-test one more time ;)

jancborchardt:
A possible additonal enhancement seems to be that when you read the mail on another device, that the next background refresh will remove the red dot from the favicon. Would that work @ChristophWurst?

ChristophWurst:
A bit tricky but definitely possible.

ChristophWurst:
As always, pleas open a ticket so that we don't forget about it.

js/views/appview.js
```diff
@@ -120,8 +120,9 @@ define(function(require) {
 			Radio.ui.trigger('document:click', event);
 		},
 		onDocumentShow: function(e) {
+			console.error('SHOW');
```
nickvergessen: 🙈

ChristophWurst: haha, whoops

skjnldsv: I though it was on purpose :p

jancborchardt:
Follow-up issue at https://github.com/nextcloud/mail/issues/507 :)

fsedarkalex:
When talking about favicon are we actually talking about the icon on the browser tab?
Am I right it should also be displayed with red dot when I am currently e.g. in contacts app?
That is what I was especially talking about.
Further I think the dot should also be visible next to mail app's icon in the header or menu list or something like that...

My results:
The red dot appears but only when mail app is active.

My additional Question:
When I am on mail apps tab I see a number of unread mails in the title. When I am somewhere else I don't. Why? Bug? :)

ChristophWurst:
> When talking about favicon are we actually talking about the icon on the browser tab?
> Am I right it should also be displayed with red dot when I am currently e.g. in contacts app?
> That is what I was especially talking about.
> Further I think the dot should also be visible next to mail app's icon in the header or menu list or something like that...

Yes, we only update the tab's favicon and not the app icon in the app menu. Updating the app icon also from other apps would mean we'd have to poll for new messages in any opened tab. That would require a lot of processing power and is not worth the additional complexity it would introduce into this app. But still, feel free to open an enhancement ticket ;)

> The red dot appears but only when mail app is active.

It appears if the app is *not* active.

> When I am on mail apps tab I see a number of unread mails in the title. When I am somewhere else (meaning when the red dot appears) I don't. Why? Bug? :)

Probably. I think we don't synchronize the number of unread messages when we sync new/changed/vanished messages in background. Please file an issue and I'll take a look at this.

fsedarkalex:
> > The red dot appears but only when mail app is active.
> It appears if the app is not active.
we should find some better way to say what we are talking about :)

"app active" = clicked mail app
"app inactive" = clicked e.g. contacts app

what you mean with app inactive is using another browser tab, right?

I will open two more "issues" :)

ChristophWurst:
> what you mean with app inactive is using another browser tab, right?

Exactly :)

PR #510, opened 2017-09-11
Always log IMAP and SMTP if debug mode is enabled

ChristophWurst:
IMAP and SMTP logs are *very* helpful to analyze bug reports. Unfortunately, most users don't enable it and thus bug reports often just list the (generic) error message from nextcloud.log, if any. Thus I'd like to enable those debug logs as soon as the instance is set to debug mode.

This lowers the number of possible settings and hopefully results in better bug reports.

@nextcloud/mail opinions? objections? Let me know :)


ChristophWurst:
No feedback/objections -> merge as soon as CI is green

Issue #481, opened 2017-08-29
Drafts: to and cc is wrongly set [$15]


jancborchardt:
@ChristophWurst as debugged. When I send an email to someone, in the drafts view my own email is put in the »to« field and the original recipient moved to »cc«:
![screenshot from 2017-08-29 13-58-14](https://user-images.githubusercontent.com/925062/29819833-9ab4cc4c-8cc2-11e7-8ae6-7fd8aca8b0f7.png)
The draft is saved correctly, but the composer somehow loads it wrongly. The from and to paramenters are correct, but the replyToList and replyCcList are wrong – they shouldn’t even exist.

PR #515, opened 2017-09-11
Fix resumed draft to/cc/bcc values

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/481

I've found another (apparently yet undiscovered *bug*): bcc values are not saved for drafts. Digging into the code showed that horde does not store bcc values. They are just temporary values, used to send the email also to those recipients. Which makes sense. Kind of. This needs more investigation. Here, I focused on the to/cc stuff mainly, which now works.


ChristophWurst:
> When the »to« is a person with name, like "Name Lastname" name@lastname.com, only the "Name Lastname" part gets saved.

Fixed. Please re-test.

ChristophWurst:
@jancborchardt just told me it works and he takes full responsibility if something breaks.

ChristophWurst:
Good to merge then, right?

jancborchardt:
Nope :D

jancborchardt:
screenshot from 2017-09-22 16-46-10

PR #517, opened 2017-09-12
AutoConfig: remove unused dependencies and extract method

ChristophWurst:
Just stumbled over that class and was wondering why it had so many dependencies. Turned out many of them aren't needed anymore. This is probably a leftover of my autoconfig refactoring a few months ago.

### DIFF
--- a/lib/Service/AutoConfig/AutoConfig.php
+++ b/lib/Service/AutoConfig/AutoConfig.php
@@ -22,50 +22,22 @@
 
 namespace OCA\Mail\Service\AutoConfig;
 
-use Horde_Imap_Client_Exception;
-use Horde_Mail_Transport_Smtphorde;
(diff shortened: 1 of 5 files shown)


ChristophWurst:
I've pushed another commit that refactors one method into its own class. This allows us to test code in isolation and keeps our code base readable and maintainable.

Autoconfig was tested manually and it still works.

ChristophWurst:
Well, while at it, there's more low hanging refactoring fruit :)

ChristophWurst:
> Can’t see any problems in viewing, sending, replying, th usual stuff. :) 👍

Well, that's obvious. Please test the *auto config*, as in try to re-add your accounts and see whether it works. Thanks :)

ChristophWurst:
@jancborchardt so did you actually re-test this?

PR #518, opened 2017-09-12
Add more CSRF checks

ChristophWurst:
I've removed the `@NoCSRFRequired` from routes that are access by AJAX calls. Those should automatically send the CSRF token.

@LukasReschke please take a quick look.


ChristophWurst:
Actually I don't know why we had those annotation there in the first place. Probably for debugging …

Gomez:
Does not break anything on a quick test.

ChristophWurst:
> Does not break anything on a quick test.

Then let's merge it. We'll notice when something breaks 😉

PR #521, opened 2017-09-13
Fit app-navigation to css guidelines

### LINKED ISSUE #516: Fit the app-navigation to css guidelines [$20]
Refs: https://docs.nextcloud.com/server/12/developer_manual/app/css.html

<bountysource-plugin>

---
There is a **[$20 open bounty](https://www.bountysource.com/issues/49232072-fit-the-app-navigation-to-css-guidelines?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #528: Favorites folder sometimes appear as sub-folder of inbox even if it's not
With master
https://github.com/nextcloud/mail/blob/632e9ed934b269ec0bf8742a92c9afe0ff4ec3ae/lib/Service/FolderMapper.php#L70-L92

Gives me:
``` php
array(
	0 => stdClass::__set_state(array(
		'id' => 'SU5CT1g=',
		'accountId' => 4,
		'name' => 'Inbox',
		'specialRole' => 'inbox',
		'unseen' => 0,
		'total' => 2,
		'isEmpty' => false,
		'noSelect' => false,
		'attributes' => array(
			0 => '\marked',
			1 => '\hasnochildren',
		) ,
		'delimiter' => '/',
		'folders' => array() ,
		'syncToken' => 'xxxxxx',
	)) ,
	1 => stdClass::__set_state(array(
		'id' => 'SU5CT1gvRkxBR0dFRA==',
		'accountId' => 4,
		'name' => 'Favorites',
		'specialRole' => 'flagged',
		'unseen' => 0,
		'total' => 0,
		'isEmpty' => true,
		'noSelect' => false,
		'attributes' => array(
			0 => '\marked',
			1 => '\hasnochildren',
			2 => '\Flagged',
		) ,
		'delimiter' => '/',
		'folders' => array() ,
		'syncToken' => NULL,
	)) ,
	2 => stdClass::__set_state(array(
		'id' => 'RHJhZnRz',
		'accountId' => 4,
		'name' => 'Drafts',
		'specialRole' => 'drafts',
		'unseen' => 0,
		'total' => 1,
		'isEmpty' => false,
		'noSelect' => false,
		'attributes' => array(
			0 => '\hasnochildren',
			1 => '\drafts',
		) ,
		'delimiter' => '/',
		'folders' => array() ,
		'syncToken' => 'xxxxxx=',
	)) ,
	3 => [REMOVED FOR READABILITY]
);
```

It's probably related to the id of the two folders:
 - SU5CT1g=
 - SU5CT1gvRkxBR0dFRA==

Which gives us respectively:
 - INBOX
 - INBOX/FLAGGED

So the favorite folder appear as subs even if it has the `hasnochildren` attribute.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49330678-favorites-folder-sometimes-appear-as-sub-folder-of-inbox-even-if-it-s-not?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skjnldsv:
Fix #516 
Fix #528

### DIFF
--- a/js/service/accountservice.js
+++ b/js/service/accountservice.js
@@ -86,7 +86,14 @@ define(function(require) {
 			if (accounts.length > 1) {
 				accounts.add({
 					accountId: -1,
-					isUnified: true
+					isUnified: true,
+					id: btoa('all-inboxes'),
+					name: t('mail', 'All inboxes'),
+					emailAddress: t('mail', 'All inboxes'),
+					specialRole: 'inbox',
+					isEmpty: false,
+					noSelect: false,
+					delimiter: '.'
 				}, {
 					at: 0
 				});
--- a/js/service/accountservice.js
+++ b/js/service/accountservice.js
@@ -90,7 +90,6 @@ define(function(require) {
 					id: btoa('all-inboxes'),
 					name: t('mail', 'All inboxes'),
 					emailAddress: t('mail', 'All inboxes'),
-					specialRole: 'inbox',
 					isEmpty: false,
 					noSelect: false,
 					delimiter: '.'
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -1,7 +1,7 @@
 {{#if emailAddress}}
 <div class="mail-account-color" style="background-color: {{accountColor emailAddress}}"></div>
 {{/if}}
-<a href="javascript:void(0)" class="icon mail-account-email" title="{{emailAddress}}">{{emailAddress}}</a>
+<a href="#" class="icon mail-account-email" title="{{emailAddress}}">{{emailAddress}}</a>
 
 {{#if hasMenu}}
 <div class="app-navigation-entry-utils">
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -1,32 +1,35 @@
 {{#if emailAddress}}
 <div class="mail-account-color" style="background-color: {{accountColor emailAddress}}"></div>
 {{/if}}
-<h2 class="mail-account-email" title="{{emailAddress}}">{{emailAddress}}</h2>
+<a href="javascript:void(0)" class="icon mail-account-email" title="{{emailAddress}}">{{emailAddress}}</a>
 
 {{#if hasMenu}}
-<ul class="app-navigation-entry-utils">
-	<li class="app-navigation-entry-utils-menu-button svg"><button></button></li>
-</ul>
-
-<ul class="app-navigation-entry-menu popovermenu bubble menu">
-	<li>
-		<a href="#" class="menuitem action action-settings permanent" data-action="Settings">
-			<span class="icon icon-rename"></span>
-			<span>{{ t 'Settings' }}</span>
-		</a>
-	</li>
-	{{#if isDeletable}}
-    <li>
-        <a href="#" class="menuitem action action-delete permanent" data-action="Delete">
-			<span class="icon icon-delete"></span>
-			<span>{{ t 'Delete account' }}</span>
-		</a>
-    </li>
-	{{/if}}
-</ul>
+<div class="app-navigation-entry-utils">
+	<ul>
+		<li class="app-navigation-entry-utils-menu-button svg"><button></button></li>
+	</ul>
+</div>
+<div class="app-navigation-entry-menu">
+	<ul>
+		<li>
+			<a href="#" class="menuitem action action-settings permanent" data-action="Settings">
+				<span class="icon icon-rename"></span>
+				<span>{{ t 'Settings' }}</span>
+			</a>
+		</li>
+		{{#if isDeletable}}
+	    <li>
+	        <a href="#" class="menuitem action action-delete permanent" data-action="Delete">
+				<span class="icon icon-delete"></span>
+				<span>{{ t 'Delete account' }}</span>
+			</a>
+	    </li>
+		{{/if}}
+	</ul>
+</div>
 {{/if}}
 
-<div class="folders with-icon"></div>
+<ul class="folders"></ul>
 {{#unless isUnifiedInbox}}
 {{#if hasFolders}}
 <span class="account-toggle-collapse">{{toggleCollapseMessage}}</span>
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -1,5 +1,5 @@
 {{#if emailAddress}}
-<div class="mail-account-color" style="background-color: {{accountColor emailAddress}}"></div>
+<div class="mail-account-color app-navigation-entry-bullet" style="background-color: {{accountColor emailAddress}}"></div>
 {{/if}}
 <a href="#" class="icon mail-account-email" title="{{emailAddress}}">{{emailAddress}}</a>
 
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -1,7 +1,7 @@
 {{#if emailAddress}}
 <div class="mail-account-color app-navigation-entry-bullet" style="background-color: {{accountColor emailAddress}}"></div>
 {{/if}}
-<a href="#" class="icon mail-account-email" title="{{emailAddress}}">{{emailAddress}}</a>
+<a href="#" class="icon mail-account-email" id="account{{accountId}}" title="{{emailAddress}}">{{emailAddress}}</a>
 
 {{#if hasMenu}}
 <div class="app-navigation-entry-utils">
--- a/css/mail.css
+++ b/css/mail.css
@@ -140,48 +140,6 @@
 }
 
 
-/* START: should be done in core/apps.css */
-
-#app-navigation .collapse {
-	display: inline-block !important;
-	opacity: .3;
-	z-index: 1;
-}
-
-#app-navigation-accounts {
-	height: calc(100% - 65px);
-	overflow-x: hidden;
-}
-
-#app-navigation ul ul {
-	display: block;
-}
-
-#app-navigation ul ul ul {
-	display: none;
-}
-
-#app-navigation .navigation-account {
-	position: relative;
-}
-#app-navigation .navigation-account .app-navigation-entry-utils-menu-button {
-	/* show the account menu toggle (three dots) by default */
-	display: inline-block;
-}
-
-#app-navigation ul ul li > a {
-	padding-left: 32px;
-}
-
-#app-navigation ul .collapsible.open:hover {
-	box-shadow: inset 0 0 3px #ddd;
-}
-
-#app-navigation ul .collapsible.open ul {
-	display: block;
-}
-/* STOP: should be done in core/apps.css */
-
 #emptycontent,
 #mail-messages #emptycontent.emptycontent-search {
 	position: static;
@@ -194,15 +152,8 @@
 }
 
 .mail-account-email {
-	display: inline-block;
-	opacity: .5;
-	padding: 20px 0 10px 25px;
-	width: calc(100% - 50px);
-	margin: initial;
-	font-size: 100%;
-	overflow: hidden;
-	text-overflow: ellipsis;
-	white-space: nowrap;
+	/* for icon color */
+	padding-left: 44px !important;
 }
 .mail-account-email:first-child {
 	display: none;
@@ -219,21 +170,13 @@
 }
 
 .mail-account-color {
-	display: inline-block;
+	display: block;
 	width: 14px;
 	height: 14px;
+	margin: 16px;
 	border-radius: 50%;
-	position: relative;
-	left: 15px;
-	top: -12px;
+	position: absolute;
 }
-/* color dot position fix for Safari */
-@media not all and (min-resolution:.001dpcm) { @media
-{
-.mail-account-color {
-        top: 2px;
-}
-}}
 
 .mail-message-account-color {
 	position: absolute;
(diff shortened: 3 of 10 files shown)


jancborchardt:
Oh by the way, cause I see this (I know that’s not how it’s supposed to look like ;):
screenshot from 2017-09-15 12-29-36

But the one thing is: The folders are indented sub-entries of the accounts. That indentation is stylistically not something which we would want in this case though. :\ The account is just a heading, but not a »folder« per se.

skjnldsv:
@jancborchardt, it's an easy fix with a single line on the mail app css. That being said, we should wait the end of my work here to really see what it would look like! :)

jancborchardt:
Okidoke! I just snook a peek :)

skjnldsv:
@jancborchardt Feeling better? 😉 
capture d ecran_2017-09-20_18-40-16

jancborchardt:
@skjnldsv not yet – this would be the correct indentation ;) As it is in the current design too. That’s why I mentioned it’s unusual, but we don’t want to have it look like a strange tree view by default.
30656174-6e204ff8-9e33-11e7-8dc4-02b7d7401847

skjnldsv:
I disagree, folders are subitems of the account, they should appear as indented.

jancborchardt:
The folders are _not_ subitems of an account. The account name is simply the header of the set of folders. :)
I’m pretty particular about this since the tree view look by default makes the app look quite old-style. Even when you only have a single account, the folders will be indented and that just looks strange.


What do you think @ChristophWurst @Gomez?

skjnldsv:
I'm okay with both! :)
I was just stating how the app outputs the divs! 😉 

If we go for NOT a subitem, the work will be far more considerable! :)

pixelipo:
I agree with @jancborchardt regarding indentation. Account name is not clickable and it's not collapsible, hence it shouldn't pretend to be a parent folder.

However, as far as headers go, account name could be styled better. Right now, it's styled to like like a folder (mailbox). It's very hard to differentiate it from actual mailboxes (inbox, sent etc.) App-navbar doesn't seem to support headers at the moment, so I don't know how it should look :)

js/templates/account.html
```diff
@@ -1,32 +1,35 @@
 {{#if emailAddress}}
-<div class="mail-account-color" style="background-color: {{accountColor emailAddress}}"></div>
+<div class="mail-account-color app-navigation-entry-bullet" style="background-color: {{accountColor emailAddress}}"></div>
```
pixelipo: Shouldn't this be rendered via CSS, similar to the way folders' icons are done?

skjnldsv: You mean the background color?

pixelipo: Not only that, the whole `.mail-account-color` could be moved over to `.mail-account-email:before`. That would make for a cleaner DOM.

skjnldsv: The bullet is now integrated into server. It's part of the css guidelines.
see https://github.com/nextcloud/server/pull/6399

jancborchardt:
> However, as far as headers go, account name could be styled better. Right now, it's styled to like like a folder (mailbox). It's very hard to differentiate it from actual mailboxes (inbox, sent etc.) App-navbar doesn't seem to support headers at the moment, so I don't know how it should look :)

It’s greyed out for now, and has the colored circle. That’s what differentiates it from the folders – any other ideas? :)


@skjnldsv so should we in the mail app simply override that one first indent? Or how to do it? :) I mean, semantically the accounts are not a »parent folder« of the mail folders, as @pixelipo also said, so they shouldn’t be like that in the markup.

skjnldsv:
If we start overriding the default server, let's drop this full pr. I don't see the point. :)
Speaking of hierarchy, I find the indentation pretty logical.
But I also see your point. We could put all items as first childs. @ChristophWurst is this an easy task?

pixelipo:
@skjnldsv Well I think we need to decide whether we want to support list headings in the app navbar. In my opinion, it's logical to have this - semantically, list usually have titles.

No app that I know of (apart from Mail) has the need for them - so far, but if you look at Calendar, you will notice that regular and webcals are separated and the "New ***" button is acting as a heading:
screenshot from 2017-09-21 23-04-19

I would also like to attach Geary sidebar - they are indenting folders within accounts, even though accounts themselves are not clickable (they are collapsible, though).
screenshot from 2017-09-21 23-05-17

skjnldsv:
Thanks @pixelipo! :)
My opinion is on the subfolder side. I like the hierarchy it induce and it's easier to have it so we can easily collapse what we don't want to see.

jancborchardt:
@pixelipo and that sidebar look is precisely what I find annoying with Geary. :) It’s complicating the view and makes the structure look more complex than it actually is.

jancborchardt:
@skjnldsv that’s the thing though – the Mail app doesn’t need to override it if the accounts are headers and not parent folders. We could add a bit of vertical whitespace like in the Calendar too, to make the separation more clear. (Note that the buttons being used as headings in the Calendar is by design.)

skjnldsv:
I'm OK with you @jancborchardt! :)
I was saying that in case if putting folders as first level entries is not possible. Because removing the ul padding is not the proper way to do it! :(

skjnldsv:
@ChristophWurst Fixed the first level unified inbox issue! :)

js/service/accountservice.js
```diff
@@ -86,7 +86,14 @@ define(function(require) {
 			if (accounts.length > 1) {
 				accounts.add({
 					accountId: -1,
-					isUnified: true
+					isUnified: true,
+					id: btoa('all-inboxes'),
+					name: t('mail', 'All inboxes'),
+					emailAddress: t('mail', 'All inboxes'),
+					specialRole: 'inbox',
```
ChristophWurst: 🙈 so you're adding folder properties to the account?

skjnldsv: what do you mean? :p

ChristophWurst: Obviously, the account does not have a property called `specialRole`. Only `Folder` objects do 😉

skjnldsv: Oh, okay! :)
I basically copied the property added into folders. I'm a bit lost with backbone & marionnette, feels harder than it could have been :p

jancborchardt:
Some remaining issues:

- [ ] How do we now fix the »everything is indented by default« issue? :)
screenshot from 2017-09-28 12-57-05
- [ ] »All inboxes« has a colored circle whereas it should have icon-inbox
- [ ] »Show all folders« toggle doesn’t have same padding (clickability and whitespace to bottom) as menu entries. Should it be a li element too or does it need to be outside of it and we need to add additional styles?
screenshot from 2017-09-28 12-57-05 copy
- [ ] The heading bar of the Settings drawer is half-transparent when opened:
screenshot from 2017-09-28 12-55-44

jancborchardt:
The core of the indentation problem is that the grouping of the account sets are also marked up as ul. If we just do that via div, it would solve the problem, no? Because only the accounts themselves should be ul, with the entries Inbox, Sent, Drafts etc being the li.

ChristophWurst:
> Removed unwanted attribute to all inbox

It's about *all* the attributes you've added. They don't belong to the account but to the actual folder: the unified inbox. I will take a closer look next week - promised!

skjnldsv:
>  The heading bar of the Settings drawer is half-transparent when opened:

Nice callout, this is a core bug!

>  The active folder (»Sent« in this case) doesn’t have the blue bar on the left:

Because it's a sub item :)

>  How do we now fix the »everything is indented by default« issue? :)

Good question. I'm not sure we can put everything as first items. This would require a big change in the backbone/marionette controllers! :/

ChristophWurst:
Regarding

> @skjnldsv Well I think we need to decide whether we want to support list headings in the app navbar. In my opinion, it's logical to have this - semantically, list usually have titles.

and

> Good question. I'm not sure we can put everything as first items. This would require a big change in the backbone/marionette controllers! :/

IMO of course, folders belong to an account and thus reflect a hierarchy. However, I'm not a big fan of indenting that in the sidebar as that seems like a lot of wasted space to me. I like the idea of having list heading/separators that we could use to separate different accounts. @skjnldsv is something like that feasible? Like something that could be used by *any* app somehow? I image that as one entry in the list with a dedicated class which makes the font a bit stronger/darker/whatsoever. We wouldn't then add it for the unified account of course, but rather for the specific ones only.

I'd gladly implement it for this app, I just have to know how to structure the HTML and I'll fight with/against backbone/marionette to achieve that 😉 

A rough structure would look like this

* Sidebar
  + Unified inbox
  + *Account 1 title heading*
  + Inbox
  + Sent
  + Other
     - Subfolder
     - Subfolder.Subfolder
  + *Account 2 title heading*
  + Inbox
  + Sent

To comply with the current design, list headings should have the ability to show an icon on the left (like the other entries in the list, I think we have the styling in place already). As before, we wouldn't go deeper than one subfolder.

Does that make sense? :)

skjnldsv:
That is perfect!!

ChristophWurst:
> That is perfect!!

Awesome. Then let me try to adjust our views to generate a HTML closely to what I've sketched above. Could be that Marionette requires us to have some additional elements due its requirement of an surrounding element and our nested view hierarchy. Should be fine though :)

skjnldsv:
@ChristophWurst good luck. The div wrapping is a pain!! :(

ChristophWurst:
> @ChristophWurst good luck. The div wrapping is a pain!! :(

Indeed they are. And from experience I can tell it's super hard to get rid of them without introducing random bugs. Thus I'd like to ask whether it would be possible to *relax* the CSS rules a bit to be a bit less strict about the HTML structure. As in: ignore some wrapping divs.

Something like
```patch
 core/css/apps.scss | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/core/css/apps.scss b/core/css/apps.scss
index 9978c38033..fa984ef90b 100644
--- a/core/css/apps.scss
+++ b/core/css/apps.scss
@@ -104,7 +104,7 @@ kbd {
 		box-sizing: border-box;
 		display: flex;
 		flex-direction: column;
-		> li {
+		li {
 			display: inline-flex;
 			flex-wrap: wrap;
 			order: 1;
@@ -201,8 +201,8 @@ kbd {
 			}
 		}
 		/* Menu and submenu */
-		> li,
-		> li > ul > li {
+		li,
+		li > ul > li {
 			position: relative;
 			width: 100%;
 			box-sizing: border-box;
```
seems to allow styling Mail according to the guidelines while still using some additional divs. @skjnldsv how bad is that? 😨

ChristophWurst:
> @skjnldsv how bad is that? 😨

To answer my own question: very bad. Now the stlye is applied to any `li` down the nested structure. Is there any fancy CSS trick to loosen the selector but still limit it to the first level of `li`s? 🤔

ChristophWurst:
Although it still looks a bit odd, I seem to be on the right path
bildschirmfoto von 2017-10-10 11-06-00

ChristophWurst:
Putting this on hold until https://github.com/nextcloud/mail/pull/521#issuecomment-335407693 has been answered. So far, so good
bildschirmfoto von 2017-10-10 11-22-59

ChristophWurst:
@skjnldsv please see https://github.com/nextcloud/mail/pull/565 😉

skjnldsv:
It's done on purpose to force the correct css guidelines. :)
So no, it would be counter productive to loosen the selector and basically allow anything to be done on the app navigation.

ChristophWurst:
> It's done on purpose to force the correct css guidelines. :)
> So no, it would be counter productive to loosen the selector and basically allow anything to be done on the app navigation.

I see. Unfortunately, that means I'm out of idea on how to solve this. I definitely won't merge all the views into one to remove the wrapping divs and I also won't work on hacks to remove the existing divs. With the tools we currently use (Backbone+Marionette), it seems very hard to obey those strict rules. And a switch to another front-end framework is not planned right now as it would require a lot of effort.

ChristophWurst:
https://github.com/nextcloud/mail/pull/565/commits/2e723e963990db716b2574b67d6a9697283ed4e6 removes two unnecessary wrapping divs. The other ones cannot be removed that easily it seems.

skjnldsv:
I'll think about what we can do on mail. On the other hand, it seems incredible that a very used engine like backbone forces users to wrap in div... I don't get their idea on this. 🤔

What is the main problem with your latest changes @ChristophWurst? The screenshot seems great?

ChristophWurst:
> What is the main problem with your latest changes @ChristophWurst? The screenshot seems great?

Only because I've adjusted the server css with the patch from above 🙈 

> On the other hand, it seems incredible that a very used engine like backbone forces users to wrap in div... I don't get their idea on this. 🤔

I guess https://github.com/jashkenas/backbone/issues/546#issuecomment-3604746

jancborchardt:
Looks very good so far! :) Much better

ChristophWurst:
I assume we can close this as #565 and #594 have been merged :)

PR #534, opened 2017-09-18
Enhance loading view wording

ChristophWurst:
* We don't use '…' anywhere for loading views
* Folder loaing message can cause problems with translations,
  e.g. for some languages the text is grammatically wrong.

Basically, it always bothers me to read "Lade Alle Posteingänge" or "Lade Sent" on my German instance 🙈

### DIFF
--- a/js/controller/foldercontroller.js
+++ b/js/controller/foldercontroller.js
@@ -113,8 +113,8 @@ define(function(require) {
 	function showFolder(account, folder, loadFirstMessage) {
 		loadFirstMessage = loadFirstMessage !== false;
 		Radio.ui.trigger('search:set', '');
-		Radio.ui.trigger('content:loading', t('mail', 'Loading {folder}', {
-			folder: folder.get('name')
+		Radio.ui.trigger('content:loading', t('mail', 'Loading folder {folder}', {
+			folder: '»' + folder.get('name') + '«'
 		}));
 
 		return new Promise(function(resolve, reject) {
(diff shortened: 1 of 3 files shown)


js/controller/foldercontroller.js
```diff
@@ -113,8 +113,8 @@ define(function(require) {
 	function showFolder(account, folder, loadFirstMessage) {
 		loadFirstMessage = loadFirstMessage !== false;
 		Radio.ui.trigger('search:set', '');
-		Radio.ui.trigger('content:loading', t('mail', 'Loading {folder}', {
-			folder: folder.get('name')
+		Radio.ui.trigger('content:loading', t('mail', 'Loading folder {folder}', {
+			folder: '»' + folder.get('name') + '«'
```
ChristophWurst: FYI I've added those here on purpose because otherwise translators have to insert them in the translated string for every language. This way we can also easily remove the quotes if we device we don't like them and can still use the old translation strings.

ChristophWurst:
Tests seem to fail …

jancborchardt:
> Basically, it always bothers me to read "Lade Alle Posteingänge" or "Lade Sent" on my German instance

Well, then it's going to be "Lade Ordner Sent" and "Lade Ordner Alle Posteingänge" which looks even stranger. ;)

If the grammar is what annoys you, we could simply _only_ display the folder name. The spinner is above it for loading feedback.

ChristophWurst:
> Well, then it's going to be "Lade Ordner Sent" and "Lade Ordner Alle Posteingänge" which looks even stranger. ;)
> 
> If the grammar is what annoys you, we could simply only display the folder name. The spinner is above it for loading feedback.

Wrong. Now it's actually *Lade Ordner »Sent«* which is much better IMO

PR #539, opened 2017-09-20
Add OCP files as composer dependency

ChristophWurst:
This allows better CI analyses where Nextcloud's files are not
available because only the app repository is cloned.

This is a little experiment, let's see how well that actually works …


ChristophWurst:
Scrutinizer analyzed the package ✅ 
bildschirmfoto von 2017-09-20 13-15-41

ChristophWurst:
> Scrutinizer — 11 new issues, 2 updated code elements

Exactly what I expected: it now finds exception classes it reported as non-existing before and complains about wrong parameter types passed to OCP classes/interfaces. Awesome 🚀 🤖 💃

Issue #543, opened 2017-09-25
[enhancement] User login


Extarys:
<details>
<summary>Disclaimer: Don't be mad - just share</summary>

```
As per a user suggestion on the forum, I'll start making enhancement suggestions
and bug reports (a lot). Hope someone can tag them for sorting. I got imagination
and time. I may have duplicates, the search function usually doesn't return things
when I search for them, even for 30 minutes, I apologies in advance, please don't
get mad. We are all trying to build great stuff here !
```

</details>

**Enhancement**

Make an option to link current user's email address with the server's own imap/smtp server.
If the option is enabled, the login screen would disappear and the inbox would be displayed.

Could also offer the user the possibility to log in with this email or another email.

Later on, an option could be added to connect to external email addresses (Google, Microsoft) to have an unified mailbox.

Also, the login form could use a little modern touch.

Issue #536, opened 2017-09-19
Configure mail app for IMAP authenticated users


rayfun:
### Steps to reproduce
1. Login to Nextcloud through IMAP auth backend
2. Go to mail app

### Expected behaviour
Mail app should be setup with auth credentials from IMAP authentication and a set of preconfigured values for host, port etc

### Actual behaviour
IMAP authentication backend works properly but mail app needs to be configured manually.

### Mail app

**Mail app version:** (see apps admin page)
0.7.3

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
Postfix 3.1.0-3/Dovecot 2.2.22-1ubuntu2.4

**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 

**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)

**Number of accounts:**

**Mail app version build date:** (only if you are using a Nightly Build)

### Server configuration

**Operating system**:
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.3 LTS"

**Web server:**
2.4.18-2ubuntu3.4
**Database:**

**PHP version:**

**Version:** (see admin page)

**Updated from an older version or fresh install:**

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your server installation folder
```

**The content of config/config.php:**

<details>
```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```
</details>

#### Client configuration
**Browser:**

**Operating system:**

##### Web server error log
```
Insert your webserver log here
```

##### Server log (data/nextcloud.log)
<details>
```
Insert your server log here
```
</details>

##### Horde IMAP log (data/horde_imap.log)
<details>
```
Insert your horde IMAP log here, see https://github.com/nextcloud/mail#debug-mode
```
</details>

##### Horde SMTP log (data/horde_smtp.log)
<details>
```
Insert your horde SMTP log here, see https://github.com/nextcloud/mail#debug-mode
```
</details>

##### Browser log
<details>
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log 
c) ...
```
[...]

PR #544, opened 2017-09-26
Enhanced doc

ChristophWurst:
Documents https://github.com/nextcloud/mail/pull/28
Fixes https://github.com/nextcloud/mail/issues/543, https://github.com/nextcloud/mail/issues/536#issuecomment-330537927

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,16 @@ And in the works for the [coming versions](https://github.com/nextcloud/mail/mil
 
 If you experience any issues or have any suggestions for improvement, use the [issue tracker](https://github.com/nextcloud/mail/issues). Please follow the [issue template](https://raw.githubusercontent.com/nextcloud/mail/master/.github/issue_template.md) so we get the info needed to debug and fix the problem. Thanks!
 
+## Development setup
+
+Just clone this repo into your apps directory (Nextcloud server installation needed). Additionally, [npm](https://www.npmjs.com/) to fetch [Node.js](https://nodejs.org/en/download/package-manager/) is needed for installing JavaScript dependencies.
+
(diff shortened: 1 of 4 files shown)


ChristophWurst:
See https://github.com/nextcloud/mail/tree/doc/enhanced-doc for a preview of what it will look like :)

jancborchardt:
Good point to move the advanced stuff out of the readme to keep it a bit cleaner. :) I would however keep the sections Installation, Developer setup info and Keyboard shortcuts. As that is the essential info and should be directly visible – the Readme is some kind of a landing page after all.

Advanced stuff like resetting, improving, further configuration etc can go in an »Advanced configuration« document I’d say.

ChristophWurst:
> Installation, Developer

IMO those actually *are* advanced topics. Usually people shouldn't have to know about the repo at all, as they install the app comfortably via the app store. They will only look for code on GitHub if they've found a bug, have questions/requests or want to contribute. The updatd readme would even list where to find what ;)

jancborchardt:
> They will only look for code on GitHub if they've found a bug, have questions/requests or want to contribute.

Ok, but then the developer setup instructions should be in the readme. :) The other things (install, keyboard shortcuts) can stay out.

ChristophWurst:
> Ok, but then the developer setup instructions should be in the readme still. :) The other things (install, keyboard shortcuts) can stay out. It’s very important to get people involved that the dead simple developer setup is immediately visible.

Done

ChristophWurst:
> Done

Apparently that was the only objection by @jancborchardt, thus I'm merging this PR now that its resolved.

ChristophWurst:
> Damn right that was my only objection ;)

PR #551, opened 2017-09-27
Cleanup address handling

ChristophWurst:
* Server-side
  + [x] Create our own `Address` class that serializes to an useful format for the client side
  + [x] Create a `AddressList` to wrap a list of addresses, e.g. for the recipients, cc and bcc
  + [x] Return only (serialized) address lists in responses
  + [x] Write and adjust tests
  + [x] Return `AddressList` in `Message::parseAddressList`
* Client-side
  + [x] Make use of address lists in messages list
  + [x] Make use of address lists in the message view
  + [x] Make use of address lists in the composer view
  + [x] Get rid of any `replyTo`, `replyCC` and similar stuff. The composer view must not have to know whether it's a reply or now. The surrounding view/logic has to handle that
  + [x] Make use of address lists in notifications
  + [x] Make use of address lists in (resumed) drafts
  + [x] Find out what's the correct/standard way of handling to/cc fields for group replies, ref comment `on reply, fill cc with everyone from to and cc except yourself`
  + [x] Test/fix reply handling (composer constructor args might have to be adjusted)
  + [x] mailto handling (to,cc and such are passed as strings rather than objects)

### DIFF
--- a/js/tests/replybuilder_spec.js
+++ b/js/tests/replybuilder_spec.js
@@ -1,0 +1,156 @@
+/* global expect */
+
+/**
+ * @copyright 2017 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2017 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+define([
+	'replybuilder',
+	'models/message',
+	'models/folder',
+	'models/account',
+	'backbone'
+], function(ReplyBuilder, Message, Folder, Account, Backbone) {
+
+	describe('ReplyBuilder', function() {
+
+		var message, messageBody, folder, account;
+
+		beforeEach(function() {
+			message = new Message();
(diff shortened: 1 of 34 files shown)


ChristophWurst:
⚠️ This PR changes cached data structures, hence developers have to manually clear their browser cache by running `sessionStorage.clear()` in the browser's dev console. ⚠️ 

On production instances, the update mechanism will take care of this as the version number increases.

jancborchardt:
Will test this in my local instance where I use my normal Mail accounts too.

Btw there’s some conflicts here.

js/tests/replybuilder_spec.js
```diff
@@ -0,0 +1,156 @@
...
+			messageBody.set('to', [b, c]);
+			messageBody.set('cc', [d, e]);
+			setEmail(message, e);
+
+			var reply = ReplyBuilder.buildReply(message, messageBody);
+
+			assertSameAddressList(reply.from, [e]);
+			assertSameAddressList(reply.to, [a, b, c]);
+			assertSameAddressList(reply.cc, [d]);
+		});
```
ChristophWurst: @jancborchardt if you could come up with a few more typical and thus important from, to, cc combinations we should have covered, please add them as comment and I'll add them.

ChristophWurst:
> For Sent and Drafts the placement of to and cc seems fine. But for the inboxes and the unified inbox too, my own email is filled in the »to« field. 💥 ;)

Fixed. This was surprisingly complicated/confusing to implement for all kind of to/from/cc combinations but I think I now cover most cases :)

ChristophWurst:
> This was surprisingly complicated/confusing

And that's the reason why I've extracted the code into its own module and have a bunch of unit tests for it.

ChristophWurst:
@jancborchardt ping :)

ChristophWurst:
FYI: I'm setting the *merge without review timeout* to one week 😉

jancborchardt:
Seems to work now, only sometimes it still says »[object Object]« before the quoted text in the content instead of the sender name.

ChristophWurst:
How's that *sometimes* triggered? Is it reproducible?

jancborchardt:
I guess it might be some caching, only for mails which I didn't open before, or only plaintext mails I didn't open before.

Did you never see that at all?

jancborchardt:
Well, let’s finally merge this anyway and deal with issues going forward. :)

ChristophWurst:
Thanks!

PR #556, opened 2017-10-03
Fix phpunit warnings and apply stricter CI checks

ChristophWurst:
Since a few weeks I'm constantly seeing four warnings whenever
I attempt to run phpunit locally. As they didn't make out CI jobs
fail, nobody seemed to care. As that's a risky situation, I've
fixed all the reported warnings and - to ensure this doesn't happen
again - I've enabled failure on warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ script:
 
   # Run PHP tests
   - cd tests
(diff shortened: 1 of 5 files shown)


ChristophWurst:
CI is happy again -> merging

Issue #553, opened 2017-09-28
Closing file chooser of »Save all to Files« leaves button inactive and loading


jancborchardt:
### Steps to reproduce
1. Go to an email with multiple attachments
2. Click on »Save all to Files«
3. Close the file chooser popup

### Expected behaviour
The button is active again.

### Actual behaviour
The »Save all to Files« button is inactive and loading:
![screenshot from 2017-09-28 12-23-16](https://user-images.githubusercontent.com/925062/30961910-18a1372a-a448-11e7-84a4-2b1d94c2828c.png)

ChristophWurst:
Fixed by #559

PR #559, opened 2017-10-03
Fix loading status for Save attachments button

haivp3010:
Fix loading status for Save attachments button when closing file chooser without choosing. This fix is for both saving all attachments as well as saving a single attachment

### DIFF
--- a/js/controller/messagecontroller.js
+++ b/js/controller/messagecontroller.js
@@ -127,8 +127,9 @@ define(function(require) {
 			OC.dialogs.filepicker(
 				t('mail', 'Choose a folder to store the attachment in'),
 				function(path) {
-					if (typeof callback == 'function')
+					if (typeof callback === 'function') {
 						callback();
+					}
 					Radio.message.request('save:cloud', account,
 						folder, messageId, attachmentId, path).then(function() {
 						if (saveAll) {
--- a/js/controller/messagecontroller.js
+++ b/js/controller/messagecontroller.js
@@ -118,6 +118,7 @@ define(function(require) {
 	 * @param {Folder} folder
 	 * @param {number} messageId
 	 * @param {number} attachmentId
+	 * @param {function} callback
 	 * @returns {Promise}
 	 */
 	function saveAttachmentToFiles(account, folder, messageId, attachmentId, callback) {
@@ -205,6 +206,7 @@ define(function(require) {
 	 * @param {Account} account
 	 * @param {Folder} folder
 	 * @param {number} messageId
+	 * @param {function} callback
 	 * @returns {Promise}
 	 */
 	function saveAttachmentsToFiles(account, folder, messageId, callback) {
--- a/js/controller/messagecontroller.js
+++ b/js/controller/messagecontroller.js
@@ -120,13 +120,14 @@ define(function(require) {
 	 * @param {number} attachmentId
 	 * @returns {Promise}
 	 */
-	function saveAttachmentToFiles(account, folder, messageId, attachmentId) {
+	function saveAttachmentToFiles(account, folder, messageId, attachmentId, callback) {
 		var saveAll = _.isUndefined(attachmentId);
 
 		return new Promise(function(resolve, reject) {
 			OC.dialogs.filepicker(
 				t('mail', 'Choose a folder to store the attachment in'),
 				function(path) {
+					callback();
 					Radio.message.request('save:cloud', account,
 						folder, messageId, attachmentId, path).then(function() {
 						if (saveAll) {
@@ -204,8 +205,8 @@ define(function(require) {
 	 * @param {number} messageId
 	 * @returns {Promise}
 	 */
-	function saveAttachmentsToFiles(account, folder, messageId) {
-		return saveAttachmentToFiles(account, folder, messageId);
+	function saveAttachmentsToFiles(account, folder, messageId, callback) {
+		return saveAttachmentToFiles(account, folder, messageId, null, callback);
 	}
 
 	return {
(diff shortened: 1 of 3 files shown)


haivp3010:
#553

ChristophWurst:
> First-time contributor

First of all, congratulations to your first PR to this repo and happy hacktoberfest 🎆 🚀 🎃

js/controller/messagecontroller.js
```diff
@@ -120,13 +120,14 @@ define(function(require) {
 	 * @param {number} attachmentId
 	 * @returns {Promise}
 	 */
-	function saveAttachmentToFiles(account, folder, messageId, attachmentId) {
+	function saveAttachmentToFiles(account, folder, messageId, attachmentId, callback) {
```
ChristophWurst: @haivp3010 don't know whether you're familiar with the concept of promises, but this function returns one. This means we get an object that we can register callback on, which are executed as soon as the asynchronous task has finished. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises for more info 😉

js/views/messageattachment.js
```diff
@@ -68,14 +68,14 @@ define(function(require) {
 			var folder = require('state').currentFolder;
 			var messageId = this.model.get('messageId');
 			var attachmentId = this.model.get('id');
-			// Loading feedback
-			this.getUI('saveToCloudButton').removeClass('icon-folder')
```
ChristophWurst: Why did you remove this? Doesn't this remove the initial loading feedback?

js/views/messageattachment.js
```diff
@@ -68,14 +68,14 @@ define(function(require) {
...
-				.addClass('icon-loading-small')
-				.prop('disabled', true);
 
 			var _this = this;
-			MessageController.saveAttachmentToFiles(account, folder, messageId, attachmentId)
-				.catch(console.error.bind(this)).then(function() {
+			MessageController.saveAttachmentToFiles(account, folder, messageId, attachmentId, function() {
+				// Loading feedback
+				_this.getUI('saveToCloudButton').removeClass('icon-folder')
```
ChristophWurst: see five lines below, the code is already there 😉 For some reason it doesn't seem to work though. Try to debug and fix that instead please

js/views/messageattachments.js
```diff
@@ -57,14 +57,14 @@ define(function(require) {
...
-				.addClass('icon-loading-small')
-				.prop('disabled', true);
 
 			var _this = this;
-			MessageController.saveAttachmentsToFiles(account, folder, messageId)
-				.catch(console.error.bind(this)).then(function() {
+			MessageController.saveAttachmentsToFiles(account, folder, messageId, function() {
+				// Loading feedback
+				_this.getUI('saveAllToCloud').removeClass('icon-folder')
+				.addClass('icon-loading-small')
```
ChristophWurst: same as above, you can find the promise callback a few lines below

js/controller/messagecontroller.js
haivp3010: I do know about promises, but in this case, I just haven't found out if that close event of the file chooser calls the reject method of the promise.

js/views/messageattachment.js
haivp3010: I wanted to make the button show loading only when user clicks Choose to choose a directory. Shouldn't the button stay the same unless user confirms saving?

js/views/messageattachment.js
haivp3010: I moved it here to only show loading when user confirms directory, as above.

haivp3010:
Thanks, I am looking into that now :)

js/controller/messagecontroller.js
ChristophWurst: I don't know that either as I haven't checked. But from the diff here on GitHub I can tell that you're invoking the callback as soon as the user selected a directory. However, the loading feedback should be there until the task of actually storing the attachments has been finished.

ChristophWurst: Wait. Ignore my comment, now I see what you did there 🙈 The callback was added for *adding* the feeback, not for removing.

ChristophWurst:
> Thanks, I am looking into that now :)

Might be that you're actually correct with your patch. I will check the code soon :)

js/controller/messagecontroller.js
haivp3010: Yeah. That's what I meant: no loading until user has chosen a directory, and loading disappears after upload done. I have checked that on my localhost and it went fine. It's just that I couldn't find a way to achieve this with promise, with the existing code. I had thought of catching the close event on the file chooser and call reject, but that would require making changes to core js files of nextcloud server as well, so I had to resort to callback like that.

haivp3010:
Thanks. I have just pushed a small patch for checking the callback. Please check back when you have time :)

haivp3010:
Any progress with this yet? @ChristophWurst

ChristophWurst:
> Any progress with this yet? @ChristophWurst

No, because I'm busy with other stuff, sorry. Can anybody of @nextcloud/mail help here please? Thanks.

js/views/messageattachment.js
jancborchardt: @haivp3010 the loading feedback should appear as soon as the button is clicked already. So if the popover does _not_ overlap the button you see the loading feedback already.

haivp3010: @jancborchardt I think that the purpose of the loader is to represent file saving status. For the file picker, it does have its own loading spinner while loading up, so why should we leave the former as loading while it's function is actually not being executed?

haivp3010:
It's ok, I was just kind of eager :) Just give it a look later when you have time. Thanks.

js/views/messageattachment.js
haivp3010: A similar situation can be seen here https://developers.google.com/drive/v3/web/savetodrive, where the Save to drive button does not show as loading while still choosing a folder.
image
The loader only shows when the file is actually being uploaded.
image

Of course, different systems have different ways to do things, but I think it's just not common practice to show the loading feedback so early while still choosing a directory.

PR #565, opened 2017-10-10
Clean up navigation views

ChristophWurst:
Some cleanups, originally written to replace #521 but no valid solution was found. IMO still an improvement and thus I would like to integrate this.

### DIFF
--- a/js/views/accountview.js
+++ b/js/views/accountview.js
@@ -41,7 +41,10 @@ define(function(require) {
 		},
 
 		regions: {
-			folders: '.folders'
+			folders: {
+				el: '.folders',
+				replaceElement: true
+			}
 		},
 
 		ui: {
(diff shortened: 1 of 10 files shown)


js/views/accountview.js
```diff
@@ -18,51 +18,88 @@ define(function(require) {
...
 			'click @ui.deleteButton': 'onDelete',
 			'click @ui.settingsButton': 'showAccountSettings',
 			'click @ui.email': 'onClick'
 		},
+
 		regions: {
-			folders: '.folders'
+			folders: {
+				el: '.folders',
+				replaceElement: true
```
ChristophWurst: @skjnldsv FYI this remove one of the wrapping elements

js/views/navigation.js
```diff
@@ -17,33 +17,43 @@ define(function(require) {
...
 	return Marionette.View.extend({
+
 		el: '#app-navigation',
+
 		regions: {
 			newMessage: '#mail-new-message-fixed',
-			accounts: '#app-navigation-accounts',
+			accounts: {
+				el: '#app-navigation-accounts',
+				replaceElement: true
```
ChristophWurst: same here

PR #569, opened 2017-10-16
Fix js test execution

ChristophWurst:
The app must not be started, otherwise random stuff breaks
as timers, http requests and DOM manipulation is done in
the background.

Found while debugging failing/erroring tests in https://github.com/nextcloud/mail/pull/551.


ChristophWurst:
CI passes -> merging

PR #583, opened 2017-10-30
Fix loading of HTML messages

ChristophWurst:
Dunno how or when this regression slipped into our code base, but I'm pretty sure it was me 🙈 

Anyway, HTML email seem to work again 🚀


jancborchardt:
@ChristophWurst giveth and @ChristophWurst taketh away! PRAISE THE LORD :crossed_swords:

Issue #577, opened 2017-10-25
Release tarball contains developement files


onny:
Hey,
the latest release tarball (mail v.0.7.3) contains developement files which shouldn't be included:
```
[...]/nextcloud/apps/mail/vendor/kwi/urllinker/.hg
```
These files cause a file integrity check error in Nextcloud.

ChristophWurst:
Whoops, that shouldn't have happened.

PR #584, opened 2017-10-30
Fix package size

ChristophWurst:
* Don't package composer dev deps
* Don't package the .phan dir
* Don't package .hg dirs

Fixes https://github.com/nextcloud/mail/issues/577


ChristophWurst:
Wow, those build are slow today …

PR #594, opened 2017-11-02
Fix CSS in app-navigation

jancborchardt:
This is pretty much done now! Needs some testing with subfolders @nextcloud/mail 

TODO where I need JS help:

- [ ] Don’t show the counter when there are 0 unread mails
- [x] Fix popover position for account settings
- [x] Clicking »Show all folders« and collapse should not reload the main content

screenshot from 2017-11-02 17-26-18

### DIFF
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -37,7 +37,7 @@
 {{#unless isUnifiedInbox}}
 {{#if hasFolders}}
 <li class="account-toggle-collapse">
-	<a href="">{{toggleCollapseMessage}}</a>
+	<a>{{toggleCollapseMessage}}</a>
 </li>
 {{/if}}
 {{/unless}}
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -1,7 +1,7 @@
 {{#unless isUnifiedInbox}}
 <li class="{{#if hasMenu}} with-menu {{/if}}">
 	{{#if emailAddress}}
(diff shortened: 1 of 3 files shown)


ChristophWurst:
> Clicking »Show all folders« and collapse should not reload the main content

Wild guess: did you somehow break the click handler binding and the app tries to navigate to `#`?

ChristophWurst:
> Don’t show the counter when there are 0 unread mails

Easy, I can do that. Doesn't have to be in this PR as that's independent.

ChristophWurst:
> Fix popover position for account settings

Why do we need js for this?

js/templates/account.html
```diff
@@ -36,6 +36,8 @@
 <div class="folders"></div>
 {{#unless isUnifiedInbox}}
 {{#if hasFolders}}
-<li class="account-toggle-collapse">{{toggleCollapseMessage}}</li>
+<li class="account-toggle-collapse">
+	<a href="">{{toggleCollapseMessage}}</a>
```
ChristophWurst: here you go. The newly added `a` element causes an unexpected navigation. The click handler should call `event.preventDefault()`.

skjnldsv: preventDefault isn't working here, wefound something else. ;)

jancborchardt:
Pushed a fix for the other two issues, the position was indeed CSS only. :)

ChristophWurst:
❤️

Issue #590, opened 2017-11-01
Prefilled reply text shows "[object Object]" instead of name of original sender


jancborchardt:
With latest Mail master and plain text mails only:

```
[object Object] – Wed., 1. November 2017 14:56
> Hello there, 
>
```

@nextcloud/mail can you see that too?

ChristophWurst:
> @nextcloud/mail can you see that too?

Yup, confirmed.

PR #600, opened 2017-11-06
Fix rendering of original sender in reply body

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/590

### DIFF
--- a/js/views/composerview.js
+++ b/js/views/composerview.js
@@ -420,7 +420,7 @@ define(function(require) {
 
 			this.$('.message-body').first().text(
 					'\n\n\n' +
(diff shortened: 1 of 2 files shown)


jancborchardt:
If realname and email are present, only the realname should be shown. If only the emai is present, show only that. That’s what we had before at least, and how it should be. :)

ChristophWurst:
> If realname and email are present, only the realname should be shown. If only the emai is present, show only that. That’s what we had before at least, and how it should be. :)

Okidoke

ChristophWurst:
Done

Issue #591, opened 2017-11-01
»to« field of a reply not correctly filled


jancborchardt:
As mentiond in https://github.com/nextcloud/mail/issues/541#issuecomment-341117259, try with an email like this:

Sent from Ali to Bob, yourself and Dani in cc

The reply field will have Ali in the to, the cc people in the cc again. But the original people from the »to«, whether it’s one or multiple, will not be in the »to« of the reply.

PR #603, opened 2017-11-06
Fix another case of reply to/cc handling

ChristophWurst:
@jancborchardt I've started fixing https://github.com/nextcloud/mail/issues/591 with writing a test and to my surprise this is correctly handled already. Please see the test code. Did I misunderstand your reproduction steps?

Local test execution blocked by https://github.com/nextcloud/mail/pull/602


jancborchardt:
Here’s a simpler testcase where I have an error still:
- Nina sends a mail to the company mailing list
→ in the reply, Nina is in the »to«, but no one is in the cc (should be company mailing list)

ChristophWurst:
> a simpler testcase

Sounds simple, yes. But extracting logic out of this is a different story 😉 

So, if you reply to an email where you're neither in the TO field nor in CC, how do you know where to answer to?

It then really depends on the reply strategy, like *answer all*, *answer to group/list* or however mail clients name it.

*answer to group* would be a sensible default IMO, but how do we know what the list is? Can we safely assume only there's only one recipient in that case and that one is the list? Or could there be the list **and** a regular recipient? Then it gets tricky because based on the address I cannot tell which one is a list and which one isn't.

jancborchardt:
We always did answer all. So in my example above, the reply should have Nina in to (cause I'm replying directly to her) and the list in cc.

ChristophWurst:
> the reply should have Nina in to (cause I'm replying directly to her)

Okay, *reply all* it is. Let me see if I can put that into code.

ChristophWurst:
Fixed.

ChristophWurst:
Blocked by #621

PR #610, opened 2017-11-13
Update npm deps

ChristophWurst:
This is the result of running `npm update`.


ChristophWurst:
Didn't explode so far, hence I'm merging this minor change now.

PR #611, opened 2017-11-13
Fix js tests by excluding build artefacts

ChristophWurst:
Yet again, js tests wouldn't run on my local instance whereas on Travis the always pass 😠 

Turns out we include the `build.js` artefact and thus the app starts up and causes all kinds of weird errors 🙈


ChristophWurst:
CI is happy and solved the local dev problem for me -> merging

Issue #613, opened 2017-11-14
Subfolders cannot be selected, parent folder is opened instead


ChristophWurst:
https://github.com/nextcloud/mail/pull/606#pullrequestreview-76539071

PR #619, opened 2017-11-20
Fix subfolder click event propagation

ChristophWurst:
This prevents unintended loading of the parent folder if a subfolder
is clicked because the event is now only handled one.

Note that for some reasons not all accounts are effected. Somehow my Gmail account's subfolder can be opened even without this fix.

Fixes https://github.com/nextcloud/mail/issues/613


ChristophWurst:
Bocked by https://github.com/julien-maurel/js-storage/pull/8

Issue #589, opened 2017-11-01
Settings auto-detect fails with "Error while creating the account: Internal Server Error"


Skymirrh:
### Steps to reproduce
Add a new Gmail account without using manual configuration.

### Expected behaviour
Gmail connection settings (imap.gmail.com/993/SSL and smtp.gmail.com/465/SSL) should be auto-detected and account created.

### Actual behaviour
Fails with `Error while creating the account: Internal Server Error`: account is not created.
When checking server log, the following error message appears `Error: Call to undefined function OCA\Mail\Service\AutoConfig\no_null()`, along with a stack trace (see below).

### Mail app
**Mail app version:** 0.7.4
**Mailserver or service:** Gmail
**Transport security - IMAP:** SSL
**Transport security - SMTP:** SSL
**Number of accounts:** 1

### Server configuration
**Operating system**: Debian 9 (Stretch)
**Web server:** nginx/1.10.3 (fpm-fcgi)
**Database:** mysql 10.1.26-MariaDB
**PHP version:** 7.0.19-1
**Nextcloud version:** 12.0.3 - 12.0.3.3
**Updated from an older version or fresh install:** Fresh install

##### Server log (data/nextcloud.log)
<details>

```
Error: Call to undefined function OCA\Mail\Service\AutoConfig\no_null()
0. /var/www/files/<REDACTED>/cloud/apps/mail/lib/Service/AutoConfig/IspDbConfigurationDetector.php - line 93: OCA\Mail\Service\AutoConfig\IspDbConfigurationDetector->detectImap(Array, '<REDACTED>@gmail.co...', '<REDACTED>', '<REDACTED>')
1. /var/www/files/<REDACTED>/cloud/apps/mail/lib/Service/AutoConfig/AutoConfig.php - line 51: OCA\Mail\Service\AutoConfig\IspDbConfigurationDetector->detectImapAndSmtp('<REDACTED>@gmail.co...', '<REDACTED>', '<REDACTED>')
2. /var/www/files/<REDACTED>/cloud/apps/mail/lib/Service/SetupService.php - line 72: OCA\Mail\Service\AutoConfig\AutoConfig->createAutoDetected('<REDACTED>@gmail.co...', '<REDACTED>', '<REDACTED>')
[...]

PR #620, opened 2017-11-20
Fix autoconfig null check

ChristophWurst:
`no_null` should be `is_null`

Fixes https://github.com/nextcloud/mail/issues/589

Testd with Gmail and now I can create a autoconfigured account again.

cc @Skymirrh


ChristophWurst:
Bocked by https://github.com/julien-maurel/js-storage/pull/8

PR #623, opened 2017-11-20
Update some js dev deps

ChristophWurst:
Found out that most of our dev deps are pretty outdated and thus decided to give them a version bump. I had to adjust some of the jsdoc to make it pass with the newer and apparently more strict jscs task.


ChristophWurst:
- [ ] Requires #622

ChristophWurst:
CI is green and it's *just* dev deps -> merging

PR #625, opened 2017-11-20
Save autoconfigured account too

ChristophWurst:
Still think it's funny how nobody noticed this 🙈 

Well, at least now we've got some tests that make sure the account is always saved.

### DIFF
--- a/lib/Service/SetupService.php
+++ b/lib/Service/SetupService.php
@@ -73,6 +73,9 @@ class SetupService {
 		if (is_null($mailAccount)) {
 			return null;
 		}
+
+		$this->accountService->save($mailAccount);
+
 		return new Account($mailAccount);
 	}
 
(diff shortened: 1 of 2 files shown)


skjnldsv:
To be fair I always end up with a timeout or 500 with the autoconfig. 🙈

ChristophWurst:
> To be fair I always end up with a timeout or 500 with the autoconfig. 🙈

Known problem. And I even know the solution to it but still haven't found the time to implement …

skjnldsv:
@ChristophWurst

OH! I always thought it was coming from my setup! ^_^

ChristophWurst:
No it's because we iterate over dozens of possible configurations without any concurrency. That takes a lot of time. Hence I'd like to move some of that logic to the client where we can make use of async requests and this get a concurrent process.

ChristophWurst:
Fix worked for me, is covered by unit tests and CI is green -> merging.

Issue #644, opened 2017-11-28
Provide opt-out setting for the new avatar feature


jancborchardt:
In the bottom left settings area we should have a »Get avatars from online sources« checkbox which is enabled by default. So that people can disable the checking on Gravatar and external domains.

(And yes, it should be enabled by default as it improves the experience by miles.)

fti7:
Please also provide an Option to do this on a Global-Level (Administrator Settings)

jancborchardt:
@fti7 please test the pull request at https://github.com/nextcloud/mail/pull/653 and comment there accordingly. :)

PR #653, opened 2017-12-04
Add opt-out setting for Gravatar integration

ChristophWurst:
- [x] Add settings controller
- [x] Add server-side settings service
- [x] Add client-side settings service
- [x] Add button and click handler to settings view
- [x] Tests
- [x] Squash fixup commits

Fixes https://github.com/nextcloud/mail/issues/644

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -96,11 +96,12 @@ $app->registerRoutes($this,
 		],
 	],
 	'resources' => [
-		'autoComplete' => ['url' => '/api/autoComplete'],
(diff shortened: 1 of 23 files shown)


ChristophWurst:
Note to anybody wondering: if you disable external avatars, already fetched ones are not removed. It just prevents the back-end from loading new avatars. IMO this is *good enough*.

ChristophWurst:
In doubt of getting reviews I'll merge this after a second manual test. To some extend the functionality has been unit tested anyway, so I'm quite confident this should work as expected.

ChristophWurst:
Seems good -> merging

PR #657, opened 2017-12-06
Test on php7.1 on Travis

ChristophWurst:
Once [php7.2 is available for Travis](https://docs.travis-ci.com/user/languages/php/), we should test against it as well.


ChristophWurst:
No objections -> merge

Issue #661, opened 2017-12-11
Undefined index: ccList


ChristophWurst:
![bildschirmfoto von 2017-12-11 21-02-24](https://user-images.githubusercontent.com/1374172/33851273-bb6aeb1a-deb6-11e7-9d5e-2cceab721d6c.png)

Happens with the latest release.

PR #668, opened 2017-12-12
Fix undefined indices

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/661


ChristophWurst:
Note that I don' really get what that code does. But now the indices do definitely exist and I could view my 'sent' folder without any issues.

PR #672, opened 2017-12-12
Allow and test against php7.2

ChristophWurst:


ChristophWurst:
CI is happy and I've used it locally on my dev instance with php7.2 too already. Should be good.

PR #681, opened 2017-12-18
Migrate away from some global js dependencies

### LINKED ISSUE #626: Missing close or choose button in file picker dialog since 0.7.5
Hey,
the file picker dialog is missing a close or save button:
2017-11-22-125411_swaygrab
But maybe this is related to core?
I'm using Nextcloud Beta 13 and the latest mail app (from git).

Best regards,
Jonas

ChristophWurst:
~~This is WIP and non-working right now. Not sure if this problem is even solvable. However, this messed up dependency issue already caused bugs, so we kinda have to find a solution for it. It's complicated somehow …~~

I could fix some of the issues. Now at least the js-part is working. The close button on the file picker is also back.

Things I couldn't migrate:
* Tooltips (they look super ugly if I use the up-to-date jquery-ui lib, not sure why)
* Imageplaceholde (it's registered as plugin on the global jquery instance. should be a library though)

Fixes #626


ChristophWurst:
Oh my … this jquery ui thingy is completely borked. Even though here we don't have any version conflict (since I intentionally relied only on global code and none of the shipped dependencies, it clearly shows that even there, the code only runs because jquery-migrate is involved. That looks like Nextcloud server already ships incompatible libraries.

bildschirmfoto von 2017-12-19 08-41-34

ChristophWurst:
@jancborchardt I'd need your help with this. Would you mind taking a look at the current state and fix whatever seems to be broken. The only thing I noticed is a strange blue border when I hover the results of the composer autocomplete list.

ChristophWurst:
@jancborchardt ping

jancborchardt:
> a strange blue border when I hover the results of the composer autocomplete list.

Fixed that. On hover it still jumps a bit but that is the same for Files, so we need to fix it in core.

ChristophWurst:
@jancborchardt thanks a lot. I assume your assignment symbolizes everything has been addressed, I can resolve the conflicts and merge?

laurentiu2:
Does not work for me in 4.7.7
screen shot 2018-01-10 at 11 03 08 am

PR #683, opened 2017-12-19
Migrate to a more sophisticated HTML to text lib

ChristophWurst:
This converts most HTML emails pretty nicely. However, some that have large numbes of URLs in them look still pretty bad. This needs some fine-tuning.

* [x] Fix messages with many URLs
* [x] Fix existing unit tests
* [x] Add more unit tests


ChristophWurst:
This seems to render HTML message replies very well. Even some pretty nasty looking HTML results in a somewhat readable format. Not perfect, but remarkably better than before.

ChristophWurst:
If you're curious about how it's configured but you're too lazy to actually test this PR, please take a look at https://github.com/nextcloud/mail/pull/683/files#diff-2f524624b8ba34c22daad0bcdfdd60be and then hit that approve button. Thanks.

PR #684, opened 2017-12-20
Add warning if replying to noreply@...

jakobsack:
This PR disable the reply field if the sender of the email is noreply@...


jancborchardt:
Huh, good one! However sometimes stuff is from a noreply and you want to forward it, or reply to a different one.

So how about when it's from a noreply, and you don't change the "to" address in the reply, the "to" field is highlighted and also has a tooltip saying "As this is a noreply email address, you are unlikely to get a reply."

jakobsack:
Good point @jancborchardt .
With the current version it looks like this:
screenshot_20171222_213805

I don't think that this is too bad.

jancborchardt:
It is quite a different and confusing break in the interfave, and as I said sometimes you wanna reply, changing the "to" address. Like from noreply@example.com to info@example.com you found on their website. ;)

jakobsack:
Hey Jan, I've changed the behaviour now. It shows the reply form but with a warning:

screenshot_20171228_124358

The system has one drawbacks: the warning does not get updated when the user changes the to addresses

jancborchardt:
Cool! Some feedback:

- Can you move it a bit up so it's directly below the "to" field
- The text should be moved to the right so it's left-aligned with all the other fields (like "to" directly above)
- Better wording: "Note that the mail came from a noreply address so your reply will probably not be read."
- Yellow would be better as a color here since red is for grave issues.
- Ideally if it's changed to a different address the notice should indeed vanish (check again if it's a noreply) to take care of the drawback you mentioned.

jakobsack:
Hi @jancborchardt , I implemented your styling suggestions.
screenshot_20180117_214604
Any comments?

I'm unable to show/hide the text depending on the recipients, it would be great if someone else could take care of this.

ChristophWurst:
- [ ] Blocked by #726, please do not merge until the 0.7.9 release is out

PR #691, opened 2018-01-02
Make account settings editable

### LINKED ISSUE #16: Account settings editable [$50]
Currently you can only "edit" them by deleting and re-adding them. There should be a 2nd option in drop down navigation (under the 3 dots) to edit account settings

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/37587252-account-settings-editable?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pierlon:
Fixes #16.

### DIFF
--- a/js/views/setupview.js
+++ b/js/views/setupview.js
@@ -87,7 +87,7 @@ define(function(require) {
 			this._config = config;
 			this.render();
 
-			return Radio.account.request('create', config).then(function() {
+			return Radio.account.request('create', {config: config}).then(function() {
 				Radio.ui.trigger('navigation:show');
 				Radio.ui.trigger('content:loading');
 				// reload accounts
--- a/js/views/setupview.js
+++ b/js/views/setupview.js
@@ -87,7 +87,7 @@ define(function(require) {
 			this._config = config;
 			this.render();
 
-			return Radio.account.request('create', {config: config}).then(function() {
+			return Radio.account.request('create', config).then(function() {
 				Radio.ui.trigger('navigation:show');
 				Radio.ui.trigger('content:loading');
 				// reload accounts
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -37,12 +37,10 @@ use OCA\Mail\Service\AliasesService;
 use OCA\Mail\Service\AutoConfig\AutoConfig;
 use OCA\Mail\Service\Logger;
 use OCA\Mail\Service\SetupService;
-use OCA\Mail\Db\MailAccount;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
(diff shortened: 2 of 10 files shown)


pierlon:
@ChristophWurst @jancborchardt ?

js/views/setupview.js
```diff
@@ -87,7 +87,7 @@ define(function(require) {
 			this._config = config;
 			this.render();
 
-			return Radio.account.request('create', config).then(function() {
+			return Radio.account.request('create', {config: config}).then(function() {
```
ChristophWurst: in `accountservice.js`, you have `data: {config: config},`. Doesn't that mean you wrap that in a object property twice? Something like `{ config: { config: { ... } } }`?

lib/Controller/AccountsController.php
```diff
@@ -131,13 +130,48 @@ public function show($accountId) {
...
+	 * @param int $id
+	 * @param array $config
+	 * @return JSONResponse
 	 */
-	public function update() {
-		$response = new Response();
-		$response->setStatus(Http::STATUS_NOT_IMPLEMENTED);
-		return $response;
+	public function update($id, $config) {
+		return $this->createAccount($config, $id);
```
ChristophWurst: I'd say let's duplicate the method above instead of abusing it for the update. This kind of core re-use makes the code pretty hard to understand and thus might introduce new bugs.

ChristophWurst: FYI the autodetected and the manual mode should be two separate routes and separate methods. This cleanup is out of scope for this PR however.

tests/Controller/AccountsControllerTest.php
```diff
@@ -212,17 +215,19 @@ public function testCreateAutoDetectSuccess() {
 	}
 
 	public function testCreateAutoDetectFailure() {
```
ChristophWurst: while you're here, please also add unit tests for the update methods 😉

js/views/setupview.js
pierlon: It was intended for easier use in the controller. My preference really... Will revert change.

pierlon:
@ChristophWurst All done 😄

ChristophWurst:
@pierlon again, thank you very much for your first contribution to this app! Don't forget to claim your well deserved bounty for fixing https://github.com/nextcloud/mail/issues/16 😉 

I'm looking forward to seeing more such PRs 👍

pierlon:
Thanks! Looking forward to contribute more PRs :wink:

PR #710, opened 2018-01-15
Update horde libs

ChristophWurst:
Just bumping minor versions. Nothing will actually change.


ChristophWurst:
This could be a dup of #705

Issue #386, opened 2017-06-07
Subfolders and Dovecot


reubenlillie:
Please pardon my confusion, @ChristophWurst and @oparoz, but has the Dovecot namespace prefixing issue (e.g., #185) been fixed via #271 or not? I know #185 was closed, but was it premature, or is my issue new? That is, does the "not working correctly" that I'm experiencing relate to what #271 attempted to fix, or is my issue, like #372, a separate kind of bug? The readme is still vague, and I'd like to know if it's something I'm doing wrong or a limitation for the current software.

In any case, I'm having difficulty similar to #372 on my instance (unless I'm also misunderstanding #372):

### Expected behaviour
"Show all folders" displays folders beyond inbox (e.g., Archive, Sent, Drafts, Junk, etc.)

### Actual behaviour
"Show all folders" displays no more folders, instead only "Collapse folders" toggle prompt.

### Mail app

**Mail app version:** 0.6.4

**Mailserver or service:** Dovecot

**Transport security - IMAP:** SSL 

**Transport security - SMTP:** SSL

**Number of accounts:** 1 (testing for development, ~250 accounts planned for client)

### Server configuration
**Operating system**: Linux server.***{DOMAIN}*** 2.6.32-696.el6.x86_64 #1 SMP Tue Mar 21 19:29:05 UTC 2017 x86_64

**Web server:** Apache (cgi-fcgi)

**Database:** mysql 5.6.35

**PHP version:** 7.1.5
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, pcntl, readline, Reflection, SPL, session, standard, cgi-fcgi, bcmath, calendar, ctype, curl, dom, mbstring, ftp, gd, gettext, imap, json, exif, mcrypt, mysqlnd, PDO, Phar, posix, pspell, SimpleXML, soap, sockets, sqlite3, tidy, tokenizer, xml, xmlwriter, xsl, zip, mysqli, pdo_mysql, pdo_sqlite, wddx, xmlreader, xmlrpc

**Nextcloud version:** 12.0.0 - 12.0.0.29

**Updated from an older Nextcloud/ownCloud or fresh install:** fresh install

**Where did you install Nextcloud from:** https://nextcloud.com/install/

**Signing status:**
<details>
	<summary>Signing status</summary>

```
[]
```

</details>

**List of activated apps:**
<details>
	<summary>App list</summary>
```
Enabled:
 - activity: 2.5.2
[...]

PR #520, opened 2017-09-13
Fix dovecot sieve special folder

ChristophWurst:
Dovecot/sieve seems to create a special 'dovecot.sieve' folder as soon
as a filter is configured. This folder, however, must not be queried
(e.g.) for a sync token. This patch makes sure that this special folder
is simply ignored. Thus it's also not shown in the UI.

Fixes https://github.com/nextcloud/mail/issues/475

@digdug3 thanks a lot for providing an account on your mail server. This helped a lot 🚀

PR #652, opened 2017-12-04
Fix mailbox prefix hierarchy

ChristophWurst:
Apparently some mail servers prefix their mailboxes by `INBOX` and
hence the hierarchy logic previously detected those prefixes as parent
folders.

Ref https://github.com/nextcloud/mail/issues/386

This only fixes the hierarchy problem. Folders will still be shown with a prefix. Since I'm fixing a problem without actually having an IMAP implementation with prefixes at hand, I'll fix this in smaller steps and let the community verify the fix.

cc @reubenlillie

PR #728, opened 2018-01-22
Handle dovecot sieve folder when subdir of INBOX

j0s:
Pull request #520 introduced special handling of dovecot's
'dovecot.sieve' folder, so that it isn't queried for a sync
token. A common configuration of dovecot is to put all folders as
subfolders of INBOX. This pull request adds support for
cases when the sieve folder is a subfolder of INBOX.


ChristophWurst:
Hey, @j0s!

Thanks a lot for your PR. I think the proper solution would be to fix https://github.com/nextcloud/mail/issues/386 because then the prefix would be removed and thus that wouldn't be an issue, right?

j0s:
That looks like the proper solution, yes, if the solution involves removing the INBOX prefix from all  subfolders. However the fix in https://github.com/nextcloud/mail/pull/652#issuecomment-352427619 does not solve the problem for me. Unfortunately I'm not comfortable enough with docker to create an image for you to test.

ChristophWurst:
I'm fine with accepting this temporary fix if

* You fix the coding style and use camel case variable names and the short array syntax
* Add a source code comment that this is a ~~hack~~ erm … I mean temporary fix 😉 
* Add a notice to #386 so that we don't forge to remove this fix once that ticket has been addressed in a PR

ChristophWurst:
Thanks a lot!

ChristophWurst:
- [ ] blocked by #735

ChristophWurst:
@j0s FYI your email address used for the commits here is not associated with your account. You might want to add it to your account, then GitHub will correctly show you as author 😉

ChristophWurst:
> First-time contributor

Congrats on your first contribution btw 🙌 Looking forward to more contributions ✨ 🚀

PR #730, opened 2018-01-23
Fix/drafts concurrency

ChristophWurst:
This fixes the following issues

* Detection of when the composer is left was buggy because it had to be triggered explicitly in many parts of the code. Instead, we now simply use the marionette `onDetroy` view event handler that triggers - you guessed it - when the composer view is destroyed. That's exactly where we have to save the last draft state if necessary.
* Saving drafts can be triggered concurrently. However, the `saveDraft` method has to be called in-order to assure we get rid of all intermediate draft versions as previous draft saving requests return the UID we need for the successor request. By facilitating a js promise to run the draft saving mechanism serial, it's ensured that we do not leave any leftovers.
* When sending a message, we clean up previous draft versions and thus also need the most recent UID to do so. Thus, I've also included the message sending into this serialization. It might be a tad slower to send a message if draft saving is still in progress, but this is the one and only clean way to handle this.

@jancborchardt please give this a try. Maybe use the app for a few ours. You're the one that actually writes emails. I just read them 😉


ChristophWurst:
- [ ] blocked by #735

ChristophWurst:
> sh -c "if [ '$TEST_JS' = 'TRUE' ]; then grunt; fi"
> Running "jscs:src" (jscs) task
> >> 58 files without code style errors.
> Running "jshint:all" (jshint) task
>    js/views/composerview.js
>     138 |            this.draftPromise = Promise.resolve();
>                                          ^ 'Promise' is not defined.
> >> 1 error in 58 files
> Warning: Task "jshint:all" failed. Use --force to continue.
> Aborted due to warnings.

ChristophWurst:
@nextcloud/mail please test. Thank you!

PR #742, opened 2018-01-29
Handle folders without delimiter gracefully

j0s:
For some IMAP servers, Horde fails to detect a delimiter. In my case INBOX returned an empty string when calling getDelimiters(). This resulted in warnings in FolderMapper::guessSpecialUse and FolderMapper::getParentId. The following commit fixes the warnings, and enables us to
properly detect that INBOX is actually an inbox.

To be clear, I don't fully understand the implications of this change, other than that it seems to fix my problems, so please inform me if this is stupid :)

### DIFF
--- a/lib/Service/FolderMapper.php
+++ b/lib/Service/FolderMapper.php
@@ -111,7 +111,9 @@ class FolderMapper {
 	 */
 	private function getParentId(Folder $folder) {
 		$delimiter = $folder->getDelimiter();
-		if ($delimiter == '') {
+		if ($delimiter === '') {
+			// If no delimiter is detected, it's probably because
+			// this is a top level folder
 			return null;
 		}
 		$hierarchy = explode($delimiter, $folder->getMailbox());
@@ -208,7 +210,7 @@ class FolderMapper {
 		];
 
 		$delimiter = $folder->getDelimiter();
-		if ($delimiter == '') {
+		if ($delimiter === '') {
 			$lowercaseId = strtolower($folder->getMailbox());
 		} else {
 			$lowercaseExplode = explode($delimiter, $folder->getMailbox(), 2);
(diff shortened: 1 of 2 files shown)


lib/Service/FolderMapper.php
```diff
@@ -110,7 +110,11 @@ public function buildFolderHierarchy(array $folders) {
 	 * @return string
 	 */
 	private function getParentId(Folder $folder) {
-		$hierarchy = explode($folder->getDelimiter(), $folder->getMailbox());
+		$delimiter = $folder->getDelimiter();
+		if ($delimiter == '') {
```
ChristophWurst: use `===` please

ChristophWurst: and adding a explanatory comment doesn't hurt 😉

lib/Service/FolderMapper.php
```diff
@@ -203,8 +207,13 @@ protected function guessSpecialUse(Folder $folder) {
 			'junk' => ['junk', 'spam', 'bulk mail'],
 		];
 
-		$lowercaseExplode = explode($folder->getDelimiter(), $folder->getMailbox(), 2);
-		$lowercaseId = strtolower(array_pop($lowercaseExplode));
+		$delimiter = $folder->getDelimiter();
+		if ($delimiter == '') {
```
ChristophWurst: `===`

j0s:
I should really get a full development environment set up, so I can run the unit tests without committing :)

j0s:
OK, this is getting out of hand. Sorry about the commit spam! Will try to figure out how to run the unit tests locally, and then submit a new pull request with the changes squashed into one commit.

i-wan:
May I ask what's the state of this issue/merge? I'm facing same problem with delimiter being null so seems the update fixing $folder->getDelimiter() being null never reached the master, right?

[index] Error: TypeError: explode() expects parameter 1 to be string, null given at <<closure>>

with error at FolderMapper.php line 214 
    undefinedundefinedexplode(null, "INBOX", 2)

ChristophWurst:
please open a new ticket so we can discuss this. what version of the mail app do you use? let us know

ChristophWurst:
locking as resolved

Issue #598, opened 2017-11-05
Add a "reply" (as opposed to "reply all") action [$20]


mathiasconradt:
NC 12.0.3

I had posted about this in the help forum: https://help.nextcloud.com/t/mail-app-design-decisions-regarding-reply-reply-all-forward-actions/23128

There is currrently no way to simply **reply** to the sender (as opposed to **reply all**) in an easy way. As opposed to other mail clients, the primary action is already preset to "reply all" rather than giving the user the choice.

This leads to the problem that if the user wants to reply to only the sender, he would need to go into the recipients line and manually mark all recipients that he doesn't want to address (if there would be a small 'x' beside each recipient where the user could at least remove a recipient via one click, it would make things easier). 
I think this behaviour (having to mark recipients manually in order to remove them / in order to just reply to the sender) is not really good from a UX perspective and also not very common or what a user would expect from a mail client (at least I didn't see it in any other mail client so far).
Even though the "reply all" might be the most probably primary action a user would take, it should not be pre-selected but giving the user a choice.

Alternatively, there should at least be an action button "reply" similar to the "forward" button, that allows the user to do a simple reply. It's an alternative, but not the best solution.

IMHO the current design seems to be good for a group messaging app, but not for a mail app.

I noticed this when I started working with the mail client now on a regular basis.

dcrobertson01:
Huh. I am trying to find out how I can reply all. All I can do is reply to sender. I need to use another mail programme to reply all. I can't even see the full list of recipients.

mathiasconradt:
@dcrobertson01 Hm, that's weird. The button is labeled "reply" but it has the function of a "reply all" at the moment (see first screenshots in my form link mentioned above). 
Sure that your email actually has multiple recipients in it when you reply? Maybe you can post a screenshot, cause there should not be a simple reply function at the moment.

dcrobertson01:
![2017-11-27-194621_1280x800_scrot](https://user-images.githubusercontent.com/283012/33254215-8672134e-d3ac-11e7-83db-11f3740c9b4b.png)

ChristophWurst:
We've discovered a few bugs recently that effected which addresses were put into the reply to/cc fields. Maybe your installed version was effected too.

freekvh:
I have the same behavior in 12.0.4. Just emailed over 100 people using just the reply button...

brzd:
Hi all

[...]

Issue #512, opened 2017-09-11
Better separation between mail and answer field


fsedarkalex:
I find it very hard to see the answer field after any email.

I think there should be some border between the actual mail and the answer field so you can easily see where the mail ends and where the reply part begins.

Furthermore I'd love to see an answer button somewhere in the header or at least fixed on the bottom of the screen so you do not need to "search it". I mean a button which will scroll down to the reply box when clicked and focus the box.

ChristophWurst:
cc @jancborchardt @skjnldsv

fsedarkalex:
Err... Noticed the button is actually there I just did not see it :ashamed:

But an idea:

What about adding a fixed footer section which will open up when used?
See attached graphic... Not thinking about reloads just an overlay...
![mailapp](https://user-images.githubusercontent.com/16449266/30271116-614bc898-96ef-11e7-8d91-94321ab31462.png)

alexanderdd:
Hm, this should actually be two separate issues.
(1) answer button - looks solved to me?!
(2) separation between received email and coposition of reply email.

I agree about (2), there should be some grey space or something.

ChristophWurst:
@jancborchardt please have a look

jancborchardt:
@ChristophWurst I still think putting an additional "Reply" button in the header is fine, as in that other pull request (where I only disagreed with removing the direct reply field). Next to it could be a 3-dot-menu with Forward, Delete etc (to also make delete work on mobile in this view).

simonbuehler:
to add to this issue:

the initial whitespace below the subject and sender down to the actual mail body is too big, the content is somehow lost on the page and there is a visual guide like a thin line missing.
Overall clean design is nice but the missing borders around header, mail, reply do not look right and are not intuitive. there is some visual ordering needed in this empty sea of whitespace!

rasos:
Several of our users get lost with the pre-filled answer section. They believe it is a duplication of the message. When we explain that you save another reload to answer then some understand the concept.

There should be at least a stronger border to the pre-filled answer section, with a headline `Answer`. The `Forward` button above should be also aligned right, like the `Send` button of the answer. 

A more advanced solution would be like https://github.com/nextcloud/mail/pull/750#issuecomment-363900595 suggests that the pre-filled answer is collapsed and opens when pressing `Answer`.

ChristophWurst:
[...]

PR #750, opened 2018-02-01
Clean up reply-all/forward handling and add reply-to-sender

ChristophWurst:
Ref https://github.com/nextcloud/mail/issues/598

TODO:
- [x] Clean up UI
   * Add reply/reply-all/forward button group to the message header
   * Fix long message subjects (fully show them)
   * Do not scroll message body under the header as that looks super confusing, esp. on mobile phones - element boundaries were not clear
   * Move floating and confusing composer send button to a fixed position
- [ ] Add simple reply
- [ ] Fix reply-all
- [ ] Fix forward message
- [ ] Test on mobile
- [ ] Fix l10n
- [ ] Squash commits

--

Current state

### DIFF
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -367,6 +367,11 @@
 .composer-fields {
 	position: relative;
 }
+.composer-bottom {
+	display: flex;
+	flex-direction: row;
+	justify-content: space-between;
+}
 input.to,
 input.cc,
 input.bcc,
@@ -442,17 +447,6 @@ label.bcc-label {
 textarea.reply {
 	min-height: 100px;
 }
-input.submit-message,
-.submit-message-wrapper {
-	position: fixed;
-	bottom: 10px;
-	right: 15px;
-}
-.submit-message-wrapper {
-	position: fixed;
-	height: 36px;
-	width: 60px;
-}
 
 .submit-message-wrapper-inside {
 	position: absolute;
@@ -466,12 +460,6 @@ input.submit-message,
 	float: left;
 }
 
-/* extra button styles */
-.submit-message.send,
-#mail_new_attachment,
-#forward-button {
-	padding: 12px;
-}
 .new-message-attachments,
 #forward-button {
 	margin-left: 30px;
@@ -557,23 +545,36 @@ input.submit-message,
 }
 
 #mail-message-header {
-	position: fixed;
-	height: 90px;
-	width: 50%;
-	z-index: 100;
-	background: -webkit-linear-gradient(rgba(255,255,255,.97), rgba(255,255,255,.9) 80%, rgba(255,255,255,0));
-	background: -o-linear-gradient(rgba(255,255,255,.97), rgba(255,255,255,.9) 80%, rgba(255,255,255,0));
-	background: -moz-linear-gradient(rgba(255,255,255,.97), rgba(255,255,255,.9) 80%, rgba(255,255,255,0));
-	background: linear-gradient(rgba(255,255,255,.97), rgba(255,255,255,.9) 80%, rgba(255,255,255,0));
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	border-bottom: 1px solid #eee;
+}
+#mail-message-header > div:first-child {
+	flex-grow: 1;
+	margin-right: 20px;
+}
+#mail-message-header > div:first-child > h2 {
+	word-wrap: break-word;
+}
+#mail-message-header > div:last-child {
+	flex-shrink: 0;
 }
 
-#mail-message-header h2,
-#mail-message-header p {
-	white-space: nowrap;
-	overflow: hidden;
-	text-overflow: ellipsis;
-	padding-bottom: 7px;
-	margin-bottom: 0;
+.button-group > button {
+	border-radius: 0px;
+	margin-right: -1px;
+	z-index: 0;
+}
+.button-group > button:hover {
+	z-index: 1;
+}
+.button-group > button:first-child {
+	border-radius: 3px 0px 0px 3px;
+}
+
+.button-group > button:last-child {
+	border-radius: 0px 3px 3px 0px;
 }
 
 #mail-message-close {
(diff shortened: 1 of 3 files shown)


ChristophWurst:
Open question: how we we handle different reply cases? If you click reply/reply-all, will that only change the prefilled values of the composer? What do we do about existing values - shall they be overwritten?

I do get @jancborchardt's arguments on why we're showing the composer right away: it shall look like a conversation. But personally, I always found the composer rather annoying as most of the time I don't use it. Moreover, all the other email clients I've checked do not show a composer right away, but only if you use a reply/forward action.

@jancborchardt @skjnldsv ideas? Would it be OK to hide the composer as default and only show/prefill it if one of the reply/reply-all/forward buttons is hit?

jancborchardt:
> how we we handle different reply cases? If you click reply/reply-all, will that only change the prefilled values of the composer? What do we do about existing values - shall they be overwritten?

Yeah, only the to and cc fields should be changed as a result of that. Forward is different as it changes the content too, so we should do it like now and put it in a new mailto

> Would it be OK to hide the composer as default and only show/prefill it if one of the reply/reply-all/forward buttons is hit?

The composer is shown by design to make it dead-simple to quickly reply to mails, especially if they are short. Reply all is the default which makes the flow so much easier, and I don’t see why to remove that. If at all, it’s a separate discussion from this.


Also additonal feedback: Let’s add the button-primary class to the »Reply to all« so it’s blue and perceived as primary. Otherwise we have 3 buttons of the exact same importance.

ChristophWurst:
> If you click reply/reply-all, will that only change the prefilled values of the composer? What do we do about existing values - shall they be overwritten?

^ @jancborchardt

jancborchardt:
Ideally we change the values but also keep anything which has been added on top of the original to or cc (or bcc)

ChristophWurst:
> Ideally we change the values but also keep anything which has been added on top of the original to or cc (or bcc)

Ideally, yes. But what shall we do now? It's quite complex to do differential updates (remove recipients when switching from reply-all to reply-sender and adding only the missing ones when switching vice versa). I don't know any immediate solution for this.
We'd also have to parse email addresses and label, which adds yet another level of complexity.

ChristophWurst:
> Also additonal feedback: Let’s add the button-primary class to the »Reply to all« so it’s blue and perceived as primary. Otherwise we have 3 buttons of the exact same importance.

IMO their importance is the same and cannot be generalized.

ChristophWurst:
Other external input on the topic: https://github.com/nextcloud/mail/issues/512#issuecomment-328493219

jancborchardt:
> IMO their importance is the same and cannot be generalized.

The default so far was reply all, and it’s clearly the usually most relevant. »Forward« is by far not as important as replying, either reply all or only replying to recipient.


> Ideally, yes. But what shall we do now?

Ok, then:
- Default prefill is reply all, so everyone is in anyway
- When switching to Reply, remove all from cc
- When switching back to Reply all, refill original cc, no need to touch »to«
- No touching of bcc, subject, or content at all
- On »Forward« we do the same as now and open a completely new composer

ChristophWurst:
> When switching to Reply, remove all from cc
> When switching back to Reply all, refill original cc, no need to touch »to«

Well, actually the `to` field is different in some contexts. Image you got a message from a mailinglist. On direct reply, the new recipient is the original sender. For lists, you reply to the list. So that cannot be ignored. Hence we have to touch it.

ChristophWurst:
My idea was to do this properly this time. I don't want to add yet another feature that somewhat works, but in reality it's more buggy than stable. If we now come up with some hacky *fingers crossed this will work in most cases* solution we have to redo it again.

ChristophWurst:
If you know one, please tell me one email client that handles different cases of replies the way you image it. I'll then model it according to that other software.

jancborchardt:
Ok, good point. Then the simplest solution is to completely overwrite to and cc on switch.

ChristophWurst:
> Ok, good point. Then the simplest solution is to completely overwrite to and cc on switch.

Well, that is actually one the hacks I really want to avoid. Switching from directy-reply to reply-all and vice versa will cause data loss/inconsistency. If a user doesn't check the to/cc fields again, they might send the reply to the wrong people.

ChristophWurst:
FYI one of the probably best designed app programs also has reply/reply-all/forward shown with equal importance. None of the buttons stands out.
bildschirmfoto von 2018-02-02 09-00-00

ChristophWurst:
Also interesting: they never directly show the reply input, not even in the thread view:
bildschirmfoto von 2018-02-02 09-04-34

ChristophWurst:
Geary, which is one of the well designed FOSS mail apps does also not show the reply composer right away and thus the UI is super simple and clean
bildschirmfoto von 2018-02-02 09-29-08

skjnldsv:
I really like the Geary UI, what is stopping us from doing something similar?

jancborchardt:
Sorry but neither macOS mail nor Geary are better than us in that respect. They show the reply, spam, mark buttons on the very top all the time. They are completely removed from the actual mail you are reading.

Again - mails are about reading and replying, both should be super simple. That's why we added the reply box by default and it's great for seamless UX.

If there's a problem with the code because of slightly related changes, we should fix that properly. Not by trying to remove a really good element simply on the basis that others don't make it as easy.

ChristophWurst:
> Sorry but neither macOS mail nor Geary are better than us in that respect. They show the reply, spam, mark buttons on the very top all the time. They are completely removed from the actual mail you are reading.

If you see email as conversation like a chat, that could make sense. But you usually reply with a new message that quotes the original one. Hence, you'll have the original message embedded in your new email. This means in our app the original message is shown twice on the same view.

I don't see the added value of showing both versions. As soon as I start replying to a message, I only look at the quoted one, regardless of whether I reply on top/bottom or inline. That's why I wouldn't care to open that message in a new, clean composer view. I'd even prefer to have it that way.

As I don't see a technical solution to adding a reply/reply-all/forward switching mechanism to the current design (see issues listed above), I'm afraid we have to further delay adding this feature until someone has found a solution. I will think about it until next week and if no better approach comes up, I'll close this PR and focus on other features.

jancborchardt:
Email is to be seen as a conversation of course.

An improvement would be to collapse the previous email quote in the compose textarea. There could be a checkbox (checked by default) saying "Include previous mail" and a button to "Reply inline" which automatically expands the quoted mail – what is currently how we do it.

Issue #560, opened 2017-10-05
Mail is not displayed, JS running ready 


foxbow:
### Steps to reproduce
Upgrade from Nextcloud 11 to 12
Upgrade mail plugin to 0.7.3
Try to check mails

### Expected behaviour
The list of folders is shown, the mail titles of the current folder are shown, the selected mail is shown

### Actual behaviour
The list of folders is shown, the mail titles of the current folder are shown, the selected mail is not shown, just the roller while one core stays busy. This blocks the whole app on Firefox. On Chrome you can still change folders and on some the mails are displayed just fine. I was not able to find a true pattern here. It has apparently nothing to do with number of mails in the folder, mail formatting (html/text/mixed) or sender. The Chrome javascript console however claims some cache error.

### General server configuration

**Operating system:** Linux J6868 2.6.32-5-xen-686 #1 SMP Sun Jul 13 02:47:03 UTC 2014 i686

**Web server:** nginx/1.6.2 (fpm-fcgi)

**Database:** mysql 5.5.57

**PHP version:** 5.6.31-1~dotdeb+7.1
<details>
        <summary>PHP-modules loaded</summary>

```
 - Core
 - date
 - ereg
 - libxml
 - openssl
 - pcre
 - zlib
 - bcmath
 - bz2
 - calendar
 - ctype
 - dba
 - dom
 - hash
 - fileinfo
 - filter
 - ftp
 - gettext
 - SPL
 - iconv
 - json
 - mbstring
 - session
 - posix
 - Reflection
 - standard
 - shmop
 - SimpleXML
 - soap
 - sockets
 - Phar
 - exif
 - sysvmsg
 - sysvsem
 - sysvshm
 - tokenizer
 - wddx
 - xml
 - xmlreader
 - xmlwriter
 - zip
 - cgi-fcgi
 - PDO
 - curl
 - gd
 - memcache
 - mysql
 - mysqli
 - pdo_mysql
 - pdo_sqlite
 - redis
 - sqlite3
 - xsl
 - mhash
 - Zend OPcache
```
</details>

### Nextcloud configuration

**Nextcloud version:** 12.0.3 - 12.0.3.3

**Updated from an older Nextcloud/ownCloud or fresh install:** 11.0.x

**Where did you install Nextcloud from:**  web-installer

**Are you using external storage, if yes which one:** files_external is disabled

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** n/a

```
 - activity: 2.5.2
[...]

PR #752, opened 2018-02-01
Remove session storage cache

ChristophWurst:
* Only used for HTML bodies, which are cached on HTTP level anyway
* Preloading of messages was broken and has to be re-evaluated
* Made the app unusable with large mailboxes - https://github.com/nextcloud/mail/issues/560

Fixes https://github.com/nextcloud/mail/issues/560

Even if it sounds like it, this should not have any effect on performance. Trust me 😉 

cc @foxbow


ChristophWurst:
> Ship it.

😀 🦄 :shipit:

PR #753, opened 2018-02-01
Remove unified inbox/mailbox relicts

ChristophWurst:
Working on https://github.com/nextcloud/mail/pull/752 made me realize that we actually have no more code paths that use those classes. Thus I'm removing `UnifiedInbox` and `UnifiedMailbox`. As those classes seem to not have been covered with tests, this will bump our test coverage.

To further clean up the code I've removed the `IAccount` interface. I thought this was a clever abstraction, but in fact many code paths assumed to get a `Account` as some methods were not part of the interface. Anyway, now that it's only one impl, the removal of the interface is the natural consequence.


ChristophWurst:
bildschirmfoto von 2018-02-01 17-20-34
✨

PR #777, opened 2018-02-12
Add avatar URL memoization to prevent duplicate requests

ChristophWurst:
* Extract external avatar logic into a service
* Add tests for the new service
* Memoize (cache) the return value of avatar URL calls and thus eliminate duplicate requests


ChristophWurst:
> Memoize (cache) the return value of avatar URL calls and thus eliminate duplicate requests

On my test system this effectively cut page load from 140 requests down to 120 -> ~1/6 fewer requests.

PR #782, opened 2018-02-15
Fix INBOX prefixing

ChristophWurst:
TODO
* [x] Remove prefix from *translated* folder names
* [x] Don't set prefixed folders as subfolders of inbox

Tested with
* Unit tests (see changes)
* Manually with a community provided mail server

Fixes https://github.com/nextcloud/mail/issues/386

### DIFF
--- a/doc/admin.md
+++ b/doc/admin.md
@@ -81,20 +81,3 @@ If you can not access your Gmail account use https://accounts.google.com/Display
 ### Outlook.com
 
 If you can not access your Outlook.com account try to enable the 'Two-Factor Verification' (https://account.live.com/proofs/Manage) and set up an app password (https://account.live.com/proofs/AppPassword), which you then use for the Nextcloud Mail app.
-
-### Dovecot IMAP
-
-If your Dovecot IMAP server prefixes all folders with `INBOX`, Nextcloud Mail does not work correctly.
-
-Check `/etc/dovecot/dovecot.conf`:
-
-```
-namespace inbox {
-        separator = .
-        # All folders prefixed
-        # prefix = INBOX.
-        prefix =
-        inbox = yes
-        type = private
-}
-```
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -79,16 +79,17 @@ class FolderMapper {
 
 	/**
 	 * @param Folder[] $folders
+	 * @param bool $havePrefix
 	 * @return Folder[]
 	 */
-	public function buildFolderHierarchy(array $folders) {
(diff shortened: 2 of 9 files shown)


ChristophWurst:
Test package:
[mail.tar.gz](https://github.com/nextcloud/mail/files/1727071/mail.tar.gz)

cc @reubenlillie @japerezULT @alexhass @douzepouze @ftoledo @andreas-wolf @winnieXY @Ninos please test the package above with your dovecot setup and tell me if the fix is working for you.

Special thanks to @ftoledo for providing a test account on their server - very much appreciated 🚀

Ninos:
Thank you for your work. For me I still cannot see the inbox subfolders :/

ChristophWurst:
@Ninos please provide exact steps to reproduce. How did you install the provided archive? Could you send me credentials to a IMAP account on your email server?

Ninos:
@ChristophWurst Sent via mail. I've already created some test folders.

Ninos:
Problem found. Subfolders are working fine with latest version, just UX is very bad, because you need to click on the icon of the inbox/draft/trash-folder. I recommend moving the arrow to the right side without hiding it, just if folder has no subfolders.

ChristophWurst:
This really sounds like you're testing with a wrong version of the app. All the above issues were addressed in this PR. But I'll test with the provided account next week. Thanks for the credentials.

Ninos:
Hmm, strange. Normally I've used the archive uploaded in this PR..

ChristophWurst:
@Ninos you don't have consistent `INBOX.` prefixing. Your mailboxes are prefixed with different tokens like `INBOX.`, `Draft.` or `Sent.`. Not why your server is configured like that but this is not what this PR attempts to fix. This PR is only about the `INBOX.*` prefix on Dovecot mailservers.

ChristophWurst:
Four days of no feedback from @reubenlillie @japerezULT @alexhass @douzepouze @ftoledo @andreas-wolf and @winnieXY and @Ninos's problem is caused by a specific setup, hence I'm merging this now since it at least fixed the problem with the one test account that was provided.

Not that since I have no possibility to reproduce those prefixed setups anymore once @ftoledo disables my temporary test account, I cannot guarantee this will work forever. As soon as we've got a running Docker image with this kind of IMAP account, we can of course add automated tests.

Ninos:
@ChristophWurst I'm using a default ispconfig & dovecot (Debian Stretch) installation, so this must be a general problem. If needed I can create a seperate issue.

ChristophWurst:
> If needed I can create a seperate issue.

Yes, please.

PR #788, opened 2018-02-19
Generate a automatically prefilled GitHub issue templates for errors in debug mode

ChristophWurst:
Automatically generate a crash report page on the client-side for
API errors that return tracing information like exception class,
error message and stack trace.

The idea is simple: most bug reports we get are of low quality. Most of them are missing logs, especially the IMAP ones which help trace down the majority of bugs. A crash report page like this would be an attempt to help people gather the necessary information to fill out the issue template more easily.

@nextcloud/mail what do you think? Would that make sense? Will people submit better bug reports?

~~Currently, the page looks like this:~~ this was the initial version, see new screenshot below.
bildschirmfoto von 2018-02-19 11-45-35
bildschirmfoto von 2018-02-19 11-45-57

I'd propose to also automatically collect the IMAP and SMTP logs if we go down this path. This has the benefit that those logs are scoped to the request where the error occurred. Currently you have to manually separate which logs belong to which request and thus makes it more time consuming to find errors in the IMAP logs.

We could even go one step further and pre-fill an issue template like @juliushaertl's app does, but on the client-side.

I've left the testing error in the folder controller so anybody can check out this branch and test the crasg report.

Ref https://github.com/orgs/nextcloud/projects/4#card-7365062


ChristophWurst:
Auto generated issue template:
bildschirmfoto von 2018-02-19 21-22-11
Of course we can add as much information as we'd like. I've greatly reduced the number of required sections because most of them are not relevant for the majority of bug reports and we always ask for more information.

I've created this tiny npm lib to build Nextcloud issue templates: https://www.npmjs.com/package/nextcloud_issuetemplate_builder https://github.com/ChristophWurst/nextcloud_issuetemplate_builder

enoch85:
Nice stuff!

jancborchardt:
Would appreciate that! :D makes it much easier to put the relevant info in the issues and not collect it all manually.

juliushaertl:
> We could even go one step further and pre-fill an issue template like @juliushaertl's app does, but on the client-side.

Maybe we can just prefill the logs section of the issue template and open up the issuetemplate app for finishing the report. I'll think a bit more about that.

ChristophWurst:
# Sample PHP exception report:
bildschirmfoto von 2018-03-12 10-57-31

# Sample JS error report:
bildschirmfoto von 2018-03-12 11-04-43

There is a lot we can improve on both, but this is the basic feature that I'd like to integrate.

ChristophWurst:
Oh, and it now directly opens the issue page on GitHub rather than copying the issue template to the clipboard. I assume that makes it even easier to report a bug.

pierreozoux:
Did you investigate https://sentry.io/welcome/ ? It is a sliently different solution to the problem to track errors, but I think it is also worth exploring?

ChristophWurst:
> Did you investigate https://sentry.io/welcome/ ? It is a sliently different solution to the problem to track errors, but I think it is also worth exploring?

https://github.com/christophwurst/nextcloud_sentry https://apps.nextcloud.com/apps/sentry :wink:

ChristophWurst:
@pierreozoux the crash reporting added by this PR focuses on user-faced bugs. The sentry integration does all the automated bug reporting of logged exceptions. They are similar but slightly different. Does that make sense?

PR #789, opened 2018-02-19
Migrate from $.ajax to native fetch API

ChristophWurst:
- [ ] Find an ergonomic way to send JSON (including necessary headers).
- [ ] Find an ergonomic way to retrieve JSON (including accept headers).


ChristophWurst:
Postponing because it's not really necessary to switch right now.

PR #796, opened 2018-02-21
fix typo

myrho:


ChristophWurst:
Hi and thanks a lot for your contribution!

We don't translate this app directly via GitHub but instead use [Transifex](https://www.transifex.com/signin/?next=/nextcloud/) for that. So please take a look at the platform and fix the typo there. The changes will be synced back after into this repository later.

Thanks :wink:

PR #801, opened 2018-02-23
Fix stretching of non-square avatars

### LINKED ISSUE #26: Non-square avatars are stretched
### Steps to reproduce
1. Add a contact with email address and avatar that is not square (contacts app)
2. Open mail-folder with email from this contact
### Expected behaviour

The avatar should be fit into the circle.
### Actual behaviour

The avatar is stretched to the size of the circle.
### Mail app

**Mail app version:** 0.5.4

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37492311-non-square-avatars-are-stretched?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sagorika1996:
Fixes #26

### DIFF
--- a/js/templates/message-list-item.html
+++ b/js/templates/message-list-item.html
@@ -5,7 +5,7 @@
 	<div class="mail-message-header">
 		<div class="sender-image avatardiv">
 			{{#if senderImage}}
-			<img class="avatar" src="{{senderImage}}" width="32px" height="32px" />
+			<img src="{{senderImage}}" width="32px" height="32px" style="object-fit: cover;"/>
 			{{else}}
 			<div class="avatar" data-user="{{sender.label}}" data-size="32"></div>
 			{{/if}}
--- a/js/templates/message-list-item.html
+++ b/js/templates/message-list-item.html
@@ -5,7 +5,7 @@
 	<div class="mail-message-header">
 		<div class="sender-image avatardiv">
 			{{#if senderImage}}
-			<img src="{{senderImage}}" width="32px" height="32px" />
+			<img class="avatar" src="{{senderImage}}" width="32px" height="32px" />
 			{{else}}
 			<div class="avatar" data-user="{{sender.label}}" data-size="32"></div>
 			{{/if}}
--- a/js/templates/message-list-item.html
+++ b/js/templates/message-list-item.html
@@ -5,7 +5,7 @@
 	<div class="mail-message-header">
 		<div class="sender-image avatardiv">
 			{{#if senderImage}}
-			<img src="{{senderImage}}" width="32px" height="32px" style="object-fit: cover;"/>
+			<img src="{{senderImage}}" width="32px" height="32px" />
 			{{else}}
 			<div class="avatar" data-user="{{sender.label}}" data-size="32"></div>
 			{{/if}}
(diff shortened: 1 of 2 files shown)


juliushaertl:
@ChristophWurst Does the mail app also support IE11? In that case we cannot use object-fit according to: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit#Browser_compatibility

ChristophWurst:
> @ChristophWurst Does the mail app also support IE11? In that case we cannot use object-fit according to: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit#Browser_compatibility

I don't think so. At least I've never paid any attention to IE compatibility. cc @nextcloud/mail

sagorika1996:
@juliushaertl object-fit is also used in Contacts app

js/templates/message-list-item.html
```diff
@@ -5,7 +5,7 @@
 	<div class="mail-message-header">
 		<div class="sender-image avatardiv">
 			{{#if senderImage}}
-			<img src="{{senderImage}}" width="32px" height="32px" />
+			<img class="avatar" src="{{senderImage}}" width="32px" height="32px" />
```
juliushaertl: It would probably be better to not use .avatar here, since that will cause the avatar jquery plugin to jump in on that img-tag as well.

js/templates/message-list-item.html
```diff
@@ -5,7 +5,7 @@
 	<div class="mail-message-header">
 		<div class="sender-image avatardiv">
 			{{#if senderImage}}
-			<img src="{{senderImage}}" width="32px" height="32px" />
+			<img src="{{senderImage}}" width="32px" height="32px" style="object-fit: cover;"/>
```
ChristophWurst: @sagorika1996 I think you might want to adjust the SCSS rather than inlining this style in the HTML :wink: 

A selector like `.sender-image img` should work for this specific avatar image. The `object-fit` from `.avatar` could be removed then, I assume.

sagorika1996: @ChristophWurst Sender's pic in the autocomplete suggestions is also stretched https://github.com/nextcloud/mail/blob/f2ecf4d4ddd011475f991509d9f45af6d0c12607/js/views/composerview.js#L522-L524
That's why I included `object-fit` in `.avatar`

MariusBluem:
> I don't think so. At least I've never paid any attention to IE compatibility. cc @nextcloud/mail

Is supported by all current major browsers like Edge, Chrome, Firefox and Safari. Should be fine ... using Contacts, Calendar, Mail or Talk with IE is harakiri anyway 😈

ChristophWurst:
@MariusBluem did you test this PR? Does it work for you?

MariusBluem:
> @MariusBluem did you test this PR? Does it work for you?

Did not experienced new bugs 😅  ... but I was not able to reproduce the one this PR wants to fix.

Issue #41, opened 2016-09-05
Sending to groups [$45]


jancborchardt:
Typing a group name into the »to« field should autocomplete from the »Groups« in the Contacts app as well. This would allow you to write emails to your »Sports« or »Work« groups quickly.

cc @aspdye @Toub @virgilekat @ttjaden since you requested that. :)

ttjaden:
Thanks @jancborchardt for this news. Maybe you have a little extra-information for me, 'cause I'm confused about the owncloud/nextcloud thing:
1. Will the Mail-App be the same for owncloud and nextcloud?
2. I tested the feature, but in mail 0.5.3 the autocomplete doenst work with the groups from the contacts apps.
Would be great if you guys really can make this work! :-)

jancborchardt:
@ttjaden the main contributors to the Mail app (@ChristophWurst is the main maintainer) moved to Nextcloud as the more community-oriented project. Hence we don’t really test on ownCloud anymore and while the app might run there it’s not tested. So we decided to also move/fork the repository.
Same happened btw to the News and Gallery apps as the maintainers moved. And likely the Calendar is going to move soon too.

Regarding the feature, it’s not there yet. This is just an issue tracking the request. :) Also this is a community project so any contribution is welcome! :rocket:

Ecphrasis:
Any news of this missing feature ? ;) Actually, rainloop and Mail don't allow to send mails to a group and it's a very missing feature...

lafrech:
I'm looking for a collaborative work platform and I'm thinking of using NextCloud. AFAIK, there is no support for mailing lists, already.

The "send to group" feature could provide that if it actually sent to `sports@groups.my_nextcloud.tld` rather than explicitly expanding to the group members.

For a complete mailing list support, it should also be able to receive e-mails to `sports@groups.my_nextcloud.tld` from the outside and redirect them.

I'm not sure what the best strategy would be. Maybe mailing lists should be a separate Nextcould app. And maybe it would be better to interface an existing solution like Mailman than reinventing the wheel.

Sorry if I'm off-topic. I thought this feature request might arise at some point, so rather mention it before the "sending to groups" feature reaches an advanced conception stage.

jancborchardt:
So @Ivansss @nickvergessen @schiessle @LukasReschke who of you implemented this in the Spreed video calls app? :) Basically here it would only be:
- adding the contact groups to the autocomplete dropdown
[...]

PR #800, opened 2018-02-23
Autocomplete nextcloud groups #41

myrho:
Autocomplete nextcloud groups in recipient fields

PR #810, opened 2018-02-27
Sending to nextcloud groups

myrho:
This PR overrides #800.

* Autocomplete nextcloud groups in recipient input fields. 
* Prefix nextcloud group recipients with "nextcloud:" to have a means of distinguishing from yet-to-implement Contacts groups recipients (#41).
* Expands groups to email addresses of group members.

Fixes https://github.com/nextcloud/mail/issues/41


lib/AppInfo/Application.php
```diff
@@ -71,6 +73,10 @@ private function initializeAppContainer() {
 
 		$container->registerAlias('ErrorMiddleware', ErrorMiddleware::class);
 		$container->registerMiddleWare('ErrorMiddleware');
+
+		$container->registerService('OCA\Mail\Service\GroupsIntegration', function($c) {
```
ChristophWurst: I don't think you have to register this here. The DI container should be able to resolve this class automatically.

lib/Service/AutoCompletion/AutoCompleteService.php
```diff
@@ -23,22 +24,28 @@
...
-	public function __construct(ContactsIntegration $ci, AddressCollector $ac) {
+	public function __construct(ContactsIntegration $ci, GroupsIntegration $gi, AddressCollector $ac) {
 		$this->contactsIntegration = $ci;
+		$this->groupsIntegration = $gi;
 		$this->addressCollector = $ac;
 	}
 
 	public function findMatches($term) {
 		$recipientsFromContacts = $this->contactsIntegration->getMatchingRecipient($term);
+		$recipientGroups = $this->groupsIntegration->getMatchingGroups($term);
```
ChristophWurst: I'd prefer to have this logic moved to a service class, maybe even introduce a facade for that. But that's a minor enhancement - this can be done later on :wink:

lib/Service/Group/AbstractGroupService.php
```diff
@@ -0,0 +1,51 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Mail\Service\Group;
+
+abstract class AbstractGroupService {
```
ChristophWurst: Looks like an interface would be the better choice here.

lib/Service/Group/NextcloudGroupService.php
```diff
@@ -0,0 +1,80 @@
...
+	public function getNamespace() {
+		return $this->namespace;
+	}
+
+	public function search($term) {
+		$groups = $this->groupManager->search($term);
+
+		return array_map(
+			function($g) {
+				return [
```
ChristophWurst: Looks like a good candidate to implement PHP's `JsonSerializable` interface for this.

lib/Service/Group/NextcloudGroupService.php
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	public function getUsers($groupId) {
+		if(!$this->groupManager->groupExists($groupId)) {
+			throw new \Exception("$groupId ({$this->getNamespace()}) does not exist");
+		}
+		$users = $this->groupManager->get($groupId)->getUsers();
+		return array_map(
+			function($user) {
+				return [
```
ChristophWurst: same here

lib/Service/GroupsIntegration.php
```diff
@@ -0,0 +1,108 @@
...
+class GroupsIntegration {
+
+	/**
+	 * The services to get groups from
+	 *
+	 * @var IGroupService[]
+	 */
+	private $groupServices = [];
+
+	public function __construct(AbstractGroupService ...$groupServices) {
```
ChristophWurst: For simplicity, I'd recommend to only inject a single group service (there's only one so far). You don't have to use the concrete impl here but use the interface (as proposed above) and register the impl as alias for the interface in `Application`.

lib/Service/GroupsIntegration.php
```diff
@@ -0,0 +1,108 @@
...
+namespace OCA\Mail\Service;
+
+use OCA\Mail\Service\Group\AbstractGroupService;
+
+class GroupsIntegration {
+
+	/**
+	 * The services to get groups from
+	 *
+	 * @var IGroupService[]
```
ChristophWurst: Does this interface really exist?

tests/Service/GroupsIntegrationTest.php
```diff
@@ -0,0 +1,192 @@
...
+class GroupsIntegrationTest extends TestCase {
+
+	private $groupService1;
+	private $groupService2;
+	private $groupsIntegration;
+
+	protected function setUp() {
+		parent::setUp();
+
+		$this->groupService1 = $this->getMockBuilder(NextcloudGroupService::class)
```
ChristophWurst: use `$this->createMock(...::class);`.

lib/Service/Group/AbstractGroupService.php
```diff
@@ -0,0 +1,51 @@
...
+	 * @return string
+	 */
+	abstract public function search($term);
+
+	/**
+	 * Get the group's namespace.
+	 *
+	 * @return string
+	 */
+	abstract public function getNamespace();
```
ChristophWurst: Do we really need these namespaces? This kinda looks like an unnecessary generalization.

lib/AppInfo/Application.php
myrho: How should the container know which `AbstractGroupService`s `GroupsIntegration` should be constructed with?

lib/Service/Group/AbstractGroupService.php
myrho: Yes!

lib/Service/Group/NextcloudGroupService.php
myrho: I ported this pattern from `Contacts\IManager::search`. It's just returning key/value pairs. Where would `JsonSerializable` fit in?

lib/AppInfo/Application.php
ChristophWurst: See below. Use an interface instead of this abstract class and register the implementation as alias of the interface here (see other services in this file).

lib/Service/Group/NextcloudGroupService.php
ChristophWurst: Nvm, just realized that `$g` is some data structure from the Nextcloud server, so nothing we can change. In that case that's fine :+1:

lib/Service/GroupsIntegration.php
myrho: I designed this so it will be extensible for Contacts groups. 
But I could reserve this feature for later when Contacts groups integration will be ready.

lib/Service/GroupsIntegration.php
myrho: It did :-)
Missed to update the docs here. 

But `AbstractGroupService` should become `IGroupService` again anyways.

lib/Service/Group/AbstractGroupService.php
myrho: Yes, I think so. There are two reasons:
1. When `GroupIntegration` once would be extended for Contacts groups, we need a means to tell apart whether it's a Nextcloud or a Contacts group, since both applications can have equally named groups. Eg. nextcloud:mygroup vs. contacts:mygroup.
2. We need to distinguish nextcloud groups from single recipients, since Nextcloud groups could look like email addresses. Eg. nextcloud:group@domain.com vs. group@domain.com ...

tests/Controller/AccountsControllerTest.php
```diff
@@ -94,6 +98,7 @@ protected function setUp() {
 		$this->appName = 'mail';
 		$this->request = $this->createMock(IRequest::class);
 		$this->accountService = $this->createMock(AccountService::class);
+		$this->groupsIntegration = new GroupsIntegration();
```
ChristophWurst: use a mock instead please

lib/AppInfo/Application.php
ChristophWurst: Please use `$container->registerAlias(IGroupService::class, NextcloudGroupService::class);` instead

tests/Service/Group/NextcloudGroupServiceTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+class NextcloudGroupServiceTest extends TestCase {
+
+	private $groupsManager;
+	private $groupService;
+
+	protected function setUp() {
+		parent::setUp();
+
+		$this->groupsManager = $this->createMock('OCP\IGroupManager');
```
ChristophWurst: use the `::class` notation

lib/Service/Group/NextcloudGroupService.php
```diff
@@ -0,0 +1,80 @@
...
+					'name' => $g->getDisplayName()
+				];
+			},
+			$groups
+		);
+	}
+
+	public function getUsers($groupId) {
+		if(!$this->groupManager->groupExists($groupId)) {
+			throw new \Exception("$groupId ({$this->getNamespace()}) does not exist");
```
ChristophWurst: throw a `OCA\Mail\Exception\ServiceException` instead

lib/Service/GroupsIntegration.php
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	 * Returns the prefix for the group service.
+	 *
+	 * @param IGroupService $gs
+	 * @return string
+	 */
+	public function servicePrefix(IGroupService $gs) {
+		if(empty($gs->getNamespace())) {
+			throw new \Exception('GroupService has no namespace');
```
ChristophWurst: use `ServiceException` instead

tests/Service/Autocompletion/AutoCompleteServiceTest.php
```diff
@@ -37,11 +39,15 @@ protected function setUp() {
 		$this->contactsIntegration = $this->getMockBuilder('\OCA\Mail\Service\ContactsIntegration')
 			->disableOriginalConstructor()
 			->getMock();
+		$this->groupsIntegration = $this->getMockBuilder('\OCA\Mail\Service\GroupsIntegration')
```
ChristophWurst: `->createMock` and the `::class` notation

tests/Service/Group/NextcloudGroupServiceTest.php
```diff
@@ -0,0 +1,142 @@
...
+				'email' => 'alice@smith.net'
+			]
+		];
+
+		$this->assertEquals($expected, $actual);
+
+	}
+
+	public function testGetUsersWrong() {
+		$this->expectException(\Exception::class);
```
ChristophWurst: needs update to `ServiceException`

lib/Service/GroupsIntegration.php
```diff
@@ -0,0 +1,108 @@
...
+class GroupsIntegration {
+
+	/**
+	 * The services to get groups from
+	 *
+	 * @var IGroupService[]
+	 */
+	private $groupServices = [];
+
+	public function __construct(IGroupService ...$groupServices) {
```
ChristophWurst: I'd prefer to simply inject a single instance as long as we have only a single implementation. Other implementations can be injected later.

tests/Service/GroupsIntegrationTest.php
```diff
@@ -0,0 +1,228 @@
...
+		$expected = "john@doe.com,bob@smith.net,mary@smith.net,alice@smith.net,jim@smith.net";
+
+		$actual = $this->groupsIntegration->expand($recipients);
+
+		$this->assertEquals($expected, $actual);
+
+	}
+
+	public function testExpandEmpty() {
+		$this->expectException(\Exception::class);
```
ChristophWurst: `ServiceException`

tests/Service/GroupsIntegrationTest.php
```diff
@@ -0,0 +1,228 @@
...
+		$expected = "john@doe.com,nons:testgroup,alice@smith.net";
+
+		$actual = $this->groupsIntegration->expand($recipients);
+
+		$this->assertEquals($expected, $actual);
+
+	}
+
+	public function testExpandWrong2() {
+		$this->expectException(\Exception::class);
```
ChristophWurst: `ServiceException`

lib/Service/Group/IGroupService.php
```diff
@@ -0,0 +1,51 @@
...
+	 * Get the group's users.
+	 *
+	 * @param string $groupId 
+	 * @return array with group's key-value pairs
+	 */
+	public function getUsers($groupId);
+
+}
+
+?>
```
ChristophWurst: - [ ] remove closing tag

ChristophWurst:
> Scrutinizer — 3 new issues

Should be addressed before merging.

ChristophWurst:
cc @nextcloud/mail for review. Would be great to have a second person testing/reviewing this new feature

ChristophWurst:
> cc @nextcloud/mail for review. Would be great to have a second person testing/reviewing this new feature

*ain't nobody got time for that*.

Merging.

ChristophWurst:
@myrho please don't forget to claim your well-deserved bounty for fixing https://github.com/nextcloud/mail/issues/41 :moneybag:

alexanderdd:
I am confused.

Please see my comment in https://github.com/nextcloud/mail/issues/41

PR #826, opened 2018-03-12
Add Babel

ChristophWurst:
Lets us use ES6 features on all browsers and especially on phantomjs, required for https://github.com/nextcloud/mail/pull/824.


ChristophWurst:
CI happy and app still works -> merging to be able to continue #824

PR #857, opened 2018-03-28
Remove duplicated function

ChristophWurst:


ChristophWurst:
Thanks for the quick review. This is tagged for the next release because I don't want to merge this kind of PR shortly before a release. Will revert this.

ChristophWurst:
Ugh, I think I'll just leave it as it is now and hope that nothing breaks ¯\\_(ツ)_/¯

PR #908, opened 2018-05-14
Add issue templates

ChristophWurst:
These are based on the old issue template, which I generalized as https://github.com/ChristophWurst/nextcloud_issue_templates which is based on https://github.com/doctrine/doctrine2/tree/master/.github/ISSUE_TEMPLATE :see_no_evil: 

See the template selection in action: https://github.com/nextcloud/twofactor_u2f/issues/new/choose

cc @nextcloud/mail @skjnldsv

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,41 +1,39 @@
-### Steps to reproduce
-1.
-2.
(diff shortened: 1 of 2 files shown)


jancborchardt:
We should say "Issue" instead of "Bug" to make it more understandable, otherwise :+1:! :)

PR #914, opened 2018-05-22
Fix return type comment

ChristophWurst:


ChristophWurst:
Smh Scrutinizer doesn't like this change and I think it's right.

Issue #532, opened 2017-09-15
Link with # is cut off at that point


jancborchardt:
### Steps to reproduce
1. Have an email with a link containing a `#`, like https://docs.nextcloud.com/server/12/admin_manual/installation/source_installation.html#example-installation-on-ubuntu-16-04-lts-server
2. Click on the link

### Expected behaviour
It correctly leads to docs.nextcloud.com/server/12/admin_manual/installation/source_installation.html **#example-installation-on-ubuntu-16-04-lts-server**

### Actual behaviour
It will cut off at the #, leading to https://docs.nextcloud.com/server/12/admin_manual/installation/source_installation.html

(This breaks some links pretty hard, for example because https://zammad.org uses the hashes to link to specific tickets.)

blizzz:
Perhaps related: also parameter indicators (`?`) are ignored, rendering e.g. confirmation links broken.

ChristophWurst:
> Perhaps related: also parameter indicators (?) are ignored, rendering e.g. confirmation links broken.

Yup, super annoying. It's some regex magic I couldn't wrap my head around IIRC, thus it's still not fixed. Sounds like what was reported in https://github.com/nextcloud/mail/issues/147.

blizzz:
> Sounds like what was reported in #147.

Paintext :D Hellos from Mr. Freud.

@ChristophWurst i did a regex for links in comments also a little while back. Also, a pain, i agree. Perhaps it does help you though? cf. https://github.com/nextcloud/server/pull/7680/commits/282567fcedf34d2b7043f90cd5b7ce18d4753720#diff-8d84b744472f51309f72bc30ae82b1b7R24

PR #915, opened 2018-05-23
preserve the uri hash when sanitizing html links

icewind1991:
Fixes https://github.com/nextcloud/mail/issues/532


icewind1991:
@ChristophWurst *poke*

PR #917, opened 2018-05-28
Add account export command

ChristophWurst:
```
$ php occ mail:account:export user123
Account 17:
- E-Mail: user123@domain.com
- Name: user123
- IMAP user: user123@domain.com
- IMAP host: mail.domain.com:143, security: tls
- SMTP user: user123@domain.com
- SMTP host: mail.domain.com:25, security: none
```

cc @jospoortvliet


ChristophWurst:
Command works, tests pass locally :+1:

PR #922, opened 2018-05-29
paginate by datetime instead of just date

icewind1991:
Currently pagination only uses the date of the "cursor", not the time, because of this it's very likely to skip over messages.

This switches the horde/imap instead to a fork, once the [upstream PR](https://github.com/horde/Imap_Client/pull/3) is reviewed and released it can be switched back to using an official release

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,10 @@
 		{
 			"type": "pear",
 			"url": "https://pear.horde.org"
+		},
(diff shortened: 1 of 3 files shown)


ChristophWurst:
Wow, this is an impressive bug and fix. Thanks a lot! Will review this soon.

Issue #147, opened 2016-10-12
E-Mail with paintext and http links shows up with invalid http links


pamapa:
### Steps to reproduce
1. Create E-Mail with plain text and add a http link "https://domain.com/track/direct.php?Nr=1020&EP=100"
2. Watch E-Mail with Nextcloud Mail
### Expected behaviour

``` html
<a href="https://domain.com/track/direct.php?Nr=1020&EPLZ=8006" rel="noreferrer" target="_blank">https://domain.com/track/direct.php?Nr=1020&EPLZ=8006</a>
```
### Actual behaviour

Looks like the code, which creates the link stops at the "&", but it should not.

``` html
<a href="https://domain.com/track/direct.php?Nr=1020" rel="noreferrer" target="_blank">https://domain.com/track/direct.php</a>&EPLZ=8006
```
### Mail app

**Mail app version:** 0.5.3
### Server configuration

**Operating system**: webhoster

**Web server:** apache

**Database:** MySQL

**PHP version:** 5.6

**Version:** 10.0.1

**Updated from an older version or fresh install:** updated from Nextcloud 9.x

**List of activated apps:**
#### Client configuration

**Browser:** Firefox 45.4.0

**Operating system:** Linux

ChristophWurst:
cc @nickvergessen 😉

Maulwurf16:
ah, and by handling urls correctly, don't forget ports :)

an url http://url.de:12345/index.php is converted into: http://url.de/index.php

FlorianFranzen:
I am not sure why part of the link is hidden at all. If I view a plain text mail, I want to see all of the plain text. That is one of the main reasons to use plain text mail.

Based on this I think two things should be fixed:
 * Fix Link detection by including & and : in regexp
 * Display full link even when link is autodetected

ChristophWurst:
Fixed by https://github.com/nextcloud/mail/pull/923

PR #923, opened 2018-05-29
let urllinker handle html escaping for plain text mails instead of doing it beforehand

icewind1991:
Currently links inside plain text mails containing `&` breaks because it get's escaped into `&amp;` before running the url linker.

The url linker's escape option handles this use case correctly.

As an example, see the spreed invite mail


ChristophWurst:
There's a ticket for this bug. Will try to find and link it.

Thanks a lot!

ChristophWurst:
Tests have to be updated to this change

```
There was 1 failure:
1) OCA\Mail\Tests\Service\HtmlTest::testLinkDetection with data set #1 ('&lt;&gt;', '&lt;&gt;')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-&lt;&gt;
+&amp;lt;&amp;gt;
/home/travis/build/nextcloud/core/apps/mail/tests/Service/HtmlTest.php:41
FAILURES!
Tests: 365, Assertions: 829, Failures: 1.
```

While at it, could you possibly add a test (if none exists) that covers the issue your PR fixes? That would be awesome.

icewind1991:
done

ChristophWurst:
@icewind1991 does this fix https://github.com/nextcloud/mail/issues/147

FlorianFranzen:
@icewind1991 Thank you for your work!

Issue #901, opened 2018-05-09
IMAP folders appear in random order


tenbob:
### Steps to reproduce
Connect to an IMAP account which has several folders.

### Expected behaviour
Folders should display in alphabetical order.

### Actual behaviour
Folders are displayed in a completely random order. I can't find a setting to change this. I know folder maintenace is scheduled as a future enhancement. But displaying the folders in a sensible order would be a quick win. 

### Mail app
0.7.10

### Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
imap.zoho.com:993

ChristophWurst:
The Nextcloud mail app offers an extensive logging system to make it easier identifying and tracking down bugs. Please enable [debug mode](https://docs.nextcloud.com/server/13/developer_manual/general/devenv.html#enabling-debug-mode) and set the log [level to debug](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/logging_configuration.html) in your admin settings. Then, try to reproduce your issue and take another look at `data/nextcloud.log`, `data/horde_imap.log` and `data/horde_smtp.log`.

ChristophWurst:
They *are* sorted. Please provide the sections of your `horde_imap.log` file where the server lists your mailboxes.

tenbob:
@ChristophWurst. I turned on debug and checked the logs, but there wasn't anything significant. Just these two lines.
~~~
Debug	mail	loading messages of folder <SU5CT1g=>
Debug	mail	no default config found
~~~
Here is a screenshot of my IMAP folders. I have a lot of folders and I have added numbers to the names so they sort the way I want. This works OK in other clients, But not in Nextcloud.

![clipboard06](https://user-images.githubusercontent.com/10346301/40250872-9159fee8-5acf-11e8-9767-f47e7196c360.jpg)

ChristophWurst:
Are these top-level folders or subfolders?

tenbob:
These are top level folders at imap.zoho.com.

Sorry. I can't find the instructions for horde logging. Can you please point me to a description.

ChristophWurst:
> Sorry. I can't find the instructions for horde logging. Can you please point me to a description.

https://github.com/nextcloud/mail/issues/901#issuecomment-387797578

tenbob:
Oops, thanks. I have now located the horde log and done some more tests.

[...]

PR #929, opened 2018-06-05
Fix folder translation and sorting order

ChristophWurst:
The `sortFolder` method expects the folders to have a `displayName` property set, which is populated on `translateAll`. Therefore, the order was wrong. Folders should first be translated and then sorted.

Fixes https://github.com/nextcloud/mail/issues/901

Special thanks to @tenbob for some awesome debugging at https://github.com/nextcloud/mail/issues/901#issuecomment-393751571 - this helped a lot :rocket:

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -33,7 +33,7 @@ class FolderMapper {
 	 * @param Account $account
 	 * @param Horde_Imap_Client_Socket $client
 	 * @param string $pattern
-	 * @return Folder
+	 * @return Folder[]
 	 */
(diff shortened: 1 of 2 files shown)


mathroc:
thank you both !

ChristophWurst:
@nextcloud/mail please review

jancborchardt:
We should sort folders by amount of unread count first, and then afterwards alphabetically. That way you have the relevant folders up top.

ChristophWurst:
> We should sort folders by amount of unread count first, and then afterwards alphabetically. That way you have the relevant folders up top.

Feel free to open a ticket for this change/enhancement. This PR is here to fix an annoying bug in the current behavior.

PR #931, opened 2018-06-06
When deleting account also delete all its aliases

PanCakeConnaisseur:
Not sure if if new `deleteAll()` should return the deleted aliases like the other methods do.

### DIFF
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -186,6 +186,7 @@ class AccountsController extends Controller {
 	 * @return JSONResponse
 	 */
 	public function destroy($id) {
+		$this->aliasesService->deleteAll($id, $this->currentUserId);
 		$this->accountService->delete($this->currentUserId, $id);
 		return new JSONResponse();
 	}
--- a/lib/Service/AliasesService.php
+++ b/lib/Service/AliasesService.php
@@ -79,4 +79,15 @@ class AliasesService {
 		return $alias;
 	}
 
+	/**
+	 * Deletes all aliases of an account.
+	 * @param int $accountId the account which aliases will be deleted
+	 * @param string $currentUserId the user whom the account belongs to
+	 */
+	public function deleteAll($accountId, $currentUserId) {
+		$aliases = $this->mapper->findAll($accountId, $currentUserId);
+		foreach ($aliases as $alias) {
+			$this->mapper->delete($alias);
+		}
+	}
 }


lib/Controller/AccountsController.php
```diff
@@ -186,6 +186,7 @@ public function update($id, $accountName, $emailAddress, $password, $imapHost, $
 	 * @return JSONResponse
 	 */
 	public function destroy($id) {
+		$this->aliasesService->deleteAll($id, $this->currentUserId);
 		$this->accountService->delete($this->currentUserId, $id);
```
ChristophWurst: I'd prefer to hide this logic here and rather do it in the account service.

lib/Service/AliasesService.php
```diff
@@ -79,4 +79,15 @@ public function delete($aliasId, $currentUserId) {
...
 	}
 
+	/**
+	 * Deletes all aliases of an account.
+	 * @param int $accountId the account which aliases will be deleted
+	 * @param string $currentUserId the user whom the account belongs to
+	 */
+	public function deleteAll($accountId, $currentUserId) {
+		$aliases = $this->mapper->findAll($accountId, $currentUserId);
+		foreach ($aliases as $alias) {
```
ChristophWurst: Not quite efficient but fine for account deletions :wink:

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/932

ChristophWurst:
> First-time contributor

Congratulations on your first contribution to this repository :raised_hands:

lib/Controller/AccountsController.php
PanCakeConnaisseur: I agree, I wanted to put it into `AccountService` as well, but could not find a way to call `AliasesService` from there. Did I miss something?

lib/Service/AliasesService.php
PanCakeConnaisseur: Agree, I wanted to avoid creating a new mapper method, but a dedicated SQL query that deletes all at once would be better.

lib/Controller/AccountsController.php
ChristophWurst: I'd suggest to inject the alias mapper in the account service.

PanCakeConnaisseur: so, extending the constructor with an additional argument as well?

ChristophWurst: Exactly

PanCakeConnaisseur: OK, I just wanted to make sure I don't blow up the constructor further needlessly.

PanCakeConnaisseur:
I started working on the PR again. Do the delete methods have to return the deleted elements? I only found [one usage](https://github.com/nextcloud/mail/blob/6634041f6a561bb903d5038987e4c1cddcf869ca/lib/Controller/AliasesController.php#L88) but don't understand what exactly it does.

ChristophWurst:
> Do the delete methods have to return the deleted elements? I found that the return value is only used once but don't understand what exactly it does.

It's possible that we're not making any use of it at the moment, but that's how REST usually works.

PR #933, opened 2018-06-07
Add support for basic "advanced search"

icewind1991:
Allows for the following searches:

- Whether or not the `read`, `unread` and `answered` flags are set (`is:unread`)
- Search in `from`, `to`, `cc`, `bcc` and `subject` (`from:somebody@example.com`)

Born from the need to find that one unread message in my inbox

### DIFF
--- a/lib/SearchHelper.php
+++ b/lib/SearchHelper.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * @copyright Copyright (c) 2018 Robin Appelman <robin@icewind.nl>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Mail;
+
+
+class SearchHelper {
+	const FLAG_MAP = [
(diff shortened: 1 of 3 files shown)


ChristophWurst:
Awesome feature and perfect timing. I was thinking about a possible implementation this week :rocket: 

One of the issues I reflected about was whether we should translate the filters. But I guess that gets unnecessarily complex and those filters are only used by power users anyway, so fine by me to have it in English only.

The second issue I see from the diff is that it won't handle terms that aren't a single word, like `subject: "work report"`, right? That would require a slightly more complex parser, I guess.

Thanks a lot!

lib/SearchHelper.php
```diff
@@ -0,0 +1,80 @@
...
+		}
+		$tokens = explode(' ', $filter);
+		$textTokens = [];
+		foreach ($tokens as $token) {
+			if (!$this->parseFilterToken($query, $token)) {
+				$textTokens[] = $token;
+			}
+		}
+		if (count($textTokens)) {
+			$query->text(implode(' ', $textTokens), false);
```
ChristophWurst: Could you please share some insight on why you're adding the tokens as text search query here as well, e.g. when they are header text like subjects?

stefan-niedermann:
@ChristophWurst

have you thought (in long term) about graphical filters, e. g. like atlassian jira does?

https://confluence.atlassian.com/jira064/files/720416597/720416653/1/1366249159018/jira-issuenavigator.png

ChristophWurst:
> have you thought (in long term) about graphical filters, e. g. like atlassian jira does?

Are those stored search queries that appear like folders? If so, that would make an awesome feature! Would you mind creating a feature request ticket for that?

lib/SearchHelper.php
icewind1991: `$textTokens` are all parts of the search query which were not parsed to have a special meaning, and just get used for "normal" text search

icewind1991:
> The second issue I see from the diff is that it won't handle terms that aren't a single word, like subject: "work report", right? That would require a slightly more complex parser, I guess.

That would make the parser significantly more complex yes

> Are those stored search queries that appear like folders

Though about that as well, would be really useful to have

lib/SearchHelper.php
ChristophWurst: You're right, I misread the condition :see_no_evil:

PR #979, opened 2018-07-23
Center align text on the redirect screen

kevgathuku:
Before:
screen shot 2018-07-23 at 18 25 05

After:
screen shot 2018-07-23 at 18 26 20

### DIFF
--- a/templates/redirect.php
+++ b/templates/redirect.php
@@ -36,7 +36,7 @@
 		</p>
 	</div>
 <?php else: ?>
-	<div class="error">
+	<div class="error" style="text-align: center;">
 		<h2><?php p($l->t('Redirect')); ?></h2>
 		<p><?php p($l->t('The link leads to %s', array($_['urlHost']))); ?></p>
 		<p><?php print_unescaped($l->t('If you don’t want to visit that page, you can return to <a href="%s">Mail</a>.',


ChristophWurst:
Looks good to me!

I've restarted the failing CI job. It looked like a temporary issue so it should pass now.

jancborchardt:
I would center-align the button (as the h2 is), but not sure about the text. :) Text blocks which are multiple lines are always nicer to read when left-aligned.

ChristophWurst:
> Please use css files

@kevgathuku please move this to a separate css file (e.g. `redirect.css`) and include it [like this](https://github.com/nextcloud/mail/blob/master/templates/index.php#L26) in `redirect.php`. Thank you :rocket:

kevgathuku:
Thanks for the pointers @ChristophWurst 
Working on it 👍

kevgathuku:
> Text blocks which are multiple lines are always nicer to read when left-aligned.

For this case it seems that center-aligning everything is better than left-aligning the longer text
cc @jancborchardt 

screen shot 2018-08-09 at 09 41 41

screen shot 2018-08-09 at 09 43 35

ChristophWurst:
@jancborchardt @skjnldsv please re-review! Thanks!

PR #984, opened 2018-07-24
Make OCA\Mail\AppInfo strict

ChristophWurst:
- [x] Requires https://github.com/nextcloud/mail/pull/983


ChristophWurst:
Tiny change and CI is happy :+1:

PR #985, opened 2018-07-24
Make OCA\Mail\Cache strict

ChristophWurst:


ChristophWurst:
> Scrutinizer — 6 new issues

Minor an can be ignored for this copy-paste class.

PR #986, opened 2018-07-24
Make OCA\Mail\Command strict

ChristophWurst:


ChristophWurst:
> Scrutinizer — 4 new issues

Not really new issues, so :+1:

PR #991, opened 2018-07-29
mail app does not send fqdn #489 - set localhost with ServerHostName

### LINKED ISSUE #489: mail app does not send fqdn
The Mail App fails when sending to a server configured to reject hosts with no FQDN. Below are some notes from trying to solve that problem:

NextCloud uses the Horde libraries for the mail app. While I don’t know zilch about PHP I thought I would at least look at the code.

The interesting part is in apps/mail/vendor/pear-pear.horde.org/Horde_Smtp/Horde/Smtp.php.

First there is a function _hello that does the EHLO exchange:

```
protected function _hello()
{
    $ehlo = $host = $this->_getHostname();
    if ($host === false) {
        $ehlo = $_SERVER['SERVER_ADDR'];
        $host = 'localhost';
    }
...
```

We want the function _getHostname:

```
protected function _getHostname()
{
    return ($localhost = $this->getParam('localhost'))
        ? $localhost
        : gethostname();
}
```

It appears that the PHP builtin gethostname() can only return the non-qualified server host name. So what’s left is the function getParam.

getParam does things to a _params array. That takes us back to the top of the file where there is a constructor filling up a params array. And there is provision for a parameter named localhost. I figured I could play with the array too. So I added a line to force params(‘localhost’) to contain my fqdn:

```
public function __construct(array $params = array())
{
    // Default values.
    $params = array_merge(array(
        'chunk_size' => self::CHUNK_DEFAULT,
        'context' => array(),
        'host' => 'localhost',
        'port' => 587,
        'secure' => true,
        'timeout' => 30,
+        'localhost' => 'myhostname.example.com'
    ), array_filter($params));
```

Save. Try sending. Bingo.

At best that’s a hack, not a fix; I’ll have to edit that file for every upgrade.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48867723-mail-app-does-not-send-fqdn?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

drajar:
@ChristophWurst pls allow this pull request to resolve #489


ChristophWurst:
Thank you!

I will try to integrate this pull request soon!

ChristophWurst:
> First-time contributor

Congratulations, btw :rocket:

ChristophWurst:
Btw, @drajar, you're very welcome to also join us at our [conference](https://eventyay.com/e/77d26f89/) where we can hack on this app together :raised_hands:

PR #994, opened 2018-07-30
Make OCA\Mail\Exception strict

ChristophWurst:


ChristophWurst:
CI is happy and so am I

Issue #981, opened 2018-07-24
App structure incompatible with Nextcloud 14


ChristophWurst:
### Expected behavior
See a properly working app structure.

### Actual behavior
Structure is pretty much borked. The content element is moved underneath the left sidebar.

Ref https://github.com/nextcloud/mail/pull/973#issuecomment-407044413
Ref https://github.com/nextcloud/server/pull/9982

cc @skjnldsv

ChristophWurst:
@nextcloud/mail I have the feeling this breaking change will make it quite hard to support Nextcloud 13 and 14 at the same time.

I'd therefore suggest to do one final release for Nextcloud 13 and jump to 14+ afterwards. Objections?

ChristophWurst:
@skjnldsv do you know a specific commit where the structure is still compatible with Nextcloud 13? I'd like to temporarily check that out to be able to give the app a good test run.

skjnldsv:
@ChristophWurst just before https://github.com/nextcloud/server/pull/9982 :p

ChristophWurst:
> @ChristophWurst just before nextcloud/server#9982 :p

hehe, tried that. Didn't quite work :man_shrugging:

skjnldsv:
@ChristophWurst then it must be something else than the structure cleanup :/
What does not work exactly?

ChristophWurst:
Nvm, I'm testing 0.8.3 on a separate instance now. This should be fine since there were hardly any changes since the last release and there weren't many bug reports either :)

PR #1001, opened 2018-07-31
Fix app structure for Nextcloud 14

ChristophWurst:
Ref https://github.com/nextcloud/server/pull/9982https://github.com/nextcloud/server/pull/9982

My attempt at fixing https://github.com/nextcloud/mail/issues/981.

cc @jancborchardt @skjnldsv input/help would be *very* appreciated. I don't know if it's this apps CSS or the structure that makes is to hard to comply the new structure, but I can't get it right so far. It's a lot worse on master, but it seems that some additional wrapping elements cause the CSS rules not to apply. I can't remove them, however, because they are required for our UI framework (Marionette) …

---

Bugs

- [x] Overall app structure is not compliant with new standard
- [x] Navigation toggle button is shown all the time
- [x] Message details are wrapped in an additional (useless) element
- [x] Message detail header is not fixed anymore
- [x] Message details got narrower as well, doesn't take up the available space
- [x] Messages list got narrower
- [x] Message delete icon in message list is misplaced
- [ ] ~~colors variables dark theme compatibility~~ will fix this in a separate PR
- [x] message list needs fixes for the answered icon and attachment icon
- [x] messages list needs fixes for the loading icon and load more message button
- [x] Need a mobile fix for the narrow screens. (when clicking a message in the list add the 'showdetails' class on the app-content-list)

### DIFF
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -342,12 +342,6 @@
 }
 
 /* editor */
-#mail-new-message-fixed {
-	z-index: 111; /* navigation menu is 110 ;-) */
-	background-color: rgba(255, 255, 255, .75);
-	padding: 10px;
-	position: relative;
-}
 #mail-new-message-list {
 	margin-top: 36px;
 }
@@ -414,7 +408,7 @@ textarea.message-body {
 	padding-right: 25%;
 }
 
-#app .message-composer .composer-fields .noreply-box {
+#content .message-composer .composer-fields .noreply-box {
 	margin-top: 0;
 	background: #fdffc3;
 	padding-left: 64px;
@@ -562,7 +556,6 @@ input.submit-message,
 }
 
 #mail-message-header {
-	position: fixed;
 	height: 90px;
 	width: 50%;
 	z-index: 100;
@@ -633,13 +626,6 @@ input.submit-message,
 	height: 100%;
 	overflow-y: auto;
 	overflow-x: hidden;
-}
-.mail-message-container {
-	position: absolute;
-	top: 0px;
-	right: 0px;
-	bottom: 0px;
-	left: 0px;
 }
 
 .new-message-attachments li {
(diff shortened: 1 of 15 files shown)


ChristophWurst:
I will take another attempt at this tomorrow morning when it's below 30°C and I can think clearly :sweat:

skjnldsv:
aaaah, the good old marionette wrapping shitstorm ^^

ChristophWurst:
> I will take another attempt at this tomorrow morning when it's below 30°C and I can think clearly sweat

It's 24.4°C here right now, so I was able to resolve that one tricky wrapping div with some official Marionette kung fu.

What I don't get is the ``<div id="app-navigation-toggle" class="icon-menu"></div>`` part. Do I have to hide/show this manually? This element is now shown all the time and takes up the full width.

ChristophWurst:
> What I don't get is the <div id="app-navigation-toggle" class="icon-menu"></div> part. Do I have to hide/show this manually? This element is now shown all the time and takes up the full width.

Nvm, just discovered that this is added/injected automatically and so I had it there twice (one hidden and once visible).

ChristophWurst:
@skjnldsv I've updated the PR description with a list of bugs that have been fixed already and those I've found but not yet tackled.

skjnldsv:
> What I don't get is the <div id="app-navigation-toggle" class="icon-menu"></div> part. Do I have to hide/show this manually? This element is now shown all the time and takes up the full width.

You shouldn't add this, it should be added by the snapjs I think :thinking:

ChristophWurst:
> You shouldn't add this, it should be added by the snapjs I think thinking

https://github.com/nextcloud/mail/pull/1001#issuecomment-409451123 :wink:

skjnldsv:
Well done !! 🤗 😊

ChristophWurst:
Thanks a lot for your help! :rocket: :hugs:

PR #1004, opened 2018-08-01
Fix sending messages withouth a folder ID (new messages)

ChristophWurst:
cc @skjnldsv since you discovered this :)


ChristophWurst:
Of course I forgot to adjust the tests :see_no_evil:

PR #1005, opened 2018-08-01
Make OCA\Mail\Service strict

ChristophWurst:


ChristophWurst:
> Scrutinizer — No new issues

Yay

Issue #593, opened 2017-11-02
List of inboxes displayed as Sent


Bables55:
### Steps to reproduce
1. Starting Mail
2. Select Sent menu from left pane

### Expected behaviour
A list of sent mails is displayed

### Actual behaviour
Exactly the same display as the inbox

### Mail app

**Mail app version:** (see apps admin page)
0.7.4
**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
My providers mail.
**Transport security - IMAP:** (None, SSL, TLS, STARTTLS) 
SSL
**Transport security - SMTP:** (None, SSL, TLS, STARTTLS)
SSL
**Number of accounts:**
1
**Mail app version build date:** (only if you are using a Nightly Build)

### Server configuration

**Operating system**:
CentOS 7.3
**Web server:**
nginx 1.11.10
**Database:**
MariaDB 5.5.56
**PHP version:**
PHP 7.1.11
**Version:** (see admin page)
12.0.3
**Updated from an older version or fresh install:**
Mail 0.7.3

#### Client configuration
**Browser:**
Firefox, Vivaldi, Safari(include iOS) 
**Operating system:**

##### Web server error log
```
There is no Mail error.
```

##### Server log (data/nextcloud.log)
<details>
```
none
```
</details>

Bables55:
Append.
It seems to also occur in the above problem, "Junk mail" or "Trash". It seems that the original content(sent, junk, trash) is displayed momentarily and the Inbox is displayed immediately.

PR #1017, opened 2018-08-08
Fix nonexistent variable access in HTML transform codestent variable …

ChristophWurst:
…access in HTML transform code

Fixes https://sentry.io/share/issue/0f0cc91816c14953a7fb73d0d4666624/.
Ref https://github.com/nextcloud/mail/issues/593#issuecomment-342025424.

I've never really touched this magic before, so I'm asking for review of the three people that have according to GitHub.


ChristophWurst:
- [ ] Blocked by #1006

PR #1018, opened 2018-08-08
Fix favicon caching on invalid paths

ChristophWurst:
This prevents the lib from writing cache files to the `vendor` dir.

- [ ] Blocked by #1006 

Ref https://sentry.rullzer.com/share/issue/b64f0d016338469f8b059ddc2f4a6e4a/.


ChristophWurst:
> This prevents the lib from writing cache files to the vendor dir.

https://github.com/ArthurHoaro/favicon/blob/02163a16798bb024b0a413aa38400226665c1b2a/src/Favicon/Favicon.php#L316-L320

PR #1019, opened 2018-08-08
Feature: print email

### LINKED ISSUE #579: Print stylesheet [$65]

thx
-d

<bountysource-plugin>

---
</bountysource-plugin>

pierlon:
Fixes #579

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ build/artifacts
 # NetBeans project
 nbproject/
 
+# PhpStorm
+.idea/
+
 translations.js
 
 # JS coverage
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -386,7 +386,7 @@ input.submit-message,
 /* extra button styles */
 .submit-message.send,
 #mail_new_attachment,
-#forward-button {
+#options > input {
 	padding: 12px;
 }
 .new-message-attachments,
--- a/css/print.scss
+++ b/css/print.scss
@@ -1,0 +1,24 @@
+@media print {
+	#reply-composer {
+		display: none !important;
+	}
+
+	#options {
+		display: none !important;
+	}
+
+	.attachments {
+		.attachment-name {
+			overflow: inherit;
+			text-overflow: inherit;
+	}
+
+		button {
+			display: none !important;
+		}
+	}
+
+    .section {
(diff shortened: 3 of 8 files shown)


tevon456:
both reviewers are offline right now

ChristophWurst:
Awesome. Will review this soonish.

cc @skjnldsv @jancborchardt for the styling aspects of this patch.

pierlon:
@ChristophWurst thanks.

skjnldsv:
@ChristophWurst  seems ok at first glance, but since we're just stripping the design out of everything not related to the mail, does it really matter? :p 
It's just displaying the email and printing in the end?  :)

ChristophWurst:
> @ChristophWurst seems ok at first glance, but since we're just stripping the design out of everything not related to the mail, does it really matter? :p
> It's just displaying the email and printing in the end? :)

:man_shrugging: :see_no_evil:

pierlon:
@ChristophWurst made changes as requested 😄

jancborchardt:
@ChristophWurst regarding placement of the print button, it might be time to add a "Reply all" button on the top right, with a 3-dot menu next to it containing Reply, Forward, Favorite, Move, Print and Delete. Also that we have this stuff working on mobile. What do you think?

ChristophWurst:
> @ChristophWurst regarding placement of the print button, it might be time to add a "Reply all" button on the top right, with a 3-dot menu next to it containing Reply, Forward, Favorite, Move, Print and Delete. Also that we have this stuff working on mobile. What do you think?

Sounds good. Any chance you have some time to help us with the design?

jancborchardt:
@ChristophWurst yes, for sure! :) The main work however is to put the actions in the popover menu as described in the design guidelines: https://docs.nextcloud.com/server/14/developer_manual/design/popovermenu.html

And left of it simply a `<button class="primary">Reply all</button>`, which would also focus on the reply form on the bottom. And both elements should be in the header element.

I can then help with the exact layout, and also the behavior on mobile. Is that what you mean?

ChristophWurst:
Closing as this needs a bit of rework for the Vue migration and I have not received any feedback within three weeks (https://github.com/nextcloud/mail/pull/1019#pullrequestreview-145537636).

Issue #960, opened 2018-06-27
Switch DB mappers to query builder mappers


ChristophWurst:
Ref https://github.com/nextcloud/server/pull/9444

ChristophWurst:
- [ ] Todo: depend on Nextcloud 14+

PR #959, opened 2018-06-26
Delete aliases when deleting account in account service

PanCakeConnaisseur:
This time deletion is initiated by `AccountService` and all aliases are deleted in one SQL query.

Also I fixed a typo in a file name.

PR #1036, opened 2018-08-16
Migrate to query builder mappers

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/960.

Required for #959.

This migration is mandatory as the old mappers are deprecated now.

### DIFF
--- a/lib/Db/CollectedAddressMapper.php
+++ b/lib/Db/CollectedAddressMapper.php
@@ -48,13 +48,11 @@ class CollectedAddressMapper extends QBMapper {
 		$dbQuery = $qb
 			->select('*')
 			->from($this->getTableName())
-			->where(
-				$qb->expr()->eq('user_id', $qb->createNamedParameter($userId)),
-				$qb->expr()->orX(
-					$qb->expr()->iLike('email', $qb->createNamedParameter("%$query%")),
-					$qb->expr()->iLike('display_name', $qb->createNamedParameter("%$query%"))
-				)
-			);
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->orX(
+				$qb->expr()->iLike('email', $qb->createNamedParameter("%$query%")),
+				$qb->expr()->iLike('display_name', $qb->createNamedParameter("%$query%"))
+			));
 
 		return $this->findEntities($dbQuery);
 	}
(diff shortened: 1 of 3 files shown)


lib/Db/CollectedAddressMapper.php
```diff
@@ -40,25 +40,34 @@ public function __construct(IDBConnection $db) {
...
-			'%' . $query . '%',
-		];
-		return $this->findEntities($sql, $params);
+		$qb = $this->db->getQueryBuilder();
+		$dbQuery = $qb
+			->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->eq('user_id', $qb->createNamedParameter($userId)),
+				$qb->expr()->orX(
```
rullzer: Below you do `andWhere` on the sexond expression. Might make sense to do it consitently one way. (I prefer the andWhere since it makes it explicit it is an AND

rullzer: To be clear I mean

```
->where(
  <1>,
  <2>);
```

vs

```
->where(<1>)
->andWhere(<2>);
```

lib/Db/CollectedAddressMapper.php
```diff
@@ -40,25 +40,34 @@ public function __construct(IDBConnection $db) {
...
-			$userId,
-			$email
-		];
-		return count($this->findEntities($sql, $params)) > 0;
+		$qb = $this->db->getQueryBuilder();
+		$dbQuery = $qb
+			->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('email', $qb->createNamedParameter($email)));
```
rullzer: The original expersssion does an ILIKE.. this is pure equal. intentional?

lib/Db/LocalAttachmentMapper.php
```diff
@@ -37,15 +37,20 @@ public function __construct(IDBConnection $db) {
...
 	 * @return LocalAttachment
 	 */
 	public function find($userId, $id) {
-		$sql = 'SELECT * FROM `' . $this->getTableName() . '` WHERE user_id = ? and id = ?';
-		$params = [$userId, $id];
+		$qb = $this->db->getQueryBuilder();
+		$query = $qb
+			->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)));
```
rullzer: The oroginal statement has `id = ?` this one not. Intentional?

ChristophWurst: Nope, my fault.

PR #1047, opened 2018-08-20
Update composer.json to PHP 7

### LINKED ISSUE #1042: PHP config in composer.json
### Question

Seeing that [v0.9.0](https://github.com/nextcloud/mail/releases/tag/v0.9.0) now requires a minimum of PHP 7, shouldn't it be reflected in the [composer.json](https://github.com/nextcloud/mail/blob/master/composer.json) file also?

pierlon:
Fixes #1042.

I chose `v7.0.13` because its the minimum supported version of PHP 7, but its security support ends in 3 months. Should `v7.1.20` be used instead?

---

PHP supported versions: https://secure.php.net/supported-versions.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,11 +7,11 @@
 	],
 	"config": {
 		"platform": {
-			"php": "5.6"
+			"php": "7.0.13"
 		}
 	},
 	"require": {
-		"php": ">=5.6.0",
+		"php": ">=7.0.13",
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> I saw you started something similar a week ago: https://github.com/nextcloud/mail/compare/fix/composer-php-version

Ha! Good catch. I knew I was working on this already but I couldn't find a PR when I saw #1042 :see_no_evil: :speak_no_evil: :hear_no_evil:

Issue #966, opened 2018-07-10
Cannot connect to G Suite accounts (and possibly Gmail accounts as well)


neonnero:
### Expected behavior
Setting up a G Suite email account (Google) using a custom domain name (other than Gmail.com) should be detected, either from the list of MX servers in DNS, or via the Autodiscover method (or, if all else fails, test with a list of common subdomains for email), and possibly redirect me to a web login, if Gmail requires this, where I can accept the Nextcloud Mail app to access my email.

### Actual behavior
Nextcloud Mail app ignores my custom autodiscover.domain.tld setup, and instead uses domain.tld as IMAP server, port 143 with STARTTLS (instead of imap.gmail.com with SSL/TLS on port 993), and one of the servers from the MX list for domain.tld (such as ALT1.ASPMX.L.GOOGLE.com) using port 25 with STARTTLS (instead of smtp.gmail.com with SSL/TLD on port 465).

Connection fails due to this incorrect setup.

Also, for pure IMAP to happen (without anything else), my Google Account setup needs to have the "Allow less secure apps" setting enabled. Otherwise, it would require a separate web login.

### Mail app

**Mail app version:** 0.8.2

**Mailserver or service:** Gmail

**Number of accounts:** 1

### Server configuration

**Operating system**: Ubuntu 16.04.4 LTS (Xenial)

**Web server:** nginx/1.10.3

**Database:** MySQL 5.7.22-0ubuntu0.16.04.1

**PHP version:** PHP 7.0.30-0ubuntu0.16.04.1

**Nextcloud Version:** 13.0.4

**Updated from an older version or fresh install:** Updated

**List of activated apps:**
```
Enabled:
  - activity: 2.6.1
  - admin_notifications: 1.0.1
  - audioplayer: 2.3.1
  - bookmarks: 0.11.0
  - bruteforcesettings: 1.1.0
  - calendar: 1.6.1
  - comments: 1.3.0
  - contacts: 2.1.5
  - dav: 1.4.7
  - drawio: 0.8.9
  - federatedfilesharing: 1.3.1
  - federation: 1.3.0
  - files: 1.8.0
  - files_pdfviewer: 1.2.1
  - files_sharing: 1.5.0
  - files_texteditor: 2.5.1
  - files_trashbin: 1.3.0
  - files_versions: 1.6.0
  - files_videoplayer: 1.2.0
  - firstrunwizard: 2.2.1
  - gallery: 18.0.0
  - logreader: 2.0.0
[...]

Issue #808, opened 2018-02-26
Mail says server denied auth when accepted


virtualdxs:
### Steps to reproduce
1. Create new mail account for Exchange 2007 server

### Expected behaviour
Successful account creation

### Actual behaviour
Error while creating the account: Creating account failed: Server denied authentication.

### Mail app

**Mail app version:** 0.7.8

**Mailserver or service:** Exchange 2007

**Number of accounts:** Creating second one

## Server configuration detail

**Operating system:** Linux 4.14.12-x86_64-linode92 #1 SMP Fri Jan 5 15:34:44 UTC 2018 x86_64

**Webserver:** Caddy/0.10.10 (+9eba91f) (fpm-fcgi)

**Database:** pgsql PostgreSQL 10.1 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 7.2.0, 64-bit

**PHP version:** 7.2.1
Modules loaded: Core, date, libxml, openssl, pcre, zlib, ctype, dom, fileinfo, filter, hash, json, mbstring, pcntl, SPL, PDO, session, posix, readline, Reflection, standard, SimpleXML, Phar, tokenizer, xml, xmlreader, xmlwriter, mysqlnd, cgi-fcgi, curl, gd, pdo_pgsql, zip

**Nextcloud version:** 13.0.0 Beta 3 - 13.0.0.8

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated

Here's the tcpflow output:
```
tcpflow: listening on lo
127.000.000.001.08143-127.000.000.001.38942: * OK The Microsoft Exchange IMAP4 service is ready.

127.000.000.001.38942-127.000.000.001.08143: 1 CAPABILITY

127.000.000.001.08143-127.000.000.001.38942: * CAPABILITY IMAP4 IMAP4rev1 AUTH=NTLM AUTH=GSSAPI AUTH=PLAIN IDLE NAMESPACE LITERAL+
1 OK CAPABILITY completed.

127.000.000.001.38942-127.000.000.001.08143: 2 AUTHENTICATE PLAIN

127.000.000.001.08143-127.000.000.001.38942: +

127.000.000.001.38942-127.000.000.001.08143: [redacted]

127.000.000.001.08143-127.000.000.001.38942: 2 OK AUTHENTICATE completed.

127.000.000.001.38942-127.000.000.001.08143: 3 CAPABILITY

[...]

PR #1048, opened 2018-08-21
Increase SMTP timeout

pierlon:
I kept getting the error: `Server denied authentication` while logging to my mail server. After some debugging, I found the SMTP timeout to be the problem.

This may fix issues #966 & #808.

### DIFF
--- a/lib/SMTP/SmtpClientFactory.php
+++ b/lib/SMTP/SmtpClientFactory.php
@@ -76,7 +76,7 @@ class SmtpClientFactory {
 			'port' => $mailAccount->getOutboundPort(),
 			'username' => $mailAccount->getOutboundUser(),
 			'secure' => $security === 'none' ? false : $security,
-			'timeout' => (int)$this->config->getSystemValue('app.mail.smtp.timeout', 2)
+			'timeout' => (int)$this->config->getSystemValue('app.mail.smtp.timeout', 10)
 		];
 		if ($this->config->getSystemValue('debug', false)) {
 			$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde_smtp.log';


lib/SMTP/SmtpClientFactory.php
```diff
@@ -76,7 +76,7 @@ public function create(Account $account): Horde_Mail_Transport {
 			'port' => $mailAccount->getOutboundPort(),
 			'username' => $mailAccount->getOutboundUser(),
 			'secure' => $security === 'none' ? false : $security,
-			'timeout' => (int)$this->config->getSystemValue('app.mail.smtp.timeout', 2)
+			'timeout' => (int)$this->config->getSystemValue('app.mail.smtp.timeout', 10)
```
ChristophWurst: This is a known problem on some setups, hence this is configurable via `config/config.php` (the `app.mail.smtp.timeout` key).

I suggest to simply change the value on your setup. The default timeout is fine IMO. cc @nextcloud/mail

pierlon: True, however the error message does not indicate the cause. The [IMAP timeout](https://github.com/nextcloud/mail/blob/master/lib/Account.php#L145) is 20 seconds.

ChristophWurst: Good point. I think the default IMAP timeout should be lower :wink:

pierlon: Yea that makes sense. Should I change this PR to lower the IMAP timeout instead?

ChristophWurst: Yes, please. How about we make both 5s by default for both? Sounds like a fair compromise to me.

Issue #1045, opened 2018-08-20
Vue


ChristophWurst:
Hey, @nextcloud/mail,

as we can all probably agree on, our current UI framework (jQuery/Backbone/Marionette) is a bit of a pain to work with. It was a great enhancement over simple Backbone views when those were the standard for old ownCloud apps, but times change and now there are way better frameworks available that would easy the long-term maintainability of this app.

Therefore, I'd like to propose the transition to Vue. This will allow us to share common components across apps and move to a more pleasant development style.

I've started gathering some brainstorming ideas for tasks that have to be worked on at https://github.com/nextcloud/mail/projects/4.

Marionette still isn't *that* bad for simple apps. However, this app has grown significantly, with highly nested data structures and lots of asynchronous actions. Naturally, this led to a hard-to-maintain state management, which I failed at certain aspects and the app is suspect to some concurrency issues (e.g. routing vs. async loading of content data). Therefore, I would *strongly* suggest to *fully* built on *Vuex* to have a one-way data flow (opposed to the highly complex event system in Backbone/Marionette) and defined state transitions.

Usually @Gomez and I spend some time working on a bigger change of this app at the conference, so I think this would be fun to hack on together at this year's [conference](https://eventyay.com/e/77d26f89/). Everybody is very welcome to join us. There's a lot work to be done :wink: 

@nextcloud/mail please let me know what you think.

Gomez:
Yeah! Sounds like a plan :) Great opportunity to dive deeper into vue and get it started.

PR #1050, opened 2018-08-22
Keyboard accessibility

jancborchardt:
This fixes that there’s currently almost no keyboard navigation feedback in the mail app.

- Fix keyboard tab feedback in app-navigation
- Fix keyboard feedback and navigability for showing more folders
- Fix keyboard feedback for message list 

Still missing is that the message list (and the 3-dot-menu in the message list) is actually interactive. @ChristophWurst I guess you know how to do that?

### DIFF
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -737,8 +737,8 @@ input.submit-message,
 				padding-left: 38px !important;
 			}
 
-			&:focus,
 			&:hover,
+			& a:focus,
 			&.active,
 			a.selected {
 				&,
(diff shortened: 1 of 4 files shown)


kevgathuku:
@jancborchardt adding a commit to handle the 3-dot menu activation on `Enter/Space` keypress

ChristophWurst:
This needs rework for #1045. Shall we close this for now?

jancborchardt:
Could we just make sure that it’s transferred to the Vue rework? Because it’s small things, but all over the place.

ChristophWurst:
You can do that too of course.

ChristophWurst:
Please re-apply the changes on master and re-submit. Closing due to lack of feedback.

PR #1063, opened 2018-08-27
Added folderfilter for noselect entries in imap-folder-structure

### LINKED ISSUE #1033: t-online imap does not show entries properly in unified inbox
### Expected behavior
Unified Inbox shows all Mails once.

### Actual behavior
Unified Inbox shows specific Mails once, and others are missing completely.

The first 10 entries on my t-online imap account are shown twice as seperate entries.
If i scroll down, new entries from web.de are loaded, but none from t-online.de

The normal inbox folder works flawless, except that there is a second inbox, which is not clickable. (The mouseover changes to "not-allowed")
This behavior is consistent over multiple accounts (from t-online).

I could not find a similar bug, is this a configuration-error on my part?

### Mail app

**Mail app version:**
 0.8.3

t-online.de
web.de

**Number of accounts:** (e.g. 1 or 2)
6

Ubuntu 16.04

Apache 2.4

**Database:**
MySQL

13.05

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62121973-t-online-imap-does-not-show-entries-properly-in-unified-inbox?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

newhinton:
This adds a filter for noselect entries in the imap folderstructure. It should fix problems related to  #1033.
Fixes  #1033


lib/IMAP/FolderMapper.php
```diff
@@ -107,6 +107,13 @@ public function buildFolderHierarchy(array $folders, bool $havePrefix): array {
 			}
 		}
 
+		//filter for \\noselect
+        foreach ($top as $folder) {
```
ChristophWurst: please use `array_filter` instead, removing elements of an array you're currently iterating on is scary :see_no_evil:

ChristophWurst: Indentation is a bit off, btw

lib/Service/MailManager.php
```diff
@@ -55,20 +55,23 @@ class MailManager implements IMailManager {
...
-	 * @param FolderMapper $folderMapper
-	 * @param MailboxPrefixDetector $prefixDetector
-	 * @param FolderNameTranslator $folderNameTranslator
-	 * @param Synchronizer $synchronizer
-	 * @param MessageMapper $messageMapper
-	 */
+    /**
+     * @param IMAPClientFactory $imapClientFactory
+     * @param FolderMapper $folderMapper
+     * @param MailboxPrefixDetector $prefixDetector
```
ChristophWurst: please revert the changes to this file

lib/IMAP/FolderMapper.php
ChristophWurst: What happens if I have a top-level mailbox that is not selectable with subfolders that are? Will the subfolders still be shown?

newhinton:
@ChristophWurst Should be fine now

ChristophWurst:
The logic was moved to the client and has to be redone. I'm also still afraid that this won't be a good solution for all setups.

PR #1076, opened 2018-09-04
Use v-tooltip instead of jQuery plugin

ChristophWurst:
@skjnldsv this one is for you.

### DIFF
--- a/js/components/Address.vue
+++ b/js/components/Address.vue
@@ -1,6 +1,6 @@
 <template>
 	<span :title="email"
-		  v-tooltip="{ content: 'label', placement: 'bottom' }">{{ label }}</span>
+		  v-tooltip.bottom="label">{{ label }}</span>
 </template>
 
 <script>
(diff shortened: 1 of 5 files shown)


js/components/Address.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<span :title="email"
-		  v-tooltip="tooltipData">{{ label }}</span>
+		  v-tooltip="{ content: 'label', placement: 'bottom' }">{{ label }}</span>
```
skjnldsv: `v-tooltip.bottom="label"`?

js/components/Address.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<span :title="email"
-		  v-tooltip="{ content: 'label', placement: 'bottom' }">{{ label }}</span>
+		  v-tooltip.bottom="label">{{ label }}</span>
```
skjnldsv: :heart:

PR #1077, opened 2018-09-04
Add Vue-based composer component

ChristophWurst:
Implements https://github.com/nextcloud/mail/projects/4#card-12257342.

TODO:
- [x] autosize magic
- [x] send/reply callbacks
- [x] draft callbacks
- [x] component properties


ChristophWurst:
Lot's of details missing, but those will be added when we wire the components with real data

PR #1082, opened 2018-09-06
Add account settings view

ChristophWurst:
cc @kevgathuku (don't worry I don't take away your ticket, I just added the basic view where all the components have to be added :wink:)


kevgathuku:
@ChristophWurst No worries. This actually makes my work easier 😅

PR #1129, opened 2018-09-20
Add empty folder view in Vue

### LINKED ISSUE #1128: View: Empty folder

kevgathuku:
Fixes #1128

### DIFF
--- a/js/components/EmptyFolder.vue
+++ b/js/components/EmptyFolder.vue
@@ -1,0 +1,12 @@
+<template>
+  <div id="emptycontent">
+    <div class="icon-mail"></div>
+    <h2>{{ t('mail', 'No messages in this folder') }}</h2>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'EmptyFolder'
+};
+</script>
--- a/js/components/MessageList.vue
+++ b/js/components/MessageList.vue
@@ -1,5 +1,6 @@
 <template>
 	<div class="app-content-list">
+		<EmptyFolder v-if="messages.length === 0"/>
 		<MessageListItem v-for="message in messages"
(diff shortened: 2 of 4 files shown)


ChristophWurst:
Awesome!

I've got a few changes locally that I haven't pushed yet that will change the data structure to be a bit more vuex-friendly. This will cause some conflicts with this, I think, but they should be easy to resolve ;)

kevgathuku:
No worries @ChristophWurst 
I can make the changes once the structure is up 👍

ChristophWurst:
resolved :rocket: 

If you add `return Promise.resolve([]);` right above https://github.com/nextcloud/mail/blob/427c2f5b95c1dfc6a25677784530cbbc78927080/js/service/MessageService.js#L2 you can simulate an empty folder.

In that state, the empty folder view is shown, but there's an error in the console about `env` being undefined. Could you look into this? Thanks!

kevgathuku:
This is the line causing the issue:
https://github.com/nextcloud/mail/blob/refactor%2Fvue/js/store.js#L77
specifically `.map(msgId => state.envelopes[msgId])`

It seems `state.envelopes` is empty at this point

ChristophWurst:
Ah, my bad. You're right, this is because of the hacked-in empty folder. Removing the envelope ids from the folder would have been the correct hack :wink: 

Anyway, this is just temporary mocking data until we're pulling real data from the server.

kevgathuku:
The workaround for the issue is that when testing empty accounts, the `envelopes` key in https://github.com/nextcloud/mail/blob/refactor%2Fvue/js/store.js#L95 
should also be set to `[]` in addition to `return Promise.resolve([]);` in the `MessageService`

PR #1170, opened 2018-09-27
Port master changes to vue rewrite branch II

ChristophWurst:


ChristophWurst:
> What about tests?

I wanted to keep this change as small is possible.

PR #1171, opened 2018-09-27
Fix duplicate files / case-sensitivity issues

kevgathuku:
On a case-insensitive filesystem e.g. on MacOS, I'm having some issues with files that have the same name but the name is in different cases and different content.
e.g. in the attached image, you can see that we have 2 files named `accountservice.js` and `AccountService.js`
Same for `FolderService.js` and `MessageService.js`

My filesystem is treating those 2 as the same file and git is having issues differentiating the files.

The PR addresses this issue by removing the old files i.e. 

- `accountservice.js`
- `folderservice.js`
- `messageservice.js`

screen shot 2018-09-27 at 19 24 03


kevgathuku:
This is a blocker for me. Due to the `git` confusion i can't switch branches, etc.

PR #1172, opened 2018-09-27
Lint code with eslint and format with prettier

kevgathuku:
Resolves #1130 

Setup largely based off reading, which also contains info on how to configure your editor
 https://medium.com/@doppelmutzi/eslint-prettier-vue-workflow-46a3cf54332f

To show errors on console, run:
`npm run lint`

To auto-fix errors, run:
`npm run lint:autofix`

- This causes a lot of changes so I've decided to hold off on actually running the auto-formatter to avoid conflicts.

The prettier config is in `package.json`

PR #1174, opened 2018-09-28
Switch to nextcloud server API lib instead of global vars

ChristophWurst:
Ref https://github.com/ChristophWurst/nextcloud-server

cc @skjnldsv

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -1,14 +1,15 @@
 import Vue from 'vue'
 import Router from 'vue-router'
+import {generateUrl} from 'nextcloud-server/dist/router'
 
-import AccountSettings from './views/AccountSettings';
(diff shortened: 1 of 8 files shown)


skjnldsv:
Already? :O

js/router.js
```diff
@@ -1,14 +1,15 @@
 import Vue from 'vue'
 import Router from 'vue-router'
+import {generateUrl} from 'nextcloud-server/dist/router'
```
splitt3r: Spaces are missing here :grin:

ChristophWurst: hehe, yeah, the coding style is quite "creative" right now. @kevgathuku has added a linter in #1172 but we're not yet enforcing it. Will try to gradually move files to a uniform style soon :wink:

PR #1185, opened 2018-10-02
Adjust redirect page design

jancborchardt:
- [ ] Requires Nextcloud 15 (https://github.com/nextcloud/server/pull/11557)

> Redirect of Mail app before & after, had no hover/focus effect either. General style adjusted like the modal in https://github.com/nextcloud/server/pull/11536
> screenshot from 2018-10-02 20-45-48
> screenshot from 2018-10-02 20-38-44

### DIFF
--- a/templates/redirect.php
+++ b/templates/redirect.php
@@ -37,16 +37,17 @@ style('mail', 'redirect');
 		</p>
 	</div>
 <?php else: ?>
-	<div class="error">
+	<div class="update">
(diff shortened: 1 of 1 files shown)


ChristophWurst:
The block is that this requires Nextcloud 15: https://github.com/nextcloud/mail/blob/8a2f8f349bf9c6cf83b1689ec0c54e76a846c37f/appinfo/info.xml#L25

jancborchardt:
:tada:

PR #1252, opened 2018-10-31
Add tabs to the setup/settings screen

ChristophWurst:
Ref https://schoberg.net/2008/10/making-better-forms-clickable-text-for-radio-button-and-checkbox-fields/


ChristophWurst:
## Current state 
bildschirmfoto von 2018-10-31 17-16-05
bildschirmfoto von 2018-10-31 17-16-10

PR #1256, opened 2018-11-02
Use action component for envelope actions

ChristophWurst:
- [x] Requires https://github.com/nextcloud/vue-components/pull/55


ChristophWurst:
bildschirmfoto von 2018-11-02 13-42-52

@skjnldsv ^ ideas?

PR #1319, opened 2018-11-15
Prevent subject prefix stacking

### LINKED ISSUE #346: Prevent stacking of Re: and Fwd: also for other languages
The easiest would be to put this whole list of the translations of `Re:` and `Fwd:` into an array to check: https://en.wikipedia.org/wiki/List_of_email_subject_abbreviations#Abbreviations_in_other_languages

The relevant file is https://github.com/nextcloud/mail/blob/master/lib/Model/ReplyMessage.php

cc @avanc if you want to work on this. :)

MrMcX:
Should fix #346 
Couldn't run the test cases because I don't have the environment set up

### DIFF
--- a/lib/Model/ReplyMessage.php
+++ b/lib/Model/ReplyMessage.php
@@ -23,23 +23,37 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Model;
 
-class ReplyMessage extends Message {
+class ReplyMessage extends Message
+{
+    public function setSubject(string $subject)
+    {
+        parent::setSubject("Re: " . ReplyMessage::replacePrefix($subject));
+    }
 
-	public function setSubject(string $subject) {
-		// prevent 'Re: Re:' stacking
-		if (strcasecmp(substr($subject, 0, 4), 'Re: ') === 0) {
-			parent::setSubject($subject);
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Thank you for this contribution, @MrMcX (or @https://github.com/simon-moser? The commits seem to come from the latter).

We're currently rewriting the front-end of this app in Vue and this logic was moved from the server to the browser. I have to think about a strategy to include these changes. Maybe we just have to move it to js.

MrMcX:
It's both me, but different accounts ;) Just forgot to change the commit mail address..

If you tell me where this part runs now, I can look into it, but I have no time to read through the whole new code

ChristophWurst:
https://github.com/nextcloud/mail/blob/225cc7626287503d246e9c1a4ad3941fe6ee78d0/src/ReplyBuilder.js#L97-L116

But tbh I'm still undecided if we really should remove all these. I'm afraid this will break the threading mechanism of some clients.

ChristophWurst:
FYI the Vue rework was integrated in master.

ChristophWurst:
This has to be re-implemented on the client side to work with the new code base -> closing.

Thank you for your effort, @MrMcX :v:

PR #1330, opened 2018-11-23
Use the actual constructed query in the AccountMapper

rullzer:
And some strict+typehinting cleanup

### DIFF
--- a/lib/Db/MailAccountMapper.php
+++ b/lib/Db/MailAccountMapper.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
 /**
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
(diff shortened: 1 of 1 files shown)


lib/Db/MailAccountMapper.php
```diff
@@ -43,26 +44,25 @@ public function __construct(IDBConnection $db) {
...
+	public function find(string $userId, int $accountId): MailAccount {
 		$qb = $this->db->getQueryBuilder();
 		$query = $qb
 			->select('*')
 			->from($this->getTableName())
 			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
 			->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($accountId)));
 
-		return $this->findEntity($qb);
+		return $this->findEntity($query);
```
ChristophWurst: whoops https://github.com/nextcloud/mail/commit/22f15363c29f1ee7227d6aab219f8e6fad932ff7#diff-bf6daa524015d90a54445ef622fd3be2R54

PR #1356, opened 2018-12-04
Drafts view

ChristophWurst:


ChristophWurst:
>     * Fix transition hiccups

Will try to tackle this in a follow-up PR. Can't figure out how to suppress the collapse-old expand-new transition of the envelope getting a new UID (and therefore having the `key` change).

PR #1359, opened 2018-12-05
Allow mail body formatting

### LINKED ISSUE #210: Allow Mailbody formating [$100]
### Steps to reproduce
1. Open Mail App
2. Click on Button "New Message"
3. Type some Text in the Textarea for Mailbody

### Expected behaviour
There should be a Buttonbar, which let me format the Mail with the most needed Formats like bold, italic, font color, font size, font name. Maybe its possible to add something like tinymce.

### Actual behaviour
Currently, its a pure Textarea, only normal Text could be entered.

### Mail app

**Mail app version: 0.60**

**Mailserver or service: Personal IMAP Server **

### Server configuration
**Operating system**: Raspbian (Debian, Jessie)

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 5.6

**Updated from an older version or fresh install:** fresh install

#### Client configuration
**Browser:** Firefox 50

**Operating system:** Windows

<bountysource-plugin>

---
There is a **[$100 open bounty](https://www.bountysource.com/issues/39482772-allow-mailbody-formating?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pierlon:
This should fix #210.

There's a problem with inline images, however. The HTML sanitizer is removing them, but I haven't gotten to look at that code as yet.


webpack.common.js
```diff
@@ -47,12 +47,21 @@ module.exports = {
 			},
 			{
 				test: /\.(svg)$/i,
+				loader: 'url-loader',
+				exclude: '/node_modules/quill/',
+			},
+			{
+				test: /\.(svg)$/i,
+				include: '/node_modules/quill/',
```
ChristophWurst: what's that used for?

ChristophWurst:
Nice, thanks a lot. 

Will review/test this once the Vue rework is done. We decided not to change any functionality like this until we've reached feature parity with the previous version.

Btw, wouldn't it be easier to use a Vue wrapper of the editor like https://www.npmjs.com/package/vue-quill-editor?

Is it possible to switch to text-only editing? This is a must.

ChristophWurst:
I see you included the "snow" theme. Would it be possible to select an appropriate theme according to the theme? Nextcloud has a dark mode and I'm sure a white editor wouldn't fit the design.

pierlon:
> Will review/test this once the Vue rework is done. We decided not to change any functionality like this until we've reached feature parity with the previous version.

Ok no problem :smile:.

> Btw, wouldn't it be easier to use a Vue wrapper of the editor like https://www.npmjs.com/package/vue-quill-editor?

I prefer to use the source rather than the wrapper (it's outdated btw).

> Is it possible to switch to text-only editing? This is a must.

Will work on that.

> I see you included the "snow" theme. Would it be possible to select an appropriate theme according to the theme? Nextcloud has a dark mode and I'm sure a white editor wouldn't fit the design.

Will work on that also.

webpack.common.js
pierlon: The svg files in from the quill module require the html-loader rather than the url-loader (they get converted to base64 when loaded with url-loader).

ChristophWurst:
@pierlon :ping_pong:

pierlon:
Hey, been busy for the past few days, so was unable to work on some of my PRs for a while. I'll push what I have for this one, although I'm not finished.

ChristophWurst:
No worries. Thank you for your effort :v:

ChristophWurst:
:ping_pong:

pierlon:
Hey @ChristophWurst, haven't been able to work on this for a long time. Hopefully the summer break allows me to :crossed_fingers:

ChristophWurst:
I'd like to help you with this now. First of all, let's update the branch and solve the conflicts. Commits incoming  …

src/components/Composer.vue
```diff
@@ -129,6 +130,7 @@
 	import Vue from 'vue'
 
 	import {findRecipient} from '../service/AutocompleteService'
+	import Editor from './Editor'
```
ChristophWurst: you forgot to include that file :D

pierlon:
Hey @ChristophWurst, looking forward to help close this issue! 

If I remember correctly, I was having some issues modifying Quill to fit the project's needs and was trying out a HTML editor called Squire (check out the branch [here](https://github.com/pierlon/mail/tree/feature/html-editor)). Squire is very lightweight and can be customized and iterated upon much easier than Quill. There are a few missing features and bugs I need to squash on that branch, but they should be easily fixable.

Let me know which editor you think would be better in the long run :running_man: .

ChristophWurst:
> Let me know which editor you think would be better in the long run running_man .

I've looked into that like three times already. Never found anything that looked a 100% satisfying but I'll give that another try.

ChristophWurst:
Superseded by #1988. Thanks for your help @pierlon!

PR #1386, opened 2018-12-12
Collapse folders only if there is less than 6 and remember collapse choice

### LINKED ISSUE #692: Show all folders if only one is behind the „Show all folders“-button [$15]
4f427aa6-4e2f-4d79-a09f-466884425ff4

@ChristophWurst ... any clue why this is flickering when clicking on „Show all folders“ 🤭

<bountysource-plugin>

---
There is a **[$15 open bounty](https://www.bountysource.com/issues/53312344-show-all-folders-if-only-one-is-behind-the-show-all-folders-button?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ALSai:
This commit fixes #692 in a more generic way :
If there are less than 6 folders in an account, it hides the "Show all folders" button ans show all of them. It fix the initial need.
More, the second commit add a feature to store the "collapse folders" button state in localStorage in browser. It's a per account setting.

### DIFF
--- a/js/templates/account.html
+++ b/js/templates/account.html
@@ -35,7 +35,7 @@
 
 <div class="folders"></div>
 {{#unless isUnifiedInbox}}
-{{#if hasFolders}}
+{{#if hasCollapseFoldersButton}}
 <li class="account-toggle-collapse">
 	<a>{{toggleCollapseMessage}}</a>
 </li>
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Awesome feature!

However, I have to inform you that this version of your front-end code is unmaintained and we're currently re-writing it with Vue. See #1045.

Let's see if we can reuse the logic.

ALSai:
I will try to do the same on the Vue branch.

PR #1405, opened 2018-12-19
Fix all linting error & implement pre commit hook.

### LINKED ISSUE #1293: Code formatting issues
### Question

I'm creating a feature from the `refactor/vue` branch, but I'm being bombarded with formatting errors from `eslint` and `prettier`. Should I format the files that I edit, or leave them be?

sahalsaad:
Fixes #1293

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "eslint-plugin-prettier": "^3.0.0",
     "eslint-plugin-vue": "^5.0.0",
     "file-loader": "^2.0.0",
-    "jsdom": "^13.0.0",
+    "jsdom": "^13.1.0",
     "jsdom-global": "^3.0.2",
     "mocha": "^5.2.0",
     "mocha-webpack": "^2.0.0-beta.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1200,9 +1200,9 @@
       },
       "dependencies": {
         "acorn": {
-          "version": "6.0.2",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-6.0.2.tgz",
-          "integrity": "sha512-GXmKIvbrN3TV7aVqAzVFaMW8F8wzVX7voEBRO3bDA64+EX37YSayggRJP5Xig6HYHBkWKpFg9W5gg6orklubhg==",
+          "version": "6.0.4",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -78,10 +80,21 @@
...
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,vue}": [
+      "npm run lint:autofix",
+      "git add"
```
ChristophWurst: will this add all the files automatically?

sahalsaad: Not all. the staged file only.

ChristophWurst: I see. What if don't want the full file to be staged? It happens rarely, but sometimes I like to `git add -p` and only add parts of a file. Will this be impossible?

I appreciate adding the commit hook but I think we should just lint on CI.

sahalsaad: I tried with the `git add -p`. It does not touch whatever you wish not to commit. So technically this only run on the staged item. My earlier explanation is wrong, not staged file but staged item(can be file/line). 

With this, we can catch the lint error earlier, does not need to wait for CI build. My self a bit annoyed when the build failed just because of lint but we can remove this if lint on CI is good enough.

ChristophWurst:
Could you please update to master?

Also, I think we should remove the commit hook for now. I'm not aware of any Nextcloud apps that use such a mechanism right now and I don't want to confuse anybody.

sahalsaad:
Hi @ChristophWurst 
Updated to master and removed the commit hook.

ChristophWurst:
Sorry for the delay. I tried to update the branch locally for a final review but it's terribly conflicting. How can I resolve these issues with the automatic linting?

sahalsaad:
> Sorry for the delay. I tried to update the branch locally for a final review but it's terribly conflicting. How can I resolve these issues with the automatic linting?

When there is any conflict on js/vue file during rebase, just accept yours changes first(because my changes just lint). After that you can run `npm run lint:autofix` to fix the linting. If there rules does not support autofix, we need to fix it manually.

ChristophWurst:
That makes sense! Thanks!

ChristophWurst:
I did something wrong here. Will try to fix ASAP

PR #1451, opened 2019-01-09
Add phan as composer dev dependency and add a composer script

ChristophWurst:


ChristophWurst:
> Should we also run it on travis then?

Yeah, I pan to add a job for it in a follow-up PR!

PR #1453, opened 2019-01-09
Simplify the bug report issue template

ChristophWurst:
For many reports we don't really care about the detailed information like logs. And most people don't get how to access those anyway, so I'd rather not ask for it unless really required and then I can answer with a saved reply.

cc @nextcloud/mail

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -4,22 +4,21 @@ about: Have you encountered a bug?
 title: ''
 labels: bug, 0. to triage
 assignees: ''
-
 ---
 
 ### Expected behavior
 Tell us what should happen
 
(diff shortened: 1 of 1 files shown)


jancborchardt:
Very good call! :) You can always save the other parts as a [saved reply](https://github.com/settings/replies) so you can quickly ask for further info.

ChristophWurst:
Yep, that is exactly what I had written in the description :see_no_evil: Maybe it just wasn't clear :wink:

jancborchardt:
Haha, busted! … well I guess it is true what they say about people not reading. :D I did scan and probably stopped after the middle of the second row.

ChristophWurst:
I guess you just scanned without active thinking and then your brain was like "hey, I have this idea..." :wink:

PR #1476, opened 2019-01-17
Update package-lock.json

jancborchardt:
This keeps being annoying on changing branches. :D Please check @nextcloud/mail if this is correct.


ChristophWurst:
@jancborchardt https://github.com/dependabot/feedback/issues/197#issuecomment-456401703 will make you happy :)

PR #1580, opened 2019-03-06
Fix linting of service modules

ChristophWurst:
Partly applying https://github.com/nextcloud/mail/pull/1405 to keep things reviewable


ChristophWurst:
the diff does not seem right

PR #1581, opened 2019-03-06
Fix linting of store modules

ChristophWurst:
Partly applying #1405 to keep things reviewable


ChristophWurst:
doesn't seem right

PR #1647, opened 2019-04-05
Add monkey patch to add an #app-navigation-toggle if it's missing

ChristophWurst:
The element *should* exist, but it doesn't in all cases.

Ref https://github.com/nextcloud/server/issues/14956 https://github.com/nextcloud/server/pull/14972

This doesn't actually work as the click handlers (which delegate to `Snap`) are missing.


ChristophWurst:
Fixed with https://github.com/nextcloud/nextcloud-vue/pull/339

PR #1669, opened 2019-04-10
Replace the legacy appstore target in Makefile with the Krankerl command

ChristophWurst:
Nobody should use this, hence removing it.


ChristophWurst:
To any packager stumbling over this: you can find the tool here: https://github.com/ChristophWurst/krankerl

PR #1671, opened 2019-04-10
Rearrange Travis CI jobs

ChristophWurst:
Goal: small and simple jobs that run fast

Considerations:
* Split jobs into *test suites*
* Do minimal setup only when really needed
* Remove outdated/broken nightly push to FTP


ChristophWurst:
One day I will migrate to https://docs.travis-ci.com/user/build-stages

PR #1673, opened 2019-04-10
Fix app navigation toggle with an update Vue components lib

ChristophWurst:
- [ ] Requires the next nextcloud-vue release


ChristophWurst:
-> merged into https://github.com/nextcloud/mail/pull/1694

PR #1705, opened 2019-04-19
Make it possible to create subfolders and show folder stats

ChristophWurst:
This adds a convenient way of adding subfolders to top-level folders without having to know the delimiter.

I had to find a secondary action to prevent the vue components from rending a single action and thus making the *add folder* form unusable. Therefore we now show some simple folder stats:

Bildschirmfoto von 2019-04-19 10-14-43

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -54,6 +54,11 @@ return [
 			'verb' => 'GET'
 		],
 		[
+			'name' => 'folders#stats',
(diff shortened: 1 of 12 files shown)


ChristophWurst:
> I had to find a secondary action to prevent the vue components from rending a single action and thus making the _add folder_ form unusable. Therefore we now show some simple folder stats:

Reported at https://github.com/nextcloud/nextcloud-vue/issues/356

PR #1771, opened 2019-05-22
Comply with our npm guidelines

skjnldsv:
Signed-off-by: John Molakvoæ (skjnldsv) <skjnldsv@protonmail.com>


skjnldsv:
@ChristophWurst I checked your krankerl settings, but it looked ok :)

Could you double check just in case as you're the sensei here? :pray:

PR #1795, opened 2019-06-04
Feature: print email

### LINKED ISSUE #579: Print stylesheet [$65]

sourenaraya:
Fixes #579


src/components/MessagePrint.vue
```diff
@@ -0,0 +1,47 @@
...
+	},
+	methods: {
+		getIframeDoc() {
+			const iframe = this.$refs.iframe
+			return iframe.contentDocument || iframe.contentWindow.document
+		},
+		replaceBody() {
+			const iframeDoc = this.getIframeDoc()
+			const iframeBody = iframeDoc.querySelectorAll('body')[0]
+			console.log({header: this.header, body: this.body})
```
ChristophWurst: debug statement? can we remove this?

sourenaraya: sorry for this, removed.

jancborchardt:
So I was initially confused because using the browser print method (via the menu or Ctrl-P) still printed the rest of the mail.

Only then I saw the "Print" button on the bottom. ;)

As mentioned in the original issue at https://github.com/nextcloud/mail/issues/579 a print stylesheet just hiding elements (like the header, `#app-navigation`, and the reply form) via CSS would be the preferred method here. It is much simpler, also works with the native print function, and we don’t introduce more controls/buttons to which are rarely used.

A good article on how to create a print stylesheet is at Smashing Magazine: https://www.smashingmagazine.com/2018/05/print-stylesheets-in-2018/

sourenaraya:
I tried this aproach, it worked fine for plaintext emails, but i dont think its possible to print HTML emails that way. The problem is, if height of email is higher than screen height, its not being printed correctly, because HTML emails is in iframe. Setting height of iframe and all parent elements to 100% doesnt help. Not sure if its possible to implement CSS only solution, but will try again.

sourenaraya:
Ok, i did some research. 
There is no CSS-only solution.
Using `beforeprint` event we could adjust size of iframe to its content height and then `afterprint` to remove explicit size, but safari doesnt support that, so we have to use polyfill.

jancborchardt:
Ah ok – good research @sourenaraya! So with the polyfill would it be possible to also catch the native browser print and Ctrl-P? Cause that would be the preferred method over a dedicated "Print" button. :)

sourenaraya:
i think so. Working on it. Do i just force-push new changes, or better close this PR and open a new one?

jancborchardt:
@sourenaraya cool, thanks so much! Regarding force push or new pull request – feel free to do whatever is better for you. :)

ChristophWurst:
> i think so. Working on it. Do i just force-push new changes, or better close this PR and open a new one?

https://dev.to/tmr232/dont-amend-fix :rocket:

sourenaraya:
see #1804

PR #1804, opened 2019-06-11
Feature: print email

### LINKED ISSUE #579: Print stylesheet [$65]

sourenaraya:
Fixes #579

new dependency introduced, not sure if that is ok.
Ctrl+P works.


sourenaraya:
any update on review?

package.json
```diff
@@ -31,6 +31,7 @@
     "nextcloud-server": "^0.15.9",
     "nextcloud-vue": "^0.11.4",
     "nextcloud_issuetemplate_builder": "^0.1.0",
+    "printscout": "^2.0.3",
```
ChristophWurst: new dep is fine :+1: but please use `"2.0.3"` instead so another `npm i` does not [accidentally update this package](https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json)

templates/index.php
```diff
@@ -24,6 +24,7 @@
  *
  */
 style('mail', 'mail');
+style('mail', 'print');
```
ChristophWurst: any chance we can inline these styles to the Vue components? We are trying to get the number of requests down in general

sourenaraya: ok, moved styles to `Message.vue`

package.json
sourenaraya: fixed, thanks

ChristophWurst:
Let's wait for @jancborchardt's review

sourenaraya:
Sorry for asking this again, but, any update on this?

jancborchardt:
Sorry, I was recently on vacation and have to get up to speed again. :) Will check it soon. @sourenaraya just for clarification: There’s no interface change to check for, only Ctrl-P and browser menu print functionality, correct?

sourenaraya:
Yep, just Ctrl+P as usual.

jancborchardt:
Works very nicely, good job @sourenaraya! Only details:
- In Firefox: If you print via the menu → Print functionality you get a preview. If you scroll there, the `mail-message-header` (with subject, sender and recipient) is still sticky so floating above the text. Maybe it should be changed to `position: relative;` for printing? It does come out fine in the PDF
- Also in Firefox/Ubuntu, when printing to PDF, is it possible to preset the file name to something, like the mail title? For me the filename is always the same as last time unless I rename.
- In the printout, the date of the email is not included. Should we have this? Might be something for a follow-up too.

In general nothing blocking and we could also merge it like this :+1:  :)

sourenaraya:
i've set `position: relative` to `mail-message-header`, thanks.

* print-to-pdf function: AFAK there is no way to suggest filename to browser. Its always `mozilla.pdf`, or something set in `about:config`
* date in the printout:  date is not present on plaintext, as well as on richtext emails, anywhere. First of all, we have to somehow c it in backend responses, or have a way to retrieve date information from backend. If there is any, i am not aware of it.

ChristophWurst:
@sourenaraya please don't forget to go to https://www.bountysource.com/issues/50806693-print-stylesheet and claim your bounty :)

sourenaraya:
Issue still marked as open on the bountysource 😂

ChristophWurst:
Ugh. Bountysource is so buggy. Sorry about that. Tried to get support via https://twitter.com/ChristophWurst/status/1159344197436104704. Let's see :pray:

PR #1817, opened 2019-06-18
Use nextcloud-l10n for translations

ChristophWurst:
Ref https://github.com/ChristophWurst/nextcloud-l10n


src/main.js
```diff
@@ -25,7 +25,7 @@ import App from './App'
 import router from './router'
 import store from './store'
 import {sync} from 'vuex-router-sync'
-import {translate, translatePlural} from 'nextcloud-server/dist/l10n'
+import {translate, translatePlural} from 'nextcloud-l10n'
```
georgehrke: Would probably be good to keep things consistent and use `{translate as t, translatePlural as n}` as well. :)
(But that can also be changed later on)

ChristophWurst: Yep, it's alias'ed below, though.

ChristophWurst: Added a note to https://github.com/ChristophWurst/nextcloud-l10n#usage for why I'm actually doing this :)

PR #1950, opened 2019-08-28
Fix next/prev message navigation when non-open message is deleted

ChristophWurst:
## Steps to reproduce

* Open a message
* Delete another one

***Expected***

Current message stays open

***Actual (before)***

App navigates to previous/next message of the deleted one

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -124,6 +124,10 @@ export default {
 				Logger.debug('envelope to delete does not exist in envelope list')
 				return
 			}
+			if (envelope.uid !== this.$route.params.messageUid) {
+				Logger.debug('other message open, not jumping to the next/previous message')
+				return
+			}
 
 			let next
 			if (idx === 0) {


StCyr:
Hi @ChristophWurst,

Unfortunately this fix doesn't work.

Either your description of the bug is not correct or your fix changes the way the app reacts because, after applying this patch, **the app navigates to the deleted message** not the previous/next one (so, though the envelope is removed from the list, the content of the deleted message is displayed). 

I think it's your description of the bug that's wrong (that's: the app works like this event without your fix) but I don't have the time to confirm this right now.

Here's the console's log when performing the test:

```
other message open, not jumping to the next/previous message EnvelopeList.vue:125
message removed actions.js:533
```
​

ChristophWurst:
> the app navigates to the deleted message

well observed. this is actually fixed in https://github.com/nextcloud/mail/pull/1949. I developed this fix locally on the other branch, but kept the features separate. if you combine the two fixes it should work as expected ;)

StCyr:
@ChristophWurst 

Yes, it works! 

#1950 + #1949 solve the issue 👍

PR #1956, opened 2019-08-28
Add a database cache for folders

ChristophWurst:
This drastically* speeds up page loads, especially with multi account setups.

Todo

- [x] Store time of last sync
- [x] Only sync folders each x hours
- [x] Update and write tests

*on my two-account test it took an avg of 8s to load a full page (showing a message as well) without this cache and an avg of 4.6 with.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -5,7 +5,7 @@
 	<name>Mail</name>
 	<summary>IMAP web client</summary>
 	<description>Easy to use email client which connects to your mail server via IMAP and SMTP.</description>
-	<version>0.15.5</version>
+	<version>0.15.6</version>
 	<licence>agpl</licence>
 	<author>Christoph Wurst</author>
 	<author>Jan-Christoph Borchardt</author>
--- a/lib/Db/MailAccount.php
+++ b/lib/Db/MailAccount.php
@@ -59,6 +59,8 @@ use OCP\AppFramework\Db\Entity;
  * @method void setOutboundPassword(string $outboundPassword)
  * @method string|null getSignature()
  * @method void setSignature(string|null $signature)
+ * @method int getLastMailboxSync()
+ * @method void setLastMailboxSync(int $time)
  */
 class MailAccount extends Entity {
 
@@ -76,6 +78,7 @@ class MailAccount extends Entity {
 	protected $outboundUser;
 	protected $outboundPassword;
 	protected $signature;
+	protected $lastMailboxSync;
 
 	/**
 	 * @param array $params
@@ -123,6 +126,8 @@ class MailAccount extends Entity {
 		if (isset($params['smtpPassword'])) {
 			$this->setOutboundPassword($params['smtpPassword']);
 		}
+
+		$this->addType('lastMailboxSync', 'integer');
 	}
 
 	/**
--- a/lib/Db/Mailbox.php
+++ b/lib/Db/Mailbox.php
@@ -1,0 +1,76 @@
+<?php declare(strict_types=1);
+
+/**
+ * @copyright 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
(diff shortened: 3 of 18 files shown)


appinfo/info.xml
```diff
@@ -5,7 +5,7 @@
 	<name>Mail</name>
 	<summary>IMAP web client</summary>
 	<description>Easy to use email client which connects to your mail server via IMAP and SMTP.</description>
-	<version>0.15.5</version>
+	<version>0.15.6</version>
```
rullzer: O come on... that is worth at least a 0.16

ChristophWurst: well, sure ;)

ChristophWurst: it's a pity we have to include the version in the migration file name. makes it a bit hard to define what the next version will be

lib/Db/Mailbox.php
```diff
@@ -0,0 +1,76 @@
+<?php declare(strict_types=1);
```
rullzer: newline or?

lib/Db/MailAccount.php
```diff
@@ -123,6 +126,8 @@ public function __construct(array $params=[]) {
 		if (isset($params['smtpPassword'])) {
 			$this->setOutboundPassword($params['smtpPassword']);
 		}
+
+		$this->addType('lastMailboxSync', 'integer');
```
rullzer: Is there a reason you only set it for this one explicitly?

rullzer: Come tot hink of it we should make this use the declarations docs properly I guess (but for server fixing that is)

lib/IMAP/MailboxSync.php
```diff
@@ -0,0 +1,141 @@
+<?php declare(strict_types=1);
```
rullzer: newline?

lib/IMAP/MailboxSync.php
```diff
@@ -0,0 +1,141 @@
...
+		$this->mailboxMapper = $mailboxMapper;
+		$this->folderMapper = $folderMapper;
+		$this->mailAccountMapper = $mailAccountMapper;
+		$this->imapClientFactory = $imapClientFactory;
+		$this->timeFactory = $timeFactory;
+		$this->logger = $logger;
+	}
+
+	public function sync(Account $account): void {
+		if ($account->getMailAccount()->getLastMailboxSync() >= ($this->timeFactory->getTime() - 7200)) {
```
rullzer: I assume it is future work to be able to force a sync? Because otherwise I can already see the bug reports come in :wink:

lib/Db/MailAccount.php
ChristophWurst: It's the only one I touch in this PR. The others are all strings and so should just with the default.

lib/IMAP/MailboxSync.php
ChristophWurst: Yep, I could imagine something like an entry in the account's dropdown with "refresh folder list". But for 99.9% of all cases it's relatively rare that folders change, I guess

lib/IMAP/MailboxSync.php
ChristophWurst: this was intentional, as explained above

lib/Db/Mailbox.php
ChristophWurst: Saw @icewind1991 using this style and I kinda like it. objections?

rullzer: Ah no fine by me. Just didn't see it like this before ;)

PriceChild:
This is brilliant! Makes the app so much more usable. Thanks @ChristophWurst

ChristophWurst:
Thank you @PriceChild! I really appreciate the feedback!

ChristophWurst:
We also appreciate ratings on https://apps.nextcloud.com/apps/mail :)

PR #1964, opened 2019-08-29
Show a warning if you are typing a message without a subject

rullzer:
Noticed this today when sending a mail.
I'm not 100% sure how I feel about blocking. But such a warning should be good enough.

It only pops up once you start typing in the body.

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -83,6 +83,9 @@
 				@keyup="onInputChanged"
 			/>
 		</div>
+		<div v-if="noSubject" class="warning">
+			{{ t('mail', 'This mail does not have a subject yet.') }}
+		</div>
 		<div v-if="noReply" class="warning noreply-box">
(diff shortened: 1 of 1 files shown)


src/components/Composer.vue
```diff
@@ -83,6 +83,9 @@
 				@keyup="onInputChanged"
 			/>
 		</div>
+		<div v-if="noSubject" class="warning">
+			{{ t('mail', 'This mail does not yet have a subject.') }}
```
```suggestion
			{{ t('mail', 'This mail does not have a subject yet.') }}
```

Not 100% sure which one is the correct one but this sounds a tad better

src/components/Composer.vue
```diff
@@ -223,6 +226,9 @@ export default {
 		isReply() {
 			return !_.isUndefined(this.replyTo)
 		},
+		noSubject() {
+			return this.subjectVal == '' && this.bodyVal != ''
```
```suggestion
			return this.subjectVal === '' && this.bodyVal !== ''
```

PR #1967, opened 2019-08-30
Enh/user server response caching

rullzer:
Todo:
- [x] Update tests
- [x] Probably kill of the JSON response (or just keep it so if you ever want to extend it you can?)
- [x] Kill of the cache trait


ChristophWurst:
>     * Probably kill of the JSON response (or just keep it so if you ever want to extend it you can?)

Fine by me if all functionality is provided by the base class meanwhile

rullzer:
Good to review :)

PR #1968, opened 2019-08-30
Clean up the code responsible for deleting a message

ChristophWurst:
Basically getting rid of Account::deleteMessage and a lot of duplicated code inside the Account class


lib/AppInfo/Application.php
```diff
@@ -38,6 +40,7 @@
 use OCA\Mail\Service\MailTransmission;
 use OCA\Mail\Service\UserPreferenceSevice;
 use OCP\AppFramework\App;
+use OCP\EventDispatcher\IEventDispatcher;
```
rullzer: Fine by me but then this becomes 17 only

ChristophWurst: Good point. 0.17.0 seems like a good last release for 16. Will increase the min version in another PR.

PR #1970, opened 2019-08-30
Persist the mailbox special use as well

ChristophWurst:
Saves some unnecessary operations

- [x] Requires https://github.com/nextcloud/mail/pull/1971
- [x] Update migration name to run after #1971

### DIFF
--- a/lib/IMAP/MailboxSync.php
+++ b/lib/IMAP/MailboxSync.php
@@ -23,6 +23,7 @@
 
 namespace OCA\Mail\IMAP;
 
+use function json_encode;
 use OCA\Mail\Account;
 use OCA\Mail\Db\MailAccountMapper;
 use OCA\Mail\Db\Mailbox;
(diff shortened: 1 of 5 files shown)


ChristophWurst:
Actual commit to review: https://github.com/nextcloud/mail/pull/1970/commits/7cbc0e5019fbc631613bb3f5468b2545e2c062db

lib/IMAP/MailboxSync.php
```diff
@@ -120,19 +121,21 @@ private function updateMailboxFromFolder(Folder $folder, Mailbox $mailbox): void
...
-		$mailbox->setId($folder->getMailbox());
+		$mailbox->setName($folder->getMailbox());
 		$mailbox->setAccountId($account->getId());
 		$mailbox->setSyncToken($folder->getSyncToken());
 		$mailbox->setAttributes(json_encode($folder->getAttributes()));
 		$mailbox->setDelimiter($folder->getDelimiter());
 		$mailbox->setMessages(0); // TODO
 		$mailbox->setUnseen(0); // TODO
 		$mailbox->setSelectable($folder->isSelectable());
+		$mailbox->setSpecialUse(json_encode($folder->getSpecialUse()));
```
rullzer: Do we know beforehand what the special uses can be? As in how long this final string can get?

ChristophWurst: A combination of https://github.com/nextcloud/mail/blob/f40aba36d6356acb23cd0fb6aa9e6584f69fdca3/lib/IMAP/FolderMapper.php#L180-L186, which are all ~4-6 chars long. And usually a folder has just one of these special uses.

The max length shouldn't be a problem.

PR #1971, opened 2019-08-30
Move away from string-based ID column as that causes problems on update

### LINKED ISSUE #1969: Mail not working after update to 0.16.0, exclamation marks left to account names
### Expected behavior
Two IMAP accounts were setup up with a previous version. They worked without a problem. 

### Actual behavior
All accounts won't work, instead of the mail folders only an exclamation mark is shown left to the account name. 
There is no output in the nextcloud log file

### Mail app

**Mail app version:** 0.16.0

**Mailserver or service:** Selft hosted, strato.de

### Server configuration

**Operating system**: Ubuntu 19.04 Host, official Docker image

**Web server:** Apache 2.4.38 as Proxy, official Docker image

**Database:** MariaDB 17.4 (docker)

**PHP version:** Official Docker Image

**Nextcloud Version:** 16.0.4, running in official Docker image

#### Client configuration
**Browser:** Firefox 68.0.2

**Operating system:** Ubuntu 19.04

ChristophWurst:
The QBMapper will automatically use int as column type for the `id` columns, hence an update on the old db schema failed with an sql error.

I've run into this before and yet I was stupid enough to make the same mistake again. Anyway, should be fine now. `name` is the new `id` and `id` is a simple auto increment column.

Fixes #1969

### DIFF
--- a/lib/Db/Mailbox.php
+++ b/lib/Db/Mailbox.php
@@ -27,12 +27,12 @@ use OCA\Mail\Folder;
 use OCP\AppFramework\Db\Entity;
 
 /**
- * @method string getId()
- * @method void setId(string $id)
+ * @method string getName()
+ * @method void setName(string $name)
  * @method int getAccountId()
  * @method void setAccountId(int $accountId)
- * @method string getSyncToken()
- * @method void setSyncToken(string $syncToken)
+ * @method string|null getSyncToken()
+ * @method void setSyncToken(string|null $syncToken)
  * @method string getAttributes()
  * @method void setAttributes(string $attributes)
  * @method string getDelimiter()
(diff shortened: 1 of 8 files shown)


ChristophWurst:
[mail.tar.gz](https://github.com/nextcloud/mail/files/3560289/mail.tar.gz)

ChristophWurst:
This bug also prevented me from updating data for https://github.com/nextcloud/mail/pull/1970. Once this branch got merged into the other feature branch the error disappeared.

lib/Db/Mailbox.php
```diff
@@ -29,6 +29,8 @@
 /**
  * @method string getId()
```
rullzer: This should also be updated

lib/Db/Mailbox.php
```diff
@@ -29,6 +29,8 @@
...
  * @method void setId(string $id)
```
rullzer: and this

Issue #1982, opened 2019-09-01
Can't recreate account (mailbox sql viloation)


sheychen290:
### Expected behavior
After updating to 0.16.0, most of my accounts were broken (exclamation mark)  without error.
When recreating one (delete then add new), client do back to main view but account isn't visible (even after hard refresh)

Logs contains an SQL error on oc_mail_mailboxes
```
[mail] Error: Doctrine\DBAL\Exception\NotNullConstraintViolationException: An exception occurred while executing 'INSERT INTO `oc_mail_mailboxes` (`id`, `account_id`, `attributes`, `delimiter`, `messages`, `unseen`, `selectable`) VALUES(?, ?, ?, ?, ?, ?, ?)' with params ["dovecot", 16, "[\"\\\\nonexistent\",\"\\\\noselect\"]", ".", 0, 0, false]:

SQLSTATE[HY000]: General error: 1364 Field 'sync_token' doesn't have a default value at <<closure>>

 0. /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php line 184
    Doctrine\DBAL\Driver\AbstractMySQLDriver->convertException("An exception oc ... e", Doctrine\DBAL\Dr ... ]})
 1. /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php line 158
    Doctrine\DBAL\DBALException::wrapException(Doctrine\DBAL\Driver\PDOMySql\Driver {}, Doctrine\DBAL\Dr ... ]}, "An exception oc ... e")
 2. /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/Connection.php line 1088
    Doctrine\DBAL\DBALException::driverExceptionDuringQuery(Doctrine\DBAL\Driver\PDOMySql\Driver {}, Doctrine\DBAL\Dr ... ]}, "INSERT INTO `oc ... )", {1: "dovecot",2: ... e})
 3. /var/www/html/lib/private/DB/Connection.php line 216
    Doctrine\DBAL\Connection->executeUpdate("INSERT INTO `oc ... )", ["dovecot",16,"[ ... e], [2,1,2,2,1,1,5])
 4. /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/Query/QueryBuilder.php line 222
[...]

PR #1983, opened 2019-09-02
Allow null values for a mailbox's sync token

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/1982

- [ ] Requires #1971


ChristophWurst:
superseded by #1971

PR #1985, opened 2019-09-02
Move away from string-based ID column as that causes problems on update

ChristophWurst:
Replaces https://github.com/nextcloud/mail/pull/1971 and hopefully triggers Travis here.


ChristophWurst:
.travis markup is wrong :see_no_evil:

PR #1986, opened 2019-09-03
Require Nextcloud 17 or 18

ChristophWurst:


thomas-mc-work:
@ChristophWurst Is that really required? I think this is a very rough decision for end users.

ChristophWurst:
Yes, really. We use an API that is only available in recent versions of Nextcloud. You can still use the app on older releases. This was a very pragmatic, technical decision.

See https://github.com/nextcloud/mail/issues/2195 for details.

PR #1991, opened 2019-09-04
Do not use lodash for undefined checks

ChristophWurst:


skjnldsv:
Any reason you're checking directly for undefined instead of just globally checking for falsy values?

ChristophWurst:
> Any reason you're checking directly for undefined instead of just globally checking for falsy values?

It's more explicit and IMO less prone to errors.

PR #1997, opened 2019-09-05
Refactor send message and save draft logic

ChristophWurst:
* Move code from `Account` class to the `MailTransmission` service
* Remove lots of duplicated code from `Account`
* Use the IMAP client factory consistently
* Decouple the two processes with events. Essential logic stays in `MailTransmission`, anything else is handled elsewhere
  * Add a `SaveDraftEvent`
  * Add a `DraftSavedEvent`
  * Add a `MessageSentEvent`
  * Add a listener to save the sent message on IMAP
  * Add a listener that ensures a draft mailbox exists before we save
  * Add a listener to delete optional draft when sending a message or saving a new draft
  * Add a listener that flags a replied message as replied
  * Add a listener that collect email addresses of sent messages
 * Add a `BootstrapSingleton` to ensure initialization code is only run twice (in some situations two `Applications` classes are created, causing duplicate events)

While this got a lot bigger than I expected, it will be necessary to allow us adding more complexity to these processes with future features. Moreover, most of the code is now unit-testable.


ChristophWurst:
To my reviews: I'm sorry if this PR is large. I swear I tried to keep it small but darn this was a huge heap of spaghetti code.

When in doubt: just approve :see_no_evil: :speak_no_evil:

ChristophWurst:
Also CI is *very* happy, the code metrics look [great](https://scrutinizer-ci.com/g/nextcloud/mail/inspections/ca2e0ebd-4271-40e4-83b6-d768df686e50) and everything's still covered by unit and integration tests.

Of course I also manually tested and things worked just like they did before.

lib/Events/MessageSentEvent.php
```diff
@@ -0,0 +1,92 @@
+<?php declare(strict_types=1);
```
kesselb: I know nextcloud does not follow psr-2 (and not psr-12) but would be nice ;) https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements

lib/Listener/AddressCollectionListener.php
```diff
@@ -0,0 +1,68 @@
...
+	private $logger;
+
+	public function __construct(AddressCollector $collector,
+								ILogger $logger) {
+		$this->collector = $collector;
+		$this->logger = $logger;
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof MessageSentEvent)) {
```
kesselb: How is that possible? https://github.com/nextcloud/mail/pull/1997/files#diff-de54363bb086805834528f6f9d0f58e7R118 the event is only listening for `MessageSentEvent`.

lib/Listener/DeleteDraftListener.php
```diff
@@ -0,0 +1,144 @@
...
+								ILogger $logger) {
+		$this->imapClientFactory = $imapClientFactory;
+		$this->mailboxMapper = $mailboxMapper;
+		$this->messageMapper = $messageMapper;
+		$this->mailboxSync = $mailboxSync;
+		$this->logger = $logger;
+	}
+
+	public function handle(Event $event): void {
+		if ($event instanceof DraftSavedEvent && $event->getDraftUid() !== null) {
```
kesselb:
```
if ($event->getDraftUid() === null) {
    return;
}
```

Makes the following code simpler

lib/Listener/DraftMailboxCreatorListener.php
```diff
@@ -0,0 +1,98 @@
...
+								MailboxSync $mailboxSync,
+								ILogger $logger) {
+		$this->mailboxMapper = $mailboxMapper;
+		$this->imapClientFactory = $imapClientFactory;
+		$this->mailboxSync = $mailboxSync;
+		$this->logger = $logger;
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof SaveDraftEvent)) {
```
kesselb: :thinking:

lib/Listener/FlagRepliedMessageListener.php
```diff
@@ -0,0 +1,98 @@
...
+								ILogger $logger) {
+
+		$this->imapClientFactory = $imapClientFactory;
+		$this->mailboxMapper = $mailboxMapper;
+		$this->messageMapper = $mapper;
+		$this->logger = $logger;
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof MessageSentEvent) || !$event->getRepliedMessageData()->isReply()) {
```
kesselb: :thinking:

lib/Events/MessageSentEvent.php
ChristophWurst: Oh yeah. Didn't know such a guideline existed. I won't change it with this PR but will consider that later, possibly with linter support

lib/Listener/AddressCollectionListener.php
ChristophWurst: While that is correct, neither my IDE nor Scrutinizer can statically know that and thus don't like any method invocations except for the base methods of `Event` ;)

lib/Listener/DeleteDraftListener.php
ChristophWurst: Yes but is statically incorrect as `getDraftUid` is only a method of `DraftSavedEvent` and not `Event`

lib/Listener/DraftMailboxCreatorListener.php
ChristophWurst: I'm consistent :P

lib/Listener/FlagRepliedMessageListener.php
ChristophWurst: Yep, even here :)

lib/Listener/DeleteDraftListener.php
```diff
@@ -0,0 +1,144 @@
...
+		$draftsMailbox = $this->getDraftsMailbox($account);
+
+		try {
+			$this->messageMapper->addFlag(
+				$client,
+				$draftsMailbox,
+				$draftUid,
+				Horde_Imap_Client::FLAG_DELETED
+			);
+		} catch (Horde_Imap_Client_Exception_NoSupportExtension|Horde_Imap_Client_Exception $e) {
```
kesselb: https://github.com/horde/Imap_Client/blob/dd6306e2e7aa77b06954c5f580401cca779ebad0/lib/Horde/Imap/Client/Exception/NoSupportExtension.php#L23-L24

`Horde_Imap_Client_Exception_NoSupportExtension` extends `Horde_Imap_Client_Exception`

Issue #2019, opened 2019-09-19
Show envelopes' recipient addresses when in a Send folder rather than sender.


StCyr:
### Feature Request

In the envelopes list, the email address mentioned is always the sender. But, wouldn't it be nice to rather show the recipients addresses?

![image](https://user-images.githubusercontent.com/8179031/65244928-e5dac400-daeb-11e9-8365-51e2ff0e6d91.png)

#### Summary

1. Detect that we are showing the envelopes of a "sent" folder
2. Show recipients addresses (ellipsed when space is short) in such case.

StCyr:
fixed by  #2020

ChristophWurst:
If you add "Fixes #2019" to the PR description it closes the ticket automatically :)

PR #2020, opened 2019-09-19
Show recipients in envelopes' list when displaying a 'sent' folder

StCyr:
Fix to https://github.com/nextcloud/mail/issues/2019

image

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -1,7 +1,7 @@
 <template>
 	<router-link class="app-content-list-item" :class="{unseen: data.flags.unseen, draft}" :to="link">
 		<div
-			v-if="showAccountColor"
+			v-if="folder.isUnified"
 			class="mail-message-account-color"
 			:style="{'background-color': accountColor}"
 		></div>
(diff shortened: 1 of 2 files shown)


src/components/Envelope.vue
```diff
@@ -96,14 +96,26 @@ export default {
...
+		addresses() {
+			if (this.folder.specialRole === 'sent') {
+
+				let recipients = [this.data.to, this.data.cc].flat().map(function(recipient) {
+					return recipient.label ? recipient.label : recipient.email
+				})
+
+				return recipients.length > 0 ? recipients.join(', ') : t('mail', 'Bcc recipients only')
+
+			} else {
```
ChristophWurst: no `else` needed as the `if` will return early

ChristophWurst:
Btw, I think you should check your git config. Github does not associate the commit with your user. The email addresses might not match.

src/components/Envelope.vue
StCyr: sorry, I don't get you: `else` will be executed when not in a 'sent folder' (or at least it's what I'm expecting :-) )

Should I amend all my commits to set the proper email address?

ChristophWurst: it's about simpler flow control:

this
```js
if (something) {
    ...
    return
} else {
   ...
}
```
versus

```js
if (something) {
    ...
    return
}

...
```

-> it's the same thing ;)

> Should I amend all my commits to set the proper email address?

Yes please :)

StCyr:
> 
> 
> Btw, I think you should check your git config. Github does not associate the commit with your user. The email addresses might not match.

I've added this email address to my github's config.

ChristophWurst:
One more lint autofix then we can merge :)

StCyr:
should be ok now

PR #1988, opened 2019-09-03
Add CKEditor as WYSIWYG editor

ChristophWurst:
Fixes #210 #306.

- [x] Blocked by https://github.com/ckeditor/ckeditor5-vue/issues/94

Todo

- [x] Add CKEdit ([balloon editor](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/balloon-editor.html))
- [x] Adjust design
- [x] Fix conflict with keyboard shortcuts
- [x] keyup handler
- [ ] keypress handler
- [x] Make it possible to disable all rich features (text-only
- [x] L10n https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/vuejs.html#ready-to-use-builds
- [x] Limit allowed HTML
- [x] Check loading of HTML drafts
- [x] Add an account setting for the default mode

PR #2039, opened 2019-09-27
Include the HTML body in the JSON message data

ChristophWurst:
Required for #1988


ChristophWurst:
> No css ? cry

[No idea how css works](https://twitter.com/markdalgleish/status/1177398336640274433)

PR #2051, opened 2019-09-30
Fix invalid envelope avatar display name

ChristophWurst:
Fixes a regression from https://github.com/nextcloud/mail/pull/2020 that causes Vue validation errors.


ChristophWurst:
Oh, it shouldn't just be an A. but now that you mentioned it I also only have A's on my dev setup. This is a bug/regression. It should be the first letter from the sender name. And we already have support for avatars from contacts

ChristophWurst:
Merging as this is better than the broken view. PRs welcome to fix the rest :)

PR #2057, opened 2019-10-01
Add missing debug log to \OCA\Mail\IMAP\IMAPClientFactory::getClient

ChristophWurst:


ChristophWurst:
Depends. Actually I'm trying to get away from the operations inside \OCA\Mail\Account, and move towards a more service-oriented design with simple DTOs to pass around. Hence \OCA\Mail\Account::getImapConnection should be removed.

PR #2058, opened 2019-10-01
Refactor imap search

ChristophWurst:


lib/Service/MailSearch.php
```diff
@@ -0,0 +1,152 @@
...
+				]
+			);
+
+			$fetchResult = $client->fetch($mailbox->getMailbox(), $fetchQuery, ['ids' => $ids]);
+		} catch (Horde_Imap_Client_Exception $e) {
+			throw new ServiceException('Could not fetch messages', 0, $e);
+		}
+
+		// TODO: do we still need this fix?
+		ob_start(); // fix for Horde warnings
```
kesselb: Urgh :see_no_evil: Do you know some of these warnings?

ChristophWurst: NONE. That is why I added the `todo` ;)

ChristophWurst: It's from https://github.com/nextcloud/mail/commit/6baf36b844ddd7dfa414515c73bfca3063249631. This could be totally useless by know.

Issue #17, opened 2016-08-31
Grouping of message threads (threading/conversations) [$320]


jancborchardt:
Similar as in Gmail, Apple Mail for iOS, or Facebook private messages or what have you. Read messages in that thread should be compressed & expandable (also like in Gmail and iOS Mail), and inside the thread the sorting should go recent messages on the bottom (because of reading direction, and that’s how comment threads work).

There should be no indentation or crap like that, just timestamp-based sorting. In the future every individual message in the thread could show a »Reply« button on hover to reply to that specific message. But for now I’d just keep it simple like a one-stream comment thread.

For reference:
- [Jamie Zawinski's message threading algorithm](http://www.jwz.org/doc/threading.html)
- ~~[Max Ogden's JS implementation of said algorithm](https://github.com/maxogden/conversationThreading-js)~~
- https://tools.ietf.org/html/rfc5256

![badge](https://api.bountysource.com/badge/issue?issue_id=37463009)

sim6:
Firstly thanks for your work.

Any thread/conversation support is essential to me. I prefer see all related mails on the same page.

jancborchardt:
@sim6 cool you enjoy the app! :) Since you are a developer, would you be interested in contributing a bit? We are an open project and anyone is welcome!

adsworth:
I would love if this thread/conversation view includes all mail regardless of the folder in which it is saved. Outlook (yes I know) does this and I find it really helps me when I get replies in my INBOX. The thread shows all the mails of the conversation regardless of the folder I saved the previous mails to.
The list should also show a link/label of the folder in which the individual mails are.

jancborchardt:
@adsworth yup, agree. Want to contribute to the bounty on this issue? ;) https://www.bountysource.com/issues/37463009-grouping-of-message-threads-threading-conversations cc @sim6 as well.

ChristophWurst:
Just in case anybody wants to work on this: please first write down how you'd implement this. There are many aspects to consider for this feature to work properly. Thanks.

nsrosenqvist:
Any action on this?

ChristophWurst:
No

jancborchardt:
@nsrosenqvist so you are very welcome to contribute! :)

PanderMusubi:
You can add this line at the top of the issue description, it will update itself

    ![badge](https://api.bountysource.com/badge/issue?issue_id=37463009)

and it looks like

[...]

PR #2064, opened 2019-10-02
Add a cache for IMAP message in the database

ChristophWurst:
This is something I've been wanting to do for a long time. So here it is. Or, well, just a few modifications towards a new architecture I would like to have with this app.

### The problem

Right now this app operates on IMAP directly. Hence the performance and capabilities completely depend on the IMAP server and its implementation. This means on many accounts pagination and search is slow because SORT is not supported. We use Horde's cache driver but it doesn't really yield and noticeable performance boost.

### The proposal

In addition to mailboxes #1956 we should keep a copy of message information in the database. Then we can display mailboxes (folders) very quickly, paginate through them and search for strings. We can keep the database sync'ed with Horde's sync mechanism. This means the sync would shift from browser->Nextcloud to Nextcloud->IMAP. That implies that we need a new mechanism for the front-end, but I have some ideas and it should be fairly simple to work that out.

Eventually it will make this app similar to how desktop apps work (in regards to storing data locally).

A rough list of things to do

- [x] Create db table for messages
- [x] Refactor message retrieval code to go through a new service (#2058?), where we can swap our the IMAP connection with a DB query
- [x] Add a mechanism that syncs a mailbox the first time it's accessed
- [x] Add a mechanism to trigger the sync of a mailbox
- [x] Respect UIDVALIDITY
- [x] ~~Don't sync the virtual flagged inbox to disk~~ fine for now. can be improved later

A list of ideas for follow-up features
- Sync mailboxes in cron as well, then you get fast page loads *and* an up to date representation of what's on IMAP
- Have a virtual mailbox for outgoing mail, allowing users to compose message even when IMAP is not available
- Store drafts locally first and only sync to IMAP later. That would eliminate the buggy, unreliable draft logic we have right now (IMAP messages are immutable, we have to replace & get a new ID every time)

### Concerns

Things that might be problematic are

* Huge mailboxes -> may take some time for the initial sync, but IMO we never supported them well and hence not a problem

cc @nextcloud/mail @rullzer for early input/feedback


ChristophWurst:
A feature where we can really see the benefit of this is #17. Not all servers have support for threads and then it gets very expensive to calculate them on demand.

ChristophWurst:
- [ ] BUG: Default accounts have to be persisted to work with background syncs

ChristophWurst:
- [ ] Follow-up: sync'ing large mailboxes from a web request may time out

ChristophWurst:
- [ ] Follow-up: if partial sync dies half way, the next partial sync will run into errors on INSERT

lib/IMAP/Search/Provider.php
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	private function convertMailQueryToHordeQuery(SearchQuery $searchQuery): Horde_Imap_Client_Search_Query {
+		$query = new Horde_Imap_Client_Search_Query();
+
+		foreach ($searchQuery->getFlags() as $flag => $set) {
+			$query->flag($flag, $set);
+		}
+
+		// TODO: text, header text
```
ChristophWurst: - [ ] TODO: fix in a follow-up

lib/Db/MessageMapper.php
```diff
@@ -0,0 +1,372 @@
...
+				 ] as $flag) {
+			if (in_array($flag, $flagKeys, true)) {
+				$key = ltrim($flag, '\\');
+				$select->andWhere($qb->expr()->eq("flag_$key", $qb->createNamedParameter($flags[$flag], IQueryBuilder::PARAM_BOOL)));
+			}
+		}
+
+		$select = $select
+			->orderBy('sent_at', 'desc')
+			->setMaxResults(20);
```
ChristophWurst: - [ ] TODO: is this ok? pagination?

ChristophWurst:
Rebased. There is one showstopper

* Migration to this branch always gives me a `mailboxes` table where `sync_new_token` is populated and thus sync doesn't do anything afterwards until a full sync is performed.

lib/Migration/Version1020Date20191002091034.php
```diff
@@ -0,0 +1,158 @@
...
+		]);
+		$mailboxTable->addColumn('sync_changed_lock', 'integer', [
+			'notnull' => false,
+			'length' => 4,
+		]);
+		$mailboxTable->addColumn('sync_vanished_lock', 'integer', [
+			'notnull' => false,
+			'length' => 4,
+		]);
+		$mailboxTable->dropColumn('sync_token');
```
ChristophWurst: @nickvergessen I'm dropping `sync_token` here and create `sync_new_token`. For some reason the new column is where the old one was

```
+---------------------+--------------+------+-----+---------+----------------+
| Field               | Type         | Null | Key | Default | Extra          |
+---------------------+--------------+------+-----+---------+----------------+
| id                  | int(11)      | NO   | PRI | NULL    | auto_increment |
| name                | varchar(255) | NO   |     | NULL    |                |
| account_id          | int(11)      | NO   | MUL | NULL    |                |
| sync_new_token      | varchar(255) | YES  |     | NULL    |                |
| attributes          | varchar(255) | NO   |     | []      |                |
| delimiter           | varchar(1)   | NO   |     | NULL    |                |
| messages            | int(11)      | NO   |     | NULL    |                |
| unseen              | int(11)      | NO   |     | NULL    |                |
| selectable          | tinyint(1)   | NO   |     | 1       |                |
| special_use         | varchar(255) | NO   |     | []      |                |
| sync_new_lock       | int(11)      | YES  |     | NULL    |                |
| sync_changed_lock   | int(11)      | YES  |     | NULL    |                |
| sync_vanished_lock  | int(11)      | YES  |     | NULL    |                |
| sync_changed_token  | varchar(255) | YES  |     | NULL    |                |
| sync_vanished_token | varchar(255) | YES  |     | NULL    |                |
+---------------------+--------------+------+-----+---------+----------------+
```

Any possible explanation for that? I would expect to find the new col near the end of the list where the other new attributes are :thinking:

ChristophWurst: (sync_new_token is also populated with values and I don't yet know where they are coming from. they are valid tokens. but I can't find the code path where we set only that one token)

nickvergessen: Well the order should not matter anyway?
But content being there already sounds strange....

ChristophWurst: I don't care about the order. I'm just wondering as I have problems with the values.

RobertZenz: @ChristophWurst There's a good chance that Doctrine detects a rename here.

RobertZenz: You sure Doctrine does not simply detect a rename here?

ChristophWurst: Yes, same conclusion @rullzer and I came to. Looks like it's trying to be smart and just renames the dropped column to one of the created once. Thus the data. Smells like https://github.com/doctrine/dbal/issues/3661 to me. The workaround is to split the migration into two, where the first one drops and the second creates. Seems to do the trick.

ChristophWurst:
- [ ] BUG: Attachment indicator missing in message list

lib/Migration/Version1020Date20191002091034.php
nickvergessen: That's quite a shocking change in their API. I guess we need to check all Migrations for this case.

ChristophWurst: If you want to have a closer look you can use commit https://github.com/nextcloud/mail/commit/72b15d77e4d573faa6d6f4e0ede23f5d6f169c17. Just set up master and add an account. Then switch to that commit and run the migration.

ChristophWurst:
Test release 1 [mail.tar.gz](https://github.com/nextcloud/mail/files/4133364/mail.tar.gz)

lib/Db/MessageMapper.php
ChristophWurst: Pagination works fine

lib/Db/MailAccountMapper.php
```diff
@@ -60,7 +60,7 @@ public function find(string $userId, int $accountId): MailAccount {
 	}
 
 	/**
-	 * @param int $id
```
nickvergessen, line 63: opsy?

PR #2071, opened 2019-10-03
Changed CC/BCC to uppercase

rakekniven:
Reported at Transifex

Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>


rakekniven:
While reading RFCs and other sources I found https://de.wikipedia.org/wiki/Header_(E-Mail)#BCC very good.

PR #2072, opened 2019-10-03
Use lodash in a functional fashion

ChristophWurst:
Checking if the bundle analyzer works …


rullzer:
This is done now in smaller steps.

PR #2089, opened 2019-10-07
fix french typography in l10n

bastiengrb:
Using your application, I noticed a small spelling error. Here is a correction.

I hope it will help you, keep going.


ChristophWurst:
Hi,

translations are handled via Transifex. Please submit the fix at https://www.transifex.com/nextcloud/nextcloud/dashboard/ and it will be synced back to this repository :)

Cheers

PR #2090, opened 2019-10-08
Do not fetch headers if we don't use them

ChristophWurst:
Debugging performance problems of #2064 I found out that we do a lot of over-fetching with our IMAP queries. Then I tried to gradually remove stuff we fetch but obviously don't use and it turns out that some expensive queries can take 85% less time if headers aren't fetched.

I could not find any problems in my manual tests. @nextcloud/mail please verify :)

### DIFF
--- a/lib/Attachment.php
+++ b/lib/Attachment.php
@@ -62,22 +62,9 @@ class Attachment {
 	private $mimePart;
 
 	private function load() {
-		$headers = [];
-
 		$fetch_query = new \Horde_Imap_Client_Fetch_Query();
 		$fetch_query->bodyPart($this->attachmentId);
 		$fetch_query->mimeHeader($this->attachmentId);
-
-		$headers = array_merge($headers, [
-			'importance',
-			'list-post',
-			'x-priority'
-		]);
-		$headers[] = 'content-type';
-
-		$fetch_query->headers('imp', $headers, [
-			'cache' => true
-		]);
 
 		// $list is an array of Horde_Imap_Client_Data_Fetch objects.
 		$ids = new \Horde_Imap_Client_Ids($this->messageId);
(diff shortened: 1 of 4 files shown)


StCyr:
'Copied the 4 changed files on my NC: mail cannot fetch messages anymore (hangs on "Loading messages").

I've got these messages in my log:

`Class OCA\Mail\Model\IMAPMessage contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (OCA\Mail\Model\IMessage::getRepliedMessage, OCA\Mail\Model\IMessage::setRepliedMessage) at /var/www/nextcloud/apps/mail/lib/Model/IMAPMessage.php#54`

And this in my apache's error log:

`[Tue Oct 08 15:56:58.632627 2019] [proxy_fcgi:error] [pid 2411:tid 140304228902656] [client 94.224.26.219:50770] AH01071: Got error 'PHP message: PHP Fatal error:  Class OCA\\Mail\\Model\\IMAPMessage contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (OCA\\Mail\\Model\\IMessage::getRepliedMessage, OCA\\Mail\\Model\\IMessage::setRepliedMessage) in /var/www/nextcloud/apps/mail/lib/Model/IMAPMessage.php on line 54'`

Using NC 16.0.5

ChristophWurst:
Hi,

thanks for testing. You can't just copy the files as then you'll miss some other recent changes. Here's a snapshot you can test with: [mail.tar.gz](https://github.com/nextcloud/mail/files/3703036/mail.tar.gz)

StCyr:
> Here's a snapshot you can test with: [mail.tar.gz](https://github.com/nextcloud/mail/files/3703036/mail.tar.gz)

Server version 17 or higher is required :-)

upgrading

ChristophWurst:
> UPDATE: Ok, it's working now (NC 17 and your mail.tar.gz package). Didn't notice speed improvement yet

Great! Yep, with the existing code there won't be a noticeable difference as we do relatively small queries on IMAP. But in #2064 I get the full mailbox at once, where this makes a huge difference.

PR #2111, opened 2019-10-16
Remove getAccount from MessagesController

rullzer:


rullzer:
I made the find function return a proper exception and fixed the tests.
@ChristophWurst have a quick look please

PR #2117, opened 2019-10-16
Remove jquery usage and replace with Axios

rullzer:


skjnldsv:
Jsunit fails

PR #2118, opened 2019-10-16
[WIP] Extend list of international re and fwd strings

### LINKED ISSUE #346: Prevent stacking of Re: and Fwd: also for other languages

jakobsack:
Data taken from https://en.wikipedia.org/wiki/List_of_email_subject_abbreviations#Abbreviations_in_other_languages

fixes #346


jancborchardt:
Seems good from what I can tell. But in right-to-left languages, would it be prepended? Or appended? cc @mabkenar for insight :)

PR #2121, opened 2019-10-16
Properly show resultst of stats

skjnldsv:
Capture d’écran_2019-10-17_00-01-35

### DIFF
--- a/src/components/NavigationFolder.vue
+++ b/src/components/NavigationFolder.vue
@@ -25,14 +25,14 @@
 		:key="genId(folder)"
 		:allow-collapse="true"
 		:icon="icon"
-		:menu-open.sync="menuOpen"
 		:title="title"
 		:to="to"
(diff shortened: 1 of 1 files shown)


jancborchardt:
Some thoughts:
- The info i and the folder name is actually quite repetitive from the info directly on the row.
- The "Mark all as read" entry below is quite useless if all are read already.
- The subline of "Mark all as read" is pretty redundant of the main row.
- How about we combine the "Number of messages" detail row and the main text of "Mark all as read" into one row?

Like so:

> **Mark all as read**
> 2 unread of 120

When it’s all read:

> **All read**
> 120 messages

Reduces the amount of info severely and groups it in a relevant way.

skjnldsv:
Jan, wanna fix tis please :)

jancborchardt:
Will try! :D

PR #2123, opened 2019-10-16
Design details

jancborchardt:
- Fix text size in empty folder message
- Don’t show unread count when it’s 0
- Reduce presence of 'Show all folders'/'Collapse' entry 

Check it @ChristophWurst @skjnldsv


skjnldsv:
Tests fails

PR #2125, opened 2019-10-17
Build message threads after each sync and store thread root ID for later reference

ChristophWurst:
This makes use of the jwz algorithm added via another PR. This focuses on the integration part, not the threading algorithm. So if any thread is assigned the wrong ID -> I don't care here as long as *some* thread root ID will be assigned :P

- [x] Requires #3337 to load real data and write back the updated root IDs
- [x] Requires https://github.com/nextcloud/mail/pull/3355

To test this:
* Switch wo this branch
* Use mail as you normally would

Expected outcome:
* Don't notice any difference
* The `thread_root_id` of your `oc_mail_messages` will slowly populate with values

Ref #17 (not fully done yet)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -32,6 +32,7 @@ use OCA\Mail\Contracts\IMailTransmission;
 use OCA\Mail\Contracts\IUserPreferences;
 use OCA\Mail\Events\BeforeMessageDeletedEvent;
 use OCA\Mail\Events\DraftSavedEvent;
+use OCA\Mail\Events\SynchronizationEvent;
 use OCA\Mail\Events\MessageDeletedEvent;
 use OCA\Mail\Events\MessageFlaggedEvent;
 use OCA\Mail\Events\MessageSentEvent;
@@ -44,6 +45,7 @@ use OCA\Mail\Listener\DeleteDraftListener;
 use OCA\Mail\Listener\DraftMailboxCreatorListener;
 use OCA\Mail\Listener\FlagRepliedMessageListener;
 use OCA\Mail\Listener\InteractionListener;
+use OCA\Mail\Listener\AccountSynchronizedThreadUpdaterListener;
 use OCA\Mail\Listener\MessageCacheUpdaterListener;
 use OCA\Mail\Listener\NewMessageClassificationListener;
 use OCA\Mail\Listener\SaveSentMessageListener;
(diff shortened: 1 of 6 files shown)


ChristophWurst:
Note to self

```
Thread roots

select id,uid,message_id,subject from oc_mail_messages where mailbox_id=747 and (parent_message_id ='' or parent_message_id is null);

Thread roots without dangling pointers

select id,uid,message_id,subject from oc_mail_messages m1 where mailbox_id=747 and ((parent_message_id = '' or parent_message_id is null) or parent_message_id not in (select distinct message_id from oc_mail_messages));
```

ChristophWurst:
We'll finish https://github.com/nextcloud/mail/projects/9 first. There are a few quirks with that feature and adding more complexity to the app right now will make debugging a nightmare. But threading will come soon :)

StCyr:
Hello, 

may I help on something here?

I've tried this to get the `In-Reply-To` header in the frontend (and it works):

```php
root@vps612142:/var/www/nextcloud/apps/mail# git diff
diff --git a/lib/Model/IMAPMessage.php b/lib/Model/IMAPMessage.php
index 180ea30d..c18c63b6 100644
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -424,6 +424,7 @@ class IMAPMessage implements IMessage, JsonSerializable {
                }
 
                $data['attachments'] = $this->attachments;
+               $data['inReplyTo'] = $this->getInReplyTo();
 
                return $data;
        }
@@ -610,10 +611,10 @@ class IMAPMessage implements IMessage, JsonSerializable {
        }
 
        /**
-        * @return IMessage|null
+        * @return string
         */
-       public function getInReplyTo() {
-               throw new Exception('not implemented');
+       public function getInReplyTo(): string {
+               return $this->getEnvelope()->in_reply_to;
        }
 
        /**
```

ChristophWurst:
I have to revisit the db design. We might want to cleverly store the id (like ours, not message id) of the parent id, so building the threads can happen fast and incrementally.

lib/Db/Message.php
```diff
@@ -134,6 +135,20 @@ public function __construct() {
 		$this->addType('updatedAt', 'integer');
 	}
 
+	public function setMessageId(?string $id): void {
+		if (empty($id)) {
+			parent::setter('messageId', [null]);
```
nickvergessen, line 140: :boom:

lib/Db/Message.php
```diff
@@ -134,6 +135,20 @@ public function __construct() {
...
+		}
+		parent::setter('messageId', [$id]);
+	}
+
+	public function setParentMessageId(?string $id): void {
+		if (empty($id)) {
+			parent::setter('parentMessageId', [null]);
```
nickvergessen, line 147: :boom:

lib/Db/Message.php
```diff
@@ -134,6 +135,22 @@ public function __construct() {
 		$this->addType('updatedAt', 'integer');
 	}
 
+	public function setMessageId(?string $id): void {
+		if (empty($id)) {
+			parent::setter('messageId', [null]);
+			return;
```
kesselb, line 141: I would use else ;)

ChristophWurst:
Let's not merge this yet. I want to also store the `References` header value with the message and give https://www.jwz.org/doc/threading.html a go.

Issue #2066, opened 2019-10-02
Sending not working to groups and new users


wdfee:
### Expected behavior
1. When typing an existing user into the "To" field, it should autocomplete the username.
2. When typing a group name into the "To" field, it should autocomplete the group name. When clicking the "send" button, it should send this mail to all users of this group.

### Actual behavior
1. The autosuggestion only works for users, who are already registered for some time. Autosuggestion does not work for new users, who were added today (and already logged in).
2. The autosuggestion for group names is working. But on sending, it breaks with error:
Error: Request failed with status code 500
(Sending to single users is working.)

### Mail app

**Mail app version:** 0.17.0

**Mailserver or service:** ionos + hosteurope

### Server configuration

**Operating system**: Ubuntu 18.0.4

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud Version:** 16.0.4 + 16.0.5

#### Client configuration
**Browser:** Firefox 69.0.1

**Operating system:** Mac OS 10.14.6

wdfee:
ok
1. solved: in the settings "sharing" section I had to to untick "Allow username autocompletion in share dialog. If this is disabled the full username or email address needs to be entered." and then reactivate this option again. Now the new users are found by autosuggestion.

wdfee:
2. log entries for failed sending:

[mail] Fatal: Sending mail failed: valid RCPT command must precede DATA
POST /index.php/apps/mail/api/accounts/4/send
from *** by *** at 2019-10-02T23:20:31+02:00

[mail] Error: Horde_Mime_Exception: valid RCPT command must precede DATA at <<closure>>

 0. /var/www/vhosts/[domain]/httpdocs/apps/mail/vendor/pear-pear.horde.org/Horde_Mime/Horde/Mime/Mail.php line 479
    Horde_Mime_Part->send("", Horde_Mime_Headers {}, Horde_Mail_Trans ... "})
 1. /var/www/vhosts/[domain]/httpdocs/apps/mail/lib/Account.php line 214
    Horde_Mime_Mail->send(Horde_Mail_Trans ... "}, false, false)
 2. /var/www/vhosts/[domain]/httpdocs/apps/mail/lib/Service/MailTransmission.php line 117
    OCA\Mail\Account->sendMessage(OCA\Mail\Model\Message {}, Horde_Mail_Trans ... "}, 105)
[...]

PR #2127, opened 2019-10-17
Fix groups expansion on umlauts

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2066

- [x] Add a test to verify the bug/fix
- [x] Fix the bug

### DIFF
--- a/tests/Service/GroupsIntegrationTest.php
+++ b/tests/Service/GroupsIntegrationTest.php
@@ -25,10 +25,14 @@ use ChristophWurst\Nextcloud\Testing\TestCase;
 use OCA\Mail\Service\GroupsIntegration;
 use OCA\Mail\Service\Group\NextcloudGroupService;
 use OCA\Mail\Exception\ServiceException;
+use PHPUnit\Framework\MockObject\MockObject;
 
 class GroupsIntegrationTest extends TestCase {
 
+	/** @var NextcloudGroupService|MockObject */
 	private $groupService1;
+
+	/** @var GroupsIntegration */
 	private $groupsIntegration;
 
 	protected function setUp() {
@@ -38,7 +42,9 @@ class GroupsIntegrationTest extends TestCase {
 		$this->groupService1->expects($this->any())
 			->method('getNamespace')
 			->willReturn('Namespace1');
-		$this->groupsIntegration = new GroupsIntegration($this->groupService1);
+		$this->groupsIntegration = new GroupsIntegration(
+			$this->groupService1
+		);
 	}
 
 	public function testGetMatchingGroups() {
@@ -76,7 +82,7 @@ class GroupsIntegrationTest extends TestCase {
 				'name' => "Bobby",
 				'email' => "bob@smith.net"
 			],
-			[ 
+			[
 				'id' => 'mary',
 				'name' => 'Mary',
 				'email' => 'mary@smith.net'
@@ -102,7 +108,7 @@ class GroupsIntegrationTest extends TestCase {
 				'name' => "Bobby",
 				'email' => "bob@smith.net"
 			],
-			[ 
+			[
 				'id' => 'mary',
 				'name' => 'Mary',
 				'email' => 'mary@smith.net'
@@ -128,7 +134,7 @@ class GroupsIntegrationTest extends TestCase {
 				'name' => "Bobby",
 				'email' => "bob@smith.net"
 			],
-			[ 
+			[
 				'id' => 'mary',
 				'name' => 'Mary',
 				'email' => 'mary@smith.net'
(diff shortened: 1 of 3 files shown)


ChristophWurst:
@wdfee @myrho please review :)

wdfee:
thanks for involving me. unfortunately I was not able to test successfully. 

First I tried just changing the line in lib/Service/GroupsIntegration.php, still in Nextcloud 16.0.5 - this alone didn't solve the bug.

Then I upgraded Nextcloud to 17.0.0 because the mail master version (and this branch) require it. In Nextcloud 17.0.0 I'm not able to activate the mail app, it always deactivates itself. I tried installing branch "fix groups expansion on umlauts" manually, then installing master zip manually, and last reinstalling version from app store.
I'll restore my backup of 16.0.5 now.

General question: is it really necessary to require the newest major version of Nextcloud for the next app update? It's so hard to get Nextcloud to any stable setup – some (important) apps are not updated yet and maybe won't work in 17, others like the mail app force upgrading to get some bug fixes for them, new bugs appear with every major Nextcloud version, and there are many major versions per year... Uff 😕

But thanks for your fixing, I guess the (de)activating issue is something different.

ChristophWurst:
@wdfee thanks for the feedback. The versions limitation is mostly a organizational problem. I do not have the time to test Mail against all Nextcloud versions, hence I just release new updates for the versions I test with. This is usually the latest major version and the current development build.

Sometimes an app may also depend on an API from server. Right now we have that with event system that was added in Nextcloud 17. This app does technically not work on 16. There is no way around.

Fixes could be backported. But then I need (community) help as well, as I have other tasks to work on except for maintaining the Mail code for old Nextcloud versions. If someone can help testing, I'm happy to do more backports.

ChristophWurst:
> However, since this part merely is about getting the group names out of the input stringe, I'd suggest to use `([^,]+)(,?)` instead. So we don't care for what's in the name at all.

Let's just fine out the allowed chars on Nextcloud group names and use that. I'll ask my colleagues for help.

wdfee:
@ChristophWurst thanks for your reply. I really would like to test this if I could. Do you think the github app version doesn't work on my install (deactivates itself) because of missing code signing that would be added through app store?
I have two setups: One staging for tests and one production, both setup by web installer of community version. On production I set the update channel to stable – here the 16.0.5 is displayed as newest, no update to 17 available yet. So from my point of view 16.0.5 is the actual version I have to use and not "old". In staging I set the update channel to beta, here 17.0.0 is displayed, but web updater doesn't work, I had to install it manually (the updater did work for all stable versions before). Unfortunately the mail app didn't work there – but I don't know if the cause is in the app or in Nextcloud server.

I'm still at the very beginning to understand Nextcloud logic and architecture. I read the developer manual again, about backporting and so on, but still don't know what I can do to help.
As I told you my roots are in WordPress development. I started there in it's early stages and I see some parallels to Nextcloud, that is still so young. I think Nextcloud has the potential to become the next "big thing", but there are many bugs to solve first and the community has to grow.
In WordPress it is so easy to become part of the community, because even beginners with only few coding skills can contribute. To test a bug patch only the plugin (like app) has to be replaced by the github branch. Developers can just perform small version check conditions to keep backwards compatibility.
I think in software development it's a good way to maintain backwards compatibility to at least two major versions.

Sorry for being so offtopic here. I'd really like to step in a little deeper, but keep stumbling.

ChristophWurst:
> Do you think the github app version doesn't work on my install (deactivates itself) because of missing code signing that would be added through app store?

No, this should be fine. Just make sure then when replacing an app to remove all previous files.

>  16.0.5 is displayed as newest, no update to 17 available yet

We are aware of this. The roll-out is done in stages. 17 isn't delivered to everyone yet.

> Unfortunately the mail app didn't work there – but I don't know if the cause is in the app or in Nextcloud server.

It should. I run Mail on my 17 server.

> I think in software development it's a good way to maintain backwards compatibility to at least two major versions.

I second that. But Nextcloud apps aren't fully independent of the Nextcloud server code. There can be some code dependencies. Right now we use an API that was only added in Nextcloud 17. There is no way we could still be compatible with 16 or older. The only thing we can do is backport (simple) fixes to older releases. But that requires someone's time to do it and I'm currently fully occupied with feature development as some critical features are still missing in this app (see https://github.com/nextcloud/mail/projects/6). Once an app is stable it's also a lot easier to maintain for many versions.

> In WordPress it is so easy to become part of the community, because even beginners with only few coding skills can contribute. To test a bug patch only the plugin (like app) has to be replaced by the github branch.

I don't fully see how this is different here. Of course if the app is not compatible with the base product version in use you will also not be successful, unless the wordpress APIs never change (don't know them). How's this not possible with Nextcloud apps? Is it about the composer/npm dependencies?

tests/Service/GroupsIntegrationTest.php
```diff
@@ -146,6 +152,83 @@ public function testExpand2() {
...
+
+		$expected = "john@doe.com,bob@smith.net,mary@smith.net";
+
+		$actual = $this->groupsIntegration->expand($recipients);
+
+		$this->assertEquals($expected, $actual);
+	}
+
+	public function testExpandLabeledGroup() {
+		$recipients = "\"test group (Nextcloud)\" <nextcloud:test group>";
```
ChristophWurst: @myrho this is what I get from the groups integration, sent to the server as is. Was the server ever able to understand/parse this?

myrho: I don't get it. What server?

wdfee:
@ChristophWurst I tried it again, using a running 17.0.0.9 beta and your newest fix now.
Unfortunately the mail app still deactivates itself.

> Just make sure then when replacing an app to remove all previous files.

Yes I did that.
1. deactivated mail app (and deactivated web instance)
2. made a backup of /apps/mail outside the root folder
3. deleted /apps/mail completely
4. downloaded branch mail-fix-groups-expansion-umlauts as zip
5. uploaded this zip to server, moved to /apps/, renamed to "mail"
6. reactivated web instance and activated mail app

The buttons for activating change, so it seems to activate the mail app. But if I navigate anywhere else, there is no mail app icon in navigation and in app list it is deactivated.

tests/Service/GroupsIntegrationTest.php
ChristophWurst: When the front-end does autocompletion, it gets an entry with
```json
{
  "label": "test group (Nextcloud)",
  "address": "nextcloud:test group"
}
```

If the user then send the email, the `to` HTTP parameters becomes `"test group (Nextcloud)" <nextcloud: test group>`.  This falls apart on the group expansion as the regex would detect `test group>` as group name.

Do you remember if we always did it this way? As in, sending the address **and** label? The group expansion would handle the address well, but the combination with label causes nothing but problems.

ChristophWurst:
@wdfee hm. There's also the problem that if you just download the zip, there won't be php (composer) dependencies, nor npm dependencies or javascript build artifacts, as we don't commit them.

wdfee:
@ChristophWurst uh, sure! I performed composer update now before uploading, now the mail app stays active after activating. The current issue isn't solved yet, but I keep on following your fixes and now I know how to test. Thanks for your work and your patience.

ChristophWurst:
Great to hear! Awesome!

I know it's not fully resolved yet, just the part where the label is not sent to the server (as mentioned in the other comment).

tests/Service/GroupsIntegrationTest.php
ChristophWurst: Sorry for the confusion caused. What I was looking for is this code: https://github.com/nextcloud/mail/blob/c89121aa5d2b3560a102dd2cb259dd017b5692dc/src/components/Composer.vue#L313. It needs an update for the new group names we allow (umlauts, spaces and others) :)

ChristophWurst:
Custom build for testers: [mail.tar.gz](https://github.com/nextcloud/mail/files/3784864/mail.tar.gz)

wdfee:
yes! issue is fixed using the custom build.
Umlauts, spaces and brackets are working in group name, mail is send to group members.

Issue #2110, opened 2019-10-16
🌈 Design fixes and polishing


skjnldsv:
## Message list
- [x] alignement of avatar → fix at https://github.com/nextcloud/mail/pull/2126, component at https://github.com/nextcloud/nextcloud-vue/pull/616
- [x] second row position → fix at https://github.com/nextcloud/mail/pull/2126, component at https://github.com/nextcloud/nextcloud-vue/pull/616
- [x] delete animation: quicker, height left at the end, issue https://github.com/nextcloud/mail/issues/1761

## App navigation
- [x] Account menu: show "delete account"
- [x] Do not show "loading" on the mail number stats
- [x] Bottom left settings can use more spacing -> #2219 
- [ ] Clicking "New message" should close nav and go to composer. On narrow screens the nav stays open and it seems like nothing happens.
- ~~When you zoom in, the nav hides but opening the hamburger menu doesn't show the nav~~ → core issue, see **Scaling issue with burger menu in split screen https://github.com/nextcloud/server/issues/17592**

## Menu button
- [x] On narrow screens, the message view has both a menu button and a back arrow button. In this view only the back arrow button is necessary and it should be in place of the menu button.
- ~~Menu button and "New message" placement~~ → core issue, see **Mobile menu toggle for navigation overlaps with content (especially in Mail & Contacts) https://github.com/nextcloud/server/issues/16934**

## Message view
- [ ] For HTML emails, the iframe doesn’t seem to fit the whole container -> #2222
- [x] In the heading of a mail below the subject, if a contact has only an email set but no name, the tooltip does not need to be shown. -> #2223
- [x] no message emptycontent h2

## Composer
- [ ] Composer overflowing right on mobile (because of attachment buttons, or from/to multiselects?) -> #2220 
- [x] Cc/Bcc maxcontrast and non-caps → pull request at https://github.com/nextcloud/mail/pull/2131
- [x] Align text (image blocking, header and mail content) → pull request at https://github.com/nextcloud/mail/pull/2131
- [x] Sticky bottom menu (send?), or have stuff at top of composer like Geary → pull request at https://github.com/nextcloud/mail/pull/2131
- [x] Put attachement buttons into the three dot menu and make this menu more visible (circle/background?) -> #2226 

## Account settings
- [x] Signature and Editor toggle at top
- [x] Editor toggle should be radio
- [x] Signature field should be bigger
- → All fixed in https://github.com/nextcloud/mail/pull/2119

@jancborchardt

[...]

Issue #1761, opened 2019-05-18
Mail deletion animation has issues with the menu popover


jancborchardt:
When you open the 3-dot-menu of a message and delete it, it seems the menu flickers, reappears, and then also gets animated/compressed. And the animation compresses the entry until a certain stage, then the last 10px are not animated. Looks quite strange:
![mail deletion animation](https://user-images.githubusercontent.com/925062/57962052-9a34c580-7913-11e9-9291-f38beee5c839.gif)

Instead the menu should vanish right away, and the deletion animation fast and smooth.

Not sure whether it’s an issue with @nextcloud/mail or @nextcloud/vue though?

skjnldsv:
Probably on the new standards here: https://github.com/nextcloud/nextcloud-vue/pull/382 :)

ChristophWurst:
I would love to remove/replace that animation anyway. The collapse never appealed to me.

jancborchardt:
We do need some feedback on deletion though, and collapsing is a great non-flamboyant way of doing that, which doesn't just jerk the entries below up.

Yes, we can improve the animation, like easing in and out, and making it quicker. But we do need an animation. :)

ChristophWurst:
:+1: on animation

:-1: on collapse :wink:

jancborchardt:
Fixed by @GretaD via https://github.com/nextcloud/mail/pull/2130 :balloon:

PR #2130, opened 2019-10-18
Improve animation on EnvelopeList when you delete an email

GretaD:
> delete animation: quicker, height left at the end, issue #1761 and here #2110

cc @jancborchardt @ChristophWurst I also removed the transform: scaleY(0), to remove that collapse effect, what do you think?

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -257,10 +257,9 @@ export default {
 #list-refreshing {
 	overflow-y: hidden;
 	min-height: 0;
-
 	transition-property: all;
 	transition-duration: 0.5s;
-	transition-timing-function: ease-in;
+	transition-timing-function: ease-in-out;
 }
 
 #list-refreshing.refreshing {
@@ -269,7 +268,7 @@ export default {
 
 .list-enter-active,
 .list-leave-active {
-	transition: all 1s;
+	transition: all var(--animation-quick);
 }
 
 .list-enter,


jancborchardt:
> I also removed the transform: scaleY(0), to remove that collapse effect, what do you think?

Some sort of animation is needed so that the list items below don’t just jump up when the mail is faded out. And a very quick collapse would be best as that’s the most seamless transition between the 2 states. :)

GretaD:
>And a very quick collapse would be best as that’s the most seamless transition between the 2 states. :)

is this better or it maybe 0.1s slower?

jancborchardt:
The speed is good, but it still has the issue that it animates down to like 10px and the last part vanishes quickly. As described with the video in issue https://github.com/nextcloud/mail/issues/1761 :)

jancborchardt:
Looking good! :) I think @ChristophWurst will also ask to squash your commits into one though, so better do that directly

src/components/EnvelopeList.vue
```diff
@@ -269,7 +268,7 @@ export default {
 
 .list-enter-active,
 .list-leave-active {
-	transition: all 1s;
+	transition: all 100ms;
```
jancborchardt:
```suggestion
	transition: all var(--animation-quick);
```

As per the variables.scss :)

GretaD: Thats what @skjnldsv would have said :)

skjnldsv: :see_no_evil:

skjnldsv:
> The speed is good, but it still has the issue that it animates down to like 10px and the last part vanishes quickly

I guess you still needs to animate the padding?

jancborchardt:
Tested again and it seems good to me now, only thing that’s left @GretaD squash into one commit. :) @skjnldsv the padding issue does not occur for me anymore, did you clear the cache?

skjnldsv:
> did you clear the cache?

I did not test, I assumed your comment was still occurring as I did not see anything related to the padding in the commits :)
I'm guessing the speed is _faking_ the instant toggle of the padding collapse ;)

jancborchardt:
> I'm guessing the speed is _faking_ the instant toggle of the padding collapse ;)

All this UI stuff, a bunch of magic! :crystal_ball:

ChristophWurst:
Congrats on the commit sha 81d8d18b. It almost rimes.

ChristophWurst:
Oh it is actually a palindrome!

jancborchardt:
### `Where were you when it happened?`

81d8d18

GretaD:
> Looking greta, let’s get this in! +1

i see what you did there :D

PR #2131, opened 2019-10-18
Composer design improvements

jancborchardt:
Lots of little fixes, but still some issues which be something for follow-up pull requests instead?

- [ ] Subject field is not automatically focused despite autofocus
- [ ] Possibly place "Draft saved" message below button? Also it could vanish after a few seconds?
- [ ] Sticky message options menu should open to the top → https://github.com/nextcloud/nextcloud-vue/issues/685
- [ ] Attachment upload buttons could go into the message options menu
- [ ] Attachment list layout needs to be rethought, especially on mobile → separate pull request I’d say
- [ ] Actions bar on bottom overlaps text, but if the previous 2 things are fixed then this is likely less of an issue.

## Before
Mail composer before

## After
Mail composer after

Please review @ChristophWurst @nextcloud/mail

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1,8 +1,8 @@
 <template>
 	<div v-if="state === STATES.EDITING" class="message-composer">
 		<div class="composer-fields mail-account">
-			<label class="to-label transparency" for="from">
-				{{ t('mail', 'from') }}
+			<label class="from-label" for="from">
+				{{ t('mail', 'From') }}
 			</label>
 			<Multiselect
 				id="from"
@@ -10,13 +10,16 @@
 				:options="aliases"
 				label="name"
 				track-by="id"
+				:searchable="false"
+				:hide-selected="true"
 				:custom-label="formatAliases"
+				:placeholder="t('mail', 'Select account')"
 				@keyup="onInputChanged"
 			/>
 		</div>
 		<div class="composer-fields">
-			<label class="to-label transparency" for="to">
-				{{ t('mail', 'to') }}
+			<label class="to-label" for="to">
+				{{ t('mail', 'To') }}
 			</label>
 			<Multiselect
 				id="to"
@@ -26,17 +29,19 @@
 				label="label"
 				track-by="email"
 				:multiple="true"
+				:placeholder="t('mail', 'Contact or email address …')"
+				:show-no-options="false"
 				@keyup="onInputChanged"
 				@tag="onNewToAddr"
 				@search-change="onAutocomplete"
 			/>
-			<a v-if="!showCC" href="#" @click.prevent="showCC = true">
-				{{ t('mail', '+ CC/BCC') }}
+			<a v-if="!showCC" class="copy-toggle" href="#" @click.prevent="showCC = true">
+				{{ t('mail', '+ Copy') }}
 			</a>
 		</div>
 		<div v-if="showCC" class="composer-fields">
-			<label for="cc" class="cc-label transparency">
-				{{ t('mail', 'CC') }}
+			<label for="cc" class="cc-label">
+				{{ t('mail', 'Copy') }}
 			</label>
 			<Multiselect
 				id="cc"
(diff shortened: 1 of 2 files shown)


ChristophWurst:
> [Work in progress]

You can also open a PR as draft now ;)

src/components/Composer.vue
```diff
@@ -26,17 +29,19 @@
...
+				:placeholder="t('mail', 'Contact or email address …')"
+				:show-no-options="false"
 				@keyup="onInputChanged"
 				@tag="onNewToAddr"
 				@search-change="onAutocomplete"
 			/>
-			<a v-if="!showCC" href="#" @click.prevent="showCC = true">
-				{{ t('mail', '+ CC/BCC') }}
+			<a v-if="!showCC" class="transparency" href="#" @click.prevent="showCC = true">
+				{{ t('mail', '+ Cc/Bcc') }}
```
ChristophWurst: It was uppercase intentionally https://github.com/nextcloud/mail/pull/2071

jancborchardt: So I changed it because both Gmail and Microsoft Outlook use "Cc" and "Bcc", and full-on uppercase just looks shouty.

And actually Apple Mail uses "Copy" and "Blind copy", which I would change it to instead as it is a lot less cryptic. :)

ChristophWurst: cc @rakekniven

rakekniven: @jancborchardt I checked older KMail and Thunderbird and Wikipedia.
Please change to other resources as well.

jancborchardt: @rakekniven Wikipedia also uses "Cc" and "Bcc" (as well as of course explaining it stands for "Carbon copy" and "Blind carbon copy") https://en.wikipedia.org/wiki/Blind_carbon_copy

By other occurences do you mean in the Mail app, or is this elsewhere? Thanks for your help! :)

jancborchardt:
Updated and added before/after screenshot. There’s some possible todos, but in the sense of improving step by step we could do them in follow-up pull requests.

skjnldsv:
Maybe add a background on the bottom toolbar?

jancborchardt:
Added a fade from main background (bottom) to transparent (top) just like we do in the filepicker. As said above in the list though I’d rather fix the real issues there though.

jancborchardt:
Anything else? What is a blocker, what should I still fix, and what can be done separately @ChristophWurst?

ChristophWurst:
> Lots of little fixes, but still some issues which be something for follow-up pull requests instead?

Please open separate tickets for these things. Small PRs are good PRs :)

jancborchardt:
@ChristophWurst it's modeled after how the subject looks when reading mails too. I'd move from bold to regular, but the other changes make it lose structure since everything looks thr same.

(The alignment is intentional – it is aligned with the mail body.)

ChristophWurst:
> @ChristophWurst it's modeled after how the subject looks when reading mails too (as well as the general h2). I'd move from bold to regular, but the other changes make it lose structure since everything looks thr same.

isn't that what *gives* structure? If one of four things looks so different it breaks the structure IMO.

> 
> (The alignment is intentional – it is aligned with the mail body.)

Fair enough.

One more thing though, aren't we also introducing this anti-pattern of having only an inline label: https://baymard.com/blog/mobile-forms-avoid-inline-labels?

jancborchardt:
> isn't that what gives structure? If one of four things looks so different it breaks the structure IMO.

Sorry, structure was not really a good word to bring it across. I mean that it loses focus and guidance.
You would have 3 inputs below each other which look exactly the same, and below that a text thing with also the same size. This basically makes it look like you are working on an Excel sheet (or German bureaucracy form) where people don’t immediately know what the important stuff is and you get lost in a sea of things looking as if they have the same importance.


> One more thing though, aren't we also introducing this anti-pattern of having only an inline label: 

That’s another reason why the subject is differently styled: It’s very clear that it is the mail subject, without really needing a label. Also because (as said) it appears exactly the same as when you read Mail, so you have that symmetry. :)

skjnldsv:
> That’s another reason why the subject is differently styled: It’s very clear that it is the mail subject, without really needing a label. Also because (as said) it appears exactly the same as when you read Mail, so you have that symmetry. :)

Could we remove the bold though, it seems a bit too much.
Maybe juste the increased font size?

jancborchardt:
As said, the bold has been like that in Mail reading for some time now, and it looks really good giving point of reference. Also, we toyed around with giving the heading in the right sidebar bold, which is a h2 as well – you remember? ;)

So let’s keep it bold, and some other parts should be bold too.

ChristophWurst:
> **Travis CI - Pull Request ** Build Errored

Rule of thumb: make CI green *before* merging. This caused a build break on all new PRs and thus a huge Dependabot backlog.

jancborchardt:
@ChristophWurst sorry! Merged cause you gave me the green light via message. Forgot to run the linter, but am also confused cause I did not touch the content of the line – I guess it is because of the container and the further indent, and the linter wants it to not exceed a certain amount of characters?

ChristophWurst:
Could be, yeah

PR #2147, opened 2019-10-21
Allow formatted replies

ChristophWurst:
- [x] Remove html-to-text conversion and let the editor handle it if necessary
- [x] Fix conversion from HTML blockquote to text-only

Test 1
* Set rich editor as default
* Open a (HTML) message
* Click reply
* See block quote formatted message body

Test 2
* Start reply
* Switch between rich and plain text mode
  * Block quotes become `> ` in simple mode
  * `> ` stay the same when switching from simple to rich

### DIFF
--- a/src/ReplyBuilder.js
+++ b/src/ReplyBuilder.js
@@ -20,11 +20,8 @@
  *
  */
 
+import moment from '@nextcloud/moment'
 import negate from 'lodash/fp/negate'
-import moment from 'moment'
-import {getLocale} from '@nextcloud/l10n'
-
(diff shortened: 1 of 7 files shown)


ChristophWurst:
Ref https://github.com/ckeditor/ckeditor5/issues/393 https://github.com/ckeditor/ckeditor5/issues/419 https://github.com/ckeditor/ckeditor5/issues/2137 for the editor hacks

ChristophWurst:
cc @matiasdelellis for review :)

PR #2155, opened 2019-10-23
Fix missing account display name and email prefill

ChristophWurst:
This was probably lost with the Vue rewrite. Nobody seemed to have missed it but it's useful during testing :)


jancborchardt:
I did actually notice but must have thought "cmooon, not another time I have to report this!" ;)

Issue #2092, opened 2019-10-10
Some messages fail to decode properly


StCyr:
### Expected behavior
Messages should be displayed nicely on screen

### Actual behavior
Some messages are not displayed nicely (probably not decoded properly) though they are displayed nicely on another webmail app.

Issue with `text/plain` content-type?

(source [email.txt](https://github.com/nextcloud/mail/files/3711370/email.txt))

mail formating: 

![image](https://user-images.githubusercontent.com/8179031/66552098-0923ec80-eb49-11e9-893b-f350d03d6729.png)

ovh formating: 

![image](https://user-images.githubusercontent.com/8179031/66552742-27d6b300-eb4a-11e9-9536-af04b7b9b215.png)

### Mail app

**Mail app version:** 0.17.0

**Mailserver or service:** OVH

### Server configuration

**Operating system**: Debian Buster/10/stable

**Web server:** Apache

**Database:** MariaDB

**PHP version:** (e.g. 7.0)

**Nextcloud Version:** 17

#### Client configuration
**Browser:** Firefox Quantum stable

**Operating system:** Windows 10

StCyr:
so, yeah, that's an issue with a plain text message. Here's the message as received by the front-end:

![image](https://user-images.githubusercontent.com/8179031/66748373-4ceb5e80-ee87-11e9-9f14-fe37236ef0ba.png)

The problem comes from the `\n` characters, but I don't know where this should be fixed (backend/frontend). Any idea @nextcloud/mail ?

ChristophWurst:
> Content-Transfer-Encoding: 8bit

Usually transfer encoding is taken into account IIRC. And Horde should thus handle it properly.

StCyr:
> 
> 
> > Content-Transfer-Encoding: 8bit
> 
> Usually transfer encoding is taken into account IIRC. And Horde should thus handle it properly.

Sorry, I don't understand: Do you mean that horde have already translated the `\n` into `<br>`?!?

ChristophWurst:
I have to attach a debugger and step through the code to give a qualified answer. I don't understand the problem either at the moment. It could be an encoding issue. It could be this app changing some text. I don't know.

StCyr:
> I have to attach a debugger and step through the code to give a qualified answer. 

That's already a good answer :-) 

This means backend should be verified first.

PR #2157, opened 2019-10-24
In MessagePlainTextBody, translates all newline chars into '<br>'

StCyr:
In MessagePlainTextBody, translates all newline chars into `<br>` to better display plain text messages.

Fixes https://github.com/nextcloud/mail/issues/2092

Eventually, I've discovered PHP's `nl2br` function. So, a better fix might be to updated the `IMAPMessage` class's getFullMessage function as follow:

```php
        public function getFullMessage(int $accountId, string $mailbox, int $id): array {
                $mailBody = $this->plainMessage;

                $data = $this->jsonSerialize();
                if ($this->hasHtmlMessage) {
                        $data['hasHtmlBody'] = true;
                        $data['body'] = $this->getHtmlBody($accountId, $mailbox, $id);
                } else {
---                        $mailBody = $this->htmlService->convertLinks($mailBody);
+++                        $mailBody = nl2br($this->htmlService->convertLinks($mailBody));
                        list($mailBody, $signature) = $this->htmlService->parseMailBody($mailBody);
                        $data['body'] = $mailBody;
                        $data['signature'] = $signature;
                }

                $data['attachments'] = $this->attachments;

                return $data;
        }
```

### DIFF
--- a/src/components/MessagePlainTextBody.vue
+++ b/src/components/MessagePlainTextBody.vue
@@ -1,6 +1,6 @@
 <template>
 	<div>
-		<div id="mail-content" v-html="body"></div>
+		<div id="mail-content" v-html="sanitizedBody"></div>
 		<div v-if="signature" class="mail-signature" v-html="signature"></div>
 	</div>
 </template>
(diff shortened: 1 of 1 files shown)


ChristophWurst:
I will double check if this conversion is correct. I've never had emails that were displayed like yours, hence I'm afraid we're not really fixing what's broken but patching it to look like it works.

ChristophWurst:
I imported your email and could reproduce the problem. I still doubted the nl2br thing but found out that I accidentally removed it recently with a little cleanup. So it's rightfully missing. I added it to a spot where it fits better as we also add the signature line breaks there: https://github.com/nextcloud/mail/pull/2167

ChristophWurst:
In any case thanks a lot for your help! I wouldn't have been so easy without your insights and the provided email!

PR #2160, opened 2019-10-24
Account settings should be a form

ChristophWurst:
This has two advantages:

* You can CTRL+Enter in any field to submit (power user)
* The browser detects the form submit and will remember values (super useful for test setups but also other users)

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -1,5 +1,5 @@
 <template>
-	<div id="account-form">
+	<form id="account-form" @submit.prevent="onSubmit">
 		<tabs
 			:options="{useUrlFragment: false, defaultTabHash: settingsPage ? 'manual' : 'auto'}"
 			cache-lifetime="0"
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> plus you know... it kind of is a form...

yep. true words are true

src/components/AccountForm.vue
```diff
@@ -228,13 +228,13 @@
 		</tabs>
 		<slot name="feedback"></slot>
 		<input type="submit" class="primary" :disabled="loading" :value="submitButtonText" @click="onSubmit" />
```
skjnldsv: Be careful about the double submit now, because an input[type=submit] will trigger a click and then the form submit :)

ChristophWurst: daaaaaaaaaayum

jancborchardt:
> Good fortune awaits those who write correct HTML.
> _– Socrates ca. 2009_

PR #2179, opened 2019-10-25
Fix unnecessary line breaks in quoted reply text

ChristophWurst:
Back-end concerns should be handled on the back-end, front-end ones on the front-end. Hence the `nl2br` on the server was a mistake (#2167). We don't only need the text for displaying, but also to build the quoted reply. The latter is where the old code fell apart, adding more line breaks than necessary. Now both displaying the original message and quoting its text for replies looks good.

### DIFF
--- a/lib/Service/Html.php
+++ b/lib/Service/Html.php
@@ -98,12 +98,12 @@ class Html {
 		$signature = null;
 		$parts = explode("-- \r\n", $body);
 		if (count($parts) > 1) {
-			$signature = nl2br(array_pop($parts));
+			$signature = array_pop($parts);
 			$body = implode("-- \r\n", $parts);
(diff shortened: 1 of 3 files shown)


jancborchardt:
@ChristophWurst how can I best test this, and what should be the expected outcome? Replying to a plain-text email?

ChristophWurst:
> @ChristophWurst how can I best test this, and what should be the expected outcome? Replying to a plain-text email?

Yes.

Before

```
Hi,
some text
```
turned into

```
> hi,
>
> some text
```
, now it's

```
> hello,
> some text
```

rullzer:
So it is a feature that is changes `hi` into `hello` :wink:

PR #2194, opened 2019-10-28
Fix loggin drafts promise errors

ChristophWurst:
A function passed to `.catch` loses its context unless bound to one explicitly, hence it's necessary to use `.catch(console.error.bind(this))` or `.catch(logger.error.bind(logger))`.

Fixes ``TypeError: this is undefined``


ChristophWurst:
Yes, will migrate one day. Today is not that day :)

PR #2203, opened 2019-10-29
Test against php beta & nightly

ChristophWurst:
The php version check will catch it, but let's try.


ChristophWurst:
ha! I knew I had a PR for this …

PR #2211, opened 2019-11-04
relocate toggle arrow button

GretaD:
ref #[2110](https://github.com/nextcloud/mail/issues/2110#issuecomment-548906027)

### DIFF
--- a/src/components/AppDetailsToggle.vue
+++ b/src/components/AppDetailsToggle.vue
@@ -39,11 +39,10 @@ export default {
 	left: 0;
 	width: 44px;
 	height: 44px;
-	z-index: 149;
-	background-color: var(--color-main-background-darker);
+	z-index: 1500;
 	cursor: pointer;
-	opacity: 0.6;
 	transform: rotate(180deg);
-	margin-top: 44px; /* under the show navigation button */
+	background-color: var(--color-main-background);
+	opacity: 1;
 }
 </style>


src/components/AppDetailsToggle.vue
```diff
@@ -39,11 +39,10 @@ export default {
 	left: 0;
 	width: 44px;
 	height: 44px;
-	z-index: 149;
-	background-color: var(--color-main-background-darker);
+	z-index: 1051;
```
ChristophWurst: 🧐🧐🧐

src/components/AppDetailsToggle.vue
```diff
@@ -39,11 +39,10 @@ export default {
...
 	cursor: pointer;
-	opacity: 0.6;
 	transform: rotate(180deg);
-	margin-top: 44px; /* under the show navigation button */
+	background-color: white;
```
ChristophWurst: what about dark theme? :)

GretaD: Thats a very good question. Working on it

src/components/AppDetailsToggle.vue
GretaD: you dont like my 1051 index number? I can make it 2000 ;)

jancborchardt:
Thanks for picking it up! :) There is the same issue in Contacts as well, mind fixing that the same way?
contacts back and menu

ChristophWurst:
> Thanks for picking it up! :) There is the same issue in Contacts as well, mind fixing that the same way?

Mind submitting a review for this one? :)

jancborchardt:
@GretaD commits should be squashed into 1, otherwise good! :)

PR #2214, opened 2019-11-05
Show itinerary info

ChristophWurst:


jancborchardt:
Very nice! Only some small issues at the moment:
- [ ] If I click one of the "Plus"-Buttons in a list with multiple itinerary steps, all of them show a spinner
- [ ] The min-width is a bit wide, meaning on smartphones stuff overflows to the right. Could be fixed by wrapping the part with the reservation code below the flight name. (Have tried with `white-space:` CSS but didn’t work?

Otherwise we should really make sure to check with different companies, platforms etc so we make sure we show itinerary info from the most common ones. :)

ChristophWurst:
Build 1a40f6310ada1e397e35820bb88e5e204706be61 [mail.tar.gz](https://github.com/nextcloud/mail/files/3920803/mail.tar.gz)

ChristophWurst:
Build 9b9bad1974ac54412488e99a5f4fdc0e9971054a [mail.tar.gz](https://github.com/nextcloud/mail/files/3921157/mail.tar.gz)

ChristophWurst:
> Otherwise we should really make sure to check with different companies, platforms etc so we make sure we show itinerary info from the most common ones. :)

This is out of scope for us. https://community.kde.org/KDE_PIM/KItinerary/Supported_Providers is the limiting factor here. Anything they support will show up here.

ChristophWurst:
Build a230ad96c6149d9253e9ddf4359db876aa020f16 [mail.tar.gz](https://github.com/nextcloud/mail/files/3922704/mail.tar.gz)

ChristophWurst:
3c6c2b70d6a9cb4775bad8918174b9fe4014381c fixes the issue that the special flagged inbox didn't work before as we tried to open an IMAP mailbox that only virtually exists. @mario you don't know that I know but you ran into this. Props to Sentry for capturing this :)

ChristophWurst:
>     * If I click one of the "Plus"-Buttons in a list with multiple itinerary steps, all of them show a spinner

Not reproducible here.

>     * The min-width is a bit wide, meaning on smartphones stuff overflows to the right. Could be fixed by wrapping the part with the reservation code below the flight name. (Have tried with `white-space:` CSS but didn’t work?

Can be fixed in a follow-up

StCyr:
Don't know how to test this.

I'm guesssing this needs Linux/KDE for it to be effective.

Anyway it builds, upgrade, and runs properly here.

ChristophWurst:
> Don't know how to test this.
> 
> I'm guesssing this needs Linux/KDE for it to be effective.

Just open an email with something like a train ticket or flight ticket.

It works out of the box, but not on all systems. If you don't run Linux x86_64 it won't show anything.

Make sure you install composer dependencies with `composer install` :)

Issue #1740, opened 2019-05-08
Not full mail visible on mobile


rullzer:
I noticed a bug when using the mail app on mobile. When viewing a message the content of the mail is to wide.

![localhost_apps_mail_](https://user-images.githubusercontent.com/45821/57375984-23732c00-719f-11e9-95c0-71ff94a0aa0d.png)

As you can see in the screenshot above.
I happens mostly in emails with log of recipients.

I found the cause of a few of the issues

1. The recipient list is not wrapping as it should
2. The composer part on the bottom
  a. The account selection is to wide and not wrapping as it should
  b. The recipeient (To, CC, BCC) are not wrapping as they should

If I remove the above mentioned elements the mail is fully readbale.

Of course making all element wrap properly is best here.
But at the least I should have a horizontal scroll bar if there is wider content ;)

ChristophWurst:
:eyes: @GretaD

jancborchardt:
This seems to be a duplicate of https://github.com/nextcloud/mail/issues/1641, closing hence.

PR #2218, opened 2019-11-05
fix full mail visibility on mobile

GretaD:
ref #1740

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -304,6 +304,7 @@ export default {
 		text-overflow: ellipsis;
 		padding-bottom: 7px;
 		margin-bottom: 0;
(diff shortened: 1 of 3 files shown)


ChristophWurst:
> Run `npm run lint:autofix` :)

Try again :speak_no_evil:

jancborchardt:
@GretaD and @violoncelloch did you coordinate on this then? Is this PR or https://github.com/nextcloud/mail/issues/2222 going to be used going forward? :)

violoncelloCH:
as also discussed with @juliushaertl and @skjnldsv the pure css approach in #2222 seems to be the better way to do this (less performance heavy as it just uses css flex)

PR #2219, opened 2019-11-06
👌 IMPROVE: navigation settings styling improvement (spacing etc.)

violoncelloCH:
image


jancborchardt:
Good stuff! :) Some additional thoughts, but can be separate:
- Should "Keyboard shortcuts" be styled as a button? Currently doesn’t look clickable.
- Then the "Add mail account" button could be primary
- The "Install the Mailvelope browser extension" should get a link underline to look clickable

violoncelloCH:
@jancborchardt I've implemented all of your suggestions:
image

PR #2220, opened 2019-11-07
🐛 FIX: various horizontal scrolling issues on mobile

violoncelloCH:


violoncelloCH:
> Run `npm run lint:autofix` :)

done :)

src/components/FolderContent.vue
```diff
@@ -13,7 +13,7 @@
 				/>
 				<NewMessageDetail v-if="newMessage" />
 				<Message v-else-if="showMessage" />
-				<NoMessageSelected v-else-if="hasMessages" />
+				<NoMessageSelected v-else-if="hasMessages && !this.isMobile" />
```
ChristophWurst:
```suggestion
				<NoMessageSelected v-else-if="hasMessages && !isMobile" />
```

No need for `this` in a few template. It's added automagically ;)

ChristophWurst:
How do I test this?

src/components/FolderContent.vue
violoncelloCH: whoops, of course; you're right...

violoncelloCH:
hmm, best would be to test this together with nextcloud/server#17843 and nextcloud/nextcloud-vue#711 because those two solve related problems... or what do you mean by "how"?

ChristophWurst:
More like what is the behavior before and after this fix :)

If we need https://github.com/nextcloud/nextcloud-vue/pull/711 then we should wait for the next release of that lib

violoncelloCH:
test it on narrow mobile layouts of 340px width to compare before and after... 
it makes the composer fit into those 340px and and also the message list is optimized for mobile..
yeah maybe it makes sense to wait for that release; on the other hand it's more an additional improvement than a requirement for this PR

violoncelloCH:
the only "functional" change is that it makes sure to not display the "no message selected" section on mobile (together with nextcloud/nextcloud-vue#711 this uses the correct breakpoint)...

PR #2222, opened 2019-11-07
🐛 FIX: message iframe vertically doesn't fit the container / available space

violoncelloCH:
Fixes https://github.com/nextcloud/mail/issues/658

### DIFF
--- a/css/mobile.scss
+++ b/css/mobile.scss
@@ -49,11 +49,6 @@
 		padding-right: 12px;
 	}
 
-	#mail-content,
-	.mail-message-attachments {
-		margin: 0 10px 50px 30px;
-	}
-
 	/* reply-forward actions align to the far right */
 	#mail-message-header #mail-message-actions {
 		margin-right: 5px;
--- a/css/mobile.scss
+++ b/css/mobile.scss
@@ -19,7 +19,6 @@
 		background: #fff;
 		width: 100%;
 		left: 0;
-		height: 100%;
 		top: 0;
 	}
 
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -67,7 +67,12 @@
 				</div>
 				<MessageHTMLBody v-if="message.hasHtmlBody" :url="htmlUrl" />
 				<MessagePlainTextBody v-else :body="message.body" :signature="message.signature" />
-				<MessageAttachments :attachments="message.attachments" />
+				<Popover v-if="message.attachments[0]" class="attachment-popover">
+					<Actions slot="trigger">
+						<ActionButton icon="icon-public icon-attachment">Attachments</ActionButton>
+					</Actions>
+					<MessageAttachments :attachments="message.attachments" />
+				</Popover>
 				<div id="reply-composer"></div>
 			</div>
 		</template>
@@ -77,6 +82,7 @@
 <script>
 import Actions from '@nextcloud/vue/dist/Components/Actions'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+import Popover from '@nextcloud/vue/dist/Components/Popover'
 import AppContentDetails from '@nextcloud/vue/dist/Components/AppContentDetails'
 import axios from '@nextcloud/axios'
 import Modal from '@nextcloud/vue/dist/Components/Modal'
@@ -107,6 +113,7 @@ export default {
 		MessageHTMLBody,
 		MessagePlainTextBody,
 		Modal,
+		Popover,
 	},
 	data() {
 		return {
@@ -287,7 +294,8 @@ export default {
 
 .mail-message-body {
 	flex-grow: 1;
-	margin-bottom: 100px;
+	margin-bottom: 20px;
+	position: relative;
 }
 
 #mail-message-header {
@@ -327,13 +335,19 @@ export default {
 	}
 }
 
-#mail-content,
-.mail-message-attachments {
-	margin: 10px 38px 50px 38px;
+.v-popover > .trigger > .action-item {
+	border-radius: 22px;
+	background-color: rgba(127, 127, 127, 0.25) !important;
 }
 
-.mail-message-attachments {
-	margin-top: 10px;
+.attachment-popover {
+	position: absolute;
+	bottom: 0;
+	right: 20px;
+}
+
+.tooltip-inner {
+	text-align: left;
 }
 
 #mail-content iframe {
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -286,6 +286,7 @@ export default {
 }
 
 .mail-message-body {
+	flex-grow: 1;
 	margin-bottom: 100px;
 }
 
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -61,7 +61,7 @@
 					</Modal>
 				</div>
 			</div>
-			<div class="mail-message-body">
+			<div :class="[message.hasHtmlBody ? 'mail-message-body mail-message-body-html' : 'mail-message-body']">
 				<div v-if="message.itineraries.length > 0" class="message-itinerary">
 					<Itinerary :entries="message.itineraries" :message-id="message.messageId" />
 				</div>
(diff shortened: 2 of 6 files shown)


jancborchardt:
@GretaD was already working on this at https://github.com/nextcloud/mail/pull/2218 :) Best coordinate on how to proceed

violoncelloCH:
so if we continue with this, there are still some things to do:
in order to remove the bottom (which preserves space for the attachments currently) and expand the iframe to the bottom of the viewport we still need to decide where to put the attachments instead. We could e.g. put an attachment indicator/button to the header which opens a popover or we could also have a floating indicator at the bottom which brings up a panel or something with the attachments. What do you think?

jancborchardt:
I think a floating indicator button for the attachments at the bottom is best, for the following reasons:
- The header is already too crowded
- Attachments are usually at the bottom
- When it’s floating, they can easily be shown
- It’s always a consistent size (just a button for "4 attachments" or in the case of 1 attachment it could show the info directly), and can expand on tap

jancborchardt:
@ChristophWurst does your :eyes:-Reaction on my comment mean that you approve? ;D

@violoncelloCH any other open questions on this, or help needed?

ChristophWurst:
> [jancborchardt](/jancborchardt) added this to the [0.19.0](/nextcloud/mail/milestone/34) milestone [38 minutes ago](#event-2820141434)

Thou shall not mess with my release schedules :)

ChristophWurst:
I'm fine with that. But sounds like a lot more work. And I'm still not sold on setting an arbitrary height on the iframe size in the hope that it is big enough. We'll then also have a much larger scrolling pane, no? The attachements are fixed, yes, but what comes after the message contents ended?

violoncelloCH:
@jancborchardt fine by me; it's time missing for me atm, not information... So if someone else wants to continue working on this I'd be fine... Maybe @GretaD ? Otherwise I'll try to find some time during the next weeks, but can't promise anything...sorry

@ChristophWurst this one explicitely doesn't set a fixed heigh but uses flex-grow which let's the iframe expand to the bottom of the viewport... Isn't this what we want to have?

ChristophWurst:
As far as I know you still can't have an iframe resize to the height of its contents dynamically, can you?

violoncelloCH:
just check out this PR and take a look how it's indeed possible :wink:

jancborchardt:
> As far as I know you still can't have an iframe resize to the height of its contents dynamically, can you?

This pull request increases the height of the iframe so the bottom edge lines up with the bottom edge of the viewport. It doesnt know about the contents of the iframe, but takes only the outside into account.

This results in what we want: Only a single scrollbar (albeit unfortunately hovering in the air).

jancborchardt:
@violoncelloch what’s left to do on this and how can I help? :)

violoncelloCH:
@jancborchardt well the basic css "magic" to make the iframe expand to the available screen height is there, however what's not yet implemented is the floating attachment button with a popover displaying the attachments (as you described in this comment https://github.com/nextcloud/mail/pull/2222#issuecomment-552147470)... (Or we merge this without the attachment part? [so there would just be a little bit of polishing needed])
Sadly I'm currently too busy with work, some personal projects, etc, so I can't promise to find time for this soon :/

szaimen:
Would someone be able to provide a test build? Then I could test this and see if it fixes #2694

jancborchardt:
@violoncelloCH ok, sounds good! :) I would say let’s finish it up with the attachment part as otherwise that would be a regression. Do you maybe have a split-up todo list for the tasks in case someone else wants to pick it up?

violoncelloCH:
so I finally made it :tada: 

there is now a button at the bottom right of the screen indicating attachment(s) and on click it opens a popover:

image

violoncelloCH:
@StCyr well it behaves totally fine in my tests. I guess this is one more time iOS misinterpreting some CSS standards for whatever reasons... 
As I don't have any iOS device at hand for testing/debugging I don't really know how I should proceed on this... Could you take a look into this?
The basic concept of this is to use `flex-grow: 1` to make the iframe dynamically expand vertically.

src/components/Message.vue
```diff
@@ -54,7 +54,12 @@
...
 				<MessageHTMLBody v-if="message.hasHtmlBody" :url="htmlUrl" />
 				<MessagePlainTextBody v-else :body="message.body" :signature="message.signature" />
-				<MessageAttachments :attachments="message.attachments" />
+				<Popover v-if="message.attachments[0]" class="attachment-popover">
+					<Actions slot="trigger">
+						<ActionButton icon="icon-public icon-attachment">Attachments</ActionButton>
+					</Actions>
+					<MessageAttachments :attachments="message.attachments" />
+				</Popover>
 				<div id="reply-composer"></div>
```
StCyr: Any idea why this "reply-composer" thingy is still here? Doesn't seem to be used...

violoncelloCH:
> I don't know what's going wrong yet, but I have identified that the issue only happens with HTML emails; plain text emails display just fine.

Well that is the point of this whole PR. It only addresses HTML mails, because plaintext never had this problem. For security reasons HTML mails have to be rendered in iframes which generate this trouble...

src/components/Message.vue
violoncelloCH: No idea... but I didn't want to touch code which I don't know what purpose it might serve... Maybe @ChristophWurst can help us here?

jancborchardt:
> so I finally made it tada
> 
> there is now a button at the bottom right of the screen indicating attachment(s) and on click it opens a popover:
> 
> image
> 
> let me know what you think about it :)

Aaaaawesome! :) Some feedback:

- [ ] The button could also use some text next to the icon, namely here "2 attachments"
- [ ] The button would look nicer centered in the view so on large screens it is directly below the message instead of off to the right
- [ ] The popover can then open centrally above the button

jancborchardt:
Another issue which just happens with plaintext mails: The alignment of the content is off to the left side.
Left current master, right this pull request:
Mail proper alignment    - - - - - - - -  Mail indentation wrong

@violoncelloCH :)

css/mobile.scss
```diff
@@ -50,11 +49,6 @@
 		padding-right: 12px;
 	}
 
-	#mail-content,
-	.mail-message-attachments {
-		margin: 0 10px 50px 30px;
-	}
```
jancborchardt: Any specific reason this was removed? This seems to cause the issue with the content being moved to the left. :)

szaimen: I think he wanted to use all the available space which could be more important then aligning all content at the same border...
What about the header of the message: couldn't it be always left-aligned and move to the right if the mobile view is activated to free the spacw for the back-button?

jancborchardt: @szaimen let’s please keep this pull request on topic. Moving the text so it sticks to the very left of the content is certainly a bug and not intentional:
image

jancborchardt:
Fixed the 2 outstanding issues, this is ready for review! @nextcloud/mail 

The only thing I was unable to fix is this which seems to need fixing in the components:

> The button could also use some text next to the icon, namely here "2 attachments"

css/mobile.scss
jancborchardt: Fixed

src/components/Message.vue
ChristophWurst: I think this can be removed in a separate PR

css/mobile.scss
violoncelloCH: this wasn't really intentional, sorry... wanted to remove the margin from the attachment section but not from the body/content, obviously...
thanks for fixing this! @jancborchardt

violoncelloCH:
thanks very much @jancborchardt for taking this up and fixing the remaining issues...
I'm sorry I didn't find the time to come back to this earlier...

jancborchardt:
No worries @violoncelloCH! Thank you so much for your work! :)

szaimen:
@ChristophWurst is there already an ETA when a new release will be available? I would really like to test this out :)

ChristophWurst:
No date set yet as we still work on stabilization, but feel free to test the betas https://github.com/nextcloud/mail/blob/v1.1.4/lib/Controller/FoldersController.php! I'll release another one soon so it will have this PR and other recent changes :)

szaimen:
Nice, I will do that! :)

rumbis:
.envelope{height:900px;}

PR #2226, opened 2019-11-09
👌 IMPROVE: put composer attachment buttons into three dot menu

violoncelloCH:
as discussed with @jancborchardt

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -123,7 +123,7 @@
 			></TextEditor>
 		</div>
 		<div class="composer-actions">
-			<ComposerAttachments ref="composerAttachments" v-model="attachments" @upload="onAttachmentsUploading" />
+			<ComposerAttachments v-model="attachments" :bus="bus" @upload="onAttachmentsUploading" />
 			<div class="composer-actions-right">
 				<p class="composer-actions-draft">
 					<span v-if="savingDraft === true" id="draft-status">{{ t('mail', 'Saving draft …') }}</span>
@@ -278,6 +278,7 @@ export default {
 			selectCc: this.cc,
 			selectBcc: this.bcc,
 			editorPlainText: this.isPlainText,
+			bus: new Vue(),
 		}
 	},
 	computed: {
@@ -358,10 +359,10 @@ export default {
 			this.saveDraftDebounced(this.getMessageData())
 		},
 		onAddLocalAttachment() {
-			this.$refs.composerAttachments.onAddLocalAttachment()
+			this.bus.$emit('onAddLocalAttachment')
 		},
 		onAddCloudAttachment() {
-			this.$refs.composerAttachments.onAddCloudAttachment()
+			this.bus.$emit('onAddCloudAttachment')
 		},
 		onAutocomplete(term) {
 			if (term === undefined || term === '') {
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -123,17 +123,26 @@
 			></TextEditor>
 		</div>
 		<div class="composer-actions">
-			<ComposerAttachments v-model="attachments" @upload="onAttachmentsUploading" />
+			<ComposerAttachments ref="composerAttachments" v-model="attachments" @upload="onAttachmentsUploading" />
 			<div class="composer-actions-right">
 				<p class="composer-actions-draft">
 					<span v-if="savingDraft === true" id="draft-status">{{ t('mail', 'Saving draft …') }}</span>
 					<span v-else-if="savingDraft === false" id="draft-status">{{ t('mail', 'Draft saved') }}</span>
 				</p>
 				<Actions>
-					<ActionText icon="icon-info">{{ t('mail', 'Message options') }}</ActionText>
-					<ActionCheckbox :checked.sync="editorPlainText" :text="t('mail', 'Plain text')">{{
-						t('mail', 'Plain text')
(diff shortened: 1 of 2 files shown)


ChristophWurst:
> as discussed with @jancborchardt

For https://github.com/nextcloud/mail/pull/2222#issuecomment-552147470?

src/components/Composer.vue
```diff
@@ -348,6 +357,12 @@ export default {
 		onInputChanged() {
 			this.saveDraftDebounced(this.getMessageData())
 		},
+		onAddLocalAttachment() {
+			this.$refs.composerAttachments.onAddLocalAttachment()
```
ChristophWurst: In general not a big fan of this coding style. This should be doable with events. Check https://stackoverflow.com/a/42634865/2239067 and https://stackoverflow.com/questions/43256978/event-from-parent-to-child-component/43258428#43258428. The former is less code, but requires a parent component (problematic for isolated testing), the latter requires a bit more code but is quite elegant and works with testing :)

src/components/Composer.vue
```diff
@@ -123,17 +123,26 @@
...
-					}}</ActionCheckbox>
+					<ActionButton icon="icon-upload" @click="onAddLocalAttachment">{{
+						t('mail', 'Upload attachment')
+					}}</ActionButton>
+					<ActionButton icon="icon-folder" @click="onAddCloudAttachment">{{
+						t('mail', 'Add attachment from Files')
+					}}</ActionButton>
+					<ActionCheckbox
+						:checked="!editorPlainText"
+						:text="t('mail', 'Enable text formatting')"
```
ChristophWurst: Please revert to the original message.

Moving the checkbox is okay, but sneaking in a new label with an unrelated change is a bit meh. Did @jancborchardt talk you into this? :D

violoncelloCH:
> > as discussed with @jancborchardt
> 
> For [#2222 (comment)](https://github.com/nextcloud/mail/pull/2222#issuecomment-552147470)?

no, actually this one is about the composer but #2222 is about the reading/displaying view...
this one is rather for #2220 to further improve the mobile composer view (and in general keep the interface clean)

src/components/Composer.vue
violoncelloCH: oh, I didn't even know that this way of doing it existed...
maybe it would even make sense to handle this in a vuex store? :thinking: 

anyway I'm happy to implement it the way you want :)

src/components/Composer.vue
violoncelloCH: hmm, @jancborchardt and me thought it would make sense to change this, because a checkbox saying "plain text" which reveals formatting possibilities when unchecked is only understandable for techie users. For accessibility reasons it would make sense to describe features in clear and precise language...
but of course if you want me to revert this I can do so or I could also create a separate PR to discuss this there if you prefer...?

jancborchardt: It was @violoncelloch's suggestion and I agree with it, also saying it could be added here since we talked about it in the same context. Yes it should be in a separate PR, my mistake. :)

ChristophWurst: Yes, @violoncelloCH, please do this in a separate PR. I'm fine with the change, just not when the PR is about *put composer attachment buttons into three dot menu* :wink:

violoncelloCH: let me do that :)

src/components/Composer.vue
violoncelloCH: so I implemented the second variant with the local event bus :)
hope this is fine now...

ChristophWurst: I'd only use Vuex for global state. These attachments are local state, only really relevant to these components.

src/components/Composer.vue
```diff
@@ -348,6 +358,16 @@ export default {
 		onInputChanged() {
 			this.saveDraftDebounced(this.getMessageData())
 		},
+		onAddLocalAttachment() {
+			this.bus.$emit('event', {
```
```suggestion
			this.bus.$emit('onAddLocalAttachment', {
```

src/components/ComposerAttachments.vue
```diff
@@ -136,6 +133,16 @@ export default {
 			this.$emit('input', this.value.filter(a => a !== attachment))
 		},
 	},
+	created() {
+		this.bus.$on('event', (args) => {
```
ChristophWurst:
```suggestion
		this.bus.$on('onAddLocalAttachment', (args) => {
```

:wink:

src/components/ComposerAttachments.vue
```diff
@@ -80,6 +77,14 @@ export default {
 			return ((uploaded / total) * 100).toFixed(1)
 		},
 	},
+	created() {
+		this.bus.$on('onAddLocalAttachment', args => {
+			this.onAddLocalAttachment()
+		})
+		this.bus.$on('onAddCloudAttachment', args => {
+			this.onAddCloudAttachment()
```
violoncelloCH: thanks @ChristophWurst ! Implemented your changes...
only thing I wasn't sure is if I could omit those `args` somehow which aren't used

ChristophWurst: Yes! ``this.bus.$on('onAddCloudAttachment', this.onAddCloudAttachment)``

violoncelloCH: ahh, makes sense... thank you!

PR #2247, opened 2019-11-12
Feature/noissue/plain text vs rich text checkbox

violoncelloCH:
this is a followup PR to #2226 
as discussed there I've split this change into a separate PR

- [x] Requires #2226


violoncelloCH:
makes sense! - changed that...

PR #2248, opened 2019-11-12
Label the unified inbox as "All inboxes"

### LINKED ISSUE #2217: Unified inbox should be labeled "Unified inbox"
Currently the unified inbox is just called "Inbox", which is a bit confusing or unclear. Ideally it should be labeled "Unified inbox" so that you directly know what it is. :)

@ChristophWurst if you point out the code where this needs to be changed, this would be a great good first issue.

StCyr:
I've simply used the existing "All inboxes" label. 

I can change it if you prefer "Unified inobx".

Fixes #2217 

### DIFF
--- a/src/l10n/MailboxTranslator.js
+++ b/src/l10n/MailboxTranslator.js
@@ -40,7 +40,11 @@ const translateSpecial = folder => {
 	}
 	if (folder.specialUse.includes('inbox')) {
 		// TRANSLATORS: translated mail box name
-		return t('mail', 'Inbox')
+		if (folder.isUnified === true) {
+			return t('mail', 'All inboxes')
+		} else {
+			return t('mail', 'Inbox')
+		}
 	}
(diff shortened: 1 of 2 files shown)


src/l10n/MailboxTranslator.js
```diff
@@ -40,7 +40,11 @@ const translateSpecial = folder => {
 	}
 	if (folder.specialUse.includes('inbox')) {
 		// TRANSLATORS: translated mail box name
-		return t('mail', 'Inbox')
+		if (folder.isUnified === true) {
+			return t('mail', folder.name)
```
ChristophWurst: This is a bit problematic. Translation strings are extracted statically from the source code.

Hence, could you please inline the `All inboxes` string so it's `return t('mail', 'All inboxes')`? Then we might also be able to remove the `name` property from the folder. I think it's not used anywhere else.

Issue #2187, opened 2019-10-27
"To" recipient field is not automatically focused for new message


jancborchardt:
On writing a new message via the "+ New message" button, the "to" recipient field should receive the focus so people can directly start typing the recipient.

The code in question is at https://github.com/nextcloud/mail/blob/master/src/components/Composer.vue#L96 → the autofocus needs to be removed from the subject and moved to the "to" field, and tested if it works. (Because it did not when it was on the subject field.)

ChristophWurst:
Why would that actually be the first input to focus? Who doesn't start with the recipient?

jancborchardt:
Good point. You are right, it should focus the "to" field first. :) Changed the title and text accordingly.

PR #2249, opened 2019-11-12
Focus on the "To" field when writing a new message.

StCyr:
Seems like the `autofocus` attribute doesn't work well with vue => 'had to call `focus()` in mounted() method

Fixes: #2187 

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -24,6 +24,7 @@
 			<Multiselect
 				id="to"
 				v-model="selectTo"
+				ref="toLabel"
 				:options="selectableRecipients"
 				:taggable="true"
 				label="label"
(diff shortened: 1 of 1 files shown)


src/components/Composer.vue
```diff
@@ -302,6 +302,9 @@ export default {
 
 		this.bodyVal = this.bodyWithSignature(this.selectedAlias, this.bodyVal)
 	},
+	mounted() {
+		this.$refs.toLabel.$el.focus()
```
ChristophWurst: Guess ``this.$refs.toLabel.focus()`` wouldn't work because it's a Vue component and not a DOM element. Nice trick :+1:

ChristophWurst:
>  [ChristophWurst](/ChristophWurst) changed the title ~FIX: Focus on the "To" field when writing a new message.~ Focus on the "To" field when writing a new message. [2 minutes ago](#event-2791929430)

@StCyr please don't prefix the title but use the labels: `bug` for bugfixes, `enhancement` for features :)

Issue #425, opened 2017-08-03
Identities for default account


fsedarkalex:
Hi another request...
I use the default account feature and it is great. One point I am missing is the ability to add identities to this default account.

This should be done by the user manually not by config.

Also mail signatures (per identity) would be great

andreas-p:
Missing for me too.

ChristophWurst:
> Missing for me too.

please use reactions like :+1: instead of comments

andreas-p:
I'm a little surprised that this issue isn't more popular, I'm seeing multiple identities everywhere around. Thumbs-upping probably won't help.

ChristophWurst:
> Thumbs-upping probably won't help.

how would the comment help?

> I'm a little surprised that this issue isn't more popular

it is always hard to guess what people need the most. Are you interested in helping?

andreas-p:
> > Thumbs-upping probably won't help.
> 
> how would the comment help?
Marking my me-too as spam isn't too friendly either...

I can live with the problem at the moment, but priority will probably get upped next year, depending on customer's requirements. I might look into that then (haven't worked with horde so far, nor do I know about its integration into NC).

ChristophWurst:
> depending on customer's requirements

Just let us know when you need help. [We're here to help](https://nextcloud.com/enterprise/)

mat-m:
@ChristophWurst Is it not a duplicate of #1722 ? Or, since this one is older, it should have been closed a while ago ?

ChristophWurst:
Yeah, actually in the past it wouldn't have been possible. But since https://github.com/nextcloud/mail/pull/2259 we can treat aliases of a provisioned account just like for any other account.

Closing as duplicate of #1722  thus.

PR #2259, opened 2019-11-15
Persist provisioned accounts

ChristophWurst:
*Default accounts* will be a bit tricky to handle in the future, as we will often need access to accounts, even if the user is not logged in (background jobs). Moreover, you can't persist any settings as they all share the same -2 id across all users.

Thus, the idea is to just persist them. Admins shall from now on configure this only through the web UI, hence adding a settings interface.

Todo

- [x] Add admin settings page
- [x] Add admin provisioning settings
  - [x] Basic settings form
  - [x] Preview for easier template creation
  - [x] Persist settings
  - [x] Update all existing provisioned accounts and create new for all missing
- [x] Provision new account when user is created
- [x] When a user accesses a provisioned account, set the (until then) missing password
- [x] Add repair step to also create these accounts from the old manual config
- [x] Prevent deletion of the default account

---

Current (2019-11-15) state

Bildschirmfoto von 2019-11-15 14-31-16


lib/Migration/MigrateProvisioningConfig.php
```diff
@@ -0,0 +1,86 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types=1);
+
+/**
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
```
rullzer: whoops

ChristophWurst: dang it, phpstorm!

andreas-p:
Came across this from closing #425 info: I wonder if I under stand the check "When a user accesses a provisioned account, set the (until then) missing password" right: The password for the (formerly called) default account will be stored in the database, just as it is done for quite a while with additional mail accounts?

ChristophWurst:
Yes, that is correct. For further questions please use https://help.nextcloud.com/c/apps/mail :)

Issue #2221, opened 2019-11-07
View Raw Source


HLFH:
### Feature Request

I should be able to view the raw source of a received email.

PR #2264, opened 2019-11-19
View raw message source

StCyr:
Should fix https://github.com/nextcloud/mail/issues/2221

TODO:

- [x] Effectively show raw message (with modal component?)
- [x] Add a nice icon
- [x] Also add action in Message component
- [x] Return raw message as html rather than json

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -267,6 +267,35 @@ class MessagesController extends Controller {
 	 * @param int $messageId
 	 *
 	 * @return HtmlResponse|TemplateResponse
+	 * @throws ServiceException
+	 */
+	public function getSource(int $accountId, string $folderId, int $messageId): JSONResponse {
+		$account = $this->accountService->find($this->currentUserId, $accountId);
+
+		$response = new JSONResponse([
+			'source' => $this->mailManager->getSource(
+				$account,
+				base64_decode($folderId),
+				$messageId
+			)
+		]);
+
+		// Enable caching
+		$response->cacheFor(60 * 60);
+
+		return $response;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 * @TrapError
+	 *
+	 * @param int $accountId
+	 * @param string $folderId
+	 * @param int $messageId
+	 *
+	 * @return HtmlResponse|TemplateResponse
 	 *
 	 * @throws ClientException
 	 */
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -27,6 +27,7 @@ use Horde_Imap_Client;
 use Horde_Imap_Client_Base;
 use Horde_Imap_Client_Data_Fetch;
 use Horde_Imap_Client_Exception;
+use Horde_Imap_Client_Exception_NoSupportExtension;
 use Horde_Imap_Client_Fetch_Query;
 use Horde_Imap_Client_Ids;
 use Horde_Imap_Client_Socket;
@@ -40,6 +41,7 @@ use OCP\ILogger;
 use function array_filter;
 use function array_map;
 use function iterator_to_array;
+use function reset;
 
 class MessageMapper {
 
@@ -256,6 +258,42 @@ class MessageMapper {
 				'add' => [$flag],
 			]
 		);
+	}
+
+	/**
+	 * @param Horde_Imap_Client_Socket $client
+	 * @param string $mailbox
+	 * @param int $id
+	 *
+	 * @return string|null
+	 * @throws ServiceException
+	 */
+	public function getSource(Horde_Imap_Client_Socket $client,
+							  string $mailbox,
+							  int $id): ?string {
+		$query = new Horde_Imap_Client_Fetch_Query();
+		$query->uid();
+		$query->fullText([
+			'peek' => true,
+		]);
+
+		try {
+			$result = iterator_to_array($client->fetch($mailbox, $query, [
+				'ids' => new Horde_Imap_Client_Ids($id),
+			]), false);
+		} catch (Horde_Imap_Client_Exception $e) {
+			throw new ServiceException("Could not fetch message source: " . $e->getMessage(), $e->getCode(), $e);
+		}
+
+		$msg = array_map(function (Horde_Imap_Client_Data_Fetch $result) {
+			return $result->getFullMsg();
+		}, $result);
+
+		if (empty($msg)) {
+			return null;
+		}
+
+		return reset($msg);
 	}
 
 	public function getHtmlBody(Horde_Imap_Client_Socket $client,
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -42,10 +42,20 @@
 						<ActionButton icon="icon-mail" @click="onToggleSeen">
 							{{ envelope.flags.unseen ? t('mail', 'Mark read') : t('mail', 'Mark unread') }}
 						</ActionButton>
+						<ActionButton
+							:icon="sourceLoading ? 'icon-loading-small' : 'icon-details'"
+							:disabled="sourceLoading"
+							@click="onShowSource"
+						>
+							{{ t('mail', 'View source') }}
+						</ActionButton>
 						<ActionButton icon="icon-delete" @click.prevent="onDelete">
 							{{ t('mail', 'Delete') }}
 						</ActionButton>
 					</Actions>
+					<Modal v-if="showSource" @close="onCloseSource">
+						<pre class="message-source">{{ rawMessage }}</pre>
+					</Modal>
 				</div>
 			</div>
 			<div class="mail-message-body">
@@ -65,6 +75,8 @@
 import Actions from '@nextcloud/vue/dist/Components/Actions'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
 import AppContentDetails from '@nextcloud/vue/dist/Components/AppContentDetails'
+import axios from '@nextcloud/axios'
+import Modal from '@nextcloud/vue/dist/Components/Modal'
 import {generateUrl} from '@nextcloud/router'
 
 import AddressList from './AddressList'
@@ -91,6 +103,7 @@ export default {
 		MessageAttachments,
 		MessageHTMLBody,
 		MessagePlainTextBody,
+		Modal,
 	},
 	data() {
 		return {
@@ -101,6 +114,9 @@ export default {
 			error: undefined,
 			replyRecipient: {},
 			replySubject: '',
+			rawMessage: '',
+			sourceLoading: false,
+			showSource: false,
 		}
 	},
 	computed: {
@@ -236,6 +252,27 @@ export default {
 				id: this.message.id,
 			})
 		},
+		async onShowSource() {
+			this.sourceLoading = true
+
+			try {
+				const resp = await axios.get(
+					generateUrl('/apps/mail/api/accounts/{accountId}/folders/{folderId}/messages/{id}/source', {
+						accountId: this.message.accountId,
+						folderId: this.message.folderId,
+						id: this.message.id,
+					})
+				)
+
+				this.rawMessage = resp.data.source
+				this.showSource = true
+			} finally {
+				this.sourceLoading = false
+			}
+		},
+		onCloseSource() {
+			this.showSource = false
+		},
 	},
 }
 </script>
@@ -346,6 +383,10 @@ export default {
 	margin-left: 4px;
 }
 
+.modal-container {
+	overflow-y: scroll !important;
+}
+
 @media print {
 	#mail-message-header-fields {
 		position: relative;
@@ -367,5 +408,10 @@ export default {
 	.mail-message-body {
 		margin-bottom: 0 !important;
 	}
+}
+
+pre.message-source {
+	font-family: monospace;
+	margin: 15px;
 }
 </style>
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -84,6 +84,11 @@ return [
 			'verb' => 'GET'
 		],
 		[
+			'name' => 'messages#getSource',
+			'url' => '/api/accounts/{accountId}/folders/{folderId}/messages/{messageId}/source',
+			'verb' => 'GET'
+		],
+		[
 			'name' => 'messages#getHtmlBody',
 			'url' => '/api/accounts/{accountId}/folders/{folderId}/messages/{messageId}/html',
 			'verb' => 'GET'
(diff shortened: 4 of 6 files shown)


ChristophWurst:
I'll try to find some time to look into this and see where we should put the logic. Hacking it into the existing already quite complex logic won't be the best choice in the long run.

https://github.com/nextcloud/mail/pull/2214 and #2064 will clean some of that logic up. We might want to wait with this feature until these large features are done :raised_hands: 

Hope this makes sense :v:

jancborchardt:
Design-wise:
- Should only be in the 3-dot menu when in the mail detail view, last item before "Delete".
- Wording "View source"
- Icon can be `.icon-details` (the info i)

StCyr:
> 
> 
> Design-wise:
> 
>     * Should only be in the 3-dot menu when in the mail detail view, last item before "Delete".
> 
>     * Wording "View source"
> 
>     * Icon can be `.icon-details` (the info i)

Fixed

jancborchardt:
Just a note @StCyr, since you are member of our organization on Github, you can also just work on branches of the main repository and don’t need a fork. :) This has the advantage that if any changes are needed or you want to hand over a pull request, other contributors can continue.

StCyr:
> 
> I'll try to find some time to look into this and see where we should put the logic. Hacking it into the existing already quite complex logic won't be the best choice in the long run.
> 
> #2214 and #2064 will clean some of that logic up. We might want to wait with this feature until these large features are done 🙌
> 
> Hope this makes sense ✌️

I let you devise if some refactoring is needed. But, for me, it's done.

ChristophWurst:
> Just a note @StCyr, since you are member of our organization on Github, you can also just work on branches of the main repository and don’t need a fork. :) This has the advantage that if any changes are needed or you want to hand over a pull request, other contributors can continue.

That's not true anymore. For a long time they changed this and you can now (unless explicitly opted out) push to the PR's fork branch as a maintainer of the destination repo. I use this regularly to clean up external contributor's branches. It works just fine adding the other user's fork as new git remote and pushing there.

jancborchardt:
> It works just fine adding the other user's fork as new git remote and pushing there.

Still much more complicated than a simple branch change. ;) All about removing the small barriers.

jancborchardt:
Nice! :) For me just 2 things are missing:
- [ ] "View source" entry should be _before_ "Delete", as Delete should always be the last element in the action list.
- [ ] Can we get the source displayed in a modal instead of a separate window and all. Not _so_ important but would be nice. :)

StCyr:
>     * [x]  Can we get the source displayed in a modal instead of a separate window and all. Not _so_ important but would be nice. :)

I've done this but Modal components don't have a vertical scrollbar and as such can't display the whole content :-(

jancborchardt:
> > ```
> > * [x]  Can we get the source displayed in a modal instead of a separate window and all. Not _so_ important but would be nice. :)
> > ```
> 
> I've done this but Modal components don't have a vertical scrollbar and as such can't display the whole content :-(

Oh they don’t? I guess only the filepicker component has that. But the modal component should for sure have that. Could you try adding `overflow-y: scroll;` manually to the relevant container? Let me know if I should look into it instead. :)

Also cc @skjnldsv @juliushaertl on how we could do this properly in the component – should be possible of course to have modals with scrollbar?

juliushaertl:
> Also cc @skjnldsv @juliushaertl on how we could do this properly in the component – should be possible of course to have modals with scrollbar?

The idea is that the modal size fits to the content size so it is more flexible to be used with a fixed content size like in the image viewer. Adding the scollbars with CSS in the apps scope is fine I'd say.

jancborchardt:
> The idea is that the modal size fits to the content size so it is more flexible to be used with a fixed content size like in the image viewer. Adding the scollbars with CSS in the apps scope is fine I'd say.

Right, I’m also thinking about e.g. the keyboard shortcuts list and the account settings. Both of these would work a bit better in a modal than the display they use right now, with left navigation still visible. But not fully relevant to this pull request, will open an issue in the components repo.

StCyr:
> Oh they don’t? I guess only the filepicker component has that. But the modal component should for sure have that. Could you try adding `overflow-y: scroll;` manually to the relevant container? Let me know if I should look into it instead. :)

it seems that the component applies a `overflow: hidden` style that overrides my settings:

image

jancborchardt:
> it seems that the component applies a `overflow: hidden` style that overrides my settings:

That’s because the style in the component is more specific as it has 2 classes. :) Simply change your rule to `.modal-wrapper .modal-container` so it has 2 classes as well. If that still doesn’t work (cause of the `data` attribute maybe), it’s fine in this case to use `overflow-y: scroll !important;` to override it.

Oh and yeah, only use overflow**-y** as we don’t want horizontal (-x) scrolling. :)

StCyr:
> If that still doesn’t work (cause of the `data` attribute maybe), it’s fine in this case to use `overflow-y: scroll !important;` to override it.

Yeah, indeed, I think the `data` attribute came into play during the tests I made.

I'll try  to make use of that `!important` stanza then

StCyr:
Ok, good for me

ChristophWurst:
> #2214 and #2064 will clean some of that logic up. We might want to wait with this feature until these large features are done raised_hands

Still blocked. But these two will be integrated soon. Will take a look at this one ASAP

lib/Model/IMAPMessage.php
```diff
@@ -153,6 +153,12 @@ public function setFlags(array $flags) {
 		throw new Exception('Not implemented');
 	}
 
+	/**
+	 */
+	public function getRawMessage() {
```
ChristophWurst: this should be a method of https://github.com/nextcloud/mail/blob/master/lib/IMAP/MessageMapper.php. See https://github.com/nextcloud/mail/blob/ec940371233c2c3630d5fb48e65e9e75203cbbc1/lib/IMAP/MessageMapper.php#L244-L309 for reference

StCyr: I shouldn't be using `$this->fetch->getFullMsg()` here. Is this also what you mean?

ChristophWurst: Yes, please don't add any more complexity to this class. This coding style is not testable. I'm migrating these methods into simple, testable service methods.

You can use `getFullMsg` in `MessageMapper`, though :)

StCyr:
> You can use getFullMsg in MessageMapper, though :)

Then my latest update to this PR might be good now :)

ChristophWurst:
I'm not sure, tbh. I tried to read the diff. But something is wrong with the commits. Could you please try to clean that up, so it's just one commit with *your* changes? :)

You moved the method from the message to the mapper, but the logic was unchanged. IMO it should be one dedicated method for the message source. We should not hack this into any other method. And the returned value shouldn't be an IMAPMessage but just a string, no?

Also, in the controller you don't have to catch all possible exceptions. It's fine to let some bubble up with the controller method is annotated with `@TrapError`. In Mail this means we'll catch the exception anyway, log it and return a proper JSON response.

StCyr:
HI @ChristophWurst 

I think I've corrupted my mail cache. Is there some command to reset it?

image

ChristophWurst:
Looks like it lost recipients? What happens if you refresh the page?

StCyr:
> 
> 
> Looks like it lost recipients? What happens if you refresh the page?

Refreshing the page doesn't help :)

bah. It's okay now. I only have 2 more interesting emails left in this state.

ChristophWurst:
So it was just a temporary bug? Or what do you mean by that?

Just to be clear: it's possible to clear the cache by dropping some data. But I'd really like to identify the problem first as this should be fixed before the release :grimacing:

StCyr:
> 
> 
> So it was just a temporary bug? Or what do you mean by that?

Oh, yes, I think it was due to a change I made while working on this PR. Change that I have now reverted.

> 
> Just to be clear: it's possible to clear the cache by dropping some data. But I'd really like to identify the problem first as this should be fixed before the release 😬

So, no need to clear the cache anymore, for what I'm concerned. Thanks :)

ChristophWurst:
Awesome :pray:

StCyr:
@ChristophWurst 

I think you might like this PR better now :)

lib/Controller/MessagesController.php
```diff
@@ -220,6 +220,55 @@ public function move($accountId, $folderId, $id, $destAccountId, $destFolderId):
...
+				$this->appName,
+				'error',
+				['message' => 'Not allowed'],
+				'none'
+			);
+		}
+
+		$htmlResponse = new HtmlResponse(
+			nl2br(
+				$this->mailManager->getRawMessage(
```
ChristophWurst: As Scrutinizer says: `The method ``getRawMessage()`` does not exist on ``OCA\Mail\Contracts\IMailManager``. Since it exists in all sub-types, consider adding an abstract or default implementation to ``OCA\Mail\Contracts\IMailManager``.`

lib/Controller/MessagesController.php
```diff
@@ -220,6 +220,55 @@ public function move($accountId, $folderId, $id, $destAccountId, $destFolderId):
...
+	/**
+	 * @NoAdminRequired
+ 	 * @NoCSRFRequired
+	 * @TrapError
+	 *
+	 * @param int $accountId
+	 * @param string $folderId
+	 * @param int $messageId
+	 *
+	 * @return JSONResponse
```
ChristophWurst: remove this, otherwise returning a `TemplateResponse` is incorrect

lib/Controller/MessagesController.php
```diff
@@ -220,6 +220,55 @@ public function move($accountId, $folderId, $id, $destAccountId, $destFolderId):
...
+					$account,
+					base64_decode($folderId),
+					$messageId
+				)
+			)
+		);
+
+		// Harden the default security policy
+		$policy = new ContentSecurityPolicy();
+		$policy->allowEvalScript(false);
```
ChristophWurst: this is deprecated. we shouldn't use this method. Why do we need a custom CSP here btw?

lib/IMAP/MessageMapper.php
```diff
@@ -241,6 +241,30 @@ public function addFlag(Horde_Imap_Client_Socket $client,
...
+		$msg = array_map(function (Horde_Imap_Client_Data_Fetch $result) use ($client, $mailbox) {
+			return new IMAPMessage(
+				$client,
+				$mailbox,
+				$result->getUid(),
+				$result
+			);
+		}, $result);
+
+		return $msg[0]->getRawMessage();
```
ChristophWurst: check for empty `$msg` otherwise this will do :boom: when the UID doesn't exist :)

lib/Model/IMAPMessage.php
```diff
@@ -155,6 +155,10 @@ public function setFlags(array $flags) {
 		throw new Exception('Not implemented');
 	}
 
+	public function getRawMessage() {
+		return $this->fetch->getFullMsg();
```
ChristophWurst: not used, please remove this method

src/components/Message.vue
```diff
@@ -46,10 +46,18 @@
...
 							{{ envelope.flags.unseen ? t('mail', 'Mark read') : t('mail', 'Mark unread') }}
 						</ActionButton>
+						<ActionButton icon="icon-details" @click="onShowSource">
+							{{ t('mail', 'View source') }}
+						</ActionButton>
 						<ActionButton icon="icon-delete" @click="onDelete">
 							{{ t('mail', 'Delete') }}
 						</ActionButton>
 					</Actions>
+					<Modal v-if="showSource" @close="onCloseSource">
```
ChristophWurst: in a follow-up we should extract this into an independent component

lib/Controller/MessagesController.php
StCyr: I've just copied what's done in `getHtmlBody()`.

Should I remove this?

lib/Model/IMAPMessage.php
StCyr: It is used, I'm sure.

src/components/Message.vue
StCyr: that's too much for me I believe ;)

lib/Controller/MessagesController.php
StCyr: Could you please confirm that my change correctly addresses your concern?

lib/Controller/MessagesController.php
ChristophWurst: yes please

lib/IMAP/MessageMapper.php
```diff
@@ -241,6 +241,34 @@ public function addFlag(Horde_Imap_Client_Socket $client,
...
+			return new IMAPMessage(
+				$client,
+				$mailbox,
+				$result->getUid(),
+				$result
+			);
+		}, $result);
+
+		if (!empty($msg)) {
+			return $msg[0]->getRawMessage();
```
ChristophWurst: so this is where `getRawMessage` is used. Can't we restructure this method to be independent of the message class? see `getHtmlBody` below.

lib/Model/IMAPMessage.php
ChristophWurst: found it :) https://github.com/nextcloud/mail/pull/2264/files#r395078054

lib/Controller/MessagesController.php
StCyr: Ok, done. Should I remove this $policy thing from `getHtmlBody` too?

lib/IMAP/MessageMapper.php
StCyr: Indeed, thanks for your thoroughfull review.

It should also be possible, I guess, to improve this method a bit so as not to use `iterator_to_array` and `array_map` as `$id` is just a simple int. What do you think?

ChristophWurst: `$id` is just an int, yep, but the horde method is designed to handle more than one result at a time, hence it always gives you an iterable IIRC.

lib/IMAP/MessageMapper.php
```diff
@@ -241,6 +241,29 @@ public function addFlag(Horde_Imap_Client_Socket $client,
...
+		]), false);
+
+		$msg = array_map(function (Horde_Imap_Client_Data_Fetch $result) {
+			return $result->getFullMsg();
+		}, $result);
+
+		if (!empty($msg)) {
+			return $msg[0];
+		} else {
+			return "Error: Could not get message source";
```
ChristophWurst: let's just return `null` when there is no result

you can just `return reset($msg)` to get the value of the first element in the array https://www.php.net/manual/en/function.reset.php

lib/Controller/MessagesController.php
ChristophWurst: Yep, looks good now :)

ChristophWurst:
Pushed a few minor fixes

* Remove use of dangerous `v-html` and use `<pre>` instead
* Show a loading spinner whlie the messages is fetched (this takes time for large messages or ones with attachments)
* Rename *raw message* to *message source* in the back-end
* Minor code cleanups

PR #2267, opened 2019-11-21
Add php7.4 support

ChristophWurst:
Not even sure if Travis or the linter have it already but let's see

### DIFF
--- a/lib/Service/Attachment/AttachmentStorage.php
+++ b/lib/Service/Attachment/AttachmentStorage.php
@@ -71,9 +71,9 @@ class AttachmentStorage {
 	public function save(string $userId, int $attachmentId, UploadedFile $uploadedFile) {
 		$folder = $this->getAttachmentFolder($userId);
 
-		$file = $folder->newFile($attachmentId);
+		$file = $folder->newFile((string) $attachmentId);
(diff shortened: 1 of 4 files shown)


lib/Service/Attachment/AttachmentStorage.php
```diff
@@ -71,9 +71,9 @@ private function getAttachmentFolder($userId): ISimpleFolder {
 	public function save(string $userId, int $attachmentId, UploadedFile $uploadedFile) {
 		$folder = $this->getAttachmentFolder($userId);
 
-		$file = $folder->newFile($attachmentId);
+		$file = $folder->newFile((string) $attachmentId);
 		$tmpPath = $uploadedFile->getTempPath();
-		if (is_null($tmpPath)) {
+		if ($tmpPath === null) {
```
georgehrke: Is this a general change or required for PHP 7.4? oO

ChristophWurst: Depends on how you want to see it. The API wants a `string` but doesn't type-hint nor check that. And we later do an array access on the first character. If you pass in php7.4+ will rightfully complain that this doesn't make sense.

So, yeah, to run the code you have to pass in the correct value. But it's not a breaking change if you already use the API correctly.

PR #2268, opened 2019-11-21
Fix ical import

ChristophWurst:
* Fix limited content type filter
* Fix Axios sending invalid requests

Thanks @georgehrke!

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -443,7 +443,7 @@ class MessagesController extends Controller {
 	 * @return boolean
 	 */
 	private function attachmentIsCalendarEvent(array $attachment): bool {
-		return $attachment['mime'] === 'text/calendar';
+		return in_array($attachment['mime'],  ['text/calendar', 'application/ics'], true);
 	}
 
 }
--- a/src/service/DAVService.js
+++ b/src/service/DAVService.js
@@ -101,12 +101,10 @@ const getCalendarData = properties => {
 export const getUserCalendars = () => {
 	var url = generateRemoteUrl('dav/calendars') + '/' + getCurrentUser().uid + '/'
 
-	return getRequestToken()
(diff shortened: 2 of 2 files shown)


lib/Controller/MessagesController.php
```diff
@@ -443,7 +443,7 @@ private function attachmentIsImage(array $attachment): bool {
 	 * @return boolean
 	 */
 	private function attachmentIsCalendarEvent(array $attachment): bool {
-		return $attachment['mime'] === 'text/calendar';
+		return in_array($attachment['mime'],  ['text/calendar', 'application/ics'], true);
```
georgehrke: Why `'application/ics'`?

ChristophWurst: That's the mime set for some emails apparently

src/service/DAVService.js
```diff
@@ -206,15 +204,11 @@ export const importCalendarEvent = url => data => {
...
-					Axios.put(
-						url + getRandomString(),
-						{data},
-						{
-							headers: {
-								contentType: 'text/calendar; charset=utf-8',
-							},
-						}
-					)
+					Axios.put(url + getRandomString(), component, {
```
georgehrke: Please add '.ics' to the filename :)

lib/Controller/MessagesController.php
ChristophWurst:
```
Content-Type: application/ics; name=meetup.ics; method=PUBLISH; 
        charset=UTF-8
Content-Disposition: attachment; filename=meetup.ics
```

ChristophWurst: https://gitlab.freedesktop.org/xdg/shared-mime-info/commit/9d73fc74d9386333de1ba49ffe7ba991ef498c37

I think this is fine :)

PR #2271, opened 2019-11-21
Fix unexpected linebreaks in mails, responsive text flow, fix #2181

### LINKED ISSUE #2181: Mails unexpectedly get line breaks
Whenever you write a mail and then save it as draft, it is saved with line breaks after 80 characters. This is very annoying as it basically destroys your text in an unfinished state.

Same thing when sending mail. You don’t expect it to be broken, and it also leads to issues when e.g. people have narrower displays or when people quote. Then it looks like this:

> Hey look at me I am writing
> an
> email and I sent it with an
> email
> app which does automatic line
> breaking
> looks great right?

jancborchardt:
Fix https://github.com/nextcloud/mail/issues/2181 and use proper responsive text flow, as most other email apps do. See research in comment https://github.com/nextcloud/mail/issues/2181#issuecomment-554183882 

## Before
Wrapping at 78 characters. Breaks quickly on mobile devices or when there’s more than 1 reply. Is also entirely unexpected:
mail wrap before

## After
Sends the mail just like you see it and expect it. And it responsively adjusts to the screen size of the client:
Mail wrap after

### DIFF
--- a/src/util/HtmlHelper.js
+++ b/src/util/HtmlHelper.js
@@ -17,7 +17,7 @@ export const htmlToText = html => {
 		noLinkBrackets: true,
 		ignoreHref: true,
 		ignoreImage: true,
-		wordwrap: 78, // 80 minus '> ' prefix for replies
+		wordwrap: false,
 		format: {
 			blockquote: function(element, fn, options) {
 				return fn(element.children, options)


ChristophWurst:
Please leave the milestone assignments to me :raised_hands:

PR #2272, opened 2019-11-22
Patch out version check on nightly builds

ChristophWurst:


.travis.yml
```diff
@@ -5,10 +5,9 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - nightly
-jobs:
-  allow_failures:
-  - php: nightly
+#jobs:
+#  allow_failures:
```
ChristophWurst: Usually I would allow failure here but since it passes I would leave it like this for now :)

Issue #1032, opened 2018-08-14
Show overlay star only for favorited mails, and action inside 3-dot-menu


jancborchardt:
Currently the list shows a favorite icon for every single mail:
![screenshot from 2018-08-14 13-25-55](https://user-images.githubusercontent.com/925062/44089023-98399b3e-9fc5-11e8-9764-f421e6dfc1d3.png)

This becomes crowded very quickly and it’s not easy to distinguish which mails are actually favorited.

Better would be to do it like in Files:
- The overlay icon (yellow) is only there for items which are actually favorited
- Favoriting and unfavoriting happens in the 3-dot menu
![screenshot from 2018-08-14 13-23-57](https://user-images.githubusercontent.com/925062/44089084-c96ff4b4-9fc5-11e8-915e-4ed9444913b2.png)

ChristophWurst:
@skjnldsv do we have any (documented) standard for this? :wink:

skjnldsv:
@ChristophWurst about the menu?
I don't like what @jancborchardt suggest because I feel like we keep adding steps to simple actions (click the menu, click the favorite...) But it's just a matter of adding the favorite entry in the menu and only displaying the favorited entry when on! ;)

ChristophWurst:
No, about

> The overlay icon (yellow) is only there for items which are actually favorited

So that we can hide the inactive stars on desktop at least.

:-1: on moving the action to the menu, it's too important to add another step.

skjnldsv:
@ChristophWurst I don't get your request, you want to just hide it totally by default?
What standard are you looking for?

ChristophWurst:
Sorry. If I understand Jan's issue correctly, it bothers him that the star icon is shown for every message, even the ones that are not flagged. I would suggest to hide the icon on desktop browsers and just show it on hover. Does that make sense?

skjnldsv:
Ah yes it does now! :)
I'll let you debate with Jan for this, but there is no standards, you need to add this manually ;)

jancborchardt:
Sorry about adding steps and such, as you know I’m also not a fan of any unnecessary complexities.

But as said:

> it’s not easy to distinguish which mails are actually favorited.

And this unfortunately makes the entire feature kind of useless in the current form. ;)  And not only on desktop, also on mobile.

We can also talk about this at the conf.

ChristophWurst:
> And this unfortunately makes the entire feature kind of useless in the current form. ;) And not only on desktop, also on mobile.

[...]

Issue #2273, opened 2019-11-22
Actions on multiple messages


StCyr:
### Feature Request

I would like to be able to perform actions (eg: delete) on multiple messages at once.

#### Summary

It looks a bit complicated to implement at first sight. So, I'd like to have some help here :-)

Here's a proposal to start the discussion.

#### Proposal 

What I think is necessary to implement to get this functionality is:

1. Implement a new store object/array to keep track of the currently selected messages 
2. Be able to divert CTRL+click and SHIFT+click from their regular behaviour in the Envelope/EnvelopList component (=> to update the above-mentioned store object for example)
3. New component to display in place of the Message/NewMessageDetail/NoMessageSelected components when multiple messages are selected. This component would show the possible actions (inspired by OWA)

#### Decision

1. The tracking of the currently selected messages must be implemented with a property of the EnvelopeList component;
2. Show an "action header" at the top of the EnvelopeList component rather than create a new component to show the possible actions.

#### TODO

- [x] create a selection array in the envelopeList component
- [x] handle ctrl/shift/meta+click events in the envelope component
- [x] show an action header at the top of the envelopeList component when multiple 
- [ ] style the action header better
- [ ] make the header's position fixed
- [X] implement "mark read/unread" actions
- [x] implement "delete" action
- [ ] automatically add the currently viewed message to the selection list
- [ ] clear the selection list when needed
- [ ] verify the algorithm used to populate the selection property
- [ ] proper icons for "mark read" and "mark unread"

ChristophWurst:
I'm pretty sure that is at least one existing ticket for this. But I can't find it right now. :laughing: 

IMO the selection is a local state. It's only relevant to the component that renders the list. This shouldn't go into a store.

Generally speaking this functionality isn't specific to this app. It doesn't make sense to have a specific implementation for every app that renders a list. Instead, this should be added to the https://github.com/nextcloud/nextcloud-vue components (https://github.com/nextcloud/nextcloud-vue/pull/616). And then apps can specify if actions are available for multi select.

:+1: on the CTRL+click behavior for power users on a desktop system.

Not sure about the 3). It could also be fine to just keep showing the previously shown message. It definitely shouldn't be the (only) place for the actions, as this element is not visible on mobile.

StCyr:
> I'm pretty sure that is at least one existing ticket for this. But I can't find it right now. 😆

I didn't find neither. Strange uh?

> 
[...]

PR #2278, opened 2019-11-23
Move favorite toggle to menu, otherwise not distinguishable

jancborchardt:
Fix https://github.com/nextcloud/mail/issues/1032

Before, basically every mail looked like it was favorited. Now we only show a star for mails which really are favorited. And the favorite toggle is moved to the menu.

This aligns it with how we do it in Files as well. :)

Before & after:
favorites list before | | | favorites now

Menu:
favorite menu

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -6,6 +6,7 @@
 			:style="{'background-color': accountColor}"
 		></div>
 		<div
+			v-if="data.flags.flagged"
 			class="app-content-list-item-star icon-starred"
 			:data-starred="data.flags.flagged ? 'true' : 'false'"
 			@click.prevent="onToggleFlagged"
@@ -28,6 +29,9 @@
 			<Moment :timestamp="data.dateInt" />
 		</div>
 		<Actions class="app-content-list-item-menu" menu-align="right">
+			<ActionButton icon="icon-starred" @click.prevent="onToggleFlagged">{{
+				data.flags.flagged ? t('mail', 'Unfavorite') : t('mail', 'Favorite')
+			}}</ActionButton>
 			<ActionButton icon="icon-mail" @click.prevent="onToggleSeen">{{
 				data.flags.unseen ? t('mail', 'Mark read') : t('mail', 'Mark unread')
 			}}</ActionButton>


ChristophWurst:
> Before, basically every mail looked like it was favorited.

Did it? There is a yellow star for starred items and a grey one for the rest.

IMO this is a primary action. I already kind of dislike that the delete button is so hidden. This makes fast mail triaging harder.

StCyr:
> 
> 
> > Before, basically every mail looked like it was favorited.
> 
> Did it? There is a yellow star for starred items and a grey one for the rest.

I also never got confused with this behaviour. And, indeed, I like being able to favorite a mail with a single click on the avatar 'star.

But I'm not against adding this action to the Actions menu also.

> 
> IMO this is a primary action. I already kind of dislike that the delete button is so hidden. This makes fast mail triaging harder.

That, on the other hand, never frustated me :-)

And, I think that we should implement actions on multiple selection if we want to support faster mail triaging (see https://github.com/nextcloud/mail/issues/2273)

StCyr:
ouch! 'just implemented the PR... .I don't like it :-(

jancborchardt:
For fast mail triaging, there are the shortcuts, no? If we make the actions visible so presently it will be very crowded.

> > Before, basically every mail looked like it was favorited.
> 
> Did it? There is a yellow star for starred items and a grey one for the rest.

Yes it does. :) _Every_ mail has a star. And even if you do have one or 2 mails favorited, you can’t go by the additional present element, but by the color of an element.

ChristophWurst:
> For fast mail triaging, there are the shortcuts, no?

Usually I do not have a keyboard attached to my phone when I read my mail :)

StCyr:
Shortcuts aren’t documented IINM.

I never use them (and don’t expect users to discover them themselves) 

EDIT: I’ve just seen that they are documented. Anyway, I didn’t know that until now and that’s why I was not using them.
I expect lambda users to have the same experience as I had.

jancborchardt:
> Usually I do not have a keyboard attached to my phone when I read my mail :)

I see you want to implement swipe shortcuts? ;) Swipe right to favorite, swipe left to archive/delete in that case.

Seriously though, I would then replace the "unstarred" state icon with an outline icon instead of filled grey, so it’s really distinguishable from the starred state. Makes sense?

ChristophWurst:
> E.g. if you delete 3 mails after another, or star 3 mails in succession etc, it could show a tooltip or so with the shortcut?

Can't help myself but think about

jancborchardt:
@ChristophWurst yes I know it sounds ridiculous. ;) But when done right, learnable UI is pretty nice. :)

PR #2292, opened 2019-11-30
doc: admin: add note about content security policy headers

nodiscc:


ChristophWurst:
Hi,

could you elaborate when this is a problem? Did you supply a stricter CSP than the one added to Nextcloud responses?

nodiscc:
Hi, sorry for the late reply, yes I set this CSP header as a baseline [[2](https://gitlab.com/nodiscc/ansible-xsrv-lamp/blob/master/templates/etc_apache2_conf-available_security.conf.j2#L111)] for all my web servers:

```apache
Header always set Content-Security-Policy "script-src 'self' 'unsafe-inline'; frame-ancestors 'none'"
```

This value (or even something more restrictive) is recommended by many hardening guides/tools [[example for nginx](https://github.com/dev-sec/ansible-nginx-hardening/blob/master/defaults/main.yml#L22)]. Nextcloud itself works fine with it because it overrides the CSP directive [[3](https://github.com/owncloud/core/issues/11925)], but as far as I know there is no way for an app to do that. Hence I think documenting it is the best we can do.

doc/admin.md
```diff
@@ -71,6 +71,11 @@ You can use the php-mail function to send mails. This is needed for some webhost
 ```php
 'app.mail.transport' => 'php-mail'
 ```
+### Content Security Policy
+This app uses iframes to display mail contents. If mail bodies do not load, check your browser console, and make sure your webserver's [Content Security Policy](https://en.wikipedia.org/wiki/Content_Security_Policy) headers allow displaying iframes if they have the same origin as the parent page (`frame-ancestors`). For example on Apache:
+```apache
+Header always set Content-Security-Policy "script-src 'self' 'unsafe-inline'; frame-ancestors 'self'"
```
rullzer: Please don't do this. Nextcloud adds it own CSP headers. Adding more here is not a good idea. If the CSP has to be modified we should do it in the response from the server.

rullzer:
> Hi, sorry for the late reply, yes I set this CSP header as a baseline [[2](https://gitlab.com/nodiscc/ansible-xsrv-lamp/blob/master/templates/etc_apache2_conf-available_security.conf.j2#L111)] for all my web servers:
> 
> ```apache-config
> Header always set Content-Security-Policy "script-src 'self' 'unsafe-inline'; frame-ancestors 'none'"
> ```
> 
> This value (or even something more restrictive) is recommended by many hardening guides/tools [[example for nginx](https://github.com/dev-sec/ansible-nginx-hardening/blob/master/defaults/main.yml#L22)]. Nextcloud itself works fine with it because it overrides the CSP directive [[3](https://github.com/owncloud/core/issues/11925)], but as far as I know there is no way for an app to do that. Hence I think documenting it is the best we can do.

Yes apps can do that ;) And they should if needed.
What exactly needs to change here?

nodiscc:
>  What exactly needs to change here?

The app needs to add `frame-ancestors 'self'` to the Content-Security-Policy header.

ChristophWurst:
> The app needs to add `frame-ancestors 'self'` to the Content-Security-Policy header.

This is already the case :)

nodiscc:
>This is already the case :)

I don't think it is... can you point me to the part of the app that sets the header? https://github.com/nextcloud/mail/search?q=frame-ancestors&unscoped_q=frame-ancestors -> no results

PR #2336, opened 2019-12-09
Make it possible to change the order of accounts in the sidebar

ChristophWurst:
Right now we show the accounts in the order they were added, because the database sorts them by primary key. However, as a user I want to change the order. Adding an account later doesn't mean it's less important.

Todo
- [x]  front-end @GretaD 
  - [x] sort accounts by `order` attribute in the sidbar
  - [x] add a "Move up" and "Move down" action in the account's action menu
  - [x] update the account order attribute locally
  - [x] send the attribute change to the server
  - [x] The first account should not have the button _move up_
  - [x] The last account should not have the button _move down_
  - [x]  _All inboxes_ is defined as _an account_, we should find a way how to exclude it and not to be count as an account.
- [x] back-end @ChristophWurst 
  - [x] add a `order` attribute to the JSON of all accounts
  - [x] add `PATCH` support for this new attribute
  - [x] create a new column in the DB
  - [x] update the row on `PATCH`
- [x] bugs
  - [x] Move doesn't change anything on the second try

### DIFF
--- a/lib/Migration/Version0210Date20191212144925.php
+++ b/lib/Migration/Version0210Date20191212144925.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version0210Date20191212144925 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$accountsTable = $schema->getTable('mail_accounts');
+		$accountsTable->addColumn('order', 'integer', [
+			'notnull' => true,
+			'length' => 4,
+			'default' => 1,
+		]);
+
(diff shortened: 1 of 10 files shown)


GretaD:
To be done:

- The first account should not have the button _move up_
- The last account should not have the button _move down_
- _All inboxes_ is defined as _an account_, we should find a way how to exclude it and not to be count as an account.

lib/Migration/Version0210Date20191212144925.php
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$accountsTable = $schema->getTable('mail_accounts');
+		$accountsTable->addColumn('order', 'integer', [
+			'notnull' => true,
+			'length' => 4,
+			'default' => 0,
```
ChristophWurst: - [ ] TODO: 1

ChristophWurst:
@nextcloud/mail review time :)

src/components/Navigation.vue
```diff
@@ -110,6 +116,14 @@ export default {
 				},
 			})
 		},
+		isFirst(account) {
+			const accounts = this.$store.getters.getAccounts()
+			return account === accounts[1]
+		},
+		isLast(account) {
+			const accounts = this.$store.getters.getAccounts()
```
skjnldsv: getters are usually not functions :)
https://vuex.vuejs.org/guide/getters.html#property-style-access

ChristophWurst: yep. for consistency reasons I'll shoot a follow-up PR in a sec

ChristophWurst: it's used in the same style in other occasions in this file

src/store/mutations.js
```diff
@@ -19,9 +19,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import identity from 'lodash/fp/identity'
```
skjnldsv: :+1:

src/store/mutations.js
```diff
@@ -39,14 +37,23 @@ const addFolderToState = (state, account) => folder => {
...
 		Vue.set(state.preferences, key, value)
 	},
 	addAccount(state, account) {
 		account.collapsed = true
 		Vue.set(state.accounts, account.id, account)
-		Vue.set(state, 'accountList', sortBy(identity)(state.accountList.concat([account.id])))
+		Vue.set(
+			state,
+			'accountList',
+			sortAccounts(state.accountList.concat([account.id]).map(id => state.accounts[id])).map(a => a.id)
```
skjnldsv: a bit hard to understand all those functions.
Adding a step with a const and some comments would helps a lot for this file :)

src/components/Navigation.vue
skjnldsv: :ok_woman:

src/components/NavigationAccount.vue
```diff
@@ -43,6 +43,12 @@
 			<ActionInput icon="icon-add" @submit="createFolder">
 				{{ t('mail', 'Add folder') }}
 			</ActionInput>
+			<ActionButton v-if="!isFirst" icon="icon-triangle-n" @click="changeAccountOrderUp">
+				{{ t('mail', 'Move Up') }}
```
jancborchardt: We always only capitalize the first word. :) So "Move up". Strangely enough, "Move down" is correct. ;)

ChristophWurst: Send a PR? :)

GretaD: good catch,thats definitely a typo, my typo, sorry.

jancborchardt: @GretaD no worries – could you fix it? :)

Issue #2337, opened 2019-12-09
Unable to save config


xf-:
### Expected behavior
Save config, use mail plugin

### Actual behavior
Unable to save the mail Adminconfig (after updating from 0.19 to 0.20 via "Apply and create/update for all users".

### Mail app

**Mail app version:** 0.20.0

**Mailserver or service:** Postfix & Dovecot

### Server configuration

**Operating system**: (Debian 10

**Web server:** Nginx

**Database:** PostgreSQL

**PHP version:** (e.g. 7.3, because Nextcloud v17 is incompatible with PHP7.4

**Nextcloud Version:**17.0.1

#### Client configuration
**Browser:** Firefox 

**Operating system:** Arch Linux

### Log

```
[index] Error: Exception: Authenticated ciphertext could not be decoded. at <<closure>>

0. PATHa/pps/mail/lib/Service/Provisioning/Manager.php line 166
   OC\Security\Crypto->decrypt("*** sensitive parameters replaced ***")
1. PATH/apps/mail/lib/Http/Middleware/ProvisioningMiddleware.php line 71
   OCA\Mail\Service\Provisioning\Manager->updatePassword("*** sensitive parameters replaced ***")
2. PATH/lib/private/AppFramework/Middleware/MiddlewareDispatcher.php line 95
   OCA\Mail\Http\Middleware\ProvisioningMiddleware->beforeController(OCA\Mail\Controller\SettingsController {}, "provisioning")
3. PATH/lib/private/AppFramework/Http/Dispatcher.php line 97
   OC\AppFramework\Middleware\MiddlewareDispatcher->beforeController(OCA\Mail\Controller\SettingsController {}, "provisioning")
4. PATH/lib/private/AppFramework/App.php line 126
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\Mail\Controller\SettingsController {}, "provisioning")
5. PATH/lib/private/AppFramework/Routing/RouteActionHandler.php line 47
   OC\AppFramework\App::main("OCA\\Mail\\Cont ... r", "provisioning", OC\AppFramework\ ... {}, {_route: "mail.settings.provisioning"})
6. <<closure>>
   OC\AppFramework\Routing\RouteActionHandler->__invoke({_route: "mail.settings.provisioning"})
7. PATH/lib/private/Route/Router.php line 297
   undefinedundefinedcall_user_func(OC\AppFramework\ ... {}, {_route: "mail.settings.provisioning"})
[...]

PR #2338, opened 2019-12-09
Fix empty password for provisioned account

ChristophWurst:
Passwords can't just be `null` but also an empty string. The code did
not handle that properly. Now checking against `empty` so both `null`
and `''` trigger shortcut evaluation and prevent decrypting of an
invalid value.

Fixes https://github.com/nextcloud/mail/issues/2337


ChristophWurst:
Thanks for the quick review. You rock :rocket:

PR #2347, opened 2019-12-11
Fix accidental provisioning when it's actually disabled

ChristophWurst:
Of provisioning is disabled, we keep the config but add `active: false` to the json. During the repair step, however, this prop wasn't checked, so the app created accounts nevertheless.

cc @nextcloud/mail


ChristophWurst:
@nextcloud/mail please review

PR #2371, opened 2019-12-17
Access accounts getter as property

ChristophWurst:
Makes Vuex cache the result and doesn't re-evaluate all the time.

So this is a performance improvement.

Thanks @skjnldsv for the hint!

@nextcloud/mail please review

### DIFF
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -57,7 +57,7 @@ export default {
 	fetchAccounts({commit, getters}) {
 		return fetchAllAccounts().then(accounts => {
 			accounts.forEach(account => commit('addAccount', account))
-			return getters.getAccounts()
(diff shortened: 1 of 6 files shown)


src/store/actions.js
```diff
@@ -155,7 +155,7 @@ export default {
 			// TODO: handle short/ending streams and show their last element as well
 			return Promise.all(
 				getters
```
skjnldsv: What's going on with this eslint comment? :p

skjnldsv: Ooh, `⏎↹↹↹↹↹` means new line and 4 tabs! Not very straightforward :p

ChristophWurst: fixed

PR #2379, opened 2019-12-17
Close #2375

### LINKED ISSUE #2375: "Delete account" should be last entry in account menu
In the account menu in the app navigation, "Delete account" is the second entry for each account.

Instead it should always be last. As always, the destructive action is last in a list.

@ChristophWurst if you point out where this would have to be changed, this would make for a nice "good first issue". :)

sayashraaj:
moved delete option at last

Fixes #2375

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -127,6 +127,17 @@ export default {
 					throw error
 				})
 		},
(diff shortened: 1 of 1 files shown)


sayashraaj:
Thanks, I will work on it!

PR #2384, opened 2019-12-18
fix image hover effect on attachment

### LINKED ISSUE #2365: No pointer cursor when hovering attachment images
If I receive a attachment it looks like this:

image

There are the two buttons of the obvious actions "download" and "save to files" but the whole attachment is also clickable. 

First, the option to click on the attachment is only indicated by a mouse pointer change over the empty area, there is no hover effect if I point at the picture but I can still click it.

Second, the click triggers a download. Would it make sense to change it to "save to files" as the default behavior? Is there any way to configure the behavior?

GretaD:
fixes #2365

### DIFF
--- a/src/components/MessageAttachment.vue
+++ b/src/components/MessageAttachment.vue
@@ -216,6 +216,10 @@ export default {
 	display: block;
 	max-width: 100%;
 	max-height: 120px;
+	cursor: pointer;
+}
+.mail-attached-image:hover {
+	opacity: 0.8;
 }
 
 .attachment-save-to-cloud,


GretaD:
> Screenshots of before/after are always appreciated for quick design review. :)


before
photo6039753462163550791
after (you cannot notice well the opacity hover, but its there :) ) printscreen will remove the cursor thats why i took a pic :D
photo6039753462163550792

ChristophWurst:
> And to make sure this works with keyboard navigation, a line with `:focus` needs to be included.

Correct me if I'm wrong. But we want to change the cursor when you hover the element. You can't trigger this with the keyboard, right?

GretaD:
> > And to make sure this works with keyboard navigation, a line with `:focus` needs to be included.
> 
> Correct me if I'm wrong. But we want to change the cursor when you hover the element. You can't trigger this with the keyboard, right?

Actually you are right, the focus with the keyboard is only for the download and save button and that already exist. You cannot trigger it with the keyboard for the image. So i'm reverting the last line

jancborchardt:
> Correct me if I'm wrong. But we want to change the cursor when you hover the element. You can't trigger this with the keyboard, right?

Yes, and the cursor change is done in `.mail-attached-image`, without any :hover or :focus. It is separate rules. :) I am talking about the :hover rule which triggers the opacity change.

> Actually you are right, the focus with the keyboard is only for the download and save button and that already exist. You cannot trigger it with the keyboard for the image.

Then we need to keep this in mind for when we have Viewer integration, cause otherwise you would have no feedback for when you can open the file. :)

jancborchardt:
> printscreen will remove the cursor thats why i took a pic :D

On Ubuntu you should have the program "Screenshot" installed, it comes with the option to include the cursor. :)

PR #2403, opened 2019-12-19
Fix automatic provisioning (for new users)

ChristophWurst:
This makes sure that every user has a provisioned account when Mail is accessed.

# Before
new user didn't have provisioned account until admin provisioned for all users

# After
also new users get their account as soon as they access the app

In the future we can use https://github.com/nextcloud/server/pull/18348 to create the account right when the user is created. But it doesn't make a difference at the moment.

### DIFF
--- a/lib/Http/Middleware/ProvisioningMiddleware.php
+++ b/lib/Http/Middleware/ProvisioningMiddleware.php
@@ -65,7 +65,12 @@ class ProvisioningMiddleware extends Middleware {
 			// Nothing to update
 			return;
 		}
+		$config = $this->provisioningManager->getConfig();
+		if ($config === null || !$config->isActive()) {
+			return;
+		}
 		try {
+			$this->provisioningManager->provisionSingleUser($config, $user);
 			$this->provisioningManager->updatePassword(
 				$user,
 				$this->credentialStore->getLoginCredentials()->getPassword()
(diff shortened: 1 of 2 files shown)


republicus:
Yes, please :)
I created a feature request issue at nextcloud/user_external that partially makes this request.
https://github.com/nextcloud/user_external/issues/120#issue-551873154 

Additionally, it would be useful if the user_external app were to also populate the user's email.

Mail app auto provision would be useful on a mail server with multiple domains by provisioning %EMAIL% instead of %USERID%_@domain.com_

That's already possible with provisioning. But, for other backend authentications, the email is not populated. So provisioning for multiple domains, while making use of %EMAIL%, doesn't work for for all backends.

ChristophWurst:
> Additionally, it would be useful if the user_external app were to also populate the user's email.

Makes sense. Mind opening a feature request?

republicus:
@ChristophWurst should the feature request you asked for be submitted in Mail?
The link I shared above is a feature request at nextcloud/user_external

I'm happy to reproduce it here if that is the right place for it.

PS. The automatic provisioning is working beautifully now.

ChristophWurst:
> My mail server's logon username credentials is the user's full email -- so the user's email is already known and passed on the first login with user_external.

Possibly the same as https://github.com/nextcloud/mail/issues/2566?

PR #2414, opened 2019-12-29
Feature contact group send

bfritscher:
Test of implementing contact group lookup and expansion
By adding a new GroupService implementation and copying some code of the ContactsIntegration.php 
This is a first attempt to fix issue #41 . Some open questions I identified the following open issues:
- To which email of the contact do we send? first or all?
- Should we hide the expansion in the BCC field?
- GroupService only seems to use email field no label?

Fixes https://github.com/nextcloud/mail/issues/41

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -24,32 +24,7 @@ declare(strict_types=1);
 
 namespace OCA\Mail\AppInfo;
 
-use OCA\Mail\Contracts\IAttachmentService;
-use OCA\Mail\Contracts\IAvatarService;
-use OCA\Mail\Contracts\IMailManager;
-use OCA\Mail\Contracts\IMailTransmission;
-use OCA\Mail\Contracts\IUserPreferences;
-use OCA\Mail\Events\BeforeMessageDeletedEvent;
-use OCA\Mail\Events\DraftSavedEvent;
-use OCA\Mail\Events\MessageSentEvent;
-use OCA\Mail\Events\SaveDraftEvent;
-use OCA\Mail\Http\Middleware\ErrorMiddleware;
-use OCA\Mail\Listener\AddressCollectionListener;
-use OCA\Mail\Listener\DeleteDraftListener;
-use OCA\Mail\Listener\DraftMailboxCreatorListener;
-use OCA\Mail\Listener\FlagRepliedMessageListener;
-use OCA\Mail\Listener\TrashMailboxCreatorListener;
-use OCA\Mail\Listener\SaveSentMessageListener;
-use OCA\Mail\Service\Attachment\AttachmentService;
-use OCA\Mail\Service\AvatarService;
-use OCA\Mail\Service\Group\IGroupService;
-use OCA\Mail\Service\Group\NextcloudGroupService;
-use OCA\Mail\Service\MailManager;
-use OCA\Mail\Service\MailTransmission;
-use OCA\Mail\Service\UserPreferenceSevice;
 use OCP\AppFramework\App;
-use OCP\EventDispatcher\IEventDispatcher;
-use OCP\Util;
 
 class Application extends App {
 
--- a/lib/Service/Group/ContactsGroupService.php
+++ b/lib/Service/Group/ContactsGroupService.php
@@ -1,0 +1,120 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @author Matthias Rella <mrella@pisys.eu>
+ * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @author Boris Fritscher <boris.fritscher@gmail.com>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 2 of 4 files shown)


jancborchardt:
@bfritscher hello and welcome to Nextcloud Mail! :) Could you add some screenshots so people see how to test? And it also helps a lot with design review – thanks! :)

bfritscher:
@jancborchardt thanks! There is not a lot to see I used the existing group autocomplete. And used it to add contact group autocomplete. So now in the mail's To field the autocomplete also provides group names from contacts. The entry is marked as (ContactGroup) instead of (nextcloud) which is for internal groups. On send the entry is expanded to all emails by the GroupsIntegration provided by @myrho in #810.

contactgroup_autocomplete

jancborchardt:
Ah right, and ContactGroup means a group as per the Contacts app, right?

Then my only feedback is to simply append "(group)" instead of the quite technical "(ContactGroup)". :)

ChristophWurst:
> Then my only feedback is to simply append "(group)" instead of the quite technical "(ContactGroup)". :)

That makes it ambiguous. It should be "Contacts" IMO.

lib/Service/Group/ContactsGroupService.php
```diff
@@ -0,0 +1,120 @@
...
+		$result = $this->contactsManager->search($groupId, ['CATEGORIES']);
+		$receivers = [];
+		foreach ($result as $r) {
+			if (!$allowSystemUsers && isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
+				continue;
+			}
+			if (!isset($r['EMAIL'])) {
+				continue;
+			}
+			$groups = preg_split('/,/', $r['CATEGORIES']);
```
ChristophWurst: this might return `false`

ChristophWurst: why not just use `explode`?

bfritscher: Good point missed explode when looking through the api. Long time since I coded in php.

jancborchardt:
> > Then my only feedback is to simply append "(group)" instead of the quite technical "(ContactGroup)". :)
> 
> That makes it ambiguous. It should be "Contacts" IMO.

Alright, that seems good. :)

lib/Service/Group/ContactsGroupService.php
ChristophWurst: No worries :)

lib/AppInfo/Application.php
```diff
@@ -44,6 +44,7 @@
 use OCA\Mail\Service\AvatarService;
 use OCA\Mail\Service\Group\IGroupService;
 use OCA\Mail\Service\Group\NextcloudGroupService;
+use OCA\Mail\Service\Group\ContactsGroupService;
```
ChristophWurst: actually not used here. you may also remove the two other group service imports

bfritscher: I do not yet get everything about use, but if its like imports it seems to me that since https://github.com/nextcloud/mail/commit/d18ef16578dc34ffd67e48bc5c61654e55461ce9 when BootstrapSingleton was created only App is really used? And everything else was moved to Bootstrap?

ChristophWurst: Yes, that is absolutely correct :)

Bildschirmfoto von 2020-01-08 10-54-28

bfritscher: Ok I squashed my edits into one commit and made a second one for the cleanup of the imports in applications. Do you need it on a different pull request?

ChristophWurst:
>  To which email of the contact do we send? first or all?

Good question. Let's send to all of them.

>  Should we hide the expansion in the BCC field?

No, I think there can be cases where you want to use a group for BCC.

> GroupService only seems to use email field no label?

IIRC we just fall back to email as label if no label is set.

lib/Service/Group/ContactsGroupService.php
```diff
@@ -0,0 +1,120 @@
...
+			}
+			$groups = explode(',', $r['CATEGORIES']);
+
+			// search matches substring but we only want full match
+			if (!in_array($groupId, $groups)) {
+				continue;
+			}
+
+			$email = $r['EMAIL'];
+			// only take first email ?
```
ChristophWurst: as commented, I think we should just use all.

bfritscher: ok changed it to iterate through all emails

ChristophWurst:
Will integrate once CI passes. Restarted a few jobs. Looks like it was a hiccup with the Docker service.

dieter-wilhelm:
> > To which email of the contact do we send? first or all?
> 
> Good question. Let's send to all of them.

I would suggest the first one (supposedly the "favourite" one)
 
> > Should we hide the expansion in the BCC field?
> 
> No, I think there can be cases where you want to use a group for BCC.

yes, I think for this "group" functionality BCC is the most import field with regards to privacy

> > GroupService only seems to use email field no label?
> 
> IIRC we just fall back to email as label if no label is set.

Pardon, what does "label" mean in this context?

thank you for your work and please tell me when this feature will be available.

lib/Service/Group/ContactsGroupService.php
dieter-wilhelm: I wouldn't iterate over all email addresses for a specific contact, often contacts have employer mail addresses or mail addresses for different purposes. I would just use the first mail address, which is supposed the favourite or public mail address of the respective contact.

ChristophWurst:
> I would suggest the first one (supposedly the "favourite" one)

Open a ticket to discuss.

> Pardon, what does "label" mean in this context?

It's a detail of the UI code. Like what we use for the user-visible text on each entry.

PR #2429, opened 2020-01-03
Removed max width for messages.  Fixes issue with nextcloud 17

JohnOberhauser:
Messages in nextcloud 17 are not taking up the entire space.

Before this PR
before

After
after

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -271,6 +271,7 @@ export default {
 <style lang="scss">
 #mail-message {
 	flex-grow: 1;
+	max-width: none !important;
 }
 
 .mail-message-body {


violoncelloCH:
Hi @JohnOberhauser 
Thank you very much for your pull request!
Hmm I'm actually not sure if this is a "bug"... Afaik this is the intended behaviour. At least Talk is implementing the same in order to not have too long lines on wide screens. If I'm right wat @jancborchardt said was that this limited with is here so you don't lose track while reading (it's easy to jump over / back lines while reading if they are too long).
Therefore I'm not sure if we really want this design-wise....

jancborchardt:
Hi @JohnOberhauser, the _width_ restriction is indeed intentional, as per https://github.com/nextcloud/mail/pull/2180
Further reading on the cognitive science basis on that at https://en.wikipedia.org/wiki/Line_length

What would be nice though is if the scrollbar would move to the very right side, so it doesn’t float in mid-air.

And still to be fixed is the _height_ restriction for HTML mails. There is a work-in-progress pull request for that at https://github.com/nextcloud/mail/pull/2222 – help is much appreciated there. :) Or do you have a different approach which works for HTML mails as well @JohnOberhauser?

jayhybrid:
If you must limit the width, at least make the heading full width please.

email

feutl:
> 
> 
> If you must limit the width, at least make the heading full width please.
> 
> email
> 
> I also mentioned this here #2449.

I also do not understand why the height is limited, this looks broken on every device I am using. also it is not consistent. the scrolling for mails and accounts looks and works different than the one for mails with their attachments.
Personally it feels quite strange,

jancborchardt:
> If you must limit the width, at least make the heading full width please.

@jayhybrid Then for really big resolutions, the title, actions and content are all over the place. Even on the screenshot you show it looks really off and mis-aligned. The width should be uniform.


----------------

> I also do not understand why the height is limited

@feutl As I mentioned above this is a separate issue that should indeed be fixed:
> And still to be fixed is the height restriction for HTML mails. There is a work-in-progress pull request for that at #2222 – help is much appreciated there. :)

jancborchardt:
Closing this in favor of https://github.com/nextcloud/mail/pull/2222 as per the comments above. :)

steeve971:
Hi, I am newby.
I have to modify wich file ?
Add wich code on wich file ?
It’s not clear for me.

jancborchardt:
@steeve971 hi, the vertical sizing issue is fixed with the latest release, nothing to do except updating the Mail app. :)

rumbis:
.envelope{height:900px;}

Issue #2428, opened 2019-12-31
Replying from the unified inbox, the wrong account is selected


jancborchardt:
### Expected behavior
If you reply from the unified inbox, the same account that you got the email for should be the one you are replying with.

### Actual behavior
It seems that instead, the first mail account in the list of accounts is selected.

### Mail app

**Mail app version:** latest master

cc @nextcloud/mail as this is quite a big papercut for 18. :)

PR #2447, opened 2020-01-07
Reply from correct account when in unified inbox

StCyr:
Fixes https://github.com/nextcloud/mail/issues/2428


jancborchardt:
Good stuff @StCyr! :)

PR #2448, opened 2020-01-07
Run lint for all PRs

ChristophWurst:
Otherwise PRs from forks don't get the checks apparently. https://github.com/nextcloud/mail/pull/2446 doesn, but @StCyr other PR with a branch in this repo worked.

PR #2454, opened 2020-01-09
Uses ReplyBuilder.js to compute the list of recipients when replying to a email.

StCyr:
this PR can also be used to test #2448

### DIFF
--- a/src/ReplyBuilder.js
+++ b/src/ReplyBuilder.js
@@ -75,22 +75,22 @@ export const buildRecipients = (envelope, ownAddress) => {
 	let to = []
 	let cc = []
 	if (recipientType === RecipientType.To) {
-		// Send to everyone except yourself plus the original sender
+		// Send to everyone except yourself, plus the original sender if not ourself
 		to = envelope.to.filter(isNotOwnAddress)
-		to = to.concat(envelope.from)
+		to = to.concat(envelope.from.filter(isNotOwnAddress))
 
 		// CC remains the same
 		cc = envelope.cc
 	} else if (recipientType === RecipientType.Cc) {
-		// Send to the same people plus the sender
-		to = envelope.to.concat(envelope.from)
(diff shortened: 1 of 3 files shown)


src/components/NewMessageDetail.vue
```diff
@@ -83,10 +89,13 @@ export default {
 					msgTo = []
 				}
 
+				var account = this.$store.getters.getAccount(message.accountId)
```
ChristophWurst: use `const`. Otherwise looks fine :)

ChristophWurst:
yay, the actions run :)

StCyr:
Here you are :-)

src/components/NewMessageDetail.vue
```diff
@@ -83,10 +89,13 @@ export default {
 					msgTo = []
 				}
 
+				const account = this.$store.getters.getAccount(message.accountId)
```
ChristophWurst: account is not used? :)

StCyr:
looks good now :-)

src/ReplyBuilder.js
```diff
@@ -75,22 +75,22 @@ export const buildRecipients = (envelope, ownAddress) => {
 	let to = []
 	let cc = []
 	if (recipientType === RecipientType.To) {
-		// Send to everyone except yourself plus the original sender
+		// Send to everyone except yourself, plus the original sender if not ourself
 		to = envelope.to.filter(isNotOwnAddress)
-		to = to.concat(envelope.from)
+		to = to.concat(envelope.from.filter(isNotOwnAddress))
```
ChristophWurst: oh that makes sense! Could we add a test case for this? :v:

StCyr: done :)

ChristophWurst:
Tip for a clean commit history and better developer experience: https://dev.to/tmr232/dont-amend-fix :)

PR #2479, opened 2020-01-14
Focus editor in replies and forwdared messagess

### LINKED ISSUE #2373: Reply issues
When you reply to an email, there are some small issues which are quite annoying when summed up:

- [x] The "To" field is focused, even though the recipient is already filled
- [x] You are prefilled in the "cc" field and need to manually remove yourself
- [ ] The text body has no space above and you need to insert newlines yourself. Expected would be focus on the first line, and 2 empty lines below until the original message.
- [ ] ~~If you are replying from the unified inbox, the wrong account is selected. This might be related to yourself being prefilled in cc~~ → separated out into https://github.com/nextcloud/mail/issues/2428

cc @nextcloud/mail I’d say this is some crucial polishing for 18.

GretaD:
Ref #2373

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -62,6 +62,10 @@ export default {
 			type: String,
 			default: '',
 		},
+		focus: {
+			type: Boolean,
+			default: false,
+		},
 	},
 	data() {
 		return {
@@ -127,6 +131,9 @@ export default {
 					priority: 'highest',
 				}
 			)
+			if (this.focus) {
+				logger.debug('focusing TextEditor')
+			}
 
 			// Set value as late as possible, so the custom schema listener is used
 			// for the initial editor model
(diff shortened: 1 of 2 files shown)


src/components/TextEditor.vue
```diff
@@ -127,6 +131,9 @@ export default {
 					priority: 'highest',
 				}
 			)
+			if (this.focus) {
+				logger.debug('focusing TextEditor')
```
jancborchardt: This should probably be removed? :)

jancborchardt:
Hmm, I can’t really confirm that it fixes anything? Also I still have myself in cc when replying to mails.

ChristophWurst:
> Also I still have myself in cc when replying to mails.

https://github.com/nextcloud/mail/pull/2454

ChristophWurst:
> Hmm, I can’t really confirm that it fixes anything?

Have you built the app? What happens when you try to reply to a message? What is focused? Do you see the debug log about the focus?

src/components/TextEditor.vue
ChristophWurst: I'm fine with leaving those as they aid debugging of bug reports.

GretaD:
@jancborchardt its working now, can you please recheck?

PR #2480, opened 2020-01-14
Allow navigation from more paths to new message composer

### LINKED ISSUE #110: Navigation from account settings to new message composer is impossible [$10 awarded]

kingjan1999:
This PR solves issue #110, making navigating from account settings / setup to the new message composer possible again.

To make this possible, the first mail account and first folder available is used when navigating back. This requires at least one account / folder present and will result in the folder contents loaded and rendered again.

Closes #110

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -316,20 +316,42 @@ export default {
 				this.editorPlainText = val.editorMode === 'plaintext'
 			}
 		},
+		'$route.params.messageUid'(newID) {
+			this.reset()
+		},
 	},
 	beforeMount() {
-		if (this.fromAccount) {
-			this.selectedAlias = this.aliases.find((alias) => alias.id === this.fromAccount)
-		} else {
-			this.selectedAlias = this.aliases[0]
-		}
-
-		this.bodyVal = this.bodyWithSignature(this.selectedAlias, this.bodyVal)
+		this.setAlias()
+		this.initBody()
 	},
 	mounted() {
 		this.$refs.toLabel.$el.focus()
+		// event is triggered when user clicks 'new message' in navigation
+		this.$root.$on('newMessage', () => {
+			this.draftsPromise
+				.then(() => {
+					return this.saveDraft(this.getMessageData)
+				})
+				.then(() => {
+					// wait for the draft to be saved before resetting the message content
+					this.reset()
+				})
+		})
+	},
+	beforeDestroy() {
+		this.$root.$off('newMessage')
 	},
 	methods: {
+		setAlias() {
+			if (this.fromAccount) {
+				this.selectedAlias = this.aliases.find((alias) => alias.id === this.fromAccount)
+			} else {
+				this.selectedAlias = this.aliases[0]
+			}
+		},
+		initBody() {
+			this.bodyVal = this.bodyWithSignature(this.selectedAlias, this.bodyVal)
+		},
 		recipientToRfc822(recipient) {
 			if (recipient.email === recipient.label) {
 				// From mailto or sender without proper label
@@ -363,12 +385,31 @@ export default {
 		saveDraft(data) {
 			this.savingDraft = true
 			this.draftsPromise = this.draftsPromise
-				.then((uid) => this.draft(data(uid)))
+				.then((uid) => {
+					const draftData = data(uid)
+					if (
+						!uid &&
+						!draftData.subject &&
+						!draftData.body &&
+						!draftData.cc &&
+						!draftData.bcc &&
+						!draftData.to
+					) {
+						// this might happen after a call to reset()
+						// where the text input gets reset as well
+						// and fires an input event
+						logger.debug('Nothing substantial to save, ignoring draft save')
+						this.savingDraft = false
+						return uid
+					}
+					return this.draft(draftData)
+				})
 				.catch(logger.error.bind(logger))
 				.then((uid) => {
 					this.savingDraft = false
 					return uid
 				})
+			return this.draftsPromise
 		},
 		onInputChanged() {
 			this.saveDraftDebounced(this.getMessageData)
@@ -438,6 +479,7 @@ export default {
 				})
 		},
 		reset() {
+			this.draftsPromise = Promise.resolve() // "resets" draft uid as well
 			this.selectTo = []
 			this.selectCc = []
 			this.selectBcc = []
@@ -446,6 +488,17 @@ export default {
 			this.attachments = []
 			this.errorText = undefined
 			this.state = STATES.EDITING
+			this.autocompleteRecipients = []
+			this.newRecipients = []
+
+			this.setAlias()
+			this.initBody()
+			Vue.nextTick(() => {
+				// toLabel may not be on the DOM yet
+				// (because "Message sent" is shown)
+				// so we defer the focus call
+				this.$refs.toLabel.$el.focus()
+			})
 		},
 		/**
 		 * Format aliases for the Multiselect
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -75,6 +75,7 @@ import AppNavigationNew from '@nextcloud/vue/dist/Components/AppNavigationNew'
 import AppNavigationSettings from '@nextcloud/vue/dist/Components/AppNavigationSettings'
 import AppNavigationSpacer from '@nextcloud/vue/dist/Components/AppNavigationSpacer'
 
+import logger from '../logger'
 import NavigationAccount from './NavigationAccount'
 import NavigationAccountExpandCollapse from './NavigationAccountExpandCollapse'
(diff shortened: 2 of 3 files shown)


jancborchardt:
@kingjan1999 nice! Is this actually also intended to fix https://github.com/nextcloud/mail/issues/2510 ?

(Will review next week, traveling atm.)

kingjan1999:
@jancborchardt It wasn't intended originally, but it should fix this issues as well now (implemented with e46dfa34).

src/components/Composer.vue
```diff
@@ -314,6 +314,8 @@ export default {
 	},
 	mounted() {
 		this.$refs.toLabel.$el.focus()
+		// event is triggered when user clicks 'new message' in navigation
+		this.$root.$on('newMessage', () => this.reset())
```
ChristophWurst: will it re-register this event handler every time a composer component is rendered? Should there be an `$off` or similar in `unmount`?

ChristophWurst: what happens to the current data in the composer? I guess we should try to save a last draft before clearing the user input, no?

kingjan1999: > will it re-register this event handler every time a composer component is rendered? Should there be an `$off` or similar in `unmount`?

Good catch, will add this.

>what happens to the current data in the composer? I guess we should try to save a last draft before clearing the user input, no?

I think this is a good question, which maybe should be even discussed in a separate issues (#2510 does not apply here, as there is no draft to save any more after the message have been sent). If the user clicks "New Message" while currently composing another message, does he want to:

1. Save the current message as a draft and create a new, separate message? In this case we not only need to save the last message as a draft (as you suggested) but also make sure to set a new message / draft uid to prevent overwriting the new message.

2. Completely discard the current message (incl. draft versions) and just start over? This is - more or less - what is currently done in this PR.

IMHO, 1) is the better one, so I changed the behavior in the most recent commit (3600208d). 

Note that this made some additional changes necessary to get the `reset()` right (The reset of the text input didn't work correctly as there was no real 2-way data binding between `Composer` and `TextEditor`. This wasn't a problem before, as reset was only used after the messages was sent, so the `TextEditor` would get re-mounted nevertheless)

ChristophWurst: yes 1) makes most sense

kingjan1999:
@ChristophWurst Can you just take another look at this if there is anything blocking?

src/components/Navigation.vue
```diff
@@ -106,15 +108,29 @@ export default {
...
+				return
+			}
+
+			this.$router
+				.push({
+					name: 'message',
+					params: {
+						accountId,
+						folderId,
+						messageUid: 'new',
```
ChristophWurst: Needs adjustment for/after https://github.com/nextcloud/mail/pull/2706/files#diff-03f4e6dc240c55861fa235e27445d65bR86

src/components/TextEditor.vue
```diff
@@ -92,6 +92,11 @@ export default {
 			},
 		}
 	},
+	watch: {
+		value(newVal) {
```
ChristophWurst: why do we need this?

kingjan1999: This is for us to be able to reset the text body.  Otherwise, when calling `reset()` the `bodyVal` in `Composer` is set to '' but the change won't be reflected to the actual text editor

src/components/Composer.vue
```diff
@@ -358,12 +371,31 @@ export default {
...
+						!draftData.to
+					) {
+						// this might happen after a call to reset()
+						// where the text input gets reset as well
+						// and fires an input event
+						logger.debug('Nothing substantial to save, ignoring draft save')
+						this.savingDraft = false
+						return uid
+					}
+					return this.draft(data(uid))
```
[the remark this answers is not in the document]
kingjan1999: Changed, thank you!

src/components/Navigation.vue
kingjan1999: Should be adjusted

kingjan1999:
@jancborchardt Thank you, fixed!

jancborchardt:
Nice! That works very well now, but I think 2 new small issues came:
- When you click reply on a mail, the "To" suggestion dropdown flashes briefly (cause focused?)
- When you then click the "New message" button, the composer loads and the "To" field is focused, but the suggestion dropdown of the previous mail you were replying to shows?

kingjan1999:
>     * When you click reply on a mail, the "To" suggestion dropdown flashes briefly (cause focused?)

I believe this is unrelated to this PR. I think this happens because `Composer` calls `this.$refs.toLabel.$el.focus()` which focuses the "To" input and then the TextEditor calling focus if `isReply` is true. This should be easy to fix (by wrapping the first `focus` call in `if(!isReply`), but I think this is out of scope for this PR. Apart from that, after merging the latest master branch, I am no longer able to reproduce this issue, so maybe it got fixed already.
 
>     * When you then click the "New message" button, the composer loads and the "To" field is focused, but the suggestion dropdown of the previous mail you were replying to shows?

Fixed! I refactored the way `reset()` works, so bugs like these shouldn't happen anymore

src/components/Composer.vue
```diff
@@ -430,14 +470,15 @@ export default {
...
 		reset() {
-			this.selectTo = []
-			this.selectCc = []
-			this.selectBcc = []
-			this.subjectVal = ''
-			this.bodyVal = ''
-			this.attachments = []
-			this.errorText = undefined
-			this.state = STATES.EDITING
+			Object.assign(this.$data, this.$options.data.apply(this))
```
ChristophWurst: is it safe to assign new props to [this.$data](https://vuejs.org/v2/api/#data)? I can't find any docs on `this.$options`. Is this even an official API?

src/components/Composer.vue
```diff
@@ -311,20 +311,42 @@ export default {
...
 	mounted() {
 		this.$refs.toLabel.$el.focus()
+		// event is triggered when user clicks 'new message' in navigation
+		this.$root.$on('newMessage', () => {
+			this.draftsPromise
+				.then(() => {
+					return this.saveDraft(this.getMessageData)
+				})
+				.then(() => {
+					// wait for the draft to be saved before resetting the message content
```
ChristophWurst: Actually we can do this in the background, right?

src/components/Composer.vue
kingjan1999: > I can't find any docs on this.$options. Is this even an official API?

[It is](https://vuejs.org/v2/api/#vm-options)

> is it safe to assign new props to this.$data? 

Well, I took this approach from the [Vue issue tracker](https://github.com/vuejs/vue/issues/702#issuecomment-308991548), so I wouldn't consider this neither experimental nor "offically recommended". I can revert back to manually resetting the fields if you feel this is more safe.

src/components/Composer.vue
kingjan1999: Not really, as the draft save has to be triggered before resetting the message content. To be more precise, `getMessageData` has to be called before reset sets the message data to ''. As we don't really have control over when the call to `getMessageData` happens, we have to wait for the whole promise to finish (instead of continuing right after the call).

ChristophWurst: You're right. This could be deferred. Fine then :)

src/components/Composer.vue
ChristophWurst: Oh, okay. I would prefer the manual way indeed.

kingjan1999: Sure! I reverted the change.

jancborchardt:
@kingjan1999 @ChristophWurst what about the failing Travis tests? :)

kingjan1999:
> @kingjan1999 @ChristophWurst what about the failing Travis tests? :)

The error message from Travis is:
```
Using default tag: latest

Error response from daemon: unexpected EOF

Makefile:59: recipe for target 'start-smtp-docker' failed

make: *** [start-smtp-docker] Error 1

The command "make start-smtp-docker" failed and exited with 2 during

```

This seems unrelated to this PR. Maybe we can restart the Travis job?

ChristophWurst:
Rebased & squashed. Let's merge once CI is done :)

ChristophWurst:
@kingjan1999 you may also join our developer discussion at https://cloud.nextcloud.com/call/89474m7g :)

Issue #2105, opened 2019-10-13
Mailbox delimiters are optional


arkanoid87:
### Expected behavior
Mail app gets loaded

### Actual behavior
Internal server error

### Mail app
0.17.0

### Server configuration
ubuntu server 18.04 + docker-compose with nextcloud:latest (17.0) + mariadb
System integrity: check

#### Client configuration
Firefox/69.0
Ubuntu 18.04

I was trying to add my first mail account into mail app (SSL on both smtp and imap) and I was encountering generic errors, so I've enabled debug in my config and I had this
```
{"reqId":"95qexcN6zulTTpsAhrqf","level":0,"time":"2019-10-13T03:27:30+00:00","remoteAddr":"<redacted>","user":"<redacted>","app":"PHP","method":"POST","url":"\/apps\/mail\/api\/accounts","message":"stream_socket_client(): SSL operation failed
with code 1. OpenSSL Error messages:\nerror:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure at \/var\/www\/html\/custom_apps\/mail\/vendor\/pear-pear.horde.org\/Horde_Socket_Client\/Horde\/Socket\/Client.php#293","userAgent":"Mozilla\/5.0 (X11; Ubuntu; Linux x86_64; rv:69.0) Gecko\/20100101 Firefox\/69.0","version":"17.0.0.9"}
{"reqId":"95qexcN6zulTTpsAhrqf","level":0,"time":"2019-10-13T03:27:30+00:00","remoteAddr":"<redacted>","user":"<redacted>","app":"PHP","method":"POST","url":"\/apps\/mail\/api\/accounts","message":"stream_socket_client(): Failed to enable crypto at \/var\/www\/html\/custom_apps\/mail\/vendor\/pear-pear.horde.org\/Horde_Socket_Client\/Horde\/Socket\/Client.php#293","userAgent":"Mozilla\/5.0 (X11; Ubuntu; Linux x86_64; rv:69.0) Gecko\/20100101 Firefox\/69.0","version":"17.0.0.9"}
{"reqId":"95qexcN6zulTTpsAhrqf","level":0,"time":"2019-10-13T03:27:30+00:00","remoteAddr":"<redacted>","user":"<redacted>","app":"PHP","method":"POST","url":"\/apps\/mail\/api\/accounts","message":"stream_socket_client(): unable to connect to
[...]

PR #2515, opened 2020-01-20
delimiter cannot be NULL

MikyWoW:
Bug #2105

### DIFF
--- a/lib/Folder.php
+++ b/lib/Folder.php
@@ -60,7 +60,7 @@ class Folder implements JsonSerializable {
 		$this->accountId = $accountId;
 		$this->mailbox = $mailbox;
 		$this->attributes = $attributes;
-		$this->delimiter = $delimiter;
+		$this->delimiter = $delimiter ?: '/';
 		$this->folders = [];
 		$this->status = [];
 		$this->specialUse = [];


lib/Folder.php
```diff
@@ -60,7 +60,7 @@ public function __construct(int $accountId, Horde_Imap_Client_Mailbox $mailbox,
 		$this->accountId = $accountId;
 		$this->mailbox = $mailbox;
 		$this->attributes = $attributes;
-		$this->delimiter = $delimiter;
+		$this->delimiter = $delimiter ?: '/';
```
ChristophWurst: See https://github.com/nextcloud/mail/issues/2105#issuecomment-576693800. I'm a bit worries about the default value.

From https://tools.ietf.org/html/rfc3501#section-7.2.2

> The hierarchy delimiter is a character used to delimit levels of
>       hierarchy in a mailbox name.  A client can use it to create child
>       mailboxes, and to search higher or lower levels of naming
>       hierarchy.  All children of a top-level hierarchy node MUST use
>       the same separator character.  A NIL hierarchy delimiter means
>       that no hierarchy exists; the name is a "flat" name.

Aha! So `null` (NIL) is actually okay.

MikyWoW: If null is valid, then this fix is probably not best solution. Maybe fix it directly on FolderMapper.php - line 219

MikyWoW: If i fix FolderMapper.php i getting SQL error:
`An exception occurred while executing 'INSERT INTO oc_mail_mailboxes .....  SQLSTATE[HY000]: General error: 1364 Field 'delimiter' doesn't have a default value.`
DB structure need change too:
`ALTER TABLE oc_mail_mailboxes MODIFY delimiter varchar(1) DEFAULT NULL;`

ChristophWurst: yes, we need to fix this at a few places. let's discuss in your ticket :)

PR #2520, opened 2020-01-21
hide subfolder for all inboxes

GretaD:
hide subfolder for All inboxes


ChristophWurst:
- [ ] Rebase once #2523 19 has ben  merged
- [ ] Fix eslint

PR #2521, opened 2020-01-21
remove mark all as read from All Inboxes

### LINKED ISSUE #2370: Hide 'mark all as read' action from unified inbox
The server doesn't know about the unified account/inbox. We should remove the action and only offer it for individual mailboxes.

cc @mario 

@GretaD maybe something for you? :)

GretaD:
fixes #2370


ChristophWurst:
- [ ] Rebase once #2523 19 has ben  merged

Issue #505, opened 2017-09-06
Remember »Show images from this sender«


jancborchardt:
Clicking the button »Show images from this sender« should only be necessary once per sender. We should remember that you want to see images from this sender, which is the point of this button. ;)

cc @nextcloud/mail

NESC1US:
it would be also nice to show the images after the button disappears, I see some octet streams, but those are never actually shown on the page.

ChristophWurst:
> its been 3 years for this bug, maybe the feature is too big to be implemented, how about removing it completely?

Just needs someone to implement it :) Are you a developer or know someone who is?

PR #2532, opened 2020-01-23
removing a button which does nothing #505

NESC1US:
ok, bug #505 fix pull request :)


ChristophWurst:
This feature work. And if it doesn't we have to fix it. No reason to remove the button :)

NESC1US:
with all due respect, it does not work, you commented on it couple days ago with this:
> Just needs someone to implement it :) Are you a developer or know someone who is?

ChristophWurst:
Yeah, the part about remembering. But the showing of images is there.

PR #2533, opened 2020-01-23
Change Copy/Blind Copy into CC/BCC

### LINKED ISSUE #2531: "+ Copy" isn't working well with translations
https://github.com/nextcloud/mail/commit/e22c7005df8fceb1b535346609c48733682ee3bd changed the composer from "+ CC/BCC" to " to "+ Copy". This isn't working well with translations and thus confuses end users.

@GretaD could you revert this specific change so we have the old label again?

GretaD:
fixes #2531


jancborchardt:
If we have to change it back, could we at least use the form "Cc/Bcc" so it’s in line with e.g. how Gmail does it and not all uppercase? :)

Issue #2489, opened 2020-01-17
Warning "KItinerary binary adapter is not available, can't extract information" in nextcloud.log after update to version 1.0.0


Nils160988:
### Expected behavior
Everything runs smooth after update of mail-app.

### Actual behavior
After app-update to version 1.0.0 the warning "KItinerary binary adapter is not available, can't extract information" shows up in the server-log, twice for each mail I open.
Also the white space in the app looks a bit odd (don't know if related).

### Mail app

**Mail app version:** 1.0.0

**Mailserver or service:** self hosted

### Server configuration

**Operating system**: Ubuntu 18.04

**Web server:** Apache 2.4.29

**Database:** mariadb 10.1.43

**PHP version:** 7.2.24

**Nextcloud Version:** 17.0.2

#### Client configuration
**Browser:** Firefox 72.0.1

**Operating system:** Ubuntu 18.04

ChristophWurst:
It's a warning. Do you have your log level set too low maybe?

Nils160988:
I have set loglevel 2.
Just thought about reporting, because I don't expect a warning for a pretty much standard installation.
But of course, I can ignore it, if not important.

ChristophWurst:
We can make it a debug mode, but yeah, seeing more entries if the log level is low is kinda expected.

Nils160988:
Found some additional error message in the apache-log:
```
sh: 1: /var/www/nextcloud/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: Permission denied
sh: 1: flatpak: not found
```
Did you introduce some new dependency?

ChristophWurst:
Yes, https://packagist.org/packages/christophwurst/kitinerary-bin and https://packagist.org/packages/christophwurst/kitinerary-flatpak

ChristophWurst:
Is there any more details on what line this fails?

Nils160988:
No, sorry. I changed loglevel to debug and also no more information.
The kitinerary-extractor binary has 644 permission, not sure if this is intended.

mzeiher:
I had the same Issue, I had to manually `chmod +x` the file `<...>vendor/christophwurst/kitinerary-bin/bin/kitinerary-extractor`, not I don't get the warning anymore, maybe the extraction does not preserve the execution permission

ChristophWurst:
Interesting! I have 744 here :thinking:

Nils160988:
[...]

PR #2534, opened 2020-01-23
Make the itinerary extractor executable in a repair step

ChristophWurst:
Workaround for https://github.com/nextcloud/server/issues/19093

Fixes https://github.com/nextcloud/mail/issues/2489

### DIFF
--- a/lib/Migration/MakeItineraryExtractorExecutable.php
+++ b/lib/Migration/MakeItineraryExtractorExecutable.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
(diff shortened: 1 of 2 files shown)


lib/Migration/MakeItineraryExtractorExecutable.php
```diff
@@ -0,0 +1,75 @@
...
+			$this->logger->warning('itinerary file doesn\'t exist');
+			$output->info('itinerary file doesn\'t exist');
+			return;
+		}
+		if (is_executable($this->file)) {
+			$this->logger->debug('itinerary is already executable');
+			return;
+		}
+		try {
+			@chmod($this->file, 744);
```
juliushaertl: chmod returns TRUE on success or FALSE on failure, so catching Throwable doesn't work here.

ChristophWurst: thanks! will check the return as well.

the try-catch was just added so this always runs through. This repair step isn't critical, thus it shouldn't block the instance repair if anything blows up.

ChristophWurst: Neat, scrutinizer detected the same

> It seems like you do not handle an error condition for chmod(). This can introduce security issues, and is generally not recommended.

Issue #2513, opened 2020-01-20
Pasting plain text inserts newlines


jancborchardt:
When you paste in plain text from a text editor amd your Mail app is also set to plain text, there's still 2 line breaks inserted for every single break.

Meaning that this message:
```
Tasks
- Bla
- Test
```

Pasted into Mail and sent will turn to
```
Tasks

- Bla

- Test
```

Related to https://github.com/nextcloud/mail/issues/2481

ChristophWurst:
:+1: IIRC there were a few issues in the CKEditor5 repo. So chances are good this has to be fixed upstream.

jancborchardt:
This is fixed in master! :) Possibly through #2538

PR #2538, opened 2020-01-24
Sets <p> margin to 0 for sent emails

### LINKED ISSUE #2481: Sending mail inserts newlines
Write a mail:
```
⏮ Last week:
- More 18 reviews and polishing
- Went through notifications
```

Send it, and see it formatted like this:
```
⏮ Last week:

- More 18 reviews and polishing

- Went through notifications
```

Priority cause it changes mail text unexpectedly. :)

StCyr:
fixes #2481 and #2513 I guess )

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -126,6 +126,20 @@ export default {
 
 			logger.debug('CKEditor editor is ready', {editor, schema})
 
+			// Set 0 pixel margin to all <p> elements
+			editor.conversion.for('downcast').add(dispatcher => {
+				dispatcher.on(
+					'insert:paragraph',
+					(evt, data, conversionApi) => {
+						const viewWriter = conversionApi.writer
+						viewWriter.setStyle('margin', '0', conversionApi.mapper.toViewElement(data.item))
+					},
+					{
+						priority: 'low',
+					}
+				)
+			})
+
 			schema.on(
 				'checkChild',
 				(evt, args) => {
--- a/src/util/HtmlHelper.js
+++ b/src/util/HtmlHelper.js
@@ -24,6 +24,9 @@ export const htmlToText = html => {
 					.replace(/\n\n\n/g, '\n\n') // remove triple line breaks
 					.replace(/^/gm, '> ') // add > quotation to each line
 			},
+			paragraph: function(element, fn, options) {
+				return fn(element.children, options) + '\n'
+			},
 		},
 	})
 


src/components/NewMessageDetail.vue
```diff
@@ -261,6 +261,7 @@ export default {
 			})
 		},
 		sendMessage(data) {
+			data.body = '<style>p{margin:0;}</style>' + data.body
```
ChristophWurst: Two remarks

1) This is problematic for plaintext messages as at this point we already have a non-hmtl version of the body AFAIK
2) The CKEditor would allow us to change the way certain elements are rendered into HTML. We could change the generated HTML there as well. Might be the cleaner approach :thinking:

StCyr: >     1. This is problematic for plaintext messages as at this point we already have a non-hmtl version of the body AFAIK

You're right. 

>     2. The CKEditor would allow us to change the way certain elements are rendered into HTML. We could change the generated HTML there as well. Might be the cleaner approach 🤔

You mean using this? https://ckeditor.com/docs/ckeditor5/latest/features/headings.html

ChristophWurst: More like this https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/conversion/conversion-extending-output.html#adding-a-css-class-to-block-elements. But the question is how to add the styles. We could either have them inline for every `<p>` or include a style block somewhere.

StCyr: I would say adding a style bock is cleaner. Isn't?

StCyr: wouldn't this be more appropriate? 

https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/conversion/conversion-preserving-custom-content.html

StCyr:
I've tried to create a ckeditor plugin to implement this functionality.

Without success till now.

If someone @nextcloud/mail with knowledge about ckeditor's internals could review my PR to try and understand what should be done ,that would be great.

ping @ChristophWurst

StCyr:
Hi @ChristophWurst 

This should be working properly now

Cyrille

src/components/NewMessageDetail.vue
ChristophWurst: Makes sense!

src/components/Composer.vue
```diff
@@ -342,7 +342,9 @@ export default {
 					bcc: this.selectBcc.map(this.recipientToRfc822).join(', '),
 					draftUID: uid,
 					subject: this.subjectVal,
-					body: this.editorPlainText ? htmlToText(this.bodyVal) : this.bodyVal,
+					body: this.editorPlainText
```
ChristophWurst: Works, but from a logical PoV the composer shouldn't have to care about this. This should be moved to the `TextEditor.vue`

ChristophWurst:
Can't test right now as I have my local setup migrated to #2064 but will give it a test once that PR is merged :)

src/components/Composer.vue
StCyr: Makes sense but I can't figure out how to implement this there.

What could probably easily be done would be to add a "style" attribute to all `<p>` elements rather than assign them a class:

`viewWriter.addClass('null-margin', conversionApi.mapper.toViewElement(data.item))`

might be changed to add an attribute rather than adding a class.

ChristophWurst: > Makes sense but I can't figure out how to implement this there.

Possibly at https://github.com/nextcloud/mail/blob/1691463ecdfa207f5e470c7592196812b76ff563/src/components/TextEditor.vue#L141 but CKEditor might remove the `style` tag. I also just realized that when we only add the style element when the message is sent, the previewed message looks different than what is actually sent.



> What could probably easily be done would be to add a "style" attribute to all `<p>` elements rather than assign them a class:

If that is easy to do I'm fine with that as well :) This will also show a correct preview of what will be sent (independent of Nextcloud's paragraph styling)

ChristophWurst:
This works fine with the rich text editing mode. But the plain text one still has line breaks for paragraphs, no?

StCyr:
> 
> 
> This works fine with the rich text editing mode. But the plain text one still has line breaks for paragraphs, no?

I think it's fixed now

PR #2539, opened 2020-01-24
Add mailto: protocol handler

### LINKED ISSUE #2535: Mailto protocol handler vanished
### Expected behavior
It should be possible for the app to register as `mailto:` protocol handler. This worked up until we migrated to Vue.

### Actual behavior
There is no handler.

### Solution
The app should [register a protocol handler]https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler). We used to do this as soon as the app starts https://github.com/nextcloud/mail/blob/v0.11.0/js/app.js#L64-L73. But I was never a fan of this as it's making the first run experience worse with an additional dialogue.

Thus we should just add a setting to the bottom left. Like a button "Register as application for mail links". When the user clicks this, we run https://github.com/nextcloud/mail/blob/v0.11.0/js/app.js#L64-L73 (maybe without the theme thing because that API is ugly :see_no_evil:).

### Mail app

**Mail app version:**

Anything after 0.11

GretaD:
fixes #2535

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -18,7 +18,11 @@
 			/>
 			<label for="gravatar-enabled">{{ t('mail', 'Use Gravatar and favicon avatars') }}</label>
 		</p>
-
+		<p>
+			<button class="icon-mail app-settings-button" @click="registerProtocolHandler">
+				{{ t('mail', 'Register as application for mail links') }}
+			</button>
+		</p>
 		<p class="icon-details app-settings-button button">
 			<router-link :to="{name: 'keyboardShortcuts'}">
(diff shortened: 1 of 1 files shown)


src/components/AppSettingsMenu.vue
```diff
@@ -18,7 +18,11 @@
 			/>
 			<label for="gravatar-enabled">{{ t('mail', 'Use Gravatar and favicon avatars') }}</label>
 		</p>
-
+		<p>
+			<button class="icon-mail-white app-settings-button primary" @click="registerProtocolHandler">
```
ChristophWurst: let's use the non-white icon and remove `primary`. This is a secondary action :)

PR #2543, opened 2020-01-24
Preserve recipient input when focus is lost

ChristophWurst:
Steps to reproduce

* Start new message
* Type something into TO, CC or BCC but don't press Enter or click the entry in the dropdown
* Focus another element (with <kbd>Tab</kbd>)
* Set focus back to TO/CC/BCC

## Before

Previous input lost

## Now

Input preserved


skjnldsv:
SHouldn't we enforce this on the components?

ChristophWurst:
Not sure. Doesn't it depend on the use case? We already have this problem where taggable doesn't work as expected here because of wrong assumptions in our wrapper.

PR #2556, opened 2020-01-26
Linting is already done by github actions

rullzer:
So no need to lint twice


rullzer:
that is what I get from looking to late at stuff :P

Issue #2559, opened 2020-01-27
Add user setting to opt-out from data collection


ChristophWurst:
For https://github.com/nextcloud/mail/projects/7.

User should have control over whether their interactions are recorded by the Nextcloud instance or not. We will never send it anywhere else, but still the user has to have the choice.

Thus we need a new checkbox in the app's settings menu, just like we have for the gravatars, and persist the opt-out on the server.

@GretaD can you add the new checkbox and wire it up with the back-end just like we do with the gravatar setting? I can take care of the backe-end changes, if any necessary. :)

ChristophWurst:
And we should then already use this setting for the existing recipient collection code.

PR #2562, opened 2020-01-27
Add user setting to opt-out from data collection

GretaD:
Add user setting to opt-out from data collection

Fixes https://github.com/nextcloud/mail/issues/2559

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -126,6 +126,7 @@ class PageController extends Controller {
 				'app-version' => $this->config->getAppValue('mail', 'installed_version'),
 				'accounts' => base64_encode(json_encode($accountsJson)),
 				'external-avatars' => $this->preferences->getPreference('external-avatars', 'true'),
+				'collect-data' => $this->preferences->getPreference('collect-data', 'true'),
 			]);
 		$this->initialStateService->provideInitialState(
 			Application::APP_ID,
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -4,6 +4,31 @@
 			{{ t('mail', 'Add mail account') }}
 		</router-link>
 
+		<p v-if="loadingOptOutSettings" class="app-settings">
+			<span class="icon-loading-small"></span>
+			{{
+				t(
+					'mail',
+					'Allow the app to collect data about your interactions. Based on this data, the app will adapt to your preferences. The data will only be stored locally.'
+				)
+			}}
+		</p>
+		<p v-else class="app-settings">
+			<input
+				id="data-collection-toggle"
+				class="checkbox"
+				type="checkbox"
+				:checked="useDataCollection"
+				@change="onToggleCollectData"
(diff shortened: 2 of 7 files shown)


src/components/AppSettingsMenu.vue
```diff
@@ -4,6 +4,22 @@
 			{{ t('mail', 'Add mail account') }}
 		</router-link>
 
+		<p v-if="loadingOptOutSettings" class="app-settings">
+			<span class="icon-loading-small"></span>
+			<!-- better label? -->
+			{{ t('mail', 'Collect data..') }}
```
ChristophWurst: My take: "Allow the app to collection data about your interactions. Based on this data, the app will adapt to your preferences. The data will only be stored locally."

GretaD:
> Lint isn't happy :)

it never is #spoiledLittleSomething

src/components/AppSettingsMenu.vue
ChristophWurst: you omitted the last two sentences on purpose? ;)

ChristophWurst:
For the initial value on page load, see

* https://github.com/nextcloud/mail/blob/8b41503c90b9ab67626208dc25b7b786906173cd/lib/Controller/PageController.php#L128
* https://github.com/nextcloud/mail/blob/dfe651eee4d712cc3f252f9a2629143d0cff9fe8/templates/index.php#L34
* https://github.com/nextcloud/mail/blob/ad29c8a46cc0bc6e783e82cc81ea5c701c66fb98/src/main.js#L62-L65

src/components/AppSettingsMenu.vue
GretaD: haha nope, i thought you were explaining to me the first line.

ChristophWurst: nah it's just a bit of text to clarify what "we" do with the data. so users know it's not sent somewhere suspicious but just the Nextcloud instance they're using

GretaD: absolutely

lib/Controller/PageController.php
```diff
@@ -126,6 +126,7 @@ public function index(): TemplateResponse {
 				'app-version' => $this->config->getAppValue('mail', 'installed_version'),
 				'accounts' => base64_encode(json_encode($accountsJson)),
 				'external-avatars' => $this->preferences->getPreference('external-avatars', 'true'),
+				'external-optOut' => $this->preferences->getPreference('external-optOut', 'true'),
```
ChristophWurst: how about just `collect-data` as key?

GretaD: sure, do you want to check it all and let me know if theres anything else i need to change?

ChristophWurst:
I have to fix the tests

> 1) OCA\Mail\Tests\Unit\Controller\PageControllerTest::testIndex
>
>OCA\Mail\Contracts\IUserPreferences::getPreference('collect-data', 'true') was not expected to be called more than once.

PR #2567, opened 2020-01-28
Add %DISPLAYNAME% as parameter and start documenting provisioning

ksteinb:
This pull request adds a %DISPLAYNAME% parameter to the provisioning and tries to document the provisioning a little bit in admin.md

### DIFF
--- a/doc/admin.md
+++ b/doc/admin.md
@@ -25,6 +25,24 @@ You can use the php-mail function to send mails. This is needed for some webhost
 ```php
 'app.mail.transport' => 'php-mail'
 ```
+### Provisioning
+to Provisioning of Mail Accounts for all user navigate to Settings -> Groupware
+there you can set the parameters the mail account
+
+There are the following parameters available:
+	%EMAIL%  will be replaced by the Email Adress of the User
+	%USERID% will be replaced by the User ID (note: on LDAP Account this is completly useless as it is an UUID Value)
+	%DISPLAYNAME% will be replace by the Displayname
+
+For LDAP Setups:  your best bet should be to use %EMAIL% both for Email and the User Names for imap and smtp
+                  But! If your IMAP and SMTP Server does not accept Email Adress for Login and requires a different value
+                  you will be in trouble, as the original Username is not easily available
+
+                  Try to convince your Mail Admins to allow also Email Address for login, otherwise:
+
+		  Workaround (but dirty): Use advanced LDAP settings an return the login name (uid or cn) to the Displayname
+                                          (no it is really not nice) and use %DISPLAYNAME%
+
 
 ## Troubleshooting
 
--- a/doc/admin.md
+++ b/doc/admin.md
@@ -27,22 +27,6 @@ You can use the php-mail function to send mails. This is needed for some webhost
 ```
 ### Provisioning
 to Provisioning of Mail Accounts for all user navigate to Settings -> Groupware
-there you can set the parameters the mail account
-
-There are the following parameters available:
-	%EMAIL%  will be replaced by the Email Adress of the User
-	%USERID% will be replaced by the User ID (note: on LDAP Account this is completly useless as it is an UUID Value)
-	%DISPLAYNAME% will be replace by the Displayname
-
-For LDAP Setups:  your best bet should be to use %EMAIL% both for Email and the User Names for imap and smtp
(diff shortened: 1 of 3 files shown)


doc/admin.md
```diff
@@ -25,6 +25,24 @@ You can use the php-mail function to send mails. This is needed for some webhost
 ```php
 'app.mail.transport' => 'php-mail'
 ```
+### Provisioning
+to Provisioning of Mail Accounts for all user navigate to Settings -> Groupware
```
ChristophWurst: I'm fine with this line, but the rest should be on the UI. Docs here are prone to get out of sync with the UI changes.

lib/Service/Provisioning/Config.php
```diff
@@ -122,6 +122,9 @@ private function buildUserEmail(string $original, IUser $user) {
 		if ($user->getUID() !== null) {
 			$original = str_replace('%USERID%', $user->getUID(), $original);
 		}
+		if ($user->getDisplayName() !== null) {
+			$original = str_replace('%DISPLAYNAME%', $user->getDisplayName(), $original);
```
ChristophWurst: Please add to https://github.com/nextcloud/mail/blob/df939b93e2993749fbdb43401acedb7cde1adf00/src/components/settings/ProvisioningSettings.vue#L246 as well

ksteinb:
ok, i will change it appropriately, thanks for the hints

src/components/settings/ProvisioningSettings.vue
```diff
@@ -243,7 +243,9 @@
 						/>
 						<br />
 						<small>{{
-							t('mail', "* %USERID% and %EMAIL% will be replaced with the user's UID and email")
+							t('mail', "* %USERID% %DISPLAYNAME and %EMAIL% will be replaced with the user's UID, DISPLAYNAME and email")
```
ChristophWurst: please remove the two lines above. admins should know the values of the three placeholders and thus be able to figure out what will work for them :)

ksteinb: is it better now? But please think of not so experienced admins, the documentation of the mail app is really sparse

ksteinb:
is the line too long now ?

src/components/settings/ProvisioningSettings.vue
ChristophWurst: It should be "* %USERID%, %DISPLAYNAME% and %EMAIL% will be replaced with the user's UID, email and display name."

Hm. I'm fine with adding more docs. But they should be written in a form that is comprehensible for everyone and not a listing of edge case recommendations.

ksteinb: LDAP isn't an edge case in large environments 

Greetings from CS3

ksteinb: could you tell me where the error in eslint comes from?  I was not yet able to setup the complete build environment locally

ChristophWurst: > Greetings from CS3

Say hi to Björn and Frank :)

> could you tell me where the error in eslint comes from? I was not yet able to setup the complete build environment locally

ESLint want the `t` call to be wrapped into multiple lines. You can run `npm run lint:autofix` to resolve this automagically.

ksteinb: Arghh, such a picky thing
i got it now (blind test, as I didn't get eslint running on my Ubuntu LTS 18.04)
but now xml-linters failed, looks like it had connection trouble to apps.nextcloud.com

jancborchardt:
@ksteinb is this ready for review again, or what’s missing? :)

ChristophWurst:
@ksteinb :ping_pong:

ChristophWurst:
closing due to lack of activity, but we can re-open easily if we decide to continue working on the change :)

src/components/settings/ProvisioningSettings.vue
wKollendorf: This should fix it
```
<small>
  {{ t('mail', '* %USERID% %DISPLAYNAME% and %EMAIL% will be replaced with the user\'s UID, display name, and email.') }}<br>
  {{ t('mail', '* In LDAP environments, %USERID% is usually meaningless if IMAP/SMTP login is not possible.') }}<br>
  {{ t('mail', '* With email login, you can work around this by provisioning the display name with the login attribute.') }}
</small>
```

PR #2569, opened 2020-01-28
Do not double-encode mailto parameters

ChristophWurst:
Steps to reproduce

* Set up maito: handler from settings menu (#2539)
* Open contacts menu
* Click the mail icon on an entry that has an email address

Before: get the new message page with an email like `user%40domain.tld` filled in
After: get new message page with `user@domain.tld` filled in

The reason for this bug is that `parse_url` does *not* URL decode the parameters, hence they ended up being encoded twice.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -166,7 +166,7 @@ class PageController extends Controller {
 
 		array_walk($params,
 			function (&$value, $key) {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
https://github.com/nextcloud/server/pull/19176 actually replaces this PR.

PR #2585, opened 2020-01-31
Fix folderStats

GretaD:
WIP - it is fixed for inbox, but not for the rest (draft, sent ....)

Error in render: "TypeError: vars is undefined"

### DIFF
--- a/src/components/NavigationFolder.vue
+++ b/src/components/NavigationFolder.vue
@@ -76,6 +76,7 @@ import ActionText from '@nextcloud/vue/dist/Components/ActionText'
 
 import {getFolderStats} from '../service/FolderService'
 import logger from '../logger'
+import {translatePlural as n} from '@nextcloud/l10n'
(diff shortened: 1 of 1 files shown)


src/components/NavigationFolder.vue
```diff
@@ -136,7 +138,7 @@ export default {
 						total: this.folderStats.total,
 					})
 				} else {
-					return n('mail', '{unread} unread of {total}', '{unread} unread of {total}', {
+					return n('mail', '{unread} unread of {total}', '{unread} unread of {total}', undefined, {
```
ChristophWurst:
```suggestion
					return n('mail', '{unread} unread of {total}', '{unread} unread of {total}', this.folderStats.total, {
```

I think

GretaD: makes sense, but still the draft sent ect are not loading.. 
[ERROR] mail: could not load folder stats

mario:
Ty for fixing this!

Issue #2586, opened 2020-01-31
Search regressions


ChristophWurst:
_Originally posted by @ChristophWurst in https://github.com/nextcloud/mail/pull/2064_

PR #2617, opened 2020-02-05
Fix searching in message cache (and IMAP)

ChristophWurst:
* Query IMAP only if necessary
* Fix pagination (when IMAP results are used)
* Fix CC and BCC search queries

Fixes https://github.com/nextcloud/mail/issues/2586

cc @nextcloud/mail for testing

Before this PR search is unreliable. Now it should yield correct results.

### DIFF
--- a/lib/Contracts/IMailSearch.php
+++ b/lib/Contracts/IMailSearch.php
@@ -24,9 +24,9 @@
 namespace OCA\Mail\Contracts;
 
 use OCA\Mail\Account;
+use OCA\Mail\Db\Message;
 use OCA\Mail\Exception\ClientException;
 use OCA\Mail\Exception\ServiceException;
-use OCA\Mail\Model\IMAPMessage;
 
 interface IMailSearch {
 
@@ -36,7 +36,7 @@ interface IMailSearch {
 	 * @param string|null $filter
 	 * @param string|null $cursor
 	 *
-	 * @return IMAPMessage[]
+	 * @return Message[]
 	 * @throws ClientException
 	 * @throws ServiceException
 	 */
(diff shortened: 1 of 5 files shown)


Ornias1993:
I think this is the right issue to bring this up:

> Query IMAP only if necessary

While you are at it, it might be worthwhile to add button/feature to force download the whole imap server. Because there is a case where querying an IMAP server isn't feasable:

- IMAP server is external
- Pipe to the imap server isn't too big (lets say 10MB/s)
- IMAP server contains GB's (lets say 4GB) of small emails

Even preloading all email from the server takes a few hours, but queries to the IMAP server might take a LONG time as well. (minutes)

If we add to this the following situation:
- Using a reverse proxy or other execution time limiter

We end up with 504 A LOT, because the proxy or other limiter isn't going to accept minutes of wait time to finish such a query.

If we, as admin, could force Nextcloud to preload the whole IMAP account, this would be sped up even byond this PR and should be relatively easy to add even within scope of this PR.

TL:DR
Not everyone is able to have a decent connection to their IMAP server, please give us the option to keep more inside Nextcloud.

ChristophWurst:
Hi @Ornias1993 :v: 

while we don't fully *mirror* IMAP yet, you'll be very happy about what we've done with #2064.

ChristophWurst:
But feel free to shoot a feature request to https://github.com/nextcloud/mail/issues/new?assignees=&labels=enhancement%2C+0.+to+triage&template=feature_request.md&title= with the overall idea.

Ornias1993:
@ChristophWurst Thanks for the reply, when i've the time i'll throw one in.
Should be relatively easy to implement considering #2064 

Just noticed #2064 was just two days ago, I hope that one solves some issues already :)

ChristophWurst:
> Just noticed #2064 was just two days ago, I hope that one solves some issues already :)

It definitely does. You'll be blown away by how responsive everything becomes :)

Ornias1993:
I've put my (fictional) money where my mouth is and made the feature request right away:
#2621

PR #1753, opened 2019-05-13
Drag'n'Drop support

StCyr:
Drag'n'Drop support implementation with https://github.com/Vivify-Ideas/vue-draggable

Remaining issues:

- [ ] It seems the code isn't properly initialized: If you load the mail apps from your browser, the drag'n'drop functionality won't work directly. You first have to perform an action (eg: selecting another email than the default one displayed) before the drag'n'drop functionality will work;
- [ ] Shouldn't navigate when the moved message is not tghe one currently displayed
- [ ] Do not expand when folder's account is not the same as envelope account
- [ ] Polishing

Issue #890, opened 2018-05-04
Multiselect / Make it possible to delete several mails at once [$35]


hitam4450:
Hi,
It would be very advantageous, if one could delete several mails at once and not only one by one !

regards, hitam4450

laurentiu2:
Definitely needed. Also add the possibility to empty Trash and Spam more easily.

Laurentiu

ChristophWurst:
To potential bounty hunters:

1. Let us know when you plan working on this
2. Let's discuss the implementation together
3. …
4. Profit

aharabara:
> To potential bounty hunters:
> 
>     1. Let us know when you plan working on this
> 
>     2. Let's discuss the implementation together
> 
>     3. …
> 
>     4. Profit

Hi @ChristophWurst . I did a POC for this feature and you can checkout it from my fork on `feature/890/bulk-operations` branch. Use `a` to select/unselect messages and `d` to deselect all messages. All selected messages can be `unseen`, `deleted` and `starred`.
You can find it [here](https://github.com/aharabara/mail/tree/feature/890/bulk-operations)

ChristophWurst:
@StCyr didn't you work on this?

@aharabara quick remarks:

* Selected state is something local. I wouldn't put that into the store.
* The multi select feature is very useful for other apps as well. We should implement this in the components of https://github.com/nextcloud/nextcloud-vue

ChristophWurst:
(wrong button, sorry)

ChristophWurst:
-> https://github.com/nextcloud/mail/issues/2273#issuecomment-567828441

StCyr:
> @StCyr didn't you work on this?

Yes, here: https://github.com/nextcloud/mail/tree/feature/2273/multi-select
 
> @aharabara quick remarks:
> 
>     * Selected state is something local. I wouldn't put that into the store.

I find it nice that he's using the 'selected' flag. I had try at first but couldn't succeed and you also told me to use something local.  

What's the purpose of the 'selected' flag? It's not even use atm....

>     * The multi select feature is very useful for other apps as well. We should implement this in the components of https://github.com/nextcloud/nextcloud-vue

Wow, how do u imagine this?

ChristophWurst:
> Wow, how do u imagine this?

We have existing components for the list and the items (though Mail uses its own implementation for the item right now), so we can add a prop `selectable` to the list or items components and if it's true then we offer this functionality. The list component will have the actions available for the selected items.

Does that make sense?

[...]

PR #2619, opened 2020-02-05
Actions on multiple message

StCyr:
This PR is to fix https://github.com/nextcloud/mail/issues/2273

#### TODO

- [ ] multiselect header bar moves with the envelopes' list on iOS
- [ ] indexes of selected envelopes get wrong when a new message is received (all indexes should be increased by 1)


jancborchardt:
> proper icons for "mark read" and "mark unread"

You can just use the `icon-mail` for both states as we do in the action menu as well. :)

jancborchardt:
Also, can you check out the conflicts and rebase on latest master? :)

ma12-co:
@StCyr, when this branch is checked out i get a 500 error while fetching the envelopes, so all my folders are empty.

Screenshot from 2020-03-03 16-23-19

StCyr:
> 
> 
> @StCyr, when this branch is checked out i get a 500 error while fetching the envelopes, so all my folders are empty.
> 
> Screenshot from 2020-03-03 16-23-19

Did you run "make dev-setup" after having checked out the branch?

Maybe the branch needs a rebase too...

ma12-co:
> Did you run "make dev-setup" after having checked out the branch?

yep

> Maybe the branch needs a rebase too...

I tried but it's not a clean one, there are conflicts to be resolved

StCyr:
> > Maybe the branch needs a rebase too...
> 
> I tried but it's not a clean one, there are conflicts to be resolved

Hmm, indeed. I'll have a deeper look

StCyr:
> > Maybe the branch needs a rebase too...
> 
> I tried but it's not a clean one, there are conflicts to be resolved

Branch rebased. Feel free to hack around :-)

src/components/EnvelopeList.vue
```diff
@@ -38,16 +61,118 @@ export default {
...
 }
 </script>
 
 <style scoped>
+.multiselect-header {
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	justify-content: center;
+	background: #0080ff;
```
jancborchardt: The background should be the same as for the files controls bar :)
```suggestion
	background-color: var(--color-main-background-translucent);
```

jancborchardt:
Heh, I left the design comment on a different issue than your original one, at https://github.com/nextcloud/mail/issues/890#issuecomment-591289668 :)

> Interface-wise:
> 
> * The action menu of each item should get an entry called "Select", which then triggers multiselect mode and selects that message.
> * In that mode, a select box is shown overlaying the avatar part of every message in the list. Clicking on the select box toggles the selection (of course), but the whole avatar part is click-sensitive for proper clickability/tappability.
> * Clicking the message part opens the message so you can check if you want to select that message. It’s important that you can still open messages in multiselect mode.
> * We show floating actions on the top just like for files. We should decide on 1 or max 2 ones which show directly there, with icon + text, and the rest shows in a 3-dot menu. Possible actions ("7" is a stand-in for the count of messages which we show as feedback. Is shorter than "Mark selected unread"):
>    * "Favorite 7": Toggles with "Unfavorite 7". If all are favorited, directly show "Unfavorite 7".
>    * "Mark 7 unread": Toggles with "Mark 7 read". If all are unread, directly show "Mark 7 read".
>    * "Move 7"
>    * "Deselect 7"
>    * "Delete 7"
> 
> 
> This seems to work well for Mail and also for e.g. Contacts and others (it’s also similar to what was discussed previously about how we do multiselect in the Files grid view).

src/components/EnvelopeList.vue
StCyr: It's just plain white?

image

StCyr:
`ActionCheckbox` doesn't seem to work well when the sole action in an `Actions` component:

Mail - Nextcloud - Mozilla Firefox 2020-03-06 16-53-20

src/components/EnvelopeList.vue
jancborchardt: @StCyr exactly, until you scroll when it overlays the list translucent. :) That’s how it should be.
In reference to my comment at https://github.com/nextcloud/mail/pull/2619#issuecomment-595652200 – if we show 1 or 2 actions directly in there like "Mark 7 unread", we can use the `primary` class on the action so it appears blue and calls attention. But the background should be plain. :)

StCyr:
Hi @nextcloud/mail 

Could you have a look at my code?

Do you know why the checkoxes won't get checked? CSS issue?

StCyr:
Hey @nextcloud/mail,

Could you have a look at my last commit (or hereunder code)? 

I'm trying 2 things for the checkbox. 

The first one (active one in the code hereunder) makes it possible to change the selection checkbox's `checked` state from the action menu, but, when clicking on the checkbox directly, the checkbox's layout isn't updated (ie: the checkbox isn't visualy checked)

The second one (the commented one in the code hereunder) works perfectly except that the user navigates to the selected message.

Do you have an idea how to fix one of these solutions? 

```js
		<input v-show="selectMode" ref="selectCheckbox" class="app-content-list-item-select-checkbox" type="checkbox" @click.prevent="onToggleSelect" />
	<!--		<input v-show="selectMode" ref="selectCheckbox" class="app-content-list-item-select-checkbox" type="checkbox" @click.stop="$emit('select', data)" /> -->
```

ma12-co:
> The second one (the commented one in the code hereunder) works perfectly except that the user navigates to the selected message.

@StCyr for me it works perfectly, it doesn't navigate to the clicked envelope.

On a different note I would suggest you to use a store for keeping track of the selected users instead of doing that in the `EnvelopeList` component. l I think it's quite a bit cleaner. then you can bind the `:checked` status of the checkbox to a computed property like:
```
checked() {
    return this.$store.getters.isEnvelopeSelected(envelope)
}
```
I've just done something similar for the selection of participants in talk's new group conversation creation form if you want to take a look:

this is the store, it's very simple:
https://github.com/nextcloud/spreed/blob/d4e6ed6ce7a3b7dfbc8170565fb68985882f7767/src/store/newGroupConversationStore.js 

this is an action dispatched from a component to update the selection: 
https://github.com/nextcloud/spreed/pull/3078/files#diff-6aa320fa8f78384ca9a71f76961c1003R230

and this is a getter: 
https://github.com/nextcloud/spreed/pull/3078/files#diff-9a8b3a61fcf6e77158448571e632fd1fR88-R89

If you have any question on this let me know :)

StCyr:
> @StCyr for me it works perfectly, it doesn't navigate to the clicked envelope.

Damned. You're right: it works perfectly here on Safari/MacOS but not on FF/MacOS.

However, though I've found that many people have also troubles with stopping event propagation on Firefox, I haven't find how to fix it here :(

> On a different note I would suggest you to use a store for keeping track of the selected users instead of doing that in the EnvelopeList component. 

Hmm... I think we already had this discussion with @jancborchardt and he was advocating exactly the opposite: don't make use of the store for this feature :)

ChristophWurst:
> Hmm... I think we already had this discussion with @jancborchardt and he was advocating exactly the opposite: don't make use of the store for this feature :)

It was me ;)

src/components/EnvelopeList.vue
```diff
@@ -1,25 +1,58 @@
...
+			</div>
+		</transition-group>
 		<transition-group name="list">
 			<div id="list-refreshing" key="loading" class="icon-loading-small" :class="{refreshing: refreshing}" />
 			<Envelope
 				v-for="env in envelopes"
 				:key="env.uid"
 				:data="env"
 				:folder="folder"
+				:selected="env.flags.selected"
```
ChristophWurst: as noted before this should be local state. like an array or object in this component with all selected items. and here you pass that info to the message.

what you're currently do is
1) use global state
2) mutate Vuex state outside a mutation. Vue should already warn you about that

StCyr: ok resolved

StCyr:
>     * Wording of delete action should include the number: "Delete 7 selected"

Simply "Delete 7" looks nicer to me

StCyr:
> Selected emails should also have the greyish background
Wouldn't be the user confused if we use the same background color as the one for the currently active message?

StCyr:
Dear @nextcloud/mail,

I've fixed most of the issue. Could you please have another look at it?

jancborchardt:
> > Selected emails should also have the greyish background
> 
> Wouldn't be the user confused if we use the same background color as the one for the currently active message?

You are right, good point! Actually the currently open mail should have `background-color: var(--color-primary-element-light);` just like we do for the currently open folder in the navigation. :)

StCyr:
> The multiselect action bar should have translucent background via background-color: var(--color-main-background-translucent);

It's already the case

> The "Select" action is better moved down in the list before "Delete", as both "Favorite" and "Mark unread" are more important – this goes for the mail action menu as well as for the multiselect action menu.

I've moved it at the far bottom of the list, after the "Delete" action.

> You are right, good point! Actually the currently open mail should have background-color: var(--color-primary-element-light); just like we do for the currently open folder in the navigation. :)

ok, done (that's `var(--color-primary-light);` actualy)

jancborchardt:
> I've moved it at the far bottom of the list, after the "Delete" action.

_Before_ Delete please, sorry – the sentence was a bit complicated. :) Delete should always be the last item in an action list.

StCyr:
> * The animation seems to be on the button, but the message list immediately moves down. The message list should also move with a quick animation.
> * Same when you click one of the actions and the bar vanishes: It should have an animation.

I don't see the problem; When I slow down the animation using `transition: all var(--animation-slow)`, everything seems to move at the same space

jancborchardt:
And code review please @nextcloud/mail :)

StCyr:
>     * When in select mode, clicking the then visible checkboxes should also select or unselect a message. This doesn’t seem to work and you have to use the dropdown to select or unselect messages.

Yes, I've already mentioned this. 

It seems like they became unfunctional when I've implemented the nextcloud design (ie: using a label)

StCyr:
closing in favor of https://github.com/nextcloud/mail/pull/1753

jancborchardt:
Did you recently rebase on master @StCyr? Possibly this "what to do after message deletion" fix by @ahstro fixed that issue with the "Message not found" note when deleting the currently open message and the ones around in a batch?
https://github.com/nextcloud/mail/pull/2697

jancborchardt:
@StCyr now that #2697 is merged, is this still blocked? Anywhere you need help on? :)

StCyr:
> @StCyr now that #2697 is merged, is this still blocked? Anywhere you need help on? :)

Yes, please help for the following:

I can't understand why:

1. The checkboxes don't work anymore as soon as I make them using the "checkbox" class (see commit https://github.com/nextcloud/mail/pull/2619/commits/8ab85e49b0dd272786bb1ccc50ab40d4c9fdb839. When it's removed checkboxes don't work anymore)
2. The envelope list is pushed down when the multiselect header bar appears, though the header bar has a z-index of 100

Best regards,

jancborchardt:
@juliushaertl @ChristophWurst @skjnldsv any idea about the checkbox issue? ^

skjnldsv:
Done!

- Please avoid using $emit in template, this is not super clean :)
- When using `:selected="prop"` you are not forwarding the change to the parent, therefore selected is not updated and can never change ;)

StCyr:
> Done!
> 
>     * Please avoid using $emit in template, this is not super clean :)
> 
>     * When using `:selected="prop"` you are not forwarding the change to the parent, therefore selected is not updated and can never change ;)

So... Maybe you could now find out why the multiselect header bar pushes the envelope list down when it shows up?!? :-D (Am I abusing you? ;-))

skjnldsv:
> So... Maybe you could now find out why the multiselect header bar pushes the envelope list down when it shows up?!? :-D (Am I abusing you? ;-))

No problem :)
What is expected? Animation?

StCyr:
> > So... Maybe you could now find out why the multiselect header bar pushes the envelope list down when it shows up?!? :-D (Am I abusing you? ;-))
> 
> No problem :)
> What is expected? Animation?

The goal is to display the multiselect header bar above the envelope list. 

But, when it appears, during the animation, it doesn't seem to be displayed above the envelope list: It pushes the envelope list down (creating an annoying displacement of the envelope list which we would like to avoid).

skjnldsv:
> But, when it appears, during the animation, it doesn't seem to be displayed above the envelope list: It pushes the envelope list down (creating an annoying displacement of the envelope list which we would like to avoid).

so it would hide the first Envelope?

StCyr:
> > But, when it appears, during the animation, it doesn't seem to be displayed above the envelope list: It pushes the envelope list down (creating an annoying displacement of the envelope list which we would like to avoid).
> 
> so it would hide the first Envelope?

Yes, it should directly hide the first envelope (the displacement is really annoying). 

There's some transparency though....

Ah... but you're right: that would probably hinder the user to click on the first envelope... what do you think @jancborchardt ?

StCyr:
Hello @nextcloud/mail folks.

Could you review this PR once again? @skjnldsv solved the issue I had with nextcloud's checkbox style and I think it's 100% functional now.

@jancborchardt still had an issue with the way the multiselect header bar appears, but I'm not sure to understand his request in the end:

1. The multiselect header bar has a z-index of 100 to shows above the envelope list;
2. However, it still should push the envelope list down when it appears in order to not hide the first envelope in the list.

So, I'm not sure what @jancborchardt is asking exactly. Please have a look at the TODO list in the first comment of this PR.

StCyr:
> Could you review this PR once again? @skjnldsv solved the issue I had with nextcloud's checkbox style and I think it's 100% functional now.

Err... There's still the problem that the header bar is not sticky on small screens (eg: iphone and firefox's reactive view)

jancborchardt:
> > Could you review this PR once again? @skjnldsv solved the issue I had with nextcloud's checkbox style and I think it's 100% functional now.
> 
> Err... There's still the problem that the header bar is not sticky on small screens (eg: iphone and firefox's reactive view)

Yeah, that seems to be the main thing – on desktop it seems all fine! :)

Sometimes I still get a "Message not found" error when I delete multiple messages of which the currently open message is part of.

@ChristophWurst what’s your call – should these issues be fixed before merging, or do we merge and then fix going forward?

StCyr:
Hi

Could we merge this branch?

I use it all the time. It's really usefull :-)

Issue #2611, opened 2020-02-04
Writing mode


ramikilany:
### Feature Request

In the nextcloud mail, there is an option which is the writing mode when I switch to Rich text mode nothing happens in the newly created message. In the end, I need to have a tool menu like the Wordpad or the menu which has the align-left, center and right for example text bold, italic and underline, and so on. and if would be an option to write from right-to--left message (Arabic for example)
#### Summary
Text menu toolbar (rich text mode)

ChristophWurst:
It's an inline editor. Focus some text and a magic toolbar will appear :)

StCyr:
there's no text-alignment and heading tools though :(

ChristophWurst:
-> https://github.com/nextcloud/mail/pull/2623

ChristophWurst:
> and if would be an option to write from right-to--left message (Arabic for example)

The editor *should* do this automatically: https://ckeditor.com/docs/ckeditor5/latest/features/ui-language.html#righttoleft-rtl-languages-support. But right now there seems to be an issue with translations.

PR #2623, opened 2020-02-06
Add headings and alignment formatting

ChristophWurst:
From https://github.com/nextcloud/mail/issues/2611#issuecomment-582392828 @StCyr 

> there's no text-alignment and heading tools though :(

While working on this I noticed that translations don't work. Setting my lang to "de" still shows me an English interface. Needs investigation -> will open a ticket.

I think @hanserasmus will appreciate this :v:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3401,6 +3401,17 @@
         "@ckeditor/ckeditor5-undo": "^16.0.0"
       }
     },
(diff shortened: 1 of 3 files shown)


ChristophWurst:
@nextcloud/mail testing welcome :)

StCyr:
works good here:

image

PR #2625, opened 2020-02-06
Save sender email for messages read/starred/sent/deleted

### LINKED ISSUE #2558: Collect user interaction data
For https://github.com/nextcloud/mail/projects/7 we need data about important interactions of a user.

- [x] Requires https://github.com/nextcloud/mail/issues/2559

As a start we should focus on the people who interacted, hence only consider the email addresses. In a step we can extend the logic to also collect subject or even the message content.

* Messages read -> save sender email
* Messages starred -> save sender email
* Messages deleted -> save sender email
* Message sent -> save all recipients (should also work for replies)

I would propose to do this in two parts
* Emit events with https://nextcloud.github.io/nextcloud-event-bus/ for the actions above -> @GretaD 
* Debounce and combine events and send to back-end -> @ChristophWurst 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/87622135-collect-user-interaction-data?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GretaD:
fixes #2558


src/components/Composer.vue
```diff
@@ -421,7 +422,11 @@ export default {
 				.then(this.getMessageData())
 				.then(data => this.send(data))
 				.then(() => logger.info('message sent'))
-				.then(() => (this.state = STATES.FINISHED))
+				.then(() => {
+					emit('mail:interaction', {type: 'message-sent', recipient: this.selectedUser.user})(
```
ChristophWurst: :thinking: These are two statements. the `emit` call and the assignment. You create a very creative syntax :)

Just make it
```
.then(() => {
  emit('mail:interaction', {type: 'message-sent', recipient: this.selectedUser.user})
  this.state = STATES.FINISHED
}
```

src/components/Envelope.vue
```diff
@@ -127,6 +128,7 @@ export default {
 	methods: {
 		onToggleFlagged() {
 			this.$store.dispatch('toggleEnvelopeFlagged', this.data)
```
ChristophWurst: use .then(...) to await the change on the server. And we should only track this when an unflagged message gets flagged.

src/components/Message.vue
```diff
@@ -232,6 +233,7 @@ export default {
 		},
 		onToggleSeen() {
 			this.$store.dispatch('toggleEnvelopeSeen', this.envelope)
+			emit('mail:interaction', {type: 'message-read', recipient: this.selectedUser.user})
```
ChristophWurst: again. use `this.$store.dispatch(...).then(() => { ... })` to await the store action

src/components/Message.vue
```diff
@@ -252,6 +254,7 @@ export default {
 
 			this.$emit('delete', this.envelope)
 			this.$store.dispatch('deleteMessage', this.envelope)
+			emit('mail:interaction', {type: 'message-delete', recipient: this.selectedUser.user})
```
ChristophWurst: wait the store action

src/components/Envelope.vue
ChristophWurst: hmmm and I'm pretty sure `this.selectedUser.user` does not work `Envelope.vue` ;)

src/components/Composer.vue
GretaD: thats what i did, but with bracket :P i like bracket

src/components/Envelope.vue
GretaD: recipient: this.recipient.email ?

ChristophWurst: sounds right!

src/components/Composer.vue
ChristophWurst: the problem is that you did `emit(...)(...)` which means you treated the return value of `emit(...)` as a function which you ten called :wink:

ChristophWurst:
We won't need this right now as we already have that data available: https://github.com/nextcloud/mail/issues/2558#issuecomment-617291935

Issue #2622, opened 2020-02-06
Envelopes don't get removed from envelopes list


StCyr:
### Expected behavior
Envelope of deleted messages should automaticaly be removed from envelope list

### Actual behavior
Envelopes don't get deleted. Refreshing the page several times doesn't even fix this.

Cliking on the envelope of a deleted message shows this:

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)

### Server configuration

**Operating system**: (e.g. Debian 8)

**Web server:** (e.g. Apache, Nginx,...)

**Database:** (e.g. MariaDB, SQLite or PostgreSQL)

**PHP version:** (e.g. 7.0)

**Nextcloud Version:** (see admin page, e.g. 13.0.2)

#### Client configuration
**Browser:** (e.g. Firefox 48)

**Operating system:** (e.g. Arch Linux)

StCyr:
related to https://github.com/nextcloud/mail/pull/2618

ChristophWurst:
Also related to #2064. I have yet to find out why this happens. Doesn't occur all the time for me. @rullzer noticed it as well.

ChristophWurst:
As a start we should probably just delete the message from the cache. Right now we only change IMAP and hope that the next sync will update the cache accordingly.

ChristophWurst:
> As a start we should probably just delete the message from the cache. Right now we only change IMAP and hope that the next sync will update the cache accordingly.

https://github.com/nextcloud/mail/pull/2634

PR #2634, opened 2020-02-10
Delete cached messages when deleting them on IMAP

ChristophWurst:
While strictly speaking we get inconsistent with the IMAP mailbox due to bypassing the sync machanism, this will show the user a more up-to-date representation of the mailbox, and making the sync a tiny bit faster (one row less to delete).

Fixes https://github.com/nextcloud/mail/issues/2622


StCyr:
Seems to work here :)

PR #2635, opened 2020-02-10
Update Docker in the hope of getting a more stable version

ChristophWurst:
Ref https://docs.travis-ci.com/user/docker/#installing-a-newer-docker-version


ChristophWurst:
same issues as before

PR #2637, opened 2020-02-10
Remove subfolders from Favorites

### LINKED ISSUE #2602: Can't add subfolder to "Favorites" mailbox
### Expected behavior
The "Favorites" mailbox (ID: INBOX/FLAGGED) is virtual and does not exist on IMAP. Thus the action to add a subfolder does not make sense.

### Actual behavior
There is an action to add a subfolder.

GretaD:
fixes #2602
now it looks like in the pic below, since the only option is mark all as read. Is this ok? @ChristophWurst 

favorite_subfolder

### DIFF
--- a/src/components/NavigationFolder.vue
+++ b/src/components/NavigationFolder.vue
@@ -51,7 +51,7 @@
 					{{ t('mail', 'Mark all messages of this folder as read') }}
 				</ActionButton>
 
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> now it looks like in the pic below, since the only option is mark all as read. Is this ok? @ChristophWurst

Does mark all as read even work for favorites? I think we can remove it.

Issue #2530, opened 2020-01-22
"All inboxes" not translated


ChristophWurst:
### Expected behavior
Get "All mailboxes" (the unified inbox) translated

### Actual behavior
See English text

@rakekniven I tried to find the string on Transifex but failed. Is it missing there? :thinking:

rakekniven:
I cannot find the string in this repo either? Which file should contain the source?

ChristophWurst:
At https://github.com/nextcloud/mail/blob/c8d90e539cb0ac41d23eee31e5d869201d5b71f7/src/l10n/MailboxTranslator.js#L44

ChristophWurst:
oh it's "All inboxes" and not "All mailboxes". Sorry.

rakekniven:
Ok, string is actually not present at Transifex.

ChristophWurst:
Okidoke. I have to check our extraction tool.

ChristophWurst:
Related: https://github.com/nextcloud/mail/issues/2290

ChristophWurst:
It's back https://github.com/nextcloud/mail/commit/88bdce8a1ffe669e5452bce001d45e766082ccb1 :rocket:

PR #2639, opened 2020-02-11
Extract js translation strings with gettext and commit them

ChristophWurst:
Not sure if this actually works, but it looks like we can have .pot files in `translationfiles/templates` and they will be picked up by the l10n bot: https://github.com/nextcloud/docker-ci/blob/8f896e1b82051c9c863496cd680935a011e05daf/translations/handleAppTranslations.sh#L44-L48

For #2530

### DIFF
--- a/.github/workflows/l10n.yml
+++ b/.github/workflows/l10n.yml
@@ -1,0 +1,19 @@
+name: Lint
+on: pull_request
+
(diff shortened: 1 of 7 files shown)


ChristophWurst:
Mostly fixed by #2645

PR #2644, opened 2020-02-11
Change forward msg

### LINKED ISSUE #2642: When you forward an email, the feedback message is wrong
Its says Reply Sent! Instead should be The email was sent or another message

GretaD:
I think new message sent, reply message and forward message should have the same feedback message

Fixes #2642

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -173,8 +173,7 @@
 		<button class="button primary" @click="onSend">{{ t('mail', 'Retry') }}</button>
 	</div>
 	<div v-else class="emptycontent">
-		<h2 v-if="!isReply">{{ t('mail', 'Message sent!') }}</h2>
-		<h2 v-else>{{ t('mail', 'Reply sent!') }}</h2>
+		<h2>{{ t('mail', 'Message sent!') }}</h2>
 		<button v-if="!isReply" class="button primary" @click="reset">
 			{{ t('mail', 'Write another message') }}
 		</button>


src/components/Composer.vue
```diff
@@ -174,7 +174,7 @@
 	</div>
 	<div v-else class="emptycontent">
 		<h2 v-if="!isReply">{{ t('mail', 'Message sent!') }}</h2>
-		<h2 v-else>{{ t('mail', 'Reply sent!') }}</h2>
+		<h2 v-else>{{ t('mail', 'Message sent!') }}</h2>
```
ChristophWurst: sure, but in that case we don't need the if-else, do we? :)

GretaD: We need one for new message v-if, and one for everything else, v-else.

ChristophWurst: Look again. It's the same element in both cases: `<h2>{{ t('mail', 'Message sent!') }}</h2>`

GretaD: Please check isReply function. If we remove v-else, reply and forward will not have a feedback message.

ChristophWurst: Let me rephrase. Instead of

```html
<h2 v-if="!isReply">{{ t('mail', 'Message sent!') }}</h2>
<h2 v-else>{{ t('mail', 'Message sent!') }}</h2>
```

we can simply do

```html
<h2>{{ t('mail', 'Message sent!') }}</h2>
```

`v-if` and `v-else` render the same element.

GretaD: dammit

PR #2648, opened 2020-02-13
Add priority inbox "light"

GretaD:
For https://github.com/nextcloud/mail/projects/7#card-31900324

Done:

- [x] Add SectionTitle Component
- [x] Add starred section
- [x] Add other section

To do: 
- [ ] Add priority section

Bugs
- [x] https://github.com/nextcloud/mail/pull/2706 Refresh pulls in unrelated elements (non-starred ones into starred and the other way around)
- [x] Starred should show *all* messages. Only *other* is paginated
- [x] First message not opened correctly
- [x] Messages sync'ed in the background won't show in PI because we use the filtered envelope lists here

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -74,6 +74,14 @@ export default {
 			type: Object,
 			required: true,
 		},
+		paginate: {
+			type: Boolean,
+			default: true,
+		},
+		openFirst: {
+			type: Boolean,
+			default: true,
+		},
 		searchQuery: {
 			type: String,
 			required: false,
@@ -111,6 +119,7 @@ export default {
 		this.bus.$on('delete', this.onDelete)
 		this.bus.$on('shortcut', this.handleShortcut)
 		this.loadMailboxInterval = setInterval(this.loadMailbox, 60000)
+		window.addEventListener('focus', this.onWindowFocus)
 	},
 	async mounted() {
 		return await this.loadEnvelopes()
@@ -120,6 +129,7 @@ export default {
 		this.bus.$off('delete', this.onDelete)
 		this.bus.$off('shortcut', this.handleShortcut)
 		this.stopInterval()
+		window.removeEventListener('focus', this.onWindowFocus)
 	},
 	methods: {
 		initializeCache() {
@@ -150,13 +160,14 @@ export default {
 					accountId: this.account.id,
 					folderId: this.folder.id,
 					query: this.searchQuery,
+					paginate: this.paginate,
 				})
 
-				logger.debug('envelopes fetched', {envelopes})
+				logger.debug(envelopes.length + ' envelopes fetched', {envelopes})
 
 				this.loadingEnvelopes = false
 
-				if (!this.isMobile && this.$route.name !== 'message' && envelopes.length > 0) {
+				if (this.openFirst && !this.isMobile && this.$route.name !== 'message' && envelopes.length > 0) {
 					// Show first message
 					let first = envelopes[0]
 
@@ -165,8 +176,8 @@ export default {
 					this.$router.replace({
 						name: 'message',
 						params: {
-							accountId: this.account.id,
-							folderId: this.folder.id,
+							accountId: this.$route.params.accountId,
+							folderId: this.$route.params.folderId,
 							filter: this.$route.params.filter ? this.$route.params.filter : undefined,
 							messageUid: first.uid,
 						},
@@ -206,8 +217,8 @@ export default {
 
 			try {
 				await this.$store.dispatch('fetchNextEnvelopePage', {
-					accountId: this.$route.params.accountId,
-					folderId: this.$route.params.folderId,
+					accountId: this.account.accountId,
+					folderId: this.folder.id,
 					envelopes: this.envelopes,
 					query: this.searchQuery,
 				})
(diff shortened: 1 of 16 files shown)


ChristophWurst:
Please update to latest master for #2669, then we can look into the "light" version of the priority inbox with *starred* and *other* sections

ChristophWurst:
> Refresh pulls in unrelated elements (non-starred ones into starred and the other way around)

- [ ] requires https://github.com/nextcloud/mail/pull/2706 so we can clean up the server-side logic

src/components/MailboxMessage.vue
```diff
@@ -86,7 +123,7 @@ export default {
 			return list.length > 0
 		},
 		showMessage() {
-			return this.hasMessages && this.$route.name === 'message'
+			return (this.folder.isPriorityInbox === true || this.hasMessages) && this.$route.name === 'message'
```
ahstro: I'm not sure if `isPriorityInbox` can have different values and we want this specifically if it's `true`, but if not we might want to skip `=== true` to unify how truthiness is checked with `this.hasMessage`? 😇

src/components/Mailbox.vue
```diff
@@ -162,8 +171,8 @@ export default {
 					this.$router.replace({
 						name: 'message',
 						params: {
-							accountId: this.account.id,
-							folderId: this.folder.id,
+							accountId: this.$route.params.accountId,
+							folderId: this.$route.params.folderId,
```
ahstro: I'm just curious: why were these changed from accessing the values directly on `this` to using the route params, while the other uses were changed the other way around? I'm not familiar with the code yet, so just trying to grok things 😅

src/i18n/MailboxTranslator.js
```diff
@@ -46,6 +46,11 @@ const translateSpecial = folder => {
 			return t('mail', 'Inbox')
 		}
 	}
+	if (folder.specialUse.includes('priority')) {
+		if (folder.isPriorityInbox === true) {
```
ahstro: Same as L89?

src/store/index.js
```diff
@@ -55,6 +62,16 @@ export default new Vuex.Store({
 				folders: [],
 				envelopeLists: {},
 			},
+			[PRIORITY_INBOX_UID]: {
+				id: PRIORITY_INBOX_ID,
+				accountId: 0,
```
ahstro: Might be out of the scope of this PR, but why are there both an `account.id` and `account.accountId`? And why can they be different?

ChristophWurst: Someone once had the idea to use `accountId` instead of `id`, while now we have both. To be honest I don't know why that is, but we should slowly move towards just `id` :)

src/components/Mailbox.vue
ChristophWurst: Hm. We definitely need it for the correct routing. But for the `fechEnvelope` call I'm not so sure anymore. I have to have another look :grin:

src/store/index.js
ahstro: Hmm, then should these not be different here? Assuming this difference is important for L205 and L302? Urgh, I hate asking questions about other people's code 😖

src/i18n/MailboxTranslator.js
GretaD: Sorry, i didnt get this one? What do you mean?

ahstro: Oh, sorry :) I meant that `=== true` isn't needed, because `if (folder.isPriorityInbox)` checks truthiness

src/store/index.js
ChristophWurst: > I hate asking questions about other people's code confounded

No, please do :)

I think we should unify accountId and id and use it consistently.

> Hmm, then should these not be different here?

No, this new inbox is also part of the unified account. Though we might reconsider the name *Unified account* now that it has more than one purpose.

ChristophWurst:
- [ ] BUG (possible on master as well) when you scroll down *lots* of request are dispatched for the next message page and it does not seem to stop

ChristophWurst:
> If you don’t have any favorites, does the section hide? (It should – I didn’t test this as I didn’t want to unfavorite all my favorites.

Can we make this a follow-up?

> The background reload spinners shift the whole list, and in this case even twice as there are 2 (1 for each list).

It's actually a foreground spinner. When the window focus is lost and acquired again, we trigger a sync to make sure any messages from background notifications are there. This is a technical requirement, there is currently no way around.

IMO we should show the spinner here. Otherwise the list will update as you attempt to perform an action, e.g. a message is deleted, the list shrinks and you click on another message.

We should enhance the spinner to not move the list instead of hiding the spinner.

jancborchardt:
> > If you don’t have any favorites, does the section hide? (It should – I didn’t test this as I didn’t want to unfavorite all my favorites.
> 
> Can we make this a follow-up?

Yep, for sure. :) Just make sure that any outstanding design review points are opened as an issue so the review does not get lost.


> > The background reload spinners shift the whole list, and in this case even twice as there are 2 (1 for each list).
> 
> It's actually a foreground spinner. When the window focus is lost and acquired again, we trigger a sync to make sure any messages from background notifications are there. This is a technical requirement, there is currently no way around.
> 
> IMO we should show the spinner here. Otherwise the list will update as you attempt to perform an action, e.g. a message is deleted, the list shrinks and you click on another message.
> 
> We should enhance the spinner to not move the list instead of hiding the spinner.

Sure. Let’s just half-overlay the spinner above the first message. This is just like Android does it as well. Of course the first message should still remain clickable – only the small part of the spinner should overlay the element.

ChristophWurst:
> Sure. Let’s just half-overlay the spinner above the first message. This is just like Android does it as well. Of course the first message should still remain clickable – only the small part of the spinner should overlay the element.

:+1: can be done independent of this PR already. Who'd like to work on this? :)

GretaD:
> > If you don’t have any favorites, does the section hide? (It should – I didn’t test this as I didn’t want to unfavorite all my favorites.
> 
> Can we make this a follow-up?

when you dont have any favorites, it looks like this:

favorites

ChristophWurst:
- [ ] BUG: vue duplicate key errors due to duplicates in fav section
- [ ] BUG: background syncs shows full inbox as "new"

ChristophWurst:
>     * BUG: background syncs shows full inbox as "new"

Can't reproduce right now. Let's integrate and fix later when we know what causes this

Issue #2649, opened 2020-02-13
Move "message list + sync" logic from folder details to a new component


ChristophWurst:
(reused for simple priority inbox and the final one -> each section has self-contained logic)

Issue #2154, opened 2019-10-23
Delete key shortcut behaves differently than action menu "Delete" action


jancborchardt:
1. Open the very first mail in a mailbox
2. Hit the delete key on your keyboard
3. Current message gets deleted as expected, but:

### Expected behavior
It should switch to the next most recent message below as the active one. (Only when you are at the very top of course. When you are in any other message below, it will move to the next newer message, as intended.)

This works fine when you use "Delete" from the action menu.

### Actual behavior
When you use the delete key shortcut, the already deleted mail keeps being active and shown in the content view.

### Mail app

**Mail app version:** current master

jancborchardt:
The relevant code which gets run on using the delete keyboard shortcut is in this block:
https://github.com/nextcloud/mail/blob/master/src/components/EnvelopeList.vue#L207-L213

It seems what is missing there is the action to go to the next newest unread mail. :)

ChristophWurst:
To fix this discrepancy we should unify the code and use the same logic for both the shortcut and the action menu.

PR #2665, opened 2020-02-18
Make the envelope list independent

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2649
Fixes https://github.com/nextcloud/mail/issues/2154

* `EnvelopeList` is now indeed a simple list of envelopes
* `Mailbox` wraps ^ and is responsible for data fetching, loading and error state
* Envelopes are not saved to the Vuex store anymore. I figured this is not necessary and the code is just a lot simpler when we keep the data locally. Also this fixes some edge cases from previous master, e.g. you can now open a message from an URL where the envelope is not shown in the list.

In general everything should work like before.

Open todo
* Make sync work for each `Mailbox` independently, but also globally for inboxes


GretaD:
1. Back button background is shown(a dirty hack we did some time ago to cover the hamburger button)
backbutton


2. On smaller screens, you click the message and nothing happens. Like the pic above.
3. On smaller screens when you click the back button, a white page is open and only hamburger menu is shown. You click the menu and the first column is shown, you type i.e inbox but nothing happens. You make the screen bigger, the messages are shown, but you cannot click them. `TypeError: "this.folder.envelopes is undefined"`
4. On big normal screen, favorite/draft/sent doesnt work, it loads infinitely .
`[ERROR] mail: could not load message` and `UID undefined is invalid`

ChristophWurst:
Closing in favor of #2669

PR #2666, opened 2020-02-18
Add third line of message preview

### LINKED ISSUE #2579: Show first line of message as preview
### Feature Request

As any other decent IMAP client would do, our should also show the start of the first line of each message in the message list.

#### Summary

See above

---

Caveat: this operation is expensive. We have to fetch the body part. HTML message need special care.

In a later stage we can optimize to omit stuff like greetings or quoted text for bottom replies. For a start, the first line is good enough.

## Implementation idea

We should fetch this data on demand, similar to the avatars. The data should then be written to the message cache #2064 (or memcache? :thinking:) as it is immutable.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/87801779-show-first-line-of-message-as-preview?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GretaD:
fix  #2579

**PC**
3rdline3
_____
**Phone**
3rdline

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -15,15 +15,19 @@
 		</div>
 		<div class="app-content-list-item-line-one" :title="addresses">
 			{{ addresses }}
-		</div>
 		<div class="app-content-list-item-line-two" :title="data.subject">
+			{{ data.subject }}
+		</div>
+			<!-- work in progress -->
+		<div class="app-content-list-item-line-three">
+			<label> {{t('mail', 'First line of message should be shown here')}}</label>
+		</div>
+		</div>
 			<span v-if="data.flags.answered" class="icon-reply" />
 			<span v-if="data.flags.hasAttachments === true" class="icon-public icon-attachment" />
 			<span v-if="draft" class="draft">
 				<em>{{ t('mail', 'Draft: ') }}</em>
 			</span>
-			{{ data.subject }}
-		</div>
 		<div class="app-content-list-item-details date">
 			<Moment :timestamp="data.dateInt" />
 		</div>
@@ -177,9 +181,15 @@ export default {
 
 // Fix layout of messages in list until we move to component
 
-.app-content-list-item-line-two,
+.app-content-list-item-line-two {
+	margin-left: -50px;
+	margin-top: -8px;
+}
 .app-content-list-item-menu {
 	margin-top: -8px;
+}
+.app-content-list-item-line-three {
+	opacity: 0.3;
 }
 
 .app-content-list-item-menu {
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -15,19 +15,19 @@
 		</div>
 		<div class="app-content-list-item-line-one" :title="addresses">
 			{{ addresses }}
-		<div class="app-content-list-item-line-two" :title="data.subject">
-			{{ data.subject }}
-		</div>
+			<div class="app-content-list-item-line-two" :title="data.subject">
+				{{ data.subject }}
+			</div>
 			<!-- work in progress -->
-		<div class="app-content-list-item-line-three">
-			<label> {{t('mail', 'First line of message should be shown here')}}</label>
+			<div class="app-content-list-item-line-three">
+				<label> {{ t('mail', 'First line of message should be shown here') }}</label>
+			</div>
 		</div>
-		</div>
-			<span v-if="data.flags.answered" class="icon-reply" />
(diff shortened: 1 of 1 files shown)


ChristophWurst:
cc @nextcloud/designers  :)

ma12-co:
In the example you bring up in the issue @ChristophWurst there are different elements in the component, but none of them has everything like we want to havef xD.
By everything I mean: name, timestamp, avatar, actions, subject, body preview

So just a couple of quick ideas: 
* If we want to have everything in the component I guess the only solution is make the component bigger (mockup 1)
* Otherwise another solution that would also declutter a bit could be moving the actions to a topbar like in files and use checkmarks to select envelopes to whom apply the actions, this would also allow bulk actions!
https://www.figma.com/file/pjJzvAeD8imKUnNKVtFBF2/Untitled?node-id=0%3A1

raimund-schluessler:
Mockup #2 looks good to me, especially since it allows bulk actions. 👍 But since this requires far more work than this PR, I guess it is something for a different PR.

skjnldsv:
Thanks for diving into this Greta :)
It still feels too dense to me. 

What about putting the time in the top right corner and the menu action in the bottom right?
Having both inline reduce your available space quite a lot :)

GretaD:
Big screen/pc
thirdlinePC


small screen/phone
thirdline

skjnldsv:
Much better :)
Maybe the second line could be closer to the third instead of closer to the first? What do you all think?

GretaD:
> Much better :)
> Maybe the second line could be closer to the third instead of closer to the first? What do you all think?

Yes sure, its not perfectly aligned, i will do that once we get the real data for the third line. Thank you for the feedback <3

ma12-co:
It looks better but imo it's still too packed, too little negative space (even if the second line is moved down). Honestly I really see no problems in having a slightly taller component if the info is worth displaying :)

jancborchardt:
Yup, just add more padding between name and subject, taller component is fine. :)
@GretaD the general layout is great otherwise, only details:
- [ ] The 3-dot menu icon needs to be vertically aligned with the preview line text. Currently it is a bit too low down
- [ ] The preview line text is just cut off, should be ellipsized

@ma12-co regarding the mockup/proposal: Keep in mind that multiselect is not a standard case, so optimizing for that by moving the actions out of every single mail makes the regular usage more difficult → we shouldn’t do that. It’s a separate thing, and a solution is proposed in the relevant issue at https://github.com/nextcloud/mail/issues/890#issuecomment-591289668

ChristophWurst:
(just dropping that we still need to work on the back-end to actually *extract* the text for the third line. I don't even know if we can for HTML messages :grin: )

ma12-co:
>  Keep in mind that multiselect is not a standard case

hmm.. To me performing actions on multiple items in a mailbox is a pretty standard thing to do these days, people is used to it from gmail, and also sometimes the 'select all' is a life saver :D

> so optimizing for that by moving the actions out of every single mail makes the regular usage more difficult

Actions for each single items are already available in the 3 dot menu near the reply button in each email, If they'd be explicit buttons instead of being grouped in a 3 dot menu, single and bulk actions would be really at hand at all times, without having to go through the step of initialising a multiselect operation as you describe in #890 or opening the 3 dot menu for single e-mail actions.

Screenshot_20200226_155810

GretaD:
Thank you @jancborchardt and @ma12-co for the feedback, will fix all that after we have the real data for the third line :)

jancborchardt:
@ma12-co nevertheless a change like this is completely separate from the topic of this pull request, which is simply to introduce a 3rd line. :) The action icon is not in the way as clear from Greta’s mockup, and if anything the Select action would need the same amount of space.

clementhk:
> (just dropping that we still need to work on the back-end to actually _extract_ the text for the third line. I don't even know if we can for HTML messages 😁 )

How does it look with a simple `strip_tags()`?

ChristophWurst:
> How does it look with a simple `strip_tags()`?
> Or `html2text/html2text` package?
> What's the up and down of them?

huh, didn't think about those. This makes sense. I'll check that out!

ChristophWurst:
Tried to rebase but it's bit tricky. For a small change like this it will be easier to re-do it from scratch :)

Issue #2434, opened 2020-01-04
Back (to emails list) button on mobile refreshes the list


Rychu-Pawel:
### Expected behavior
Pressing the back button on an email view takes user back to the already loaded list of emails without refreshing it.

### Actual behavior
The Back button triggers emails list refresh. It worsen user experience as the list is refreshed all over again when reading emails. UX would be much smoother without this feature. 

### Mail app

**Mail app version:** 0.17.0

**Mailserver or service:** imap

### Server configuration

**Operating system**: Ubuntu 18.04

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.2.24

**Nextcloud Version:** 16.0.7.1

#### Client configuration
**Browser:** Firefox

**Operating system:** Windows

PR #2669, opened 2020-02-20
Make envelope lists more versatile

### LINKED ISSUE #2649: Move "message list + sync" logic from folder details to a new component

ChristophWurst:
This is a remake of https://github.com/nextcloud/mail/pull/2665.

Fixes #2649
Fixes https://github.com/nextcloud/mail/issues/2434

Before we had two types of envelope lists: ordinary and search. This was a bit limiting, e.g. search lists were disposed of quite often and having two types of searches at the same time was impossible.

This reorganizes the components a bit. And adds a `Mailbox` abstraction that should be the reusable building block for future UI features (#2648). The store got a little overhaul. And some unit tests for the most important actions/mutations.

It's generally not as radical as my previous approach in #2665, but that failed hard with keeping data in sync across mailboxes, different views and background operations.


GretaD:
(small screen)
1. The back button with white backgroud is still there
backbuton
2. When you click back button a white page will be open(the menu should be open instead) 
3. When you click inbox after you've click back button, no email is shown. You need to click "all inboxes" for the email to be shown
4. After the email are shown(point 3) and you click in one of them, the email doesnt open to read.

PR #2688, opened 2020-02-25
Make the initial mailbox sync interruptable and resumable

### LINKED ISSUE #2578: Resume partial message cache sync
### Expected behavior
Message cache should be fault tolerant and able to resume (or just redo) a failed partial sync.

### Actual behavior
An aborted partial sync can leave behind *some* new inserted messages, but an old sync token. Thus the next partial sync attempt will try to insert the same messages and run into duplicate key constraint errors.

Updates and deletes do not have this problem.

### Mail app

**Mail app version:** 

One after #2064 has been merged.

Originally discovered at https://github.com/nextcloud/mail/pull/2064#issuecomment-543087394

---

A possible solution could be to just delete all UIDs of new message before we insert. As we have locking this should also be relatively save with any other ongoing caching operations.

ChristophWurst:
Fixes #2578 

With large mailboxes the initial sync is slow. Therefore it's prone to run into timeouts. I was hoping cron to be my last resort, but it turns out we can't rely on it as the instance might not use system cron but web cron. Then the same timeout rules apply as for any other AJAX requests. Thus we need an alternative.

I've changed the sync logic to limit the cache population to x (currently 1000 but I'll increase) message per run. Then it just stops without setting the syn tokens. The next sync attempt will detect the missing cache and start over. However, it will also check for the highest known UID in the database, and only fetch the ones that are newer than this message. It relies on the fact that the previous inital sync attempt did not leave any holes. But that shouldn't happen usually. Messages are also inserted in ascending order and both messages and recipients are inserted in the same transaction. Thus this should be a save way to resume the operation with the previous progress. Eventually one of the attempts will finish and no more messages can be fetched. Then the tokens will be set. From then on, partial sync can be used.

This also fixes the /sync request to be a POST. It was okay-ish to do it as GET for the old Horde sync as it was indeed stateless. But now it's changing server state, thus POST seems more appropriate. Moreover we send the list of known UIDs as parameter and AFAIK there are limits in URL length while with POST body we can send more data.

Steps to test

* Close the app
* Run `update oc_mail_mailboxes set sync_new_token=null, sync_new_lock=null, sync_changed_token=null, sync_changed_lock=null, sync_vanished_token=null, sync_vanished_lock=null;`, `delete from oc_mail_messages;` and `delete from oc_mail_recipients;`
* Open the app and browse to a large inbox

You should then see the "indexing … might take longer … " message. and in the browser network tab you see /sync requests. lots of 202s and then finally a 200. then the mailbox content should show.

Todo

- [x] A bit more testing
- [x] Find sweet spot of messages per request to optimize performance and still avoid timeouts
- [x] Ask @nickvergessen for a test run on his large mailbox(es)

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -62,6 +62,23 @@ class MessageMapper extends QBMapper {
 		return $uids;
 	}
 
+	public function findHighestUid(Mailbox $mailbox): ?int {
+		$query = $this->db->getQueryBuilder();
+
+		$query->select($query->createFunction('MAX(' . $query->getColumnName('uid') .  ')'))
+			->from($this->getTableName())
+			->where($query->expr()->eq('mailbox_id', $query->createNamedParameter($mailbox->getId())));
+
+		$result = $query->execute();
+		$max = (int) $result->fetchColumn(0);
+		$result->closeCursor();
+
+		if ($max === 0) {
+			return null;
+		}
+		return $max;
+	}
+
 	public function insertBulk(Message ...$messages): void {
 		$this->db->beginTransaction();
 
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -56,7 +56,7 @@ return [
 		[
 			'name' => 'folders#sync',
 			'url' => '/api/accounts/{accountId}/folders/{folderId}/sync',
-			'verb' => 'GET'
+			'verb' => 'POST'
 		],
 		[
 			'name' => 'folders#markAllAsRead',
--- a/lib/Controller/FoldersController.php
+++ b/lib/Controller/FoldersController.php
@@ -125,6 +125,10 @@ class FoldersController extends Controller {
 				}, $uids),
 				!$init
 			);
+
+			if ($syncResponse->isIncomplete()) {
+				return \OCA\Mail\Http\JsonResponse::fail([], Http::STATUS_ACCEPTED);
+			}
 		} catch (MailboxNotCachedException $e) {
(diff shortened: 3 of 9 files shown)


ChristophWurst:
[mail.tar.gz](https://github.com/nextcloud/mail/files/4254639/mail.tar.gz)

lib/Db/MessageMapper.php
```diff
@@ -62,6 +62,23 @@ public function findAllUids(Mailbox $mailbox): array {
 		return $uids;
 	}
 
+	public function findHighestUid(Mailbox $mailbox): ?int {
+		$query = $this->db->getQueryBuilder();
+
+		$query->select($query->createFunction('MAX(' . $query->getColumnName('uid') .  ')'))
```
nickvergessen: max is available in 18+,
maybe leave yourself a TODO?

ChristophWurst: Yes I used the function until I saw the failure on stable17 :)

nickvergessen:
Didn't review the code really, but I ran it on my 80k folders and they seem to work fine

ahstro:
>Didn't review the code really, but I ran it on my 80k folders and they seem to work fine

Bold move, my friend 🙇

nickvergessen:
> Bold move, my friend 🙇

Since I don't depend on Nextcloud Mail it's not as bold as it sounds, and someone has to try it (we ran into timeouts before that is why this follow up is now needed

ChristophWurst:
Testing in production > unit testing :grinning:

PR #2697, opened 2020-02-26
Fix "post-message deletion"-navigation

### LINKED ISSUE #2154: Delete key shortcut behaves differently than action menu "Delete" action

### LINKED ISSUE #2492: Add 'cancel' button to discard draft

ahstro:
This PR contains two adjacent but ultimately disconnected bugs. Sorry! If y'all prefer harder separation in the future let me know.

This fixes 1) navigating to the wrong message (one step too far down in the list) after deleting a message, and 2) not navigating at all.

I also threw in some minor clean-up for good measure/bad GitHub etiquette. At least the commits are separate 😇

Edit: I'm not sure if there's an issue related to this. I stumbled upon the bugs when trying to understand enough to fix #2492 (getting there). I tried finding a related issue, but couldn't..

Edit 2:
Fixes #2154

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -317,7 +317,6 @@ export default {
 			}
 		},
 		onDelete({envelope}) {
-			const envelopes = this.envelopes
 			const idx = findIndex(propEq('uid', envelope.uid), this.envelopes)
 			if (idx === -1) {
 				logger.debug('envelope to delete does not exist in envelope list')
@@ -329,13 +328,7 @@ export default {
 				return
 			}
 
-			let next
-			if (idx === 0) {
-				next = envelopes[idx + 1]
-			} else {
-				next = envelopes[idx - 1]
-			}
-
+			const next = this.envelopes[idx === 0 ? idx : idx - 1]
 			if (!next) {
 				logger.debug('no next/previous envelope, not navigating')
 				return
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -131,10 +131,7 @@ export default {
 		onToggleSeen() {
 			this.$store.dispatch('toggleEnvelopeSeen', this.data)
 		},
-		onDelete(e) {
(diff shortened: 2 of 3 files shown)


jancborchardt:
Hi @ahstro, good stuff! :) Is it correct that this aims to fix **Delete key shortcut behaves differently than action menu "Delete" action #2154**? Or what is the expected before/after behavior?

src/components/Mailbox.vue
```diff
@@ -329,13 +328,7 @@ export default {
...
 			}
 
-			let next
-			if (idx === 0) {
-				next = envelopes[idx + 1]
-			} else {
-				next = envelopes[idx - 1]
-			}
-
+			const next = this.envelopes[idx === 0 ? idx : idx - 1]
```
ChristophWurst: should it be `idx === 0 ? idx + 1 : idx - 1` here?

ChristophWurst: or just `1` instead of `idx + 1`

ahstro: No, that was one of the bugs I found 😊 Since we removed the 0th item of the array, the new 0th item would be the desired message, while the 1st would be one down the list. Not sure if that sentence made sense to anyone other than me 😅

ahstro:
@jancborchardt Ah, there ya' go, that's a related issue! That's one of the two bugs this fixes 😊 The other is the one I'm crudely trying to describe in @ChristophWurst's review. Sorry if my communication skills are off...

PR #2700, opened 2020-02-27
Remove factory feature of getMessageData (or enlighten me)

ahstro:
So, I really had a question about the code, but I thought I'd structure it in the form of a PR. Is there a reason for `getMessageData` to have that extra function-layer? To me, this seems like it would work, but I'm used to functional programming, so I'm guessing it might have something to do with how JavaScript handles mutability? Sorry if this is not the right forum for this.. 🙇

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -338,21 +338,19 @@ export default {
 				return `"${recipient.label}" <${recipient.email}>`
 			}
 		},
-		getMessageData() {
-			return uid => {
-				return {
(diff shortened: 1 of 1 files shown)


ahstro:
Ooh, that sounds familiar from back when I was battling React in the past. Just tested it, and seems like Vue is making that bind magic happen behind the scenes. Vue: 1️⃣ React: 0️⃣

ChristophWurst:
Okay, all good then :)

PR #2701, opened 2020-02-27
[WIP] Add action buttons for saving and discarding draft

### LINKED ISSUE #2492: Add 'cancel' button to discard draft

ahstro:
Working on #2492, but done for the day, so thought I'd open a PR with my progress and ask some questions at the same time 😊 Also, if there are any concerns, shout 'em out!

#### 🧭 Where to navigate?
Currently L402 and L425 navigate to `{name: 'home'}`, which I was hoping would be the same as just loading the app and letting some already written code put me at an appropriate place, but turns out just leaves me with a blank screen. Before I start figuring out how to get somewhere, I should probably know where I want to go?

#### 🚮 How to delete a draft?
This seems to be a different state from deleting a draft from the drafts-folder. On L409-420 I have access to the message data, but would need to have access to the `envelope` to delete it the old way. I'm not 100% yet, but I'm thinking this is something that is created and stored on the server? Haven't gotten enough mental overview of the code to know how to get a hold of it here.

Fixes #2492

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -154,6 +154,9 @@
 					<ActionButton icon="icon-download" @click="onSaveAndClose">
 						{{ t('mail', 'Save and close draft') }}
 					</ActionButton>
+					<ActionButton icon="icon-delete" @click.prevent="onDelete">
+						{{ t('mail', 'Discard draft') }}
+					</ActionButton>
 				</Actions>
 				<div>
 					<input
@@ -172,6 +175,11 @@
 	<Loading v-else-if="state === STATES.SAVING" :hint="t('mail', 'Saving …')" />
 	<div v-else-if="state === STATES.SAVED" class="emptycontent">
 		<h2>{{ t('mail', 'Saved!') }}</h2>
+		<p>{{ t('mail', 'Redirecting …') }}</p>
+	</div>
+	<Loading v-else-if="state === STATES.DISCARDING" :hint="t('mail', 'Discarding …')" />
+	<div v-else-if="state === STATES.DISCARDED" class="emptycontent">
+		<h2>{{ t('mail', 'Draft was discarded!') }}</h2>
 		<p>{{ t('mail', 'Redirecting …') }}</p>
 	</div>
 	<div v-else-if="state === STATES.ERROR" class="emptycontent">
@@ -220,6 +228,8 @@ const STATES = Object.seal({
 	FINISHED: 4,
 	SAVING: 5,
 	SAVED: 6,
+	DISCARDING: 7,
+	DISCARDED: 8,
 })
 
 export default {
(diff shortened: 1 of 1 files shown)


src/components/Composer.vue
```diff
@@ -372,6 +393,37 @@ export default {
...
+				.then(() => wait(1000)) // Time to register the "success"-message
+				.then(() => this.$router.push({name: 'home'})) // TODO: Navigate to better place
+		},
+		onDelete() {
+			this.state = STATES.DISCARDING
+			this.draftsPromise
+				.then(this.getMessageData)
+				.then(message => {
+					// TODO: Somehow delete the draft
+					// We've got `message`, but we need an `envelope`. I'm not 100%
```
ChristophWurst: Essentially both represents a message. And in the old code base before the Vue rewrite we used just *message* for both objects that had the *from*, *to*, *cc*, *bcc*, *subject*, *flags*,... data, and the objects that contained the body data and attachments. Thus I renamed the former to gain some clarity in code.

I think we can just refactor the action to take *account id*, *folder id* and *message id* rather than an object. Then it will work with both envelopes and messages. Ironically, `deleteMessage` currently takes an `envelope` :see_no_evil:

src/components/Composer.vue
```diff
@@ -372,6 +393,37 @@ export default {
...
+					// on the distinction yet, but `envelope` seems to have some
+					// necessary parameters which we don't have access to here.
+					// const envelope = magic(message)
+
+					// I'm not sure if this is needed. It only seems to affect the
+					// Mailbox.vue, which I doooon't think is relevant from within
+					// a draft? Maaybe?
+					// this.$emit('delete', {envelope})
```
ChristophWurst: It's needed when you *resume a draft*, as in open an existing draft in your *Drafts* mailbox. Then you'll see the message composer and work on something that is actually displayed in the second column. In that case it makes sense to emit the `delete` event, so that the message is removed from the list.

ChristophWurst:
> Currently L402 and L425 navigate to `{name: 'home'}`, which I was hoping would be the same as just loading the app and letting some already written code put me at an appropriate place, but turns out just leaves me with a blank screen. Before I start figuring out how to get somewhere, I should probably know where I want to go?

This navigation logic dates back to when I did my first attempts with Vue. So this might be an anti pattern. But we use the `created` lifecycle method to magically show the correct content (account/folder). So when you are already showing a mailbox and the composer and navigate "back" to `home` nothing will happen as Vue won't create a new instance of `Home.vue` and thus this routing logic won't be triggered.

Again, we might use Vue unconventionally here. I'll have to do a few tests to see if we can improve this. We could possibly `watch` some route parameters and re-trigger the logic.

src/components/Composer.vue
ahstro: Ah, thank you for the clarification! 🙌 I'll look into a refactor and block this PR with another one. One potential problem I see is that I have neither `folderId` nor `messageId` at this stage, just the `draftUID` and `accountId`. I'll figure out if that's a problem when I start working on it 😅

>  Ironically, deleteMessage currently takes an envelope 🙈

Yeah, that tripped me up at first 😂

ahstro:
> we use the created lifecycle method to magically show the correct content (account/folder). So when you are already showing a mailbox and the composer and navigate "back" to home nothing will happen as Vue won't create a new instance of Home.vue and thus this routing logic won't be triggered.

Yeah, I was poking around there, thinking of maybe moving it to the `mounted`-stage instead, but didn't feel comfortable enough with Vue's lifecycle to start playing around with it. I'm not sure what `watch` is, but it sounds promising!

ChristophWurst:
I'm closing this for now as it hadn't gotten any update in almost half a year. We will still work on this, but just not now. So when that is the case we can just re-apply the changes to the then current master, as far as possible.

Cheers :v:

Issue #2508, opened 2020-01-19
Cannot add recipients


republicus:
This bug makes Nextcloud Mail nearly unusable from within Nextcloud.
I noticed the latest comment at the App Store is the same complaint.

### Expected behavior
Entering recipients' address into **To:** field adds recipients' email.

### Actual behavior
**Intermittently** (frequently) it becomes impossible to add recipients.
1. Typing the address may return no results or blank results -- leaving no option to add recipient.
2. Address not in contacts or previously used _may_ not be added.

![image](https://user-images.githubusercontent.com/47370709/72673357-501e5980-3a37-11ea-9a42-0b611dfb3f14.png)

It should be possible to enter text without engaging the mouse to add a recipient.
Currently even when it does behave - I must still click the **+**  button to add a recipient; or removing the **check** icon to remove a recipient.

Behavior like adding tags may be ideal where pressing Comma, Semicolon, Tab, and/or Return should add the address.

The current behavior is quite intermittent and annoying. I would even prefer a free form where email is consistently able to be sent.

### Mail app

**Mail app version:** 1.0.0
**Mailserver or service:** Postfix/Dovecot

### Server configuration

**Operating system**: Debian 10
**Web server:** Apache
**Database:** MariaDB
**PHP version:** PHP 7.3.11-1
**Nextcloud Version:** 17.0.2

#### Client configuration
**Browser:** Chrome 74.0.3729.169
**Operating system:** Windows 10

DecaTec:
Same here: When trying to set the recipient's mail address, it often let's you type some letters, but then the whole auto completion of the mail address seems to be broken.

Some more information from the browser's developer console:

```
TypeError: "t.customLabel(...) is null"
    optionKeys https://mydomain.de/apps/mail/js/mail.7.2cef98fdf1f1cf7cca86.js:391
    optionKeys https://mydomain.de/apps/mail/js/mail.7.2cef98fdf1f1cf7cca86.js:391
    get https://mydomain.de/apps/mail/js/mail.js?v=3e3e5628-9:7
    evaluate https://mydomain.de/apps/mail/js/mail.js?v=3e3e5628-9:7
    _n https://mydomain.de/apps/mail/js/mail.js?v=3e3e5628-9:7
    isExistingOption https://mydomain.de/apps/mail/js/mail.7.2cef98fdf1f1cf7cca86.js:391
[...]

Issue #1973, opened 2019-08-31
Contact selection behave weirdly


devcrafting:
### Expected behavior
I type an email address, it should take it, but I can't.

### Actual behavior
When I start a new message, I click in "To", I have a "list is empty". I start typing email address, I see a + sign with what I am typing. If I slow down just a bit, it stops keeping new characters (eg I typed "clement@xxx.com", but it keeps only "+ cle" (an invalid address). From this point, I cannot add any address (even with copy/paste)...also the suggestion list always display contacts and an empty entry...

I got several errors in the browser console both in Firefox and Chrome :

```
background sync failed Error: "Request failed with status code 500"
    exports https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:87
    exports https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:94
    onreadystatechange https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:87
mail.js:94:13808
```

and only in Firefox

```
TypeError: "t.customLabel(...) is null"
    optionKeys https://domain.com/custom_apps/mail/js/mail.2.a6edd5281a08a69e52fe.js:470
    optionKeys https://domain.com/custom_apps/mail/js/mail.2.a6edd5281a08a69e52fe.js:470
    get https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    evaluate https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    bn https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    isExistingOption https://domain.com/custom_apps/mail/js/mail.2.a6edd5281a08a69e52fe.js:470
    filteredOptions https://domain.com/custom_apps/mail/js/mail.2.a6edd5281a08a69e52fe.js:470
    get https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    evaluate https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    bn https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    render https://domain.com/custom_apps/mail/js/mail.2.a6edd5281a08a69e52fe.js:470
    _render https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
    r https://domain.com/custom_apps/mail/js/mail.js?v=7efbea2b-0:15
[...]

PR #2703, opened 2020-02-27
Label recipients correctly in Composer.vue

ahstro:
I was getting some juicy errors when trying to add recipients, so I dug into it. Turns out `vue-multiselect` wasn't a huge fan of their default `customLabel` returning `null` with our data, so I wrote our own. It's very recipient specific - assuming there will be a `.email` - but also somewhat lenient - falling back on hoping `option` is a string (which maybe could blow up if it's a truthy non-string); worst of both worlds? Anyway, here's more code to review 🤷

#### Juicy errors 🧃
errors

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -290,7 +290,9 @@ export default {
 			return this.$store.getters.accounts.filter(a => !a.isUnified)
 		},
 		selectableRecipients() {
-			return this.newRecipients.concat(this.autocompleteRecipients)
+			return this.newRecipients
+				.concat(this.autocompleteRecipients)
+				.map(recipient => ({...recipient, label: recipient.label || recipient.email}))
 		},
 		isReply() {
 			return this.to.length > 0


ChristophWurst:
When does the error actually occur? We possibly should also fix the code that leads to this condition, though I'm fine hardening the label logic to be more tolerant :)

ahstro:
I'm not 100% (don't code at night, kids), but I\m pretty sure the offending line is [this one](https://github.com/shentao/vue-multiselect/blob/master/src/multiselectMixin.js#L378), where `vue-multiselect` tries to call `.toString()` on our `null`. Now, I'm not entirely sure whether they want to handle `null` or not, because [here](https://github.com/shentao/vue-multiselect/blob/master/src/multiselectMixin.js#L14) they kinda just pretend it's all good 😅 I'll open an issue to make them aware of it, so they can decide how they want to handle it.

ChristophWurst:
Okay, thanks for the insights :) I'm not so much worried about the handling in the vue-multiselect, but why this app passes `null` as value. Every recipient should have at least an email address. And optionally a label. If there is no label, we use the email as label here IIRC.

ahstro:
The `options` I saw passed to the multiselect was an array of objects like this:
```
{
  email: "foo@bar.org",
  id: 1, 
  label: null,
}
```
Their code essentially boiled down to `option.label` with no fallback if it was falsy or non-string. I don't know if we want to explicitly set the `label` to the value of `email` on the backend or something?

ChristophWurst:
We could `.map` over the entries in `selectableRecipients` of `Composer.vue` to fill in a value for `label` in case it's `null` :thinking:

ahstro:
That was how I first tested it, and yeah it works too. This felt more _intended_, but I really don't know 🤷 I love `.map`, so if you prefer that, I'm aλλ in

ChristophWurst:
> This felt more _intended_

True. Let's do it with `map` then :)

ahstro:
Ah, found a related issue!
Fixes #2508

ChristophWurst:
> Ah, found a related issue! Added the priority label because an issue had it. Hope I'm not overstepping anything bowing_man
> Fixes #2508, #1973

Wow, that would be amazing. Do you have insights on why this PR fixes those issues? Could you reproduce them before?

ahstro:
>Wow, that would be amazing. Do you have insights on why this PR fixes those issues? Could you reproduce them before?

The errors and related behavior I was experiencing seemed to be the exact same as those. Closing them, but if they keep happening, I'll eat my word and get back to work on a proper fix 😇

PR #2704, opened 2020-02-28
Add a public Talk channel for developers

ChristophWurst:
cc @ahstro :)


ChristophWurst:
Also @greta @StCyr please join us there for any questions :)

PR #2706, opened 2020-02-28
Remove the heavy-weight virtual flagged mailbox

ChristophWurst:
Currently we have this feature of a virtual `\flagged` inbox, where only flagged messages from an inbox are shown. This isn't a lightweight implementation, but actually stores a mailbox and all its contents in the database. Thus causes lots of duplication. Even worse, the flagged view of the inbox can get outdated when flags change, while the data is never cleaned up.

For the priority inbox I needed a way to display a filtered view of any mailbox because we also want to show a unified list of all flagged messages. The solution was to "simply" query the database for it, while operating on the real inbox.

So I figured that we should probably get rid of the old, heavyweight flagged mailbox and replace it with a lightweight filtered view.

This PR does

* [x] remove the old implementation and makes sure to force a mailbox cache update when the user visits the app for the first time after the update (developers please run `php occ migrations:execute mail 1030Date20200228105714`, I don't want to bump the version number for this)
* [x] adds a lightweight filtered view instead
* [x] Filter sync requests according to the `query` of the currently shown mailbox


lib/IMAP/FolderMapper.php
```diff
@@ -25,18 +25,27 @@
...
 use Horde_Imap_Client_Socket;
 use OCA\Mail\Account;
 use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\Folder;
-use OCA\Mail\SearchFolder;
 
 class FolderMapper {
 
+	/**
+	 * This is a temporary workaround for when the sieve folder is a subfolder of
```
ChristophWurst: "temporary"

ChristophWurst:
- [ ] BUG ``[vue-router] missing param for named route "message": Expected "filter" to match "[^\/]+?", but received ""``

ChristophWurst:
- [ ] BUG background sync for inboxes pulls in *all* messages if another mailbox is opened as it does not send the known UIDs

ChristophWurst:
One remaining bug that because visible with the fixed active state of navigation items from nc-vue: when in the starred inbox, the inbox shows as active as well. this is because we use the non-exact route matching and the starred inbox route is thus a descendant of the inbox route.

ChristophWurst:
- [ ] remove `\OCA\Mail\Db\Mailbox::getMailbox`

ChristophWurst:
@nextcloud/mail please review and test :)

jancborchardt:
> I have no idea how to test this change :-)

Same here – does anything change visually? :)

ChristophWurst:
> Same here – does anything change visually? :)

Nope. If you can't make out a difference as a user, I guess this can seen as a positive test result :)

ChristophWurst:
Just test if the favorites inbox works as before. And then somone please have a look at the code for any obvious flaws.

ChristophWurst:
self-reviewed the code (some time passed and I have bad memory, so it's like another person wrote the code), this is partly covered by tests and CI is happy. Let's merge and fix any bugs later on.

PR #2708, opened 2020-02-29
Refactor 'deleteMessage'-action to take just ids

ahstro:
Refactors the action `deleteMessage` to take `actionId`, `folderId`, and `id` instead of a full envelope. This paves the way for #2701 by allowing drafts to be deleted without access to the related envelope.

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -237,7 +237,11 @@ export default {
 				envelope: this.envelope,
 				message: this.message,
 			})
-			this.$store.dispatch('deleteMessage', this.message)
+			this.$store.dispatch('deleteMessage', {
+				accountId: this.message.accountId,
+				folderId: this.message.folderId,
+				id: this.message.id,
+			})
 		},
 	},
 }
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -502,30 +502,27 @@ export default {
 			newUid: uid,
 		})
 	},
-	deleteMessage({getters, commit}, envelope) {
-		const folder = getters.getFolder(envelope.accountId, envelope.folderId)
-		commit('removeEnvelope', {
-			accountId: envelope.accountId,
-			folderId: envelope.folderId,
-			id: envelope.id,
-		})
+	deleteMessage({getters, commit}, {accountId, folderId, id}) {
+		commit('removeEnvelope', {accountId, folderId, id})
 
-		return deleteMessage(envelope.accountId, envelope.folderId, envelope.id)
+		return deleteMessage(accountId, folderId, id)
 			.then(() => {
-				commit('removeMessage', {
-					accountId: envelope.accountId,
-					folder,
-					id: envelope.id,
-				})
+				const folder = getters.getFolder(accountId, folderId)
+				if (!folder) {
+					logger.error('could not find folder', {accountId, folderId})
+					return
+				}
+				commit('removeMessage', {accountId, folder, id})
 				console.log('message removed')
 			})
 			.catch(err => {
 				console.error('could not delete message', err)
-				commit('addEnvelope', {
-					accountId: envelope.accountId,
-					folderId: envelope.folderId,
-					envelope,
-				})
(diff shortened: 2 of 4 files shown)


src/components/Message.vue
```diff
@@ -237,7 +237,7 @@ export default {
 				envelope: this.envelope,
 				message: this.message,
 			})
-			this.$store.dispatch('deleteMessage', this.message)
+			this.$store.dispatch('deleteMessage', this.message.accountId, this.message.folderId, this.message.id)
```
ChristophWurst: I like it especially in this file. We currently fetch the `message` object *and* the `envelope` object. While we need the latter just for passing it to the actions. If more actions work with key-only, we can remove this overhead :rocket:

src/store/actions.js
```diff
@@ -502,30 +502,19 @@ export default {
...
 		})
 	},
-	deleteMessage({getters, commit}, envelope) {
-		const folder = getters.getFolder(envelope.accountId, envelope.folderId)
-		commit('removeEnvelope', {
-			accountId: envelope.accountId,
-			folderId: envelope.folderId,
-			id: envelope.id,
-		})
+	deleteMessage({getters, commit}, accountId, folderId, id) {
```
ChristophWurst: For consistency with the other actions I'd suggest to also use the object destruction like `{accountId, folderId, id}` for `deleteMessage`

src/store/actions.js
```diff
@@ -502,30 +502,19 @@ export default {
...
-		return deleteMessage(envelope.accountId, envelope.folderId, envelope.id)
+		return deleteMessage(accountId, folderId, id)
 			.then(() => {
-				commit('removeMessage', {
-					accountId: envelope.accountId,
-					folder,
-					id: envelope.id,
-				})
+				const folder = getters.getFolder(accountId, folderId)
+				commit('removeMessage', {accountId, folder, id})
```
ChristophWurst: Wouldn't hurt have a `undefined` check here. Right now we *should* always have the folder. But if we don't have it locally, we can just carry on as well.

src/store/actions.js
```diff
@@ -502,30 +502,19 @@ export default {
...
 			})
 			.catch(err => {
 				console.error('could not delete message', err)
-				commit('addEnvelope', {
-					accountId: envelope.accountId,
-					folderId: envelope.folderId,
-					envelope,
-				})
+				const envelope = getters.getEnvelope(accountId, folderId, id)
+				commit('addEnvelope', {accountId, folderId, envelope})
```
ChristophWurst: Similar here, add an `undefined` check and if we don't have the envelope locally we don't need to commit any change

src/store/actions.js
ahstro: Cool! 👍 In that case, do you have any preference on how to pass in the arguments?

#### Explicitly picking the relevant properties
```
this.$store.dispatch('deleteMessage', {
	accountId: this.message.accountId,
	folderId: this.message.folderId,
	id: this.message.id,
})
```
#### Sending in the whole object
```
this.$store.dispatch('deleteMessage', this.message)
```
#### Or using something like Ramda's `pick` to make a copy
```
this.$store.dispatch('deleteMessage', pick(['accountId', 'folderId', 'id'], this.message))
```

I think there's some maintainability and grokability value in the explicitness of the first and last options, but the middle option looks cleaner 😅

ChristophWurst: I like all of them :see_no_evil: 

The first one is more consistent to how we pass the objects to other actions right now, though.

src/store/actions.js
ahstro: Oh, right, I keep forgetting that things can just be undefined in JS 😂 On it!

src/store/actions.js
ahstro: Went with the first one before I saw this message, so guess that's the one I'm going with for now 😅

src/store/actions.js
ahstro: I'm not 100% this is what you meant? Also, added some logging which.. double check please 😇

PR #2709, opened 2020-02-29
Fix normalization imports

ahstro:
Got a bit too eager with #2707 and missed the _normalize_ vs _normalize**d**_ distinction. I went with _normalize**d**_, because two vs. one, but if y'all prefer the other way around, I'll change quickly 😊

Also, putting the priority label on this, because as it stands now, `master` is broken. My bad, y'all 🙇


ChristophWurst:
Whoopsie :)

PR #2796, opened 2020-03-26
Make it possible to mark messages important

### LINKED ISSUE #2963: mail_messages.mailbox_id should be an INT
### Expected behavior
Be able to join mail_messages and mail_mailboxes on postgres.

### Actual behavior
```
SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: integer = character varying

LINE 1: ... INNER JOIN "oc_mail_mailboxes" "mb" ON "mb"."id" = "m"."mai...

                                                             ^

HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

### Mail app

**Mail app version:**

Anything after #2064

ChristophWurst:
For https://github.com/nextcloud/mail/projects/7
Fixes #2963

- [x] Requires https://github.com/nextcloud/mail/pull/2800

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -62,7 +62,7 @@ module.exports = {
 				test: /\.(svg)$/i,
 				use: [
 					{
-						loader: 'url-loader'
+						loader: 'svg-inline-loader'
 					}
 				],
 				exclude: path.join(__dirname, 'node_modules', '@ckeditor')
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -12,7 +12,7 @@
 - **🙈 We’re not reinventing the wheel!** Based on the great [Horde](http://horde.org) libraries.
 - **📬 Want to host your own mail server?** We don’t have to reimplement this as you could set up [Mail-in-a-Box](https://mailinabox.email)!
 	]]></description>
-	<version>1.3.3</version>
+	<version>1.4.0</version>
 	<licence>agpl</licence>
 	<author>Christoph Wurst</author>
 	<author>Roeland Jago Douma</author>
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -331,6 +331,9 @@
 .icon-drafts {
 	@include icon-color('drafts', 'mail', $color-black);
 }
+.icon-important {
+	@include icon-color('important', 'mail', $color-black);
+}
 .icon-sent {
 	@include icon-color('sent', 'mail', $color-black);
 }
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -331,9 +331,6 @@
 .icon-drafts {
 	@include icon-color('drafts', 'mail', $color-black);
 }
-.icon-important {
-	@include icon-color('important', 'mail', $color-black);
-}
 .icon-sent {
 	@include icon-color('sent', 'mail', $color-black);
 }
--- a/img/important.svg
+++ b/img/important.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 0 24 24" width="24"><path d="M0 0h24v24H0V0z" fill="none"/><path d="M3.5 18.99l11 .01c.67 0 1.27-.33 1.63-.84L20.5 12l-4.37-6.16c-.36-.51-.96-.84-1.63-.84l-11 .01L8.34 12 3.5 18.99z"/></svg>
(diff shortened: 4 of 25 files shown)


jancborchardt:
As icon, you can use the same icon that Gmail uses: https://material.io/resources/icons/?search=label_important&icon=label_important&style=baseline

If a message is important, we can overlay it on the avatar just like the Favorite star.

ChristophWurst:
I can't find that icon in https://www.npmjs.com/package/vue-material-design-icons. What's its name?

ChristophWurst:
Bildschirmfoto von 2020-03-26 19-48-23

For now I went with a subtle red background

jancborchardt:
> I can't find that icon in https://www.npmjs.com/package/vue-material-design-icons. What's its name?

It’s on the link / in the url: `label_important`. If that library doesn’t have it, just download it from the page and put it in the repository. :)

ChristophWurst:
Sure, I just hoped that we could include the icon as simple as https://github.com/nextcloud/mail/blob/13fef2ad267f259b3ec4d8fd10583c8009a7b2a8/src/components/itinerary/FlightReservation.vue#L10

ChristophWurst:
What do ya think

Bildschirmfoto von 2020-03-27 11-05-18

jancborchardt:
Looking good! Could use some more opacity, and ideally a slight white border around the icon to separate it from the avatar – same for the favorites icon.

But how do we do this so it works with the dark theme? Should I make it 2 icons each, white border for normal, dark border for dark theme?

ChristophWurst:
> But how do we do this so it works with the dark theme? Should I make it 2 icons each, white border for normal, dark border for dark theme?

How does it work with the other icons? Don't we do this automagically?

jancborchardt:
> > But how do we do this so it works with the dark theme? Should I make it 2 icons each, white border for normal, dark border for dark theme?
> 
> How does it work with the other icons? Don't we do this automagically?

We currently do not do this visual separation for any icons which overlay other things. Cases where it’s similar:
- Files: The favorites icon overlays as well, but has no visual separation. Would look nicer there too, so we could actually do this in the core favorites icon.
- Notifications: On new notifications, the red dot on the bell icon is part of the icon and the bell has a cutout – it’s manually made as different icons, as the background of the header also needs to shine through.
- Android: On Android we have a white border around the star icon, and need to adjust this now for dark mode.
- Text: The avatars overlap and have a 2px border which visually separate them. We can do this here as they are HTML elements and not images.

cc @skjnldsv @juliushaertl is it possible to add this border programmatically, and adjust it based on the theme – maybe like the shadow and the black/white version? it should be a hard 1 or 2px border of the background color, just like the avatars in the Text app.

skjnldsv:
> cc @skjnldsv @juliushaertl is it possible to add this border programmatically, and adjust it based on the theme – maybe like the shadow and the black/white version? it should be a hard 1 or 2px border of the background color, just like the avatars in the Text app.

you can fill the whole icon if you want, the border, don't think so.
Unless you use inline svg in the html code, then you can with css vars

GretaD:
Now it looks like this, after the styling fixup.

white_imp_star
IMPOTAT-FGLAG

Not related to this but should we maybe change also the star icon and add a border, or remove the border from the important one, just to have the consistency..

webpack.common.js
```diff
@@ -62,7 +62,7 @@ module.exports = {
 				test: /\.(svg)$/i,
 				use: [
 					{
-						loader: 'url-loader'
+						loader: 'svg-inline-loader'
```
GretaD, line 65: @ChristophWurst, 
@skjnldsv proposed to import svg as string but it's configured with url-loader in the webpack config, can we move on with svg-inline-loader? Because otherwise we have to find with another way how to import it

jancborchardt:
> Not related to this but should we maybe change also the star icon and add a border, or remove the border from the important one, just to have the consistency.. cc @jancborchardt

Yep, already talked to @skjnldsv that of course the star icon should have the same style with border. He mentioned the standardization of list items to be a blocker for that – correct me if that was wrong @skjnldsv. :)

skjnldsv:
> He mentioned the standardization of list items to be a blocker for that – correct me if that was wrong @skjnldsv. :)

Yep, this is coming from server, until we have a nice dedicated component in #nextcloud-vue, there is no proper way to do it (inline svg + css variables)
@ma12-co had some components for the talk app, could we make them available everywhere on vue components & allow us to add inline svg around the avatar (like they're doing with the video icon & star?

ma12-co:
>until we have a nice dedicated component in #nextcloud-vue, there is no proper way to do it (inline svg + css variables)

I hope I'll be able to give some love to this once the call view in talk is done

webpack.common.js
ChristophWurst: > url-loader works like file-loader, but can return a DataURL if the file is smaller than a byte limit.

so it already inlines in most cases. But fine by if we switch to the svg-inline-loader if that doesn't break anything else. IIRC the CKEditor needed some special loaders. Let's test that carefully so it doesn't break :)

ChristophWurst:
@nextcloud/mail please give this a test. The basic marking as important should *just work*. The automatic classification is very basic. Let's get this implementation in and see how it performs. Then we can improve it.

ChristophWurst:
- [ ] BLOCKED by #2963 

```
SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: integer = character varying

LINE 1: ... INNER JOIN "oc_mail_mailboxes" "mb" ON "mb"."id" = "m"."mai...

                                                             ^

HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

jancborchardt:
O lala, we’re getting closer! :) Congrats everyone :tada:

PR #2801, opened 2020-03-26
Fix message source view wrapping and selection

jancborchardt:
Before, the message source was horizontally overflowing, and it wasn’t possible to select anything.

Now, that’s all fixed and there’s a nice descriptive header. :)
Mail message source

Please check @nextcloud/mail


StCyr:
nice :-)

PR #2802, opened 2020-03-26
 Show all folders if only one would be collapsed 

### LINKED ISSUE #692: Show all folders if only one is behind the „Show all folders“-button [$15]

xMartin:
closes #692

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -49,11 +49,6 @@
 					/>
 					<NavigationFolder
 						v-if="!group.account.isUnified && item.specialRole === 'inbox'"
-						v-show="
-							!group.isCollapsible ||
-							!group.account.collapsed ||
-							SHOW_COLLAPSED.indexOf(item.specialRole) !== -1
-						"
 						:key="item.key + '-starred'"
 						:account="group.account"
 						:folder="item"
@@ -113,15 +108,11 @@ export default {
 					(folder) => SHOW_COLLAPSED.indexOf(folder.specialRole) === -1
 				)
 				const isCollapsible = nonSpecialRoleFolders.length > 1
-				const foldersToShow = folders.filter(
-					(folder) =>
-						!isCollapsible || !account.collapsed || SHOW_COLLAPSED.indexOf(folder.specialRole) !== -1
-				)
 
 				return {
 					id: account.id,
-					account: account,
-					folders: foldersToShow,
+					account,
+					folders,
 					isCollapsible,
 				}
 			})
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -38,14 +38,22 @@
 				/>
 				<template v-for="item in group.folders">
 					<NavigationFolder
-						v-show="!group.account.collapsed || SHOW_COLLAPSED.indexOf(item.specialRole) !== -1"
+						v-show="
+							!group.isCollapsible ||
+							!group.account.collapsed ||
+							SHOW_COLLAPSED.indexOf(item.specialRole) !== -1
+						"
 						:key="item.key"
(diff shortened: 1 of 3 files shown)


src/components/NavigationAccountExpandCollapse.vue
```diff
@@ -27,7 +27,7 @@
 import AppNavigationItem from '@nextcloud/vue/dist/Components/AppNavigationItem'
 
 export default {
-	name: 'NavigationAccount',
+	name: 'NavigationAccountExpandCollapse',
```
ChristophWurst: good catch :)

src/components/Navigation.vue
```diff
@@ -100,15 +108,21 @@ export default {
...
-				const folders = this.$store.getters
-					.getFolders(account.id)
-					.filter((folder) => !account.collapsed || SHOW_COLLAPSED.indexOf(folder.specialRole) !== -1)
-				//.map(folderToEntry)
+				const folders = this.$store.getters.getFolders(account.id)
+				const nonSpecialRoleFolders = folders.filter(
+					(folder) => SHOW_COLLAPSED.indexOf(folder.specialRole) === -1
+				)
+				const isCollapsible = nonSpecialRoleFolders.length > 1
+				const foldersToShow = folders.filter(
```
ChristophWurst: is this filter necessary? The template does that already in the `v-show`, right?

xMartin: I think you're right. It got introduced in 47ff8d0 in the template and I cannot see when this would ever have an effect. Why is it being used on the `show` attribute? Is there some kind of animation planned?

jancborchardt: > Is there some kind of animation planned?

Ideally yes, the "Expand/Collapse" control should move down to the bottom in a quick animation, revealing the folders. We have `--animation-quick:100ms;` or `--animation-slow:300ms;` for that, and the -slow is probably the fitting one here.

Also :wave: @xMartin :D

ChristophWurst: Yeah, but let's do the animation in a separate PR :)

xMartin: @jancborchardt 👋

xMartin: @ChristophWurst I cleaned it up. The starred inbox is always shown anyway if it's rendered so I removed the condition there. Actually there's a bug that the inbox is always rendered twice. Not sure how it's supposed to work but I see it as a seperate problem.

ChristophWurst: > ctually there's a bug that the inbox is always rendered twice. Not sure how it's supposed to work but I see it as a seperate problem.

Fixed via https://github.com/nextcloud/mail/pull/2803 :)

ChristophWurst: >  The starred inbox is always shown anyway if it's rendered so I removed the condition there.

Good catch. That check is indeed useless :)

ChristophWurst:
@xMartin you may also join as at https://cloud.nextcloud.com/call/89474m7g for development related discussions!

PR #2822, opened 2020-03-29
Add option to add share link to a file

### LINKED ISSUE #52: Add files from Nextcloud as link [$15]
As we previously talked about, it would be useful to be able to put shared links of files from your Nextcloud directly in your email.

This makes sending mails faster cause attachments don't need to be uploaded. And your files don't leave your server.
When choosing files from your Nextcloud, there could be a split button with the choice of either »Attach« or »Add as shared link«.

<bountysource-plugin>

---
There is a **[$15 open bounty](https://www.bountysource.com/issues/37682663-add-files-from-nextcloud-as-link?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gary-kim:
Resolves #52

This adds the option to share a link to a file from Files.

Also, I was thinking it might be worth including the file sharing service in `@nextcloud/dialogs`.

### DIFF
--- a/src/components/ComposerAttachments.vue
+++ b/src/components/ComposerAttachments.vue
@@ -1,5 +1,6 @@
 <!--
   - @copyright 2018 Christoph Wurst <christoph@winzerhof-wurst.at>
+  - @copyright 2020 Gary Kim <gary@garykim.dev>
   -
   - @author 2018 Christoph Wurst <christoph@winzerhof-wurst.at>
   -
@@ -41,12 +42,14 @@
 <script>
 import map from 'lodash/fp/map'
 import trimStart from 'lodash/fp/trimCharsStart'
+import {getRequestToken} from '@nextcloud/auth'
 import {translate as t} from '@nextcloud/l10n'
 import {getFilePickerBuilder} from '@nextcloud/dialogs'
 import Vue from 'vue'
 
 import Logger from '../logger'
 import {uploadLocalAttachment} from '../service/AttachmentService'
+import {shareFile} from '../service/FileSharingService'
 
 export default {
 	name: 'ComposerAttachments',
@@ -80,6 +83,7 @@ export default {
 	created() {
 		this.bus.$on('onAddLocalAttachment', this.onAddLocalAttachment)
 		this.bus.$on('onAddCloudAttachment', this.onAddCloudAttachment)
+		this.bus.$on('onAddCloudAttachmentLink', this.onAddCloudAttachmentLink)
 	},
 	methods: {
 		onAddLocalAttachment() {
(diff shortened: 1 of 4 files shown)


src/service/FileSharingService.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Makes a share link for a given file or directory.
+ * @param {string} path The file path from the user's root directory
+ * @param {string} token The conversation's token
+ * e.g. `/myfile.txt`
+ */
+const shareFile = async function (path, token) {
+	try {
+		return axios.post(generateOcsUrl('apps/files_sharing/api/v1', 2) + 'shares', {
```
ChristophWurst: don't you miss the `await` here? :thinking:

src/service/FileSharingService.js
```diff
@@ -0,0 +1,62 @@
...
+			error.response.data.ocs &&
+			error.response.data.ocs.meta &&
+			error.response.data.ocs.meta.message
+		) {
+			console.error(`Error while sharing file: ${error.response.data.ocs.meta.message || 'Unknown error'}`)
+			showError(error.response.data.ocs.meta.message)
+		} else {
+			console.error(`Error while sharing file: Unknown error`)
+			showError(t('mail', 'Error while sharing file'))
+		}
```
ChristophWurst: we should still (re)throw an error here so the caller knows this wasn't successful

src/components/ComposerAttachments.vue
```diff
@@ -133,12 +136,27 @@ export default {
 				.then((path) => this.emitNewAttachment(this.fileNameToAttachment(path)))
 				.catch((error) => Logger.error('could not choose a file as attachment', {error}))
 		},
+		onAddCloudAttachmentLink() {
+			const picker = getFilePickerBuilder(t('mail', 'Choose a file to share as a link')).build()
+
+			return picker
+				.pick(t('mail', 'Choose a file to share as a link'))
+				.then(async (path) => {
+					const url = (await shareFile(path, getRequestToken())).data.ocs.data.url
```
ChristophWurst: I think the `.data.ocs.data.url` details should go into the service

ChristophWurst:
> Also, I was thinking it might be worth including the file sharing service in `@nextcloud/dialogs`

I think sharing is so complex it would deserve its own package, no? Also this isn't a dialog but just some abstraction of the OCS API

src/service/FileSharingService.js
gary-kim: That's from [here](https://github.com/nextcloud/spreed/blob/bc22c21cf70a6243e7df7d838d49018c61468050/src/services/filesSharingServices.js) actually and now that you mention it, yeah, I think it should await there. I'll change it while moving `.data.ocs.data.url`.

gary-kim:
> I think sharing is so complex it would deserve its own package, no?

Your next package when you've got some time? :)

gary-kim:
> Squash and I'll :shipit:

Done!

Also, why is the `:shipit:` emoji a squrriel in a suit and top hat?

ChristophWurst:
> Also, why is the `:shipit:` emoji a squrriel in a suit and top hat?

I don't know, but the internet has answers: https://www.quora.com/On-GitHub-what-is-the-significance-of-the-Ship-It-squirrel :grinning:

PR #2834, opened 2020-03-31
Drag'n'Drop support 

StCyr:
Drag'n'Drop support implementation with https://github.com/Vivify-Ideas/vue-draggable

Remaining issues:

- [ ] It seems the code isn't properly initialized: If you load the mail apps from your browser, the drag'n'drop functionality won't work directly. You first have to perform an action (eg: selecting another email than the default one displayed) before the drag'n'drop functionality will work;
- [ ] Shouldn't navigate when the moved message is not tghe one currently displayed
- [ ] Do not expand when folder's account is not the same as envelope account
- [ ] Polishing

This PR is a continuation of https://github.com/nextcloud/mail/pull/1753


StCyr:
Hello @jancborchardt 

I don't know if D&D support is the best option we have to classify e-mails.

I was also thinking about adding a "move" action in the same way how it is done in the "file" app:

image

jancborchardt:
> Hello @jancborchardt
> 
> I don't know if D&D support is the best option we have to classify e-mails.
> 
> I was also thinking about adding a "move" action in the same way how it is done in the "file" app:

Yes, absolutely! :) We should have a "Move" action:
- In the action menu in the message list, as well as in the header in the detail view.
- Both times, the action should be directly below the "Mark read/unread" action
- It should open a modal with the folder list to pick the destination folder,
- Ideally also separated by accounts so you could move messages between different accounts, if that is possible. In that case it would be good if the account where the mail is currently in is sorted up top so if you just want to move inside of that account, you can quickly do that without scrolling.

jancborchardt:
@StCyr fully agree → I moved my feedback into the issue which already exists about that: https://github.com/nextcloud/mail/issues/255#issuecomment-591362380 :)

StCyr:
This is very old and the move feature introduced in https://github.com/nextcloud/mail/pull/3591 will solve the issue this PR tried to solve

PR #2837, opened 2020-03-31
Add missing bus prop for settings page

ChristophWurst:
`bus` is a required prop, so we have to specify it everywhere we use the component.

To test open the account settings. On master you'll see a js error, on this branch there is none.


gary-kim:
Oops, didn't think to check if the `TextEditor` was used anywhere else. :see_no_evil:

Issue #2699, opened 2020-02-27
Replying to HTML message in plain text misses > characters


jancborchardt:
I’m getting an HTML mail and reply in plain text only. The mail looks like this for example:

```
Hello,

this is a message. 
```

## Expected

When I reply or forward, the message is prefilled into the field like this:
```
"Name" name@example.com – 20. Februar 2020 12:15
> Hello,
> 
> this is a message. 
```

## What actually happens
There are no `>` characters at all which would separate the original message from my reply:
```
"Name" name@example.com – 20. Februar 2020 12:15

Hello,



this is a message. 
```

Setting to priority as this makes replying to HTML mails very cumbersome.

ahstro:
I've been giving this a cursory glance, and running `this.originalBody` through `htmlToText` before sending to `buildReplyBody` gives the output we'd like. Then it'd be a matter of getting access to the editor mode in `NewMessageDetail.vue` and formatting accordingly.

However, this might be a red herring, since I realized that when toggling "Enable formatting", the quotes disappear. Not sure if we want to tackle that directly somehow, or if we want to close this issue ASAP?

ChristophWurst:
Thanks a lot for looking into this issue.

Regardless of this specific bug, I have the feeling that our handling of plaintext, HTML text and the conversions between those two is very fragile. We've had it a few times where we fixed one bug. And at another spot we created three new ones.

So, yeah, we could hack this into the new code. But we're not really solving a problem. As we don't have a new release in sight as long as https://github.com/nextcloud/mail/projects/9 doesn't work reliably, we got some time to do it right :)

IMO we need a better abstraction of *text*. Also we should revisit on which level we have which type of text.

I have not put much thought into this yes, but what about a little rich object like

```js
{
    type: 'html',
    encoded: true,
    value: '...'
}
```

and changing the current conversion helpers and components to consume and emit this structure. Then we are more aware of what the value we're passing around represents. And we can unit tests (some) of the logic.

@ahstro what do you think? Does that sound reasonable?

ahstro:
Yeah, I completely agree that a better abstraction of text would be nice! Not just passing along strings and counting on context, but having a dedicated _type_ feels like a great improvement to me 😁

[...]

PR #2838, opened 2020-03-31
Abstract text into an object with known format

ChristophWurst:
So, @ahstro, this should be more or less what we discussed in https://github.com/nextcloud/mail/issues/2699#issuecomment-593405824. We now have an abstract *Text* type that can be plain or html and functions around it. This shall make it possible for us to just pass around an object of this type and always be able to check the type or run the correct operations on it.

For https://github.com/nextcloud/mail/issues/2699 I noticed that we use HTML for plain text replies all the way down to the composer component. So we actually quote the original message, but I guess ckeditor removes it because blockquote is not an allowed formatting in plain text emails. I yet  have to find a solution for that, but I think the refactoring done here is worth considering for integration because, as discussed, these problems with different formats and our fragile handling are not ideal.

@jancborchardt you may help me smoke test this. Just use the app to write, reply and draft messages both in HTML and plain text mode. It should work just like on master. My gut feeling tells me I missed one or two spots, but it's not very easy to find in the code.


GretaD:
1. Click on favorites folder, open a favorite email, try to reply that email. The envelope list changes to inbox and the message spins infinitely to reply. error: TypeError: "e.replace is not a function"
2. When i reply a html message, the previews message disappear and only my signature is shown.
3. I cannot reply to a plaintext messages because it loads infinitely.

jancborchardt:
Replying to HTML mails I don’t get _any_ part of the original message body transferred into the reply composer. Error in console:
```
[Vue warn]: Invalid default value for prop "body": Props with type Object/Array must use a factory function to return the default value.

found in

---> <Composer> at src/components/Composer.vue
       <AppContentDetails>
         <NewMessageDetail> at src/components/NewMessageDetail.vue
           <AppContent>
             <MailboxMessage> at src/components/MailboxMessage.vue
               <Content>
                 <Home> at src/views/Home.vue
                   <App> at src/App.vue
                     <Root> vue.runtime.esm.js:619

```

ChristophWurst:
> Invalid default value for prop "body": Props with type Object/Array must use a factory function to return the default value

You weren't testing the latest version I think. This was fixed in https://github.com/nextcloud/mail/commit/5c1c0d8bd08945a459b40577e024af74a7375a54 :man_shrugging:

Issue #2830, opened 2020-03-30
Add an option to open a message in a new window


GretaD:
### Feature Request
Add an option to open a message in a new window

#### Summary
We dont have an option to open the message in a new window

ChristophWurst:
I think we'd just need an entry in the actions menu. We have the deep URL for a message. You can already open it in a new tab with <kbd>Ctrl</kbd>+click

jancborchardt:
Closing as we won’t introduce this action for now:
- No clear problem statement
- Way less important than the other actions
- We have more actions which are way more important still to come (like Select and Move, and more)

PR #2839, opened 2020-03-31
Open email in a new tab

GretaD:
ref #2830 
@jancborchardt what icon we should use for this?

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -32,6 +32,8 @@
 			<ActionButton icon="icon-starred" @click.prevent="onToggleFlagged">{{
 				data.flags.flagged ? t('mail', 'Unfavorite') : t('mail', 'Favorite')
 			}}</ActionButton>
+			<ActionButton icon="icon-mail" @click="onNewTab">
+				{{ t('mail', 'Open in a new tab') }}</ActionButton>
 			<ActionButton icon="icon-mail" @click.prevent="onToggleSeen">{{
 				data.flags.unseen ? t('mail', 'Mark read') : t('mail', 'Mark unread')
 			}}</ActionButton>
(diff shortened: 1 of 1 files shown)


jancborchardt:
I would say we do not need this action at all, it’s not really essential and we should really beware of putting lots of actions in the menu. I also haven’t seen an issue report about that – does it solve a specific problem that people encounter?

Also cc @ChristophWurst?

GretaD:
> I would say we do not need this action at all, it’s not really essential and we should really beware of putting lots of actions in the menu. I also haven’t seen an issue report about that – does it solve a specific problem that people encounter?
> 
> Also cc @ChristophWurst?

Yes, it is requested this and also for the composer to be able to open it in a new window. #2830

jancborchardt:
> Yes, it is requested this and also for the composer to be able to open it in a new window(on forum, and i open the ticket here). #2830

Even after searching I can not find the forum post, do you have a link?

ChristophWurst:
https://help.nextcloud.com/t/message-display-window-and-rendering/75444/3

GretaD:
> > Yes, it is requested this and also for the composer to be able to open it in a new window(on forum, and i open the ticket here). #2830
> 
> Even after searching I can not find the forum post, do you have a link?

CHristoph suggested to open a ticket for all those that didnt exist: https://help.nextcloud.com/t/message-display-window-and-rendering/75444/3, i did, and start working on some

jancborchardt:
@GretaD ah, thanks! :) Many of the points in there are valid, and some of them are nicely enough already in the works – but this one I’d say no to.

We already have a bunch of long standing design issues open and we should possibly focus on those before we get new ones in where we don’t yet know if we actually want to have it. :) https://github.com/nextcloud/mail/issues?q=is%3Aopen+is%3Aissue+label%3Adesign

jancborchardt:
Closing this as discussed and we won’t introduce this action for now.

PR #1774, opened 2019-05-23
Consolidate message reply buttons

StCyr:
This PR intends to replace the "Send" and "Forward" buttons by a single action menu ("..." menu) providing these functionalities plus the "Reply all" function

To do

* [x] Position actions at the top right
* [x] Expand the popover menu to a vertical group of buttons

PR #2846, opened 2020-04-02
Fix missing In-Reply-To header for replies

ChristophWurst:
When a message is answered, we have to set the in-reply-to header, so other clients can build the correct message thread. Evolution, for example, builds threads only based on the header, thus @ma12-co's replies on company emails always show up as new messages for me.

Steps to reproduce
* Open app
* Open a message
* Reply to it
* Open the sent message and inspect the source (action menu on the top right #2264)

On master: no `In-Reply-To` header found
Here: `In-Reply-To` is set to the message ID of the previous message

This might be an unnoticed regression of https://github.com/nextcloud/mail/pull/1774

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -260,6 +260,10 @@ export default {
 			type: Function,
 			required: true,
 		},
+		replyTo: {
+			type: Object,
+			default: () => undefined,
+		},
 	},
 	data() {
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Bildschirmfoto von 2020-04-02 09-41-52

PR #2871, opened 2020-04-06
Fix signature settings layout, simplify HTML

jancborchardt:
Before:
- Not obvious that it’s a text field
- Field way too narrow and not high enough
- Unnecessary containers in the HTML

We could still improve it by adding autosize back so the field will always be as high as what is in it, but not sure this works with CKEditor? @ChristophWurst 

signature before

After:
signature after

### DIFF
--- a/src/components/SignatureSettings.vue
+++ b/src/components/SignatureSettings.vue
@@ -25,22 +25,18 @@
 		<p class="settings-hint">
 			{{ t('mail', 'A signature is added to the text of new messages and replies.') }}
 		</p>
-		<p>
-			<TextEditor v-model="signature" :html="true" :placeholder="t('mail', 'Signature …')" :bus="bus" />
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> We could still improve it by adding autosize back so the field will always be as high as what is in it, but not sure this works with CKEditor? @ChristophWurst

We don't need autosize, I think. We should be able to just style the editor element.

GretaD:
> Excellent. This confused a few people :)
> 
> Run `npm run lint:autofix` please

Did you test it? I can save it, but it doesnt show to the email after

ChristophWurst:
> Did you test it?

Nope

ChristophWurst:
Please debug it. See what is sent to the server and saved in oc_mail_accounts. Then we know which part fails.

PR #2873, opened 2020-04-06
Lower the memory footprint of the initial message cache sync

### LINKED ISSUE #2855: v1.3.0 Beta 3: cron.php: Allowed memory size of 134217728 bytes exhausted
> This is my first bog report in github / for nextcloud. Tried the search, but found nothing.

### Expected behavior
I use NextCloud 18.0.3 on a hosted webspace (bitpalast.net) with cron support. (https://docs.nextcloud.com/server/18/admin_manual/configuration_server/background_jobs_configuration.html#cron). The month before testing Mail v1.3.0 Beta 2 there were no cron.php errors.
After testing Mail v1.3.0 Beta 2, I received random daily errors from executing cron.php. All errors referring to nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/.
Before testing 1.3.0 I've got no errors from executing cron.php.

After receiving the errors, I tried to play with folloging settings in Plesk:

- PHP: memory_limit 512M (max from hosting provider), 256M, 128M
- PHP: handler type 1) FPM by Apache 2) FPM by nginx (https://docs.plesk.com/en-US/onyx/customer-guide/70742/#choosing-the-handler-type)

 

### Actual behavior
All erros till today:
> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 655360 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 179

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 528384 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 179

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 528384 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 179

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 528384 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 179

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 655360 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Fetch/Results.php on line 178

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 528384 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 179

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 528384 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 369

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 528384 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php on line 179

> PHP Fatal error:   Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/vhosts/xxx/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Idna/Horde/Idna.php on line 59

### Mail app

**Mail app version:** 1.3.0 (Beta 3)

**Mailserver or service:** IMAP from Googlemail and IMAP from hosted webserver on same machine.

### Server configuration

**Operating system**: Linux lech 3.10.0-957.1.3.el7.x86_64/hosted with Plesk

**Web server:** Apache / nginx (tried Plesk config: 1) FPM by Apache 2) FPM by nginx https://docs.plesk.com/de-DE/onyx/customer-guide/websites-und-domains/hostingeinstellungen/einstellungen-f%C3%BCr-webskripting/phpeinstellungen.70742/) before was FPM by nginx

**Database:** MariaDB Version 10.3 (or newer)

**PHP version:** 7.3.16

#### Client configuration
**Browser:** N/A

**Operating system:** N/A

ChristophWurst:
The initial message sync has to fetch potentially large amounts of data
and insert that into the database. To work around limitations with sync
requests triggered by web requests the process had already been made
interruptable and resumable. This means we never insert all the data
right away. Yet, the IMAP code fetched all UIDs before we capped it to a
maximum number of results per sync attempt. Depending on the mailbox
size this operation could require and allocate a lot of memory. On some
setup with lower memory limits, the process was aborted by the web
server due to a php memory exhaustion.

This patch modifies the IMAP code to optimize the memory usage by
limiting the amount of data that is fetched with each initial sync
attempt. The algorithm works as follows.

IMAP allows us to search in a range with a lower an upper bound UID.
While we know the highest known UID from the current cache values, we
can't derive the range for the next page from that as UIDs are not
continuous but might have holes due to deleted messages. If we assume
that messages of a mailbox are roughly distributed equally across the
assigned UIDs we can guess the max UID for the next range.
So we ask the server for min and max UIDs. The min or our known highest
UID is always the lower bound. Then we can calculate the distribution
rate from the min, max and number of messages and build the upper bound.
On everage this will fetch about the expected number of messages. It
could be more, but it could also be less. It shouldn't matter in most
cases.
The only downside of this approach is that we don't know if we got back
the last page. Previously we could just check if the page limit was
reached. Now we have to do an additional query. Only if that is empty,
we know that there is no more data to fetch.

Fixes #2855 

## Performance analysis

| **Measurement** | **Old** | **New** | **Difference** |
| -- | -- | -- | -- |
| Time initial sync 1 [s] | 241.55 | 208.54 | 13.6% less |
| Time initial sync 2 [s] | 253.63 | 209.56 | 17.3% less |
| Time initial sync 3 [s] | 242.27 | 211.27 | 12.8% less |
| **Average time initial sync [s]** | 245.78 | 209.79 | 14.6% less |
| Memory usage [MB] | 221 | 185 | 16.3% less |
| Estimated IMAP data transferred¹ [MB] | 28 | 18 | 35.7% less |

So, I would argue this is quite an improvement. Server with low mem are less likely to run out of it, servers with slow CPUs are less likely to run into the timeout. Remote IMAP servers (quite likely) are less likely causing the sync to limit the performance.

**A note on the overall performance metrics here: this is my dev setup. It uses xdebug, hence php significantly slower than in production. All logs are enabled, hence lots of unnecessary disk IO is performed. The IMAP server is rather large (58k messages) and on a remote server with relatively poor performance. Production instances perform quicker.**

¹ Cleared `data/horde_imap.log` and measured its size after the sync

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -23,12 +23,14 @@ declare(strict_types=1);
 
 namespace OCA\Mail\IMAP;
 
+use Generator;
 use Horde_Imap_Client;
 use Horde_Imap_Client_Base;
 use Horde_Imap_Client_Data_Fetch;
 use Horde_Imap_Client_Exception;
 use Horde_Imap_Client_Fetch_Query;
 use Horde_Imap_Client_Ids;
+use Horde_Imap_Client_Search_Query;
 use Horde_Imap_Client_Socket;
 use Horde_Mime_Mail;
 use Horde_Mime_Part;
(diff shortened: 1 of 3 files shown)


ChristophWurst:
With logging and xdebug disabled I can sync the same IMAP server in 103s and only 135MB of memory.

lib/IMAP/MessageMapper.php
```diff
@@ -75,40 +77,94 @@ public function find(Horde_Imap_Client_Base $client,
...
+				'all' => true,
+			];
+		}
+
+		// The inclusive range of UIDs
+		$totalRange = $max - $min + 1;
+		// Here we assume somewhat equally distributed UIDs
+		// +1 is added to fetch all messages with the rare case of strictly
+		// continuous UIDs and fractions
+		$estimatedPageSize = (int)(($totalRange / $total) * $maxResults) + 1;
```
skjnldsv, line 128:
```suggestion
		$estimatedPageSize = (int) (($totalRange / $total) * $maxResults) + 1;
```

You use space in the file above, I'm fine without, but pick one style :grin:

lib/IMAP/MessageMapper.php
```diff
@@ -75,40 +77,94 @@ public function find(Horde_Imap_Client_Base $client,
 	 *
 	 * @param int|null $highestKnownUid
```
skjnldsv, line 78: Missing @param  maxResults

PR #2874, opened 2020-04-07
Fix handling of plain/html replies in plain/html

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2699

Basically we didn't take the source and destination format into consideration when building the editor body. There are four different use cases that need to be covered. We only had three.

Please test

* Reply to a plain text message with your editor mode set to plain text
* Reply to a plain text message with your editor mode set to rich text
* Reply to a HTML text message with your editor mode set to plain text
* Reply to a HTML text message with your editor mode set to rich text


jancborchardt:
Good stuff, thanks a lot @ChristophWurst!

PR #2891, opened 2020-04-08
Fix the size of the source of raw msg

GretaD:
Now it looks like this. The close button doesnt show. We need to fix that, but how, should we put it inside container?
The large version only makes it taller so no big difference.
source

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -53,7 +53,7 @@
 							{{ t('mail', 'Delete') }}
 						</ActionButton>
 					</Actions>
-					<Modal v-if="showSource" @close="onCloseSource">
+					<Modal v-if="showSource" :size="isMobile ? 'large' : 'full'" @close="onCloseSource">
 						<div class="section">
 							<h2>{{ t('mail', 'Message source') }}</h2>
 							<pre class="message-source">{{ rawMessage }}</pre>
@@ -98,6 +98,7 @@ import MessagePlainTextBody from './MessagePlainTextBody'
 import Loading from './Loading'
 import logger from '../logger'
 import MessageAttachments from './MessageAttachments'
+import isMobile from '@nextcloud/vue/dist/Mixins/isMobile'
 
 export default {
 	name: 'Message',
@@ -115,6 +116,7 @@ export default {
 		Modal,
 		Popover,
 	},
+	mixins: [isMobile],
 	data() {
 		return {
 			loading: true,


ChristophWurst:
> The close button doesnt show

@skjnldsv to the rescue :)

GretaD:
> > The close button doesnt show
> 
> @skjnldsv to the rescue :)

FYI, if you click on the top right side, it will be closed

skjnldsv:
https://github.com/nextcloud/mail/blob/714a87d20c30e3b84b90dbba3ef62cf35b81aac2/src/components/Message.vue#L58

I guess you also need to make the content larger, the modal sizes are only max-width/height iirc

skjnldsv:
> FYI, if you click on the top right side, it will be closed

Most likely a cache issue for the icon?

skjnldsv:
> > We want to keep the modal style even on mobile, and not go to fullscreen. :) We just need to make sure that on mobile it uses more width and possibly height.

Looking at the code this is what Greta did?

jancborchardt:
> > > We want to keep the modal style even on mobile, and not go to fullscreen. :) We just need to make sure that on mobile it uses more width and possibly height.
> 
> Looking at the code this is what Greta did?

On mobile, "full" mode is used. What I’m saying is we should _always_ use "large" mode and never "full".

The issue is that "large" does leave quite some space to the sides on mobile and that should be reduced a bit to show more of the content of the modal.

skjnldsv:
> The issue is that "large" does leave quite some space to the sides on mobile and that should be reduced a bit to show more of the content of the modal.

Maybe full could be reduce a bit? Or add a 4th size?

GretaD:
> On mobile, "full" mode is used with this pull request. What I’m saying is we should _always_ use "large" mode and never "full".
> 
> The issue is that "large" does leave quite some space to the sides on mobile and that should be reduced a bit to show more of the content of the modal.

I was pro large too, till i tested it, the large size just increases the height, so no big difference. So adding a 4th one maybe its a good option as @skjnldsv proposed.

GretaD:
@skjnldsv @jancborchardt so what to do here, change the component to fix the large size? Or add a 4th one|?

skjnldsv:
> @skjnldsv @jancborchardt so what to do here, change the component to fix the large size? Or add a 4th one|?

I'll let Jan decide :)

jancborchardt:
What is the issue with just making the `large` size (which is used in master now, as it should be) fill a bit more of the screen space on mobile?

Just up the max-width and max-height a bit, or reduce the outer margins – depending on how it's implemented.

We do not need a 4th size, and this should be fixed for the `large` component anyway – not only in Mail.

GretaD:
> What is the issue with just making the `large` size (which is used in master now, as it should be) fill a bit more of the screen space on mobile?

Nothing wrong with it, we are putting out options and deciding whats the best. So I'm moving on and changing the large size. :+1:

jancborchardt:
> > What is the issue with just making the `large` size (which is used in master now, as it should be) fill a bit more of the screen space on mobile?
> 
> Nothing wrong with it, we are putting out options and deciding whats the best. So I'm moving on and changing the large size. +1

Cool, thanks! :) Then this pull request is not needed, as the changes will be made in nextcloud-vue, correct? (Because I just tested and currently on large screens, it uses the full-screen display, which is not what we want.)

jancborchardt:
Closing in favor of the component pull request at https://github.com/nextcloud/nextcloud-vue/pull/1028

PR #2893, opened 2020-04-08
Do not navigate from a message to just the folder

ChristophWurst:
As discovered by @jancborchardt the router navigated from a *message route* to the *folder route*.

Let me elaborate. We have deep links for folders and messages. When a folder is opened we automagically redirect to the first message in the list once that got loaded. If you then click on the folder, the app routed back to just the folder. The active message was reset and a "No message selected" appeared. This was not desired by @jancborchardt.

To test

* Open the app and wait until the first message is shown
* Click the folder name

On master: message is unselected, you see placeholder text
Here: you still see your message. The URL might flash shortly but it doesn't have an effect logically.

The solution is merely a hack. If anyone know a better way please shoot.

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -150,6 +150,17 @@ export default {
 	},
 	watch: {
 		$route(to, from) {
+			if (
+				from.name === to.name &&
+				Number.parseInt(from.params.accountId, 10) === Number.parseInt(to.params.accountId, 10) &&
+				from.params.folderId === to.params.folderId &&
+				from.params.messageUid === to.params.messageUid &&
+				from.params.filter === to.params.filter
+			) {
+				logger.debug('navigated but the message is still the same')
+				return
+			}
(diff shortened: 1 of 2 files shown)


jancborchardt:
> > This was not desired by @jancborchardt.

The @jancborchardt is pleased with this fix.

Agree with @kingjan1999 that the short URL flash is a bit strange though, and would be cool if fixable. Not a very big deal though.

-------------

> Actually I'd expect that when clicking on a folder, regardless if it is currently selected or not, the first message is loaded (this could be used as some kind of "back to top"). 

@kingjan1999 good point – I’d say though that we don’t navigate away as this seems unexpected? We can still see if that is actually needed, but for now I’d say let’s only fix the issue. :)

ChristophWurst:
Thanks for the feedback. I like the idea with going back to the top. Let's do that in a follow-up.

Can we get this in for v1.3?

PR #2894, opened 2020-04-09
Use eslint config

skjnldsv:
Still some errors left.
I separated the commits so it's easier to drop the fix one and re do it again.
I guess we should wait after your upcoming release?

Btw, you're still supporting 7.1?
I see it running on your tests :thinking:


ChristophWurst:
> Btw, you're still supporting 7.1?

Yes, for stable17.

> I guess we should wait after your upcoming release?

Yeah, this will cause conflicts with the open PRs I'm afraid

skjnldsv:
> Yeah, this will cause conflicts with the open PRs I'm afraid

Let's wait, this is not urgent :)

ChristophWurst:
Time to get this in now that we finished a new release. I'll have to go through the linter errors to fix them.

skjnldsv:
Rebased! Still needs some fixes here and there than lint:fix couldn't do automatically @ChristophWurst 
Let's do this as soon as you get back so it's not a nightmare to rebase again :see_no_evil:

skjnldsv:
Merging, I took your rebase as a yes.
All tests were green, blame me when you get back if it created anything bad :stuck_out_tongue_closed_eyes:

Issue #2910, opened 2020-04-15
Loading messages takes "forever" since upgrade to 1.3.0


loppanloppis:
### Expected behavior
My e-mails should load and show

### Actual behavior
It seems stuck on "Loading Messages. Indexing your messages. This can take a bit longer for larger mailboxes."

### Mail app

**Mail app version:** 1.3.0

**Mailserver or service:** one.com (IMAP)

### Server configuration

**Operating system**: Ubuntu 18.04

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.2.24

**Nextcloud Version:** 18.0.3

#### Client configuration
**Browser:** Chromium 79.0.3945.130

**Operating system:** Ubuntu 18.04

Hi, I'm fairly sure this issue is not identical to #2909 or #2908. Please excuse me if it is.

I'd like to mention that this e-mail account holds a few GB so I except it would take time to index, but the errors give me hints that it's not indexing at all :

[mail] Error: OCA\Mail\Exception\ServiceException: Sync failed:  at <<closure>>

0. /var/www/nextcloud/apps/mail/lib/Service/Sync/SyncService.php line 107
   OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 118}, 42, [], true)
1. /var/www/nextcloud/apps/mail/lib/Controller/FoldersController.php line 131
   OCA\Mail\Service\Sync\SyncService->syncMailbox(OCA\Mail\Account {}, "INBOX", 42, [], false, null)
2. /var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 170
   OCA\Mail\Controller\FoldersController->sync(4, "SU5CT1g=", [], true, null)
3. /var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 99
   OC\AppFramework\Http\Dispatcher->executeController(OCA\Mail\Controller\FoldersController {}, "sync")
4. /var/www/nextcloud/lib/private/AppFramework/App.php line 125
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\Mail\Controller\FoldersController {}, "sync")
5. /var/www/nextcloud/lib/private/AppFramework/Routing/RouteActionHandler.php line 47
   OC\AppFramework\App::main("OCA\\Mail\\Cont ... r", "sync", OC\AppFramework\ ... {}, {accountId: "4", ... "})
[...]

Issue #2908, opened 2020-04-15
Sync failed


crdxfire:
### Expected behavior
Sync with iCloud acct

### Actual behavior
Sync failed

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)
1.3.0

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
iCloud

### Server configuration

**Operating system**: (e.g. Debian 8)
Ubuntu 18.04

**Web server:** (e.g. Apache, Nginx,...)
Apache

**Database:** (e.g. MariaDB, SQLite or PostgreSQL)
MadiaDB

**PHP version:** (e.g. 7.0)
7.2..24

**Nextcloud Version:** (see admin page, e.g. 13.0.2)
18.0.3

#### Client configuration
**Browser:** (e.g. Firefox 48)
Several tested, mainly Brave

**Operating system:** (e.g. Arch Linux)
macOS 10.15.4

[mail] Error: OCA\Mail\Exception\ServiceException: Sync failed:  at <<closure>>

0. /var/www/nextcloud/apps/mail/lib/Service/Sync/SyncService.php line 107
   OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 267}, 42, [], true)
1. /var/www/nextcloud/apps/mail/lib/Controller/FoldersController.php line 131
   OCA\Mail\Service\Sync\SyncService->syncMailbox(OCA\Mail\Account {}, "INBOX", 42, [], false, null)
2. /var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 170
   OCA\Mail\Controller\FoldersController->sync(12, "SU5CT1g=", [], true, null)
3. /var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 99
   OC\AppFramework\Http\Dispatcher->executeController(OCA\Mail\Controller\FoldersController {}, "sync")
4. /var/www/nextcloud/lib/private/AppFramework/App.php line 125
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\Mail\Controller\FoldersController {}, "sync")
5. /var/www/nextcloud/lib/private/AppFramework/Routing/RouteActionHandler.php line 47
   OC\AppFramework\App::main("OCA\\Mail\\Cont ... r", "sync", OC\AppFramework\ ... {}, {accountId: "12" ... "})
[...]

PR #2912, opened 2020-04-16
Fix handling of service exceptions

ChristophWurst:
We have special handling for IncompleteSyncException in the controller,
but since the exception handling always wrapped the original exception
in a service exception, the type cast to the incomplete sync exception
didn't work. Now, it just doesn't wrap unless it's something other than
a service exception.

How to test: open a very large mailbox for the first time.

On master: sync returns HTTP 500 due to unhandled exception
Here: 202 is returned, client code knows it just has to send another request. Everything works flawless.

Fixes https://github.com/nextcloud/mail/issues/2910
Fixes https://github.com/nextcloud/mail/issues/2908


ChristophWurst:
@schu-r @rmkeen @loppanloppis @crdxfire

Issue #2909, opened 2020-04-15
Cannot load mail anymore since upgrade to 1.3.0


tjareson:
### Expected behavior
Mail should get loaded and displayed when clicking on the header.

### Actual behavior
The mail headers are getting loaded after minutes. 
When selecting an email it takes another 1-2 minutes, then it display "Not found   Could not load your message"
Occasionally the email get displayed like every 4th try or so.

Same mail account (IMAP account on MXROUTE) works with other mail clients (thunderbird, iPhone) w/o problems. 

### Mail app

**Mail app version:** 1.3.0

**Mailserver or service:** MXRoute (IMAP)

### Server configuration

**Operating system**: Ubuntu server 16.04

**Web server:** Apache (Nextcloud Docker) and nginx reverse proxy

**Database:** MySQL 10.0.38

**PHP version:** 	7.3.16

**Nextcloud Version:** official Nextcloud Docker 18.0.3

#### Client configuration
**Browser:** Opera, Google Chrome, Firefox

**Operating system:** Ubuntu 16.04, Win 10

Logfile content, when it doesn't work:
OCA\Mail\Exception\ServiceException: Sync failed: Mail server denied authentication.

    /var/www/html/custom_apps/mail/lib/Service/Sync/SyncService.php - line 107:

    OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox { id: 74}, 42, [ 40,39,38,3 ... 1], false)

    /var/www/html/custom_apps/mail/lib/Controller/FoldersController.php - line 124:

    OCA\Mail\Service\Sync\SyncService->syncMailbox(OCA\Mail\Account {}, "INBOX", 42, [ 40,39,38,3 ... 1], true, null)

    /var/www/html/lib/private/AppFramework/Http/Dispatcher.php - line 170:

    OCA\Mail\Controller\FoldersController->sync(5, "SU5CT1g=", [ 40,39,38,3 ... 1], false, null)

    /var/www/html/lib/private/AppFramework/Http/Dispatcher.php - line 99:

    OC\AppFramework\Http\Dispatcher->executeController(OCA\Mail\Con ... {}, "sync")

    /var/www/html/lib/private/AppFramework/App.php - line 125:

    OC\AppFramework\Http\Dispatcher->dispatch(OCA\Mail\Con ... {}, "sync")

[...]

PR #2920, opened 2020-04-16
Fix lower bound when fetching a range of IMAP messages for the sync

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2909

When we fetch messages for the initial sync, we use a UID range. There is a bit of logic involved to only fetch roughly `$maxResults` messages, and we might have a known offset `$highestKnownUid` to take into account for resumed sync's. The code previously always set the lower range bound to `$highestKnownUid + 1`, so for the very first sync attempt the range was like `0:$maxResults`. If the UIDs of a mailbox start *after* `$maxResults`, no messages were returned. And the logic entered an infinite retry loop.

Now, the new logic starts at `$min`, the lowest known UID reported by the server or at `$highestKnownUid`, whichever is higher. The upper bound is calculated based on the lower bound.

If my train of thought is not derailing, this should give the correct range to fetch. I can still correctly sync all messages with my test accounts.

TODO
- [x] Fix the bug
- [x] Add tests to verify the fix
- [x] ~~Remove debug logs once we have the confirmation this works from https://github.com/nextcloud/mail/issues/2909#issuecomment-614690050~~ leaving them in case we have to debug this another time

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -71,18 +71,18 @@ class MessageMapper {
 
 	/**
 	 * @param Horde_Imap_Client_Socket $client
-	 * @param Mailbox $mailbox
+	 * @param string $mailbox
 	 *
 	 * @param int $maxResults
-	 * @param int|null $highestKnownUid
+	 * @param int $highestKnownUid
 	 *
 	 * @return array
 	 * @throws Horde_Imap_Client_Exception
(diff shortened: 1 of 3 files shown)


ChristophWurst:
@skjnldsv remember
Bildschirmfoto von 2020-04-16 16-59-22

:speak_no_evil:

Issue #2915, opened 2020-04-16
New sent msgs doesn't get displayed in sent folder anymore (1.3.0) but are saved there


tjareson:
### Expected behavior
Sent mail should show up in sent-folder.

### Actual behavior
Sent mail do not get displayed in sent folder.
But: they are stored there. If I check the same mail account with e.g. thunderbird or iphone they are visible in sent-folder.
But: in nextcloud-mail folder info these sent mails are counted correctly, but just not displayed in the list.

That is creating a bit of confusion for the users right now, as they are not sure, if they have sent a message or not and also cannot forward anything they have sent out. 

**Mail app**
Mail app version: 1.3.0

**Mailserver or service:** MXRoute (IMAP)

**Server configuration**
**Operating system:** Ubuntu server 16.04

**Web server**: Apache (Nextcloud Docker) and nginx reverse proxy

**Database:** MySQL 10.0.38

**PHP version**: 7.3.16

**Nextcloud Version:** official Nextcloud Docker 18.0.3

**Client configuration**
**Browser:** Opera, Google Chrome, Firefox

**Operating system:** Ubuntu 16.04, Win 10

tjareson:
Additional comment: the sent mails showed up in the sent folder meanwhile. Unfortunately I cannot tell, after how long. 
I wrote another test mail, again it doesn't show up in sent on nextcloud mail (since 5 min...), but with all other mail clients. 
So far I couldn't find any link to cron times. cron.php is handled by crontab in my system setup.

ChristophWurst:
What happens if you press `r` in the sent messages list? Will it pop up?

ChristophWurst:
My guess is that the database cache is not updated immediately. But your cron might trigger the sync. The next time you open the folder, the message will be there.

tjareson:
It gets visible within like 10-20 minutes (couldn't check it every minute...)
With 'r' it gets loaded immediately indeed. 
The confusing part is: it does not update when I switch from inbox to send for some reason.

tjareson:
If I send myself an email from another account, the inbox folder gets updated on the spot (I tested it twice to be sure that it wasn't happened shortly before a cron job kicks in)
So somehow inbox folder seems to be handled differently than e.g. sent folder.

beccon4:
Version 1.3.1 still suffers from that: I write an email - send it off and it disappears entirely.

tjareson:
oh, I was still on 1.3.0. My bad...

[...]

PR #2934, opened 2020-04-17
Sync mailbox not just inbox

GretaD:
Fixes https://github.com/nextcloud/mail/issues/2915

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -87,6 +87,7 @@ export default {
 			loadingMore: false,
 			loadingEnvelopes: true,
 			loadingCacheInitialization: false,
+			loadMailboxInterval: undefined,
 		}
 	},
 	computed: {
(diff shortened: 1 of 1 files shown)


src/components/Mailbox.vue
```diff
@@ -349,6 +364,23 @@ export default {
...
+		if (this.account.isUnified || this.folder.specialRole === 'inbox') return
+
+		try {
+			await this.$store.dispatch('syncInboxes')
+
+			logger.debug("Mailbox sync'ed in background")
+		} catch (error) {
+			logger.error('Background sync failed: ' + error.message, {error})
+		}
+		clearInterval(this.loadMailbox)
```
ChristophWurst, line 378: this means `loadMailbox` will only be called once, right?

if you use setInterval then only clear it when the component is destroyed.

src/components/Mailbox.vue
```diff
@@ -348,6 +364,23 @@ export default {
 				},
 			})
 		},
+		async loadMailbox() {
+			//when the account is unified or inbox, return nothing, else sync the mailbox
+			if (this.account.isUnified || this.folder.specialRole === 'inbox') {
+				return
+			}
+			try {
+				await this.$store.dispatch('syncInboxes')
```
ChristophWurst, line 373: this should be
```
				await this.$store.dispatch('syncEnvelopes', {
					accountId: this.$route.params.accountId,
					folderId: this.$route.params.folderId,
					query: this.searchQuery,
				})
```

PR #2935, opened 2020-04-17
Make it possible to clear the message cache as user

ChristophWurst:
This isn't something we need, but as long as there are a few dodgy edge cases to fix, it would be great to offer the users the option to clear their cache.

TODO
- [x] Only show when debug mode is enabled.

### DIFF
--- a/src/components/NavigationFolder.vue
+++ b/src/components/NavigationFolder.vue
@@ -41,7 +41,6 @@
 					{{ statsText }}
 				</ActionText>
 
-				<!-- TODO: make *mark as read* available for all folders once there is more than one action -->
 				<ActionButton
 					v-if="folder.specialRole !== 'flagged'"
 					icon="icon-mail"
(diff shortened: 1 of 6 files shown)


ChristophWurst:
@nextcloud/mail please review & test

src/components/NavigationFolder.vue
```diff
@@ -59,6 +58,21 @@
...
+					v-if="debug && !account.isUnified && folder.specialRole !== 'flagged'"
+					icon="icon-settings"
+					:title="t('mail', 'Clear cache')"
+					:disabled="clearingCache"
+					@click="clearCache"
+				>
+					{{
+						t(
+							'mail',
+							'Clear all cached data. Useful when you experience problems with synchronization. Use with care.'
```
jancborchardt, line 72: Looks good, but why the "use with care"? Possibly shorter and more explanatory:

> Clear locally cached data, in case there are issues with synchronization.

ChristophWurst: Because it deletes a lot of data that has to be recreated. Users shouldn't do this just for fun. But they will do that anyway, so :+1: on the wording

ChristophWurst:
This is a per mailbox cache. If we combine it for each account, it's just even more work for the server and more likely to cause problems with some mail providers. Big :-1: on that.

This is just for debugging. No average user would ever see this. Can we get this in to debug issues on production instances please?

Issue #2930, opened 2020-04-17
Mailbox refresh hang forever after migration from 1.1.4 to 1.3.2


doc75:
### Expected behavior
Mailbox opened without any issue and list of mail appears after a couple of seconds

### Actual behavior
list of message never appear, the refresh circle stays forever

Here is the error log (mail address have been removed and replaced by `<text explaining the content remove and approximate length>`:

[error_clean.log](https://github.com/nextcloud/mail/files/4492199/error_clean.log)

Seems related to mail with a lot of recipients.

### Mail app

**Mail app version:** 1.3.2 (right after migration from a 1.1.4 which was working like a charm)

**Mailserver or service:** Dovecot (personal installation, the same one used with 1.1.4)

### Server configuration

**Operating system**: Alpine Linux (official Nextcloud docker image)

**Web server:** Nginx

**Database:** PostgreSQL

**PHP version:** 7.3.16

**Nextcloud Version:** 17.0.25 (official docker image)

#### Client configuration
**Browser:** Firefox 75.0

**Operating system:** Ubuntu 18.04

ChristophWurst:
> Seems related to mail with a lot of recipients.

Do you know the specific message that causes this? Mind sending an exported .mbox (via Thunderbird or Evolution) to my email? I would like to inspect that header. I think it's malformed.

The app inserts one entry per recipient. But somehow in your case many of them are crammed into a single item. Thus the error

```
"Message": "Sync failed for 1:INBOX: An exception occurred while executing 'INSERT INTO \"oc_mail_recipients\" (\"message_id\", \"type\", \"label\", \"email\") VALUES(?, ?, ?, ?)' with params [310, 1, <list of mail address separated by 2 commas and the last one surrounded by double quotes (number of char in the string is about 1900)>]:\n\nSQLSTATE[22001]: String data, right truncated: 7 ERROR:  value too long for type character varying(255)",
```

FleischKarussel:
I can confirm this, using Nextcloud 18.0.3 and mail 1.3.2.
I am using mailbox.org als IMAP provider.

ChristophWurst:
>  confirm this

what exactly are you confirming?

doc75:
> Do you know the specific message that causes this? Mind sending an exported .mbox (via Thunderbird or Evolution) to my email? I would like to inspect that header. I think it's malformed.

Mail sent with information.
Thanks.

zpintar:
[...]

PR #2955, opened 2020-04-20
Limit the size of recipient labels

ChristophWurst:
Ref https://github.com/nextcloud/mail/issues/2930

This is more likely to be triggered by an invalid message header as in the ticket, but even if the real label is that long, we won't have the space to show it anyway. So let's just limit it to the width of the db table.

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -36,6 +36,7 @@ use OCP\IDBConnection;
 use function array_combine;
 use function array_keys;
 use function array_map;
+use function mb_substr;
 
 class MessageMapper extends QBMapper {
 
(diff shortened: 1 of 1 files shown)


doc75:
@ChristophWurst patch tested on my full mailbox (containing at least the 2 corrupted messages) and it is working fine. Waiting for the official release.

Thanks again for this great App and the good work done.

lib/Db/MessageMapper.php
```diff
@@ -146,7 +147,7 @@ public function insertBulk(Message ...$messages): void {
 
 					$qb2->setParameter('message_id', $messageId, IQueryBuilder::PARAM_INT);
 					$qb2->setParameter('type', $type, IQueryBuilder::PARAM_INT);
-					$qb2->setParameter('label', $recipient->getLabel(), IQueryBuilder::PARAM_STR);
+					$qb2->setParameter('label', mb_substr($recipient->getLabel(), 0, 255, 'UTF-8'), IQueryBuilder::PARAM_STR);
```
nickvergessen, line 150: Should be enough without specifying the charset, see https://github.com/nextcloud/server/pull/19634

PR #2970, opened 2020-04-22
Migrate mail_messages.mailbox_id to an INT

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2963

TODO
- [ ] Basis migrations structure
- [ ] Drop the original unique index on [uid, mailbox_id]
- [ ] Create a new unique index
- [ ] Handle migration errors if possible


ChristophWurst:
This is actually not the one :)

https://github.com/nextcloud/mail/pull/2796 has the drop and recreate logic

PR #2972, opened 2020-04-23
Use all available group services

ChristophWurst:
When we just had one implementation of a group service we could just
inject the implementation when the interface is used. However, as
another implementation was added, it is necessary to explicitly require
the implementations to be able to use all of them.

Before the patch the contacts group integration overwrote the nextcloud group integration.

Fixes https://github.com/nextcloud/mail/pull/810

@nextcloud/mail

### DIFF
--- a/lib/Service/GroupsIntegration.php
+++ b/lib/Service/GroupsIntegration.php
@@ -23,7 +23,9 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Service;
 
+use OCA\Mail\Service\Group\ContactsGroupService;
 use OCA\Mail\Service\Group\IGroupService;
+use OCA\Mail\Service\Group\NextcloudGroupService;
 use OCA\Mail\Exception\ServiceException;
 
 class GroupsIntegration {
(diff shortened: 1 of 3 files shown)


lib/Service/GroupsIntegration.php
```diff
@@ -35,8 +37,11 @@ class GroupsIntegration {
 	 */
 	private $groupServices = [];
 
-	public function __construct(IGroupService $groupService) {
-		$this->groupServices[] = $groupService;
+	public function __construct(ContactsGroupService $contactsGroupService, NextcloudGroupService $nextcloudGroupService) {
+		$this->groupServices = [
```
bfritscher, line 41: So we decide on ordering of the services here?

If any service fails nothing is returned in getMatchingGroups, but that is outside the scope of this issue.

myrho: Code looks good.

However, I wanted to do an integration test on the GUI, but I'm not able to install the mail app since `composer install` fails on requiring `kwi/urllinker` which seems to be a private repo?

ChristophWurst: > fails on requiring `kwi/urllinker` which seems to be a private repo?

Are you sure it fails? I think that is just a warning. You can perhaps ignore that.

ChristophWurst: > If any service fails nothing is returned in getMatchingGroups, but that is outside the scope of this issue.

Yes and yes :)

Issue #2967, opened 2020-04-22
PHP encounters a non-numeric value at /var/www/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php#317


hansbaer443:
### Expected behavior
After conecting to a new IMAP account mail-app is presenting all mails in this inbox.

### Actual behavior
Mail-app shows "Keine Nachrichten in diesem Ordner" (in english "No messages in this folder") and the log shows "A non-numeric value encountered at /var/www/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php#317"

### Mail app

**Mail app version:** 1.3.3

**Mailserver or service:** Icloud mail (imap.mail.me.com)

### Server configuration

**Operating system:** Debian 10

**Web server:** Apache/2.4.38

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud Version:** 18.0.3
#### Client configuration
**Browser:** Firefox 75.0

**Operating system:** Windows 8.1

ChristophWurst:
> the log shows

Mind sharing the full log entry?

hansbaer443:
Here's the full log entry I just removed IP address, Username und browser ID: `{"reqId":"PWz0sokJNTDT0hofR4mO","level":3,"time":"2020-04-22T08:41:14+00:00","remoteAddr":"IPv6adress","user":"username","app":"PHP","method":"POST","url":"/apps/mail/api/accounts/3/folders/RHJhZnRz/sync","message":"A non-numeric value encountered at /var/www/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Imap_Client/Horde/Imap/Client/Ids.php#317","userAgent":"Mozilla/5.0 (Windows NT 6.3; Win64; x64; rv:75.0) Gecko/20100101 Firefox/75.0","version":"18.0.3.0","id":"ID"}`

Other IMAP accounts work fine. I had this issue only with a iCloud mailaccount.

ChristophWurst:
Thanks. So it's caused by our code here: https://github.com/nextcloud/mail/blob/e5f0e3591a44a08b36c96a52eb8d9f219bb8de2c/lib/IMAP/MessageMapper.php#L152. Somehow `$upper` is not an int (as I would expect) and therefore the numeric operation in horde fails.

[...]

PR #2975, opened 2020-04-23
Cast min/max/count numbers to integers

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2967


ChristophWurst:
Does not seem to be the problem.

Issue #2976, opened 2020-04-23
Hourly background job is raising exception linked to sync issue


doc75:
### Expected behavior
Background job shall end the sync (partial or full) without exception

### Actual behavior
Every hour I can see the 2 same errors raised:
<details>
  <summary>OCA\Mail\Exception\IncompleteSyncException</summary>

```json
{
  "reqId": "IXaVQTaqXm9b3wcTGEkK",
  "level": 2,
  "time": "2020-04-23T18:45:44+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "mail",
  "method": "",
  "url": "--",
  "message": {
    "Exception": "OCA\\Mail\\Exception\\IncompleteSyncException",
    "Message": "Initial sync is not complete for 4:INBOX",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/custom_apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php",
        "line": 179,
        "function": "runInitialSync",
        "class": "OCA\\Mail\\Service\\Sync\\ImapToDbSynchronizer",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Mail\\Account"
          },
          {
            "id": 72,
            "__class__": "OCA\\Mail\\Db\\Mailbox"
          }
        ]
      },
      {
        "file": "/var/www/html/custom_apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php",
        "line": 107,
        "function": "sync",
        "class": "OCA\\Mail\\Service\\Sync\\ImapToDbSynchronizer",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Mail\\Account"
          },
          {
            "id": 72,
            "__class__": "OCA\\Mail\\Db\\Mailbox"
          },
          42,
          null,
          false
        ]
      },
      {
        "file": "/var/www/html/custom_apps/mail/lib/BackgroundJob/SyncJob.php",
        "line": 86,
        "function": "syncAccount",
        "class": "OCA\\Mail\\Service\\Sync\\ImapToDbSynchronizer",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Mail\\Account"
          }
        ]
      },
      {
        "file": "/var/www/html/lib/public/BackgroundJob/Job.php",
        "line": 77,
        "function": "run",
        "class": "OCA\\Mail\\BackgroundJob\\SyncJob",
        "type": "->",
        "args": [
          {
[...]

PR #2979, opened 2020-04-24
Fix UID validity change logic

ChristophWurst:
- [x] Requires https://github.com/nextcloud/mail/pull/2978

    UID validity is a way for the IMAP server to signal that the UIDs - that
    are usually persistend - have changed. Whenever this happens clients
    have to wipe their cached data and read the IMAP data anew. Since making
    the initial sync process resumable, it does not clear the messages of a
    mailbox as these might be from a previous sync attempt. Hence, we have
    to manually clear the cache on a UID validity change before we invoke
    the initial sync routine.


ChristophWurst:
@nextcloud/mail please review carefully :eyes: and let me know if you have any questions :pray:

ChristophWurst:
Confirmed to be working at https://github.com/nextcloud/mail/issues/2976#issuecomment-619265642

PR #2980, opened 2020-04-24
Log the progress of an incomplete sync

ChristophWurst:
This is helpful to determine whether the initial sync just takes long or
got stuck.

Ref https://github.com/nextcloud/mail/issues/2976#issuecomment-618833115


ChristophWurst:
Confirmed to be working at https://github.com/nextcloud/mail/issues/2976#issuecomment-619265642

PR #3007, opened 2020-04-27
Change emptycontent on mail

### LINKED ISSUE #2961: Change how favorite section is shown when there's no favorites messages.
Right now it looks like this:

favorite

@nextcloud/designers , any idea?

GretaD:
fixes #2961 
before

beforefav

after
favorite

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -27,7 +27,8 @@
 		:hint="t('mail', 'Loading messages')"
 		:slow-hint="t('mail', 'Indexing your messages. This can take a bit longer for larger mailboxes.')"
 	/>
-	<EmptyMailbox v-else-if="envelopes.length === 0" key="empty" />
+	<EmptyMailboxSection v-else-if="isPriorityInbox && !hasMessages" key="empty"></EmptyMailboxSection>
+	<EmptyMailbox v-else-if="!hasMessages" key="empty" />
 	<EnvelopeList
 		v-else
 		:account="account"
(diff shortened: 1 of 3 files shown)


src/components/EmptyMailbox.vue
```diff
@@ -1,7 +1,6 @@
 <template>
 	<div id="emptycontent">
-		<div class="icon-mail"></div>
-		<h2>{{ t('mail', 'No messages in this folder') }}</h2>
+		<p> <slot /></p>
```
ChristophWurst, line 3: Imo this is a bit strange. The `<EmptyMailbox>` is only used once, so the slot value can just be moved here.

GretaD:
> This looks really good, but I think we should keep the old style for individual folders.
> 
but its too big :) 
> Can we have a new component for `EmptyMailboxSection` or similar for the PI and the old (unchanged) component for any other?

I dont have a strong opinion about the current emtymailbox to be honest, except that is too big :) 
I agree on the new component and will work on it today.

src/components/Mailbox.vue
```diff
@@ -27,7 +27,8 @@
 		:hint="t('mail', 'Loading messages')"
 		:slow-hint="t('mail', 'Indexing your messages. This can take a bit longer for larger mailboxes.')"
 	/>
-	<EmptyMailbox v-else-if="envelopes.length === 0" key="empty" />
+	<EmptyMailboxSection v-else-if="!hasMessages && !folder.isPriorityInbox" key="empty"></EmptyMailboxSection>
```
ChristophWurst, line 30: :eyes: :eyes: :eyes:

GretaD: !!!!!!!!!

src/components/Mailbox.vue
```diff
@@ -102,6 +106,17 @@ export default {
 		envelopes() {
 			return this.$store.getters.getEnvelopes(this.account.id, this.folder.id, this.searchQuery)
 		},
+		hasMessages() {
```
ChristophWurst, line 109: I know this is/was necessary in `MailboxMessage` but have you tried `return this.envelopes.length !== 0`? Since we use the list in other places inside this component, that should actually work just fine

ChristophWurst: also because that is exactly what was used in the template :)

Issue #1501, opened 2019-01-27
Accounts not deleted when Nextcloud account is deleted


anojht:
Hello, while testing we discovered that externally authenticated users with the same username but different domains have access to mailbox data from first user. This is after the user is first removed from the admin UI, before setting up the external authentication for a different domain with the same username.

This presents a huge security issue for users/admins who are using the mail app with external users authenticated to their instance should they ever decide to add other domains for external authentication or switch domains.
### Steps to Reproduce
1. Setup external user support app with IMAP `userid@domain.tld ` for authentication.
2. Mail app auto sets up their mailbox once the user logs in.
3. Delete the user from admin UI.
4. Setup external user support app with IMAP `sameuserid@differentdomain.tld` for authentication.
5. Login with the user and click the Mail app.

### Expected behavior
Removing IMAP authenticated external user account should remove their mailbox data

### Actual behavior
Mail app retains and displays the old user's mail account, folders, emails, and receives new emails alongside the new account.

### Mail app

**Mail app version:** 0.11.0

**Mailserver or service:** Any email service

### Server configuration

**Operating system**: unRAID 6.6.6

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.2.13

**Nextcloud Version:** 15.0.2

#### Client configuration
**Browser:** Chrome 71.0.3578.98

**Operating system:** Windows 10

ChristophWurst:
> Setup external user support app with IMAP `userid@domain.tld ` for authentication

Are you talking about the IMAP user back-end? That is not provided by this app but https://github.com/nextcloud/user_external.

> Mail app auto sets up their mailbox.

It does not unless you configure that. And even then it crates the automatic account dynamically, so if you change the config it will log the user into the new mail server automatically.

ChristophWurst:
Just to clarify: are you talking about this feature https://github.com/nextcloud/mail/blob/master/doc/admin.md#automatic-account-creation?

anojht:
> > Setup external user support app with IMAP `userid@domain.tld ` for authentication
> 
> Are you talking about the IMAP user back-end? That is not provided by this app but https://github.com/nextcloud/user_external.
> 
> > Mail app auto sets up their mailbox.
> 
> It does not unless you configure that. And even then it crates the automatic account dynamically, so if you change the config it will log the user into the new mail server automatically.

[...]

PR #3009, opened 2020-04-27
Delete all user's accounts when the user is deleted

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/1501

When a Nextcloud user is deleted, the Mail accounts remained. This shall help clean up their data.

- [x] Requires 18+

@nextcloud/mail

### DIFF
--- a/lib/Listener/UserDeletedListener.php
+++ b/lib/Listener/UserDeletedListener.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


lib/Listener/UserDeletedListener.php
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof BeforeUserDeletedEvent)) {
+			// Unrelated
+			return;
+		}
+
+		$user = $event->getUser();
+		foreach ($this->accountService->findByUserId($user->getUID()) as $account) {
```
rullzer, line 56: Would it make sense to move this to a background job?
I'm just thinking about the case where a user gets deleted but has a few huge mailboxes. Could still take some time for this all to get done then.

Tsuroerusu:
My company provides hosted Nextcloud and I thought I would give my two cents:

I think this should be considered a priority issue. The reason is that if one of my customers wish to cancel their account, I am legally obliged to delete their data as per the GDPR. My terms of service state that this happens after 3 days as that is how long my daily backups run. However it would appear that, as of now, some data related to their account remains and that could get me into some hot water.

ChristophWurst:
> My company provides hosted Nextcloud

Are you a customer or partner of Nextcloud GmbH? Then please get in contact with us via the support system.

Tsuroerusu:
> > My company provides hosted Nextcloud
> 
> Are you a customer or partner of Nextcloud GmbH? Then please get in contact with us via the support system.

Nope, I am not, I just wanted to provide my input as to why I think this pull request is important as the present behaviour could have legal ramifications. Keep up the good work! :-)

lib/Listener/UserDeletedListener.php
ChristophWurst: Yeah, once we actually do the full cleanup right away. My current plan is to do it lazily in a background job.

johanehnberg: In that case a design consideration could be to do this for non-existing users rather than marking deleted users? This way, it will retroactively also clean up those that were deleted before the fix and other potential future database inconsistencies.

PR #3012, opened 2020-04-27
Improve important message estimation slightly

ChristophWurst:
* Instead of binary *one overrules all other* approach this uses
  weighted scores so it's always a mix of high estimations and not one
  extreme value.
* Exclude most special mailboxes so the trashbin and similar are not
  marked important.

### DIFF
--- a/lib/Listener/NewMessageClassificationListener.php
+++ b/lib/Listener/NewMessageClassificationListener.php
@@ -25,12 +25,21 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Listener;
 
+use Horde_Imap_Client;
 use OCA\Mail\Events\NewMessagesSynchronized;
 use OCA\Mail\Service\Classification\MessageClassifier;
 use OCP\EventDispatcher\Event;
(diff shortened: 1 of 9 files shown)


lib/Listener/NewMessageClassificationListener.php
```diff
@@ -25,13 +25,22 @@
...
 
 class NewMessageClassificationListener implements IEventListener {
 
+	private const EXEMPT_FROM_CLASSIFICATION = [
+		Horde_Imap_Client::SPECIALUSE_ARCHIVE,
+		Horde_Imap_Client::SPECIALUSE_DRAFTS,
+		Horde_Imap_Client::SPECIALUSE_JUNK,
+		Horde_Imap_Client::SPECIALUSE_SENT,
+		Horde_Imap_Client::SPECIALUSE_TRASH,
+	];
```
jancborchardt, line 42: Good stuff, this will already fix a big part of the issue I see. :)

ChristophWurst:
Obsoleted by #3029 and #3128

Issue #2991, opened 2020-04-24
Mail Sync is stucked


doc75:
This is a split from #2976 to focus on the partial Sync part.
Following #2980 it appears in the warning message that my Sync is stucked.

**Warning**: The test were done with patch #2978, #2979 and #2980 on top of v1.3.3

### Expected behavior
Background job shall end the sync.

### Actual behavior
Every hour I can see that the number of message cached is not moving:  33989

<details>
  <summary>OCA\Mail\Exception\IncompleteSyncException</summary>

```
{
  "reqId": "XKTFd95F3tvowpUdhaGs",
  "level": 2,
  "time": "2020-04-24T22:16:03+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "mail",
  "method": "",
  "url": "--",
  "message": {
    "Exception": "OCA\\Mail\\Exception\\IncompleteSyncException",
    "Message": "Initial sync is not complete for 4:INBOX (33989 of 52533 messages cached).",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/custom_apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php",
        "line": 193,
        "function": "runInitialSync",
        "class": "OCA\\Mail\\Service\\Sync\\ImapToDbSynchronizer",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Mail\\Account"
          },
          {
            "id": 72,
            "__class__": "OCA\\Mail\\Db\\Mailbox"
          }
        ]
      },
      {
        "file": "/var/www/html/custom_apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php",
        "line": 108,
        "function": "sync",
        "class": "OCA\\Mail\\Service\\Sync\\ImapToDbSynchronizer",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Mail\\Account"
          },
          {
            "id": 72,
            "__class__": "OCA\\Mail\\Db\\Mailbox"
          },
          42,
          null,
          false
        ]
      },
      {
        "file": "/var/www/html/custom_apps/mail/lib/BackgroundJob/SyncJob.php",
        "line": 86,
        "function": "syncAccount",
        "class": "OCA\\Mail\\Service\\Sync\\ImapToDbSynchronizer",
        "type": "->",
        "args": [
          {
            "__class__": "OCA\\Mail\\Account"
          }
        ]
      },
      {
[...]

PR #3013, opened 2020-04-28
Fix endless initial sync due to empty partial page

ChristophWurst:
The initial message cache sync estimates the UID range of the next
    message page so it can fetch messages efficiently. There was an
    uncovered edge case where the next page might not return any message,
    simple because there is a larger gap between the existing UIDs.
    Therefore the sync process was aborted as incomplete over and over as it
    could not make any progress and the number of known messages never
    reached the total number of messages on IMAP. This patch changes the
    logic so it retries the next page if the current one is empty. To break
    a possibly infinite recursion there is a check about reaching the
    highest UID reported by the server.

Fixes https://github.com/nextcloud/mail/issues/2991

@doc75 please test https://patch-diff.githubusercontent.com/raw/nextcloud/mail/pull/3013.patch

@nextcloud/mail for review

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -140,19 +140,32 @@ class MessageMapper {
 
 		$query = new Horde_Imap_Client_Fetch_Query();
 		$query->uid();
+		$fetchResult = $client->fetch(
+			$mailbox,
+			$query,
+			[
+				'ids' => new Horde_Imap_Client_Ids($lower . ':' . $upper)
+			]
+		);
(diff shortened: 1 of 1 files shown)


doc75:
@ChristophWurst +1 sync is making progress now . It is taking some time, but progressing every hour.

ChristophWurst:
> @ChristophWurst sync is making progress now . It is taking some time, but progressing every hour.

Progress is the important part. As we saw in the log `33989 of 52533` indicates that it will be done soon as with every attempt it fetches an average of 10k messages. If that user opens the web interface it will load much faster.

Thanks for testing :v:

Issue #2971, opened 2020-04-22
Mail App Unexpected error during account creation


nestorf5:
### Expected behavior
Open email account

### Actual behavior
Cannot connect

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)

### Server configuration

**Operating system**: Debian 9

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.4

**Nextcloud Version:** 18.0.3

#### Client configuration
**Browser:** chrome, edge, firefox

**Operating system:** W10

nestorf5:
No matter what user and/or email account , always says: Unexpected error during account creation.
Tried imap and smtp with and without security (tls, ssl, none)

Operating system: Debian 9
Web server: Apache
Database: MariaDB
PHP version: 7.4.4
Nextcloud Version: 18.0.3

Client:
Same thing with Windows and debian running different browser

ChristophWurst:
> ### Expected behavior
> 
> Tell us what should happen
> ### Actual behavior
> 
> Tell us what happens instead
> ### Mail app
> 
> **Mail app version:** (see apps admin page, e.g. 0.5.3)
> 
> **Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
> ### Server configuration
> 
> **Operating system**: (e.g. Debian 8)
> 
> **Web server:** (e.g. Apache, Nginx,...)
> 
> **Database:** (e.g. MariaDB, SQLite or PostgreSQL)
> 
> **PHP version:** (e.g. 7.0)
> 
> **Nextcloud Version:** (see admin page, e.g. 13.0.2)
> #### Client configuration
> 
> **Browser:** (e.g. Firefox 48)
> 
> **Operating system:** (e.g. Arch Linux)

The idea of this bug template is to use it :wink: 

The Nextcloud mail app offers an extensive logging system to make it easier identifying and tracking down bugs. Please enable [debug mode](https://docs.nextcloud.com/server/stable/developer_manual/general/devenv.html#enabling-debug-mode) and set the log [level to debug](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/logging_configuration.html) in your admin settings. Then, try to reproduce your issue and take another look at `data/nextcloud.log`, `data/horde_imap.log` and `data/horde_smtp.log`.

cptMikky:
[...]

PR #3016, opened 2020-04-28
Add notification callback prototype

cptMikky:
A "how I think this might be done" solution for logging mentioned in https://github.com/nextcloud/mail/issues/2971.

Not mergeable, it's only supposed to represent the general idea.

### DIFF
--- a/lib/IMAP/IMAPClientFactory.php
+++ b/lib/IMAP/IMAPClientFactory.php
@@ -82,6 +82,7 @@ class IMAPClientFactory {
 						'verify_peer' => $this->config->getSystemValueBool('app.mail.verify-tls-peer', true),
 						'verify_peer_name' => $this->config->getSystemValueBool('app.mail.verify-tls-peer', true),
 					],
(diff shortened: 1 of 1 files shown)


jancborchardt:
@nestorf5 @Anarch157a @filipiff @dengste could you test and review this pull request? You were active in the issue this is supposed to fix: **Mail App Unexpected error during account creation #2971**

Thanks a lot!

ChristophWurst:
> Not mergeable, it's only supposed to represent the general idea.

@jancborchardt the intention was never to have anything mergeable but just to demonstrate the concept.

dengste:
@jancborchardt Sorry for being dense, but I'm not familiar with PHP development. I've applied the patch, where and for what should I look? I'm not seeing any difference in the logs.

cptMikky:
@dengste this concept, as is, doesn't extend the logs. It just shows how the notification callback works. If you want to test it, you need to replace `var_dump` with some actual logging code and copy it under each `case STREAM_NOTIFY_...` clause to get results in log.

PR #3027, opened 2020-04-30
Fix query in often contacted sender classifier

ChristophWurst:
We want to compare the recipient type, not the ID.

@nextcloud/mail 

@jancborchardt I think that is the reason why all new messages are important.


ChristophWurst:
@jancborchardt this is the bug that caused all messages to be marked as important

jancborchardt:
Had it running for the last few days and it seemed much better! :)

Issue #2987, opened 2020-04-24
No translated description in the settings


Valdnet:
### Expected behavior
Translated description of the setting.

### Actual behavior
No translated description in the settings.
**Not available for translation on Transifex**.

![obraz](https://user-images.githubusercontent.com/47037905/80219766-39ddf600-8643-11ea-99b5-b75ba8c93429.png)

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)
1.3.3

ChristophWurst:
Attempted fix is at https://github.com/nextcloud/mail/pull/3028. Let's see if the bot will pick up the string now. This will take 1-2 days.

ChristophWurst:
It's there now https://github.com/nextcloud/mail/blob/724956b26077a05b87c81c2cb49e5083e8dbf8c2/l10n/pl.js#L37

Valdnet:
@ChristophWurst Cool. Thanks a lot 😉.

PR #3028, opened 2020-04-30
Fix the missing translation from settings

GretaD:
Ref #2987


src/components/AppSettingsMenu.vue
```diff
@@ -71,6 +66,10 @@ export default {
 	data() {
 		return {
 			loadingAvatarSettings: false,
+			text: t(
```
ChristophWurst, line 69: Would it be possible not to wrap the line here or was that done by the linter?

GretaD: yes, i ran the linter, so its linters fault

ChristophWurst:
once we know this works we can backport

Issue #2745, opened 2020-03-09
Mailvelope doesn't enter a message


Optica1:
### Expected behavior
When selecting 'Write secure mail' using the mailvelope button, I enter the contact and the message.
Next I select encrypt. At this moment the encrypted message should appear in the message section and the contact in the 'to' section.

### Actual behavior
Nothing happens. No message appears, so I have nothing to send. 

I can open encrypted messages in my inbox using mailvelope, so no issue there.

### Mail app
Version 1.1.3
Mailserver is infomaniak.com using IMAP.

### Server configuration

This is setup on infomaniak who use Installatron (multi-platform application installer)

**Operating system**: unknown

**Web server:** Unknown

**Database:** unknown

**PHP version:** 7.2

**Nextcloud Version:** 18.0.1

#### Client configuration
**Browser:** Firefox 73.0.1

**Operating system:** Ubuntu

PR #3052, opened 2020-05-04
Fix mailvelope integration by switching to the API version

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2745

- [x] Send new message
  - [x] Disable link attachments for encrypted email
  - [ ] ~~Disable unencrypted attachments?~~ let's leave them for now
- [x] Read encrypted message
- [x] Reply/forward encrypted message
- [x] Move Mailvelope hint from the settings menu to the composer actions

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -103,13 +103,20 @@
 				@keyup="onInputChanged"
 			/>
 		</div>
-		<div v-if="noReply" class="warning noreply-box">
+		<div v-if="noReply" class="warning noreply-warning">
 			{{ t('mail', 'This message came from a noreply address so your reply will probably not be read.') }}
+		</div>
+		<div v-if="mailvelope.keysMissing.length" class="warning noreply-warning">
+			{{
+				t('mail', 'The following recipients do not have a PGP key: {recipients}.', {
+					recipients: mailvelope.keysMissing.join(', '),
+				})
+			}}
 		</div>
 		<div class="composer-fields">
 			<!--@keypress="onBodyKeyPress"-->
 			<TextEditor
-				v-if="editorPlainText"
+				v-if="!encrypt && editorPlainText"
 				key="editor-plain"
 				v-model="bodyVal"
 				name="body"
@@ -120,7 +127,7 @@
 				@input="onInputChanged"
 			></TextEditor>
 			<TextEditor
-				v-else
+				v-else-if="!encrypt && !editorPlainText"
 				key="editor-rich"
 				v-model="bodyVal"
 				:html="true"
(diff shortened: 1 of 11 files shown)


ChristophWurst:
Bildschirmfoto von 2020-05-05 20-16-02

Bildschirmfoto von 2020-05-05 20-16-33

Big thanks to @toberndo for the many pointers. I finally got this working and I think this looks fabulous!

@nextcloud/mail I changed it so the 'Send' button changes to 'Send unencrypted'/'Encrypt and send' **only when Mailvelope is detected**. And also the option to encrypt will only appear when Mailvelope is there.

ChristophWurst:
> Reply/forward encrypted message

@toberndo will Mailvelope allow our app to ever access the decrypted message body so we could use it for a reply or forward? Would it even make sense? :thinking:

ChristophWurst:
I think the hint makes more sense here instead of the bottom left settings. This is shown when Mailvelope is not there, once it's detected we show the option to encrypt.

Bildschirmfoto von 2020-05-05 21-56-57

toberndo:
> @toberndo will Mailvelope allow our app to ever access the decrypted message body so we could use it for a reply or forward? Would it even make sense? 🤔

Currently the API does not allow that kind of access. It could be possible to implement such a method by explicitly asking the user inside our container if they would like to grant access to the content of the editor, once requested by the mail client. But it's a question if this can be communicated properly and so far we follow the simple message: cleartext never leaves Mailvelope.

What you can do with the API is to implement a reply encrypted of forward encrypted scenario. For this we have the `quotedMail` parameters: https://github.com/mailvelope/mailvelope/blob/master/src/client-API/client-api.js#L116
Here you can provide a PGP message to the editor that will be decrypted and displayed as quoted content in the editor.

It would be great to have also PGP/MIME support (for encrypted attachments). Do you have access to the MIME headers in the frontend to identify such messages?

Another enhancement could be to early warn the user about non-existing public key of the recipient. This is how this looks like in the Mailvelope editor:

Bildschirmfoto 2020-05-06 um 11 02 28

So that you don't get an error message at the end when you click the send button.
This can be implemented with `mailvelope.getKeyring().validKeyForAddress(['info@example.com'])`.

src/components/Composer.vue
```diff
@@ -146,16 +147,26 @@
...
 						@check="editorPlainText = false"
 						@uncheck="editorPlainText = true"
 						>{{ t('mail', 'Enable formatting') }}</ActionCheckbox
 					>
+					<ActionCheckbox
+						v-if="mailvelopeAvailable"
+						:checked="encrypt"
+						@check="encrypt = true"
+						@uncheck="encrypt = false"
+						>{{ t('mail', 'Encrypt the message with Mailvelope.') }}</ActionCheckbox
```
jancborchardt, line 165: Only have this nitpick that it would be nicer and shorter like this:
> Encrypt message with Mailvelope

(No dot at end, cut "the".)

Otherwise beautiful, really great integration! :)

ChristophWurst:
> So that you don't get an error message at the end when you click the send button.
> This can be implemented with `mailvelope.getKeyring().validKeyForAddress(['info@example.com'])`.

Done and works nicely :+1: 

@nextcloud/mail time for a review :)

PR #3067, opened 2020-05-06
Add mark as junk on emails

### LINKED ISSUE #23: Flag messages as spam/ham (not spam) [$15]
An easy and common way to move messages to "Junk" folder or from there back to "Inbox" folder.

Original ticket: #15 

<bountysource-plugin>

---
There is a **[$15 open bounty](https://www.bountysource.com/issues/37490538-flag-messages-as-spam-ham-not-spam?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GretaD:
Fixes #23


ChristophWurst:
I pushed a few fixes

* Let's only use the `$junk` flag and ignore `$notjunk` for now
* Could be my other email clients but my emails had the deprecated `junk` keyword instead of `$junk`. For better interoperability Mail now sets and clears both of these. I hope server won't throw an error, but we'll see.
* Cleaned up our usage of the seen flag. It was negated and used as `unseen` in the interface. On change we had to convert it back. I've now adapted the front-end to also consistently use the IMAP convention.

lib/Model/IMAPMessage.php
```diff
@@ -644,7 +645,10 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
 		$msg->setFlagFlagged(in_array(Horde_Imap_Client::FLAG_FLAGGED, $flags, true));
 		$msg->setFlagSeen(in_array(Horde_Imap_Client::FLAG_SEEN, $flags, true));
 		$msg->setFlagForwarded(in_array(Horde_Imap_Client::FLAG_FORWARDED, $flags, true));
-		$msg->setFlagJunk(in_array(Horde_Imap_Client::FLAG_JUNK, $flags, true));
+		$msg->setFlagJunk(
+			in_array(Horde_Imap_Client::FLAG_JUNK, $flags, true) ||
+			in_array('junk', $flags, true)
```
clementhk, line 650: Where's this flag from?
https://mailarchive.ietf.org/arch/msg/morg/m6dRIwLStfK__7aIHNJtOTPqxfU/

lib/Service/MailManager.php
```diff
@@ -292,17 +291,17 @@ public function flagMessage(Account $account, string $mailbox, int $uid, string
 		}
 
 		// Only send system flags to the IMAP server as other flags might not be supported
-		if (in_array(strtolower("\\$flag"), self::ALLOWED_FLAGS, true)) {
-			try {
-				$normalized = '\\' . strtolower($flag);
+		$imapFlags = self::ALLOWED_FLAGS[$flag] ?? [];
```
clementhk, line 294: Are you sure `strtolower()` can be removed?
I see in the standard that it is `$Junk`, but Horde probably `strtolower()` already?

lib/Service/MailManager.php
```diff
@@ -292,17 +291,17 @@ public function flagMessage(Account $account, string $mailbox, int $uid, string
...
+		try {
+			foreach ($imapFlags as $imapFlag) {
 				if ($value) {
-					$this->messageMapper->addFlag($client, $mb, $uid, $normalized);
+					$this->messageMapper->addFlag($client, $mb, $uid, $imapFlag);
```
clementhk, line 298: `\` is removed from flag?
It's now `$junk` instead of `\$junk`?

lib/Model/IMAPMessage.php
clementhk: Just saw your comment

ChristophWurst: It's the deprecated keyword, but I found that it's still being used on my personal account. I can't say if it's written by the mail server or another client, but I think it makes sense to try to stay compatible with that as well.

lib/Service/MailManager.php
ChristophWurst: yes, because that is what we get from the front-end and there we always use the lower case flags

lib/Service/MailManager.php
ChristophWurst: No, only registered flags start with `\`. `$Junk` is a keyword.

ChristophWurst: To add to the confusion: we do mix the terms flags/keywords here. Ideally we could name `junk` keyword, but that would probably just complicate the code and lead to two otherwise identical code paths.

clementhk: So the old `$normalized = '\\' . strtolower($flag);` didn't make sense?

ChristophWurst: Only for the flags, yeah. Thus is was fine until now where we add the first keyword. If we set `$Junk` as `\$Junk` the server might just raise an error. Or if it does allow the flag then it has a wrong meaning.

ChristophWurst:
- [ ] Needs https://github.com/nextcloud/mail/pull/3075 for greenCI

src/components/Envelope.vue
```diff
@@ -43,7 +49,10 @@
 				data.flags.important ? t('mail', 'Mark unimportant') : t('mail', 'Mark important')
 			}}</ActionButton>
 			<ActionButton icon="icon-mail" @click.prevent="onToggleSeen">{{
-				data.flags.unseen ? t('mail', 'Mark read') : t('mail', 'Mark unread')
+				data.flags.seen ? t('mail', 'Mark unread') : t('mail', 'Mark read')
+			}}</ActionButton>
+			<ActionButton icon="icon-junk" @click.prevent="onToggleJunk">{{
+				data.flags.junk ? t('mail', 'Unmark junk') : t('mail', 'Mark junk')
```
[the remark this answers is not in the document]
GretaD: fine by me

ChristophWurst:
> "Mark _as_ spam" as mentioned would be nicer, but this works too.

Let's have another tiny PR after this :)

GretaD:
rebased and squashed

ChristophWurst:
> rebased and squashed

Mine was squashed already, actually. It was two commits because the second was a mildly related cleanup. I did not want to mix both :)

PR #3071, opened 2020-05-07
Discussion: what's the down side of caching email?

clementhk:
This PR is not to be merged but wonder what's the downside of doing this.
I cannot live without this change now as everything opens for me instantly now, but it is clearly just a hack as some parameters could be changed and doesn't match with the cached version.

But I'm wondering if we are actually using the changeable parameters (flags?) when showing the emails.

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -231,7 +231,11 @@ class MessagesController extends Controller {
 			return $this->enrichDownloadUrl($accountId, $folderId, $messageId, $a);
 		}, $json['attachments']);
 
-		return new JSONResponse($json);
+		$response = new JSONResponse($json);
+		$response->addHeader('Cache-Control', 'private, max-age=7776000');
+		$response->addHeader('Pragma', 'private');
+
+		return $response;
 	}
 
 	/**
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -402,6 +402,7 @@ export default {
 				const unifiedFolder = getters.getUnifiedFolder(folder.specialRole)
 
 				syncData.newMessages.forEach((envelope) => {
+					fetchMessage(accountId, folderId, envelope.id)
 					commit('addEnvelope', {
 						accountId,
 						folderId,


ChristophWurst:
Prefetching makes sense, but IMO we should somehow detect if the browser is busy with other operations and only do this when it idles. We have quite a high number of requests (after page load) already and thus run into the issue where browsers reach their max concurrent connections.

Something related, though on a different level: now that Mail has a database cache and thus knows when messages are *new* it could pre-fetch the messages in cron so the bodies do not have to be downloaded from IMAP when the user comes back and opens the web front-end.

clementhk:
As a hack I only set it to pre-fetch new emails so it is safe for my own usage.
I'm more wondering about commit https://github.com/nextcloud/mail/pull/3071/commits/1ad5183e1eca864defb70d1553e4d0ab977e6587 if it make sense to cache this in the browser?

ChristophWurst:
> I'm more wondering about commit [1ad5183](https://github.com/nextcloud/mail/commit/1ad5183e1eca864defb70d1553e4d0ab977e6587) if it make sense to cache this in the browser?

Yes, that does make sense. IMAP messages are immutable. So the only thing that might change is how we transform them. But that is very theoretical and would only affect dev instances, where one disables caching anyway.

clementhk:
> > I'm more wondering about commit [1ad5183](https://github.com/nextcloud/mail/commit/1ad5183e1eca864defb70d1553e4d0ab977e6587) if it make sense to cache this in the browser?
> 
> Yes, that does make sense. IMAP messages are immutable. So the only thing that might change is how we transform them. But that is very theoretical and would only affect dev instances, where one disables caching anyway.

Thanks!
Do you want this commit into master?

For the pre-fetch queue I'm gonna leave it to you guys as I'm not very sure the best way forward unless you to use the hack for now.
There are a few ways to have max concurrent requests https://github.com/axios/axios/issues/1010 if this is what you mean.

ChristophWurst:
> Thanks!
> Do you want this commit into master?

Yes, please. See https://github.com/nextcloud/mail/blob/7f32cfe6570d112fb748c9a97590e2b9d6e8aafb/lib/Controller/AvatarsController.php#L85 for how we do caching

> For the pre-fetch queue I'm gonna leave it to you guys as I'm not very sure the best way forward unless you to use the hack for now.

Could you open a ticket? I wouldn't spend too much effort on this right now but maybe we can collect some input over time.

> There are a few ways to have max concurrent requests [axios/axios#1010](https://github.com/axios/axios/issues/1010) if this is what you mean.

Partly. I don't want to limit the number of requests in axios as the browser will already limit it. I would rather want something like *only do this if there are 5 or less active request at the moment*.

clementhk:
> > Thanks!
> > Do you want this commit into master?
> 
> Yes, please. See
> 
> https://github.com/nextcloud/mail/blob/7f32cfe6570d112fb748c9a97590e2b9d6e8aafb/lib/Controller/AvatarsController.php#L85
> 
> for how we do caching

I saw that, but I hesitated because 
1) it uses "public" which means non-browser could cache it?
2) "must-revalidate" but there's no Etag?

I might be wrong on both, they're based on a quick search.

clementhk:
I'll open another PR.

Issue #2982, opened 2020-04-24
Background syncs shows full inbox as "new"


ChristophWurst:
### Expected behavior
Background sync should be clever enough to not show any new messages after the initial sync and later on also just the actual diff that recently changed.

### Actual behavior
xxx new messages notification. Like all of the inbox was new. Seen in #2648, but not sure how to trigger.

ChristophWurst:
![Bildschirmfoto von 2020-05-04 14-25-33](https://user-images.githubusercontent.com/1374172/80965631-2d138c00-8e13-11ea-850f-41202d108043.png)

Happens when I reload the page in the individual inbox.

ChristophWurst:
As far as I understand this happens when a sync is dispatched before the first page is loaded, hence the UID array is empty and the server returns *everything* as if all of those message were new

PR #3072, opened 2020-05-07
Optimize the background sync logic

ChristophWurst:
Sooo this is what I whined about in the dev channel today

Bildschirmfoto von 2020-05-07 19-29-55

Long story short: keeping inboxes up to date is a non-trivial task. Current master does too many sync's IMO, so this reduces the number of (unnecessary) roundtrips a bit.

Most noticeable is that when you lose focus and go back to the Mail window/tab it does not trigger a sync. But only periodically in the background.

The background jobs does these consecutive steps:
* Sync all individual inboxes
* Sync priority inbox sections
* Show the notification

So when you, as a user, see the notification and switch to Mail, the messages will be there already.

Fixes https://github.com/nextcloud/mail/issues/2982 for me. There might still be edge cases but this seems to make it a lot better on  my dev instance.

@nextcloud/mail please give this a test :pray:

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -131,7 +131,6 @@ export default {
 		this.bus.$on('delete', this.onDelete)
 		this.bus.$on('shortcut', this.handleShortcut)
 		this.loadMailboxInterval = setInterval(this.loadMailbox, 60000)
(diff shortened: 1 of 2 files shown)


ChristophWurst:
- [ ] Requires https://github.com/nextcloud/mail/pull/3075 for CI

PR #3074, opened 2020-05-07
Open inbox when clicking on account item

### LINKED ISSUE #2731: UX improvement: Show Inbox after clicking on account name
### Feature Request

UX improvement: Show Inbox after clicking on account name

#### Summary

Currently when user clicks on an account name, settings for that account are displayed. Inbox display is definitely more useful and likely more needed than some settings that could be available via click on three dots / settings anyway.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/89721180-ux-improvement-show-inbox-after-clicking-on-account-name?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheTimeWalker:
This changes the account item to open the inbox (or technically the first folder) in the list instead of the settings screen.

The only quirk that it currently has is that the account is also marked as active with the inbox. Not sure why, any suggeestion or edits to fix that are welcome.

Closes #2731

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -27,7 +27,8 @@
 		:icon="iconError"
 		:menu-open.sync="menuOpen"
 		:title="account.emailAddress"
-		:to="settingsRoute"
+		:to="firstFolderRoute"
+		exact="true"
 	>
 		<!-- Color dot -->
 		<AppNavigationIconBullet v-if="bulletColor" slot="icon" :color="bulletColor" />
@@ -87,6 +88,10 @@ export default {
 			type: Object,
 			required: true,
 		},
+		firstFolder: {
+			type: Object,
+			required: true,
+		},
 		isFirst: {
 			type: Boolean,
 			default: false,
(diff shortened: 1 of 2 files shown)


src/components/NavigationAccount.vue
```diff
@@ -78,6 +78,10 @@ export default {
 			type: Object,
 			required: true,
 		},
+		inbox: {
```
ChristophWurst, line 81: let's use `firstFolder` to avoid confusion later :)

ChristophWurst:
> The only quirk that it currently has is that the account is also marked as active with the inbox. Not sure why, any suggeestion or edits to fix that are welcome.

I think it is because the router does not match exactly. You could try with `:exact="true"` https://github.com/nextcloud/nextcloud-vue/blob/e93851f077ea0e2cfe7d0f2d4c11f7a7b66f3975/src/components/AppNavigationItem/AppNavigationItem.vue#L237-L244

src/components/NavigationAccount.vue
TheTimeWalker: I have renamed it to firstFolder (and firstFolderRoute

PR #3097, opened 2020-05-11
Add background database cleanup

ChristophWurst:
Follow-up to https://github.com/nextcloud/mail/pull/3009

Ideally this would all be real foreign keys with cascading delete, but we're not there yet.


ChristophWurst:
@johanehnberg @Tsuroerusu please test. You can download a patch from https://patch-diff.githubusercontent.com/raw/nextcloud/mail/pull/3097.patch

Issue #3066, opened 2020-05-06
Messages in (priority) inbox take up too much vertical space


ChristophWurst:
### Expected behavior
Have a good overview of my incoming messages, but not just the important ones but possibly also from other sections.

### Actual behavior
I see ~20 messages of a mailbox, even on a large screen. For the priority inbox this means I don't even see my favorites. Other messages are hard to spot as I have to scroll down.

This will be even worse once we have #2666 

### Mail app

**Mail app version:**

v1.4+

![Bildschirmfoto von 2020-05-06 09-06-27](https://user-images.githubusercontent.com/1374172/81146765-33bc1380-8f79-11ea-8183-34c941d8218f.png)

cc @GretaD @jancborchardt @ma12-co @skjnldsv

ChristophWurst:
What if we limit the numbers of messages shown for important and favs? Gmail does that as well. They even paginate. But I think we can have a simple collapse/expand logic that will always show the top x messages for important/other. Then you can see if there is anything new there but you'll also see the *other* messages.

skjnldsv:
Or what about a compact mode?
Gmail does it as well no?
I've seen lots of clients have it too :thinking: 
Would just switch to a 44px line with avatar - title - menu ? :shrug:

ChristophWurst:
> 44px line with avatar - title - menu ?

But what would that be for an email? avatar - sender -subject - menu? I don't think we can get that into a single line with the fixed width of our second column with meaningful text.

skjnldsv:
Don't know, would be enough for me :)
Thuunderbird display the email & subject for me ad I manage just fine :shrug: 
PS: I just checked Outlook, they display entries bigger than we do, 80px and up to 130px :scream: 
![Capture d’écran_2020-05-06_10-43-43](https://user-images.githubusercontent.com/14975046/81156687-7fc18500-8f86-11ea-867c-a011456d00d0.png)

jancborchardt:
> What if we limit the numbers of messages shown for important and favs? Gmail does that as well. They even paginate. But I think we can have a simple collapse/expand logic that will always show the top x messages for important/other. Then you can see if there is anything new there but you'll also see the _other_ messages.

Yep, this is what we should do. Limit it to 5 for each section (except the bottom "Other" where we can simply show more to fill up the rest) and then show an entry "Show all important" and "Show all favorites" below that. It could be remembered, so if you want to always see the important ones it will always do that.

[...]

PR #3098, opened 2020-05-12
Make priority inbox sections collapsible

ChristophWurst:
Limit priority inbox sections to five messages. More is shown by expanding the list.

This helps with #3066 but it does not solve the problem. I still can not see *Other messages* without scrolling.

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -3,12 +3,21 @@
 		<transition-group name="list">
 			<div id="list-refreshing" key="loading" class="icon-loading-small" :class="{refreshing: refreshing}" />
 			<Envelope
-				v-for="env in envelopes"
+				v-for="env in envelopesToShow"
 				:key="env.uid"
 				:data="env"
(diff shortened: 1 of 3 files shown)


jancborchardt:
> This helps with #3066 but it does not solve the problem. I still can not see _Other messages_ without scrolling.

I’d say that’s expected, no? Important and favorite are more important than other, so we shouldn’t cut off from those just to be able to show some of "Other"?

ChristophWurst:
https://github.com/nextcloud/mail/issues/3066#issuecomment-624674393

jancborchardt:
> [#3066 (comment)](https://github.com/nextcloud/mail/issues/3066#issuecomment-624674393)

What do you mean by the link? With this pull request, it fully works as expected, no?

ChristophWurst:
It first sounded like you didn't like the collapse logic now

ChristophWurst:
IMO the vertical space is still a lot, but let's leave it for now

PR #3099, opened 2020-05-12
Fix loading when a message is deleted for Priority Inbox

### LINKED ISSUE #3096: Delete action behaving differently in priority inbox
The delete action does nothing in the priority inbox. Or rather it does delete, but there’s no feedback collapse animation as in the unified inbox. A refresh is needed to see the effect.

GretaD:
fixes #3096


GretaD:
fixed here #3109

Issue #3080, opened 2020-05-09
Large inbox folder causes the vanished ID check to send too large of a UID SEARCH command.


MJDSys:
### Expected behavior
My background sync to happen normally without exceptions.  Also a normal (non-forced) sync using the occ command to complete normally.

### Actual behavior
During the normal sync I get an exception thrown: ```Sync failed for 1:INBOX: IMAP error reported by server.```.  Enabling debugging and digging deeper, it appears during the deleted message partial sync, some ~20000 UIDS are checked for existence.  Horde sends the complete list to my IMAP server, which responds with an error saying argument too long (error from horde_log: ```S: 13 BAD Error in IMAP command UID SEARCH: Too long argument (0.001 + 0.001 + 0.001 secs).```).  This causes an exception to be thrown, cancelling my sync.

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)
1.3.4
**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
Custom setup using Dovecot 2.3.4.1 from Debian.

### Server configuration

**Operating system**: (e.g. Debian 8)
Debian 10
**Web server:** (e.g. Apache, Nginx,...)
Nginx
**Database:** (e.g. MariaDB, SQLite or PostgreSQL)
PostgreSQL
**PHP version:** (e.g. 7.0)
7.3
**Nextcloud Version:** (see admin page, e.g. 13.0.2)
18.0.4
#### Client configuration
**Browser:** (e.g. Firefox 48)
Command line
**Operating system:** (e.g. Arch Linux)
Debian 10

I believe the issue is with the vanished UID check using the full list of UIDs from my mailbox.  I think it would be better to chunk the list for each request.  I can try to whip up a PR if this solution is wanted.  A quick workaround I tested was to slice the UID list to 2000 elements to test, and that did allow the sync to progress normally (I realize of course that breaks the deleted mail check).

doc75:
Seems pretty similar to #3024 (even if error message raised by horde is slightly different)

MJDSys:
@doc75 Maybe?  I'll try to whip up the PR (need to get my PHP dev environment setup again), maybe it will help fix several of these issues.  #2929 Has a similar comment as well.

Please close this issue if there is a similar you rather track this bug with.  I thought I managed to check the issues with my searching.

doc75:
[...]

PR #3103, opened 2020-05-13
Compress UID sequences to ranges

ChristophWurst:
For https://github.com/nextcloud/mail/issues/3080


ChristophWurst:
Obsolete as of https://github.com/nextcloud/mail/issues/3080#issuecomment-629006288

PR #3112, opened 2020-05-14
Mozartize the html purifier

ChristophWurst:
TODO
- [ ] Find out why Mozart does not put `HTMLPurifier` and other classes into a namespace.


ChristophWurst:
No progress with https://github.com/coenjacobs/mozart/issues/44, so I'm closing this for now.

PR #3128, opened 2020-05-18
Work around the cold start problem by falling back to rules

ChristophWurst:
We need good training data to create a high quality classifier from the
user's data. If the data set does not contain enough important messages,
the ML would learn that few things are important and thus always develop
a bias towards non-important classification. To solve this we fall back
to a lower quality rule-based classifier that uses the same features as
the ML classifier. The importance is determined by hard-coded thresholds
of the individual features.

As soon as a classifier has been trained it will be used for the eternity.

At some point we could detect that a mailbox shrinks again and remove the previously trained classifier. I'm not sure about that yet. It should not matter for 99% of our users.

@nextcloud/mail as discussed


jancborchardt:
How can the cold start best be tested? Removing one of the tables, or is there a reset command?

ChristophWurst:
Easiest is to remove on of your account. Then send an email to yourself or wait until some messages arrive. The rules should mark some messages as important. If there are >30 messages the ML will take over. You should not be able to tell when that happens, actually :)

PR #3129, opened 2020-05-18
Paginate all priority inbox sections

ChristophWurst:
For the first design we said pagination is only necessary for *Other messages* but as the number of messages in an inbox almost always grows this is not a real solution. The collapse thingy recently added helps with the vertical space consumed, but as soon as you expand the browser has to render lots of data, making the whole interface lag for a few secs.

@jancborchardt  and I decided to do pagination. Now *Important* and *Favorites* will have a *Load more* button, *Others* will still load more through scrolling.

TODO
- [x] Bug with hiding the button when the list end is reached -> button is hidden before already

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -27,7 +27,7 @@
 		<transition-group name="list">
 			<div id="list-refreshing" key="loading" class="icon-loading-small" :class="{refreshing: refreshing}" />
 			<Envelope
-				v-for="env in envelopesToShow"
+				v-for="env in envelopes"
 				:key="env.uid"
 				:data="env"
 				:folder="folder"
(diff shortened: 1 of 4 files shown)


jancborchardt:
> Bug with hiding the button when the list end is reached -> button is hidden before already

Still seem to get this bug, sometimes I press "Load more" and no new results are loaded. Also it’s collapsed and _then_ new things are loaded, which is a bit strange. Instead the text could change to "Loading" and show a spinner so the list is not collapsed and then expanded again?

ChristophWurst:
> Still seem to get this bug, sometimes I press "Load more" and no new results are loaded. Also it’s collapsed and _then_ new things are loaded, which is a bit strange. Instead the text could change to "Loading" and show a spinner so the list is not collapsed and then expanded again?

The problem is that I could not figure out why it's behaving like that.

ChristophWurst:
@jancborchardt @GretaD please try again. I think the new logic should cover the edge cases as well. Also the *Load more* element will be hidden while the code is loading more. The loading icon will show up instead.

PR #3140, opened 2020-05-20
Add tooltip for important filter

### LINKED ISSUE #3134: Add explanation about priority inbox's important section
As a user I want to know what this important flag means and why it's added by the system. This should go next to the section title and open on hover or click.

GretaD:
fixes #3134 
info1

----
info2

a simple ~~div~~ popover to add the tooltip, @jancborchardt please send me a more accurate text for the tooltip :) @ChristophWurst ~~I hope you're fine with the div ;)~~ i changed to popover

### DIFF
--- a/src/components/MailboxMessage.vue
+++ b/src/components/MailboxMessage.vue
@@ -19,7 +19,18 @@
 					:bus="bus"
 				/>
 				<template v-else>
-					<SectionTitle class="app-content-list-item important" :name="t('mail', 'Important')" />
+					<li class="app-content-list-item">
+						<SectionTitle class="important" :name="t('mail', 'Important')" />
+						<Popover trigger="hover focus">
+							<button slot="trigger" class="button icon-info"></button>
+							{{
+								t(
+									'mail',
+									'Messages will automatically be marked as important based on which messages you interacted with or marked as important. In the beginning you might have to manually change the importance to teach the system, but it will improve over time.'
+								)
+							}}
+						</Popover>
+					</li>
 					<Mailbox
 						class="nameimportant"
 						:account="unifiedAccount"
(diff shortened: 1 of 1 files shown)


jancborchardt:
Proposal for a text:

> Messages will automatically be marked as important based on what you interact with and what you mark as important. You can manually mark or unmark messages as important to further train the algorithm.

What do you think @ChristophWurst @GretaD, does this properly describe what is happening?

ChristophWurst:
> You can manually mark or unmark messages as important to further train the algorithm.

"The system will improve over time. In the beginning you might have to manually change the importance a few types, but the algorithm will learn your patterns."

jancborchardt:
Adjusted based on that, what do you think?

> Messages will automatically be marked as important based on which messages you interacted with or marked as important. In the beginning you might have to manually change the importance to teach the system, but it will improve over time.

GretaD:
> * [x]  I thought it was going to be a popover instead of a tooltip? Much nicer when it’s clickable rather than an instant hover
> * [x]  The button is too present, it shouldn’t have a background, only the icon.
> * [x]  There is a lot of space below the icon until the first mail


@jancborchardt 
info-info

jancborchardt:
- [ ] Now the button doesn’t have the proper needed padding anymore (44*44px min size)
- [ ] As a result it’s not possible to focus via keyboard tab
- [ ] Doesn’t give any feedback on hover or focus
- [ ] Is not vertically aligned with the text, could be a few pixels further up

ma12-co:
> Popover is possibly too wide for mobile but can’t test because it is impossible to go back to the message list on mobile

@jancborchardt  in theory the popover should not overflow the available space, so it should resize itself to fit any screen

GretaD:
> > Popover is possibly too wide for mobile but can’t test because it is impossible to go back to the message list on mobile
> 
> @jancborchardt in theory the popover should not overflow the available space, so it should resize itself to fit any screen

thats exactly what happens. I tested it

jancborchardt:
> > Seems to open it on hover again instead of on click/tap, but fine with that
> 
> But you specifically told me not to shown when clicked but only when hovered and focused.
> I fixed the tab thingy. For some reasons it doesnt show the background color when focus on tab but we can fix that with another PR. I would say to merge this one for 1.4V

Sure, works :)

Regarding the click vs hover, I said above in the first review:

> Much nicer when it’s clickable rather than an instant hover

Or where did the miscommunication come from?

PR #3144, opened 2020-05-21
Allow toggling subscription status for mailboxes

dehnhardt:
Signed-off-by: Holger Dehnhardt <holger@dehnhardt.org>
This PR adds the ability to change the subscription status for each non virtual folder.
A littlebit problematic is the synchronisation because we need to change 
* the IMAP-Folder
* the database cache
* the folder attributes in the vue store

It all works, but it seems a bit complicated to me. Is there a way to reload the 'Folder' object within the NavigationFolder?


dehnhardt:
I do not see what the error has to do with my changes:  "The command "make start-smtp-docker" failed and exited with 2 during ."

lib/Controller/FoldersController.php
```diff
@@ -234,4 +234,21 @@ public function create(int $accountId, string $name) {
...
+	 * @throws ServiceException
+	 */
+	public function subscribeFolder(int $accountId, string $folderId, bool $subscribed) {
+		$account = $this->accountService->find($this->currentUserId, $accountId);
+
+		if (empty($accountId) || empty($folderId)) {
+			return new JSONResponse(null, Http::STATUS_BAD_REQUEST);
+		}
+
+		return new JSONResponse($this->mailManager->subscribeFolder($account, base64_decode($folderId), $subscribed));
```
ChristophWurst, line 252: See https://github.com/nextcloud/mail/pull/3144/checks?check_run_id=697235572. This returns `void`, so it can't be used for the response data. You could just call the method and afterwards return a response with `null` as data.

Alternatively, and more tolerant to faults is that this `subscribeFolder` returns the new subscription state. Like we could transport back if the change was not performed by the IMAP server for whatever reason.

dehnhardt: Now the subscription status is returned from backend subscribeFolder method and the frontend is updated accordingly.

appinfo/routes.php
```diff
@@ -74,6 +74,11 @@
 			'url' => '/api/accounts/{accountId}/folders/{folderId}/stats',
 			'verb' => 'GET'
 		],
+		[
+			'name' => 'folders#subscribeFolder',
+			'url' => '/api/accounts/{accountId}/folders/{folderId}/subscribeFolder',
```
ChristophWurst, line 79:
```suggestion
			'url' => '/api/accounts/{accountId}/folders/{folderId}/subscription',
```

This is a bit cosmetic/idealistic. But REST routes should be about resources, not actions. This goes more into that direction in my opinion. What do you think?

lib/Contracts/IMailManager.php
```diff
@@ -126,4 +126,14 @@ public function markFolderAsRead(Account $account, string $folderId): void;
...
+
+	/**
+	 * @param Account $account
+	 * @param string $mailbox
+	 * @param bool $subscribed
+	 *
+	 * @throws ClientException
+	 * @throws ServiceException
+	 */
+	public function subscribeFolder(Account $account, string $mailbox, bool $subscribed): bool;
```
ChristophWurst, line 138:
```suggestion
	public function updateSubscription(Account $account, string $mailbox, bool $subscribed): bool;
```

Again, rather cosmetic than a functional change. But this name makes it clear that both *subscribe* and *unsubscribe* are handled there. `subscribeFolder` sounds a bit to specific to me.

lib/Controller/FoldersController.php
ChristophWurst: The constructor of `JSONResponse` expects an array. can we change the boolean to something like

```
[
    'subscribed' => true,
]
```
?

lib/Service/MailManager.php
```diff
@@ -282,6 +282,37 @@ public function markFolderAsRead(Account $account, string $folderId): void {
...
+			throw new ClientException("Mailbox $mailbox does not exist", 0, $e);
+		}
+		try {
+			$client->subscribeMailbox($mailbox, $subscribed);
+		} catch (Horde_Imap_Client_Exception $e) {
+			throw new ServiceException("Could not set subscription status for mailbox $mailbox on IMAP: " . $e->getMessage(), $e->getCode(), $e);
+		}
+		$folders = $this->folderMapper->getFolders($account, $client, $mailbox);
+		if (count($folders) !== 1) {
+			throw new ServiceException("Could not identify IMAP Folder: " . $e->getMessage(), $e->getCode(), $e);
```
ChristophWurst, line 310: `$e` is not defined. There is no previous exception.

src/components/NavigationFolder.vue
```diff
@@ -69,6 +69,15 @@
...
 					{{ t('mail', 'Clear locally cached data, in case there are issues with synchronization.') }}
 				</ActionButton>
+
+				<ActionCheckbox
+					v-if="notVirtual"
+					:checked="isSubscribed"
+					:disabled="changeSubscription"
+					@update:checked="subscribeFolder"
+				>
+					{{ t('mail', 'Subscribed') }}
```
ChristophWurst, line 79: Should it just be `Subscribe`? Not sure :thinking:

src/components/NavigationFolder.vue
```diff
@@ -263,6 +278,24 @@ export default {
...
+				})
+				.then(() => {
+					this.changeSubscription = false
+					logger.info('folder subscribed: ' + subscribed)
+				})
+				.catch((error) => {
+					logger.error('could not update subscription', {error})
+					this.changeSubscription = false
+					throw error
+				})
```
ChristophWurst, line 297: this is fine as it is. You might improve this code a bit with `async` and `await`. Then you can do a `try`, `catch` and `finally`, where you do the cleanup step `this.changeSubscription = false` in the `finally block` to remove the duplication.

You may leave the code as it is as well :)

src/store/mutations.js
```diff
@@ -217,4 +217,20 @@ export default {
 	removeMessage(state, {accountId, folderId, id}) {
 		Vue.delete(state.messages, normalizedMessageId(accountId, folderId, id))
 	},
+	subscribeFolder(state, {accountId, folderId, isSubscribed}) {
```
ChristophWurst, line 220: again I would favor `changeFolderSubscription` or similar

src/store/mutations.js
```diff
@@ -217,4 +217,20 @@ export default {
...
+		console.info('modifyFolderSubscription')
```
ChristophWurst, line 221: I assume this is merely for debugging. please remove it :)

appinfo/routes.php
dehnhardt: Sounds reasonable, will change it.

lib/Contracts/IMailManager.php
dehnhardt: I alway struggle with names;-) Will change it

lib/Service/MailManager.php
dehnhardt: Ok. I still missed that. Sorry...

src/components/NavigationFolder.vue
dehnhardt: The text is next to the checkbox that shows the current status. So I would leave it like this.

src/components/NavigationFolder.vue
dehnhardt: Ok, something to learn for me. Will change that.

src/store/mutations.js
dehnhardt: Ok.

src/store/mutations.js
dehnhardt: Ok.

dehnhardt:
@ChristophWurst: All should be fixed/changed now. (Hopefully...)

lib/Service/MailManager.php
ChristophWurst: don't worry :)

src/components/NavigationFolder.vue
ChristophWurst: Fine by me :)

lib/Contracts/IMailManager.php
```diff
@@ -128,6 +128,17 @@ public function markFolderAsRead(Account $account, string $folderId): void;
...
+	/**
+	 * @param Account $account
+	 * @param string $mailbox
+	 * @param bool $subscribed
+	 *
+	 * @return array
+	 * @throws ClientException
+	 * @throws ServiceException
+	 */
+	public function updateSubscription(Account $account, string $mailbox, bool $subscribed): array;
```
ChristophWurst, line 140: I'm sorry for the confusion. But here the `bool` return type is preferred. We should only change the response data structure in the controller.

lib/Service/MailManager.php
```diff
@@ -285,6 +285,38 @@ public function markFolderAsRead(Account $account, string $folderId): void {
...
+		} catch (Horde_Imap_Client_Exception $e) {
+			throw new ServiceException("Could not set subscription status for mailbox $mailbox on IMAP: " . $e->getMessage(), $e->getCode(), $e);
+		}
+		$folders = $this->folderMapper->getFolders($account, $client, $mailbox);
+		if (count($folders) !== 1) {
+			throw new ServiceException("Could not identify IMAP Folder $$mailbox");
+		}
+		$attrributes = $this->mailboxSync->updateMailboxAttributes($folders[$mailbox], $mb);
+		$subscribed = in_array('\subscribed', $attrributes);
+		return ['subscribed' => $subscribed];
```
ChristophWurst, line 317: this is what should go into the controller :)

src/store/actions.js
```diff
@@ -197,6 +202,17 @@ export default {
 			})
 		)
 	},
+	changeFolderSubscription({commit}, {accountId, folderId, subscribed}) {
+		logger.debug('Toggle subscription for account: ' + accountId + ' folder ' + folderId + ' to ' + subscribed)
+		return changeFolderSubscription(accountId, folderId, subscribed).then((res) => {
+			var isSubscribed = res.subscribed
```
ChristophWurst, line 208: you can use `const`. But actually it would be also easier if you use `isSubscribed: res.subscribed` below

src/components/NavigationFolder.vue
```diff
@@ -263,6 +278,22 @@ export default {
...
+				await this.$store.dispatch('changeFolderSubscription', {
+					accountId: this.account.id,
+					folderId: this.folder.id,
+					subscribed: subscribed,
+				})
+			} catch (error) {
+				logger.error('could not update subscription', {error})
+				throw error
+			} finally {
+				this.changeSubscription = false
```
ChristophWurst, line 294: awesome :)

src/store/actions.js
dehnhardt: Great, did not know it works this way.

dehnhardt:
@ChristophWurst: Thanks for all your tipps, I hope it's ok now;-)

ChristophWurst:
Yes, that looks great now. After an exhausting week I do not trust my reviewer's :eyes: anymore, though, so I'll leave the integration for next week :wink: 

Thanks a lot!

src/store/mutations.js
```diff
@@ -217,4 +217,19 @@ export default {
 	removeMessage(state, {accountId, folderId, id}) {
 		Vue.delete(state.messages, normalizedMessageId(accountId, folderId, id))
 	},
+	changeFolderSubscription(state, {accountId, folderId, isSubscribed}) {
+		const uid = normalizedFolderId(accountId, folderId)
+		const folder = state.folders[uid]
+		var attributes = folder.attributes
```
ChristophWurst, line 223: can be `const` because you mutate the variable only internally, but do not assign a new value

dehnhardt:
@ChristophWurst This was a merge hell. Rebases to resolve the merge conflict and there were many... The remaining ESLint errors are out of my scope. Can you please have a look at it? 
I will squash afterwards. ;-) 
Thanks Holger

dehnhardt:
I need help with squashing here: My frst attempt of an interactive rebase lead to missing changes. As I id a rebase on head yesterday I followed this howto https://blog.oddbit.com/post/2019-06-17-avoid-rebase-hell-squashing-wi/ - which wors wonderful but does not reduce the commits :-(
No an interactive rebase does not change anything.
My solution would be to open a new pull request, but then all conversation is lost.
Are there any other solutions?

PR #3155, opened 2020-05-26
Migrate from php-ml to rubix/ml

ChristophWurst:


lib/Service/Classification/ImportanceClassifier.php
```diff
@@ -350,19 +365,10 @@
 		 * Ref https://en.wikipedia.org/wiki/F1_score
 		 */
 		$this->logger->debug("classification report: " . json_encode([
-			'recall' => $report->getRecall(),
-			'precision' => $report->getPrecision(),
-			'f1Score' => $report->getF1score(),
+			'recall' => $recallImportant,
```
nickvergessen, line 368: Is this correct? You use the same var twice?

nickvergessen, line 368: $precisionImportant?

ChristophWurst: right. fixed :)

ChristophWurst:
CI fails because of https://github.com/nextcloud/mail/pull/3160

Issue #2989, opened 2020-04-24
Subject not considered for search


ChristophWurst:
### Expected behavior
"subject:test" should give me messages that partly match the string

### Actual behavior
Nothing

PR #3156, opened 2020-05-27
Add search by subject

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2989

To reproduce try to search with `subject:something` where `something` is part of a known email in the current mailbox. On master you won't get any matching result. Here you will.

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -322,6 +322,16 @@ class MessageMapper extends QBMapper {
 			);
 		}
 
+		if ($query->getSubject() !== null) {
+			$select->andWhere(
+				$qb->expr()->iLike(
+					'subject',
+					$qb->createNamedParameter('%' . $query->getSubject() . '%', IQueryBuilder::PARAM_STR),
+					IQueryBuilder::PARAM_STR
+				)
+			);
+		}
+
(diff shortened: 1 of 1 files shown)


lib/Db/MessageMapper.php
```diff
@@ -322,6 +322,16 @@ public function findUidsByQuery(Mailbox $mailbox, SearchQuery $query, ?int $limi
 			);
 		}
 
+		if ($query->getSubject() !== null) {
+			$select->andWhere(
+				$qb->expr()->iLike(
+					'subject',
+					$qb->createNamedParameter('%' . $query->getSubject() . '%', IQueryBuilder::PARAM_STR),
```
rullzer, line 329: Just thinking... should subject be escaped?

ChristophWurst: I guess. But I naively assumed that our db abstraction does that with every parameter. Is that not the case? Checking …

nickvergessen: Well depends, do you want people to be able to use placeholders?
Should they search for "he_lo" or "healo", "heblo", ... when they type "he_lo"?
If the first one, you need to escape, if the second one, dont

PR #3158, opened 2020-05-27
Display account quota

### LINKED ISSUE #3062: Display IMAP quota
### Feature Request

As a user I want to know how much storage my emails are taking up and how much more space my server allows me to consume. Should go into the account's actions menu or the settings. cc @jancborchardt 

#### Summary

n/a

GretaD:
Fixes #3062 Add quota for mail

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -29,12 +29,16 @@
 		:title="account.emailAddress"
 		:to="firstFolderRoute"
 		:exact="true"
+		@update:menuOpen="onMenuToggle"
 	>
 		<!-- Color dot -->
 		<AppNavigationIconBullet v-if="bulletColor" slot="icon" :color="bulletColor" />
 
 		<!-- Actions -->
 		<template #actions>
+			<ActionText v-if="!account.isUnified" icon="icon-info" :title="t('mail', 'Quota')">
+				{{ quotaText }}
+			</ActionText>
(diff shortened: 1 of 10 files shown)


ChristophWurst:
Some quick research
* Quota is not always supported. There is an IMAP capability `QUOTA` that we have to check. If that is not there we either show nothing or why no quota is shown.
* ``\Horde_Imap_Client_Base::getQuota`` will get us the `usage` and `limit`

src/components/NavigationAccount.vue
```diff
@@ -140,6 +149,19 @@ export default {
 		iconError() {
 			return this.account.error ? 'icon-error' : undefined
 		},
+		quotaText() {
+			if (this.quota === undefined) {
+				return t('mail', 'Loading …')
+			}
+			if (this.quota === false) {
+				return t('mail', 'Not supported by the server')
```
dehnhardt, line 157: I would prefer not to show anything if quota is not supported or not available for any reason

ChristophWurst: Our reasoning would be that this makes it more clear. Like if you were looking for this info as a user and the interface just hides it, you might think that it's hidden somewhere else, while you actually would have discovered it already.

In the end both is fine.

ChristophWurst: If you have strong feeling about this please do not hesitate to open a ticket about it :)

dehnhardt: No, not at all ;-)

PR #3165, opened 2020-05-29
Downgrade mochapack

ChristophWurst:
It breaks our CI, so let's use the old version until that got resolved.


ChristophWurst:
> **Travis CI - Pull Request ** Successful in 8m — Build Passed

PR #3175, opened 2020-06-02
Browser Message Cache

clementhk:
I really need this to be able to use mail app full time, please give input for merging or if anyone can work on something like this on other directions.
I'm open to making this optional in config file.

### DIFF
--- a/src/service/MessageService.js
+++ b/src/service/MessageService.js
@@ -53,7 +53,7 @@ export function fetchEnvelopes(accountId, folderId, query, cursor, limit) {
 			params,
 		})
 		.then((resp) => {
-			resp.data.forEach(message => {
+			resp.data.forEach((message) => {
 				prefetchMessage(accountId, folderId, message.id)
 			})
 			return resp.data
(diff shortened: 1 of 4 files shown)


src/service/MessageService.js
```diff
@@ -52,7 +52,12 @@ export function fetchEnvelopes(accountId, folderId, query, cursor, limit) {
 		.get(url, {
 			params,
 		})
-		.then((resp) => resp.data)
+		.then((resp) => {
+			resp.data.forEach((message) => {
```
ChristophWurst, line 56: would it make sense to queue this and fetch one message after the other?

src/service/MessageService.js
```diff
@@ -87,6 +96,11 @@ export async function syncEnvelopes(accountId, folderId, uids, query, init = fal
 	}
 }
 
+export async function prefetchMessage(accountId, folderId, id) {
+	fetchMessage(accountId, folderId, id)
```
ChristophWurst, line 100: I think you want to `await` both lines here to limit the nr of concurrent requests a bit

clementhk:
> This will currently send off n*2 requests when the initial message page is loaded. Like for the unified inbox that is m*n*2. If you have three accounts (=m) it will load 20 messages each (=n) and therefore cause 3*20*2=120 requests. This will slow down the interface and block any other requests (icons for example) as the browser has to process this.
> 
> I'm fine with prefetching, but this implementation is too aggressive IMO. Let's do it sequentially instead. The browser can then process any other requests interleaved. The UI will feel more responsive for other actions and the UX will be better in general.
> 
> Hope this makes sense ✌️

Prefetching one by one now in a queue, is this what you mean?

PR #3180, opened 2020-06-04
Remove bountysource references

ChristophWurst:
Due to the many issues we've been having with the platform, we should try to keep this off the radar of people new to the project. The problem of existing bounties is still to handle. But we must not advertise this anymore for new tickets.

cc @karlitschek @jospoortvliet @nextcloud/mail


karlitschek:
👍

PR #3227, opened 2020-06-09
Show unread messages in bold

ChristophWurst:
Otherwise all messages look the same.

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -246,6 +246,9 @@ export default {
 	z-index: 50; // same as icon-starred
 }
 
+.app-content-list-item:not(.seen) {
+	font-weight: bold;
+}
 .app-content-list-item.selected {
 	background-color: var(--color-background-dark);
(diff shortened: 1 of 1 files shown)


jancborchardt:
> When did this break, it was fine just last week, nc nc nc.

Found it – this seems to have slipped through in https://github.com/nextcloud/mail/commit/01b01891d7d9d98b0f9d76c5f5b1ba7c2c372722#diff-03f4e6dc240c55861fa235e27445d65bL219

Pull request https://github.com/nextcloud/mail/pull/2619 by @StCyr which @ChristophWurst and I both reviewed. Oops :D

ChristophWurst:
Yes, I found that as well. I might have accidentally *optimized this away* during rebase or squash :)

GretaD:
> Yes, I found that as well. I might have accidentally _optimized this away_ during rebase or squash :)

what bothers me is that i didnt realize smth was wrong even though i have like 100 unread messages :D

PR #3228, opened 2020-06-09
Fix icon color in dark theme

### LINKED ISSUE #3224: Folder icons in dark mode
### Expected behavior
Icons are visible

### Actual behavior
Icons are black

Bildschirmfoto von 2020-06-09 10-17-16

### Mail app

**Mail app version:**
1.4.0-rc2 I think

GretaD:
fixes #3224 
// The problem is caused because of this regression here: https://github.com/nextcloud/server/blob/892589746dd0f758358adbc66a1bd729a8374e55/lib/private/Template/IconsCacher.php#L198 
@skjnldsv should we maybe change it to check only for fill and not fill-rule?

### DIFF
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -318,31 +318,18 @@
 }
 
 /* icons */
-.icon-inbox {
-	@include icon-color('inbox', 'mail', $color-black);
-}
-.icon-important {
-	@include icon-color('important', 'mail', $color-black);
-}
+@include icon-black-white('inbox', 'mail', 1);
+@include icon-black-white('important', 'mail', 1);
+
 .icon-flagged {
 	@include icon-color('star', 'mail', $color-black);
 }
-.icon-drafts {
-	@include icon-color('drafts', 'mail', $color-black);
-}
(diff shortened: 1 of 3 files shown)


css/mail.scss
```diff
@@ -318,31 +318,18 @@
...
-}
-.icon-important {
-	@include icon-color('important', 'mail', $color-black);
-}
+@include icon-black-white('inbox', 'mail', 1);
+@include icon-black-white('important', 'mail', 1);
+
 .icon-flagged {
 	@include icon-color('star', 'mail', $color-black);
 }
```
GretaD, line 326: this one doesnt work as `@include icon-black-white('star', 'mail', 1);` is it maybe because in some places it is referred as flagged and in come as star? @skjnldsv ?

GretaD:
can we open a new issue for that please and lets merge this one?

jancborchardt:
> can we open a new issue for that please and lets merge this one?

Sure :) opened "Further dark theme fixes #3231" and let’s get this in!

PR #3234, opened 2020-06-10
Fix inconsistend wording Spam vs Junk

### LINKED ISSUE #3225: Mark as junk vs mark as spam
The label for marking an email as spam is inconsistent.
In the email list its called "Mark as spam":
F90F5795-ACF8-4D8E-BC71-634920054E0A-2

In the message view it's called "Mark junk":
A2F2634D-6337-4853-A9D5-26AA07293210-2

georgehrke:
fixes #3225

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -44,7 +44,7 @@
 						</ActionButton>
 
 						<ActionButton icon="icon-junk" @click="onToggleJunk">
-							{{ envelope.flags.notjunk ? t('mail', 'Mark not junk') : t('mail', 'Mark junk') }}
+							{{ envelope.flags.junk ? t('mail', 'Mark not spam') : t('mail', 'Mark as spam') }}
 						</ActionButton>
 						<ActionButton
 							:icon="sourceLoading ? 'icon-loading-small' : 'icon-details'"


src/components/Message.vue
```diff
@@ -44,7 +44,7 @@
 						</ActionButton>
 
 						<ActionButton icon="icon-junk" @click="onToggleJunk">
-							{{ envelope.flags.notjunk ? t('mail', 'Mark not junk') : t('mail', 'Mark junk') }}
+							{{ envelope.flags.notjunk ? t('mail', 'Mark not spam') : t('mail', 'Mark as spam') }}
```
georgehrke, line 47: Shouldn't this also be `.flags.junk` like in 
https://github.com/nextcloud/mail/blob/fcf2fd0b444af2f8fb9b03fff18d324f5d10ed8e/src/components/Envelope.vue#L59

I could find occurrences of `notjunk` in the server, but this is the only place where its used in js/vue.

ChristophWurst: > Shouldn't this also be `.flags.junk` like in

Yes it should. Though technically both keywords might be used on IMAP …

georgehrke: Fixed

ChristophWurst: I think would also explain why @blizzz experienced that the Junk state does not change as you click the button. Because the `notjunk` is not there, so it's not a reactive object in Vue terms

ChristophWurst: Yep, this fixes the reactivity bug :)

PR #3247, opened 2020-06-13
Set file picker to copy mode.

### LINKED ISSUE #2922: 1.3.1 the ability to create new folder while saving attachments has disappeared
### Expected behavior
while saving attachments to server, it was possible to create new folder

### Actual behavior
that has disappeared since 1.3.1

### Mail app

**Mail app version:** 1.3.1
**Mailserver or service:** Hetzner

### Server configuration

**Operating system**: Debian 9

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.3.2
**Nextcloud Version:** 18.0.3
#### Client configuration
**Browser:** Firefox 75

**Operating system:** Mac Catalina 10.15.3
Screenshot 2020-04-16 at 17 02 19

miceno:
Fixes #2922 by using the file picker using COPY type.

### DIFF
--- a/src/components/MessageAttachment.vue
+++ b/src/components/MessageAttachment.vue
@@ -148,6 +148,7 @@ export default {
 				.addMimeTypeFilter('httpd/unix-directory')
 				.setModal(true)
 				.setType(1)
+				.allowDirectories(true)
 				.build()
 
 			return picker
--- a/src/components/MessageAttachments.vue
+++ b/src/components/MessageAttachments.vue
@@ -81,6 +81,7 @@ export default {
 				.addMimeTypeFilter('httpd/unix-directory')
 				.setModal(true)
 				.setType(1)
+				.allowDirectories(true)
 				.build()
 
 			const saveAttachments = (id) => (directory) => {


miceno:
Rerun

ChristophWurst:
The issue with this is that the file picker button will now say `Copy` instead of `Choose`, right?

miceno:
Yes ;) In fact, you can say it is copying from email to nextcloud :D

--
“If you think that hiring a professional to do the job is expensive, try hiring an amateur." - Orestes Sanchez

ChristophWurst:
> The issue with this is that the file picker button will now say `Copy` instead of `Choose`, right?

soooooo what can we do about this?

miceno:
I am not too familiar with Nextcloud code, but I think the way to solve it is by changing it back to a Chooser filepicker, and allowing directories option, like this:

```js
const picker = getFilePickerBuilder(t('mail', 'Choose a folder to store the attachment in'))
				.setMultiSelect(false)
				.addMimeTypeFilter('httpd/unix-directory')
				.setModal(true)
				.setType(1)
				.allowDirectories(true)
				.build()
```
I will update the PR if you feel it is the way to solve it.

ChristophWurst:
Sure we can try that :)

miceno:
Thank you for your feedback!

I will need 3-4 days for it, I am busy these days. I will come back later.

ChristophWurst:
No rush. Take your time :)

miceno:
I have updated the code.

ChristophWurst (approved):
Awesome!

Issue #3058, opened 2020-05-05
Renaming folders


ChristophWurst:
### Feature Request

A user should be able to rename mailboxes.

#### Summary

n/a

jancborchardt:
Design spec:

- In the actions menu of a folder, directly below "Mark all as read" can be an action "Rename folder"
- Icon can be `icon-rename`
- Clicking it will replace the entry with an input field (just like the proposal for adding a folder at https://github.com/nextcloud/mail/issues/2798)
- The input field should have the current folder name prefilled, completely selected so you can choose to adjust the name, or just start typing to replace it.

PR #3260, opened 2020-06-15
Add rename mailbox option

GretaD:
Fixes https://github.com/nextcloud/mail/issues/3058

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -50,7 +50,6 @@
 					@click="markAsRead">
 					{{ t('mail', 'Mark all messages of this mailbox as read') }}
 				</ActionButton>
-
 				<ActionButton
 					v-if="!editing && top && !account.isUnified && mailbox.specialRole !== 'flagged'"
 					icon="icon-folder"
@@ -58,6 +57,17 @@
 					{{ t('mail', 'Add subfolder') }}
 				</ActionButton>
 				<ActionInput v-if="editing" icon="icon-folder" @submit.prevent.stop="createMailbox" />
+				<ActionButton
+					v-if="renameLabel && !hasSubMailboxes && !account.isUnified"
+					icon="icon-rename"
+					@click.prevent.stop="openRenameInput">
+					{{ t('mail', 'Edit name') }}
+				</ActionButton>
+				<ActionInput
+					v-if="renameInput"
+					icon="icon-rename"
+					:value.sync="mailboxName"
+					@submit.prevent.stop="renameMailbox" />
 				<ActionText v-if="showSaving" icon="icon-loading-small">
 					{{ t('mail', 'Saving') }}
 				</ActionText>
@@ -101,6 +111,7 @@ import { getMailboxStatus } from '../service/MailboxService'
 import logger from '../logger'
 import { translatePlural as n } from '@nextcloud/l10n'
 import { translate as translateMailboxName } from '../i18n/MailboxTranslator'
+import { showInfo } from '@nextcloud/dialogs'
 
 export default {
 	name: 'NavigationMailbox',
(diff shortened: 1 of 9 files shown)


ChristophWurst:
What was the status here? Please add this info to the description of your PRs so others get a better understanding.

If everything is done from your side please let me know and I'll rebase to latest master.

src/components/NavigationMailbox.vue
```diff
@@ -317,6 +333,34 @@ export default {
 				}
 			)
 		},
+		async renameMailbox(event) {
+			this.renameInput = false
+			this.showSaving = false
+
+			const newName = event.target.querySelector('input[type=text]').value
```
ChristophWurst, line 340: can we use two-way binding instead to a `data` prop? https://vuejs.org/v2/guide/forms.html

src/service/MailboxService.js
```diff
@@ -46,3 +46,12 @@ export const deleteMailbox = async(id) => {
 
 	await axios.delete(url)
 }
+export async function patchMailbox(accountId, mailboxId, data) {
+	const url = generateUrl('/apps/mail/api/accounts/{accountId}/mailboxes/{mailboxId}', {
```
ChristophWurst, line 50:
```suggestion
	const url = generateUrl('/apps/mail/api/mailboxes/{mailboxId}', {
```

not 100% sure but I think you're using an outdated URL format. Check `appinfo/routes.php` for comparison.

src/store/mutations.js
```diff
@@ -218,4 +218,19 @@ export default {
 	deleteAlias(state, { account, alias }) {
 		account.aliases.splice(account.aliases.indexOf(alias), 1)
 	},
+	renameMailbox(state, { accountId, mailbox, newName }) {
+		const oldId = mailbox.id
+
+		// Remove the old entry
+		Vue.delete(state.mailboxes, mailbox.databaseId(accountId, oldId))
```
ChristophWurst, line 225: `mailbox.databaseId` is an `int`. You're using it as function.

src/store/mutations.js
```diff
@@ -218,4 +218,19 @@ export default {
...
+		// Update the ID
+		mailbox.id = btoa(newName)
```
ChristophWurst, line 227: we don't need bas64 encoding anymore

src/store/mutations.js
```diff
@@ -218,4 +218,19 @@ export default {
...
+		// Set the new entry
+		Vue.set(state.mailbox, mailbox.databaseId(accountId, mailbox.id), mailbox)
+
+		// Update the reference
+		const oldNormalized = mailbox.databaseId(accountId, oldId)
```
ChristophWurst, line 232: again, this makes no sense

src/store/mutations.js
```diff
@@ -218,4 +218,19 @@ export default {
...
+		state.accounts[accountId].mailboxes = state.accounts[accountId].mailboxes.filter(id => id !== oldNormalized)
+		state.accounts[accountId].mailboxes.push(mailbox.databaseId(accountId, mailbox.id))
```
ChristophWurst, line 234: same here

src/store/mutations.js
GretaD: oh right, i just copy paste to replace normalizedFolderId

src/components/NavigationMailbox.vue
```diff
@@ -43,21 +43,31 @@
 				</ActionText>
 
 				<ActionButton
-					v-if="mailbox.specialRole !== 'flagged' && !account.isUnified"
```
ChristophWurst, line 46: was this intentional?

ChristophWurst (approved):

src/components/NavigationMailbox.vue
GretaD: no, fixed it

PR #3263, opened 2020-06-16
Make the importance classifier training more debuggable

ChristophWurst:
There are production systems where the ML training produces a model that works rather poorly. This should help debug this black box a bit.

This PR does two things
* Add a [decorator](https://en.wikipedia.org/wiki/Decorator_pattern) for the `ILogger` that also logs to the console output
* Add more debug logs to the training procedure


tacruc:
LGTM, lets wait for travis.

ChristophWurst:
> All checks have passed

:)

Issue #3298, opened 2020-06-24
References header not being set on replies


mjog:
### Expected behavior

When replying to a message, Mail should set the `References` header so that other MUAs can reconstruct the threading and display the message as a reply.

See https://cr.yp.to/immhf/thread.html for an indication of what existing mailers typically support.

### Actual behavior

When replying to a message, Mail only sets the `In-Reply-To` header, so other MUAs will not show the reply in the same thread as the original.

### Mail app

**Mail app version:** 1.4.0

**Mailserver or service:** Dovecot

### Server configuration

**Operating system**: Debian 10

**Web server:** Apache

**Database:** MariaDB

**PHP version:** Whatever comes with Debian 10

**Nextcloud Version:** 19.0.0

#### Client configuration
**Browser:** GNOME Web 3.36.2

**Operating system:** Ubuntu 20.04

ChristophWurst:
Thank a lot for this ticket. I was not aware of that header, only the in-reply-to one.

ChristophWurst:
FYI I've added you as a collaborator. Thanks a lot for your detailed reports. I really appreciate those :v:

ChristophWurst:
>             If a message contains a References header line, then use the
>             Message IDs in the References header line as the references.
> 
>             If a message does not contain a References header line, or
>             the References header line does not contain any valid
>             Message IDs, then use the first (if any) valid Message ID
>             found in the In-Reply-To header line as the only reference
>             (parent) for this message.

Found at https://tools.ietf.org/html/rfc5256#section-3 while digging into https://github.com/nextcloud/mail/pull/2125. So yeah, we totally lack this.

ChristophWurst:
PR is at https://github.com/nextcloud/mail/pull/3300

PR #3300, opened 2020-06-24
Add the missing `References` header for sent messages

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3298

cc @nextcloud/mail for review


ChristophWurst:
Bildschirmfoto von 2020-06-24 16-29-04

proof that it works

PR #3306, opened 2020-06-24
Delete old draft after switching accounts

### LINKED ISSUE #2511: Drafts stay in "Drafts" folder when switching accounts
1. Compose a new message
2. Notice you are not on the right account
3. Use the account switcher
4. Send the email

### Actual behavior
The last draft state of the message will stay in the "Drafts" folder of the first account.

### Expected behavior
When you switch the account, the draft is moved from the first account to the account you switched to. Thus when you send the mail, no draft version is left.

OhanaM:
Here is my first attempt on fixing issue #2511. My idea is to "check if change account then if yes delete old draft". I am still working on it to think of a better solution.

Fixes #2511


ChristophWurst:
> I am still working on it to think of a better solution.

If you need any help please let us know :)

ChristophWurst:
closing due to lack of activity

PR #3307, opened 2020-06-25
Fix folder creation design

### LINKED ISSUE #2798: Folder creation design review
Some feedback on the "Add folder" process – also valid for the "Add subfolder" flow:
- [ ] By default, it should only show as normal entry, not directly as input field cause that’s too much noise. Just make it so that on click it converts into the input field and automatically focuses. Then the placeholder can also be "Folder name".
- [ ] A more fitting icon for the "Add folder" entry is `icon-folder`
- [ ] When the folders are collapsed and you add a folder, you don’t see any feedback. So in this case, the folders should automatically expand.
- [ ] Seems like a strange bug possibly connected to the caching: The first folder I created shows the same emails as the inbox, but starting from message number 16 only. But when I click one, it says "Not found: Could not load your message"

GretaD:
Fixes #2798 

- Change icon into icon-folder for

- [x] Folder

- [x] Subfolder

- By default, it should only show as normal entry

- [x] Folder

- [x] Subfolder

- [x] When the folders are collapsed and you add a folder, you don’t see any feedback. So in this case, the folders should automatically expand.

### DIFF
--- a/src/components/NavigationFolder.vue
+++ b/src/components/NavigationFolder.vue
@@ -25,10 +25,12 @@
 		:id="genId(folder)"
 		:key="genId(folder)"
 		:allow-collapse="true"
+		:menu-open.sync="menuOpen"
 		:force-menu="true"
 		:icon="icon"
 		:title="title"
 		:to="to"
+		:open.sync="showSubFolders"
 		@update:menuOpen="onMenuToggle"
 	>
 		<!-- actions -->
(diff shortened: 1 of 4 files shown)


src/components/NavigationFolder.vue
```diff
@@ -133,6 +142,9 @@ export default {
 			folderStats: undefined,
 			loadingMarkAsRead: false,
 			clearingCache: false,
+			addInput: false,
```
ChristophWurst, line 145: instead of three boolean props we can also have one of type string and assign something like `hidden`, `editing` and `saving`

PR #3329, opened 2020-06-27
Hide "all inbox" if unified is enabled but only one exists

### LINKED ISSUE #3285: All Inboxes shown when only one account exists
### Expected behavior

When there is only a single account, show only the account's inbox.

### Actual behavior

When there is only a single account, two inboxes are shown - All Inboxes and the account's inbox, which are the same.

### Mail app

**Mail app version:** 1.4.0

**Mailserver or service:** Dovecot

### Server configuration

**Operating system**: Debian 10

**Web server:** Apache

**Database:** MariaDB

**PHP version:** Whatever comes with Debian 10

**Nextcloud Version:** 19.0.0

#### Client configuration
**Browser:** GNOME Web 3.36.2

**Operating system:** Ubuntu 20.04

TheTimeWalker:
The visible function of an unified inbox now checks if there's only one single account and hides itself if it's true

Fixes #3285

### DIFF
--- a/src/components/NavigationFolder.vue
+++ b/src/components/NavigationFolder.vue
@@ -145,8 +145,9 @@ export default {
 	computed: {
 		visible() {
 			return (
-				this.account.showSubscribedOnly === false
-				|| (this.folder.attributes && this.folder.attributes.includes('\\subscribed'))
+				(this.account.showSubscribedOnly === false
+				|| (this.folder.attributes && this.folder.attributes.includes('\\subscribed')))
+				&& this.isUnifiedButOnlyInbox
 			)
 		},
 		title() {
@@ -203,6 +204,10 @@ export default {
 				}
 			}
 			return t('mail', 'Loading …')
+		},
+		isUnifiedButOnlyInbox() {
+			if (!this.folder.isUnified) return true
+			return this.folder.specialUse.includes('inbox') && this.$store.getters.accounts.length > 2
 		},
 	},
 	methods: {


src/components/NavigationFolder.vue
```diff
@@ -197,6 +198,15 @@ export default {
 			}
 			return t('mail', 'Loading …')
 		},
+		isUnifiedButOnlyInbox() {
+			if (!this.folder.isUnified) return false
+			return (
+				this.folder.specialUse.includes('inbox') &&
+				this.folder.isUnified &&
+				this.$store.getters.accounts[1] ===
```
ChristophWurst, line 206: you could just check if there are more than two accounts :see_no_evil:

TheTimeWalker: Oh my, you're right. Obviously you're not getting it if unified inbox is disabled. I was thinking too far. 🙊

ChristophWurst: Could you adjust your patch? :)

TheTimeWalker: Yes, sorry for the long wait!

TheTimeWalker: I've changed it to check if there's more than two accounts and simplified it a bit by not using `!` and checking `isUnified` double

TheTimeWalker:
There we go, it's now ready again 😃

ChristophWurst (approved):

ChristophWurst:
We recently had a discussion in a design review. This topic was brought up. While we played with the idea of hiding the all inboxes, we think that it should remain as an item in the list. Otherwise the first account and the priority inbox would merge together, though they are two separate aspects. There is the general inboxes, like as an abstraction of your IMAP contents, and then there is the raw access to what is stored on IMAP.

Hope this makes some sense :)

TheTimeWalker:
Hi, that's okay! I would suggest adding that answer to the issue if that is by design so it's not staying open as it is "fixed" in that case 😄

ChristophWurst:
Ah yes, thanks!

PR #3337, opened 2020-07-01
Store the `references`, `in-reply-to` and the thread root id

ChristophWurst:
For #17, extracted from https://github.com/nextcloud/mail/pull/2125.

This stores the raw values of `In-Reply-To` and `References` in the database as well as the `Message-ID` of the root element of a message thread. The default is the own (but parsed) message-id, like if the message was its own root. In a follow-up PR I'll finish the algorithm (possibly in #2125) to update this reference if a new root is found. The goal is to have the root as stable as possible and the threading algorithm should make that possible in most cases, we we only have to update this field once.

@nextcloud/mail please review

Steps to test
1) Check out the branch
2) Run the migration
3) See that new messages have these three attributes populated

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -25,16 +25,20 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Db;
 
+use Horde_Mail_Rfc822_Identification;
 use JsonSerializable;
 use OCA\Mail\AddressList;
 use OCP\AppFramework\Db\Entity;
 use function in_array;
+use function json_encode;
 
 /**
  * @method void setUid(int $uid)
  * @method int getUid()
- * @method void setMessageId(string $id)
  * @method string getMessageId()
+ * @method string|null getReferences()
+ * @method string|null getInReplyTo()
+ * @method string|null getThreadRootId()
  * @method void setMailboxId(int $mailbox)
  * @method int getMailboxId()
  * @method void setSubject(string $subject)
@@ -83,6 +87,9 @@ class Message extends Entity implements JsonSerializable {
 
 	protected $uid;
 	protected $messageId;
+	protected $references;
+	protected $inReplyTo;
+	protected $threadRootId;
 	protected $mailboxId;
 	protected $subject;
 	protected $sentAt;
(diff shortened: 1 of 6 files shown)


lib/Db/Message.php
```diff
@@ -134,6 +140,31 @@ public function __construct() {
...
+	public function setInReplyTo(?string $inReplyTo): void {
+		$this->setFieldIfNotEmpty('inReplyTo', $inReplyTo);
+	}
+
+	public function setThreadRootId(?string $messageId): void {
+		$parsed = new Horde_Mail_Rfc822_Identification($messageId);
+		$this->setFieldIfNotEmpty('threadRootId', $parsed->ids[0] ?? null);
+	}
+
+	private function setFieldIfNotEmpty(string $field, ?string $value): void {
```
nickvergessen, line 160: but it does set it (to null) when it is empty ;)

lib/Db/MessageMapper.php
```diff
@@ -116,6 +119,12 @@ public function insertBulk(Message ...$messages): void {
 		foreach ($messages as $message) {
 			$qb1->setParameter('uid', $message->getUid(), IQueryBuilder::PARAM_INT);
 			$qb1->setParameter('message_id', $message->getMessageId(), IQueryBuilder::PARAM_STR);
+			$inReplyTo = $message->getInReplyTo();
+			$qb1->setParameter('in_reply_to', $inReplyTo, $inReplyTo === null ? IQueryBuilder::PARAM_NULL : IQueryBuilder::PARAM_STR);
```
nickvergessen, line 123: can you check this before and use `$qb->expr()->isNull()` instead?

ChristophWurst: I don't get it. Change what?

nickvergessen: Ignore, this is on insert... not comparing

lib/Migration/Version1050Date20200624101359.php
```diff
@@ -0,0 +1,40 @@
...
+		$schema = $schemaClosure();
+
+		$messagesTable = $schema->getTable('mail_messages');
+		$messagesTable->getColumn('message_id')->setLength(1023);
+		$messagesTable->addColumn('references', 'text', [
+			'notnull' => false,
+		]);
+		$messagesTable->addColumn('in_reply_to', 'string', [
+			'notnull' => false,
+			'length' => 1023,
```
rullzer, line 31: I'm assuming this is somehow guaranteed by the RFC?

lib/Model/IMAPMessage.php
```diff
@@ -630,6 +645,9 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
 
 		$msg->setUid($this->getUid());
 		$msg->setMessageId($this->getMessageId());
+		$msg->setReferences($this->getRawReferences());
```
rullzer, line 648: I'm just thinking here if it would make sense to split them already and just store it as json or something? Basically process on input. So that you don't have to do it every time afterwards? But just a suggestion.

lib/Migration/Version1050Date20200624101359.php
ChristophWurst: The length? It's based on https://stackoverflow.com/questions/30079128/maximum-internet-email-message-id-length

lib/Model/IMAPMessage.php
ChristophWurst: yep, I was also thinking about that

ChristophWurst: Let's just do that

ChristophWurst: done

rullzer: But this is the other way around right? Now you still store them flat and process on setting it?

ChristophWurst: They are parsed and set *once* during the insert. What would you like to see different? The layer at which the parsing happens?

rullzer: ah I'm just stupid then

PR #3338, opened 2020-07-02
Revive aliases feature

### LINKED ISSUE #1722: Unable to specify alias
### Actual behavior
Hey,
I'm using the latest version 0.14 and somehow the alias feature has vanished. I'm unable to define or select an alias for my account.

Best regards,
Jonas

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/73573289-unable-to-specify-alias?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F44154351&utm_medium=issues&utm_source=github).
</bountysource-plugin>

IchbinkeinReh:
Fixes #1722

### DIFF
--- a/lib/Db/Alias.php
+++ b/lib/Db/Alias.php
@@ -23,6 +23,8 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Db;
 
+use JsonSerializable;
+
 use OCP\AppFramework\Db\Entity;
 
 /**
@@ -33,7 +35,7 @@ use OCP\AppFramework\Db\Entity;
  * @method void setAlias(string $alias)
  * @method string getAlias()
  */
-class Alias extends Entity {
+class Alias extends Entity implements JsonSerializable {
 
 	/** @var int */
 	protected $accountId;
@@ -48,5 +50,13 @@ class Alias extends Entity {
 		$this->addType('accountId', 'int');
 		$this->addType('name', 'string');
 		$this->addType('alias', 'string');
+	}
+
+	public function jsonSerialize(): array {
+		return [
+			'id' => $this->getId(),
+			'name' => $this->getName(),
+			'alias' => $this->getAlias()
+		];
 	}
 }
--- a/src/components/AliasSettings.vue
+++ b/src/components/AliasSettings.vue
@@ -1,0 +1,136 @@
+<!--
+  - @copyright 2020 Patrick Bender <patrick@bender-it-services.de>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<div>
+		<ul class="aliases-list">
+			<li v-for="curAlias in aliases" :key="curAlias.id">
+				<strong>{{ curAlias.name }}</strong> &lt;{{ curAlias.alias }}&gt;
+
+				<button class="icon-delete" @click="deleteAlias(curAlias)"></button>
+			</li>
+		</ul>
+
+		<div>
+			<input
+				v-if="addMode"
+				id="alias-name"
+				v-model="alias.aliasName"
+				type="text"
+				:placeholder="t('mail', 'Name')"
+				:disabled="loading"
+			/>
+
+			<input
+				v-if="addMode"
+				id="alias"
+				ref="email"
+				v-model="alias.alias"
+				type="email"
+				:placeholder="t('mail', 'Mail Address')"
+				:disabled="loading"
+			/>
+		</div>
+
+		<div>
+			<button v-if="!addMode" class="primary icon-add" @click="enabledAddMode">
+				{{ t('mail', 'Add alias') }}
+			</button>
+
+			<button
+				v-if="addMode"
+				class="primary"
+				:class="loading ? 'icon-loading-small-dark' : 'icon-checkmark-white'"
+				:disabled="loading"
+				@click="saveAlias"
+			>
+				{{ t('mail', 'Save') }}
+			</button>
+		</div>
+	</div>
+</template>
+
+<script>
+import logger from '../logger'
+import Vue from 'vue'
+
+export default {
+	name: 'AliasSettings',
+	props: {
+		account: {
+			type: Object,
+			required: true,
+		},
+	},
+	data() {
+		return {
+			addMode: false,
+			loading: false,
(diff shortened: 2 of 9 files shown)


l10n/de.js
```diff
@@ -11,6 +11,7 @@ OC.L10N.register(
     "Auto" : "Automatisch",
     "Name" : "Name",
     "Mail Address" : "E-Mail-Adresse",
+    "Aliases" : "Aliase",
```
ChristophWurst, line 14: Please revert any l10n updates. These are done via Transifex and a cron job, not manually.

lib/Db/Alias.php
```diff
@@ -49,4 +51,13 @@ public function __construct() {
...
 		$this->addType('alias', 'string');
 	}
+
+	public function jsonSerialize() {
+		$result = [
+			'id' => $this->getId(),
+			'name' => $this->getName(),
+			'alias' => $this->getAlias()
+		];
+		return $result;
```
ChristophWurst, line 61: no need for the variable, return it directly :)

lib/Db/Alias.php
```diff
@@ -49,4 +51,13 @@ public function __construct() {
 		$this->addType('name', 'string');
 		$this->addType('alias', 'string');
 	}
+
+	public function jsonSerialize() {
```
ChristophWurst, line 55: you mad add a `: array` return type hint here

src/components/AliasSettings.vue
```diff
@@ -0,0 +1,136 @@
...
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<div id="aliases" class="section">
+		<h2>{{ t('mail', 'Aliases') }}</h2>
+		<div id="aliases-list">
+			<table class="grid">
```
ChristophWurst, line 24: @jancborchardt is there a way around table?

src/components/AliasSettings.vue
```diff
@@ -0,0 +1,136 @@
...
+	props: {
+		account: {
+			type: Object,
+			required: true,
+		},
+	},
+	data() {
+		return {
+			loading: false,
+			aliases: this.account.aliases,
```
ChristophWurst, line 79: this should be a `computed` property so it updates with vuex mutations

src/components/AliasSettings.vue
```diff
@@ -0,0 +1,136 @@
...
+			alias: {aliasName: '', alias: ''},
+		}
+	},
+	methods: {
+		deleteAlias(alias) {
+			this.loading = true
+
+			this.$store
```
ChristophWurst, line 87: I would recommend `async`/`await` if you are familiar with that, you can use promises as well if you are more comfortable with that

src/components/Composer.vue
```diff
@@ -335,7 +335,23 @@ export default {
 	},
 	computed: {
 		aliases() {
-			return this.$store.getters.accounts.filter((a) => !a.isUnified)
+			let cnt = 0
+			const aliases = []
+			const accounts = this.$store.getters.accounts.filter((a) => !a.isUnified)
+			accounts.forEach((account) => {
+				account.aliasId = null
```
ChristophWurst, line 342: :warning: this is :fish:y. You should not mutate any data from the store. but you assign new props here. Vue will definitely warn you about this in the dev build. Any change of the state should go through an action and a mutation.

perhaps what you want to do is `.map(...)` the accounts to a new structure, as in, create new object

src/views/AccountSettings.vue
```diff
@@ -51,6 +53,7 @@ export default {
 		EditorSettings,
 		Navigation,
 		SignatureSettings,
+		AliasSettings,
```
ChristophWurst, line 56: you'll get bonus points if you arrange the props alphabetically :wink:

ChristophWurst:
Btw @IchbinkeinReh I've added you as collaborator on the repo :)

jancborchardt:
Nice work! :) Some design feedback:

- [ ] We don’t need a separate section here, we can simply integrate it into the name + email display at the top:
    image
- [ ] This also means it doesn’t need to be a table but can simply be a list → also referring to what @ChristophWurst said above.
- [ ] The name input should go first, and it can be prefilled with the display name, with the email field then being focused.

l10n/de.js
IchbinkeinReh: Done

lib/Db/Alias.php
IchbinkeinReh: Done

lib/Db/Alias.php
IchbinkeinReh: Done

src/views/AccountSettings.vue
IchbinkeinReh: Done

IchbinkeinReh:
@jancborchardt  What do you think about this?
ichbinkeinreh de_cloud_index php_apps_mail_ (1)
ichbinkeinreh de_cloud_index php_apps_mail_

src/components/AliasSettings.vue
IchbinkeinReh: Thanks for the suggestion. Async await really makes the code much more readable

src/components/Composer.vue
```diff
@@ -335,7 +335,24 @@ export default {
 	},
 	computed: {
 		aliases() {
-			return this.$store.getters.accounts.filter((a) => !a.isUnified)
+			let cnt = 0
+			const aliases = []
+			const accounts = this.$store.getters.accounts.filter((a) => !a.isUnified)
+			accounts.forEach((account) => {
```
ChristophWurst, line 341: I have the feeling something like `flatMap` would be more appropriate here. Which attributes of the account/alias do we need? Is a full deep copy necessary or can we just build new objects from the existing data?

IchbinkeinReh: The selectedAlias which contains the selected value from the to field is used in 10 places in the composer. But you are right, not all data from the account is needed. Only the following attributes are needed: editorMode, signature, id, aliasId

IchbinkeinReh: @ChristophWurst BTW flatmap has no IE support is that a problem?

ChristophWurst: You can use https://lodash.com/docs/4.17.15#flatMap but we also have Babel so it should polyfill if flatMap is an ES standard

IchbinkeinReh:
@ChristophWurst Can this PR be merged now? Or didnt you have time to look into the changes yet?

ChristophWurst:
I'll have a look very soon :pray:

jancborchardt:
By the way @IchbinkeinReh we have a Nextcloud Talk instance with a channel for the Nextcloud Mail team – I can invite you there with a guest account if you like (via the email address + name from your commit)? :)

IchbinkeinReh:
@jancborchardt Hey, thanks but its not necessary. I probably won't work at anything related to nextcloud mail the next time. Only wanted to revive the aliases and test my VueJS skills ;)
But if you guys need a helping hand I could do some paid contract work. Just contact me by mail if that's the case

ChristophWurst:
@MarZi235 use https://help.nextcloud.com/c/apps/mail/35 for questions please.

MarZi235:
Thank you, Christoph. Posted https://help.nextcloud.com/t/is-the-alias-feature-dropped/159972

PR #3339, opened 2020-07-02
Fixed possible nullpointer exception in ImportantMessagesExtractor

IchbinkeinReh:
Sometimes one of my emails account does not load the latest emails. The log is then full of this message "Call to a member function getEmail() on null" [error.log](https://github.com/nextcloud/mail/files/4861634/error.log). This PR fixes the nullpointer exception. The issue seems to have the same cause as #851


ChristophWurst:
I think this is a duplicate of https://github.com/nextcloud/mail/pull/3269

IchbinkeinReh:
Yes, i think you are right. I will close the PR

PR #3353, opened 2020-07-06
Make sure all commands return an int

ChristophWurst:
For https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#symfony-update

### DIFF
--- a/lib/Command/CleanUp.php
+++ b/lib/Command/CleanUp.php
@@ -47,10 +47,9 @@ class CleanUp extends Command {
 		$this->setDescription('clean up all orphaned data');
 	}
 
-	/**
-	 * @return void
-	 */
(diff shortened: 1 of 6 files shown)


lib/Command/CleanUp.php
```diff
@@ -47,10 +47,9 @@ protected function configure() {
 		$this->setDescription('clean up all orphaned data');
 	}
 
-	/**
-	 * @return void
-	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
```
nickvergessen, line 50:
```suggestion
	protected function execute(InputInterface $input, OutputInterface $output): int {
```
?

ChristophWurst: sure

PR #3355, opened 2020-07-07
Implement the jwz threading algorithm

ChristophWurst:
Extracted from https://github.com/nextcloud/mail/pull/2125 to make the PR more reviewable.

This is basically an implementation of https://www.jwz.org/doc/threading.html but using a few tricks from https://tools.ietf.org/html/rfc5256 like the root container (as opposed to a list of root containers).

The algorithm here doesn't do anything yet, the integration will follow with #2125. All this patch is add the algorithm and a set of tests that ensure some basic cases work as expected. I would say we have to see if and how well it works and take a closer look once we have the real data. On paper it's a bit hard to reason about the correctness. But not crashing is already a good achievement I guess.

For those who really want to dig into it I left comments for the respective sections on https://www.jwz.org/doc/threading.html in each of the sections of the algorithm. On a high level there is one method that builds the threads and that splits the complexity in five private methods for each of the main steps.

Ref https://github.com/nextcloud/mail/projects/11#card-41435877

### DIFF
--- a/lib/IMAP/Threading/ThreadBuilder.php
+++ b/lib/IMAP/Threading/ThreadBuilder.php
@@ -1,0 +1,220 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


lib/IMAP/Threading/ThreadBuilder.php
```diff
@@ -0,0 +1,236 @@
...
+		// Return the children with reset numeric keys
+		return array_values($rootContainer->getChildren());
+	}
+
+	/**
+	 * @param Message[] $messages
+	 *
+	 * @return Container[]
+	 */
+	private function buildIdTable($messages): array {
```
nickvergessen, line 79:
```suggestion
	private function buildIdTable(array $messages): array {
```

lib/IMAP/Threading/ThreadBuilder.php
```diff
@@ -0,0 +1,236 @@
...
+		/** @var Container[] $idTable */
+		$idTable = [];
+
+		foreach ($messages as $message) {
+			/** @var Message $message */
+			// Step 1.A
```
nickvergessen, line 85: Describe the steps for your future self?

ChristophWurst: It's all at https://www.jwz.org/doc/threading.html, I didn't want to copy everything over

PR #3373, opened 2020-07-17
Show previous and successive messages of the same thread

GretaD:

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -73,6 +73,16 @@
 					</Modal>
 				</div>
 			</div>
+			<router-link v-for="threadMessage in previousThread"
+				:key="threadMessage.id"
+				:to="threadRoute(threadMessage)"
+				class="icon-mail">
+				<div class="address">
+					{{ threadMessage.from[0].label }}
+				</div>
+				<!-- TODO: instead of subject it should be shown the first line of the message #2666 -->
+				{{ threadMessage.subject }}
+			</router-link>
 			<div :class="[message.hasHtmlBody ? 'mail-message-body mail-message-body-html' : 'mail-message-body']">
 				<div v-if="message.itineraries.length > 0" class="message-itinerary">
 					<Itinerary :entries="message.itineraries" :message-id="message.messageId" />
@@ -90,6 +100,16 @@
 				</Popover>
 				<div id="reply-composer" />
 			</div>
+			<router-link v-for="threadMessage in successiveThread"
+				:key="threadMessage.id"
+				:to="threadRoute(threadMessage)"
+				class="icon-mail">
+				<div class="address">
+					{{ threadMessage.from[0].label }}
+				</div>
+				<!-- TODO: instead of subject it should be shown the first line of the message #2666 -->
+				{{ threadMessage.subject }}
+			</router-link>
 		</template>
 	</AppContentDetails>
 </template>
@@ -104,7 +124,10 @@ import Modal from '@nextcloud/vue/dist/Components/Modal'
 import { generateUrl } from '@nextcloud/router'
 
 import AddressList from './AddressList'
-import { buildRecipients as buildReplyRecipients, buildReplySubject } from '../ReplyBuilder'
+import {
+	buildRecipients as buildReplyRecipients,
+	buildReplySubject,
+} from '../ReplyBuilder'
 import Error from './Error'
 import { getRandomMessageErrorMessage } from '../util/ErrorMessageFactory'
 import { html, plain } from '../util/text'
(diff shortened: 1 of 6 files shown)


GretaD:
> Looks good! Does it work? :)

almost :) its missing some front end polishing

src/components/Message.vue
```diff
@@ -21,6 +21,16 @@
...
 				</div>
+				<Actions default-icon="icon-mail" style="display: block;">
+					<ActionButton
+						v-for="message in thread"
+						:key="message.id"
+						icon="icon-mail"
+						{{
+						message.subject
+						}}
+					></ActionButton>
```
juliushaertl, line 32:
```suggestion
					>{{ message.subject }}</ActionButton>
```

lib/Controller/MessagesController.php
```diff
@@ -222,6 +222,15 @@ public function getBody(int $accountId, string $folderId, int $messageId): JSONR
 			base64_decode($folderId),
 			$messageId
 		);
+
+		// FIXME: This is probably not a sane way to map the IMAP message to the database
```
juliushaertl, line 226: @ChristophWurst Please have a look at this once you're back. I wasn't able to dig too deep into it to see if there is a better way. Aparently the Message component just gets the data directly from IMAP and therefore the database id that is required for fetching the thread is missing there.

GretaD:
@jancborchardt right now, the threading looks like this:
thread_test

I did a test structure with the button, just so you could have something to test. 
What is left from my side(not design related) is the link between the thread and the actual message - so when you click, it shows the whole body reply. Im poking you now, just so you have more time to think :)

What I need from you is to please tell me, what's the best way to show the threading. Google has it like this: https://lh5.googleusercontent.com/GExDeK52zJZTAb-xfJJv2lHMC3CO1yNP1NkacanRStlMebBWUKqGtBytx82dgTYsWvsPHYIQc6bjVaJOkUCLdpWqwSRSr_gONGx9DSzm0kYYEueXxKlK3N4n6ygv1aWX
I like the google idea but we have to see how we can integrate it. 
Nevertheless, please tell me your idea, and if you can show it with image examples, i would be very grateful :).

jancborchardt:
@GretaD did you see the "Hey" video? This is how threading should ideally look like (it’s very similar to how Gmail does it): https://www.youtube.com/watch?v=UCeYTysLyGI&t=19m45s

So that entire thread can be shown as 1 entry in the MessageList, and then the individual mails below each other, newest on the bottom, and every read one should be displayed as collapsed (just like Gmail and Hey do it).
That way we don’t need the extra menu in the top right at all. :)

Ideal would be if we could also collapse any quoted messages or signatures from previous messages to save space and reduce necessary scrolling, but that is maybe a further step.

GretaD:
> @GretaD yes, doing it like Gmail sounds great! :) Also another good reference is Hey: https://www.youtube.com/watch?v=UCeYTysLyGI&t=19m45s
> 
> Ideal would be if we could also collapse any quoted messages or signatures from previous messages to save space and reduce necessary scrolling, but that is maybe a further step.

<3 thank you, i will let you know when its ready for a final review, so you dont lose time to check my fixups.

GretaD:
BUG

- [ ] Fix Iframe for MessageHTMLBody

ChristophWurst:
> Ideal would be if we could also collapse any quoted messages or signatures from previous messages to save space and reduce necessary scrolling, but that is maybe a further step.

Yep, this is definitely a later step as it's not a trivial problem to solve. We either have to rewrite the original message in a way that the iframe displays with collapsible quotes or we need some JS inside the iframe. Both paths are rather complex and I would like to focus on the grouping and navigation between related messages first.

ChristophWurst:
Oooooooooof … I'm in the middle of a 15min rebase battle … now I realize @GretaD you didn't push the changes we did together?! Do you still have them?

ChristophWurst:
Rebased. Now this works with the latest changes from #3439, yay.

lib/Controller/MessagesController.php
ChristophWurst: Yes, my bad. The restructuring from #3439 fixed this. But good catch :)

ChristophWurst:
Jumping between messages from the same thread works very nicely already :)

ChristophWurst:
Layout-wise we currently have

* msg1
* msg2
* msg3 (currently active one, but still shown as link)
* msg4
* content of msg3

… if you open a message in the middle of a thread. But it should be

* msg1
* msg2
* content of msg3
* msg4

… like in gmail/hey. Does that make sense? not sure how well the messages *newer* than the current message will render when they are down below, but let's find out :)

ChristophWurst:
This is very basic but let's integrate and improve the design step by step :+1:

ChristophWurst (approved):

PR #3388, opened 2020-07-24
Implement dashboard panel

juliushaertl:
| Empty | List |
|-------|------|
| image | image |


src/views/Dashboard.vue
```diff
@@ -0,0 +1,120 @@
...
+					targetUrl: generateUrl(`/apps/mail/#/accounts/${accountId}/folders/${mailbox}/message/${accountId}-${mailbox}-${uid}`),
+					mainText: item.from ? item.from[0].label : '',
+					subText: item.subject,
+					message: item,
+				}
+			}
+		},
+	},
+	mounted() {
+		// TODO: check if there is a more sane way to query this and if other mailboxes should be fetched as well
```
juliushaertl, line 93: @ChristophWurst I'm not sure if there is a way to get the important mail without fetching each account individually, but i assume this is good for a first attempt.

juliushaertl:
Ready for review and testing

src/views/Dashboard.vue
```diff
@@ -0,0 +1,120 @@
...
+				</template>
+			</DashboardWidgetItem>
+		</template>
+		<template v-slot:empty-content>
+			<div class="empty-content">
+				<img class="empty-content__image" :src="emptyImage">
+				<p class="empty-content__text">
+					{{ t('mail', 'No messages found yet') }}
+				</p>
+				<a v-if="accounts.length === 2" :href="accountSetupUrl" class="button">{{ t('mail', 'Setup an account') }}</a>
```
jancborchardt, line 42: Tiny detail: "Set up an account" cause it’s a verb in this case. :)

jancborchardt:
I’m seeing this for the status icons – even though I have nextcloud-vue-dashboard linked:
image

-----

And just cause we have that on the panel integrations like GitHub etc – would be really nice to have an action menu there with direct actions for "Reply", "Mark not important" and "Delete". :)

juliushaertl:
> I’m seeing this for the status icons – even though I have nextcloud-vue-dashboard linked:

Did you pull the recent changes of the lib?

> And just cause we have that on the panel integrations like GitHub etc – would be really nice to have an action menu there with direct actions for "Reply", "Mark not important" and "Delete". :)

Let's do small steps here. I'm not too familiar with the mail app code so I'd love look into this once  @ChristophWurst and @GretaD are back :wink:

src/views/Dashboard.vue
```diff
@@ -0,0 +1,120 @@
...
+	},
+	mounted() {
+		// TODO: check if there is a more sane way to query this and if other mailboxes should be fetched as well
+		this.accounts.forEach((account) => {
+			fetchEnvelopes(account.accountId, btoa('INBOX'), 'is:important', undefined, 10).then((messages) => {
+				messages = messages.map((message) => ({ ...message, accountId: account.accountId, mailbox: btoa('INBOX') }))
+				this.messages = this.messages !== null ? [...this.messages, ...messages] : messages
+				this.fetchedAccounts++
+			})
+		})
```
eneiluj, line 100: Messages could be sorted by date. They are currently grouped by account.

They have a `dateInt` attribute.

juliushaertl:
Ready for review. Latest library is also used now so no need for npm link anymore.

src/views/Dashboard.vue
ChristophWurst: I think there is not, because we currently do all the merging logic on the client-side

lib/AppInfo/Application.php
```diff
@@ -59,6 +60,7 @@
 use OCA\Mail\Service\UserPreferenceSevice;
 use OCP\AppFramework\App;
 use OCP\AppFramework\IAppContainer;
+use OCP\Dashboard\RegisterWidgetEvent;
```
ChristophWurst, line 63: I think we need 19+ for this, but currently we allow https://github.com/nextcloud/mail/blob/424624f24d864b372275f2938c6ab7bb17fd673c/appinfo/info.xml#L34

lib/Dashboard/MailWidget.php
```diff
@@ -0,0 +1,94 @@
...
+
+	/** @var IL10N */
+	private $l10n;
+	/** @var IURLGenerator */
+	private $urlGenerator;
+
+	public function __construct(IL10N $l10n, IURLGenerator $urlGenerator, AccountService $accountService, IInitialState $initialState, $userId) {
+		$this->l10n = $l10n;
+		$this->urlGenerator = $urlGenerator;
+		$this->accountService = $accountService;
```
ChristophWurst, line 45: prop does not exist

lib/Dashboard/MailWidget.php
```diff
@@ -0,0 +1,94 @@
...
+		$this->initialState = $initialState;
```
ChristophWurst, line 46: prop does not exist

lib/Dashboard/MailWidget.php
```diff
@@ -0,0 +1,94 @@
...
+		$this->userId = $userId;
```
ChristophWurst, line 47: prop does not exist

src/views/Dashboard.vue
```diff
@@ -0,0 +1,120 @@
...
+			emptyImage: imagePath('mail', 'newsletter.svg'),
+			accountSetupUrl: generateUrl('/apps/mail/#/setup'),
+		}
+	},
+	computed: {
+		loading() {
+			return this.fetchedAccounts < this.accounts.length
+		},
+		importantMessages() {
+			return this.messages ? this.messages.sort((a, b) => b.dateInt - a.dateInt).slice(0, 7) : []
```
ChristophWurst, line 78: eslint is right: `Array.sort` modifies the original array, we should copy the array or use `sort` from lodash

lib/AppInfo/Application.php
ChristophWurst: it will work with older releases as well as we only use it as string :+1:

PR #3430, opened 2020-08-11
Migrate to Nextcloud 20's bootstrap mechanism

ChristophWurst:
* Move registration stuff to `register`
* Move navigation to info.xml

This is a breaking change, thus the bump to min 20. This is unfortunate but we'll need it for the unified search and it's beneficial for the dashboard integration as well.


appinfo/info.xml
```diff
@@ -31,7 +31,7 @@
 	<screenshot>https://user-images.githubusercontent.com/1374172/79554966-278e1600-809f-11ea-82ea-7a0d72a2704f.png</screenshot>
 	<dependencies>
 		<php min-version="7.2" max-version="7.4" />
-		<nextcloud min-version="18" max-version="20" />
+		<nextcloud min-version="20" max-version="20" />
```
thomas-mc-work, line 34: I sincerely hope that this requirement shall not be applied in the next release! That would lock out many users (again).

ChristophWurst: I wish that was possible, but that would block us from having *unified search* and the dashboard integration. I'm afraid there isn't really a way around this.

But we do have stable branches and backports right now, so older releases will get fixes. You just need the newer release for the new features.

This goes by the same argumentation as last time. If we rely on a new API, then we need to also require the first version of Nextcloud that has it.

ChristophWurst: To be very clear: I'm not changing this for fun. We need this for https://github.com/nextcloud/mail/pull/3433, which is a planned feature for v1.5.

ChristophWurst: I discussed this with @rullzer shortly and we really do not have any viable options. If we continue to release this app for older versions, we can't add new features like the unified search. But if we bump the min version, anyone who keeps their Nextcloud updated will also get the new features. Those who stay on the older but stable releases of Nextcloud will continue to get roughly the same version of Mail as before. So you'll have two option in this case: go with latest Nextcloud and all the features around it, or stick with older releases and accept the lack of new changes.

I genuinely hope this makes sense and is acceptable.

PR #3433, opened 2020-08-11
Add a unified search provider 

ChristophWurst:
- [x] Requires https://github.com/nextcloud/mail/pull/3430
- [x] Requires simpler deep links by database ID

Follow-ups
* #3432 
* https://github.com/nextcloud/mail/pull/3436

Proof that it works

Bildschirmfoto von 2020-08-25 13-33-13

PR #3436, opened 2020-08-11
Try to find all text tokens als in the subject

ChristophWurst:
- [ ] Requires https://github.com/nextcloud/mail/pull/3433


ChristophWurst:
Integrated into #3433

Issue #3434, opened 2020-08-11
l10n: No text string to translate in Transifex


Valdnet:
### Expected behavior
Translated information text

### Actual behavior

No translated information text in the native language. The text string is not present in Transifex.
![obraz](https://user-images.githubusercontent.com/47037905/89882189-c39bda00-dbc6-11ea-8e88-49d050671beb.png)

https://github.com/nextcloud/mail/blob/9a17879c6523696b722136ea5224de6d4033bb9f/src/components/MailboxMessage.vue#L27

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)
1.4.1

**Nextcloud Version:** (see admin page, e.g. 13.0.2)
19.0.0

ChristophWurst:
Thanks! Fix is at https://github.com/nextcloud/mail/pull/3437

PR #3437, opened 2020-08-11
Fix l10n extraction from wrapped translation text

ChristophWurst:
FIxes https://github.com/nextcloud/mail/issues/3434


GretaD (approved):
Oh this again :/

PR #3456, opened 2020-08-18
Mapping "STARTTLS" to "tls" for autoconfig XML compatible with Thunderbird

benks-io:
As Thunderbird expects the string "STARTTLS" from the autoconfig XML*, but Mail App internally uses "tls", starttls is now being mapped to tls.

*source: https://wiki.mozilla.org/Thunderbird:Autoconfiguration:ConfigFileFormat


ChristophWurst (approved):
Awesome!

ChristophWurst:
I've manually set the DCO check to pass for a tiny change like this, but please sign off your commit next time :)

PR #3461, opened 2020-08-21
Fix accessibility problems

GretaD:
Some small fixes to improve accessibility


GretaD:
@jancborchardt do you want to test it before merge?

ChristophWurst:
We'll just have @jancborchardt do a re-test of the whole app once you're done :)

PR #3486, opened 2020-08-27
Change keyboardshortcut into modal

GretaD:
some small polishing left

- [x] Remove the client-side route that is now obsolete
- [x] Remove the server-side route and controller method that is now obsolete

### DIFF
--- a/src/views/KeyboardShortcuts.vue
+++ b/src/views/KeyboardShortcuts.vue
@@ -1,73 +1,74 @@
 <template>
-	<Content app-name="mail">
-		<Navigation />
-		<AppContent>
-			<AppContentDetails>
-				<h2>{{ t('mail', 'Keyboard shortcuts') }}</h2>
+	<Modal
+		size="large"
+		:title="t('mail', 'Keyboard shortcut')"
+		@close="$emit('close')">
+		<h2>{{ t('mail', 'Keyboard shortcuts') }}</h2>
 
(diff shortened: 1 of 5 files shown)


src/views/KeyboardShortcuts.vue
```diff
@@ -1,73 +1,77 @@
...
-		<Navigation />
-		<AppContent>
-			<AppContentDetails>
-				<h2>{{ t('mail', 'Keyboard shortcuts') }}</h2>
+	<Modal>
+		<div
+		class="modal-content"
+		size="large"
+		:title="t('mail', 'Keyboard shortcut')"
+		@close="$emit('close')">
```
skjnldsv, line 7:
```suggestion
	<Modal>
		size="large"
		:title="t('mail', 'Keyboard shortcut')"
		@close="$emit('close')">
```

ChristophWurst: :musical_note:  *if you like it then you shoulda put a ``<div/>`` on it* :musical_note:

ChristophWurst:
Btw I think we should now put `src/views/KeyboardShortcuts.vue` to `src/components` as it's not a standalone view anymore

GretaD:
> Btw I think we should now put `src/views/KeyboardShortcuts.vue` to `src/components` as it's not a standalone view anymore

i agree, should we include this in this PR?

ChristophWurst (approved):
Looks great!

PR #3528, opened 2020-09-02
Fix keyboardshortcut scroll

### LINKED ISSUE #3524: Keyboards modal does not scroll but hides overflow
### Expected behavior
Screens small in height should still show all shortcuts.

### Actual behavior
Overflow is hidden, no scrollbar is shown.

Bildschirmfoto von 2020-09-02 11-27-10

This could be an issue in our styling but also from nc-vue.

### Mail app

GretaD:
fixes  #3524

### DIFF
--- a/src/views/KeyboardShortcuts.vue
+++ b/src/views/KeyboardShortcuts.vue
@@ -64,11 +64,15 @@ export default {
 <style lang="scss" scoped>
 ::v-deep .modal-wrapper .modal-container {
 	display: block;
-	overflow: hidden;
+	overflow: scroll;
 	transition: transform 300ms ease;
 	border-radius: var(--border-radius-large);
 	box-shadow: 0 0 40px rgba(0,0,0,0.2);
 	padding: 30px 40px 20px;
+}
+dt {
+	width: auto;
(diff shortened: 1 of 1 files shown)


src/views/KeyboardShortcuts.vue
```diff
@@ -64,11 +64,15 @@ export default {
...
 ::v-deep .modal-wrapper .modal-container {
 	display: block;
-	overflow: hidden;
+	overflow: scroll;
 	transition: transform 300ms ease;
 	border-radius: var(--border-radius-large);
 	box-shadow: 0 0 40px rgba(0,0,0,0.2);
 	padding: 30px 40px 20px;
 }
+dt {
```
ChristophWurst, line 73: @skjnldsv I remember something about slower performance if the selector is an element rather than an ID or class. Is this the case?

ChristophWurst (approved):

src/views/KeyboardShortcuts.vue
skjnldsv: It was something about id vs classes. But shouldn't be that significant for such low amount of doms

PR #3542, opened 2020-09-04
Update FolderMapper.php

wackerm:
We have to use a imap mail server, which does support special use tags bot does indeed not advertise the special use capabilities.
So for a quick fix i suppose to add German folder names to guess special use here, which helps us here..

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -217,10 +217,10 @@ class FolderMapper {
 	protected function guessSpecialUse(Folder $folder): void {
 		$specialFoldersDict = [
 			'inbox' => ['inbox'],
-			'sent' => ['sent', 'sent items', 'sent messages', 'sent-mail', 'sentmail'],
-			'drafts' => ['draft', 'drafts'],
-			'archive' => ['archive', 'archives'],
-			'trash' => ['deleted messages', 'trash'],
+			'sent' => ['sent', 'sent items', 'sent messages', 'sent-mail', 'sentmail', 'gesendet'],
+			'drafts' => ['draft', 'drafts', 'entwürfe'],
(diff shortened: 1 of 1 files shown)


ChristophWurst:
The concept makes sense, but this does not scale. It should use the translation system so we can use this trick for all languages, not just German.

wackerm:
I understand your point using the translate system, as it makes sense in almost any language based string.
But since the naming of the inbox folders could vary not only for language but also depending on the imap server config it is any way insecure to trust on names and only a dirty fallback. 
Usually the imap server should use "flags" (RFC LIST Extension "special-use" https://tools.ietf.org/html/rfc6154) to mark those special folders. But since the horde imap library is lacking support for the former unofficial but widely adopted XLIST extension by Google, even some not so exotic imap servers will not be recognized.
The fallback sequentially has to compare folder names with defined lists and has to make assumptions based on the original folder names (that could be different from those shown in the clients). Since those folders are sometimes created by the mail user agents, especially outlook or thunderbird, they use different default names. Sometimes the names are translated, some times not.
For e.g. on a German imap server you could find the names "draft", "junk" or "sent" as folder names, but also "entwürfe", "werbung" or "gesendete objekte" or anything else...
The proposed pull request though is only a small and dirty fix, that helped us and may help others. It may get quite complex to support multiple translations for a folder name in a specific order to guess a best match. Especially since those names can be changed by the mail clients...
In my opinion therer should be a configuration setting and not a somehow blown up guessing engine ;-) But this is another story...

ChristophWurst:
> In my opinion there should be a configuration setting and not a somehow blown up guessing engine ;-) But this is another story...

Almost done via https://github.com/nextcloud/mail/pull/3820.

I'm clsoing this :)

Issue #3432, opened 2020-08-11
Avatars in unified search


jancborchardt:
Right now we show the Mail app icon for each search result, which doesn’t add any additional information. Ideally it should look exactly like in the Mail app, showing avatars/Gravatars/favicons/placeholders etc.

![89879971-f5f80800-dbc3-11ea-804d-74d2925c61ab](https://user-images.githubusercontent.com/925062/89880255-538c5480-dbc4-11ea-9d04-3f083f06ee87.png)

cc @skjnldsv for unified search.

skjnldsv:
Maybe srcset @ChristophWurst ?
So you could provide multiple thumbnailsUrl and it would automatically fallback to the second one if not available?

skjnldsv:
Or you implement your own avatarProvider endpoint that fallback, that way this is just a long request from the front showing an icon while the avatar is loading.

ChristophWurst:
Something like that, yes. But the fallback should probably be a colored two-letter avatar to be in line with the design of the message ist in main app view.

I'll have to analyze the current logic before I can make any plans for a solution. Let's finish the base search implementation first.

StCyr:
closed by https://github.com/nextcloud/mail/pull/3563

PR #3563, opened 2020-09-07
Show avatars in search results

StCyr:
Fix to #3432

### DIFF
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -96,21 +96,31 @@ class Provider implements IProvider {
 			array_map(function (Message $message) {
 				$formattedDate = $this->dateTimeFormatter->formatDateTimeRelativeDay($message->getSentAt(), 'short');
 				$sender = $message->getFrom()->first();
+
 				if ($sender !== null && $sender->getLabel() !== null) {
 					$subline = $sender->getLabel() . ' – ' . $formattedDate;
 				} else {
 					$subline = $formattedDate;
 				}
 
+				if ($sender !== null && $sender->getEmail() !== null) {
+					$from = $sender->getEmail();
+				} else {
+					$from = null;
+				}
+
 				return new SearchResultEntry(
-					'',
+					is_null($from) ? '' : $this->urlGenerator->linkToRoute('mail.avatars.image', [
+						'email' => $from,
+					]),
 					$message->getSubject(),
 					$subline,
 					$this->urlGenerator->linkToRouteAbsolute('mail.page.thread', [
 						'mailboxId' => $message->getMailboxId(),
 						'id' => $message->getId(),
 					]), // TODO: deep URL
-					'icon-mail'
+					'icon-mail',
+					!is_null($from)
 				);
 			}, $messages),
 			$last->getSentAt()


StCyr:
It seems the generation of the two-letters avatar is handled by a vue component (https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Avatar/Avatar.vue). 

Though, it's not a two-letter avatar, just a single letter one...

jancborchardt:
Awesome @StCyr! :) Also cc @GretaD @skjnldsv for help on the avatar placeholder component ^

skjnldsv:
> Though, it's not a two-letter avatar, just a single letter one...

It could just be outdated avatar in your instance appdata?

StCyr:
> 
> 
> > Though, it's not a two-letter avatar, just a single letter one...
> 
> It could just be outdated avatar in your instance appdata?

I don't understand you: The code at https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Avatar/Avatar.vue creates a single letter avatar fallback.

In fact, I don't think there's anything we can do on the front-end side to create a fallback avatar.

skjnldsv:
> I don't understand you: The code at [https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Avatar/Avatar.vue](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Avatar/Avatar.vue?rgh-link-date=2020-09-10T08%3A42%3A08Z) creates a single letter avatar fallback.

I misunderstood you. I assumed it was fixed but you were still saying one char. So I thought about old generated avatars :)

StCyr:
> In fact, I don't think there's anything we can do on the front-end side to create a fallback avatar.

And, that's an issue because I don't think we want to re-create this functionality (fallback avatar generation) on the back-end side. Do we @jancborchardt ?

lib/Search/Provider.php
```diff
@@ -103,7 +103,9 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
 				}
 
 				return new SearchResultEntry(
-					'',
+					$this->urlGenerator->linkToRoute('mail.avatars.image', [
```
nickvergessen, line 106: since it's an "avatar" I think it should be rounded, so set the last parameter of the `SearchResultEntry` to true

lib/Search/Provider.php
```diff
@@ -103,14 +103,17 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+						'email' => $message->getFrom()->first()->getEmail(),
```
ChristophWurst, line 107: 1) There might not be a sender
2) The sender might not have an email

Does not makes sense, I know, but there are systems where this happens for unknown reasons …

ChristophWurst (changes requested):

lib/Search/Provider.php
StCyr: Yes, I'll add error handling :-)

StCyr: done

lib/Search/Provider.php
```diff
@@ -96,21 +96,31 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+					]),
 					$message->getSubject(),
 					$subline,
 					$this->urlGenerator->linkToRouteAbsolute('mail.page.thread', [
 						'mailboxId' => $message->getMailboxId(),
 						'id' => $message->getId(),
 					]), // TODO: deep URL
-					'icon-mail'
+					'icon-mail',
+					true
```
nickvergessen, line 123:
```suggestion
!is_null($from)
```

jancborchardt:
> And, that's an issue because I don't think we want to re-create this functionality (fallback avatar generation) on the back-end side. Do we @jancborchardt ?

Is this still an issue? I would assume the avatar component just takes care of that?

If you ask me of course I would say that it’s best if avatars behave the same everywhere, no matter if in the conversation list, threading view, bubbles in the header, Dashboard widget or search. :)

StCyr:
> > And, that's an issue because I don't think we want to re-create this functionality (fallback avatar generation) on the back-end side. Do we @jancborchardt ?
> 
> Is this still an issue? I would assume the avatar component just takes care of that?

Mail's Avatar 'service only downloads an icon from an URL, nothing else:

https://github.com/nextcloud/mail/blob/fac564324ab595dc152782f2cd915aab8d60cb06/lib/Service/AvatarService.php#L144

And, I don't think Mail can change anything to the unified search results in the frontend.

jancborchardt:
Ok, then I guess let’s just do it in Mail as good as it can be. :) Mail’s avatar service would plug the additional gravatar/favicon part into the avatar component, and otherwise it’s the same? @skjnldsv

lib/Search/Provider.php
```diff
@@ -96,21 +96,31 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
 
+				if ($sender !== null && $sender->getEmail() !== null) {
+					$from = $sender->getEmail();
+				} else {
+					$from = null;
+				}
+
 				return new SearchResultEntry(
-					'',
+					is_null($from) ? '' : $this->urlGenerator->linkToRoute('mail.avatars.image', [
```
ChristophWurst, line 113: that will give you a 404 if no avatar exists

ChristophWurst (changes requested):

lib/Search/Provider.php
ChristophWurst: ah but it works. magic.

ChristophWurst (approved):
Tested & works :)

PR #3569, opened 2020-09-09
Change account settings to modal

GretaD:
- [x] Remove the client-side route that is now obsolete
- [x] Remove the server-side route and controller method that is now obsolete
- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/1493
- [x] Modal background overlay remains after closing the modal


src/views/AccountSettings.vue
```diff
@@ -1,57 +1,50 @@
...
-				<p>
-					<strong>{{ displayName }}</strong> &lt;{{ email }}&gt;
-					<a
-						v-if="!account.provisioned"
-						class="button icon-rename"
-						href="#account-form"
-						:title="t('mail', 'Change name')" />
-				</p>
-				<AliasSettings v-if="!account.provisioned" :account="account" />
+	<Modal
```
ChristophWurst, line 2: time to move this from `views` to `components` as well :)

src/components/NavigationAccount.vue
```diff
@@ -64,13 +64,13 @@
 				{{ t('mail', 'Remove account') }}
 			</ActionButton>
 		</template>
+		<AccountSettings v-if="displayAccountSettings" @close="closeAccountSettings" />
```
ChristophWurst, line 67: I'd like to see the `account` passed as prop here instead of resolving it internally via the URL parameter. This way the modal will actually work, because there currently is no ``this.$route.params.accountId``

GretaD:
The master of Modal aka @skjnldsv, can you please help a bit here, I'm clueless why modal its not working. If you click Account Settings nothing happens and no error is shown.

skjnldsv:
DOn't you want to wait for Marco's SettingsModal?

GretaD:
> DOn't you want to wait for Marco's SettingsModal?

ohh was not aware that she was creating one. Sure :)

skjnldsv:
> ohh was not aware that he was creating one. Sure :)

https://github.com/nextcloud/nextcloud-vue/pull/1286 :)

ChristophWurst:
I think ESLint wants to tell you someting :wink:

ChristophWurst (approved):
Code looks good, didn't test yet

src/components/AccountSettings.vue
```diff
@@ -0,0 +1,135 @@
...
+			<h2>{{ t('mail', 'Account settings') }}</h2>
+		</AppSettingsSection>
+		<p>
+			<strong>{{ displayName }}</strong> &lt;{{ email }}&gt;
+			<a
+				v-if="!account.provisioned"
+				class="button icon-renamesettings"
+				href="#account-form"
+				:title="t('mail', 'Change name')" />
+		</p>
```
ma12-co, line 37: Can this block go into an appsettingsection of its own?

ChristophWurst:
not blocked anymore

src/components/AccountSettings.vue
GretaD: yes, fixed

ChristophWurst (changes requested):
* You can only open the modal once
* The pencil icon next to the email address does not work
* The old settings route is not removed from `appinfo/routes.php`

ma12-co (approved):
:tada:

ma12-co (approved):
Needs rebase and fixup

GretaD:
>     * You can only open the modal once
> 
>     * The pencil icon next to the email address does not work
> 
>     * The old settings route is not removed from `appinfo/routes.php`

fixed

jancborchardt (changes requested):
- [ ] The auto/manual switch doesn’t show properly:
   image
- [ ] The headings "Alias settings", "Signature settings" and "Editor settings" are not needed. (I assume removing them will also fix the indent?)

ChristophWurst:
>     * The headings "Alias settings", "Signature settings" and "Editor settings" are not needed. (I assume removing them will also fix the indent?)

so we're craming everything into a single stream again? I really like this change because it gives a better overview and allows you to find the relevant section more quickly.

jancborchardt:
> so we're craming everything into a single stream again? I really like this change because it gives a better overview and allows you to find the relevant section more quickly.

- It is a single stream even with this change, just with a navigation?
- The 3 first sections are directly visible in the modal, so the navigation is not really needed
- The headings were duplicating the actual headings in place "Account settings", "Signature" and "Writing mode", so this results in more noise, not more structure
- What we could actually do is move the massive part of connection settings on the bottom to something like 2nd-level settings.

ChristophWurst:
The account form styling of the tabs is indeed a regression of this PR. It's fine on master. See https://github.com/nextcloud/mail/pull/1252/files?w=1 for the original PR that added and and the stylings that seem to have lost their effect.

PR #3586, opened 2020-09-14
Fix threading design

GretaD:
- [x]  Fix the left space of threading(where the icon-mail starts)

- [x]  Fix the height on HTML messages

- [x] Fix timestamp and subject

- [x] remove space before comma

- [x] remove not active reply button

~~- [ ] add the non active reply button on the 3 dot menu~~

### DIFF
--- a/src/components/AddressList.vue
+++ b/src/components/AddressList.vue
@@ -2,8 +2,8 @@
 	<span>
 		<template v-for="(entry, idx) in entries">
 			<Address :key="entry.email" :email="entry.email" :label="entry.label" /><!--
-			--><span v-if="idx === entries.length - 2" :key="'split' + entry.email">{{ t('mail', 'and') }}</span><!--
-			--><span v-else-if="idx + 1 < entries.length" :key="'split' + entry.email">, </span>
+			--><span v-if="idx === entries.length - 2" :key="'split' + entry.email">{{ t('mail', 'and') }}</span>
+			<span v-else-if="idx + 1 < entries.length" :key="'split' + entry.email">, </span>
 		</template>
 	</span>
 </template>
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -21,25 +21,27 @@
 
 <template>
 	<div>
-		<div>
+		<div class="envelope--header">
 			<Avatar v-if="envelope.from && envelope.from[0]"
 				:email="envelope.from[0].email"
 				:display-name="envelope.from[0].label"
 				:disable-tooltip="true"
-				:size="25" />
+				:size="40" />
 			<router-link
 				:to="route"
 				event=""
 				class="left"
 				@click.native.prevent="$emit('toggleExpand', $event)">
 				<span class="sender">{{ envelope.from[0].label }}</span>
-				<span class="preview">
-					<!-- TODO: instead of subject it should be shown the first line of the message #2666 -->
-					{{ envelope.subject }}
-				</span>
+				<div class="subject">
+					<span class="preview">
+						<!-- TODO: instead of subject it should be shown the first line of the message #2666 -->
+						{{ envelope.subject }}
+					</span>
+				</div>
 			</router-link>
 			<div class="right">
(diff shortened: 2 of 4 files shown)


GretaD:
@jancborchardt hi, can you please tell me which one is better from the options below. During the design call you proposed to remove the button from the not active ones and keep the active primary one. It looks more or less like this (1)

(1)
primarybutton

The (2) i moved the subject to the second line which in the future will be replaced by the first line of message. Design wise it looks better rather than having recipient and subject in one line.
(2)
newthreadindes

Instead of having (1) i would propose (3) since to me it looks more intuitive which message is active/open.
(3)
threadingnobutton

What do you think? (I am aware of the space problem at time stamp :) )

jancborchardt:
Looks great @GretaD, very good proposal! :) We can even go a bit further than your Proposal 3 and
- For collapsed messages, put the "Reply" action into the 3-dot menu, so that the list gets less cluttered.
- Style the Reply button as `.primary` for one of the entries, for example for the last opened message? Cause otherwise we don’t have any primary action in the view. It’s good to have one though. :)

GretaD:
> Looks great @GretaD, very good proposal! :) We can even go a bit further than your Proposal 3 and
> 
>     * For collapsed messages, put the "Reply" action into the 3-dot menu, so that the list gets less cluttered.
I like the idea to remove it, but i dont think its necessary to put it on 3dot menu, because i dont see any case scenario when someone would reply an un-open message. Google and hey doesnt even have the 3dot menu when the message its not open.
> 
>     * Style the Reply button as `.primary` for one of the entries, for example for the last opened message? Cause otherwise we don’t have any primary action in the view. It’s good to have one though. :)
The active/open/expanded one its still a primary button: https://github.com/nextcloud/mail/pull/3586/files#diff-d794712fba69cdd3a34a47f67fca32f5L43

i just removed the background color and the borders.

jancborchardt:
> Google and hey doesnt even have the 3dot menu when the message its not open.

Alrighty, then let's just do the same. :) Simpler is better.

> i just removed the background color and the borders.

Yep yep – I mean though that 1 button should be the "blue action button", so that should probably be the one of the last opened one. That way it's very clear where the actionable button is.

ChristophWurst:
> For collapsed messages, put the "Reply" action into the 3-dot menu, so that the list gets less cluttered.

- [ ] BUG: There is no "Reply" in that menu

ChristophWurst:
Bildschirmfoto von 2020-09-21 14-24-12
Bildschirmfoto von 2020-09-21 14-21-30

IMO the positioning of the avatar is not ideal, the avatar and the two lines of text next to it should be one the same line or *vertically centered* if that explains it better.

Also long subjects go underneath the reply button

Bildschirmfoto von 2020-09-21 14-24-59

src/components/ThreadEnvelope.vue
```diff
@@ -21,7 +21,7 @@
 
 <template>
 	<div>
-		<div>
+		<div style="display: flex">
```
juliushaertl, line 24: No inline please, just give the div a proper class

GretaD: i didnt know how to call it. how about `leftandrigh` :smiley:

juliushaertl: How about something like thread--header :)

src/components/AddressList.vue
```diff
@@ -2,7 +2,7 @@
 	<span>
 		<template v-for="(entry, idx) in entries">
 			<Address :key="entry.email" :email="entry.email" :label="entry.label" /><!--
-			--><span v-if="idx === entries.length - 2" :key="'split' + entry.email">{{ t('mail', 'and') }}</span><!--
+			--><span v-if="idx === entries.length - 2" :key="'split' + entry.email">{{ t('mail', ' and') }}</span><!--
```
ChristophWurst, line 5: instead of this space, remove the html comment between the two `span`s.

ChristophWurst (changes requested):
Rest looks fine now :)

ChristophWurst (approved):

Issue #3481, opened 2020-08-25
Incorporate IMAP for the unified search results


ChristophWurst:
### Feature Request

The current unified search implementation (https://github.com/nextcloud/mail/pull/3433) does not search in bodies, as we just query the cached data in the database and we don't have the body there.

One of the technical limitations of cache-only search is that on IMAP we can't easily search across all mailboxes.

BUT there is a special IMAP mailbox attribute `\All` and we could use this mailbox (if it exists) to search in there. This won't give us a deep link to the message in the original mailbox, but at least we're able to find it.

As a fallback we could look into the INBOX, because there is a high chance people store most of their emails there, so we might find the message(s) the user is looking for there.

#### Summary

Use `\All` (or `INBOX` if `\All` is not available) mailbox to search email bodies on IMAP.

dumblob:
The outlined solution sounds like a good idea. I'm all for it.

StCyr:
Hello @ChristophWurst ,

I've had a look at this. 

It seems that there is already some things related to IMAP search.

In particular there's this `getIdsLocally` function which seems to be function `getIdsGlobally` counterpart for IMAP search:

https://github.com/nextcloud/mail/blob/530d911fb5f0f5e2c040bdf0b9bdcdfbfbf9fecd/lib/Service/Search/MailSearch.php#L168

So, could it be possible that implementing the functionality be as easy as creating a new `findMessagesLocally` function, counterpart of `findMessagesGlobally` (which calls `getIdsGlobally` and is called by mail 'search provider 'search function) and updating mail 'search provider 'search function to also call this new `findMessagesLocally` function?!?

Maybe the IMAP search should be called as a background job though, as it may slow down mail 'search provider 'search function a lot?

ChristophWurst:
> Maybe the IMAP search should be called as a background job though, as it may slow down mail 'search provider 'search function a lot?

That's the original problem that we're facing with IMAP search. It's slow. We can't do anything in background because we need the results when we send them back to the client. There is nothing like partial or deferred results. Everything has to be ready at the same time.

So yes, once we add back search on IMAP, the search will be noticeably slower.

ChristophWurst:
@StCyr IIRC we discussed https://github.com/nextcloud/mail/pull/3589 in the conf hackweek call. Our findings was that IMAP search is quite unreliable in performance as it depends on where the mail server is located and how fast it can answer a text search request.

So possible workarounds were discussed
[...]

PR #3589, opened 2020-09-15
imap in unified search results

StCyr:
Proof of concept of adding IMAP search to unified search (https://github.com/nextcloud/mail/issues/3481).

Works for INBOX of the user's first account only atm.

What do you think @ChristophWurst ?

### DIFF
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -30,6 +30,7 @@ use OCA\Mail\Contracts\IMailSearch;
 use OCA\Mail\Db\Message;
 use OCP\IDateTimeFormatter;
 use OCP\IL10N;
+use OCP\ILogger;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\Search\IProvider;
@@ -52,12 +53,17 @@ class Provider implements IProvider {
 	/** @var IURLGenerator */
 	private $urlGenerator;
 
+	/** @var ILogger */
+	private $logger;
+
 	public function __construct(IMailSearch $mailSearch,
 								IL10N $l10n,
+								ILogger $logger,
 								IDateTimeFormatter $dateTimeFormatter,
 								IURLGenerator $urlGenerator) {
 		$this->mailSearch = $mailSearch;
 		$this->l10n = $l10n;
+		$this->logger = $logger;
 		$this->dateTimeFormatter = $dateTimeFormatter;
 		$this->urlGenerator = $urlGenerator;
 	}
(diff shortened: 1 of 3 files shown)


lib/Search/Provider.php
```diff
@@ -76,13 +82,31 @@ public function getOrder(string $route, array $routeParameters): int {
 
 	public function search(IUser $user, ISearchQuery $query): SearchResult {
 		$cursor = $query->getCursor();
+
+		// Search in local cache (doesn't contain the boday of emails)
```
ChristophWurst, line 86:
```suggestion
		// Search in local cache (doesn't contain the body of emails)
```

:wink:

lib/Search/Provider.php
```diff
@@ -76,13 +82,31 @@ public function getOrder(string $route, array $routeParameters): int {
...
 		$messages = $this->mailSearch->findMessagesGlobally(
```
ChristophWurst, line 87:
```suggestion
		$messagesFromCache = $this->mailSearch->findMessagesGlobally(
```

lib/Search/Provider.php
```diff
@@ -76,13 +82,31 @@ public function getOrder(string $route, array $routeParameters): int {
...
 			$user,
 			$query->getTerm(),
 			empty($cursor) ? null : ((int) $cursor),
 			$query->getLimit()
 		);
 
+		// Search in body of emails on the IMAP server
+		$messages2 = $this->mailSearch->findMessagesLocally(
```
ChristophWurst, line 95:
```suggestion
		$messagesFromImap = $this->mailSearch->findMessagesLocally(
```

lib/Service/Search/MailSearch.php
```diff
@@ -136,6 +139,34 @@ public function findMessages(Account $account,
...
+								 ?int $cursor,
+								 ?int $limit): array {
+		$query = $this->filterStringParser->parse($filter);
+		if ($cursor !== null) {
+			$query->setCursor($cursor);
+		}
+
+		$account = $this->accountService->findByUserId($user->getUID())[0];
+
+		$mailbox = $this->mailboxMapper->find($account, 'INBOX');
```
ChristophWurst, line 163: this should catch if there is no INBOX (no matter how unlikely that sounds)

StCyr:
HI @ChristophWurst ,

thanks for your review but I think we don't want to include IMAP search in unified search: This will probably slows things to a crawl.

What's your opinion?

ChristophWurst:
That's true. I'm not sure how to best handle it. We could make the body search on IMAP optional. like only if the user searches `body:xyz` but that is not very user friends and almost impossible to discover.

Alternatively there could be a second search provider so the main one does not get slowed down by IMAP search. But that is subpar in terms of UX as there would be two search results for Mail.

ChristophWurst:
I'd say we continue the discussion at https://github.com/nextcloud/mail/issues/3481 until we found an acceptable solution.

I agree that IMAP search for every search query slows down loading of your results significantly for most users. Only those with local mail clients and very fast search implementations will get quick results.

PR #3590, opened 2020-09-15
Fix avatar threading

GretaD:
use avatar instead of icon-mail
I will be styled properly here #3586

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -21,7 +21,12 @@
 
 <template>
 	<div>
-		<div class="icon-mail">
+		<div>
+			<Avatar v-if="envelope.from && envelope.from[0]"
+				:email="envelope.from[0].email"
+				:display-name="envelope.from[0].label"
+				:disable-tooltip="true"
+				:size="25" />
 			<router-link
 				:to="route"
 				event=""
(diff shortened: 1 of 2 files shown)


GretaD:
@jancborchardt can you please review this PR so we can merge it and then style it properly here:  #3586

src/components/ThreadEnvelope.vue
```diff
@@ -321,5 +308,6 @@ export default {
 			margin-left: 4px;
 		}
 	}
-}
+	.avatar {
```
juliushaertl, line 311: Not needed :wink:

src/components/ThreadEnvelope.vue
```diff
@@ -21,7 +21,12 @@
 
 <template>
 	<div>
-		<div class="icon-mail">
+		<div class="avatar">
```
juliushaertl, line 24: This is no longer the avatar but avatar + sender/subject, but since there is no styles defined anyways you can probably just drop the class

juliushaertl (commented):
Small comments, otherwise :+1:

src/components/ThreadEnvelope.vue
GretaD: leftovers :) thank you

src/components/ThreadEnvelope.vue
```diff
@@ -21,7 +21,12 @@
 
 <template>
 	<div>
-		<div class="icon-mail">
+		<div>
+			<Avatar v-if="envelope.from"
```
skjnldsv, line 25:
```suggestion
			<Avatar v-if="envelope.from && envelope.from[0]"
```

Maybe this to be safe?

jancborchardt (approved):
Tested and seems good!

PR #3585, opened 2020-09-14
Revert Bump css-loader from 3.6.0 to 4.3.0

GretaD:
Reverts #3579

PR #3593, opened 2020-09-16
Switch to style-loader

juliushaertl:
Better approach than https://github.com/nextcloud/mail/pull/3585 Should fix styling issues that recent bumps of css-loader caused.

Ideally mail could of course just use https://github.com/nextcloud/webpack-vue-config but I'll leave up to you :grin:


GretaD (approved):
thank youuu

PR #3594, opened 2020-09-16
Inject styles for using native fonts in html mails

juliushaertl:
This will use our native font stack as default fonts in the HTML mail renderings, though it is kind of a hack it seems to work quite well and turns bad browser fonts into something nicer :wink: 

Before:
image

After:
image

### DIFF
--- a/lib/Http/HtmlResponse.php
+++ b/lib/Http/HtmlResponse.php
@@ -32,6 +32,11 @@ class HtmlResponse extends Response {
 	/** @var string */
 	private $content;
 
+	private $injectedStyles = <<<EOF
+* { font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', Arial, 'Noto Color Emoji', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'; }
+EOF;
+
+
 	public function __construct(string $content) {
 		parent::__construct();
 		$this->content = $content;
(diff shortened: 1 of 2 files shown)


jancborchardt (approved):
Quite nice, simple fix to improve HTML mails a lot. :)

juliushaertl:
Tests should also be happy now

lib/Http/HtmlResponse.php
```diff
@@ -43,6 +48,6 @@ public function __construct(string $content) {
 	 * @return string the file contents
 	 */
 	public function render(): string {
-		return $this->content;
+		return '<style>' . $this->injectedStyles . '</style>' . $this->content;
```
ChristophWurst, line 51: :thinking: Is this compliant to HTML standards? I think `$this->content` will most likely start with a `<html>`, right? So it might render because browsers are very flexible and fault tolerant, …

juliushaertl: Yeah it is not compliant at all, however HTML emails in general are in most cases doing weird stuff anyways trying to cover random email clients, so I'd rather stick with this simple noncompliant approach than introducing complex parsing logic that tries to figure out the head element or insert one if there isn't any.

ChristophWurst: Fair enough :)

I did not mean to reject your solution, I just wanted to make sure we are aware that this could potentially cause issues depending on how flexible the rendering in the browser is.

ChristophWurst (approved):

lib/Http/HtmlResponse.php
juliushaertl: Just for reference, officially it could be put in head or body

https://www.w3.org/TR/html52/document-metadata.html#the-style-element
> Contexts in which this element can be used:
>     Where metadata content is expected. 
>     In a noscript element that is a child of a head element. 
>     In the body, where flow content is expected. https://www.w3.org/TR/html52/dom.html#flow-content

PR #3595, opened 2020-09-16
Document the mail.app.verify-tls-peer setting

### LINKED ISSUE #3372: Documentation does not list features and configuration documentation (e.g. admin documentation) is very vague and seems like it might be incomplete
### Expected behavior
I don't know how to configure NextCloud Mail and what it can and cannot do. What are the expected use cases (e.g. can admins pre-configure a mailbox for nextcloud users with a certain pattern or are users expected to configure an account from scratch?)

### Actual behavior
More documentation would be needed

StCyr:
Fix #3372

### DIFF
--- a/doc/admin.md
+++ b/doc/admin.md
@@ -25,6 +25,11 @@ You can use the php-mail function to send mails. This is needed for some webhost
 ```php
 'app.mail.transport' => 'php-mail'
 ```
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Thanks a lot!

Issue #3297, opened 2020-06-24
Uploaded attachments included as application/octet-stream MIME part


mjog:
### Expected behavior

When attaching a file from the local computer, the MIME part's content type should reflect the actual content type of the file, e.g. `image/jpeg` if attaching a JPEG image, `application/pdf` if attaching a PDF, and so on.

This allows recipient's MUAs to display previews, open the attachments in the correct application, and so on.

### Actual behavior

All files attached from the local computer are included as MIME parts with content-type `application/octet-stream`:

```
Content-Type: application/octet-stream; name=Ritas_Takeaway.pdf
Content-Disposition: attachment; filename=Ritas_Takeaway.pdf
Content-Transfer-Encoding: base64

[snip: base64 data]
```

### Mail app

**Mail app version:** 1.4.0

**Mailserver or service:** Dovecot

### Server configuration

**Operating system**: Debian 10

**Web server:** Apache

**Database:** MariaDB

**PHP version:** Whatever comes with Debian 10

**Nextcloud Version:** 19.0.0

#### Client configuration
**Browser:** GNOME Web 3.36.2

**Operating system:** Ubuntu 20.04

StCyr:
Ah! it seems that OCP\Files\File's getMimeType function doesn't work for uploaded attachment.

Modifying the code as follow:

```php
root@vps612142:/var/www/nextcloud/apps/mail# git diff lib
diff --git a/lib/Service/MailTransmission.php b/lib/Service/MailTransmission.php
index e8ff9d135..76249c7b1 100644
--- a/lib/Service/MailTransmission.php
+++ b/lib/Service/MailTransmission.php
@@ -317,6 +317,8 @@ class MailTransmission implements IMailTransmission {
                try {
                        list($localAttachment, $file) = $this->attachmentService->getAttachment($userId, $id);
                        $message->addLocalAttachment($localAttachment, $file);
+                       $this->logger->debug('***** LOCAL ********************************************');
+                       $this->logger->debug($file->getMimeType());
                } catch (AttachmentNotFoundException $ex) {
                        $this->logger->warning('ignoring local attachment because it does not exist');
                        // TODO: rethrow?
@@ -350,6 +352,8 @@ class MailTransmission implements IMailTransmission {
 
                if (!is_null($file)) {
                        $message->addAttachmentFromFiles($file);
+                       $this->logger->debug('***** CLOUD ********************************************');
+                       $this->logger->debug($file->getMimeType());
                }
        }
 }
```

[...]

PR #3606, opened 2020-09-18
Add mime type to attachments

StCyr:
Fix for #3297 

TODO:
- [x] Get mime type of uploaded files
- [ ] Get mime type of files attached from Files;
- [x] Update DB to store mime type in table oc_mail_attachments;
- [ ] Actually make use of that mime type information in the composed email.

I would like to have a confirmation of this approach as I believe I've to touch the DB 'schema

### DIFF
--- a/lib/Contracts/IAttachmentService.php
+++ b/lib/Contracts/IAttachmentService.php
@@ -32,7 +32,7 @@ interface IAttachmentService {
 	/**
 	 * Save an uploaded file
 	 */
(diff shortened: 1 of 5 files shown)


ChristophWurst:
> I would like to have a confirmation of this approach as I believe I've to touch the DB 'schema

I guess this is fine :+1:

StCyr:
nothing to do with the actual issue :-)

PR #3607, opened 2020-09-18
Fix action input label size

### LINKED ISSUE #3567: Input submit button has different size
### Expected behavior
The button has the same size as input field

### Actual behavior

Screenshot from 2020-09-08 13-08-18

GretaD:
fixes #3567

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -272,3 +272,12 @@ export default {
 	},
 }
 </script>
+<style lang="scss" scoped>
+
+::v-deep .action-input {
+
+	&__label {
+		box-sizing: border-box;
+	}
+}
+</style>


ChristophWurst:
I've also seen this in calendar. I don't think we have ot fix this here but in nc-vue, no?

GretaD:
> I've also seen this in calendar. I don't think we have ot fix this here but in nc-vue, no?

this it is fixed here: https://github.com/nextcloud/nextcloud-vue/commit/f9c77d73eb372fd9dfb1fc281076679e7a9936b6 actually, but it doesnt apply to mail

ChristophWurst:
> this it is fixed here: [nextcloud/nextcloud-vue@f9c77d7](https://github.com/nextcloud/nextcloud-vue/commit/f9c77d73eb372fd9dfb1fc281076679e7a9936b6) actually, but it doesnt apply to mail

That fix is only in 2.6.5. See the linked tag on that page. We have to wait for https://github.com/nextcloud/mail/pull/3617.

ChristophWurst:
> That fix is only in 2.6.5. See the linked tag on that page. We have to wait for #3617.

@GretaD please re-evaluate

GretaD:
fixed

PR #3637, opened 2020-09-22
Gets mime type of uploaded attachments

StCyr:
Fixes https://github.com/nextcloud/mail/issues/3297

### DIFF
--- a/lib/Controller/LocalAttachmentsController.php
+++ b/lib/Controller/LocalAttachmentsController.php
@@ -47,7 +47,7 @@ class LocalAttachmentsController extends Controller {
 	 * @param string $UserId
 	 */
 	public function __construct(string $appName, IRequest $request,
-								IAttachmentService $attachmentService, $UserId) {
+					IAttachmentService $attachmentService, $UserId) {
 		parent::__construct($appName, $request);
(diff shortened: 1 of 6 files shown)


lib/Controller/LocalAttachmentsController.php
```diff
@@ -40,16 +41,21 @@ class LocalAttachmentsController extends Controller {
 	/** @var string */
 	private $userId;
 
+	/** @var ILogger */
+	private $logger;
```
ChristophWurst, line 45: prop is never read

ChristophWurst (changes requested):

lib/Controller/LocalAttachmentsController.php
StCyr: yes, I've used it to debug, but now I don't need it anymore. Is it really important to remove it? I thought it would be good to leave it for the next developer who happens to need it.

ChristophWurst: It's dead code, so yeah, it should be removed. Clean code is good code :broom:

ChristophWurst (approved):

Issue #3628, opened 2020-09-22
Extract threading info to reproduce threading bugs more easily


ChristophWurst:
### Feature Request

v1.5 will have threading and we still occasionally see related messages that are not grouped correctly. It's not trivial to get the relevant info from the database, so I think we could add a debug-mode-only threading output. Like for each message you see

* Message ID
* References header
* In-Reply-To header

and people can gather this info for the messages they expect to see grouped. Then we can put this data into our unit test set and debug the algorithm. This might the only sane way to handle problems that are so reliant on the actual user data that we typically don't have access to.

#### Summary

Make threading bugs easier to reproduce.

PR #3638, opened 2020-09-23
Add a debug-only file download for threading data

ChristophWurst:
Threading still doesn't work for some messages. To debug this more
easily, especially for non devs, this adds a new action menu entry to
download all relevant threading data as a simple JSON. The maintainers
can then look at those files, put them into unit tests and see why they
messages don't group.

Again, this entry won't be shown on production instances but only if
debug mode is enabled.

Bildschirmfoto von 2020-09-23 09-54-05
Bildschirmfoto von 2020-09-23 09-54-01

Fixes https://github.com/nextcloud/mail/issues/3628

@nextcloud/mail in other words if we get bug reports with *threading doesn't work* we'll ask for the files of each of the messages that are expected to be shown in one thread. Then we take a closer look.

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -30,12 +30,14 @@ use JsonSerializable;
 use OCA\Mail\AddressList;
 use OCP\AppFramework\Db\Entity;
 use function in_array;
+use function json_decode;
 use function json_encode;
 
 /**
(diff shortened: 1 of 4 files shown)


StCyr (commented):
works as expected

PR #3641, opened 2020-09-23
Show inline avatars for all participants in a threading

GretaD:
lalalal

- [x] Remove addresses shown in text and keep avatar+text

- [x] Bit of space between the avatar+username chips would be good – e.g. 4px

- [x] Clicking an avatar should redirect you to a new message to this person. Just like clicking the address did before
- [x] There should be a way to see the email for the recipients that have a label. Can't we show that via the tooltip?

- [x] Remove components that are not used anywhere


GretaD:
@jancborchardt @ChristophWurst please have a look and let me now if all is good

jancborchardt:
Very cool! :) Bit of space between the avatar+username chips would be good – e.g. 4px. Otherwise nice!

src/components/Thread.vue
```diff
@@ -7,9 +7,13 @@
 					<h2 :title="threadSubject">
 						{{ threadSubject }}
 					</h2>
-					<p class="transparency">
-						<AddressList :entries="threadParticipants" />
```
ChristophWurst, line 11: also :fire: the address list and address components if they are now unused :)

ChristophWurst (changes requested):

ChristophWurst (changes requested):
Two more things that are missing

- [ ] Clicking an avatar should redirect you to a new message to this person. Just like clicking the address did before
- [ ] There should be a way to see the email for the recipients that have a label. Can't we show that via the tooltip?

ChristophWurst (approved):

PR #3642, opened 2020-09-23
Add debug output to sync processes triggered via the CLI

ChristophWurst:
Just like a CLI priority inbox model training gives all the details, we
want to have the same to diagnose slow/faulty account syncs. This
changes the console logger adapter for the PSR logger and adds it to the
sync process.

### DIFF
--- a/lib/Service/Sync/ImapToDbSynchronizer.php
+++ b/lib/Service/Sync/ImapToDbSynchronizer.php
@@ -46,7 +46,7 @@ use OCA\Mail\IMAP\Sync\Synchronizer;
 use OCA\Mail\Model\IMAPMessage;
 use OCA\Mail\Support\PerformanceLogger;
 use OCP\EventDispatcher\IEventDispatcher;
-use OCP\ILogger;
+use Psr\Log\LoggerInterface;
 use Throwable;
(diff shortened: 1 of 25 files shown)


ChristophWurst:
Bildschirmfoto von 2020-09-23 21-08-55

:fire:

lib/Service/Sync/ImapToDbSynchronizer.php
```diff
@@ -232,7 +236,8 @@ public function sync(Account $account,
 		if (!$batchSync) {
 			$this->dispatcher->dispatchTyped(
 				new SynchronizationEvent(
-					$account
+					$account,
+					$this->logger
```
nickvergessen, line 240: All other usages in this method use `$logger` not `$this->logger`?

nickvergessen (approved):
ƪ(˘⌣˘)ʃ

lib/Service/Sync/ImapToDbSynchronizer.php
ChristophWurst: Yeah because that is a different path, currently just triggered via a controller/service and not the CLI

PR #3644, opened 2020-09-23
Add a flag that indicates whether an inbox should be sync'ed

ChristophWurst:
The cron background sync can be expensive with accounts that have many
mailboxes. As it turns out other clients like Thunderbird also don't
look into *all* mailboxes to check for new email. Instead they only do
that for INBOX by default and let the user pick more mailboxes if they
wish. We should do the same.

This adds a simple flag. Only the inbox and mailboxes that have this
flag set will get a sync in background.

Any other mailbox can still be used, but the sync only happens if the
user has the mailbox open.

This will bring down the load on instances with many accounts,
especially if those have many mailboxes.

- [x] Requires https://github.com/nextcloud/mail/pull/3642


nickvergessen:
Didn't check, but looks good and like it could solve the issue.

Issue #3639, opened 2020-09-23
Collapse quoted text in the threaded view of plain text messages


ChristophWurst:
### Feature Request

Inside the thread you, you don't really need the previous email's text as you can simply expand it there.

The exception would be inline replies. Then we should show the original text.

#### Summary

Collapse/hide quoted text in plain text emails.

---

See https://github.com/nextcloud/mail/issues/3640 for HTML emails.

PR #3648, opened 2020-09-24
Collapse plain text signatures into a detail element

ChristophWurst:
Signatures are a bit of clutter when you view a conversation. So let's
collapse them and only show on click.

The logic is that we try to find the first *paragraph* in the signature. Because I found you can't assume the first line to be too meaningful. Ideally this would be the name, but some have a greeting followed by the name.

If this doesn't work we fall back to the first non-empty line. This should still be okay-ish.

Luckily there is an HTML element we can use for this: `<details>`.

If the design should be improved then let's do a follow-up. This PR
focuses on the splittig logic.

Ref https://github.com/nextcloud/mail/issues/3639

---

### Before

Bildschirmfoto von 2020-09-24 08-46-48

### Now

Bildschirmfoto von 2020-09-24 09-26-03

Bildschirmfoto von 2020-09-24 09-26-11

### DIFF
--- a/src/components/MessagePlainTextBody.vue
+++ b/src/components/MessagePlainTextBody.vue
@@ -1,11 +1,16 @@
 <template>
 	<div>
-		<div id="mail-content" v-html="htmlBody" />
-		<div v-if="signature" class="mail-signature" v-html="htmlSignature" />
+		<div id="mail-content" v-html="nl2br(body)" />
+		<details v-if="signature" class="mail-signature">
+			<summary>{{ signatureSummaryAndBody.summary }}</summary>
+			<span v-html="nl2br(signatureSummaryAndBody.body)" />
+		</details>
 	</div>
 </template>
 
 <script>
+const regFirstParagraph = /(.+\n\r?)+(\n\r?)+/
+
 export default {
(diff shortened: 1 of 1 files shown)


src/components/MessagePlainTextBody.vue
```diff
@@ -1,11 +1,16 @@
 <template>
 	<div>
-		<div id="mail-content" v-html="htmlBody" />
-		<div v-if="signature" class="mail-signature" v-html="htmlSignature" />
+		<div id="mail-content" v-html="nl2br(body)" />
+		<details v-if="signature" class="mail-signature">
+			<summary>{{ signatureSummaryAndBody.summary }}</summary>
+			<span v-html="nl2br(signatureSummaryAndBody.body)" />
```
juliushaertl, line 6: Is the body sanitized somewhere?

ChristophWurst: yes, at https://github.com/nextcloud/mail/blob/3802480378f6160d1029169afa3334665bd76699/lib/Model/IMAPMessage.php#L436

GretaD (approved):
tested and it works

PR #3650, opened 2020-09-24
Collapse quoted text in plain text messages

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3639

The logic is very simple. It will collapse all quoted text. As a future enhancement we can try to detect inline replies and don't collapse there.

---

### Before

Bildschirmfoto von 2020-09-24 10-55-07

### After

Bildschirmfoto von 2020-09-24 10-54-51


GretaD (approved):
very very kull

Issue #3634, opened 2020-09-22
Active/hover feedback of messages in thread


ChristophWurst:
### Feature Request

![Bildschirmfoto von 2020-09-22 15-32-55](https://user-images.githubusercontent.com/1374172/93888976-f3232380-fce8-11ea-85f6-15f76ab461cf.png)

#### Summary

Do what @ma12-co suggests :)

PR #3651, opened 2020-09-24
Improve thread header hover design and align text

ChristophWurst:
It's a nice little touch to show a light background color for the
hovered thread header, as suggested by Marco.

Moreover this aligns the email text with the thread subject and the
header text.

Fixes https://github.com/nextcloud/mail/issues/3634

Bildschirmfoto von 2020-09-24 11-37-15


GretaD (approved):
kull

PR #3536, opened 2020-09-03
Fix missing In-Reply-To and References header for replies

ChristophWurst:
To reproduce this, either inspect the message source of a reply written with our app, or see how the threading algorithm fails to associate messages of the same thread.

PR #3666, opened 2020-09-28
Add a test for @StCyr's thread with a reply

ChristophWurst:
In attempt to debug a threading problem I wrote this test so I could
find the bug. But it shows that the threading algorithm itself is fine.

Ref https://github.com/nextcloud/mail/pull/3536#issuecomment-687088753

### DIFF
--- a/tests/Unit/IMAP/Threading/ThreadBuilderTest.php
+++ b/tests/Unit/IMAP/Threading/ThreadBuilderTest.php
@@ -578,4 +578,34 @@ class ThreadBuilderTest extends TestCase {
 			$this->abstract($result)
 		);
 	}
+
+	public function testRealWorldThreadWithReply(): void {
+		$messages = [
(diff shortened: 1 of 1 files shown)


StCyr:
Hello @ChristophWurst 

Thanks for this test. 

In the mean time, I've experienced other issues with threading: I almost posted a bug but then realised it might also be caused by me changing git branches. So, I'm waiting a little bit before posting new issues related to threading

Cyrille

ChristophWurst:
> In the mean time, I've experienced other issues with threading: I almost posted a bug but then realised it might also be caused by me changing git branches. So, I'm waiting a little bit before posting new issues related to threading

No worries. If you think you identified something you can download the debug info with the new entry from https://github.com/nextcloud/mail/pull/3638 for all messages that should be in the same thread and I'll have a look :)

PR #3668, opened 2020-09-28
Fix plain text format message

GretaD:
Before
plaintext_pre

_______________________
After
plaintext_wrap


ChristophWurst (approved):
Whoooopsie :)

Issue #3622, opened 2020-09-21
Important *i* popover its not correctly format 


GretaD:
![infoimportan](https://user-images.githubusercontent.com/12728974/93757925-d7941c00-fc07-11ea-83b5-77bc0a489261.png)

GretaD:
@ma12-co we need your help here :)
The popover is broken and i have no idea why, I tried to _force_ style .popover, .popover__inner and wrapper, and nothing happens. The only thing that worked was if i add `max-width=xxx;` but then you need to fix a lot of other things and that defeats the purpose of using popover.

ChristophWurst:
https://github.com/nextcloud/nextcloud-vue/commit/6721935f3dea33033dba185b2282fedc0bd08724#diff-14e3a3e00c1ab75657b21af1459b20d8 is the one commit that changed the component recently

nickvergessen:
https://github.com/nextcloud/spreed/commit/5edeed2cf50b23be24574dcdb1371e2f7466d6a4

GretaD:
fixed ##3682

PR #3672, opened 2020-09-28
Import style

GretaD:
tentative fix for #3622


src/main.js
```diff
@@ -32,6 +32,8 @@ import Nextcloud from './mixins/Nextcloud'
 import router from './router'
 import store from './store'
 import { fixAccountId } from './service/AccountService'
+// Styles
+import '@nextcloud/dialogs/styles/toast.scss'
```
ChristophWurst, line 36: oh, this is about toasts, no? There is no relation to the popover from `@nextcloud/vue`

PR #3680, opened 2020-09-29
Fix collapsing of quoted text with empty lines

ChristophWurst:
An empty line in a quote looks like ">". The regex didn't match that,
hence every paragraph became a quoted text element of its own. Now the
whole text is combined.

---

### Example email

Bildschirmfoto von 2020-09-29 15-52-59

### What it looked like before

Bildschirmfoto von 2020-09-29 15-53-46

### How it looks now

Bildschirmfoto von 2020-09-29 15-56-01

### DIFF
--- a/src/components/MessagePlainTextBody.vue
+++ b/src/components/MessagePlainTextBody.vue
@@ -25,7 +25,7 @@ export default {
 	},
 	computed: {
 		enhancedBody() {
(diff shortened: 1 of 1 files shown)


GretaD (approved):

jancborchardt (commented):
Good stuff, also just stumbled over that.

Issue #3675, opened 2020-09-29
Attachments popover styling


ChristophWurst:
### Expected behavior

Smooth design.

### Actual behavior

![Bildschirmfoto von 2020-09-29 09-38-47](https://user-images.githubusercontent.com/1374172/94527724-b64eb380-0237-11eb-9305-fd197d242ddb.png)
![Bildschirmfoto von 2020-09-29 09-38-52](https://user-images.githubusercontent.com/1374172/94527719-b51d8680-0237-11eb-950e-913bd50706f6.png)

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)

Latest master

PR #3681, opened 2020-09-29
Improve design of attachments popover

jancborchardt:
Fixes https://github.com/nextcloud/mail/issues/3675

Before | After
-|-
attachments single before|attachments single
attachments multiple before|Attachments multiple


ChristophWurst (approved):
:rocket:

GretaD (approved):

Issue #3676, opened 2020-09-29
Opening an envelope action menu navigates to that envelope


ChristophWurst:
### Expected behavior
It should be possible to open the actions menu of an envelope without opening the thread.

### Actual behavior
The click event triggers a navigation.

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)

Latest master

GretaD:
@juliushaertl yesterday during the mail call, we tried to fix this but we couldnt find a way how to. Can you please have a look, this is why we wanted you to join the call :)

PR #3687, opened 2020-10-02
Do not navigate when clicking an envelope's menu

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3676


ma12-co (approved):

ma12-co:
I think you can also just use `@click.prevent`, without passing in a method

Issue #3689, opened 2020-10-03
deleting a threadEnvelope doesn't navigate to next one


StCyr:
### Expected behavior
When deleting a threadEnvelope, mail should navigate to next one

### Actual behavior
When deleting a threadEnvelope, the envelope is correctly removed from the envelope list but mail doesn't navigate to the next one.

### Mail app

**Mail app version:** 1.5 (master)

StCyr:
well, there should be 2 behaviours:

1- There's only one message in the thread: In this case mail should navigate to the next envelope in the EnvelopeList
2- There are several messages in the thread: In this case mail should naviagate to the next threadEnvelope in the thread

ChristophWurst:
This seems to be obsolete. If you open a thread with more than one message then these behaviors apply:

1. Expanding another message in the thread and deleting it removes that very message. The rest stays the same.
2. Deleting the newest message (the one clicked in the list) deletes the message and navigates you to the next/newer thread in the list

PR #3690, opened 2020-10-03
Fix/3689/navigate on delete thread

StCyr:
fix to https://github.com/nextcloud/mail/issues/3689

Please review

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -20,6 +20,7 @@
 				:envelope="env"
 				:mailbox-id="$route.params.mailboxId"
 				:expanded="expandedThreads.includes(env.databaseId)"
+				:bus="bus"
 				@toggleExpand="toggleExpand(env.databaseId)" />
 		</template>
 	</AppContentDetails>
@@ -44,6 +45,12 @@ export default {
 		ThreadEnvelope,
 	},
 
+	props: {
+		bus: {
+			type: Object,
+			required: true,
+		},
+	},
 	data() {
 		return {
 			loading: true,
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -58,7 +58,9 @@
 				</template>
 			</AppContentList>
 			<NewMessageDetail v-if="newMessage" />
-			<Thread v-else-if="showThread" @delete="deleteMessage" />
+			<Thread v-else-if="showThread"
+				:bus="bus"
+				@delete="deleteMessage" />
 			<NoMessageSelected v-else-if="hasEnvelopes && !isMobile" />
 		</div>
 	</AppContent>
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -178,6 +178,10 @@ export default {
 		MoveModal,
 	},
 	props: {
+		bus: {
+			required: true,
+			type: Object,
+		},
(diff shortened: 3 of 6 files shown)


StCyr:
oups, there's a commit that shouldn't be there... hold on

StCyr:
> oups, there's a commit that shouldn't be there... hold on

Ok resolved. sorry

src/components/Thread.vue
```diff
@@ -50,26 +51,23 @@ export default {
 			message: undefined,
 			errorMessage: '',
 			error: undefined,
+			thread: [],
```
ChristophWurst, line 54: :-1: this makes the thread non-reactive

src/components/Thread.vue
```diff
@@ -50,26 +51,23 @@ export default {
...
-			return this.$store.getters.getEnvelopeThread(this.$route.params.threadId)
-		},
 		threadParticipants() {
 			const recipients = this.thread.flatMap(envelope => {
 				return envelope.from.concat(envelope.to).concat(envelope.cc)
 			})
 			return uniqBy(prop('email'), recipients)
 		},
 		threadSubject() {
-			const thread = this.thread
```
ChristophWurst, line 67: unrelated change?

ChristophWurst (changes requested):
We should update the thread data in the store than mangling with it in a local copy. If you use this solution you'll run into the same issue if you navigate somewhere else and back: the thread change is undone.

src/components/Thread.vue
StCyr: yes, unrelated change. 
just, I think this line doesn't serve any purpose...?

src/components/Thread.vue
StCyr: new version proposed. Shall I create a new PR?

StCyr:
I think I've figured out how thread info were to be managed.

Could you have a look at this PR again @ChristophWurst ?

ChristophWurst (approved):
So right now when deleting a message you

* fetch the thread of this message (1 additional request)
* fetch the thread of all messages in this thread (n additional requests)
* dispatch n+1 mutations

Can't we do it a bit more optimistic and either just update all known envelopes from a message's `thread` property or all other messages with the same `mailboxId` and `threadRootId`?

StCyr:
> 
> 
> So right now when deleting a message you
> 
>     * fetch the thread of this message (1 additional request)
> 
>     * fetch the thread of all messages in this thread (n additional requests)
> 
>     * dispatch n+1 mutations
 
Yes, I agree that may be a lot of requests.

> 
> Can't we do it a bit more optimistic and either just update all known envelopes from a message's `thread` property or all other messages with the same `mailboxId` and `threadRootId`?

Wouldn't it be better that the backend provides the thread property directly on a fetchEnvelope call?

ChristophWurst:
> Wouldn't it be better that the backend provides the thread property directly on a fetchEnvelope call?

Ideally, yes. But we have to be careful as this easily increases the runtime complexity from O(1) to O(n+1) if we have to load the thread in one query at a time for every message. The relational database won't allow us to have the list of messages with the threads in one query.

ChristophWurst:
@StCyr conflicts and no update in a while – are you still working on this?

StCyr:
No. You may close it if you like

PR #3692, opened 2020-10-03
Migrate controllers to the PSR logger interface

ChristophWurst:
This fixes some usages of the deprecated ILogger.


StCyr (approved):

eneiluj (approved):
Thanks for reminding me to do it everywhere i can.

Issue #3691, opened 2020-10-03
Error in render: "InvalidCharacterError: String contains an invalid character"


StCyr:
### Expected behavior
Message should render properly

### Actual behavior
Some messages don't render at all and the following log appears:

```
[Vue warn]: Error in render: "InvalidCharacterError: String contains an invalid character"

found in

---> <ThreadEnvelope> at src/components/ThreadEnvelope.vue
       <AppContentDetails>
         <Thread> at src/components/Thread.vue
           <AppContent>
             <MailboxThread> at src/components/MailboxThread.vue
               <Content>
                 <Home> at src/views/Home.vue
                   <App> at src/App.vue
                     <Root> vue.runtime.esm.js:619
DOMException: String contains an invalid character
```

I can share the whole message 'source with you if you like

### Mail app

**Mail app version:** 1.5 (master)

StCyr:
fixed by https://github.com/nextcloud/mail/pull/3695

PR #3695, opened 2020-10-03
Replace btoa function by proper base64 encoder

### LINKED ISSUE #3693: Opening mail fails with: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
### Expected behavior
Mail is displayed correctly and no js error occurs

### Actual behavior
When i opened a mail i got the following js error and the email is not displayed.

mail.js:formatted:5776 DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at a.threadingFile (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:530:173675)
    at fn.get (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:26610)
    at fn.evaluate (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:27760)
    at a.threadingFile (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:29669)
    at a.r (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:813:2867)
    at a.e._render (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:35296)
    at a.r (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:64441)
    at fn.get (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:26610)
    at new fn (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:26528)
    at e (https://ichbinkeinreh.de/cloud/apps/mail/js/mail.js:36:64455)

### Mail app

**Mail app version:** v1.5.0

**Mailserver or service:** Strato

### Server configuration

**Operating system**: Ubuntu 20.04

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.4

**Nextcloud Version:** 20.0.0

#### Client configuration
**Browser:** Chrome 81

**Operating system:** Ubuntu 20.04

IchbinkeinReh:
Fixes #3693


ChristophWurst:
How does that make a difference?

IchbinkeinReh:
> Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.

As the error message says the browser internal btoa function is not capable to encode characters that are outside of Latin1 range. The package i added can handle arbitrary utf-8 strings and does therefore not have this problem

StCyr:
Compiled here and it solves the issues I have too (see https://github.com/nextcloud/mail/issues/3691 for my bug report)

IchbinkeinReh:
downgraded js-base64 to version 2.6.4 because of compatibility with internet explorer because the readme of js-base64 states that:
`In version 3.0 js-base64 switch to ES2015 module so it is no longer compatible with legacy browsers like IE [...]`

Also i replaced the atob function on account deserialization with Base64.decode because it has the same UTF-8 problem. In the atob case it does not throw an error but leads to garbage data. See this example from js-base64:
`let utf8  = '小飼弾'
Base64.encode(utf8);              // 5bCP6aO85by+
Base64.atob(        '5bCP6aO85by+');// 'å°é£¼å¼¾' which is nonsense`

package-lock.json
```diff
@@ -4744,7 +4744,7 @@
     },
     "domelementtype": {
       "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/domelementtype/-/domelementtype-1.3.1.tgz",
+      "resolved": "http://registry.npmjs.org/domelementtype/-/domelementtype-1.3.1.tgz",
```
ChristophWurst, line 4747: :-1:

ChristophWurst:
> downgraded js-base64 to version 2.6.4 because of compatibility with internet explorer because the readme of js-base64 states that:
> `In version 3.0 js-base64 switch to ES2015 module so it is no longer compatible with legacy browsers like IE [...]`

Then we should just transpile this one dependency. It's possible exclude this from the Babel exclude like we do for ckeditor: https://github.com/nextcloud/mail/blob/46605fd85835ad289be2fe46b73349751f68205b/webpack.common.js#L52

ChristophWurst:
Also please squash your commits into one once this is done :)

package-lock.json
IchbinkeinReh: Didnt change that manually. npm did that automatically. But yeah will undo this change

ChristophWurst: No worries. Just undo this :)

ChristophWurst (approved):
Awesome :v:

PR #3720, opened 2020-10-05
Migrate remaining usage of ILogger to the PSR logger interface

ChristophWurst:
A clean code is a fast code as cyclist say.


ChristophWurst:
Merging to make CI green :)

PR #3728, opened 2020-10-05
Fix displaying of mails with empty from field

IchbinkeinReh:
Signed-off-by: Patrick Bender <patrick@bender-it-services.de>

For example my ubuntu server sends such emails with empty from field when it sends mails for cron jobs that ran


ChristophWurst (approved):
Good catch!

PR #3744, opened 2020-10-07
Reset cache for mailbox on resetCache call

IchbinkeinReh:
After pressing resetCache in the frontend for a mailbox i had messages that disappeared from nextcloud mail after some time. When i disabled caching the issue was gone. I think PR might fix the cause

Signed-off-by: Patrick Bender <patrick@bender-it-services.de>


IchbinkeinReh:
Ok unfortunately this PR did not fix my problem of vanishing emails. Will disable memcache.local on my system for now until there are further hints what causes the issue

ChristophWurst:
> Ok unfortunately this PR did not fix my problem of vanishing emails.

closing thus

Issue #3743, opened 2020-10-07
Layout issue whith large To: field


StCyr:
### Expected behavior
It's 2020 and people still use the To: or Cc: field to list a large amount of email addresses. Mail should protect itself from that

### Actual behavior
When displaying a message with large To: or Cc: field, this field overlaps other elements on the page:

![Untitled](https://user-images.githubusercontent.com/8179031/95305535-63ec4300-0886-11eb-9ad0-a96b2e341ed7.png)

### Mail app

**Mail app version:** master (1.5)

ChristophWurst:
@jancborchardt @GretaD that is the theoretical issue we discussed about showing max one line with those avatars :)

StCyr:
I've fixed locally with:

```
.avatar-header {
        max-height: 80px;
        overflow-y: auto;
}
```

GretaD:
> I've fixed locally with:
> 
> ```
> .avatar-header {
>         max-height: 80px;
>         overflow-y: auto;
> }
> ```
Thank you, do you want to shot a PR? I should I do it?

StCyr:
No problem. I’ll do it

PR #3745, opened 2020-10-07
Prevent Thread's avatar header to overlap message's sender and subject.

StCyr:
Fixes https://github.com/nextcloud/mail/issues/3743

With this commit, the .avatar-header component is max 80px tall and in case content overflow a scrollbar will be available.

Untitled

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,121 @@
 {
   "name": "nextcloud-mail",
   "version": "1.7.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "@babel/cli": {
-      "version": "7.12.1",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.12.1.tgz",
-      "integrity": "sha512-eRJREyrfAJ2r42Iaxe8h3v6yyj1wu9OyosaUHW6UImjGf9ahGL9nsFNh7OCopvtcPL8WnEo7tp78wrZaZ6vG9g==",
-      "requires": {
-        "@nicolo-ribaudo/chokidar-2": "^2.1.8",
-        "chokidar": "^3.4.0",
+  "packages": {
+    "": {
+      "name": "nextcloud-mail",
+      "version": "1.7.0",
+      "license": "agpl",
+      "dependencies": {
+        "@bundle-analyzer/webpack-plugin": "^0.5.1",
+        "@ckeditor/ckeditor5-alignment": "~23.1.0",
+        "@ckeditor/ckeditor5-basic-styles": "~23.1.0",
+        "@ckeditor/ckeditor5-block-quote": "~23.1.0",
+        "@ckeditor/ckeditor5-build-balloon": "~23.1.0",
+        "@ckeditor/ckeditor5-dev-utils": "23.6.1",
+        "@ckeditor/ckeditor5-dev-webpack-plugin": "^23.6.1",
+        "@ckeditor/ckeditor5-editor-balloon": "~23.1.0",
+        "@ckeditor/ckeditor5-essentials": "~23.1.0",
+        "@ckeditor/ckeditor5-heading": "~23.1.0",
+        "@ckeditor/ckeditor5-link": "~23.1.0",
+        "@ckeditor/ckeditor5-paragraph": "~23.1.0",
+        "@ckeditor/ckeditor5-theme-lark": "~23.1.0",
+        "@ckeditor/ckeditor5-vue": "^1.0.3",
+        "@nextcloud/auth": "^1.3.0",
+        "@nextcloud/axios": "^1.5.0",
+        "@nextcloud/dialogs": "^3.1.1",
+        "@nextcloud/files": "^1.1.0",
+        "@nextcloud/initial-state": "^1.2.0",
+        "@nextcloud/l10n": "^1.4.1",
+        "@nextcloud/logger": "^1.1.2",
+        "@nextcloud/moment": "^1.1.1",
+        "@nextcloud/router": "^1.2.0",
+        "@nextcloud/vue": "^3.2.0",
+        "@nextcloud/vue-dashboard": "^1.0.1",
+        "@riophae/vue-treeselect": "^0.4.0",
+        "@vue/babel-preset-app": "^4.5.8",
+        "color-convert": "^2.0.1",
+        "core-js": "^3.7.0",
+        "davclient.js": "git+https://github.com/owncloud/davclient.js.git#0.2.1",
+        "debounce-promise": "^3.1.2",
+        "dompurify": "^2.2.2",
+        "html-to-text": "^5.1.1",
+        "ical.js": "^1.4.0",
+        "iframe-resizer": "^4.2.11",
+        "js-base64": "^3.6.0",
+        "lodash": "^4.17.20",
+        "md5": "^2.3.0",
+        "nextcloud_issuetemplate_builder": "^0.1.0",
+        "postcss-loader": "^3.0.0",
+        "printscout": "2.0.3",
+        "ramda": "^0.27.0",
+        "raw-loader": "^4.0.2",
+        "stylelint": "^13.7.2",
+        "v-tooltip": "^2.0.3",
+        "vue": "^2.6.12",
+        "vue-autosize": "^1.0.2",
+        "vue-click-outside": "^1.1.0",
+        "vue-infinite-scroll": "^2.0.2",
+        "vue-material-design-icons": "^4.11.0",
+        "vue-on-click-outside": "^1.0.3",
+        "vue-reactive-refs": "^0.0.2",
+        "vue-router": "^3.4.9",
+        "vue-scroll": "^2.1.13",
+        "vue-shortkey": "^3.1.7",
+        "vue-slide-up-down": "^2.0.1",
+        "vue-tabs-component": "^1.5.0",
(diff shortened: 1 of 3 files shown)


jancborchardt:
Could we force the bubbles to be max 1 line, and say e.g.:

> Christoph Wurst, Greta Doci, Cyrille Bollu & 14 more

The "& 14 more" could be a bubble too and open an ActionMenu which could show the other people.

That would look much nicer than another scrollable area. :)

StCyr:
> Could we force the bubbles to be max 1 line, and say e.g.:
> 
> > Christoph Wurst, Greta Doci, Cyrille Bollu & 14 more
> 
> The "& 14 more" could be a bubble too and open an ActionMenu which could show the other people.
> 
> That would look much nicer than another scrollable area. :)

That's a very good idea I think.

I'll have a look at this

StCyr:
hmmm it's not as easy as `text-overflow: ellipsis` :-D

StCyr:
What about just showing the avatars with no text? 

Since they have fixed size, we can calculate before rendering the max amount of avatars to show

StCyr:
It works with the help of the `v-reactive-refs` library.

There's stil an issue when one scrolls the message verticaly though 😢

StCyr:
This works quite well now. A review @ChristophWurst or @st3iny ?

st3iny (commented):
Very nice! However, I have 2 pieces of feedback:

1. The display of the show more bubble is kind of hard to read. Maybe insert line breaks after each email address?

header-display

2. The layout is not reactive. E.g. when increasing the window width it won't show more participants. When reducing the width the obsolete participants will vanish but the show more bubble won't appear.

StCyr:
>     1. The display of the show more bubble is kind of hard to read. Maybe insert line breaks after each email address?

I think I will be able to insert RecipientBubbles directly un the pop over; v-tooltip is awesome 🏆 

>     1. The layout is not reactive. E.g. when increasing the window width it won't show more participants. When reducing the width the obsolete participants will vanish but the show more bubble won't appear.

It's now reactive

StCyr:
I think I'm done with this:

image

Can someone use some css-fu to better style the **+38** span? @st3iny @jancborchardt

GretaD:
It looks very nice. Same small things:
1. The color of `avatar-more` should be `var(--color-background-dark)` not hard coded
2. The cursor when you hover over the +38(avatar-more) should be a pointer.
3. The popover width should be just a bit wider so instead of having a long list we have a shorter one, so we can get as much info as we can without scrolling. (This can also be done with a follow up PR later)

GretaD (approved):
Very niceee.

Please merge after Jan's review

StCyr:
@jancborchardt, review please?

package-lock.json
```diff
@@ -1,91 +1,15130 @@
 {
   "name": "nextcloud-mail",
   "version": "1.7.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
ChristophWurst, line 4: what version of npm do you use?

StCyr: 7.0.9 (Debian bullseye)

ChristophWurst: Did you install a custom version? https://blog.npmjs.org/post/631877012766785536/release-v700 it's been release a month ago but my node v15.1.0 still comes with npm 6.14.8.

What's the output for
```
node --version
which npm
```
?

We'll have to migrate the lockfile at some point but I'm curious why you have the new npm already.

StCyr: cyrille@vm-cyrille-debian:/var/www/nextcloud/apps/mail$ npm --version
7.0.9
cyrille@vm-cyrille-debian:/var/www/nextcloud/apps/mail$ which npm
/usr/bin/npm
cyrille@vm-cyrille-debian:/var/www/nextcloud/apps/mail$ dpkg -S /usr/bin/npm
npm: /usr/bin/npm
cyrille@vm-cyrille-debian:/var/www/nextcloud/apps/mail$ apt show npm
Package: npm
Version: 7.0.11+ds-1
...

st3iny (commented):
I really like the new popover :+1: 

There still is an issue with the responsiveness when resizing: Sometimes, the show more bubble won't show at all and sometime it flickers while resizing.

thread-header-flicker

StCyr:
There's no need for the vue-reacive-ref package eventualy

st3iny (approved):
Works fine now! Very nice :+1: 

My npm is not happy about the lock file though: `This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!`.

My linux distro is still on the npm cli LTS branch (6.x).

StCyr:
> My npm is not happy about the lock file though: `This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!`.

I've tried to revert the file to lockfileVersion = 1. Can you test again?

st3iny:
The error is gone but the diff of the lock is huge. Could you reset this file to the version from master?

@ChristophWurst How shall we proceed with the npm update? This breaking change of the lockfile version is kind of awkward for open source projects because we are all using different operating systems and linux distros.

StCyr:
> @ChristophWurst How shall we proceed with the npm update? This breaking change of the lockfile version is kind of awkward for open source projects because we are all using different operating systems and linux distros.

Apparently one needs nvm (https://stackoverflow.com/questions/64813775/is-there-any-way-to-fix-package-lock-json-lockfileversion-so-npm-uses-a-specific)

StCyr:
> The error is gone but the diff of the lock is huge. Could you reset this file to the version from master?

Yes, you're right. I'll do that

st3iny:
> Apparently one needs nvm (https://stackoverflow.com/questions/64813775/is-there-any-way-to-fix-package-lock-json-lockfileversion-so-npm-uses-a-specific)

Nvm seems to be a good solution especially for new maintainers. Didn't know about this tool until now.

If npm v7 gets mainstream support on all operating systems we can switch again I guess.

st3iny:
@jancborchardt Could you review this feature again please? All of the bugs are ironed out and your approval is required for merging :)

src/components/Thread.vue
```diff
@@ -94,8 +130,43 @@ export default {
 	},
 	created() {
 		this.resetThread()
+		window.addEventListener('resize', this.updateParticipantsToDisplay)
```
ChristophWurst, line 133: should we debounce this a bit? I'm afraid a browser will actually fire this event many times as the user resizes. And so for every pixel this is recalculated

src/components/Thread.vue
```diff
@@ -121,6 +192,8 @@ export default {
 		async resetThread() {
 			this.expandedThreads = [this.threadId]
 			await this.fetchThread()
+			// this.$nextTick(() => this.updateParticipantsToDisplay())
```
ChristophWurst, line 195: dead code -> :fire:

src/components/Thread.vue
```diff
@@ -304,6 +377,26 @@ export default {
...
+	padding: 0px 0px 1px 1px;
+	border-radius: 10px;
+	cursor: pointer;
+}
+.avatar-hidden {
+	visibility: hidden;
+}
+.popover-inner {
+	max-width: 500px;
+	border-radius: 10px;
```
ChristophWurst, line 397: obsolete once we use the popover from nc/vue

ChristophWurst (changes requested):

src/components/Thread.vue
StCyr: My tests with FF latest didn't show so much events fired... Could you test a little too?

src/components/Thread.vue
```diff
@@ -30,6 +50,7 @@
 <script>
 import AppContentDetails from '@nextcloud/vue/dist/Components/AppContentDetails'
 import { prop, uniqBy } from 'ramda'
+import { VPopover } from 'v-tooltip'
```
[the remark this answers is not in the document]
StCyr: will do

src/components/Thread.vue
ChristophWurst: Let's please add the debounce in any case. We can't test this with every possible browser. You can use https://lodash.com/docs/4.17.15#debounce for this.

src/components/Thread.vue
StCyr: the `max-width` attribute still looks necessary to me

src/components/Thread.vue
StCyr: debounced with a 500ms delay

src/components/Thread.vue
ChristophWurst: right :)

ChristophWurst (approved):
Code looks excellent

Please squash into one commit :)

src/components/Thread.vue
```diff
@@ -94,8 +122,43 @@ export default {
 	},
 	created() {
 		this.resetThread()
+		window.addEventListener('resize', debounce(500, this.updateParticipantsToDisplay))
+	},
+	beforeDestroy() {
+		window.removeEventListener('resize', debounce(500, this.updateParticipantsToDisplay))
```
ChristophWurst, line 128: this doesn't actually work because every time you call `debounce(500, this.updateParticipantsToDisplay)` you will get a *new function*. See what we did with `saveDraft` and `saveDraftDebounced` in `Composer.vue`. Hope this makes it more clear.

StCyr: should be good now

GretaD (approved):
:100:

PR #3751, opened 2020-10-08
Fix invalid falsable returns detected by Psalm

ChristophWurst:
This could potentially have caused a bug.


rullzer (approved):

eneiluj (approved):
Let's suggest a new return type to Php:
``` php
fun() : ?!string
```
:wink:

ChristophWurst:
> ```
> fun(): ?!string
> ```

with php8 we can in fact type this with a union type. Like I think `?string|boolean` should work

PR #3753, opened 2020-10-08
Update psalm config and add a workflow

ChristophWurst:
Nothing enforced yet, but just getting this all set up so checks can run for every commit.


rullzer (approved):
:woman_dancing:

PR #3756, opened 2020-10-08
Show important/starred/read status inside a thread

### LINKED ISSUE #3626: Show important/starred/read status inside a thread
### Feature Request

We reflect certain flags visually in the envelope list. We should do the same in the thread view.

Bildschirmfoto von 2020-09-22 10-18-41

^ this message is important. It should show the important icon here as well.

#### Summary

n/a

GretaD:
fixes #3626 

- [x] important 
//fix the clicking problem, that removes the important icon instead of opening the message

- [x] unread

- [x] favorite

note: Thread has the recipient name  in bold, envelope has normal text, maybe we should have normal text also in the thread

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -27,10 +27,22 @@
 				:display-name="envelope.from[0].label"
 				:disable-tooltip="true"
 				:size="40" />
+			<div
+				v-if="envelope.flags.important"
+				class="app-content-list-item-star icon-important"
+				:data-starred="envelope.flags.important ? 'true' : 'false'"
+				@click.prevent="onToggleImportant"
+				v-html="importantSvg" />
+			<div
+				v-if="envelope.flags.flagged"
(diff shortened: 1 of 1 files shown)


src/components/ThreadEnvelope.vue
```diff
@@ -375,9 +389,44 @@ export default {
 	}
 	.left {
 		flex-grow: 1;
+		//TODO: fix the important svg size
+		//hack to fix the important.svg overlapping the sender and subject.
+		z-index: 10000;
```
GretaD, line 394: @skjnldsv i need you help. For some reasons the svg of the important icon takes too much space. When i hover the thread, it is actually hovered the important icon. Can you please check whats wrong? i added z-index but doesn't really fix it.

skjnldsv: Might be capped by parent zIndex?

skjnldsv (approved):

PR #3757, opened 2020-10-08
Fix remaining psalm errors, add baseline and CI test matrix for all Nextcloud versions

ChristophWurst:
@nickvergessen this is what we discussed a few week ago if all goes well. My OCP package will then only be required on CI. Here I'm leaving it for now because Scrutinizer still does our coverage tracking. But in general apps don't need to have it as default require anymore.


rullzer (approved):
:elephant:

PR #3761, opened 2020-10-08
Fix saving aliases

### LINKED ISSUE #3752: No Name in Alias, not in the configuration sreen, nor in a new mail after srored a new alias
When I enter a new alias (name and e-mail address) and save it, there is an empty field with a trash can above it. When I then write a new mail, I can choose the original name or an empty field.

### Mail app
MailApp 1.5.0

### Server configuration
Debian BusterApache
MariaDB
Nextcloud Version: 20

#### Client configuration
Firefox 81.0
Linux Mint

IchbinkeinReh:
fixes #3752

### DIFF
--- a/src/service/AliasService.js
+++ b/src/service/AliasService.js
@@ -2,13 +2,12 @@ import { generateUrl } from '@nextcloud/router'
 import axios from '@nextcloud/axios'
 
 export const createAlias = async(account, data) => {
-	const url = generateUrl('/apps/mail/api/aliases')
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Tip: Github does not create a link for issues references in the title. I would suggest to just use text in the title and add the reference to the commit description or the PR description. Then PR titles are also more human readable.

ChristophWurst (approved):
Makes sense. THanks a lot!

PR #3781, opened 2020-10-09
Set Psalm to error level 4 and update baseline file

ChristophWurst:
More errors to fix but also more to silence as they are out of scope for us.

Fixes were submitted separately as https://github.com/nextcloud/mail/pull/3779.

We need https://github.com/nextcloud/server/pull/23144 for many of them.

### DIFF
--- a/tests/psalm-baseline.xml
+++ b/tests/psalm-baseline.xml
@@ -1,14 +1,107 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="3.16@d03e5ef057d6adc656c0ff7e166c50b73b4f48f3">
-  <file src="lib/Controller/AvatarsController.php">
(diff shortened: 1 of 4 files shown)


tests/psalm-baseline.xml
```diff
@@ -39,4 +137,10 @@
       <code>$transport</code>
     </InvalidArgument>
   </file>
+  <file src="vendor/cerdic/css-tidy/class.csstidy.php">
+    <InvalidDocblock occurrences="2">
+      <code>public function parse_string_list($value) {</code>
```
ChristophWurst, line 142: unsure why psalm report this, but I sent a PR in any case https://github.com/Cerdic/CSSTidy/pull/61

kesselb (approved):

Issue #3824, opened 2020-10-16
Message body not properly formatted when forwarding emails


StCyr:
### Expected behavior
When forwarding an email, the body of the message should contain the original message separated from the new message (yet to be  written) by a line, usualy in the form "On <data>, <original sender> wrote:".

### Actual behavior
When forwarding emails, the body of the message is just a plain copy of the original message. 

### Mail app

**Mail app version:** master (1.5)

ChristophWurst:
As explained before this is a limitation of the editor. It will discard anything it can't deal with. Hence the simplified formatting. As soon as the editor can cope with more complex HTML, the forwarded message will come closer to the original.

StCyr:
> As explained before this is a limitation of the editor. It will discard anything it can't deal with. Hence the simplified formatting. As soon as the editor can cope with more complex HTML, the forwarded message will come closer to the original.

No, no, that discussion is not related to this issue.

The issue here is that this line isn't called when I forward a message:

https://github.com/nextcloud/mail/blob/b8867af8217da5c95a002c500150b1b39c126a6a/src/components/Composer.vue#L473

ChristophWurst:
I'm very sorry that I misunderstood the issue.

PR #3825, opened 2020-10-16
Fixes an issue where forwarded messages' body isn't built properly

StCyr:
Fixes https://github.com/nextcloud/mail/issues/3824

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -480,9 +480,9 @@ export default {
 				this.bodyVal = this.bodyWithSignature(
 					this.selectedAlias,
 					buildReplyBody(
-						this.editorPlainText ? toPlain(this.originalBody) : toHtml(this.originalBody),
-						this.original.from[0],
-						this.original.dateInt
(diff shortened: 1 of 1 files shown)


ChristophWurst:
I'll have to check this in detail. I recall that we have to be careful with what data we use. There was a bit of back and forth to get all things working. You could be totally right, but I'd like to be 100% sure.

src/components/Composer.vue
```diff
@@ -473,9 +473,9 @@ export default {
 				this.bodyVal = this.bodyWithSignature(
 					this.selectedAlias,
 					buildReplyBody(
-						this.editorPlainText ? toPlain(this.originalBody) : toHtml(this.originalBody),
```
ChristophWurst, line 476: `originalBody` doesn't even exist in this context :facepalm:

ChristophWurst: `original` neither

ChristophWurst (approved):
Yes!

PR #3828, opened 2020-10-16
OCA.Search is deprecated

StCyr:

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -110,7 +110,6 @@ export default {
 	},
 	data() {
 		return {
-			alive: false,
 			// eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


src/components/MailboxThread.vue
```diff
@@ -157,11 +157,6 @@ export default {
 	},
 	created() {
 		this.alive = true
-
-		window.addEventListener('DOMContentLoaded', (event) => {
-			// eslint-disable-next-line no-new
-			new OCA.Search(this.searchProxy, this.clearSearchProxy)
```
ChristophWurst, line 163: those two methods can be :fire:  as well, right?

ChristophWurst (changes requested):

ChristophWurst (approved):
Yay :)

PR #3831, opened 2020-10-17
Fixes 'New message' doesn't work from Drafts folder

StCyr:
The "New message" button doesn't work when clicked from the Drafts folder.

This PR solves this.

### DIFF
--- a/src/components/NewMessageDetail.vue
+++ b/src/components/NewMessageDetail.vue
@@ -144,12 +144,18 @@ export default {
 				logger.debug('detected navigation to current (new) draft UID, not reloading')
 				return
 			}
-			logger.debug('the  draft ID changed, we have to fetch the draft', {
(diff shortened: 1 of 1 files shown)


src/components/NewMessageDetail.vue
```diff
@@ -144,12 +144,18 @@ export default {
...
 			}
-			logger.debug('the  draft ID changed, we have to fetch the draft', {
-				currentId: this.draft.databaseId,
-				newId: to.params.draftId,
-			})
 
-			this.fetchMessage()
+			if (to.params.threadId === 'new') {
+				// User pressed the 'New message button'
+				this.draft = undefined
```
ChristophWurst, line 150: won't this discard the current draft without saving?

ChristophWurst: @StCyr :ping_pong:

PR #3832, opened 2020-10-17
fix a bug when navigating away from a draft that's not yet loaded

StCyr:
If you navigate away from a draft that's currently loading, but not loaded yet, an error is thrown on the browser console.

This PR fixes that

### DIFF
--- a/src/components/NewMessageDetail.vue
+++ b/src/components/NewMessageDetail.vue
@@ -145,7 +145,7 @@ export default {
 				return
 			}
 			logger.debug('the  draft ID changed, we have to fetch the draft', {
-				currentId: this.draft.databaseId,
+				currentId: this?.draft?.databaseId ?? 'no draft currently loaded',
 				newId: to.params.draftId,
 			})
 


src/components/NewMessageDetail.vue
```diff
@@ -145,7 +145,7 @@ export default {
 				return
 			}
 			logger.debug('the  draft ID changed, we have to fetch the draft', {
-				currentId: this.draft.databaseId,
+				currentId: (this.draft !== undefined) ? this.draft.databaseId : 'no draft currently loaded',
```
ChristophWurst, line 148:
```suggestion
				currentId: this?.draft?.databaseId: 'no draft currently loaded',
```
if you don't mind :)

ChristophWurst (approved):
Looks good!

src/components/NewMessageDetail.vue
StCyr: you're right: it's nicer this way.

PR updated

Issue #3597, opened 2020-09-17
Stuck context menu


rcargnelutti:
**Expected behavior**
When inside the “Inbox”, when the option “delete” in the context menu is selected, the context menu vanishes, and the message that was selected is immediately moved to the “Trash”.

**Actual behavior**
When I use the delete option, in the context menu, inside “Sent”, “Drafts” and “Trash”, the context menu gets stuck. It won’t vanish until I click outside the menu area and the selected message stays in the box. After a while the message will appear on the “Trash”, not immediately as expected.

**Mail app version**: 1.4.1
**Mail Server or service**: MXRoute (IMAP)
**Server configuration**
**Operating system**: CentOS 7.6
**Web server**: Apache Version 2.4.46
**Database**: MySQL 10.3.24
**PHP version**: 7.3.6
**Nextcloud Version**: 19.0.3
**Client configuration**
**Browser**: Google Chrome Version 84 / 85 (official) 64 bits and Firefox Version 80.0.1 (64-bits)
**Operating system**: Ubuntu 20.04, Mint 20, Debian 10 (buster) and Windows 10 Pro

ChristophWurst:
Possibly related to https://github.com/nextcloud/nextcloud-vue/issues/1384#issuecomment-690202738

StCyr:
can't reproduce this issue in the "Sent" and "Trash" folder with latest release.

But the "Drafts" folder seems to have problems idd

StCyr:
Hello @rcargnelutti ,

We have merged a change that should fix your issue. Next version of Mail (1.6) should see your issue resolved.

BR,

Cyrille

PR #3843, opened 2020-10-19
Make menus automaticaly close after clicking on an action button 

StCyr:
Fix to https://github.com/nextcloud/mail/issues/3597

Though I couldn't reproduce the issue here above, I've experienced this issue in the ThreadEnvelope component.

This PR makes the menu automaticaly close after clicking on **any** action button in ThreadEnvelope, EnvelopeList, and Envelope.


ChristophWurst (approved):
Awesome :)

PR #3845, opened 2020-10-19
Fix font menu size

GretaD:
before
accountmenub
after
accountmenu
before
threadmenub
after
threadmenu
before
mailboxmenub
After
mailboxmenu

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5818,7 +5818,7 @@
     },
     "domelementtype": {
       "version": "1.3.1",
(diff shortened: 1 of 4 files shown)


ChristophWurst:
I'm not sure. Why do we need to have our own font size for this? Is there no standard from nc-vue?

GretaD:
> I'm not sure. Why do we need to have our own font size for this? Is there no standard from nc-vue?

it is font-size=normal i think. But it is not applying to all our actions in menus. I agree, it should be nc-vue standard.

skjnldsv:
Yeah, please fix it from the components directly :)

GretaD:
sure

PR #3849, opened 2020-10-20
Update the authors list after v1.5

ChristophWurst:
Ref https://github.com/nextcloud/mail/graphs/contributors?from=2020-06-16&to=2020-10-03&type=c

This was originally based on that data, so let's keep it up to date. @rullzer you should contribute more :P

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -15,9 +15,10 @@
 	<version>1.5.0</version>
 	<licence>agpl</licence>
 	<author>Christoph Wurst</author>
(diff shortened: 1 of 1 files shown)


StCyr (approved):

ChristophWurst:
Wait … wrong time range. It's actually https://github.com/nextcloud/mail/graphs/contributors?from=2020-06-17&to=2020-10-03&type=c

But then we have two on rank 5 :thinking:

ChristophWurst:
@dehnhardt welcome to the club :)

Issue #2733, opened 2020-03-07
Ability to load images in signatures 


JohnMaree:
### Feature Request

The ability for signatures for each account to have images is a common feature in most email clients.

#### Summary

The ability for signatures for each account to have images is a common feature in most email clients. 
As an example:
![compact-logo-1](https://user-images.githubusercontent.com/12199888/76165774-6d7fc200-6162-11ea-998b-d780a1284790.png)

ChristophWurst:
As briefly discussed in the dev channel, we need a permanent storage for these images. And then transform them into attachments when the messages is sent.

As a step in between we could start with external images only. This isn't ideal. But for some cases like corporate logos, this might actually just do the job.

ChristophWurst:
Another thing that we *could* do is use ckeditor's base64 "storage". Of course this bloats messages, but for small images (like, again, logos) this could be fine. Maybe worth a consideration? :thinking:

JohnMaree:
> Another thing that we _could_ do is use ckeditor's base64 "storage". Of course this bloats messages, but for small images (like, again, logos) this could be fine. Maybe worth a consideration? 🤔

Good idea! What makes this more appealing is the fact that typically people stray away form large signatures as loading them becomes a pain if they are too large. So one could always specify a size limit?

ChristophWurst:
> So one could always specify a size limit?

Absolutely!

hanserasmus:
Would it make sense to have an interim solution where the image is hosted on a third-party site like imgur.com, and have the URL pasted in the signature so that it shows up and "looks like" the image is added to the signature itself? Something similar to what google is allowing in it's gmail settings? Just until you have figured out the way forward into making it a clean integration to core code?

ChristophWurst:
We could allow external images already, I think.

It's basically what I meant to explain with

> As a step in between we could start with external images only. This isn't ideal. But for some cases like corporate logos, this might actually just do the job.

The app wouldn't need to care where the images are stored. We'd just need to find a way to not make the browser block these images with our strict CSP.

hanserasmus:
If you could allow them already, it would go a long way for users! I understood what you meant, just wanted clarification. Because we could create an image with all the details normally found in the signature, like the example @JohnMaree posted in the OP, and just link that in there. 

I tried adding images a while back (like 3 weeks or so) and it was not working, but might be that one needs a specific syntax for the mail client/editor to understand this imported image?

CATER-COM:
[...]

PR #3858, opened 2020-10-21
Add base64 image insert support

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2733 in the most simple form.

I also experimented with external images but the issue is that while composing you won't see them because the direct requests to other domains is blocked by CSP.

As a small enhancement I want to try the workaround from https://github.com/ckeditor/ckeditor5/issues/5657 to limit the size of the pasted images as it bloats the signatures/messages quite a bit. Unfortunately the base64 plugin has no size limit option.


dumblob:
Do you think the size limit would be a reason to postpone this feature? For me I have to say we don't care much as even if it's megabytes, we'd go for it as recipients are "used to slow email openings".

ChristophWurst:
Please leave the decision of when to ship a feature to the maintainers of the application :v:

ChristophWurst:
Turns out this is a bad idea: https://github.com/nextcloud/mail/issues/2733#issuecomment-720989398

Issue #3859, opened 2020-10-21
Cannot save several attachments to File


StCyr:
### Expected behavior
When you have an email with several attachments, there's a button "Save all to File". Clicking on this button should save all attachments to the selected location.

### Actual behavior
Attachments are not saved.

### Mail app

**Mail app version:** master (1.5)

### Server configuration

**Nextcloud Version:** 20.0.0

#### Client configuration
**Browser:** FF latest

ChristophWurst:
check the network log and nextcloud.log. any errors?

PR #3879, opened 2020-10-23
Fix to cannot save several attachments to Files

StCyr:
This fixes https://github.com/nextcloud/mail/issues/3859

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -462,7 +462,7 @@ class MessagesController extends Controller {
 		if ($attachmentId === '0') {
 			// Save all attachments
 			/* @var $m IMAPMessage */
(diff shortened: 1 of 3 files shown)


lib/Controller/MessagesController.php
```diff
@@ -462,7 +462,7 @@ public function saveAttachment(int $id,
 		if ($attachmentId === '0') {
 			// Save all attachments
 			/* @var $m IMAPMessage */
-			$m = $folder->getMessage($id);
+			$m = $folder->getMessage($message->getUid());
```
ChristophWurst, line 465: good catch!

ChristophWurst (approved):

ChristophWurst:
rebased for latest CI

PR #3897, opened 2020-10-26
Add the posibility to move mailboxes

GretaD:

### DIFF
--- a/src/components/MailboxPicker.vue
+++ b/src/components/MailboxPicker.vue
@@ -18,9 +18,9 @@
 				</div>
 			</span>
 			<div class="mailbox-list">
-				<ul v-if="mailboxes.length > 0">
+				<ul v-if="filteredMailboxes.length > 0">
 					<li
-						v-for="box in mailboxes"
+						v-for="box in filteredMailboxes "
 						:key="box.databaseId"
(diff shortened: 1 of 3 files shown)


ChristophWurst:
To bring this to life

* Update to latest master once https://github.com/nextcloud/mail/pull/3943 is in
* Update `nextcloud/vue` to 3.1
* Move the modal into the `#extra` slot of the app nav item

Then it will properly show for all mailboxes :)

GretaD:
@ChristophWurst should i also rename the movemodal.vue into movemessagemodal.vue and include here or as a separate PR?

src/components/MailboxPicker.vue
```diff
@@ -77,6 +77,14 @@ export default {
 			type: String,
 			default: t('mail', 'Choose'),
 		},
+		pickedMailbox: {
+			type: Object,
+			required: false,
```
ChristophWurst, line 82: needs a default (as function) https://vuejs.org/v2/guide/components-props.html#Prop-Validation

ChristophWurst:
Needs rebase

ChristophWurst (approved):

ChristophWurst:
Minor detail I noticed while testing https://github.com/nextcloud/mail/issues/4045

PR #3913, opened 2020-10-30
Sort phpunit tests by defect

ChristophWurst:
Then the failing ones will show right away when re-running the tests
locally.

As seen at https://laravel-news.com/tips-to-speed-up-phpunit-tests.

### DIFF
--- a/tests/phpunit.integration.xml
+++ b/tests/phpunit.integration.xml
@@ -25,6 +25,7 @@
 		 timeoutForSmallTests="900"
 		 timeoutForMediumTests="900"
 		 timeoutForLargeTests="900"
+		 cacheResult="true"
 		>
 	<testsuite name='Mail app tests'>
 		<directory suffix='test.php'>Integration</directory>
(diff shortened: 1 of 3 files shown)


tests/phpunit.integration.xml
```diff
@@ -25,6 +25,7 @@
 		 timeoutForSmallTests="900"
 		 timeoutForMediumTests="900"
 		 timeoutForLargeTests="900"
+		 cacheResult="true"
```
kesselb, line 28: That's only for localhost I guess.

ChristophWurst: We have the cache file in .gitignore anyway. It doesn't make a difference for CI as the file doesn't exist there.

kesselb:
> It is also a good idea to add the cache file .phpunit.result.cache to your .gitignore so that it does not end up being committed to your repository.

kesselb (approved):

tests/phpunit.integration.xml
ChristophWurst: Turns out this is even `true` by default now https://phpunit.readthedocs.io/en/9.3/configuration.html

PR #3943, opened 2020-11-02
Revert partial update of CKEditor to v23.1

ChristophWurst:
Dependabot was motivated this weekend and bumped some packages of CKEditor from 23.0 to 23.1. But only *some*. Hence on master you get a duplicate module error and the page won't render.

This reverts the bumps. I'll have to do a manual round of updates later.


st3iny (approved):
Tested it on a clean environment (`make clean && make dev-setup`) and ckeditor works as expected.

PR #3944, opened 2020-11-02
Migrate the phpunit config

ChristophWurst:
Phpunit suggests this, so let's do it.

>  Warning:       Your XML configuration validates against a deprecated schema.
> Suggestion:    Migrate your XML configuration using "--migrate-configuration"!


ChristophWurst:
coverage breaks

PR #3946, opened 2020-11-02
Build a dev version for a dev build, not a production version

ChristophWurst:


st3iny (approved):
Faster dev-setup :rocket:

Issue #1025, opened 2018-08-11
Remember expanded/collapsed state of navigation folders


trueshanti:
Please add a option to autoexpand the 1st-level folder view.

Even with only the default (IMAP-)folders (favorites, sent, drafts, spam, trash) only the first 3 folders are visible .. i have to expand the view to see the rest on every reload.

#

Raising the visible entries from 3 to 6 is not what i would like to see .. i would prefer an option to show all folders that reside on the main-level.

thanks for support

stefan-niedermann:
Maybe instead just remembering which folders are expanded is a better solution?

trueshanti:
@stefan-niedermann that would be even more gorgeous

PR #3952, opened 2020-11-03
Store collapsed-folder settings

brueckner:
_Hi! This is my first Open Source PR and I am excited! If you have any suggestions, they are warmly welcomed!_

---

## Problem
The UI doesn't store the last (un-)collapsed state of the folders for an account. If the user is making heavy use of email folders, they'll most likely want to have them uncollapsed at all times. At the moment they have to click on "Show all folders" on every reload/revisit.

## Suggested Solution
Store the last chosen state in the database.

## Approach
First I was thinking about a simple flag in `oc_mail_accounts` and while that would certainly work, it would "pollute" the table with UI settings, which gets worse the more settings are added down the line. Instead I stored the settings in `oc_preferences`, which is already used by the mail app to store the `collect-data` flag for example.

I though it might be useful to be able to store more than just the state of collapsible folders down the line, so tried to make it extendable by storing a JSON object like that:

| userid   | appid | configkey                   | configvalue |
|----------|-------|------------------------|--------------------------|
| username | mail  | account-settings | `{"1":{"accountId":1,"collapsed":false,"some-other-setting":true},"2":{"accountId":2,"collapsed":true}}`|

And that's basically it. The `account-settings` are now stored when using the toggle button. Upon reload, the [`PageController.php`](https://github.com/nextcloud/mail/compare/master...dotplex:enhancement/store-collapsed-folder-settings?expand=1#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9b) loads the settings and attaches them to [`templates/index.php`](https://github.com/nextcloud/mail/compare/master...dotplex:enhancement/store-collapsed-folder-settings?expand=1#diff-5258161aaab90de3916c094462a8b5035d08c02eff481989342aeca0a45c4be5) just like `collect-data` etc. The settings are then picked up in [`main.js`](https://github.com/nextcloud/mail/compare/master...dotplex:enhancement/store-collapsed-folder-settings?expand=1#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79) where they're merged with the actual account object. 

---

### Additional notes

- I wanted to pass an empty array as a default fallback to `getPreference()` in [`PageController.php`](https://github.com/nextcloud/mail/compare/master...dotplex:enhancement/store-collapsed-folder-settings?expand=1#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9b) like this: `$this->preferences->getPreference('account-settings', []);` – however... the return value would be `null` instead of the empty array. Is that intentional? Seems like a bug to me.
- I wasn't sure about the use of ternary operators here and there in terms of coding standards for this project. Please let me know if you want me to refactor these. 
- I wasn't able to set up the integration tests properly yet, so I can't tell if they would pass or not. Any help is much appreciated!

Fixes https://github.com/nextcloud/mail/issues/1025

### DIFF
--- a/src/components/NavigationAccountExpandCollapse.vue
+++ b/src/components/NavigationAccountExpandCollapse.vue
@@ -25,6 +25,7 @@
 
 <script>
 import AppNavigationItem from '@nextcloud/vue/dist/Components/AppNavigationItem'
+import logger from '../logger'
 
 export default {
 	name: 'NavigationAccountExpandCollapse',
@@ -51,8 +52,21 @@ export default {
 		},
 	},
 	methods: {
-		toggleCollapse() {
-			this.$store.commit('toggleAccountCollapsed', this.account.id)
+		async toggleCollapse() {
+			logger.debug('toggling collapsed folders for account ' + this.account.id)
+			try {
+				await this.$store.commit('toggleAccountCollapsed', this.account.id)
+				await this.$store
+					.dispatch('setAccountSetting', {
+						accountId: this.account.id,
+						key: 'collapsed',
+						value: this.account.collapsed,
+					})
+			} catch (error) {
+				logger.error('could not update account settings', {
+					error,
+				})
+			}
 		},
 	},
 }
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -148,6 +148,10 @@ export default {
 			return account
 		})
 	},
+	setAccountSetting({ commit, getters }, { accountId, key, value }) {
+		commit('setAccountSetting', { accountId, key, value })
+		return savePreference('account-settings', JSON.stringify(getters.getAllAccountSettings))
+	},
 	deleteAccount({ commit }, account) {
 		return deleteAccount(account.id).catch((err) => {
 			console.error('could not delete account', err)
@@ -166,6 +170,7 @@ export default {
 		console.debug(`mailbox ${prefixed} created for account ${account.id}`, { mailbox })
 		commit('addMailbox', { account, mailbox })
 		commit('expandAccount', account.id)
+		commit('setAccountSetting', { accountId: account.id, key: 'collapsed', value: false })
 		return mailbox
 	},
 	moveAccount({ commit, getters }, { account, up }) {
--- a/src/store/getters.js
+++ b/src/store/getters.js
@@ -31,6 +31,9 @@ export const getters = {
 	getAccount: (state) => (id) => {
 		return state.accounts[id]
 	},
+	getAllAccountSettings: (state) => {
+		return state.allAccountSettings
+	},
 	accounts: (state) => {
 		return state.accountList.map((id) => state.accounts[id])
 	},
--- a/src/store/mutations.js
+++ b/src/store/mutations.js
@@ -77,7 +77,7 @@ export default {
 		Vue.set(state.preferences, key, value)
 	},
 	addAccount(state, account) {
-		account.collapsed = true
+		account.collapsed = account.collapsed ?? true
 		Vue.set(state.accounts, account.id, account)
 		Vue.set(
 			state,
@@ -109,6 +109,16 @@ export default {
 	},
 	expandAccount(state, accountId) {
 		state.accounts[accountId].collapsed = false
+	},
+	setAccountSetting(state, { accountId, key, value }) {
+		const accountSettings = state.allAccountSettings.find(settings => settings.accountId === accountId)
+		if (accountSettings) {
+			accountSettings[key] = value
+		} else {
+			const newAccountSettings = { accountId }
+			newAccountSettings[key] = value
+			state.allAccountSettings.push(newAccountSettings)
+		}
 	},
 	addMailbox(state, { account, mailbox }) {
 		addMailboxToState(state, account, mailbox)
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -128,6 +128,8 @@ class PageController extends Controller {
 			$accountsJson[] = $json;
 		}
 
+		$accountSettings = $this->preferences->getPreference('account-settings', json_encode([]));
+
 		$user = $this->userSession->getUser();
 		$response = new TemplateResponse($this->appName, 'index',
 			[
@@ -136,6 +138,7 @@ class PageController extends Controller {
 				'accounts' => base64_encode(json_encode($accountsJson)),
 				'external-avatars' => $this->preferences->getPreference('external-avatars', 'true'),
 				'collect-data' => $this->preferences->getPreference('collect-data', 'true'),
+				'account-settings' => base64_encode($accountSettings),
 			]);
 		$this->initialStateService->provideInitialState(
 			Application::APP_ID,
--- a/src/main.js
+++ b/src/main.js
@@ -72,9 +72,21 @@ store.commit('savePreference', {
 	value: getPreferenceFromPage('collect-data'),
 })
 
+const accountSettings = JSON.parse(Base64.decode(getPreferenceFromPage('account-settings')))
 const accounts = JSON.parse(Base64.decode(getPreferenceFromPage('serialized-accounts')))
 accounts.map(fixAccountId).forEach((account) => {
-	store.commit('addAccount', account)
+	const settings = accountSettings.find(settings => settings.accountId === account.id)
+	if (settings) {
+		delete settings.accountId
+		Object.entries(settings).forEach(([key, value]) => {
+			store.commit('setAccountSetting', {
+				accountId: account.id,
+				key,
+				value,
+			})
+		})
+	}
+	store.commit('addAccount', { ...account, ...settings })
 })
 
 export default new Vue({
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -50,6 +50,7 @@ export default new Vuex.Store({
 			},
 		},
 		accountList: [UNIFIED_ACCOUNT_ID],
+		allAccountSettings: [],
 		mailboxes: {
 			[UNIFIED_INBOX_ID]: {
 				id: UNIFIED_INBOX_ID,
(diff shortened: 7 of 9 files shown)


ChristophWurst:
> I wanted to pass an empty array as a default fallback to `getPreference()` in [`PageController.php`](https://github.com/nextcloud/mail/compare/master...dotplex:enhancement/store-collapsed-folder-settings?expand=1#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9b) like this: `$this->preferences->getPreference('account-settings', []);` – however... the return value would be `null` instead of the empty array. Is that intentional? Seems like a bug to me.

Indeed. This should work.

> I wasn't able to set up the integration tests properly yet, so I can't tell if they would pass or not. Any help is much appreciated!

You can use the `make start-docker` to pull the test image and run it locally. Then you can use `composer run test:integration:dev` to invoke phpunit. This should work.

src/components/Navigation.vue
```diff
@@ -43,7 +43,7 @@
 									!group.account.collapsed ||
 									!isCollapsed(group.account, item)
 							"
-							:key="item.databaseId"
+							:key="'mailbox-' + item.databaseId"
```
ChristophWurst, line 46: looks like an unrelated change, no? and the databaseId should be unique enough. Did you just add this for clarity or to fix a bug?

src/components/NavigationAccountExpandCollapse.vue
```diff
@@ -53,6 +54,16 @@ export default {
 	methods: {
 		toggleCollapse() {
 			this.$store.commit('toggleAccountCollapsed', this.account.id)
+			this.$store
+				.dispatch('updateAccountSetting', {
+					accountId: this.account.id,
+					key: 'collapsed',
+					value: this.account.collapsed
```
ChristophWurst, line 61: tip: we use ESLint for the js formatting. You can `npm run lint:fix` to have everything in the desired format automatically :)

brueckner:
> You can use the `make start-docker` to pull the test image and run it locally. Then you can use `composer run test:integration:dev` to invoke phpunit. This should work.

Thanks for the info! I'll give that a go.

src/components/NavigationAccountExpandCollapse.vue
ChristophWurst: please use async and await (and a try-catch-finally) instead of explicit promises. You'll still find traces of promises in our code base but we generally try to use the modern coding conventions

src/store/getters.js
```diff
@@ -31,6 +31,9 @@ export const getters = {
 	getAccount: (state) => (id) => {
 		return state.accounts[id]
 	},
+	getAccountSettings: (state) => (id = null) => {
+		return id ? state.accountSettings[id] : state.accountSettings
```
ChristophWurst, line 35: I'm not sure if we also need the variant with id=null. Can we make the id mandatory and always return `state.accountSettings[id]`

src/store/mutations.js
```diff
@@ -77,7 +77,7 @@ export default {
 		Vue.set(state.preferences, key, value)
 	},
 	addAccount(state, account) {
-		account.collapsed = true
+		account.collapsed = account.collapsed !== undefined ? account.collapsed : true
```
ChristophWurst, line 80: `account.collapsed ?? true` for short

src/store/mutations.js
```diff
@@ -110,6 +110,11 @@ export default {
 	expandAccount(state, accountId) {
 		state.accounts[accountId].collapsed = false
 	},
+	updateAccountSetting(state, { accountId, key, value }) {
```
ChristophWurst, line 113: let's call it `setAccountSetting` because it also works with values that were not set before

ChristophWurst (changes requested):
Thanks for this and congrats to your first ever FLOSS contribution :)

I've left a few comments and remarks. Generally I'm on board with your approach. I too think that we shouldn't pollute the accounts table with too much UI-specifics and preferences and I like the approach with the serialized object.

What we'll need to pay attention to is phantom updates. Like when one process updates the preferences, we shouldn't overwrite the full object but just patch the one property that is sent. This will prevent concurrency issues when we later have more than one account preference.

Please also see the ticket that I linked. You're implementing a feature that was requested in the past.

src/components/Navigation.vue
brueckner: Ah, I forgot to mention that. Yea, this is unrelated indeed. I've been getting a duplicate key warning:

Mail_-_Nextcloud

ChristophWurst: Ah now I see the issue. We display accounts and mailboxes. Hence there are indeed conflicts because they can have an overlap of IDs.

Good catch. Mind submtiting this as separate PR? :)

brueckner: Not at all :) I'll take care of the other comments first and will submit this separately afterwards.

src/store/getters.js
brueckner: Hm, I am using it [here](https://github.com/nextcloud/mail/pull/3952/files#diff-e01eb159ce5ee0617f7ca141ba6abc5eaf0dd0e4e9c89bfd77979ff31f63c0b6R153) to grab the settings for all accounts to store them. Maybe a separate method like `getAllAccountSettings` for clarity's sake?

ChristophWurst: Oh I see. I didn't consider that `savePreference` is *global*, as in that it's not specific for one account. Hence you have to send this in a batch.

But yes, let's have a getter for this specific object and one for each account's settings for clarity :+1:

brueckner: Hm, so... I wasn't actually using `getAccountSettings()` for anything other than fetching ALL account settings after a refactor yesterday. I renamed the method to `getAllAccountSettings()` in order to be specific and removed the parameter. If a need for something like `getAccountSettings(account.id)` comes up, we could add it down the line. Right now it would be unused.

ChristophWurst: Makese sense, thanks

brueckner:
> > I wanted to pass an empty array as a default fallback to `getPreference()` in [`PageController.php`](https://github.com/nextcloud/mail/compare/master...dotplex:enhancement/store-collapsed-folder-settings?expand=1#diff-70e5309fcd3311d771b3db9c93490ea270fd5894769093765ec37edb68e5dd9b) like this: `$this->preferences->getPreference('account-settings', []);` – however... the return value would be `null` instead of the empty array. Is that intentional? Seems like a bug to me.
> 
> Indeed. This should work.

I am not sure if I have been seeing ghosts, but I just tried it again and it works as intended, so I [changed it](https://github.com/nextcloud/mail/pull/3952/commits/5bc84f5e08d58330940a1cb07d80ee11cc58059e) to use the proper fallback value instead of the ternary.

ChristophWurst (approved):
Code looks good but didn't test (yet) :)

st3iny (changes requested):

src/store/actions.js
```diff
@@ -148,6 +148,10 @@ export default {
 			return account
 		})
 	},
+	setAccountSetting({ commit, getters }, { accountId, key, value }) {
+		commit('setAccountSetting', { accountId, key, value })
+		savePreference('account-settings', JSON.stringify(getters.getAllAccountSettings))
```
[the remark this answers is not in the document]
ChristophWurst: We could use `Object.values()` indeed to strip the IDs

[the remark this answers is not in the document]
brueckner: Thanks for the feedback! I just noticed that doing a rebase right away is not the best for incremental reviews... 🙄 

Anyway, I decided to save it as an array of objects now, instead of an object with stringified ids. So the stored array now looks like this:

```json
[{"accountId":75,"collapsed":true},{"accountId":76,"collapsed":false}]
```

When merging the settings with the account (in main.js), I am simply grabbing the proper ones using `find`:

```javascript
const settings = accountSettings.find(settings => settings.accountId === account.id)
```

Most notably, [`setAccountSetting`](https://github.com/nextcloud/mail/pull/3952/commits/a150c9867256abb1a17ed4021d29c8f3a5285dbd#diff-7c0816590b8c43968bdb997a4d7ab7a803f17f160a132376e0b3aa1c9859d2f5R113-R122) has changed quite a bit in the process:

```javascript
setAccountSetting(state, { accountId, key, value }) {
	const accountSettings = state.allAccountSettings.find(settings => settings.accountId === accountId)
	if (accountSettings) {
		accountSettings[key] = value
	} else {
		const newAccountSettings = { accountId }
		newAccountSettings[key] = value
		state.allAccountSettings.push(newAccountSettings)
	}
},
```

... which takes care of updating existing settings or pushing new settings if there haven't been any for an account yet.

Lastly, I changed `state.accountSettings` to `state.allAccountSettings` for consistency.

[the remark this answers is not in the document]
st3iny: The settings are saved to the database and persist one reload. However, `state.allAccountSettings` is not set on load in `main.js` (only the account state via `addAccount()`).

You could do something like:

```diff
diff --git a/src/main.js b/src/main.js
index 409b7f6f..018906c2 100644
--- a/src/main.js
+++ b/src/main.js
@@ -76,6 +76,16 @@ const accountSettings = JSON.parse(Base64.decode(getPreferenceFromPage('account-
 const accounts = JSON.parse(Base64.decode(getPreferenceFromPage('serialized-accounts')))
 accounts.map(fixAccountId).forEach((account) => {
 	const settings = accountSettings.find(settings => settings.accountId === account.id)
+	if (settings) {
+		delete settings.accountId
+		Object.entries(settings).forEach(([key, value]) => {
+			store.commit('setAccountSetting', {
+				accountId: account.id,
+				key,
+				value,
+			})
+		})
+	}
 	store.commit('addAccount', { ...account, ...settings })
 })
 

```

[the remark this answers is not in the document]
brueckner: Thanks Richard, I'll take that as-is. :-)

I made the change and pulled in the latest changes from master.

st3iny (approved):
Works fine

PR #3962, opened 2020-11-03
Drop the unused IMailBox interface

ChristophWurst:
We don't use this anywhere. Time to :fire: it.

- [x] Requires https://github.com/nextcloud/mail/pull/3945


GretaD (approved):
:droplet:

PR #3963, opened 2020-11-03
Remove the composer intall from the Makefile

ChristophWurst:
It's better to have composer installed globally anyway. And the latest composer doesn't work with Mail. So I'd say the Makefile can do without.


StCyr (approved):

StCyr:
👍

PR #3964, opened 2020-11-03
Remove dead methods from the Mailbox class

ChristophWurst:
Because nobody needs dead code.


StCyr (approved):
'Had to rebase it here localy (to benefit from the very-fresh Makefile changes).

It works though

Issue #3061, opened 2020-05-05
Edit/compose as new


ChristophWurst:
### Feature Request

Allow users to compose a new message based on an existing one. Should most likely be an action on the envelope or the actions in the message details.

#### Summary

n/a

ChristophWurst:
1) Proof that THunderbird has this 
![Bildschirmfoto vom 2020-11-03 20-13-14](https://user-images.githubusercontent.com/1374172/98030487-5b4d4500-1e11-11eb-8228-8a852a86084d.png)

2) Proof that Evolution has this
![Bildschirmfoto vom 2020-11-03 20-13-40](https://user-images.githubusercontent.com/1374172/98030503-64d6ad00-1e11-11eb-8160-ee49fc001923.png)

3) Proof that outlooks has this: https://superuser.com/a/579642
4) Proof that iMail has this:
![](https://ericleeclark.com/wp-content/uploads/2018/11/Screen-Shot-2018-11-10-at-6.43.36-AM-854x385.png)

jancborchardt:
Thanks for the references! Then let’s call it "Edit as new message".

(Also from a list we take as references, let’s remove Evolution and instead take Gmail.)

PR #3965, opened 2020-11-03
Allow composing new messages based on existing ones

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3061


ChristophWurst:
@jancborchardt three questions

1) At what position should we place this new item?
2) What should be the name of the action?
3) Which icon should we use?

Here's my draft:
Bildschirmfoto von 2020-11-03 19-54-28

st3iny (approved):
Code looks good. Tested on my local instance and it works well.

PR #3967, opened 2020-11-03
Translate special mailbox labels in tooltips

brueckner:
Small improvement to translate the title/label of a mailbox tooltip:

Pasted_Image_03_11_20__19_50

This change will pass `mailbox.name` through the translation interface, but only if the mailbox has a `specialRole`. That way, custom folder names remain untouched.

Currently, the following string translations are missing: `INBOX`, `Deleted Messages`, `Sent Messages`.

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -38,7 +38,7 @@
 				<ActionText
 					v-if="!account.isUnified && mailbox.specialRole !== 'flagged'"
 					icon="icon-info"
-					:title="mailbox.name">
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Erm … this is actually intended. It's supposed to show the "real" name of the mailbox. The translated one is already visible on the left. Otherwise this is just duplicating the label without any benefit :wink:

brueckner:
Oh, alright! I see the point. Shall I just close this, or what's the deal with the "0. to triage" label?

ChristophWurst:
yeah guess I'll close. I set the triage label so I later know this is a PR were we are unsure if it makes sense or can be finished :)

PR #3969, opened 2020-11-03
Catch no account in envelope and console log spam

ChristophWurst:
For the past few weeks we've seen LOADS of `this.account is undefined` errors in the browser console. I really don't get why this happens. But I also want the error to go away. So here's my highly sophisticated *fix*.


GretaD (approved):
farewell `this.account` error, farewell foreeeever

Issue #3916, opened 2020-10-30
Switching from one account's setting to the other doesn't update/clear all UI elements


ChristophWurst:
Noticed in https://github.com/nextcloud/mail/pull/3820 by @GretaD, e.g. the selected mailboxes do not update, tho you can't cross-select one account's mailboxes on the other. It's weird, tho.

PR #3986, opened 2020-11-04
Prevent account settings reuse when account changes

ChristophWurst:
Vue is clever and always tries to recycle a component when certain
criteria is met. Apparently it also does it for components with objects
as props, which led to the problem that when you navigated from one
account's settings to another account's setting, the mailbox default
didn't update and you actually saw mailboxes of the previous account.

By key-ing the components we can express that the rendering depends on
the account ID and every time this value changes, Vue has to throw away
the instance and render a new one.

Fixes https://github.com/nextcloud/mail/issues/3916

Ref https://michaelnthiessen.com/force-re-render/

### DIFF
--- a/src/views/AccountSettings.vue
+++ b/src/views/AccountSettings.vue
@@ -12,11 +12,11 @@
 						href="#account-form"
 						:title="t('mail', 'Change name')" />
 				</p>
-				<AliasSettings v-if="!account.provisioned" :account="account" />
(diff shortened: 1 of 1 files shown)


st3iny (commented):
Tested on local instance and works as intended. Eslint is not happy though.

GretaD (approved):
fixed the lint

PR #3989, opened 2020-11-04
Revert partial update of CKEditor to v23.1 again

ChristophWurst:
Follow-up for https://github.com/nextcloud/mail/pull/3943 because it happened again …


GretaD (approved):
yes, this fixes it.

Issue #3994, opened 2020-11-04
Mails not syncing (iconv() exception)


pReya:
### Expected behavior
Mails should start to sync and should be visible in my inbox.

### Actual behavior
Inbox stays empty, syncing never finishes. Mails don't show up.

### Mail app

**Mail app version:**
1.5.2

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)
Uberspace (https://manual.uberspace.de/mail-mailboxes.html)

### Server configuration

**Operating system**: (e.g. Debian 8)
CentOS 7

**Web server:** (e.g. Apache, Nginx,...)
Apache

**Database:** (e.g. MariaDB, SQLite or PostgreSQL)
MariaDB

**PHP version:** (e.g. 7.0)
7.4

**Nextcloud Version:** (see admin page, e.g. 13.0.2)
20.0.1

#### Client configuration
**Browser:** (e.g. Firefox 48)
Firefox 82

**Operating system:** (e.g. Arch Linux)
Mac OS 10.15

My log shows the following error:

```
[mail] Error: OCA\Mail\Exception\ServiceException: Sync failed for 2:INBOX: iconv() expects parameter 3 to be string, null given at <<closure>>

0. /var/www/virtual/username/html/apps/mail/lib/Service/Sync/SyncService.php line 121
   OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 16}, OC\AppFramework\ScopedPsrLogger {}, 42, [], true)
1. /var/www/virtual/username/html/apps/mail/lib/Controller/MailboxesController.php line 161
   OCA\Mail\Service\Sync\SyncService->syncMailbox(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 16}, 42, [], false, "is:important")
2. /var/www/virtual/username/html/lib/private/AppFramework/Http/Dispatcher.php line 169
   OCA\Mail\Controller\MailboxesController->sync(16, [], true, "is:important")
3. /var/www/virtual/username/html/lib/private/AppFramework/Http/Dispatcher.php line 100
   OC\AppFramework\Http\Dispatcher->executeController(OCA\Mail\Control ... {}, "sync")
4. /var/www/virtual/username/html/lib/private/AppFramework/App.php line 152
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\Mail\Control ... {}, "sync")
[...]

PR #3995, opened 2020-11-04
Handle null email address before iconv conversion

ChristophWurst:
`iconv` doesn't accept `null`.

Fixes https://github.com/nextcloud/mail/issues/3994


st3iny (approved):
This one is hard to test but the code looks good :D

Issue #3057, opened 2020-05-05
Enforce link shares over attachments after a certain threshold


ChristophWurst:
### Feature Request

There should be an option to prevent users from attaching large files. Link shares should be used instead.

#### Summary

n/a

PR #3998, opened 2020-11-04
Allow admins to limit the size of attachements

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3057

Todo
- [x] Front-end size check
- [x] Back-end setting
- [x] Sane default value?

### DIFF
--- a/src/components/ComposerAttachments.vue
+++ b/src/components/ComposerAttachments.vue
@@ -47,13 +47,19 @@
 import map from 'lodash/fp/map'
 import trimStart from 'lodash/fp/trimCharsStart'
 import { getRequestToken } from '@nextcloud/auth'
-import { translate as t } from '@nextcloud/l10n'
-import { getFilePickerBuilder } from '@nextcloud/dialogs'
+import { formatFileSize } from '@nextcloud/files'
+import prop from 'lodash/fp/prop'
+import { getFilePickerBuilder, showWarning } from '@nextcloud/dialogs'
+import sum from 'lodash/fp/sum'
+import sumBy from 'lodash/fp/sumBy'
+import { translate as t, translatePlural as n } from '@nextcloud/l10n'
+
 import Vue from 'vue'
 
-import Logger from '../logger'
-import { uploadLocalAttachment } from '../service/AttachmentService'
+import logger from '../logger'
+import { getFileSize } from '../service/FileService'
 import { shareFile } from '../service/FileSharingService'
+import { uploadLocalAttachment } from '../service/AttachmentService'
 
 export default {
 	name: 'ComposerAttachments',
@@ -65,6 +71,10 @@ export default {
 		bus: {
 			type: Object,
 			required: true,
+		},
+		uploadSizeLimit: {
+			type: Number,
+			default: 0,
 		},
 	},
 	data() {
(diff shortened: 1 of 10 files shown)


templates/index.php
```diff
@@ -29,6 +29,7 @@
 ?>
 
 <input type="hidden" id="debug-mode" value="<?php p($_['debug'] ? 'true' : 'false'); ?>">
+<input type="hidden" id="attachment-size-limit" value="<?php p($_['attachment-size-limit']); ?>">
```
ChristophWurst, line 32: we should use initial state. will send a follow-up PR

ChristophWurst:
cc @johnnyq

GretaD:
this error message popup everytime i want to attach smth, even if its 1kb.
error_attachment
also it says 0B, shouldnt it be 3(thats the limit right)?

src/components/ComposerAttachments.vue
```diff
@@ -93,22 +99,58 @@ export default {
...
 		onLocalAttachmentSelected(e) {
 			this.uploading = true
 
 			Vue.set(this, 'uploads', {})
 
+			const toUpload = Object.values(e.target.files).reduce((acc, file) => {
+				return acc + file.size
+			}, 0)
+			const newTotal = toUpload + this.totalSizeOfUpload()
+			logger.debug('checking upload size limit', {
```
ChristophWurst, line 133: @GretaD what does this line log for you?

GretaD: [DEBUG] mail: checking upload size limit 
Object { app: "mail", uid: "greta", existingUploads: 0, toUpload: 20436, limit: "0", newTotal: 20436 }

GretaD: also, sorry i forgot to include in the first comment, when i click New Message is shows this: 
```
[Vue warn]: Invalid prop: type check failed for prop "uploadSizeLimit". Expected Number with value 0, got String with value "0".

found in

---> <ComposerAttachments> at src/components/ComposerAttachments.vue
```

ChristophWurst: We need to cast. On it.

ChristophWurst: Please try again.

st3iny:
The check is not done when adding a file from the cloud as a direct attachment. I looked through the code of the filepicker and it only returns a path. Implementing this would require some form of api call to get the file size. Do we want this?

ChristophWurst:
> The check is not done when adding a file from the cloud as a direct attachment. I looked through the code of the filepicker and it only returns a path. Implementing this would require some form of api call to get the file size. Do we want this?

It's kind of inevitable, I think. Or at least I don't know an alternative off the top of my head :wink: 
If we one day have a modern file picker we can maybe return file info objects instead of just paths. That would help a lot.

GretaD:
When i upload a file that is above limit, on UI it doesnt show any warning, just the console error
```
[ERROR] mail: could not upload all attachments 
app: "mail"
error: Error: Request failed with status code 500
uid: "greta"
```

st3iny (approved):
Works for me. Cloud attachments are now checked as well.

ChristophWurst:
> error: Error: Request failed with status code 500

@GretaD do you get the same response as @st3iny in https://github.com/nextcloud/mail/issues/4217?
What does Nextcloud log say?

ChristophWurst:
Merging in the hope that the attachments bug is not related to my change :crossed_fingers: 

@GretaD pleast test again and check your logs. Then open a ticket with the details.

Issue #4000, opened 2020-11-04
Mails with same sent_at value wont display (only one of them)


stavros-k:
### Expected behavior
Display all mails

### Actual behavior
Diplays only one mail if more than 1 mail with same sent_at value exists

### Mail app

**Mail app version:** 1.6.0 (see apps admin page, e.g. 0.5.3)

**Mailserver or service:** Outlook (e.g. Outlook, Yahoo, Gmail, Exchange,...)

### Server configuration

**Operating system**: Unraid 6.8.3 (e.g. Debian 8)

**Web server:** Nginx (e.g. Apache, Nginx,...)

**Database:** PostgreSQL (e.g. MariaDB, SQLite or PostgreSQL)

**PHP version:** 7.3.24 (e.g. 7.0)

**Nextcloud Version:** 20.0.1 (see admin page, e.g. 13.0.2)

#### Client configuration
**Browser:** Chrome 86.0.4240.111(e.g. Firefox 48)

**Operating system:** Windows 10 (e.g. Arch Linux)

I have a folder with notifications from sites that items are back in stock again. Noticed some of them wasn't displaying. 
Checked db from pgAdmin and noticed all those had the same sent_at value, changed one of them and it displayed.

ChristophWurst:
Interesting observation. Where I can imagine this to be a problem is the pagination, because the `sentAt` is used as the cursor. Does the overlap appear at the last element of a page?

stavros-k:
I marked with yellow the 2 mails that appear on the webUI. 

It looks like it selects a random mail from the group of same sent_at values. Neither the 1st or the last.
If i manually change the sent_at value to a different one they appear normally.

![2](https://user-images.githubusercontent.com/47820033/98248348-7937b880-1f7d-11eb-997c-10912c9157f2.PNG)
![1](https://user-images.githubusercontent.com/47820033/98248352-7a68e580-1f7d-11eb-90ce-cdd30ad776b2.PNG)

ChristophWurst:
This is a mailbox with just two messages? Once you changed the sent_at in the DB, the UI will always show both messages, even after reload?

stavros-k:
This mailbox contains the selected rows on the db (8mails) 
Changed the sent_at on db, refreshed webui and all mails appeared, restarted docker, run "occ mail:account:sync 3 -f" and they still appear correctly. So it persists reboots.
Only way to make it get the original sent_at value is to remove mails and the mailbox from db and refetch

[...]

PR #4015, opened 2020-11-06
Fix envelope timestamp overlap

ChristophWurst:
The code assume that timestamps would be unique. Of course that is not
always the case.

Fixes https://github.com/nextcloud/mail/issues/4000

cc @stavros-k


GretaD:
while testing, this error is shown many many time, with a different number
dublicatekeys

ChristophWurst:
> while testing, this error is shown many many time, with a different number

Indeed. In the PI.

StCyr (approved):

PR #4016, opened 2020-11-06
Clean up HTML response

ChristophWurst:
Follow-up cleanup for https://github.com/nextcloud/mail/pull/3904.

~~I think Psalm won't like that we use an unknown class.~~ suppressed. There currently is no public API for the nonce.


st3iny (commented):
Code looks good (very elegant) and still works fine. But what about the static code checks failing?

ChristophWurst:
> But what about the static code checks failing?

Fixed. Had to tell Psalm that we use a class that it can't find.

st3iny (approved):

PR #4043, opened 2020-11-09
Bump @nextcloud/vue to v3

ChristophWurst:
Tested and couldn't find any breakage.


StCyr (approved):
I didn't find any issue neither

Issue #4037, opened 2020-11-07
Unified search leads to poor UX in Mail when searching for emails


severin-lemaignan:
Searching for email in nextcloud 20+ is done through the Unified search box.

However, the resulting user experience is not great: when you are in the mail app and you are search for an email, your search results are 'polluted' by results from others nextcloud apps that are irrelevant in the context or searching an email, and the email search result do not always show at the top:

![Screenshot from 2020-11-07 13-03-20](https://user-images.githubusercontent.com/385379/98442063-67c7eb00-20fa-11eb-8ec3-1dea6cb7f318.png)

Worst, in many cases, these other apps are faster to return a result, which means that the first visible results are not even related to the task ('searching for an email').

Power-users might be aware of `in:mail` in order to limit the search results to emails, but you should not need to be a power user to search in your emails.

=> It seems to me that, while nice on the paper, the 'unified search' approach has some real downsides in practice.

Possible ways forward:
- when in the mail app, the unified search bar only returns results from emails by default (with the option to extend to the other apps via eg clicking a buttom)
- have a dedicate search bar for the emails
- ...?

StCyr:
>     * when in the mail app, the unified search bar only returns results from emails by default (with the option to extend to the other apps via eg clicking a buttom)

I agree with this solution, though it could be also the other way round (have an option to limit the search to only mails)

ChristophWurst:
Found the issue.

Issue #3658, opened 2020-09-25
unified search repeats last entry in the list when loading more results


StCyr:
### Expected behavior
Unified search should show correct results

### Actual behavior
When you load more results, unified search repeats the last entry in the list:

![image](https://user-images.githubusercontent.com/8179031/94288145-6205b400-ff57-11ea-8d5a-319d3e7432eb.png)

### Mail app

**Mail app version:** master

ChristophWurst:
Oh, I think that might be because the app search logic that is also used to load the ordinary envelope list assumes that the cursor element is not visible, hence the overlap is OK there.

I will have a look.

StCyr:
I can't reproduce this issue anymore

PR #4044, opened 2020-11-09
Always show Mail search results first

ChristophWurst:
If a search request origins from a Mail page, we should show Mail
results on top.

Inspect the result of ``/ocs/v2.php/search/providers?from=%2Fapps%2Fmail%2Fbox%2F3232%2Fthread%2F285830`` and you'll see that before we returned the default order number, now it will return `-1`  when you are on a Mail page.

Fixes https://github.com/nextcloud/mail/issues/4037

### DIFF
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -71,6 +71,11 @@ class Provider implements IProvider {
 	}
 
 	public function getOrder(string $route, array $routeParameters): int {
+		if (strpos($route, Application::APP_ID . '.') === 0) {
+			// Active app, prefer Mail results
+			return -1;
+		}
+
 		return 20;
 	}
 


StCyr:
Could it be somehow linked to https://github.com/nextcloud/mail/issues/3658 ?

ChristophWurst:
> Could it be somehow linked to #3658 ?

no, that ticket is about pagination. this PR is purely about the order.

StCyr:
doesn't seem to work here.

On development instance:

1- Files contains one file named "welcome.txt"
2- Mail has knowledge of several emails containing the straing "welcome" in their subject
3- A search on "welcome" shows Files' result before Mail's results

image

ChristophWurst:
> Inspect the result of `/ocs/v2.php/search/providers`

Please post a screenshot of that.

StCyr:
> 
> 
> > Inspect the result of `/ocs/v2.php/search/providers`
> 
> Please post a screenshot of that.

image

image

Strange: It looks like it didn't do anything... Maybe I need to flush my browser's cache?...

ChristophWurst:
No, not by opening the URL with the browser. There should be a request to this URL when you reload the *mail page*. Inspect the response of that request.

StCyr:
> No, not by opening the URL with the browser. There should be a request to this URL when you reload the _mail page_. Inspect the response of that request.

I'm sorry but the result is exactly the same:

image

And, I did rebuild Mail (make dev-setup + make build-js) once again before reloading the page, just to be sure....

lib/Search/Provider.php
```diff
@@ -71,6 +71,11 @@ public function getName(): string {
 	}
 
 	public function getOrder(string $route, array $routeParameters): int {
+		if (strpos($route, Application::APP_ID . '.') === 0) {
```
ChristophWurst, line 74: @StCyr please attach a debugger here so we can get the value of ``$route`` here

StCyr: image

ChristophWurst: :open_mouth: oh

What is the full URL sent (without domain)?

StCyr: http://192.168.0.128/nextcloud/ocs/v2.php/search/providers?from=/nextcloud/apps/mail/box/12/thread/1063

ChristophWurst: It's likely an issue because your nextcloud is hosted in a subdirectory. You'll experience the same if you have Talk installed. The search won't show Talk results on top, while it does with installations that have Nextcloud running in the root directory.

GretaD:
looks like it renders first the one that matches completely the word.

search_mail_first

ChristophWurst:
> looks like it renders first the one that matches completely the word.

No, see https://github.com/nextcloud/mail/pull/4044#issuecomment-724559869. Inspect that request on your instance. Every provider has an order. It whould be -5 on a Mail page and 20 on any other page.

lib/Search/Provider.php
StCyr: yeah, apparently, this occurs here: 

	public function findMatchingRoute(string $url): array {
https://github.com/nextcloud/server/blob/769f59ec657a4dcb993550ef33341015d5ea7942/lib/private/Route/Router.php#L245

ChristophWurst: Thanks for looking into it and opening the ticket!

StCyr (approved):
Let's approve this PR. I guess it works with nextcloud instances hosted at the webroot

ChristophWurst:
> Let's approve this PR. I guess it works with nextcloud instances hosted at the webroot

@brueckner do you maybe have your dev instance hosted in the root of localhost?

st3iny (approved):
Just tested and it works (my test instance is hosted at /).

PR #4047, opened 2020-11-10
Fix the back arrow to work properly

GretaD:

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -76,7 +76,7 @@ export default {
 			error: undefined,
 			expandedThreads: [],
 			participantsToDisplay: 999,
-			resizeDebounced: debounce(500, this.updateParticipantsToDisplay)
+			resizeDebounced: debounce(500, this.updateParticipantsToDisplay),
 		}
 	},
 
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -1,5 +1,7 @@
 <template>
 	<Content v-shortkey.once="['c']" app-name="mail" @shortkey.native="onNewMessage">
+		<AppDetailsToggle v-if="isMobile && isThreadShown"
+			@close="hideMessage" />
 		<Navigation />
 		<MailboxThread v-if="activeAccount" :account="activeAccount" :mailbox="activeMailbox" />
 	</Content>
@@ -7,6 +9,7 @@
 
 <script>
 import Content from '@nextcloud/vue/dist/Components/Content'
+import AppDetailsToggle from '../components/AppDetailsToggle'
 
 import isMobile from '@nextcloud/vue/dist/Mixins/isMobile'
 import logger from '../logger'
@@ -19,6 +22,7 @@ export default {
 		Content,
 		MailboxThread,
 		Navigation,
+		AppDetailsToggle,
 	},
 	mixins: [isMobile],
 	computed: {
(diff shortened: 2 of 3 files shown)


src/components/Thread.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<AppContentDetails id="mail-message">
+	<div id="mail-message">
```
ChristophWurst, line 2: the previous component is a `div` as well. Is the `app-content-details` class a problem?

Ref https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppContentDetails/AppContentDetails.vue#L24

GretaD: the idea is to remove the AppContentDetails component, and to move all the logic at Home.vue. That was my understanding on how Contact app does. Right now the back arrow is shown currently, it just doesnt work :confounded:

src/views/Home.vue
```diff
@@ -1,5 +1,9 @@
 <template>
 	<Content v-shortkey.once="['c']" app-name="mail" @shortkey.native="onNewMessage">
+		<div v-if="isMobile"
+			class="toggle icon-confirm"
+			href="#"
+			@click="$emit('close')" />
```
juliushaertl, line 6: This doesn't work yet, because close is not catched by the parent component. You need to trigger a route navigation as it was done previously in https://github.com/nextcloud/mail/blob/3f1e1302619adee986aa6ab90b752265c696dc02/src/components/MailboxThread.vue#L160-L166

This would then be basically the same approach as in contacts: https://github.com/nextcloud/contacts/commit/c1ec4fce4182cfad4533481464ebe8e7ab766650#diff-d2191123f77aadc793b4343d3561021532924c0b1df3117a4707aecfd0bc012b

juliushaertl (changes requested):
See comment :wink:

src/views/Home.vue
```diff
@@ -1,12 +1,15 @@
 <template>
 	<Content v-shortkey.once="['c']" app-name="mail" @shortkey.native="onNewMessage">
+		<AppDetailsToggle v-if="isMobile && showEnvelope"
```
juliushaertl, line 3: I'd probably rename that to `isThreadShown` to make it a bit clearer when this is visible.

src/views/Home.vue
```diff
@@ -31,6 +35,9 @@ export default {
 		menu() {
 			return this.buildMenu()
 		},
+		showEnvelope() {
+			return (this.$route.params.threadId)
```
juliushaertl, line 39:
```suggestion
			return this.$route.params.threadId
```

src/views/Home.vue
```diff
@@ -94,8 +101,18 @@ export default {
 				},
 			})
 		},
+		hideMessage() {
+			this.$router.replace({
+				name: 'mailbox',
+				params: {
+					mailboxId: this.$route.params.mailboxId,
+					filter: this.$route.params.filter ? this.$route.params.filter : undefined,
```
juliushaertl, line 109:
```suggestion
					filter: this.$route.params?.filter,
```

src/views/Home.vue
GretaD: whaaaattt ups

ChristophWurst (approved):

Issue #3951, opened 2020-11-03
Synchronize non background synchronized mailbox when they are opened


ChristophWurst:
### Feature Request

As of https://github.com/nextcloud/mail/pull/3644 we don't sync every mailbox in background anymore. This means as you open those with the sync set to *off* you'll see an outdated representation of the mailbox. Only the background sync every 30s will eventually update the state.

In the case of those mailboxes we should, however, trigger the sync right away, so the view gets updated ASAP and the users sees the actual contents of that mailbox.

Ideally the sync state should be reflected in the UI so the user is aware what the application is doing.

#### Summary

Sync mailboxes without background sync when they are opened by the user.

PR #4048, opened 2020-11-10
Drag and Drop (revisited)

brueckner:
## Adds drag & drop support for moving emails to folders

First off, I know there has been a previous attempt (#2834) to implement this and it has been closed in favor of #3591. This PR doesn't aim to replace the newly implemented modal window to move mails, but rather to complement it. It in fact uses methods introduced in #3591 to actually move the mails.

This approach also makes use of the multi-selection feature, to enable dragging of single and multiple emails.

I have recorded a quick **video** to showcase the feature (too bad I cant attach it right here, Nextcloud to the rescue!): 
https://cloud.dotplex.com/s/RHNGPL8xCzspCkr

**Sneak peek:**

Screenshot_10_11_20__12_18

### Features
- dragging and dropping of single or multiple mails
- customized draggable-ghost to match the general style
  - with count of mails being dragged, their subject and sender
- visual feedback for invalid drop-targets (faded out)
- collapsed folders of the same account expand automatically when starting to drag
- support for subfolders
- dragging to trash/junk is possible
- source and target folders are synced right away once the mails have been dropped

### Code
This is dependency-free. I am **not** using any external libraries for this. I dug into the [HTML5 drag and drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API). I've studied #2834 and the [library that was used there](https://github.com/Vivify-Ideas/vue-draggable) and figured out that in order for this to work, event listeners would have to be attached in a more dynamic manner, since mails are loaded asynchronously. I was trying to make it work with vue-draggable, as well as some other libraries, but each of them would spawn a different set of challanges. So in the end I decided to do it "by hand", which has the huge benefit of being able to tailor the feature exactly to our needs.

### It's a draft!
In order to get this to work, I had to do some low-level attaching of eventListeners and dom manipulations (to create the draggable-ghost). It's basically exactly what vue-draggable does, just with less overhead. I would love to wrap that up in two [custom directives](https://vuejs.org/v2/guide/custom-directive.html) (one for the draggables, one for the drop-targets), but I've never done that before and ran into some issues. So, before pouring more time into this, I wanted to make sure that this feature would be accepted in the first place. 

### How does this work?
There are `dragstart` and `dragend` eventListeners attached to an envelope in `Envelope.vue`. On the other hand, there are `dragenter`, `dragover`, `dragleave` and `drop` eventListeners on each folder/mailbox in `NavigationMailbox.vue` that matches the criteria for a valid drop-target.

On `dragstart`, each envelope being dragged is added to the global $store and a drag-ghost is created. The HTML5 API to customize that is kinda wierd... you have to create an element to your liking and attach it to the DOM. The browser will then take a "screenshot" of that element, at which point you can remove the element from the DOM again. So in practice: you create the element, attach it to the dom, set it as a [DragImage](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setDragImage) and remove it right away. 🤷‍♂️

Once you drop the mail(s) on a valid drop target (mailbox), the mails to be moved are pulled from the $store and processed. It would be possible to make them disappear right away, but I went with a slight fade-out until they are actually moved. In theory, it would be possible to use the [HTML5 DataTransfer object](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer) instead of the $store, but that API is pretty wierd as well. Good luck passing objects... :/

### Improvements
- Code isolation – as stated before. Maybe using custom directives? Glad for any suggestion.
- Styles: I blatently stuck the styling for the draggable-ghost into `App.vue`, as it **needs to be** global. Of course, that's not the right place. How/where is [`mail.scss`](https://github.com/nextcloud/mail/blob/master/css/mail.scss) compiled? I'd like to add it there.

### Trojan Horse 🐴
[This commit](https://github.com/nextcloud/mail/commit/4f5e1d89664477214ac440db33f6ac48dc23af96) has nothing to do with the feature, but it neccessary for it to work. I need to pass an event from an envelope to a mailbox/folder. Apparently there is an event bus, but it is created in [`MailboxThread.vue`](https://github.com/nextcloud/mail/commit/4f5e1d89664477214ac440db33f6ac48dc23af96#diff-2ef5622c7090ce9f7e480022e8f1217a0c811b608ecf6335e71758ec3c43768eL115) and only passed down as a prop to its' subcomponents. Instead, I am proposing [this pattern](https://medium.com/@andrejsabrickis/https-medium-com-andrejsabrickis-create-simple-eventbus-to-communicate-between-vue-js-components-cdc11cd59860) to make it available whereever neccessary. I'd submit this as a separate PR once out of draft-phase.

## Conclusion
I strongly believe that this is a **must have** for an email client. Every user will try to drag an email. We are just used to it from any other client out there. However, I am not married to the specifics of this implementation and am glad for any help with this!

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -1,5 +1,16 @@
 <template>
-	<router-link class="app-content-list-item" :class="{seen: data.flags.seen, draft, selected: selected}" :to="link">
+	<router-link
+		v-draggable-envelope="{
+			accountId: data.accountId ? data.accountId : mailbox.accountId,
+			mailboxId: data.mailboxId,
+			envelopeId: data.databaseId,
+			draggableLabel: `${data.subject} (${data.from[0].label})`,
+			selectedEnvelopes,
+		}"
+		class="app-content-list-item"
+		:class="{seen: data.flags.seen, draft, selected: selected}"
+		:to="link"
+		:data-envelope-id="data.databaseId">
 		<div
 			v-if="mailbox.isUnified"
 			class="mail-message-account-color"
@@ -133,6 +144,7 @@ import { matchError } from '../errors/match'
 import NoTrashMailboxConfiguredError
 	from '../errors/NoTrashMailboxConfiguredError'
 import logger from '../logger'
+import { DraggableEnvelopeDirective } from '../directives/drag-and-drop/draggable-envelope'
 
 export default {
 	name: 'Envelope',
@@ -143,6 +155,9 @@ export default {
 		Avatar,
 		Moment,
 		MoveModal,
+	},
+	directives: {
+		draggableEnvelope: DraggableEnvelopeDirective,
 	},
 	props: {
 		data: {
@@ -160,6 +175,11 @@ export default {
 		selected: {
 			type: Boolean,
 			default: false,
+		},
+		selectedEnvelopes: {
+			type: Array,
+			required: false,
+			default: () => [],
 		},
 	},
 	data() {
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -98,6 +98,7 @@
 				:mailbox="mailbox"
 				:selected="selection.includes(env.databaseId)"
 				:select-mode="selectMode"
+				:selected-envelopes="selectedEnvelopes"
 				@delete="$emit('delete', env.databaseId)"
 				@update:selected="onEnvelopeSelectToggle(env, index, ...$event)"
 				@select-multiple="onEnvelopeSelectMultiple(env, index)" />
@@ -125,6 +126,7 @@ import { matchError } from '../errors/match'
 import NoTrashMailboxConfiguredError
 	from '../errors/NoTrashMailboxConfiguredError'
 import { differenceWith } from 'ramda'
+import dragEventBus from '../directives/drag-and-drop/util/dragEventBus'
 
 export default {
 	name: 'EnvelopeList',
@@ -202,7 +204,20 @@ export default {
 				})
 		},
 	},
+	mounted() {
+		dragEventBus.$on('envelopesDropped', this.unselectAll)
+	},
+	beforeDestroy() {
+		dragEventBus.$off('envelopesDropped', this.unselectAll)
+	},
 	methods: {
+		isEnvelopeSelected(idx) {
+			if (this.selection.length === 0) {
+				return false
+			}
+
+			return this.selection.includes(idx)
+		},
 		markSelectedSeenOrUnseen() {
 			const seen = !this.areAllSelectedRead
 			this.selectedEnvelopes.forEach((envelope) => {
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -24,6 +24,11 @@
 		v-if="visible"
 		:id="genId(mailbox)"
 		:key="genId(mailbox)"
+		v-droppable-mailbox="{
+			mailboxId: mailbox.databaseId,
+			accountId: mailbox.accountId,
+			isValidDropTarget,
+		}"
 		:allow-collapse="true"
 		:menu-open.sync="menuOpen"
 		:force-menu="true"
@@ -142,6 +147,8 @@ import logger from '../logger'
 import { translatePlural as n } from '@nextcloud/l10n'
 import { translate as translateMailboxName } from '../i18n/MailboxTranslator'
 import { showInfo } from '@nextcloud/dialogs'
+import { DroppableMailboxDirective as droppableMailbox } from '../directives/drag-and-drop/droppable-mailbox'
+import dragEventBus from '../directives/drag-and-drop/util/dragEventBus'
 
 export default {
 	name: 'NavigationMailbox',
@@ -153,6 +160,9 @@ export default {
 		ActionCheckbox,
 		ActionInput,
 		MoveMailboxModal,
+	},
+	directives: {
+		droppableMailbox,
 	},
 	props: {
 		account: {
@@ -189,7 +199,6 @@ export default {
 			renameInput: false,
 			mailboxName: this.mailbox.displayName,
 			showMoveModal: false,
-
 		}
 	},
 	computed: {
@@ -262,6 +271,34 @@ export default {
 		isSubscribed() {
 			return this.mailbox.attributes && this.mailbox.attributes.includes('\\subscribed')
 		},
+		isDroppableSpecialMailbox() {
+			if (this.filter === 'starred') {
+				return false
+			}
+			return ![
+				this.account.draftsMailboxId,
+				this.account.sentMailboxId,
+			].includes(this.mailbox.databaseId)
+		},
+		isActive() {
+			return this.$route.params.mailboxId === this.mailbox.databaseId
+		},
+		isValidDropTarget() {
+			if (this.isActive) {
+				return false
+			}
+			return this.isDroppableSpecialMailbox || (!this.mailbox.specialRole && !this.account.isUnified)
+		},
+	},
+	mounted() {
+		dragEventBus.$on('dragStart', this.onDragStart)
+		dragEventBus.$on('dragEnd', this.onDragEnd)
+		dragEventBus.$on('envelopesMoved', this.onEnvelopesMoved)
+	},
+	beforeDestroy() {
+		dragEventBus.$off('dragStart', this.onDragStart)
+		dragEventBus.$off('dragEnd', this.onDragEnd)
+		dragEventBus.$off('envelopesMoved', this.onEnvelopesMoved)
 	},
 	methods: {
 		/**
@@ -443,6 +480,38 @@ export default {
 		},
 		onCloseMoveModal() {
 			this.showMoveModal = false
+		},
+		onDragStart({ accountId }) {
+			if (accountId !== this.mailbox.accountId) {
+				return
+			}
+			this.$store.commit('expandAccount', accountId)
+			this.showSubMailboxes = true
+		},
+		onDragEnd({ accountId }) {
+			if (accountId !== this.mailbox.accountId) {
+				return
+			}
+			this.showSubMailboxes = false
+		},
+		onEnvelopesMoved({ mailboxId, movedEnvelopes }) {
+			if (this.mailbox.databaseId !== mailboxId) {
+				return
+			}
+			const openedMessageHasBeenMoved = movedEnvelopes.find((movedEnvelope) => {
+				return movedEnvelope.envelopeId === this.$route.params.threadId
+			})
+			// navigate to the mailbox root
+			// if the currently displayed message has been moved
+			if (this.$route.name === 'message' && openedMessageHasBeenMoved) {
+				this.$router.push({
+					name: 'mailbox',
+					params: {
+						mailboxId: this.$route.params.mailboxId,
+						filter: this.$route.params?.filter,
+					},
+				})
+			}
 		},
 	},
 }
--- a/src/directives/drag-and-drop/draggable-envelope/draggable-envelope.js
+++ b/src/directives/drag-and-drop/draggable-envelope/draggable-envelope.js
@@ -1,0 +1,104 @@
+import dragEventBus from '../util/dragEventBus'
+import defer from 'lodash/defer'
+
+export class DraggableEnvelope {
+
+	constructor(el, componentInstance, options) {
+		this.el = el
+		this.options = options
+		this.registerListeners.bind(this)(el)
+		this.setInitialAttributes()
+	}
+
+	setInitialAttributes() {
+		this.el.setAttribute('draggable', 'true')
+		this.el.classList.add('draggable-envelope')
+	}
+
+	update(el, instance) {
+		this.options = instance.options
+	}
+
+	registerListeners(el) {
+		el.addEventListener('dragstart', this.onDragStart.bind(this))
+		el.addEventListener('dragend', this.onDragEnd.bind(this))
+	}
+
+	removeListeners(el) {
+		el.removeEventListener('dragstart', this.onDragStart)
+		el.removeEventListener('dragend', this.onDragEnd)
+	}
+
+	onDragStart(event) {
+		const { accountId, mailboxId, selectedEnvelopes } = this.options
+
+		event.dataTransfer.clearData()
+		event.dataTransfer.effectAllowed = 'move'
+
+		const envelopes = []
+		if (selectedEnvelopes.length) {
+			// handle dragged selection mode items
+			selectedEnvelopes.forEach((envelope, index) => {
+				envelopes.push({
+					accountId,
+					mailboxId,
+					envelopeId: envelope.databaseId,
+					draggableLabel: `${envelope.subject} (${envelope.from[0].label})`,
+				})
+			})
+		} else {
+			// handle single dragged item
+			const { envelopeId, draggableLabel } = this.options
+			envelopes.push({ accountId, mailboxId, envelopeId, draggableLabel })
+		}
+
+		event.dataTransfer.setData('text/plain', JSON.stringify(envelopes))
+		this.attachGhost({ event, envelopes })
+
+		dragEventBus.$emit('dragStart', {
+			accountId,
+			mailboxId,
+			itemCount: envelopes.length,
+		})
+	}
+
+	onDragEnd(event) {
+		dragEventBus.$emit('dragEnd', { accountId: this.options.accountId })
+	}
+
+	attachGhost({ event, envelopes }) {
+		const baseClass = 'draggable-envelope-ghost'
+		const dragNode = document.createElement('div')
+		dragNode.classList.add(baseClass)
+
+		const counterNode = document.createElement('span')
+		counterNode.classList.add(`${baseClass}--counter`)
+		const textCountNode = document.createTextNode(envelopes.length)
+		counterNode.appendChild(textCountNode)
+		dragNode.appendChild(counterNode)
+
+		const labelWrapperNode = document.createElement('div')
+		labelWrapperNode.classList.add(`${baseClass}--label-wrapper`)
+
+		envelopes.forEach(envelope => {
+			const labelNode = document.createElement('div')
+			labelNode.classList.add(`${baseClass}--label-wrapper--label`)
+			const textLabelNode = document.createTextNode(envelope.draggableLabel)
+			labelNode.appendChild(textLabelNode)
+			labelWrapperNode.appendChild(labelNode)
+		})
+
+		dragNode.appendChild(labelWrapperNode)
+		document.body.appendChild(dragNode)
+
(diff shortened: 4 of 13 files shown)


st3iny:
Awesome! In my opinion this is a huge usability improvement. I will look at the code and test as soon as I can :)

ChristophWurst:
:exploding_head:

ChristophWurst:
@jancborchardt please have a look at the design :)

brueckner:
So, does anyone happen to have an opinion on this? 
> ### It's a draft!
> In order to get this to work, I had to do some low-level attaching of eventListeners and dom manipulations (to create the draggable-ghost). It's basically exactly what vue-draggable does, just with less overhead. I would love to wrap that up in two [custom directives](https://vuejs.org/v2/guide/custom-directive.html) (one for the draggables, one for the drop-targets), but I've never done that before and ran into some issues. So, before pouring more time into this, I wanted to make sure that this feature would be accepted in the first place. 

I'm eager to put time into optimizations, but would like to put that time into the "right" thing.

st3iny (commented):
First of all, I really like your changes. Dragging works very well, feels nice and fits the design of nextcloud. The code looks clean overall but I commented on some minor things here and there. The new global event bus is a great addition but might be out of scope of this PR. If this is blocking the review/merge we could move this to a follow up PR (but keep the usages in the dragging related code here).

When moving an envelope the thread is not updated. E.g. moving a collapsed envelope to another mailbox and then expand this envelope in the thread will cause it to not load because the attached message/envelope does not exist anymore. I had a similiar issue in my move modal code. You could use the new global event bus and update the thread after moving via `this.fetchThread()` or navigate away when the thread root was moved. Take a look at my solution https://github.com/nextcloud/mail/pull/3915/files#diff-ee27750a392035af626e82a2c822204689adf614396d005d18281faf701da601 (especially at the `onMove()` method)

The draggable ghost does not have a transparent background but this seems to be an issue with the proprietary nvidia linux drivers and compositing and is not related to your code.

src/App.vue
```diff
@@ -59,3 +59,56 @@ export default {
 	},
 }
 </script>
+
+<style lang="scss">
+// TODO: move to mail.scss
+// The drag-ghost is appended to the body
+// and cannot be styled from within a
+// component with scoped CSS
```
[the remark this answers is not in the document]
brueckner: 🤔  Where/How is mail.scss compiled though? It's not picked up by the default watcher process.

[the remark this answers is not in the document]
ChristophWurst: By Nextcloud server in the background. But we're trying to move away from this. Please try to put everything into a component scss so it's bundled with the app :v:

ChristophWurst:
> In theory, it would be possible to use the [HTML5 DataTransfer object](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer) instead of the $store, but that API is pretty wierd as well. Good luck passing objects... :/

I remember facing the same problem when I once tried to revive drag and drop in an early version of Mail with Vue. Now that we streamlined our data structures a bit we don't have to pass around objects anymore because messages can be identified by unique integer IDs (was a combination of account id, mailbox name and message UID before). So, can we pass the array of IDs? Is that easier than an *object*? If a string works we can also serialize the ID array to a json string. :thinking:

brueckner:
It would be charming indeed to detach this thing from the store. My first approach was to attach a `DataTransferItem` for each Envelope in a [`DataTransferItemList`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItemList). I remember it feeling a little hacky and it felt much cleaner using the store. However, passing a serialized array could simplify this a lot.  I'm gonna look into it again and will report back.

brueckner:
Okay, so this API is melting my brain. Apparently, you shouldn't be able to access the data attached to the dataTransfer object on `dragover`, but only in the `drop` event handler (for security reasons). Oddly enough, it DOES work in Firefox, but not in Webkit. Here's a StackOverflow reply that describes the issue: https://stackoverflow.com/a/31922258

That's a bummer, because I am trying to prevent the user from dropping the mail into the same folder it came from. People are recommending global variables in that case (so... the store in our case). 🤷‍♂️ 

FWIW – attaching a stringified JSON and parsing it in `onDrop` works like a charm!

The other reason the current implementation is relying on the global store is for each Mailbox in the sidebar to detect wether it is a valid drop target or not (and to disable/fade it out in that case). Without using the store, I'd have to emit an event to inform the mailboxes about the origin of envelopes being dragged instead. Which is... I don't know. It's just shifting the load from the store over to events. I prefer the data-driven approach using the store tbh.

What do you think?

ChristophWurst:
> Okay, so this API is melting my brain. Apparently, you shouldn't be able to access the data attached to the dataTransfer object on `dragover`, but only in the `drop` event handler (for security reasons). Oddly enough, it DOES work in Firefox, but not in Webkit. Here's a StackOverflow reply that describes the issue: https://stackoverflow.com/a/31922258
> 
> That's a bummer, because I am trying to prevent the user from dropping the mail into the same folder it came from. People are recommending global variables in that case (so... the store in our case). man_shrugging

Yeah guess we have to use the store then. Or we just don't do the move (and show a hint to the user via a Toast) if source=destination.



> The other reason the current implementation is relying on the global store is for each Mailbox in the sidebar to detect wether it is a valid drop target or not (and to disable/fade it out in that case). Without using the store, I'd have to emit an event to inform the mailboxes about the origin of envelopes being dragged instead. Which is... I don't know. It's just shifting the load from the store over to events. I prefer the data-driven approach using the store tbh.

The dynamic drop target is required for the move from the unified/priority inbox to their real mailboxes, right? Because otherwise the current mailbox that is open should never be a valid drop target.

brueckner:
Alright, I refactored it into **two** directives `v-draggable-envelope` and `v-droppable-mailbox` and moved most of the code into these, along with the new `DragEventBus`. All of these live in this folder structure:

```
- directives/
- - drag-and-drop/
- - - draggable-envelope/  <--- v-draggable-envelope
- - - droppable-mailbox/   <--- v-droppable-mailbox
- - - util/                <--- DragEventBus
```

I still need to move the styles, but the most important change is the extraction into directives.

**Also:** I am not using the store to temporarily persist the envelopes anymore. I am using local state within the directives in combination with the DataTransferObject: `event.dataTransfer.setData('text/plain', JSON.stringify(envelopes))`.

It works faily well like that!

I also still need to look into @st3iny 's comment:
> When moving an envelope the thread is not updated. E.g. moving a collapsed envelope to another mailbox and then expanding this envelope in the thread will cause it to not load because the attached message/envelope does not exist anymore. I had a similar issue in my move modal code. You could use the new global event bus and update the thread after moving via this.fetchThread() or navigate away when the thread root was moved. Take a look at my solution https://github.com/nextcloud/mail/pull/3915/files#diff-ee27750a392035af626e82a2c822204689adf614396d005d18281faf701da601 (especially at the onMove() method)

Oh, there is one addition in terms of usability: when starting the drag, not only is the list of folders expanded, I am now also expanding all folders and their subfolders, which helps alot I think. When dragging is done, the subfolders are collapsed again.

I'm sure I forgot to mention something, but I am stoked about the fact that I managed to separate most of the functionality into these two directives. 🥳  Could you please have a look at that?

brueckner:
Okay, I'm thrilled to announce that I think I'm done. 🔥

---

The issue @st3iny mentioned (thanks for that by the way, I actually didn't notice that myself!) is resolved here: https://github.com/nextcloud/mail/pull/4048/commits/173201db194f465a6e346333f6b03ddc92bf311c. I decided to check if the user is currently viewing a message and if that `threadId` is in the list of moved messages. If it is, I'm navigating to the mailbox root without selecting a message. I think that's better than opening the next message automatically, because it doesn't trigger a load that would potentially delay the next move-operation.

--- 

As for the styles, I moved **all** related styling to separate scss files within the directive's folder. To avoid potential styling collisions (since they're not scoped anymore), the classes and stylable attributes are now fairly specific.

I also refactored and moved around some code. The `v-droppable-mailbox` now emits an `envelopesMoved` event after the messages have been moved and puts the responsibility of what's supposed to happen afterwards back into the "hands" of the [`NavigationMailbox`](https://github.com/nextcloud/mail/pull/4048/files#diff-039d56a63d245f0bb54616f6098247d1b2ca095374b671f8c426f9c02e4e9bbcR486).

Oh, I also reverted my global EventBus changes, which were still lingering in there... ;-)

---

It would be great if you took a final look at the whole thing one more time, @st3iny and @ChristophWurst, before I bring it up to speed with master and squash everything.

jancborchardt (approved):
Really great design-wise @brueckner, amazing work! :)

src/components/Envelope.vue
```diff
@@ -133,6 +144,7 @@ import { matchError } from '../errors/match'
 import NoTrashMailboxConfiguredError
 	from '../errors/NoTrashMailboxConfiguredError'
 import logger from '../logger'
+import { DraggableEnvelopeDirective } from '../directives/drag-and-drop/draggable-envelope'
```
ChristophWurst, line 147:
```suggestion
import { DraggableEnvelopeDirective as draggableEnvelope } from '../directives/drag-and-drop/draggable-envelope'
```

and …

src/components/Envelope.vue
```diff
@@ -144,6 +156,9 @@ export default {
 		Moment,
 		MoveModal,
 	},
+	directives: {
+		draggableEnvelope: DraggableEnvelopeDirective,
```
ChristophWurst, line 160: …

```suggestion
		draggableEnvelope,
```

for concise code :)

src/components/EnvelopeList.vue
```diff
@@ -124,6 +125,7 @@ import MoveModal from './MoveModal'
 import { matchError } from '../errors/match'
 import NoTrashMailboxConfiguredError
 	from '../errors/NoTrashMailboxConfiguredError'
+import DragEventBus from '../directives/drag-and-drop/util/DragEventBus'
```
ChristophWurst, line 128:
```suggestion
import dragEventBus from '../directives/drag-and-drop/util/DragEventBus'
```

Because it's an instance/object, not a type

src/components/EnvelopeList.vue
```diff
@@ -190,6 +192,9 @@ export default {
 			return this.selection.map((idx) => this.envelopes[idx])
 		},
 	},
+	mounted() {
+		DragEventBus.$on('envelopesDropped', this.unselectAll)
+	},
```
ChristophWurst, line 197: please also add an unmount and `$off` for extra clean code :sparkles:

src/components/NavigationMailbox.vue
```diff
@@ -142,6 +147,8 @@ import logger from '../logger'
 import { translatePlural as n } from '@nextcloud/l10n'
 import { translate as translateMailboxName } from '../i18n/MailboxTranslator'
 import { showInfo } from '@nextcloud/dialogs'
+import { DroppableMailboxDirective } from '../directives/drag-and-drop/droppable-mailbox'
```
ChristophWurst, line 150:
```suggestion
import { DroppableMailboxDirective as droppableMailbox } from '../directives/drag-and-drop/droppable-mailbox'
```

like above :)

src/components/NavigationMailbox.vue
```diff
@@ -262,6 +271,23 @@ export default {
 		isSubscribed() {
 			return this.mailbox.attributes && this.mailbox.attributes.includes('\\subscribed')
 		},
+		isDroppableSpecialMailbox() {
+			return ['inbox', 'junk', 'trash'].includes(this.mailbox.specialRole)
```
ChristophWurst, line 275: the flags aren't always accurate. we should use the account preferences from https://github.com/nextcloud/mail/issues/2035, like `account.sentMailboxId`

src/components/NavigationMailbox.vue
```diff
@@ -262,6 +271,23 @@ export default {
...
+				return false
+			}
+			return this.isDroppableSpecialMailbox || (!this.mailbox.specialRole && !this.account.isUnified)
+		},
+	},
+	mounted() {
+		DragEventBus.$on('dragStart', this.onDragStart)
+		DragEventBus.$on('dragEnd', this.onDragEnd)
+		DragEventBus.$on('envelopesMoved', this.onEnvelopesMoved)
 	},
```
ChristophWurst, line 291: `$off` counterpart missing

src/directives/drag-and-drop/draggable-envelope/draggable-envelope.js
```diff
@@ -0,0 +1,103 @@
...
+
+		dragNode.appendChild(labelWrapperNode)
+		document.body.appendChild(dragNode)
+
+		event.dataTransfer.setDragImage(dragNode, 0, 15)
+
+		// the item can be removed immediately, because the
+		// browser will take a "screenshot" of the dragImage
+		// upon initialization to be used while dragging
+		setTimeout(() => {
```
ChristophWurst, line 98: You could also import https://lodash.com/docs/4.17.15#defer but it's essentially the same

ChristophWurst (commented):
Rest of the code looks good. Still have to wrap my head around those directives and the other logic but it looks solid :)

src/components/NavigationMailbox.vue
brueckner: I turned it around into a blacklist, using `sentMailboxId` and `draftsMailboxId`. Additionally, I am checking if `this.filter === 'starred'` to exclude the "Favorites"  mailbox as well (the move doesn't list the favorites either).

brueckner:
Thanks for the review @ChristophWurst, I've implemented all of the remarks. I'll hit the "Ready for review" button to move it out of drafting phase. 🚀

ChristophWurst:
Please rebase and squash this awesomeness into a single commit, then we can merge :rocket:

brueckner:
Ready to roll! 🚀  🥳

src/components/NavigationMailbox.vue
```diff
@@ -444,6 +481,40 @@ export default {
...
+			if (accountId !== this.mailbox.accountId) {
+				return
+			}
+			this.showSubMailboxes = false
+		},
+		onEnvelopesMoved({ mailboxId, movedEnvelopes }) {
+			if (this.mailbox.databaseId !== mailboxId) {
+				return
+			}
+			this.$store.dispatch('syncEnvelopes', { mailboxId })
```
ChristophWurst, line 501: I think we don't need/want that, do we? The destination mailbox is always a different one than the current mailbox. And we don't care about keeping another mailbox updated (on the front-end).

@GretaD found out that this action can run into https://github.com/nextcloud/mail/blob/f37124c3def79072bbcd2a32d71eded92533117a/lib/Controller/MailboxesController.php#L171. That is, a sync request is sent for a mailbox that had never been sync'ed before. In this case the js code is actually supposed to do a full initialization sync request. So, that is of course out of scope for the drag and drop feature, but this dispatched action could potentially carry out a very expensive operation because the initial sync isn't happening fast for larger mailboxes.

Hence I'd say we remove the explicit sync of the destination mailbox for now. When the user navigates there, the background sync will update the mailbox anyway. https://github.com/nextcloud/mail/issues/3951 is a bit of a missing piece here, but again, off topic :wink: 

@brueckner does that make sense or am I missing something?

GretaD (approved):
It works :+1:  see [comment](https://github.com/nextcloud/mail/pull/4048/files#r530947133)

src/components/NavigationMailbox.vue
brueckner: Hmmm, I see the point. However, before I implemented triggering the sync it was pretty irritating to me to move a mail to its new destination and not seeing it. Sometimes it would take quite some time until the background sync hit. Until then – from the perspective of an unaware user – the mail is just... gone. I don't know about you, but I would be pretty angry if a mail app (seemingly) made my mails disappear 👀 . Also: wouldn't the background-sync essentially do the same expensive operation as well, just... delayed? ;)

_(Disclaimer: I didn't look at the background sync yet)_ Couldn't we reset the timer of a background sync based on a move-operation's triggered sync or something along those lines? 

OR... yes! Hear me out... What if we dropped the sync-trigger and _marked_ every mailbox that was the target of a recent move-operation instead (something like `outdated`). If a user clicks on a mailbox that is marked `outdated`, as sync is triggered immediately. That way, the user will be greeted by a loading spinner and glad to find that all their mails have been moved successfully. On the other hand, all `outdated` mailboxes the user does **not** navigate to will not trigger a sync (until the regular background sync hits). 

That way I can move a thousand mails to 20 different mailboxes while only ever triggering one sync once I navigate to one of the 20 `outdated` mailboxes. In addition to that, the background sync will remove the outdated flag, thus preventing an unnecessary sync.

👯

ChristophWurst: Are you describing https://github.com/nextcloud/mail/issues/3951? :thinking:

brueckner: Ha! Hell yea! :)

So in terms of what's described in #3951 we could just disable the sync on any move-target-mailbox in the store without persisting that setting?

brueckner: Well, wait. It would just have to be disabled for like 30 seconds, right? If the user has bg sync enabled, moves a mail into the folder and **doesn't** open the mailbox right away afterwards, the background sync should still to its' thing on regular schedule.

ChristophWurst: While #3951 is very specific about non-background sync'ed mailboxes we can discuss always firing off a sync based on a *last sync* attribute on the mailbox, as you suggested.

> If the user has bg sync enabled, moves a mail into the folder and **doesn't** open the mailbox right away afterwards, the background sync should still to its' thing on regular schedule.

I think that depends on how much we want to keep mailboxes updated in the front-end code if they aren't open right now. I'd say if the front-end never fetched a list of messages, there is no point in spending energy on updating this mailbox. The user might have 30 mailboxes subscribed, this could be a lot of work for the server.

What matters most is that the currently open mailbox is relatively up to date. Same applies to when you switch to another mailbox and the messages are loaded from the database cache: there should be a sync to ensure we always show the latest data.

ChristophWurst: Btw we have the same problems/questions for drafts (they should show up if you switch mailbox) and also sent messages.

I'm all in for improving that. But can we keep those two things separate?

brueckner: Yea sure – so should I just remove the sync-trigger? We'd just have to make sure to have a solution for it when this gets released, right?

ChristophWurst: > We'd just have to make sure to have a solution for it when this gets released, right?

Absolutely. We can discuss this the next few days (maybe in a call) and then have a solid fix for the next release :)

brueckner: Sound good to me! I'll remove the trigger :)

brueckner: Removed, rebased on master and squashed. ✅

ChristophWurst (approved):
Works really well :rocket: :pray: :+1:

PR #4050, opened 2020-11-10
Fix auto manual account styling

GretaD:
beforeautomanual
after
automanual

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -350,17 +350,17 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.tabs-component-tabs {
+::v-deep .tabs-component-tabs {
 	display: flex;
 }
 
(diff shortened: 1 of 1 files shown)


src/components/AccountForm.vue
```diff
@@ -350,17 +350,17 @@ export default {
 </script>
 
 <style lang="scss" scoped>
```
ChristophWurst, line 352: is it because of scoped? https://github.com/nextcloud/mail/commit/5bed88353a5dda88d2ed16541a97b4cb04f68cb2#diff-df99c6b542ce6db7885b569ee77d0a922c4622893b5fe34a4790e888005753ceR352

GretaD: Yes i think so, because if you remove the _scoped_ the styling goes back to normal.

ChristophWurst (approved):

PR #4058, opened 2020-11-11
Fix icon styling for reply button

### LINKED ISSUE #4009: Action primary button loses the styling in small screen
### Expected behavior
Button is shown currently 

### Actual behavior
button_small screen

### Mail app

**Mail app version:** 1.6

GretaD:
The reply button icon doesnt show properly on small screen, because the `class=action-label` was missing after we moved to thread.
fixes #4009 

before
button_small screen

after
after_button_smallscreen

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -62,7 +62,7 @@
 						primary: expanded,
 					}"
 					class="button">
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

ChristophWurst:
do we need a backport?

PR #4059, opened 2020-11-11
Fix app install on CI

ChristophWurst:


GretaD:
how to test this?

ChristophWurst:
This is a test for server. Right now CI fails. Once the migration issues are fixed everything should be green here again

PR #4065, opened 2020-11-13
Fix/3829 Saving Signatures

### LINKED ISSUE #3829: Signature in Mail not saved - no errors/warnings
### Expected behavior
After entering a signature and pressing "Save Signature" it should be stored and used for newly composed e-mails.

### Actual behavior
pressing "Save signature" does not result in storage of the signature in the database:
| signature | last_mailbox_sync | editor_mode
| | 1602850502 | plaintext |

"Save signature" button remains "pressed" until any further user action is carried out
Bildschirmfoto zu 2020-10-16 14-12-46

no error or warnings are given
XHRPUThttps://XXXXX/nextcloud/index.php/apps/mail/api/accounts/2/signature
[HTTP/2 200 OK 94ms]
XHRPOSThttps://XXXXX/nextcloud/index.php/apps/mail/api/mailboxes/13/sync
[HTTP/2 200 OK 665ms]

### Mail app

1.5.0

mail.sprit.org

### Server configuration

Linux debian 4.19.0-11-amd64 #1 SMP Debian 4.19.146-1 (2020-09-17) x86_64 GNU/Linux

Apache/2.4.38 (Debian)

10.3.23-MariaDB-0+deb10u1 Debian 10

**PHP version:** (e.g. 7.0)
7.3.19-1~deb10u1

20.0.0

#### Client configuration
FireFox 81 (64bit)

Ubuntu Linux odroid 4.9.236-97 #1 SMP PREEMPT Tue Sep 15 15:01:17 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux

brueckner:
Fixes #3829 

### Problem
Signature changes are not stored in the database

### Cause
The text editor's model binding to `this.signature`...
```javascript
<TextEditor v-model="signature" ...
```
... collides with using a computed property:
```javascript
signature() {
    return this.account.signature ? toHtml(detect(this.account.signature)).value : ''
},
```

### Solution
Initialize `this.signature` with and empty string in `data()` and assign the account's stored signature in `created()`. This will properly set the initial value and allows the model binding on `<TextEditor />` to update the value.

### DIFF
--- a/src/components/SignatureSettings.vue
+++ b/src/components/SignatureSettings.vue
@@ -63,12 +63,11 @@ export default {
 		return {
 			loading: false,
 			bus: new Vue(),
+			signature: '',
 		}
 	},
-	computed: {
-		signature() {
-			return this.account.signature ? toHtml(detect(this.account.signature)).value : ''
-		},
+	created() {
+		this.signature = this.account.signature ? toHtml(detect(this.account.signature)).value : ''
 	},
 	methods: {
 		deleteSignature() {


ChristophWurst:
IIRC it was a computed prop to catch updates of the account prop. https://github.com/nextcloud/mail/pull/3986 should fix this. But maybe we should have a watcher just in case? :thinking:

brueckner:
Do you mean a watcher on the account prop which overrides `this.signature` with `newAccount.signature` on change? I can't quite figure out a scenario for this. I think this would even cause potential problems If someone were to change their signature, not save the change and then make a change to some other preference on the same page. That would trigger the watcher and override the (changed but unsaved) signature value with the previous (saved) value.

ChristophWurst:
Yeah, I think you're right. Both scenarios are edge cases. The current solution will work just fine :)

ChristophWurst (approved):

st3iny (approved):
Saving works again :tada:

Issue #3784, opened 2020-10-09
Can not activate Mail app after migration 19 to 20


daraelmin:
### Expected behavior
I was hoping to find my apps after the migration.
When activating Mail it should activate the app
When clicking "download and activate" it should activate the app

### Actual behavior
Mail was disabled during migration (after causing an error during migration)
Upon reactivation, a prompt appears saying `An error occurred during the request. Could not process the request.`
When I deleted and reinstalled the app, the same warning appeared after downloading the app.

### Mail app

**Mail app version:** (see apps admin page, e.g. 0.5.3)

**Mailserver or service:**Infomaniak 

### Server configuration

**Operating system**: ?

**Web server:** Apache 2.4.34

**Database:** MariaDB

**PHP version:** 7.4

**Nextcloud Version:** 20.0.0

#### Client configuration
**Browser:** all

**Operating system:** all

ChristophWurst:
Check data/nextcloud.log

daraelmin:
[...]

PR #4086, opened 2020-11-16
Require vendor autoloader in class definition

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3784


StCyr (approved):
seems to work here (no issue running make dev-setup)

PR #4087, opened 2020-11-16
Throw exception if repair step is run without app registration

ChristophWurst:
To debug https://github.com/nextcloud/mail/issues/3784


jancborchardt:
@ChristophWurst this was only for debugging, so we can close this, right?

ChristophWurst:
absolutely :)

Issue #3999, opened 2020-11-04
Sync fails due to iconv error on alpine linux


tfmm:
### Expected behavior
Inbox loads with list of messages

### Actual behavior
Inbox or other folders never load, get a constant `Loading messages` message where message list should be. 

### Mail app

**Mail app version:** 1.6.0

**Mailserver or service:** custom IMAP from a server running Virtualmin

### Server configuration

**Operating system**: NC Docker Container from linuxserver.io - 

**Web server:** NC Docker Container from linuxserver.io - Nginx

**Database:** PostgreSQL

**PHP version:** 7.3.24

**Nextcloud Version:** 20.0.1

#### Client configuration
**Browser:** Chrome, Chromium, and Vivaldi, all latest versions

**Operating system:** Ubuntu 20.10 and Fedora 33

ChristophWurst:
Check your nextcloud.log for errors

tfmm:
Meant to mention, there are no errors in nextcloud.log or in the nextcloud logging interface.

ChristophWurst:
Okay let's try it differently: please try the sync from the CLI with `occ mail:account:sync <id>`, the `<id>` can be discovered by looking into the `oc_mail_accounts` table.

tfmm:
Ok, so getting an error with the manual sync: 

```
root@55623887849b:/config/www/nextcloud# sudo -u abc php occ mail:account:sync 6
[debug] Initial sync 6:INBOX - fetch all messages from IMAP took 2s

In ImapToDbSynchronizer.php line 235:
                                                                                                                           
  Sync failed for 6:INBOX: Return value of OCA\Mail\Address::getEmail() must be of the type string or null, bool returned  
                                                                                                                           

In Address.php line 78:
                                                                                                  
  Return value of OCA\Mail\Address::getEmail() must be of the type string or null, bool returned  
                                                                                                  

mail:account:sync [-f|--force] [--] <account-id>
```

Looking at the accounts table, everything appears to be proper data types.

ChristophWurst:
it's because of https://github.com/nextcloud/mail/pull/4007. we just fixed that :) you can download a patch from https://patch-diff.githubusercontent.com/raw/nextcloud/mail/pull/4007.patch

once applied the sync should work.

tfmm:
Still getting the same errors after patching the file and restarting the docker container.  For good measure I also re-installed the mail app and re-patched, no change.

ChristophWurst:
[...]

PR #4099, opened 2020-11-18
Catch when iconv returns false

ChristophWurst:
This happens when iconv doesn't work as expected, e.g. on Alpine.
Instead of causing a type error on the return value we can show a more
useful exception.

Ref https://github.com/nextcloud/mail/issues/3999


st3iny (approved):
Code looks good. It's this iconv thing again so I don't know how to test.

ChristophWurst:
Me neither :shrug:

Issue #3940, opened 2020-10-31
e-mail list column is too narrow, bulky and generally lacks in visual clarity


cptMikky:
#### Summary

As a user don't like the current layout in terms of font size, text margins and column width.

The folder list is unnecessarily wide while the e-mail list is too narrow and bulky, hiding too much information needlessly (even not-too-long subject lines are truncated) and squeezing a lot of information into a small space.

The e-mail content column, on the other hand, is so wide that the side margins are about 1/3 of the content width of unused space each (on a common desktop widescreen display) which gives more than enough space to extend the e-mail list column without sacrificing anything.

The "bulkiness" feel of the e-mail list column is, for me, caused mainly by these factors:

1. the column is too narrow which causes information loss and compression (see above)
2. the font of both Sender and Subject is unnecessarily large
3. there is no visual clue to separate the individual e-mails 
4. the distinction between read and unread e-mails is not visually strong enough
5. the attachment icon makes the list look visually inconsistent

#### Preferred solution

I suggest these fixes respective to the above mentioned factors:

1. allow manual resize of the column(s) (or make the e-mail list column about twice as wide)
2. decrease font size by 1-2 points and make the subject font color darker (or black, same as sender)
3. add a thin light line between the entries. This and 2. may allow for slightly tighter margins making the list more compact.
4. make unread e-mail background color slightly darker or a different color
5. have a fixed place for the paperclip icon that doesn't change visual arrangement of the text

ChristophWurst:
@jancborchardt 👀

dumblob:
I considered this a low-priority, so I didn't want to clutter this bug tracker. But it appeared here, so I feel obliged to raise my hand as I have exactly the same issues as @cptMikky . Making it less "bulkier" in one way or another would be very helpful for everyday users.

brueckner:
I fully agree! **However**, I believe this is one of those things that heavily depend on the user's taste and use-case. Personally, I couldn't care less about avatars in an email client to be honest, but it might be useful for others to visually scan for the email they're looking for. 

What I am trying to say is: there is no "one size fits all" solution here, I think.

Instead, I think this is something that needs to be **configurable**, just like in other email clients. Take Apple Mail for example:

<img width="548" src="https://user-images.githubusercontent.com/459512/98093903-d867d100-1e88-11eb-9ddb-9f907f1ec256.png">

I think it is reasonable to have "shiny" default settings (that is: spacious, avatars, etc.) with "power user" settings to make things more compact.

[...]

Issue #567, opened 2017-10-13
Long IMap Folder Name is unreadable


andreas-wolf:
### Steps to reproduce
1. create long imap folder name

### Expected behaviour
Name should be fully readable or at least the folder column should be resizeable/horizonal scrollable

### Actual behaviour
Name is cut off.
All I can see are folder names like 
INBOX.Archiv.Orders.Outgoing.A...
INBOX.Archiv.Orders.Outgoing.A...
INBOX.Archiv.Orders.Outgoing.A...

### Mail app
**Mail app version:** (see apps admin page)
0.7.3

ChristophWurst:
Do you see the full folder name if you hover it?

andreas-wolf:
In the left colum there is no hover text at all. Neither at e.g. the Sent folder nor at the other folders.

```Html
   <li class="ui-droppable"><a class="folder
	
	" href="/index.php/apps/mail/#accounts/1/folders/I_REMOVED_THE_ID%3D%3D">
	INBOX.Archiv.xxxxx.xxxxx.xxxxxxxxxx
</a>
<div class="folders"><ul></ul></div></li>
```
In the A-tag there are two tabs after class="folder
I've tested with Firefox 56.0.1 and Chrome Version 61.0.3163.100

Furthermore I see that the folders are not in alphabetical order:

INBOX.Archiv.Orders.R...
INBOX.Archiv
INBOX.Archiv.Orders.A..

ftoledo:
I have same behavior on 0.7.4 over Firefox 57
![recibidos ragnarok docksud com ar mail bbs cloud](https://user-images.githubusercontent.com/95441/32997555-7912f48a-cd70-11e7-8a06-f2cbc94467b8.png)

ChristophWurst:
this should be better with recent versions of Mail

PR #4123, opened 2020-11-24
Resizable Columns

brueckner:
- [x] BLOCKED requires a release of `@nextcloud/vue` that includes https://github.com/nextcloud/nextcloud-vue/pull/1759, e.g. `4.0.0-alpha.5`

**This pull request adds the capability to resize all columns inside Mail within a certain range that makes sense. Chosen column widths are saved automatically.**

---

Here's a demo GIF (3mb, 40sec):

nextcloud-mail-resizable-columns

---

## Implementation

It is using [splitpanes](https://antoniandre.github.io/splitpanes/) for the heavy lifting, which is pretty simple, seems well maintained and Vue3 compatible, if someone ever dares to upgrade... 😉

The settings are stored the same way as the `collect-data` flag for example and passed into the Vue instance straight from the template. Saving the settings is triggered on an event called `resized`.

Due to the structure of the components, there are two "stacked" `<Splitpanes>`, each with two `<Pane>` elements. The first is in `Home.vue` and controls the very left mailbox nav column `<Navigation>` and the main content region `<MailboxThread>`. The main content region holds the envelope list (middle column) and main thread/mail detail view (right column), which are wrapped by the second `<Splitpanes>` element in `MailboxThread.vue`.

There had to be some minor styling changes to existing elements due to the structural change. Visually, nothing should have changed except for the 7px wide column resize bars between the panes.

I chose to store the defaults for `size`, `min` and `max` in a local data object:

```
paneDefaults: {
	aside: {
		size: 25,
		min: 15,
		max: 40,
	},
	main: {
		size: 75,
		min: 60,
		max: 85,
	}
}
```

... these could be set as properties on the `<Pane>` elements directly, but I wanted to have them (comparable) in one spot, which makes changing them a little easier. If that's too verbose for you, we could move it to the `<Pane>`, that's fine with me.

_Note: If you look into the diff of `src/components/MailboxThread.vue` it seems like there are a lot of changes in the template section, but I actually just wrapped the other components with no changes to them whatsoever._ 🤷‍♂️

Related issues: #4098 #567 

Fixes https://github.com/nextcloud/mail/issues/3940


jancborchardt:
@brueckner we need to keep in mind that "resizable" means that every single person need to do it manually. :)

While we can talk about making the columns resizable, our main concern should be a great default experience, on all kinds of different screen sizes, without manual modification (also possibly back and forth as window size and such can be changed). So this could e.g. better be done by using "vw" (viewport width) units to define the width of the navigation and conversation list – as I mentioned in the design review call.

And any of these global layout changes we also need to do globally, in the Vue components, so they are automatically applied to all other apps. :)

cc @ChristophWurst @GretaD

ChristophWurst:
Yep, I would also suggest that we start with a better default layout that works well with most resolutions. The manual resizing can be an additional enhancement.

brueckner:
> @brueckner we need to keep in mind that "resizable" means that every single person need to do it manually. :)
> While we can talk about making the columns resizable, our main concern should be a great default experience, on all kinds of different screen sizes, without manual modification [...]

I believe this implementation does exactly that. We **can** provide a better default **and** users are able to modify it to their liking. If people don't care about it, it doesn't get in the way either, because it is super unobstrusive.

Keep in mind that pretty much **any** email application allows you to do that. Apple Mail, Thunderbird, Outlook, Rainloop, Roundcube, OpenXchange (the one Ionos uses), etc.. That makes it a **basic** feature users have come to expect from an email client and it was one of the first things I personally tried to do with Nextcloud Mail as well.

> So this could e.g. better be done by using "vw" (viewport width) units to define the width of the navigation and conversation list – as I mentioned in the design review call.

It is using percentage-based values, which – in this case – results in the exact same behaviour as using `vw`.

jancborchardt:
Sure, as said I'm not strictly against the resizability :) just that we should make sure we get the basics right for regular people before we directly focus on customizability.

- It should be done in the Vue components instead of only the Mail app. E.g. the width of the AppNavigation should carry over to other apps like Calendar, Contacts, etc and AppContentList also to Contacts, cause otherwise it would look very strange with different widths for different apps.
- We need to make sure the defaults work and look great on different screen sizes. Ref https://github.com/nextcloud/nextcloud-vue/issue/1561
- Currently this introduces further vertical bars which look oldish and introduce visual noise. Can we do without them? E.g. macOS Mail doesn't have visible drag bars as it calls unnecessary attention to a feature the majority of people should ideally not need.

Does that sound good? :)

brueckner:
Thanks for the clarification :) I am 100% on board with the fact that it needs to look good by default! I'd argue that each app might have different optimal defaults though, so `min`, `max` and (initial) `size` should be controllable from "the outside".

The way to go about this is moving the whole thing to globally used components (like AppNavigation and AppContentList), right? I didn't look at Calendar, Contacts, etc. yet, but they're all Vue Apps, correct? 

I am wondering about the process here, as this needs to touch multiple apps and probably multiple different teams for approval? Is it the app-team's responsibility to implement that, or is it mine? I guess I am just a little afraid that this gets too complicated and dies down because of that.

The most important part is increasing the potential maximum width of the envelope list on large screens. So, if it makes things easier, we could also think about just:
- improving the responsive styling of the second column by simply increasing the 300px limit [mentioned here](https://github.com/nextcloud/nextcloud-vue/issues/1561) to something reasonable (not as good as resizing imho, but will be a big improvement with close to zero effort)
- implementing this resizable columns feature only on the inner level (making only columns 2 and 3 resizable while maintaining the AppNavigation's width)

> Currently this introduces further vertical bars which look oldish and introduce visual noise. Can we do without them? E.g. macOS Mail doesn't have visible drag bars as it calls unnecessary attention to a feature the majority of people should ideally not need.

Yea, we can reduce this to 1px width without handles. It'll be a little harder to use that way though.

ChristophWurst:
> The way to go about this is moving the whole thing to globally used components (like AppNavigation and AppContentList), right? I didn't look at Calendar, Contacts, etc. yet, but they're all Vue Apps, correct?

Through the components it becomes kind of global.

> I am wondering about the process here, as this needs to touch multiple apps and probably multiple different teams for approval? Is it the app-team's responsibility to implement that, or is it mine? I guess I am just a little afraid that this gets too complicated and dies down because of that.

So most of our components evolve and grow organically as we need them in our apps. There is no top down design process, it's more like we push ideas from apps to a common standard (with a few modifications or generalizations) and then other apps adapt to this. E.g. the Talk app recently received a new settings UI based on a modal component and now we are adopting this for Mail as well to streamline the app designs.

I'm by far not a designer and also not making the decisions when it comes to UI changes but I think this topic of the adjusted column widths and/or making them resizable would be a good topic to discuss in the weekly Nextcloud design call. @jancborchardt would that make sense and could you schedule this topic?

jancborchardt:
> Thanks for the clarification :) I am 100% on board with the fact that it needs to look good by default! I'd argue that each app might have different optimal defaults though, so `min`, `max` and (initial) `size` should be controllable from "the outside".

In the interest of keeping things simple first and seeing if we need additional complexity later, let’s start out simple with the same size being used for all apps. Otherwise again, people would need to adjust it for every single app individually. (Also here as well – the defaults should work fine for 95%.)

> The way to go about this is moving the whole thing to globally used components (like AppNavigation and AppContentList), right? I didn't look at Calendar, Contacts, etc. yet, but they're all Vue Apps, correct?

Most of this stuff was already explained by @ChristophWurst – the only thing we need to look out for is that some apps are _not_ Vue apps yet, most notably Files. However Files only has the navigation and sidebar, not a second column.

> * improving the responsive styling of the second column by simply increasing the 300px limit [mentioned here](https://github.com/nextcloud/nextcloud-vue/issues/1561) to something reasonable (not as good as resizing imho, but will be a big improvement with close to zero effort)

Yep, that would be "improving the default" as mentioned. :) That’s step 1 anyway.

> * implementing this resizable columns feature only on the inner level (making only columns 2 and 3 resizable while maintaining the AppNavigation's width)

Yep, that could be a good start! :) It would also resolve the issue that Files doesn’t use Vue yet, and then we also don’t need to take into account the right sidebar.

So in this case only apps which use the 3-column-layout would be affected, which from the core apps is Mail and Contacts – great small start for this which would solve the main issues without introducing too much overhead or being blocked by external factors.


------

> I'm by far not a designer and also not making the decisions when it comes to UI changes but I think this topic of the adjusted column widths and/or making them resizable would be a good topic to discuss in the weekly Nextcloud design call. @jancborchardt would that make sense and could you schedule this topic?

Not sure it does make sense for a design call to be honest. @brueckner referenced in https://github.com/nextcloud/mail/pull/4123#issuecomment-733631423 that basically all other apps have this capacity to some extent, so it does make sense for us as well as an expert feature.
As said I’m not against it, I just want to make sure that the default which the majority of people use is getting as much (and more) focus as the customization solution for the few expert users. :) Hence @brueckner’s latest suggestions are great: Start small, with only touching the 2nd column, both fixing its default width as well as implementing resizability – both in the components which will then be used by Mail and Contacts.

vince-from-nice:
> > @brueckner we need to keep in mind that "resizable" means that every single person need to do it manually. :)
> > While we can talk about making the columns resizable, our main concern should be a great default experience, on all kinds of different screen sizes, without manual modification [...]
> 
> I believe this implementation does exactly that. We **can** provide a better default **and** users are able to modify it to their liking. If people don't care about it, it doesn't get in the way either, because it is super unobstrusive.
> 

It's so true !

Thanks to Nextcloud, I almost managed to completely degoogleifyed myself. The only thing left is Gmail but as it stands, the interface of the mail app is almost unusable for me. I am ready to make efforts by losing responsiveness and especially the perfect threading of Gmail. But frankly the limit of the width of the 2nd column is much too small and it drives me crazy. We just have space to have a few characters to display emails titles. It's even more frustrating on 4k displays which are more and more common. Personally I would like to leave something like 40% max for the 3rd column showing the body of the messages. But everyone should be able to choose this setting to their liking, as is the case as has been said for the vast majority of email clients (well not for Gmail but they don't even have a 3rd column) .

Sorry if my comment is not very constructive but it comes from the heart of a future user of the mail app.

Apparently the solution by leaving the size of the 1st column fixed should do the trick: it is certainly simpler and above all the size of this 1st column has never been a problem unlike the 2nd. So I hope that the PR can be merged soon.

Either way happy new year and a big thank you to the Nextcloud mail app developers and to @brueckner.

brueckner:
Hey! We've been (and are) busy with other projects but decided to keep pushing this further, as this is close to being finished.

I updated the PR and **removed** the possibility to resize the first column. The second (envelope list) and third (mail thread view) columns are still resizable, as in the original proposal. Another notable change is of visual nature. I have removed the drag-handle, as suggested by @jancborchardt. The drag handle area is now invisible, but wide enough (9px with 4px left and 4px right of the vertical 1px wide column separator). The 9px width make it comfortable enough to grab and resize the column, without having to "snipe" for the 1px border.

I was gonna create another quick video, but apparently QuickTime keeps crashing for some reason... 🤷‍♂️ 

Also: `psalm` seems to be complaining in build pipeline. "It works on my machine" ™️   though... am I missing something?

Screenshot - iTerm app - 2021-01-13 at 11-28

jancborchardt:
Thanks for picking it up again @brueckner!

So when testing for responsiveness by zooming, I encounter one bug, and one behavior which is a bit strange:
- Bug: Sometimes a big piece of whitespace appears above the message list, on certain zoom levels (using Firefox latest)
- The width behavior of the middle column is great when you change the width of the viewport, but when you zoom in/out it seems strange, changing the size of everything else, but keeping the width kind of in place. Maybe that’s the proper behavior or intended, just seems a bit off? @ChristophWurst @ma12-co what do you think?

https://user-images.githubusercontent.com/925062/104626473-435a3300-5696-11eb-9e95-281e0103be3d.mp4


------------

For comparison, here is current master:

https://user-images.githubusercontent.com/925062/104626469-405f4280-5696-11eb-9162-7edb7a63e7ba.mp4

brueckner:
Thanks for testing this, @jancborchardt ! I was able to reproduce that in Firefox (Chrome looks fine) and it's an odd one. It only happens when zooming in, not when zooming out (you could zoom in until the sidebar disappears, zoom back out a few steps and it will not show up). Also, when the bar appears and you do a reload, it will be gone, altho the zoom-step is still the same. The `50px` offset is coming from the core's precompiled `server.css`:

```
.app-content-list {
  position: -webkit-sticky;
  position: sticky;
  top: 50px;
  ...
}
```

I've seen a few Firefox-bugs in relation to `position: sticky` already, so I think this could be one of them. 

Adding ...
```
.app-content-list {
  top: 0;
}
```
... in the `MailboxThread.vue` component fixes it and doesn't seem to impact anything else. Here's a before & after:

NCM - Firefox - 2021-01-20 at 11-02

NCM - Firefox - 2021-01-20 at 11-06

brueckner:
Heya, did you get a chance to take a look at this with my latest fix @jancborchardt ?

Anything else that needs to be done with this @ChristophWurst ?

GretaD:
@brueckner I tested it and all looks good. I think what is left to be done is that, this should move to nc/vue. Quoting Jan here 👇

>     * It should be done in the Vue components instead of only the Mail app. E.g. the width of the AppNavigation should carry over to other apps like Calendar, Contacts, etc and AppContentList also to Contacts, cause otherwise it would look very strange  with different widths for different apps.
>

brueckner:
Thanks a lot for testing @GretaD! 

> * It should be done in the Vue components instead of only the Mail app. E.g. the width of the AppNavigation should carry over to other apps like Calendar, Contacts, etc and AppContentList also to Contacts, cause otherwise it would look very strange  with different widths for different apps.

Hmm, I remember a conversation where we talked about that not being important for now, if we limit the resizability to the inner columns of the Mail app, because no other apps are currently using a three-column layout anyway. 🤔 

Is that not the case anymore?

jancborchardt:
@brueckner I would say that @ChristophWurst is the better person to comment regarding that, but I mentioned in 2 comments here from November that we should do this in the components, so this is standardized and e.g. the Contacts app can also benefit from this as per user expectation. :)

benks-io:
I understand, that it's a good way to make features available at a higher level before putting energy and time in implementing it for a lower one. Here there is already a working solution available.
As discussed, the resizability is limited to the second and third column of the Mail app. This provides a necessary improvement for daily and professional use of the Mail app, which is an exclusion criteria for some users trying out an app probably with raised expectations as this is a featured one.

Migrating the above solution to nc/vue seems to be a major intervention and probably still needs some time and developers to be done. With that perspective I'd like to suggest the pull request to be merged and make the resizability available for an upcoming Version. I suppose there is a way to overwrite or switch off this feature by nc/vue when it's available later on.

Enabling this would bring the usability improvement to users, still having accounts on instances with earlier versions of Nextcloud which may give the Mail App and the NC "HUB" Idea more acceptance sooner.

luvis:
I'm very glad this is already being discussed. I also believe this to be an important default feature in any email app. Both the option to resize columns and perhaps also the option to change from diagonal to horizontal split view. Either way, as @benks-io mentioned, when there is already a working (good) solution available, I would suggest using it even if there is an even better solution down the road, especially if it's quite far down the road which seems likely in this case. So +1 for implementing it now =)

ChristophWurst:
Alright, so who's down with maintaining this for at least five upcoming months? That is roughly the time we currently planned with other roadmap features. Maintenance includes finishing the task of moving this to nc/vue, including it here again after the release upstream and taking care of any bugs and regressions thar arise.

luvis:
@ChristophWurst I wish I was, however I'm no developer. Whilst hoping @brueckner is up for the task ;) I thought about your comment about changing the default view. As I am no developer, I tried dabbling with stylish and came up with a few changes that I believe gave it a better overview and made it more readable. This might be the wrong place for it, if so I'm sorry. I just thought if it's easy to implement the css, maybe that's a start at least? Anyway, I've pasted the css here, sorry if the code is bad. I changed the spacing, removed the duplicate message header to free up some space since I don't see the need for two and made the remaining one sticky instead and also changed the font size a bit. If this helps at all, I'm glad I could do something. Screenshot below.

.app-content-list {
    min-width: 425px;
    max-width: 500%;
    left: 1px;
}
.app-navigation {
    width: 200px!important;
    left: 22px;
}
.content {
    font-size: 90%!important;
}
.app-navigation-toggle {
    top: 12px!important;
    margin-right: -10px!important;
}
.app-navigation-new button {
    width: 93%!important;
    padding-left: 34px;
    margin-right: 19px!important;
    margin-left: 5px;
    margin-left: -4px;
    background-color: #1fa1ea;
    border-color: #1fa1ea;
    color: white;
}
.envelope--header {
    top: 50px;
    position: sticky;
    display: flex;
    margin-bottom: 0px!important;
    border-bottom: 1px solid #dbdbdb!important;
    background-color: white;
    z-index: 1;
}
#mail-thread-header {
    visibility: hidden;
    position: unset!important;
    padding: 0px 0!important;
    height: 0px!important;
}
.subject {
    font-weight: 700;
    font-size: medium;
}
nextcloud-mail-screenshot

dumblob:
@luvis thanks, I think it's a decent design (though I can't judge whether it fits Nextcloud).

But please, **never do any sticky stuff** in UIs (even material design strongly recommends at max **one** widget/button/input/... which is sticky). On smaller screens (notebooks, tablets, smartphones, ...) it completely unnecessarily eats lots space. The current/former solution with headers is far better just because it's not sticky. Otherwise yes, duplicate info is useless, so saving on that side is always welcome :wink:.

luvis:
@dumblob 
Thanks for letting me know not to use sticky. The reason I chose to make it sticky is actually because the default view had a sticky header already. I just thought it was unnecessary to have two headers in the same message since it takes up a lot of space so I removed the old sticky one and made the header that includes the reply-button sticky instead. So no space is lost, only gained =)

dumblob:
@luvis oh, then I apologize, I kind of misunderstood (I didn't test your code). In that case the current "sticky" shall also be changed to non-sticky if possible.

szaimen:
In order to bring this PR forward: 
Since the mail team doesn't seem to want to support this PR unless it is provided upstream via the nextcloud/vue library, would you, @brueckner, be so nice to implement this upstream? :) 
I know, it was already a lot of work and the community is already very thankful that this PR has come this far, so understandable if this is too much work :)

ChristophWurst:
It's not about what we *want* but what we have time for :)

We discussed from the beginning that this should go into the components lib, so yeah, now it's some additional work to move it there now that it was done here.

brueckner:
Thanks for the supporting words everyone! 🎉 

I'll look into moving this into the components lib this week. It's kind of a daunting task though to be honest, as I need to make sure this works across the board in all the other apps that implement this component. 🤷‍♂️  Having worked on the Mail app exclusively up until now, this makes me feel a little anxious. Anyway... I will report back.

ChristophWurst:
> Having worked on the Mail app exclusively up until now, this makes me feel a little anxious.

No worries, we are here to support you :)

brueckner:
Alright, so this is not that straight forward. 

The abstracted components in **nextcloud/nextcloud-vue** don't have access to the global store for obvious reasons. Moving the Splitpane implementation to [AppContent](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppContent/AppContent.vue) means two things: 

a) current user settings have to be passed into the component as a prop
b) resize events have to be emitted up to the parent

This leaves the Mail app – once again – in charge of reading from and writing to the database to load and store the settings. Consequently, this leaves any other app that wants to use the resizable columns in charge of doing the same thing with their own implementation, which kind of defeats the purpose of this whole abstraction-operation.

Another issue I ran into is a wrapping `div` on the top level inside `<AppContent>`. This causes a problem with using slots, because slots can only be used on the main level of a component. In order for this to work, I need to create two named slots: `main` and `sidebar`, but they _cannot_ be created like this: 

```js
<AppContent>
    <div id="app-content-wrapper> <----- offending div on first level
        <template slot="sidebar">...</tempalte>
        <template slot="main">...</tempalte>
   </div>
</AppContent>
```

Intead, it must look like that:


```js
<AppContent>
    <template slot="sidebar">...</tempalte>
    <template slot="main">...</tempalte>
</AppContent>
```

... which means that the `<div id="app-content-wrapper />` needs to be moved into the `AppContent` component, but I don't know if that breaks the other apps implementing it... 🤯 

---

I feel like the second problem can be solved somehow, but the first one kills it for me. Well, at least it kills the current approach of storing the settings in the database. **What if** we use `localStorage` instead? Then we could move the responsibility to read/write into the `AppContent` component, making the whole thing self-contained. No props or emitters needed. The only prop we _could_ introduce is a key to identify the app that implements `AppContent` and scope the settings that way. If a key is not passed down as a prop, it will use the global column sizes.

This approach does have the obvious downsides of `localStorage` though and it transforms `AppContent` from a dumb component into a smart one, which is maybe not what you guys have envisioned for components in nextcloud/nextcloud-vue. 🤷‍♂️ 

Any thoughts on that @ChristophWurst?

brueckner:
Bump. Any thoughts on this? This is kind of a crossroads-decision and I'd love to get some feedback before marching on into the wildernes, only to find out I was walking in the wrong direction. ;)

ma12-co:
cc @juliushaertl and @skjnldsv

jancborchardt:
@brueckner from the user perspective, using `localStorage` would mean that while the resizing of the column would be the same between Mail and Contacts, it would be only for that specific device / browser, correct? I’d say that’s fine, and we can always improve on that behavior in the future. The important part is that it’s solid across apps. :)

brueckner:
> @brueckner from the user perspective, using `localStorage` would mean that while the resizing of the column would be the same between Mail and Contacts, [...]

It **could** be the same, but doesn't have to. I'd argue that a column with email subjects might need more space than a list of contact names. So I think I'd prefer to have a seperate setting per app.

> [...] it would be only for that specific device / browser, correct? 

Yes, exactly. That's one potential downside, but maybe even a good thing if I think about it. Let's say I've got a 13" Laptop for work and a 27" screen for my home computer – I might want to have different settings for these. In that case, using localStorage would be even better than tying it to the account.

skjnldsv:
> ... which means that the `<div id="app-content-wrapper />` needs to be moved into the `AppContent` component, but I don't know if that breaks the other apps implementing

Fine for me, would just be released as a major.
We just need to allow devs to put stuff to the default slot.

```vue
<AppContent>
	<div>whatever data would be in the default slot</div>
	<AppContentList #list></AppContentList>
	<AppContentDetails #details></AppContentDetails>
</AppContent>
```
___

:warning: But **careful**, your example are mixing multiple concepts. The app-content-wrapper is here to contain the details and the list of the content (or just the content) The AppContent is here to contain the Navigation, the Content, and the Sidebar.
Here is a complete example:

```vue
<Content app-name="forms">
	<AppNavigation>
		<AppNavigationItem title="My title" icon="icon-category-enabled">
		...
	</AppNavigation>
	<AppContent>
		<AppContentList #list></AppContentList>
		<AppContentDetails #details></AppContentDetails>
	</AppContent>
	<AppSidebar title="cat-picture.jpg"></AppSidebar>
</Content>

```

ChristophWurst:
@brueckner I would like to have some *open hacking* again tomorrow in our public Groupware call. We've done this the last few Fridays and it helped us discuss a few changes or help each other with problems. Are you available tomorrow to join us? Then we can talk this through and hopefully make some progress :)

brueckner:
> @brueckner I would like to have some _open hacking_ again tomorrow in our public Groupware call. We've done this the last few Fridays and it helped us discuss a few changes or help each other with problems. Are you available tomorrow to join us? Then we can talk this through and hopefully make some progress :)

Well... turns out the Kindergarten had to close down today due to some COVID-cases, so I'm not sure if I can squeeze in _open hacking_ today. I'll hang out in the chat today though.

Anyway - I actually worked on it quite a bit yesterday and I think I've got a pretty good and clean solution already. I'm just doing some cleanup now and will push it here & create another PR in the `nextcloud/nextcloud-vue` repo today.

brueckner:
Alright, so here goes:

I was able to drastically simplify this by using `localStorage`. Most of it is now happening in the parent component. All that was left to do in the Mail app was to put the column contents into their respective slots, which can be seen in [MailboxThread.vue](https://github.com/nextcloud/mail/pull/4123/files#diff-2ef5622c7090ce9f7e480022e8f1217a0c811b608ecf6335e71758ec3c43768e). The only notable change (along some minor CSS) is the removal of `<div id="app-content-wrapper">` as mentioned [here](https://github.com/nextcloud/mail/pull/4123#issuecomment-784135998), which has now been moved up to the parent.

Here's the accompanying PR with more details on the new storage approach in `nextcloud/nextcloud-vue`: https://github.com/nextcloud/nextcloud-vue/pull/1759

ChristophWurst (approved):
Tested, works and code looks good!

ChristophWurst:
set as draft so nobody merges accidentally before https://github.com/nextcloud/nextcloud-vue/pull/1759 is merged and integrated here

brueckner:
Pushed a small update to rename `aside` to `sidebar` in sync with the latest changes in https://github.com/nextcloud/nextcloud-vue/pull/1759.

ChristophWurst:
If we update to `4.0.0-alpha.5` of nc/vue then we can integrate this. As far as I know it needs a few minor adjustments to work with that version. We can take https://github.com/nextcloud/contacts/pull/2246 as reference implementation.

skjnldsv:
4.0.0 is published

ChristophWurst:
If I'm not mistaken @GretaD couldn't push to your fork repo @brueckner and therefore created https://github.com/nextcloud/mail/pull/5196

GretaD:
this is implemented here: #5201 so i guess we can close this one?

GretaD:
🎉🎉🎉

Issue #3760, opened 2020-10-08
"Register as application for mail links" feature does not work


IchbinkeinReh:
### Expected behavior
After i clicked "Register as application for mail links" and then clicked on a website on a mailto link i get redirected to my nextcloud instance and see the compose new mail window with the to field and optional subject and body prefilled

### Actual behavior
After i clicked "Register as application for mail links" and then clicked on a website on a mailto link i get redirected to my nextcloud instance and see my email accounts

Mail app
Mail app version: v1.5.0

Mailserver or service: Strato

Server configuration
Operating system: Ubuntu 20.04

Web server: Apache

Database: MariaDB

PHP version: 7.4

Nextcloud Version: 20.0.0

Client configuration
Browser: Chrome 81

Operating system: Ubuntu 20.04

ChristophWurst:
This might be a regression due to the routing change of https://github.com/nextcloud/mail/pull/3439 because at the generated redirect we still use the old hash routing: https://github.com/nextcloud/mail/blob/f8543ada0b781a8451fa930a62a38666df2e5114/lib/Controller/PageController.php#L221

PR #4128, opened 2020-11-26
Fix mailto route

GretaD:
fixes: #3760

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -221,10 +221,17 @@ class PageController extends Controller {
 			function (&$value, $key) {
 				$value = "$key=" . urlencode($value);
 			});
-
-		$hashParams = '#mailto?' . implode('&', $params);
-
-		$baseUrl = $this->urlGenerator->linkToRoute("mail.page.index");
-		return new RedirectResponse($baseUrl . $hashParams);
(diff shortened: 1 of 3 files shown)


ChristophWurst:
I think we have to define mailto in https://github.com/nextcloud/mail/blob/master/appinfo/routes.php as well and create a "fake controller method" like we have for the other routes that all show the index in PageController

lib/Controller/PageController.php
```diff
@@ -221,10 +221,16 @@ public function compose(string $uri): RedirectResponse {
 			function (&$value, $key) {
 				$value = "$key=" . urlencode($value);
 			});
-
-		$hashParams = '#mailto?' . implode('&', $params);
```
ChristophWurst, line 225: you still need the `$params`. You can try passing them as second argument of `linkToRoute`  because we might add them to the URL then. If that doesn't work append this to the `$baseUrl` like before but without the `#mailto?` part :)

ChristophWurst (approved):
Didn't test but code looks good :)

ChristophWurst:
Tests fail now

ChristophWurst:
now squash it :)

ChristophWurst (approved):

PR #4130, opened 2020-11-26
Change account settings into modal

GretaD:
it looks like this: 
accountsettings

ref #3569


GretaD:
Opened a new pr because the last one was too old and too many conflict to resolve :)

src/components/AccountSettings.vue
```diff
@@ -0,0 +1,176 @@
...
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<AppSettingsDialog
+		:open.sync="showSettings">
+		<AppSettingsSection
+			:title="t('mail', 'Account settings')">
```
ChristophWurst, line 29: is it possible to omit the title?

src/components/AccountSettings.vue
```diff
@@ -0,0 +1,176 @@
...
+			<strong>{{ displayName }}</strong> &lt;{{ email }}&gt;
+			<a
+				v-if="!account.provisioned"
+				class="button icon-rename"
+				href="#account-form"
+				:title="t('mail', 'Change name')" />
+		</AppSettingsSection>
+		<AppSettingsSection :title="t('mail', 'Alias')">
+			<AliasSettings v-if="!account.provisioned" :account="account" />
```
ChristophWurst, line 38: we should move them up to the general account settings (as discussed with @jancborchardt in the dev chat) so the account email and aliases appear like one list

src/components/AccountSettings.vue
```diff
@@ -0,0 +1,176 @@
...
+				{{ t('mail', 'A signature is added to the text of new messages and replies.') }}
+			</p>
+			<SignatureSettings :account="account" />
+		</AppSettingsSection>
+		<AppSettingsSection :title="t('mail', 'Writing mode')">
+			<p class="settings-hint">
+				{{ t('mail', 'Preferred writing mode for new messages and replies.') }}
+			</p>
+			<EditorSettings :account="account" />
+			<div v-if="!account.provisioned" class="section">
```
ChristophWurst, line 51: account settings are not part of the writing mode, right?

src/components/AccountSettings.vue
GretaD: i tried actually since was easier to keep the titles where they were, but it doesnt work. Thats why i moved the desc here. I will consult it @ma12-co

ma12-co: If you remove both the sections and left navigation, you're better off with a plain modal component here imo

ma12-co: Ok, I spoke with @GretaD and she told me she removed the titles from your custom components and added them in the settingsection components, so all good :)

GretaD: Cool, so lets keep the sections with title since its not possible to remove them and I guess we will have the same result.  I am not going back to modal after I worked on it, and then worked on it again to change into AppSettingsDialog component 😇

jancborchardt:
Cool!  :) Design feedback:

- [ ] Clicking on the "edit" button of the main account still tries to load the `#account-form` route, leading to an endless spinner in the main screen behind the modal
- [ ] Interactions like deleting aliases, saving signatures and such takes a _long_ time to load in the beginning, multiple seconds?
- [ ] The "Writing mode" radio buttons as well as the "Mail server" form are wrapped in a div with the class `.section` which gives it unneeded and inconsistent white space (see screenshot below). So that class can/should be removed. :)

An issue to be fixed in the component @ma12-co: `.app-settings` has padding which results in the content being cut off on the top and bottom as well as the scrollbar to float in the air
image

GretaD:
> Cool! :) Design feedback:
> 
>     * [ ]  Clicking on the "edit" button of the main account still tries to load the `#account-form` route, leading to an endless spinner in the main screen behind the modal
> 
edit_account

>     * [ ]  The "Writing mode" radio buttons as well as the "Mail server" form are wrapped in a div with the class `.section` which gives it unneeded and inconsistent white space (see screenshot below). So that class can/should be removed. :)
> 
Will remove it.
>     * [ ]  Interactions like deleting aliases, saving signatures and such takes a _long_ time to load in the beginning, multiple seconds? Probably unrelated to this change though, still a bit strange.
> 
signature_alias

About the delay, i cannot reproduce it. all looks good at my end. Also the edit button, as in the gif takes you directly to the form without any spiner.

jancborchardt:
> About the delay, i cannot reproduce it. all looks good at my end.

Now I can also not reproduce it anymore – it only seemed to be the first few times I tried to set it. Might have just been a setup-specific issue, never mind.

> the edit button, as in the gif takes you directly to the form without any spiner.

Here’s how it looks like for me: Causes refreshing the mail which is currently open in the background (this as it happens in the Priority inbox):
Peek 2020-11-30 23-30

src/components/AccountSettings.vue
```diff
@@ -133,6 +134,12 @@ export default {
 					throw error
 				})
 		},
+		handleClick() {
+			this.$refs.accountForm.$el.scrollIntoView({
+				behavior: 'smooth',
```
ChristophWurst, line 139: :v:

src/components/EditorSettings.vue
```diff
@@ -21,10 +21,6 @@
 
 <template>
 	<div class="section">
```
jancborchardt, line 23:
```suggestion
	<div>
```

This leads to the additional whitespace around the radio buttons, no? :)

jancborchardt (approved):
Looks nice, thanks @GretaD! :)

ma12-co (commented):
:tada:

Issue #4148, opened 2020-11-29
Cron mail sync failed: Special mailbox drafts does not exist


acsfer:
After upgrade to **20.0.2** from 19.0.5, and Mail **1.7.0**, it seems **Cron** is facing a problem.
Hundreds of messages like these in the Log about Drafts mailbox:

```
{
  "reqId":"Jduo7SDk7s4SXwleadga",
  "level":3,
  "time":"2020-11-29T10:25:12+01:00",
  "remoteAddr":"",
  "user":"--",
  "app":"mail",
  "method":"",
  "url":"--",
  "message":{
    "Exception":"OCP\\AppFramework\\Db\\DoesNotExistException",
    "Message":"Special mailbox drafts does not exist",
    "Code":0,
    "Trace":[
      {
        "file":"/apps/mail/lib/Listener/MailboxesSynchronizedSpecialMailboxesUpdater.php",
        "line":74,
        "function":"findSpecial",
        "class":"OCA\\Mail\\Listener\\MailboxesSynchronizedSpecialMailboxesUpdater",
        "type":"->"
      },
      {
        "file":"/lib/private/EventDispatcher/ServiceEventListener.php",
        "line":76,
        "function":"handle",
        "class":"OCA\\Mail\\Listener\\MailboxesSynchronizedSpecialMailboxesUpdater",
        "type":"->"
      },
      {
        "file":"/3rdparty/symfony/event-dispatcher/EventDispatcher.php",
        "line":251,
        "function":"__invoke",
        "class":"OC\\EventDispatcher\\ServiceEventListener",
        "type":"->"
      },
      {
        "file":"/3rdparty/symfony/event-dispatcher/EventDispatcher.php",
        "line":73,
        "function":"callListeners",
        "class":"Symfony\\Component\\EventDispatcher\\EventDispatcher",
        "type":"->"
      },
      {
        "file":"/lib/private/EventDispatcher/EventDispatcher.php",
        "line":86,
        "function":"dispatch",
        "class":"Symfony\\Component\\EventDispatcher\\EventDispatcher",
        "type":"->"
      },
      {
        "file":"/lib/private/EventDispatcher/EventDispatcher.php",
        "line":98,
        "function":"dispatch",
        "class":"OC\\EventDispatcher\\EventDispatcher",
        "type":"->"
      },
      {
        "file":"/apps/mail/lib/IMAP/MailboxSync.php",
        "line":125,
        "function":"dispatchTyped",
[...]

PR #4149, opened 2020-11-30
Handle special mailbox sync when they don't exist

ChristophWurst:
When we update the account's special mailbox after a mailbox sync it's possible that any of the three special mailboxes don't exist. In that case we don't assign anything and carry on. This is fine.

Fixes https://github.com/nextcloud/mail/issues/4148


brueckner (approved):
Looks good to me 👌

PR #3208, opened 2020-06-05
Fix formatting when it is checked individually from the account default

GretaD:
fixes #3114

PR #4154, opened 2020-11-30
Fix initializing the editor mode

ChristophWurst:
In the `data` of the Composer component we initialize the editor mode
with 'html' and update the mode as soon as the alias has been set (this
is done async). The uninitialized value is `-1` and not `undefined`,
hence before the patch the editor more always switch to `html` even when
the account was set to `plaintext`.

To test set your account to plain text editing and then either start a new message or start a reply. On master the editor uses formatting, on this branch it doesn't.

Regression of https://github.com/nextcloud/mail/pull/3208.


GretaD (approved):
Tested and works as expected

PR #4160, opened 2020-12-01
Switch to vue2 package of the CKEditor

ChristophWurst:
The main package targets Vue 3, so let's use the Vue 2 version for now until we update the full app.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1969,10 +1969,10 @@
         "lodash-es": "^4.17.15"
       }
     },
-    "@ckeditor/ckeditor5-vue": {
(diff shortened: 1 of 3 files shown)


StCyr:
Ouch: Not a single toolbar element anymore here (FF latest):

image

ChristophWurst:
Well there are two tools: *undo* and *redo* :wink: 

Do you have formatting enabled? Check the ... menu on bottom left

StCyr:
ahahah 🤦 thanks, works great 👍

StCyr (approved):

PR #4161, opened 2020-12-01
Automaticaly sync mailbox on opening

StCyr:
Fixes https://github.com/nextcloud/mail/issues/3951

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -139,6 +139,9 @@ export default {
 		},
 		mailbox() {
 			this.loadEnvelopes()
+				.then(() => {
+					this.sync(false)
+				})
 		},
 		searchQuery() {
 			this.loadEnvelopes()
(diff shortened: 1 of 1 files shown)


GretaD:
right now it looks like this. Did we move the overlay spinner and kept only the load more one? @ChristophWurst 
sync

StCyr:
This PR is somewhat linked to this one => https://github.com/nextcloud/mail/pull/3686

Change made to Composer.vue in https://github.com/nextcloud/mail/pull/3686 should probably be reverted when both PR are implemented

ChristophWurst:
> right now it looks like this. Did we move the overlay spinner and kept only the load more one? @ChristophWurst

The overlay spinner at the top of the list is expected. It shows every time a mailbox is refreshed.

ChristophWurst (approved):
Works really well

Issue #4104, opened 2020-11-19
Various issues when writing emails with latest versions of Firefox


doc75:
### Expected behavior

Writing mail, I should be able to delete text by using delete or backspace key. I also should be able to copy/paste text.

### Actual behavior

When I answer a mail, I can delete 1 or 2 chars with backspace and then it stops working and I cannot anymore delete chars (even with delete key). The only solution is to select text with mouse and type another text to overwrite it (though the final result that the recipient will receive may not look the same as what you see).

Copying text to the clipboard doesn't work neither.

Issues are not reproducible in Chromium 84.

### Mail app

**Mail app version:** 1.4.1

**Mailserver or service:** Personal server (Dovecot + Postfix)

### Server configuration

**Operating system**: Official Docker fpm image 

**Web server:** Nginx

**Database:** PostgreSQL

**PHP version:** Official docker image

**Nextcloud Version:** 19.0.4

#### Client configuration
**Browser:** Firefox 83

**Operating system:** Windows

doc75:
I can find in the web console this type of error when the issue occurs:
```
Uncaught CKEditorError: a is undefined                                                                    mail.js:1:64080
    <anonymous> VueJS
    value range.js:348
    value deletecommand.js:107
    value model.js:805
    value model.js:231
    value deletecommand.js:76
    decorate observablemixin.js:255
    fire emittermixin.js:209
    e observablemixin.js:259
    value commandcollection.js:68
    value editor.js:288
    value delete.js:55
    fire emittermixin.js:209
    c deleteobserver.js:87
    u deleteobserver.js:50
    fire emittermixin.js:209
    value domeventobserver.js:96
    value keyobserver.js:28
    value domeventobserver.js:79
    fire emittermixin.js:209
    r emittermixin.js:235
    construct self-hosted:2653
    s mail.js:1
    r mail.js:1
    construct self-hosted:2653
    f mail.js:5
    d ckeditorerror.js:66
    value ckeditorerror.js:118
    value model.js:236
    value deletecommand.js:76
    decorate observablemixin.js:255
    fire emittermixin.js:209
    e observablemixin.js:259
    value commandcollection.js:68
    value editor.js:288
    value delete.js:55
    fire emittermixin.js:209
    c deleteobserver.js:87
    u deleteobserver.js:50
    fire emittermixin.js:209
    value domeventobserver.js:96
    value keyobserver.js:28
    value domeventobserver.js:79
    fire emittermixin.js:209
[...]

PR #4162, opened 2020-12-02
Debug the CKEditor crashes

ChristophWurst:
DO NOT MERGE

This is for demonstration only, isolating the minimal CKEditor instance that breaks on user input on Firefox. See https://github.com/nextcloud/mail/issues/4104 for more details.

:eyes: @nextcloud/mail


ChristophWurst:
Fix at #4163

Issue #4169, opened 2020-12-02
ThreadEnvelope's reply button doesn't work as expected


StCyr:
### Expected behavior
Pressing the "reply" button of a ThreadEnvelope should create a reply for that specific message.

### Actual behavior
Pressing the "reply" button of a ThreadEnvelope should create a reply for the Envelope (ie: all ThreadEnvelope reply buttons reply to the same message, the message corresponding to the currently selected envelope)

### Mail app

**Mail app version:** master

PR #4170, opened 2020-12-02
Fixes an issue where the reply button of the ThreadEnvelopes would not reply to the correct message

StCyr:
Fixes https://github.com/nextcloud/mail/issues/4169


ChristophWurst (approved):
Oh! Good catch!

PR #4177, opened 2020-12-02
Fix attachment icon and wording

GretaD:

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -161,9 +161,9 @@
 							t('mail', 'Add attachment from Files')
 						}}
 					</ActionButton>
-					<ActionButton :disabled="encrypt" icon="icon-folder" @click="onAddCloudAttachmentLink">
+					<ActionButton :disabled="encrypt" icon="icon-public" @click="onAddCloudAttachmentLink">
 						{{
-							t('mail', 'Add attachment link from Files')
+							addShareLink
 						}}
 					</ActionButton>
 					<ActionCheckbox
@@ -351,6 +351,7 @@ export default {
 				keysMissing: [],
 			},
 			editorMode: 'html',
+			addShareLink: t('mail', 'Add share link from {productName} Files', { productName: OC?.theme?.name ?? 'Nextcloud' }),
 		}
 	},
 	computed: {
@@ -817,13 +818,10 @@ export default {
 	background-position: 12px center;
 	margin-left: 4px;
 }
-</style>
-
-<style lang="scss" scoped>
 ::v-deep .multiselect .multiselect__tags {
 	border: none !important;
 }
 .submit-message.send.primary.icon-confirm-white {
-color: var(--color-main-background);
+	color: var(--color-main-background);
 }
 </style>


src/components/Composer.vue
```diff
@@ -817,13 +817,10 @@ export default {
 	background-position: 12px center;
 	margin-left: 4px;
 }
-</style>
-
-<style lang="scss" scoped>
```
GretaD, line 822: I dont know what happened here and how did we miss it but I didnt do it :person_fencing:

src/components/Composer.vue
```diff
@@ -161,9 +161,9 @@
 							t('mail', 'Add attachment from Files')
 						}}
 					</ActionButton>
-					<ActionButton :disabled="encrypt" icon="icon-folder" @click="onAddCloudAttachmentLink">
+					<ActionButton :disabled="encrypt" icon="icon-public" @click="onAddCloudAttachmentLink">
 						{{
-							t('mail', 'Add attachment link from Files')
+							t('mail', 'Add share link from Nextcloud Files')
```
ChristophWurst, line 166: this doesn't work with theming. could you add a placeholder in the translation string and substitude with ``OC?.theme?.name ?? 'Nextcloud'``? Then it should pick up the theming name and fall back to Nextcloud if nothing is set.

ChristophWurst (changes requested):

ChristophWurst:
cc @benks-io because that was discussed in our design review call :)

src/components/Composer.vue
GretaD: doesnt everyone calls their cloud Nextcloud? :P

benks-io: definitely not ;)

benks-io (changes requested):
@GretaD 
It should be called "Add share link" may be even "Add share link to file from $NAME"....
The Term "share link" is also used in the files app share menu.

GretaD:
> @GretaD
> It should be called "Add share link" may be even "Add share link to file from $NAME"....
> The Term "share link" is also used in the files app share menu.

I copy pasted the one that we agreed during the call. Its on the notes. But Im fine to change it into this.

GretaD:
> > @GretaD
> > It should be called "Add share link" may be even "Add share link to file from $NAME"....
> > The Term "share link" is also used in the files app share menu.
> 
> I copy pasted the one that we agreed during the call. Its on the notes. But Im fine to change it into this.

I take this back because i have accidentally removed the word _share_ when i added the {productName}.

src/components/Composer.vue
```diff
@@ -351,6 +351,7 @@ export default {
 				keysMissing: [],
 			},
 			editorMode: 'html',
+			addShareLink: t('mail', 'Add share link from {productName} Files', { productName: OC.theme.name ?? 'Nextcloud' }),
```
ChristophWurst, line 354:
```suggestion
			addShareLink: t('mail', 'Add share link from {productName} Files', { productName: OC?.theme?.name ?? 'Nextcloud' }),
```

because that will also work if the global variable OC or any of the sub properties don't exit

ChristophWurst (approved):

PR #4206, opened 2020-12-07
Syncs Sents mailbox after sending an email (when it's the mailbox currently open)

### LINKED ISSUE #3598: Refresh sent mailbox after sending an email
**Expected behavior**
After sending a message, it will appear “instantly” in the “Sent” box.

**Actual behavior**
When sending email, it becomes available “instantly” in the “Sent” box, while using RainLoop, Afterlogic, and the Mail Android App version 6.1.22.3 on a smartphone with Android 10, but that isn’t the case using the Mail app.
Using the Mail App you would have to wait for an automatic refresh (like 30 seconds or so) OR you can press the R key, and force the update. Even if I logout of Nextcloud and login again, the mail that I’ve sent still won’t show in the “Sent” box.
After sending an email, it is essential that it become available in the “Sent” Box within a reasonable period, similar to what happens in the “Inbox”.
Is it possible to implement an update time for the “Sent” box similar to what is used in the “Inbox”?
Or maybe you could use the same call that is used when you press the R in the keyboard, and make it run when you change boxes.
Looking forward to your guidance on that matter. Thanks!

**Mail app version**: 1.4.1
**Mail Server or service**: MXRoute (IMAP)
**Server configuration**
**Operating system**: CentOS 7.6
**Web server**: Apache Version 2.4.46
**Database**: MySQL 10.3.24
**PHP version**: 7.3.6
**Nextcloud Version**: 19.0.3
**Client configuration**
**Browser**: Google Chrome Version 84 / 85 (official) 64 bits and Firefox Version 80.0.1 (64-bits)
**Operating system**: Ubuntu 20.04, Mint 20, Debian 10 (buster) and Windows 10 Pro

StCyr:
Fix #3598

Replaces https://github.com/nextcloud/mail/pull/3686

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -632,7 +632,7 @@ export default {
 
 			this.state = STATES.UPLOADING
 
-			return this.attachmentsPromise
+			await this.attachmentsPromise
 				.then(() => (this.state = STATES.SENDING))
 				.then(() => this.draftsPromise)
 				.then(this.getMessageData)
@@ -653,6 +653,18 @@ export default {
 					})
 					this.state = STATES.ERROR
 				})
+
+			// Sync sent mailbox when it's currently open
+			const account = this.$store.getters.getAccount(this.selectedAlias.id)
+			if (parseInt(this.$route.params.mailboxId, 10) === account.sentMailboxId) {
+				setTimeout(() => {
+					this.$store.dispatch('syncEnvelopes', {
+						mailboxId: account.sentMailboxId,
+						query: '',
+						init: false,
+					})
+				}, 500)
+			}
 		},
 		reset() {
 			this.draftsPromise = Promise.resolve() // "resets" draft uid as well


src/components/Composer.vue
```diff
@@ -653,6 +653,18 @@ export default {
 					})
 					this.state = STATES.ERROR
 				})
+
+			// Sync sent mailbox when it's currently open
+			const account = this.$store.getters.getAccount(this.selectedAlias.id)
+			if (this.$route.params.mailboxId === account.sentMailboxId) {
```
ChristophWurst, line 659: isn't ``this.$route.params.mailboxId`` a `string` and the `sentMailboxId` a `Number`? I think we have to use `parseInt(this.$route.params.mailboxId, 10) === account.sentMailboxId`

src/components/Composer.vue
```diff
@@ -632,7 +632,7 @@ export default {
 
 			this.state = STATES.UPLOADING
 
-			return this.attachmentsPromise
+			this.attachmentsPromise
```
ChristophWurst, line 635: I think you want to `await` sending the message. right now you trigger they sync just 500ms after *starting the send request*

src/components/Composer.vue
```diff
@@ -653,6 +653,18 @@ export default {
 					})
 					this.state = STATES.ERROR
 				})
+
+			// Sync sent mailbox when it's currently open
+			const account = this.$store.getters.getAccount(this.selectedAlias.id)
+			if (this.$route.params.mailboxId === account.sentMailboxId) {
+				return setTimeout(() => {
```
ChristophWurst, line 660: don't return the timer. if the method returns anything then it should be a promise

ChristophWurst (changes requested):

StCyr:
@ChristophWurst , All your remarks make sense and have been addressed in my amended commit

ChristophWurst (approved):

Issue #4179, opened 2020-12-03
Do not show "all inboxes" when there's only one account configured


benks-io:
Reopening this Topic again (#3285) as it was newly discussed during last contributor week in Mail design review.

To reduce unnecessary visual load, in the Navigation bar, "all inboxes" should only be visible, if there is more than one account configured.

I don't see why priority inbox would merge together, like mentioned in https://github.com/nextcloud/mail/pull/3329#issuecomment-684911514.

PR #4207, opened 2020-12-07
Remove "All inboxes" if one account is setup

GretaD:
All inboxes should be shown only if we have more than one account set up on our mail app.

fixes: #4179
ref: #3329

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -204,8 +204,8 @@ export default {
 	computed: {
 		visible() {
 			return (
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@TheTimeWalker now it was decided that we want to do this :speak_no_evil:

ChristophWurst (approved):
Code looks good, didn't test

StCyr (approved):
Works good: without the patch I see the "All inboxes", with the patch I don't see it (I've only one account)

Issue #3213, opened 2020-06-08
a11y: Leverage Mail app accessibility


yplassiard:
### Feature Request

Although non critical, it could speed up the way blind and visually impaired users can use this app. With fewchanges within templates, the mail app navigation could change from "somewhat difficult to use" to "smooth and handy".

#### Summary

Using ARIA attributes, it could be very easy to make this app  more accessible.

##### Message list

- Adds a role="heading" aria-level="2" to the div containing the list of messages so that the user can quickly skip to the list.
- For each list item, encapsulate the From, Subject and Date into a role="heading" aria-level="3" to quickly jump to a specific message.
- In the same element, add the "Unread" text before the sender's name if the message is unread.
- Identify the global action buttons (new message, delete, ...) within either a complementary region or add a level 5 heading before the button bar.

##### Message content

- Encapsulate the message's body within a region explicitly named "message body" (to be localized), or prepend the message body with a h3 title.

##### behavior

- when a message is open, setting the focus to its subject(currently marked as "h2" title) could be a good way to save time. Note that you'd need to use the tabindex attribute to make this element focusable for screen reader. If you do not want to explicitly set a tab index order for all elements of the page, a tabindex="0" is fine.
- Make loading and error template appear with a role="alert" so that screen reader can read them when the message is visible. This can also be good to have this role="alert" for any error or information messages that are present on the page, like "message sent" for example.

Should you need any help implementing this, feel free to contact me.

jancborchardt:
Hi @yplassiard, awesome list of improvements, thank you so much! Sorry for the late reply – since you offered your help, is this still something you are interested in doing for example contributing via a pull request?

Also cc @GretaD who could help regarding frontend.

GretaD:
Hello @yplassiard, I started a PR here #4211 with your first 2 points, you can check it out and push changes if you have some time. I will check it out later this week again.

GretaD:
The only point here left to be done is > _the "Unread" text before the sender's name if the message is unread._

PR #4211, opened 2020-12-08
Fix mail app accessibility

GretaD:
ref: #3213

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -44,10 +44,12 @@
 					@click.shift.prevent="onSelectMultiple" />
 			</p>
 		</div>
-		<div class="app-content-list-item-line-one" :title="addresses">
+		<div class="app-content-list-item-line-one"
+			:title="addresses">
 			{{ addresses }}
 		</div>
-		<div class="app-content-list-item-line-two" :title="data.subject">
+		<div class="app-content-list-item-line-two"
+			:title="data.subject">
 			<span v-if="data.flags.answered" class="icon-reply" />
 			<span v-if="data.flags.hasAttachments === true" class="icon-public icon-attachment" />
 			<span v-if="draft" class="draft">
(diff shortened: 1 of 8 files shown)


src/components/Envelope.vue
```diff
@@ -44,18 +44,24 @@
 					@click.shift.prevent="onSelectMultiple" />
 			</p>
 		</div>
-		<div class="app-content-list-item-line-one" :title="addresses">
+		<div class="app-content-list-item-line-one"
+			:title="addresses"
+			role="heading"
+			:aria-label="3">
```
jancborchardt, line 50:
```suggestion
			:aria-level="3">
```

Just a slight error – this needs to be level instead of label – label is the actual description. :)

jancborchardt: Ah and these should also go into the router-link instead, to only have it once instead of 3 times.

ChristophWurst:
@GretaD could you fix the eslint issue? Then I'd say we get this in and start the next item on @yplassiard's list :)

src/components/Message.vue
```diff
@@ -20,9 +20,11 @@
   -->
 
 <template>
-	<div :class="[message.hasHtmlBody ? 'mail-message-body mail-message-body-html' : 'mail-message-body']">
-		<div v-if="message.itineraries.length > 0" class="message-itinerary">
-			<Itinerary :entries="message.itineraries" :message-id="message.messageId" />
+	<div role="region" :aria-label="t('mail','Message body')">
```
jancborchardt, line 23: We don't need an extra div there? The role and aria-label can aimply be added to the existing div. :)

jancborchardt (changes requested):
Seems good, just the 1 enhancement where we don't need the extra div. :)

jancborchardt:
And @yplassiard if you like, check this out and confirm this fixes the issues you reported. Thanks!

jancborchardt (approved):

PR #4212, opened 2020-12-09
forwarding a message from a threadEnvelope's forward menu action doesn't forward the correct message

StCyr:
https://github.com/nextcloud/mail/pull/4170 was incomplete: forwarding a message from a threadEnvelope's forward menu action has the same problem


GretaD (approved):
tested and it works

PR #4235, opened 2020-12-13
Makes links highlighted in the ckeditor

StCyr:
image

Without this change, the link (the first "cyrille") would be black as a regular text


GretaD (approved):
tested and it works just fine

PR #4245, opened 2020-12-14
Only use one AppDetailsToggle button and properly hide it if the navigation is open

### LINKED ISSUE #4209: On small screen app-navigation sidebar doesnt hide
### Expected behavior
When i click **_new message_** and im on small screen, the navigation sidebar should go away.

### Actual behavior
new_message_app

### Mail app

**Mail app version:** latest

juliushaertl:
Should fix #4209 

There was something strange going on with the two instances of the component. This simplifies the code to only use one instance and properly hide the button under the navigation. If the sidebar is closed and the AppDetailsToggle (back arrow) is visible we properly hide the navigation toggle now.

### DIFF
--- a/src/components/AppDetailsToggle.vue
+++ b/src/components/AppDetailsToggle.vue
@@ -47,15 +47,17 @@ export default {
 	line-height: 17px;
 	transform: rotate(180deg);
 	background-color: var(--color-main-background);
-	z-index: 2000;
-&:active,
-&:hover,
-&:focus {
-	 opacity: 1;
- }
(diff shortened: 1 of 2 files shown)


GretaD:
@ChristophWurst i think this should be in for 1.8. This PR fixes 90% of the problem. The only thing left is that when you click _new message_ and you're on the small screen, the `app-navigation` should go away automatically. But, with this PR at least you have the button to remove it manually and nothing is overlapping.

GretaD (approved):

GretaD:
/ backport v1.8

PR #4248, opened 2020-12-15
Fix defaults where mail stores drafts and sent

GretaD:
We missed a setting on the way, while changing Account Settings into Modal.

deafults

### DIFF
--- a/src/components/AccountDefaultsSettings.vue
+++ b/src/components/AccountDefaultsSettings.vue
@@ -20,39 +20,22 @@
   -->
 
 <template>
-	<div class="section">
-		<h2>{{ t('mail', 'Defaults') }}</h2>
-		<p class="settings-hint">
-			{{
-				t('mail', 'Here you can select where Nextcloud Mail stores your drafts as well as sent and deleted messages.')
-			}}
+	<div>
+		<p>
+			{{ t('mail', 'Drafts are saved in:') }}
 		</p>
-		<table>
-			<tr>
-				<td>
-					{{ t('mail', 'Draft messages are saved to:') }}
-				</td>
-				<td>
-					<MailboxInlinePicker v-model="draftsMailbox" :account="account" :disabled="saving" />
-				</td>
-			</tr>
-			<tr>
-				<td>
-					{{ t('mail', 'Sent messages are saved to:') }}
-				</td>
-				<td>
-					<MailboxInlinePicker v-model="sentMailbox" :account="account" :disabled="saving" />
-				</td>
-			</tr>
-			<tr>
-				<td>
-					{{ t('mail', 'Deleted messages are moved to:') }}
-				</td>
-				<td>
-					<MailboxInlinePicker v-model="trashMailbox" :account="account" :disabled="saving" />
-				</td>
-			</tr>
-		</table>
+		<MailboxInlinePicker v-model="draftsMailbox" :account="account" :disabled="saving" />
+
+		<p>
+			{{ t('mail', 'Sent messages are saved in:') }}
+		</p>
+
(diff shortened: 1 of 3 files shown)


GretaD:
@jancborchardt on settings its also the defaults, which I missed it while working on it. The feature was done by Christoph while i was working on the modal, and we were suppose to style it when the modal thing was done. And here we are :)

ChristophWurst (approved):

src/components/AccountDefaultsSettings.vue
```diff
@@ -20,40 +20,28 @@
...
-				<td>
-					<MailboxInlinePicker v-model="trashMailbox" :account="account" :disabled="saving" />
-				</td>
-			</tr>
-		</table>
-	</div>
+	<table>
+		<tr>
+			<td>
+				{{ t('mail', 'Draft messages are saved to:') }}
```
jancborchardt, line 26:
```suggestion
				{{ t('mail', 'Drafts are saved in:') }}
```

src/components/AccountDefaultsSettings.vue
```diff
@@ -20,40 +20,28 @@
...
+				<MailboxInlinePicker v-model="draftsMailbox" :account="account" :disabled="saving" />
+			</td>
+		</tr>
+		<tr>
+			<td>
+				{{ t('mail', 'Sent messages are saved to:') }}
```
jancborchardt, line 32:
```suggestion
				{{ t('mail', 'Sent messages are saved in:') }}
```

src/components/AccountSettings.vue
```diff
@@ -47,7 +47,14 @@
 			</p>
 			<EditorSettings :account="account" />
 		</AppSettingsSection>
-
+		<AppSettingsSection :title=" t('mail', 'Defaults')">
```
jancborchardt, line 50:
```suggestion
		<AppSettingsSection :title=" t('mail', 'Default folders')">
```

src/components/AccountSettings.vue
```diff
@@ -47,7 +47,14 @@
...
+			<p class="settings-hint">
+				{{
+					t('mail', 'Here you can select where Nextcloud Mail stores your drafts as well as sent and deleted messages.')
```
jancborchardt, line 53:
```suggestion
					t('mail', 'The folders to use for drafts, sent messages and deleted messages.')
```

src/components/AccountDefaultsSettings.vue
```diff
@@ -20,40 +20,28 @@
...
-				<td>
-					{{ t('mail', 'Deleted messages are moved to:') }}
-				</td>
-				<td>
-					<MailboxInlinePicker v-model="trashMailbox" :account="account" :disabled="saving" />
-				</td>
-			</tr>
-		</table>
-	</div>
+	<table>
```
jancborchardt, line 23: Since we are now in a narrower modal with the dropdown below the text, instead of the whole table structure, a simple `<p>` for the text will do (followed by the MailboxInlinePicker). That way there’s also a bit of vertical spacing between the picker and the following text.

jancborchardt (dismissed):
Some wording suggestions and HTML structure comment. And the dropdowns look a bit off, too high, leading to the text not vertically aligned? (Also compare with the mail server settings inputs further down.)

GretaD:
@jancborchardt better? 👇 
befaultsss

jancborchardt:
Bit of vertical spacing between the sets of text+input would be nice but it’s good to go like this too. :) :+1:

GretaD:
> Bit of vertical spacing between the sets of text+input would be nice but it’s good to go like this too. :) +1

great, i will fix it with a follow up and fix also the drop down which can be shown better.

jancborchardt (approved):

PR #4250, opened 2020-12-16
Add undo option for trusted senders

### LINKED ISSUE #4176: Add the undo option for the "Remember images for this sender"
### Feature Request

Right now, we have the option to remember a sender as trusted and all the messages coming from them and have images, the images will be shown. We should be able to undo this option when we dont want to see the images anymore.

#### Summary

Ref: #4103

GretaD:
fixes #4176

### DIFF
--- a/lib/Controller/TrustedSendersController.php
+++ b/lib/Controller/TrustedSendersController.php
@@ -83,4 +83,18 @@ class TrustedSendersController extends Controller {
 
 		return JsonResponse::success(null);
 	}
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 *
+	 * @return JsonResponse
+	 */
+	public function list(): JsonResponse {
+		$list = $this->trustedSenderService->getTrusted(
+			$this->uid
+		);
+
+		return JsonResponse::success($list);
+	}
 }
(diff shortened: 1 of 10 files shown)


GretaD:
@jancborchardt right now, the list looks like this. Has no styling so far. Please let me know how do you think it should look.

jancborchardt:
Only some improvements:
- We always only capitalize the first word of a heading, so "Trusted senders"
- The "List" text can be removed, and instead we should have a settings-hint which says: "Images in mails from these senders will be loaded"
- The button should be called "Remove" instead of "Hide"

ChristophWurst (approved):

lib/Controller/TrustedSendersController.php
```diff
@@ -83,4 +83,18 @@ public function removeTrust(string $email): JsonResponse {
...
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 *
+	 * @return JsonResponse
+	 */
+	public function list(): JsonResponse {
+		$list =	$this->trustedSenderService->getTrusted(
+			$this->uid,
```
ChristophWurst, line 95:
```suggestion
			$this->uid
```

trailing commas are only allowed in php7.3+

Issue #3060, opened 2020-05-05
Forward as attachment


ChristophWurst:
### Feature Request

We currently only support cited forwards, but it would be helpful to also allow the original message to be attached.

#### Summary

n/a

bernatnan:
Hi, with this "feature", the attached message will not lost the "links"?

Now, when a mail is forwarded, it lost the "html" (links and others)

ChristophWurst:
yes, the attachment will have the unmodified original message.

bernatnan:
Perfect! Will be a great improve. If you need any test, tell me. Thanks!

ChristophWurst:
WIP PR is at https://github.com/nextcloud/mail/pull/4251

bernatnan:
Hi @ChristophWurst , if I haven't understood wrong, this "issue" has been added int he new release 1.8.0, but I don't see it. I'm wrong?

ChristophWurst:
Please use https://help.nextcloud.com/c/apps/mail/35 for questions. You find the forward option in the multiselect menu.

PR #4251, opened 2020-12-16
Add forward as attachment

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3060

This adds a "Forward X as attachment" action to the multiselect.

Note that our client can't show embedded messages. We can send them, but you'll need another client to verify that the attached embedded message is correct.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -366,9 +366,9 @@ class MessageMapper {
 	 * @return string|null
 	 * @throws ServiceException
 	 */
-	public function getSource(Horde_Imap_Client_Socket $client,
-							  string $mailbox,
-							  int $uid): ?string {
+	public function getFullText(Horde_Imap_Client_Socket $client,
+								string $mailbox,
+								int $uid): ?string {
 		$query = new Horde_Imap_Client_Fetch_Query();
 		$query->uid();
 		$query->fullText([
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -601,7 +601,7 @@ class IMAPMessage implements IMessage, JsonSerializable {
 	 *
 	 * @return void
 	 */
-	public function addForwardedAttachment(string $name, string $content): void {
+	public function addRawAttachment(string $name, string $content): void {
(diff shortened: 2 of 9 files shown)


StCyr:
Works good.

Though, I have issues with the UX. Why putting this action in the 'multiselect toolbar' only?

To be more generic, I start to be confused between the various menus, the menu for the threadEnvelopes, the menu for the Envelopes, the menu in the 'multiselect toolbar'. They all present *almost* the same actions and I sometime have to open 2 menus before finding the one that offers the action I'm looking for (for example, I'm missing the 'forward' action in the Envelope menu). Do you share this feeling?

szaimen:
Sry for intervening, but I would like to have this option as a default when forwarding mails (which would solve the too much options problem) since normal forwarding breaks any html formatting which is not acceptable, imo.
Maybe this could be a follow up?

ChristophWurst:
> Though, I have issues with the UX. Why putting this action in the 'multiselect toolbar' only?

Did you see my gap analysis in the dev chat? I'm fine adding it somewhere else as well, but we have to be careful with the many actions we already have. The list is getting longer and longer and I'd argue that forward of attachment is rather a niche feature.



> To be more generic, I start to be confused between the various menus, the menu for the threadEnvelopes, the menu for the Envelopes, the menu in the 'multiselect toolbar'. They all present _almost_ the same actions and I sometime have to open 2 menus before finding the one that offers the action I'm looking for (for example, I'm missing the 'forward' action in the Envelope menu). Do you share this feeling?

Those two should be identical indeed. The only difference is that *Select* is only available in the envelope list, show source makes most sense for the message detail. The rest should be uniform.

ChristophWurst:
> Sry for intervening, but I would like to have this option as a default when forwarding mails (which would solve the too much options problem) since normal forwarding breaks any html formatting which is not acceptable, imo.
> Maybe this could be a follow up?

We can discuss making this a configurable option of the user or email account. Could you open a feature request please?

StCyr (approved):
Works good. 

New issue created for streamlining the menus

jancborchardt (approved):
Nothing further to add to your comment @ChristophWurst, sounds good! :)

PR #4257, opened 2020-12-18
Fix Webpack issue with CKEditor's dynamic translation imports

ChristophWurst:
When building master there was a webpack warning that an import uses an expression. Webpack support this syntax, but not with all module systems. The problem was that our Babel was configured to output *commonjs* modules. Those can't do this. Hence the mismatch.

Now here you can build without that warning. The bundle is also split into chunks again (there was only large bundles on master of several MB). And when you switch your language to non-English, the CKEditor will be translated (assuming there is a translation for that language, they might not cover the same languages as we do).

Others apps' Babel config for comparison:
https://github.com/nextcloud/contacts/blob/d5a1314d8c0a72bdbc433e680e61c012ce85d0c5/babel.config.js#L7-L11
https://github.com/nextcloud/text/blob/5b9d41ed10ac41cf587ea562a94519f9c173be03/babel.config.js#L20-L25


StCyr (commented):
This build properly but I don't know how to test: "when you switch your language to non-English" <= how do you do that? Firefox setting?

Also, there's still an error related to this asset:

```
[CKEditorWebpackPlugin] Error: Too many JS assets has been found during the compilation. You should use one of the following options to specify the strategy:
- use `addMainLanguageTranslationsToAllAssets` to add translations for the main language to all assets,- use `buildAllTranslationsToSeparateFiles` to add translation files via `<script>` tags in HTML file,- use `translationsOutputFile` to append translation to the existing file or create a new asset.For more details visit https://github.com/ckeditor/ckeditor5-dev/tree/master/packages/ckeditor5-dev-webpack-plugin.
```

ChristophWurst:
Change your Nextcloud to French in the personal settings.

That CKEditor warnin needs to be addressed as well, but it's a different problem.

StCyr:
yes I'd figured it out :-)

StCyr (approved):

Issue #2412, opened 2019-12-23
Can not create account  Request failed with status code 500


kalle-ankan:
Hello

When I try to create an account so do I get this error: Request failed with status code 500 

I use the latest nextcloud and mail. It has been like this for an long time though a year or so..

I noticed in the horde_imap.log:

C: 12 STATUS procmail.log (MESSAGES UIDNEXT UIDVALIDITY)
S: 12 NO [CANNOT] Mailbox isn't a valid mbox file (0.001 + 0.000 secs).
>> Command 12 took 0.048 seconds.
>> ERROR: mail server explicitly reporting an error.
C: 13 LOGOUT
S: * BYE Logging out
S: 13 OK Logout completed (0.001 + 0.000 secs).

The server that I try to connect to uses Dovecot: dovecot-imapd  1:2.3.4.1-5+deb10u1

The same mail server works well with thunderbird and roundcube webmail.

Any idea how to debug it more or something to try?

ChristophWurst:
It tries to open procmail.log. What is this mailbox?

kalle-ankan:
On 1/7/20 2:21 PM, Christoph Wurst wrote:
>
> It tries to open procmail.log. What is this mailbox?
>

Hi

Procmail was installed and configured to do  server side filtering on 
incoming email.

When i removed procmail from the server and disabled server side 
filtering on incoming email, then it started to work.

I do not know why nexcloud mail reads procmail.log since it was not an 
subscribed mailbox.

ChristophWurst:
> I do not know why nexcloud mail reads procmail.log since it was not an subscribed mailbox.

Hm. We currently list all mailboxes. We don't take the subscripted state into consideration. But IMO the IMAP server should set the `\noselect` attribute on that mailbox if clients are not allowed to access it.

stweil:
@kalle-ankan, maybe you can try the following patch which worked for our instance:

    diff --git a/lib/IMAP/FolderMapper.php b/lib/IMAP/FolderMapper.php
    index 4f677f85..6f5b7656 100644
    --- a/lib/IMAP/FolderMapper.php
    +++ b/lib/IMAP/FolderMapper.php
    @@ -68,6 +68,11 @@ class FolderMapper {
                                    return null;
                            }
     
    +                       if (strpos($mailbox['mailbox']->utf8, 'shared.') === 0) {
    +                               // Workaround for shared Horde folders.
    +                               return null;
    +                       }
    +
                            return new Folder(
[...]

PR #4318, opened 2021-01-07
ignore folders which cause errors on access

jmechnich:
This is a proposed fix for the problems with inaccessible folders as discussed in #2412

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -62,10 +62,22 @@ class FolderMapper {
 			'special_use' => true,
 		]);
 
-		return array_filter(array_map(function (array $mailbox) use ($account) {
+		return array_filter(array_map(function (array $mailbox) use ($account, $client) {
 			if (in_array($mailbox['mailbox']->utf8, self::DOVECOT_SIEVE_FOLDERS, true)) {
 				// This is a special folder that must not be shown
 				return null;
+			}
+
+			try {
+				$client->status($mailbox["mailbox"]);
+			} catch (Horde_Imap_Client_Exception $e) {
+				// ignore folders which cause errors on access
+				// (i.e. server-side system I/O errors)
+				if (in_array($e->getCode(), [
+					Horde_Imap_Client_Exception::UNSPECIFIED,
+				], true)) {
+					return null;
+				}
 			}
 
 			return new Folder(


stweil (approved):

lib/IMAP/FolderMapper.php
```diff
@@ -62,12 +62,19 @@ public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
...
 			if (in_array($mailbox['mailbox']->utf8, self::DOVECOT_SIEVE_FOLDERS, true)) {
 				// This is a special folder that must not be shown
 				return null;
 			}
 
+			// ignore folders which cause errors on access (i.e. server-side SYSTEM I/O ERROR)
+			try {
+				$client->status($mailbox["mailbox"]);
+			} catch (Horde_Imap_Client_Exception $e) {
+				return null;
```
ChristophWurst, line 75: so this will also effectively make any IMAP hiccups hide the mailbox until the next mailbox sync

can we catch the specific error code? see the constants on ``Horde_Imap_Client_Exception``, we can compare those to $e->getCode().

ChristophWurst (changes requested):
Catches too much IMO, like also timeout errors or wrong password errors

lib/IMAP/FolderMapper.php
jmechnich: Thanks Christoph, I completely agree...just wasn't sure how to get the error code properly (https://dev.horde.org/api/master/lib/Imap_Client/classes/Horde_Imap_Client_Exception.html only mentions the setter-Method). Let me check and update the request.

ChristophWurst: >  (https://dev.horde.org/api/master/lib/Imap_Client/classes/Horde_Imap_Client_Exception.html only mentions the setter-Method)

I think that's because the `getCode` isn't overwritten, they just pass the code down into php's `\Exception` base class

jmechnich: The change is in (conveniently enough the relevant error in our case is UNSPECIFIED).

ChristophWurst: Well, I guess that is OK. The more common errors like timeout or authentication will have a distinct error code.

lib/IMAP/FolderMapper.php
```diff
@@ -62,12 +62,24 @@ public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
...
 			}
 
+			try {
+				$client->status($mailbox["mailbox"]);
+			} catch (Horde_Imap_Client_Exception $e) {
+				// ignore folders which cause errors on access
+				// (i.e. server-side system I/O errors)
+				if (in_array($e->getCode(), [
+					Horde_Imap_Client_Exception::UNSPECIFIED,
+				])) {
```
ChristophWurst, line 78: please pass `true` as third param of `in_array`. It's unlikely that php would do weird casting with this particular code but static analysis will thank us for this.

Ref https://www.php.net/manual/en/function.in-array.php

jmechnich: Done.

ChristophWurst (approved):

Issue #1467, opened 2019-01-14
read confirmation & acknowledgment of receipt


Zayccmar:
### Feature Request

Hello i use the mail app by my nextcloud enviorment, and i must say that is a wonderfull app but what i missing are a funktion to set a  read confirmation & or a  acknowledgment of receipt when i type an new Mail or i get a reply to a mail

it is possible to integrate this fetures by the next release?

best regrads 

Markus

ChristophWurst:
> it is possible to integrate this fetures by the next release?

Sure. Pull request are welcome anytime :v:

ChristophWurst:
You can find more information about app development in the manual: https://docs.nextcloud.com/server/15/developer_manual/

Cheers

Zayccmar:
> > it is possible to integrate this fetures by the next release?
> 
> Sure. Pull request are welcome anytime ✌️

Hi thank you
can you help me i don´t have do a pull request before and this are my first visit at github as a member

best regrads 

Markus

ChristophWurst:
You can find great GitHub docs at https://help.github.com/.

Zayccmar:
> You can find great GitHub docs at https://help.github.com/.

Thank you

but when i understand it right the pull request are for developer, but i am not a developer and so i can not make a example for the right source code, by this i have made a issue with a feature request.

so now i don´t know what i must do.

best regrads

ChristophWurst:
I you're not a developer than the chances of this feature being added to the next release are quite low. Maybe you know a friend who is a coder than they can help :)

You can try to add a bounty at https://www.bountysource.com/issues/68667195-read-confirmation-acknowledgment-of-receipt but there is no guarantee that anyone will have a look. But it sometimes helps.

ChristophWurst:
PR is at https://github.com/nextcloud/mail/pull/4343 :rocket:

PR #4343, opened 2021-01-12
Message delivery notification

kesselb:
Fix https://github.com/nextcloud/mail/issues/1467

**Compose a new message**

image

**Message with mdn request**

image

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -34,6 +34,7 @@ use Exception;
 use OC\Security\CSP\ContentSecurityPolicyNonceManager;
 use OCA\Mail\Contracts\IMailManager;
 use OCA\Mail\Contracts\IMailSearch;
+use OCA\Mail\Contracts\IMailTransmission;
 use OCA\Mail\Contracts\ITrustedSenderService;
 use OCA\Mail\Db\Message;
 use OCA\Mail\Exception\ClientException;
@@ -96,6 +97,9 @@ class MessagesController extends Controller {
 	/** @var ITrustedSenderService */
 	private $trustedSenderService;
 
+	/** @var IMailTransmission */
+	private $mailTransmission;
+
 	/**
 	 * @param string $appName
 	 * @param IRequest $request
@@ -111,6 +115,7 @@ class MessagesController extends Controller {
 	 * @param IURLGenerator $urlGenerator
 	 * @param ContentSecurityPolicyNonceManager $nonceManager
 	 * @param ITrustedSenderService $trustedSenderService
+	 * @param IMailTransmission $mailTransmission
 	 */
 	public function __construct(string $appName,
 								IRequest $request,
@@ -125,7 +130,8 @@ class MessagesController extends Controller {
 								IMimeTypeDetector $mimeTypeDetector,
 								IURLGenerator $urlGenerator,
 								ContentSecurityPolicyNonceManager $nonceManager,
-								ITrustedSenderService $trustedSenderService) {
+								ITrustedSenderService $trustedSenderService,
+								IMailTransmission $mailTransmission) {
 		parent::__construct($appName, $request);
 
 		$this->accountService = $accountService;
@@ -141,6 +147,7 @@ class MessagesController extends Controller {
 		$this->mailManager = $mailManager;
 		$this->nonceManager = $nonceManager;
 		$this->trustedSenderService = $trustedSenderService;
+		$this->mailTransmission = $mailTransmission;
 	}
 
 	/**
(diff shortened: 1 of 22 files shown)


lib/Controller/MessagesController.php
```diff
@@ -336,6 +343,43 @@ public function move(int $id, int $destFolderId): JSONResponse {
...
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $id);
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+//		 @Todo: Enable later again
+//		if ($message->getFlagMdnsent()) {
```
ChristophWurst, line 367: :+1:

lib/Controller/MessagesController.php
```diff
@@ -336,6 +343,43 @@ public function move(int $id, int $destFolderId): JSONResponse {
...
+//			return new JSONResponse([], Http::STATUS_PRECONDITION_FAILED);
+//		}
+
+		try {
+			$this->mailTransmission->sendMdn($account, $mailbox, $message);
+			$this->mailManager->flagMessage($account, $mailbox->getName(), $message->getUid(), 'mdnsent', true);
+		} catch (ServiceException $ex) {
+			// @Todo: Make sure to throw a service exception
```
ChristophWurst, line 375: indeed. `sendMdn` should have a documented exception

lib/Db/MessageMapper.php
```diff
@@ -560,6 +564,10 @@ public function findIdsByQuery(Mailbox $mailbox, SearchQuery $query, ?int $limit
 			}
 		}
 
+		if (isset($flags[Horde_Imap_Client::FLAG_MDNSENT])) {
+			$select->andWhere($qb->expr()->eq('flag_mdnsent', $qb->createNamedParameter('flag_mdnsent', IQueryBuilder::PARAM_BOOL)));
```
ChristophWurst, line 568: can we add this flag to the array above at L550-558?

ChristophWurst: If it's just about the leading `$` then we can maybe add `'mdnsent'`, right?

lib/Model/IMAPMessage.php
```diff
@@ -327,6 +338,7 @@ private function loadMessageBodies(): void {
 		$fetch_query->flags();
 		$fetch_query->size();
 		$fetch_query->imapDate();
+		$fetch_query->headers('mdn', ['disposition-notification-to']); // @Todo: peek = true, cache = true?
```
ChristophWurst, line 341: yes, at least the peek one ;)

lib/Model/NewMessageData.php
```diff
@@ -163,4 +171,8 @@ public function getAttachments(): array {
 	public function isHtml(): bool {
 		return $this->isHtml;
 	}
+
+	public function requestMdn(): bool {
```
ChristophWurst, line 175: isMdnRequested? sounds more like a predicate

src/components/MdnRequest.vue
```diff
@@ -0,0 +1,47 @@
...
+			type: Object,
+		},
+	},
+	data() {
+		return {
+			hasMdnRequest: false,
+		}
+	},
+	mounted() {
+		this.hasMdnRequest = !this.message.flags.mdnsent && this.message.dispositionNotificationTo && this.message.dispositionNotificationTo.length > 0
```
ChristophWurst, line 30: any reason this computation isn't done in `data`? You should have access to the prop there

src/components/MdnRequest.vue
```diff
@@ -0,0 +1,47 @@
...
+	},
+	methods: {
+		async sendMdn() {
+			logger.debug('send return receipt')
+
+			await sendMdn(this.message.databaseId)
```
ChristophWurst, line 36: wrap with try-catch and use `showError` from `@nextcloud/dialogs` to show a toast on error

ChristophWurst (changes requested):

src/components/MdnRequest.vue
```diff
@@ -1,11 +1,18 @@
...
-		}}
-		<button @click="sendMdn">
-			{{ t('mail', 'Notify the sender') }}
-		</button>
+	<div v-if="hasMdnRequest || mdnSent" id="mail-message-has-mdn-request">
+		<span v-if="hasMdnRequest">
+			{{
+				t('mail', 'The sender of this message has asked to be notified when you read this message.')
+			}}
+			<button @click="sendMdn">
```
ChristophWurst, line 7: it might be good if we disable the button as long as the async request is awaited in the background

lib/Db/MessageMapper.php
kesselb: https://github.com/nextcloud/mail/pull/4343/commits/f2a19f96caeb79e8551014febb659efb1b8e7940

src/components/MdnRequest.vue
kesselb: yep

ChristophWurst:
Works nicely!

Some smaller things

- [ ] Let's update the envelope/message's flag on the client-side immediately. If you navigate to another message right now after sending the confirmation and than back to that message you see the button again, simply because the flag isn't updated yet by the background sync.

src/components/MdnRequest.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+	<div v-if="hasMdnRequest || mdnSent" id="mail-message-has-mdn-request">
+		<span v-if="hasMdnRequest">
```
ChristophWurst, line 3:
```suggestion
		<span v-if="hasMdnRequest && !mdnSent">
```

src/components/MdnRequest.vue
```diff
@@ -0,0 +1,63 @@
...
+			{{
+				t('mail', 'The sender of this message has asked to be notified when you read this message.')
+			}}
+			<button :disabled="loading" @click="sendMdn">
+				{{ t('mail', 'Notify the sender') }}
+			</button>
+		</span>
+		<span v-else>
```
ChristophWurst, line 11:
```suggestion
		<span v-else-if="hasMdnRequest && mdnSent">
```

ChristophWurst (commented):
Otherwise you will always see the button after a page refresh or navigating away any back :wink:

src/components/MdnRequest.vue
```diff
@@ -0,0 +1,61 @@
...
+	name: 'MdnRequest',
+	props: {
+		message: {
+			required: true,
+			type: Object,
+		},
+	},
+	data() {
+		return {
+			mdnSent: this.message.flags.mdnsent,
```
ChristophWurst, line 32: oh and this should be moved into a *computed* property so it becomes reactive and updates when the flag changes

src/components/MessagePlainTextBody.vue
```diff
@@ -1,6 +1,7 @@
 <template>
-	<div>
-		<div id="mail-content" v-html="nl2br(enhancedBody)" />
+	<div id="mail-content">
+		<MdnRequest :message="this.message" />
```
ChristophWurst, line 3:
```suggestion
		<MdnRequest :message="message" />
```

no `this` in templates :)

ChristophWurst (changes requested):
Please increase the version to beta-3 in info.xml

ChristophWurst (approved):
Code looks good, didn't give this the final test but will do so soon!

GretaD (approved):
Tested and works great.

PR #4353, opened 2021-01-13
Fix dark theme on small screen

GretaD:
When resizing the screen while on dark theme, the dark color turns white in some screen sizes. This looks like fixes the problem. Please test it to see if i missed something.


ChristophWurst (approved):
Whohoo

Issue #4256, opened 2020-12-18
Streamline action menus


StCyr:
### Feature Request

Streamline action menus

#### Summary

Action menus for the Envelope and for the ThreadEnvelopee components contains *almost* the same actions that make them a bit confusing: One sometimes looks first in a menu that has not the action.

![image](https://user-images.githubusercontent.com/8179031/102630080-60a9e900-414c-11eb-8b6d-3f2984e5eda2.png)

It would bit good if both menus were streamlined a bit. The proposal from @ChristophWurst seems good: both menus should contains all same actions except: (1) the Envelope menu shouldn't show the "show message source" action, and (2) the ThreadEnvelope menu shouldn't show the "Select" action.

The multiselect menu should be treated differently: No need for more streamlining for it atm

PR #4359, opened 2021-01-15
Streamline envelope action menus

StCyr:
Fixes https://github.com/nextcloud/mail/issues/4256

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -120,7 +120,8 @@ export default {
 	},
 	computed: {
 		accountColor() {
-			return calculateAccountColor(this.account?.emailAddress ?? '')
+			const account = this.$store.getters.getAccount(this.data.accountId)
+			return calculateAccountColor(account?.emailAddress ?? '')
 		},
 		draft() {
 			return this.data.flags.draft
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -60,103 +60,32 @@
 		<div class="app-content-list-item-details date">
 			<Moment :timestamp="data.dateInt" />
 		</div>
-		<Actions class="app-content-list-item-menu"
-			menu-align="right"
-			event=""
-			@click.native.prevent>
-			<ActionButton icon="icon-important"
-				:close-after-click="true"
-				@click.prevent="onToggleImportant">
-				{{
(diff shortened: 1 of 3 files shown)


src/components/Envelope.vue
```diff
@@ -58,103 +58,32 @@
...
-		<MoveModal v-if="showMoveModal"
-			:account="account"
-			:envelopes="[data]"
-			@close="onCloseMoveModal" />
+		<MenuEnvelope
+			:envelope=data
+			:mailbox=mailbox
+			:isSelected=selected
+			:can-select=true
+			:with-show-source=false
```
ChristophWurst, line 66: can-show-source?

src/components/MenuEnvelope.vue
```diff
@@ -0,0 +1,321 @@
+<!-- Standard Actions menu for Envelopes -->
+<template>
+	<div class="app-content-list-item-menu">
```
ChristophWurst, line 3: I'm pretty sure this class should only be used in the *app content list* (envelope list) and no in the details on the right

src/components/MenuEnvelope.vue
```diff
@@ -0,0 +1,321 @@
...
+			this.showMoveModal = true
+		},
+		onCloseMoveModal() {
+			this.showMoveModal = false
+		},
+	},
+}
+</script>
+
+<style>
```
ChristophWurst, line 308: lang=scss?

ChristophWurst (commented):
Great move!

some minor comments. also eslint is your friend, let is fix some syntax issues :)

src/components/Envelope.vue
StCyr: with-select 😉

GretaD:
All looks pretty good. The only thing i noticed that was shown wrong, is the account color. I have 2 accounts, but the colour at the messaged(second column) is always pink. On master is all good.

rect825

StCyr:
> 
> All looks pretty good. The only thing i noticed that was shown wrong, is the account color. I have 2 accounts, but the colour at the messages(second column) is always pink. On master is all good.
> 

Hey! Good catch!

It's fixed now

GretaD (approved):
all looks good. Tested and it works

Issue #1742, opened 2019-05-10
Font color in Dark Theme Mode


menaceone:
As stated in #1363 I open a seperate Issue for the font behaviour in Dark Theme mode:

### Expected behavior
Having Light Grey Font Color in Mails when Dark Theme is activated.

### Actual behavior
Black Font Color on Black Background.
![Black Font Color](https://user-images.githubusercontent.com/1039753/49704650-ed447700-fc15-11e8-98ba-4b047932f4ec.png)

### Mail app

**Mail app version:** 0.14.0

**Mailserver or service:** independent from server/service

### Server configuration

**Operating system**: Ubuntu 18.04

**Web server:** Apache

**Database:** MySql

**PHP version:** 7.3.5

**Nextcloud Version:** 16.0.0.9

#### Client configuration
**Browser:** Firefox 66

**Operating system:** Windows

ChristophWurst:
I think this is a task for our *Head of Dark Mode @skjnldsv*

skjnldsv:
I guess it can be overridden by the html in the mail itself? Otherwise using the css variable works as-is :)

menaceone:
You mean, every Mail declares it's own font color?

skjnldsv:
@menaceone it's possible yes

tanghus:
Another example and a suggestion.

Here's a mail, where the html part doesn't declare a background color:

![nextcloud-mail-html](https://user-images.githubusercontent.com/533222/64694915-2ce90980-d49b-11e9-8a98-b007c3e9167f.png)

Most parties (spam/your friends/authorities/etc) sending html mail, assume they will always have a light canvas to paint on.

Suggestion:
- If pure text/plain show light on dark
- If it's multipart html and plain:
    - Select either (based on settings or a switch in the reader, like with loading images?).
        - If plain, show light on dark.
        - If html, show html on white background. The html part can override it.

rigrig:
(copy of https://github.com/nextcloud/mail/issues/2482#issuecomment-597758859 because that issue was closed as duplicate)

Maybe you could use
```css
#message-container {
    filter: invert(100%);
}
```
To simply invert message colors completely?
It might need some fiddling to double-invert images (and maybe other things?)
(And people replying with "my comments in red" will become even more of an annoyance)

Actual screenshots: (inverted by adjusting CSS through Firefox developer console)
| Regular  | Dark theme | Dark theme Inverted |
| ---------- | ---------------- | --------------------------- |
[...]

PR #4376, opened 2021-01-19
Fix the dark theme text color

GretaD:
Fixes https://github.com/nextcloud/mail/issues/1742

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -58,20 +58,7 @@ export default {
 		scout.on('afterprint', this.onAfterPrint)
 	},
 	mounted() {
-		iframeResizer({
-			onInit: () => {
-				const getCssVar = (key) => ({
-					[key]: getComputedStyle(document.documentElement).getPropertyValue(key),
-				})
-
-				// send css vars to client page
-				this.$refs.iframe.iFrameResizer.sendMessage({
-					cssVars: {
-						...getCssVar('--color-main-background'),
-					},
-				})
-			},
-		}, this.$refs.iframe)
+		iframeResizer({}, this.$refs.iframe)
 	},
 	beforeDestroy() {
 		scout.off('beforeprint', this.onBeforePrint)
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -1,7 +1,7 @@
 <template>
 	<div id="mail-content">
 		<MdnRequest :message="message" />
-		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content">
+		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content" style="color: #000000">
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
 			<button @click="onShowBlockedContent">
 				{{ t('mail', 'Show images from this sender') }}
@@ -67,7 +67,7 @@ export default {
 				// send css vars to client page
 				this.$refs.iframe.iFrameResizer.sendMessage({
 					cssVars: {
-						...getCssVar('--color-main-text'),
+						...getCssVar('--color-main-background'),
 					},
 				})
 			},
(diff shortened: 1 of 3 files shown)


mwalbeck:
I have tried out the fix and there is a small issue with the text colour of "#mail-message-has-blocked-content". "--color-primary-text" is set to white for both the dark and light themes which results in light text on a light background.

Screenshot_20210209_162832

Here it is with a dark text colour

Screenshot_20210209_162855

Other than that, it works as intended.

GretaD:
@mwalbeck thank you for testing. But im not sure if i understood what do you mean. On both themes the "block contect string" it is readable. Do you mean it should be lighter?

mwalbeck:
For me the text colour of the string "The images have been blocked to protect your privacy." is (```#ffffff```) and is therefore not visible at all.

I did a bit of digging in the server code and I may have found out why. The CSS variable ```--color-primary-text``` has a default value of ```#ffffff``` but it is dependent on the colour value set in the Theming app. If a colour that has a good contrast with white is chosen, like the default blue colour, then ```--color-primary-text``` is set to ```#ffffff```. But if a colour that has a bad contrast with white is chosen, for example green, then ```--color-primary-text``` is set to ```#000000```.

So if I set the colour in the Theming app to something like ```#5FC917``` then this fix works as intended and the string mentioning the blocked images is shown with text colour: ```#000000```. But with the default blue colour the string is shown with text colour: ```#ffffff```.

I think a hardcoded colour value may be best here as we want it to always be dark since it will be shown on a light background no matter what theme is being used.

GretaD:
> For me the text colour of the string "The images have been blocked to protect your privacy." is (`#ffffff`) and is therefore not visible at all.
> 
> I did a bit of digging in the server code and I may have found out why. The CSS variable `--color-primary-text` has a default value of `#ffffff` but it is dependent on the colour value set in the Theming app. If a colour that has a good contrast with white is chosen, like the default blue colour, then `--color-primary-text` is set to `#ffffff`. But if a colour that has a bad contrast with white is chosen, for example green, then `--color-primary-text` is set to `#000000`.
> 
> So if I set the colour in the Theming app to something like `#5FC917` then this fix works as intended and the string mentioning the blocked images is shown with text colour: `#000000`. But with the default blue colour the string is shown with text colour: `#ffffff`.
> 
> I think a hardcoded colour value may be best here as we want it to always be dark since it will be shown on a light background no matter what theme is being used.

ok, i dont have any message to test your "edge case" but you are right on your findings. With `--color-primary-text` all my html messages with blocked content were shown ok in both themes. But i changed it as you suggested. Please have a look.

mwalbeck:
The latest commit works as intended on my end. Cheers!

src/components/MessageHTMLBody.vue
```diff
@@ -67,7 +67,7 @@ export default {
 				// send css vars to client page
 				this.$refs.iframe.iFrameResizer.sendMessage({
 					cssVars: {
-						...getCssVar('--color-main-text'),
+						...getCssVar('--color-main-background'),
```
st3iny, line 70: Just sending another css variable to the client page won't change anything on its own. If we scrap the injection of the text color and just force a white background, this whole block can be removed (including the `getCssVar()` method).

See the injected css for reference (`css/html-response.css`):
```css
* {
	font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', Arial, 'Noto Color Emoji', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
}

body {
	color: var(--color-main-text);
}
```

The whole body block can be removed aswell from the css file.

st3iny (approved):
Tested on my local instance and mails are displayed correctly when using a dark theme. Forcing a white background works a lot better than my former injection hack.

ChristophWurst:
Let's see if the backport works, then we can do this for 1.8.x, otherwise we have this only in 1.9

Issue #3146, opened 2020-05-22
Support strong authentication


temuccio:
### Feature Request

Include support strong authetication alternative to PLAIN and LOGIN such us:

- CRAM-MD5
- GSSAPI
- DIGEST-MD5
- MD5
- OAUTH10A
- OAUTHEBEARER
- SCRAM-SHA-1 
- NTLM

#### Summary

In my case I have try to connect mail with my server that support CRAM-MD5 and DIGEST-MD5 without positive response.
It is interesting to include this support such as most important clients of mail for extend the usage.
I'm happy to collaborate to implement this Feature.
Have nice day.

ChristophWurst:
Adding this to the end of our roadmap

ChristophWurst:
> I'm happy to collaborate to implement this Feature.

Sounds good :) I assume the Horde libs can already handle this. So we should only need a new structure for the authentication method (new attribute inthe accounts table?) and a switch in the user interface.

temuccio:
Hi @ChristophWurst, sorry for my delay.
I have check Horde libs and it support this authentication.
For your interested, I have check that in:
`vendor/pear-pear.horde.org/Horde_imap_Client/Horde/Imap/Client/Socket/Pop3.php`
and
`vendor/pear-pear.horde.org/Horde_Smtp/Horde/Smtp.php`
is implemented the strong authentication.
As you annunced, it is necessary to add a "select" for decide what is the method of authetication.
How you pass the parameter to Hord?

ChristophWurst:
> new attribute inthe accounts table

basically that. and then just read it through a getter on the account class in the IMAP client factory.

kesselb:
Hello :wave:, looked into this issue today. 

**CRAM-MD5 / DIGEST-MD5**: Is already supported. If the connection to the imap server is unencrypted and CRAM-MD5 or DIGEST-MD5 are supported.

**OAUTHEBEARER**: Would be nice for Gmail users. Code to use XOAUTH2 for authentication to IMAP and SMTP is there but the process to obtain an access token via OAUTH needs to be done. I guess that's somehow similar to https://github.com/nextcloud/integration_google. Additional information: https://developers.google.com/gmail/imap/xoauth2-protocol / https://developers.google.com/identity/protocols/oauth2. 

**SCRAM-SHA-1**: https://github.com/nextcloud/mail/pull/4377

[...]

PR #4377, opened 2021-01-19
Add horde_stringprep to support SCRAM-SHA-1

kesselb:
Related to #3146 

Horde_Imap_Client supports the SCRAM authentication mechanism but Horde_Stringprep as dependency is required (https://github.com/horde/Imap_Client/blob/e5302ee8b238f03564d0ea4177625214f79a3ebe/lib/Horde/Imap/Client/Auth/Scram.php#L98-L101).

If the server supports SCRAM-SHA-1 and the connection is unencrypted we are now able to use it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
 	"config": {
 		"platform": {
 			"php": "7.3"
-		}
+		},
+		"sort-packages": true
 	},
 	"require": {
 		"php": ">=7.3",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -40,7 +40,8 @@
 		"pear-pear.horde.org/horde_util": "^2.5.8@stable",
 		"pear-pear.horde.org/horde_smtp": "^1.9.5@stable",
 		"psr/log": "^1",
-		"rubix/ml": "0.2.4"
+		"rubix/ml": "0.2.4",
+		"pear-pear.horde.org/horde_stringprep": "^1.0"
```
ChristophWurst, line 44: bonus points for moving the line up in alphabetical order :P

kesselb: Added the sort-packages flag. Next time composer will do for us ;)

composer.json
```diff
@@ -13,7 +13,8 @@
 	"config": {
 		"platform": {
 			"php": "7.3"
-		}
+		},
+		"sort-packages": true
```
ChristophWurst, line 17: :grinning:

ChristophWurst (approved):

ChristophWurst:
@kesselb rebase

kesselb:
@ChristophWurst done. Small version bump for php-cs-fixer and new github url for webmozart.

ChristophWurst:
you rebased away your DCO sign off

kesselb:
Hi @Neustradamus,

> Can you add other SCRAM?
> SHA-256/SHA-512/SHA3-512 and -PLUS variants too?

This depends on Horde ImapClient[^1] and their peer dependencies.

[^1]: https://github.com/horde/Imap_Client/blob/e5302ee8b238f03564d0ea4177625214f79a3ebe/lib/Horde/Imap/Client/Auth/Scram.php

Neustradamus:
@kesselb: I wish you a Happy New Year 2024!

Recently, there have been a lot of updates in PEAR about SCRAM.

It is possible to look for Horde and Nextcloud?

PR #4382, opened 2021-01-21
Unify action menu icons

GretaD:
@jancborchardt the version of icon-star-dark is lighter than other icons. I used the same as on talk. Should we keep it like this?
favorite_star
favoritegold

### DIFF
--- a/src/components/MenuEnvelope.vue
+++ b/src/components/MenuEnvelope.vue
@@ -37,7 +37,7 @@
 					envelope.flags.important ? t('mail', 'Mark unimportant') : t('mail', 'Mark important')
 				}}
 			</ActionButton>
-			<ActionButton icon="icon-starred"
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Are the screenshots before and after? Why is the second one still yellow?

GretaD:
> Are the screenshots before and after? Why is the second one still yellow?

thats how it should be. When its a favorite message- its dark, when its unfavorite- its gold. Its the same on talk and files.
I double checked on design channel. What we have now on mail is gold all the time.

jancborchardt:
Yeah, the grey star should have the same grey tone as the others. :)

Slightly related as it’s about the actions menu: The "Edit as new message" shouldn’t be that far up, no? I’d say we should put it as second-to-last action right before Delete

ChristophWurst (approved):

PR #4401, opened 2021-01-25
Fix recipient selection

GretaD:
To improve a bit the multiselection on mail

needs  - [x] https://github.com/nextcloud/nextcloud-vue/pull/1718


src/components/Composer.vue
```diff
@@ -29,11 +29,14 @@
 				:taggable="true"
 				label="label"
 				track-by="email"
+				:limit="4"
 				:multiple="true"
 				:placeholder="t('mail', 'Contact or email address …')"
 				:clear-on-select="false"
+				:close-on-select="false"
```
miaulalala, line 36: Can we change this to true so the recipient picker closes when pressing enter?

miaulalala (changes requested):

miaulalala (approved):

PR #4405, opened 2021-01-25
Ignore subfolders when guessing special role of mailbox

### LINKED ISSUE #4404: Subfolders shouldn't be taken into account for special roles
### How to reproduce
Add new email account that has a subfolder called "inbox" and or "drafts". E.g.
```
    - INBOX
    - Drafts
    - ...
    - MyProject
      - Inbox
      - Accounting
      - 2020
      - Drafts
    - Other Folder
```

### Expected behavior
The top-level mailboxes "INBOX" and "Drafts" are getting a dedicated icon and are moved to the top. Other folders, especially the subfolders "MyProject/Inbox" and "MyProject/Drafts" are treated and displayed as normal folders where they are.

### Actual behavior
Subfolders "MyProject/Inbox" and "MyProject/Drafts" are treated as special role mailboxes and show up as an additional inbox (or drafts folder resp.) with the special icon etc on the top level (see screenshot).

Bildschirmfoto 2021-01-25 um 16 55 18

**Warning:** Careful when testing  code changes. The special role is assigned when adding the account to Nextcloud and will not be refreshed when reloading.

### Mail app

**Mail app version:** 1.8.0 - master (c63a38f99ca2d1e061c2a3bcbe05552098d93b01)

**Mailserver or service:** Host Europe, I think it's Dovecot IMAP

### Server configuration

**PHP version:** 7.4

**Nextcloud Version:** 20.0.5 - master (a308c9be90ba174154727c4239cd968707d4c7e8)

#### Client configuration
**Browser:** Firefox 84

xMartin:
fixes #4404

The issue is caused by using only the last segment of the mailbox path to detect special use (e.g. inbox, sent, junk) and that creates false positive cases for nested folders with certain names.

This PR removes this which fixes it for me. But I'm not sure why this is being done. The code is very old (oldest occurrence from 2017 I found is here: https://github.com/nextcloud/mail/blob/55ff9146da707ef702521df23eb63764c4c9d21d/lib/Mailbox.php#L394). There might be cases where this approach is useful to some extend. Somewhere else in a comment it is mentioned that there are cases like "INBOX.Archive.2020". I don't know. If you look at the code you will see that it is also only working to the second level (`explode` is getting a "2" as the third argument) so someone might have written this for some special use-case.

**Note:** As noted in the issue, this code is not run on page load, but when adding a new account and then persisted to settings and read from an encoded string from the HTML on load. Mind this when testing.

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -224,8 +224,7 @@ class FolderMapper {
 			'junk' => ['junk', 'spam', 'bulk mail'],
 		];
 
-		$lowercaseExplode = explode($folder->getDelimiter(), $folder->getMailbox(), 2);
-		$lowercaseId = strtolower(array_pop($lowercaseExplode));
+		$lowercaseId = strtolower($folder->getMailbox());
 		foreach ($specialFoldersDict as $specialRole => $specialNames) {
 			if (in_array($lowercaseId, $specialNames)) {
 				$folder->addSpecialUse($specialRole);
--- a/tests/Unit/IMAP/FolderMapperTest.php
+++ b/tests/Unit/IMAP/FolderMapperTest.php
@@ -241,9 +241,6 @@ class FolderMapperTest extends TestCase {
 			->method('getSpecialUse')
 			->willReturn([]);
 		$folders[0]->expects($this->once())
-			->method('getDelimiter')
-			->willReturn('.');
-		$folders[0]->expects($this->once())
 			->method('getMailbox')
 			->willReturn('Sent');
 		$folders[0]->expects($this->once())


ChristophWurst:
> There might be cases where this approach is useful to some extend. Somewhere else in a comment it is mentioned that there are cases like "INBOX.Archive.2020". I don't know. If you look at the code you will see that it is also only working to the second level (`explode` is getting a "2" as the third argument) so someone might have written this for some special use-case.

Right, that is the only edge case that I could think of. We generally handle this a bit better now and I found that *personal namespace* is the technical term for the `INBOX.` prefixes.

I'll have to check where we do this and how. IIRC it moved to the js-side. On that node I'm not even sure to what extend we actually still need this distinction on the back-end. Moreover we now have a user setting for *special* mailboxes. So this is all a bit confusing. I will look into it and want to avoid rushing a change that could cause trouble for others.

ChristophWurst:
> If you look at the code you will see that it is also only working to the second level (`explode` is getting a "2" as the third argument) so someone might have written this for some special use-case

If I'm not mistaken this has historic reasons. Until a few months ago we only showed mailboxes at maximum two levels. Now we show the full hierarchy as tree. This could explain why the 2 was there.

lib/IMAP/FolderMapper.php
```diff
@@ -224,8 +224,7 @@ protected function guessSpecialUse(Folder $folder): void {
 			'junk' => ['junk', 'spam', 'bulk mail'],
 		];
 
-		$lowercaseExplode = explode($folder->getDelimiter(), $folder->getMailbox(), 2);
```
ChristophWurst, line 227: I checked current `main` and that code is no longer there :+1:

ChristophWurst:
Should be resolved already. Thanks a lot nevertheless!

xMartin:
The issue still exists for me with Mail version 3.2.2.

ChristophWurst:
Unfortunately the PR will have to be redone as too much code was moved around.

lib/IMAP/FolderMapper.php
xMartin: https://github.com/nextcloud/mail/blob/ec576848a9f0d3527720232f611f36107334e9c2/lib/IMAP/FolderMapper.php#L245

PR #4406, opened 2021-01-25
Fix html signature format balloon

GretaD:

### DIFF
--- a/src/components/SignatureSettings.vue
+++ b/src/components/SignatureSettings.vue
@@ -136,3 +136,8 @@ export default {
 	margin-bottom: 23px;
 }
 </style>
+<style>
+.ck-balloon-panel {
+	z-index: 10000 !important;
+}
+</style>


ChristophWurst:
on a side note I think that means we have to be double-careful with future updates of CKEditor as the css selectors could change or their base styles that we have to enforce here

src/components/AccountSettings.vue
```diff
@@ -37,7 +37,7 @@
 		</AppSettingsSection>
 		<AppSettingsSection :title="t('mail', 'Signature')">
 			<p class="settings-hint">
-				{{ t('mail', 'A signature is added to the text of new messages and replies.') }}
+				{{ t('mail', 'A signature is added to the text of new messages and replies. Select the text to format it.') }}
```
GretaD, line 40: I think this line makes it more friendly for people to understand how to format it

ChristophWurst (approved):

webcoreinc:
Where is the email sig setting in NC? I can't find this anywhere.

ChristophWurst:
@webcoreinc https://help.nextcloud.com/c/apps/mail/35 is the place for questions.

Issue #4403, opened 2021-01-25
Wrong attachment is sent when adding an attachment from Files


StCyr:
### Expected behavior
Proper attachment sould be sent

### Actual behavior
Another attachment is sent (not random: if you send 3 times an email with fileA as attachment, it will consistenly be fileB which will be sent, not once fileB, then another one fileC,...)

The "send" request sent to the server contains the info about fileA. So, it seems the problem occurs after that point.

### Mail app

**Mail app version:** 1.8

ChristophWurst:
> f you send 3 times an email with fileA as attachment, it will consistenly be fileB which will be sent

if you only ever picked file a, where does file b come from?

ChristophWurst:
@davidpr78 @st3iny could anyone test the v1.8.1? is this fixed? we got a new and similar report with https://github.com/nextcloud/mail/issues/4464 but I can't reproduce

PR #4407, opened 2021-01-25
fixes an issue where attachment sent would not be the one selected

StCyr:
Could you please test this fix to #4403 ?

For some reason, the attachment popover won't open anymore on my development setup but it doesn't make much sense (some caching issue maybe?...)

### DIFF
--- a/src/components/ComposerAttachments.vue
+++ b/src/components/ComposerAttachments.vue
@@ -103,15 +103,6 @@ export default {
 		onAddLocalAttachment() {
 			this.$refs.localAttachments.click()
 		},
-		fileNameToAttachment(name, size, id) {
-			return {
-				fileName: name,
-				displayName: trimStart('/')(name),
-				id,
-				size,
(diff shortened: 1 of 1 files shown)


GretaD (approved):
This pr fixes the problem. I tested it and it works

src/components/ComposerAttachments.vue
```diff
@@ -157,8 +148,15 @@ export default {
 
 				return uploadLocalAttachment(file, progress(file.name)).then(({ file, id }) => {
 					logger.info('uploaded')
-					return this.emitNewAttachments([this.fileNameToAttachment(file.name, file.size, id)])
+					return this.emitNewAttachments([{
+						fileName: file.name,
+						displayName: trimStart('/')(name),
```
ChristophWurst, line 153: `name` is undefined. this broke local attachments

StCyr: yes I'd discovered that too. but you fix it already I believe

Issue #3723, opened 2020-10-05
Do not auto read mark


jookk:
### Feature Request
Open mail app in Nextcloud. If you have new unread messages, it is actually
automatically shown / opened and marked as read. Very easy can be that
message unseen.

#### Summary

n/a

ChristophWurst:
?

jookk:
Open mail app in Nextcloud. If you have new unread messages, it is actually 
automatically shown / opened and marked as read. Very easy can be that 
message unseen.

Dňa 5. októbra 2020 14:42:25 používateľ Christoph Wurst 
<notifications@github.com> napísal:

StCyr:
Don't we just want a small delay before marking the mail read (ie: the message must be open for at least <x> seconds for the read mark to be applied)?

ChristophWurst:
> Don't we just want a small delay before marking the mail read (ie: the message must be open for at least seconds for the read mark to be applied)?

Sure :+1:

PR #4418, opened 2021-01-27
Sets a delay of 2 seconds before marking a message as read

StCyr:
Fix to #3723 

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -133,6 +133,7 @@ export default {
 			error: undefined,
 			message: undefined,
 			importantSvg,
+			seenTimer: undefined,
 		}
 	},
 	computed: {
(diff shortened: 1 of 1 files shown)


src/components/ThreadEnvelope.vue
```diff
@@ -212,7 +219,11 @@ export default {
 				logger.debug(`message ${this.envelope.databaseId} fetched`, { message })
 
 				if (!this.envelope.flags.seen) {
-					this.$store.dispatch('toggleEnvelopeSeen', { envelope: this.envelope })
+					logger.info('Starting timer to mark message as seen/read')
+					this.seenTimer = setTimeout(() => {
+						this.$store.dispatch('toggleEnvelopeSeen', { envelope: this.envelope })
+						this.seenTimer = undefined
```
ChristophWurst, line 225: I don't think this is necessary

ChristophWurst (approved):

src/components/ThreadEnvelope.vue
StCyr: what? to keep track of the timerId?

This is needed, I've test it.

I believe the timer is attached to the window, so it's not destroyed when one navigates away from the message

ChristophWurst: No I mean just the part of unsetting the value. I don't think calling clearTimer on a timer ID that has already fired will do any harm. But leave it as is, it's just a nitpick :)

Issue #3880, opened 2020-10-23
Can not access mail account in nextcloud after password change on imap server


RainerEmrich:
### Expected behavior
There should be an option to set the new password.

### Actual behavior
Mail account is not shown anymore in nextcloud. The cronjob gives "OCA\Mail\Exception\ServiceException: IMAP errorMail server denied authentication." Can not access the mail account, so I even can not delete it.

### Mail app

**Mail app version:**  1.4.1

**Mailserver or service:** imap.1und1.de

### Server configuration

**Operating system**: Ubuntu 18.04

**Web server:** Apache 2.4.46

**Database:** MariaDB 10.4.15

**PHP version:** 7.3.23

**Nextcloud Version:** 19.0.4

#### Client configuration
**Browser:** Firefox 82.0

RainerEmrich:
By the way, is there a known procedure to get rid of this dangling mail account. I tried "occ mail:clean-up", but this didn't work.

ChristophWurst:
https://github.com/nextcloud/mail/pull/4419 fixes  this

PR #4419, opened 2021-01-28
Make first mailbox of navigation account optional

GretaD:
Fixes https://github.com/nextcloud/mail/issues/3880

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -106,7 +106,7 @@ export default {
 		},
 		firstMailbox: {
 			type: Object,
-			required: true,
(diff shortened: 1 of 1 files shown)


src/components/NavigationAccount.vue
```diff
@@ -106,7 +106,7 @@ export default {
 		},
 		firstMailbox: {
 			type: Object,
-			required: true,
+			required: false,
```
ChristophWurst, line 109: > Prop 'firstMailbox' requires default value to be set

you can make that `undefined`

ChristophWurst (approved):

Issue #3908, opened 2020-10-29
Stuck at "loading messages" at 1.5.1


stavros-k:
### Expected behavior
Load Mails

### Actual behavior
Stuck on Loading Messages

### Mail app

**Mail app version: 1.5.1** (see apps admin page, e.g. 0.5.3)

**Mailserver or service: @Outlook.com** (e.g. Outlook, Yahoo, Gmail, Exchange,...)

### Server configuration

**Operating system**: Unraid 6.8.3, Docker 19.03.5 (e.g. Debian 8)

**Web server: nginx** (e.g. Apache, Nginx,...)

**Database: PostgreSQL** (e.g. MariaDB, SQLite or PostgreSQL)

**PHP version: 7.3.23** (e.g. 7.0)

**Nextcloud Version: 20.0.1** (see admin page, e.g. 13.0.2)

#### Client configuration
**Browser: Chrome 86.0.4240.111** (e.g. Firefox 48)

**Operating system: Windows 10** (e.g. Arch Linux)

After updating to 1.5.1, two servers (same machine, different docker) promted me to reenter mail account password. After that it stuck at "Loading Messages", They have pulled the Folder structure, if i click any folder, it displays how many mails are inside.

I tried the following:

Deleted the mail account
run `occ mail:clean-up`
disabled and removed app
reinstalled, added the account again
but i still get “Loading messages”.

Opened the database with pgAdmin
the account is added, but no new messages appear on the db.

Tried running `occ mail:account:sync 10 -f` (my current account id) and got:

> [debug] account is up to date, skipping mailbox sync
> 
> In MailboxLockedException.php line 33:
> 
> 331 is already being synced

After 3 hours i still get “331 is already being synced” so it doesn’t download anything…

Nextcloud log level to 0, doesn’t show anything for the mail app
On chromes network dev tab i see a tof on “sync” in the list and the preview is:

> 
> {status: “fail”,…}
> 
> data: {message: “401 is already being synced”, type: “OCA\Mail\Exception\MailboxLockedException”}
> 
> 1. message: "401 is already being synced"
> 2. type: "OCA\Mail\Exception\MailboxLockedException"
> status: “fail”

Looks like it affects other people too
https://help.nextcloud.com/t/mail-app-v1-5-1-stucks-on-loading-messages/96535/2

If you need any more info let me know

ChristophWurst:
as noted on the forums the locked mailbox is not an error

anyhting in your nextcloud.log that indicates an **error**?

[...]

PR #4442, opened 2021-02-01
Makes sure email subject is valid UTF-8

### LINKED ISSUE #2968: Can't read mailbox (spinner running continuously)  when foreign characters are in Sender's address
### Expected behavior
When foreign characters are in  e-mail sender's e-mails , mails are shown in the mailbox's lists.

### Actual behavior
When this particular e-mail sender's e-mails are in mail header, app hang and can not show mailbox content (spinner running all the time). Examples (bold characters): stru**è**nog@exampledomain.com, vo**ð**enje@exampledomain.com, gra**ð**.dnevnika@exampledomain.com...

error log:
`Caused by OCA\Mail\Exception\ServiceException: Sync failed for 26:INBOX: An exception occurred while executing 'INSERT INTO "oc_mail_recipients" ("message_id", "type", "label", "email") VALUES(?, ?, ?, ?)' with params [634589, 0, "\x73\x74\x72\x75\xe8\x6e\x6f\x67\x40\x69\x70\x7a\x2e\x68\x72", "\x73\x74\x72\x75\xe8\x6e\x6f\x67\x40\x69\x70\x7a\x2e\x68\x72"]: SQLSTATE[22021]: Character not in repertoire: 7 ERROR: invalid byte sequence for encoding "UTF8": 0xe8 0x6e 0x6f`

### Mail app

**Mail app version:** 1.33

**Mailserver or service:** Postfix

### Server configuration

**Operating system**: Debian 9.12

**Web server:** Apache2 - 2.4.25

**Database:** PostgreSQL 9.6.17 on x86_64-pc-linux-gnu

**PHP version:** 7.3.17

**Nextcloud Version:** 18.0.3

#### Client configuration
**Browser:** Google Chrome 81.x

**Operating system:** Windows 10

StCyr:
Fixes #2968 

### DIFF
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -273,7 +273,7 @@ class IMAPMessage implements IMessage, JsonSerializable {
 	 * @return string
 	 */
 	public function getSubject(): string {
-		return $this->getEnvelope()->subject;
+		return iconv("UTF-8", "UTF-8//IGNORE", $this->getEnvelope()->subject);
 	}
 
 	/**


ChristophWurst (approved):
Makes sense!

Just one idea from https://github.com/nextcloud/mail/issues/3908#issuecomment-762377816: what if we first try to convert without the ignore option and only if the result is `false` we do another conversation with ignore?

like

```php
$utf8 = iconv('UTF-8', 'UTF-8', $subject);
if ($utf !== false) {
  return $utf8;
}
return iconv('UTF-8', 'UTF-8//IGNORE', $subject);
```

this means that the installations where the ignore part is problematic will less likely run into issues, only if there really is invalid data.

StCyr:
> 
> 
> Makes sense!
> 
> Just one idea from [#3908 (comment)](https://github.com/nextcloud/mail/issues/3908#issuecomment-762377816): what if we first try to convert without the ignore option and only if the result is `false` we do another conversation with ignore?
> 
> like
> 
> ```
> $utf8 = iconv('UTF-8', 'UTF-8', $subject);
> if ($utf !== false) {
>   return $utf8;
> }
> return iconv('UTF-8', 'UTF-8//IGNORE', $subject);
> ```
> 
> this means that the installations where the ignore part is problematic will less likely run into issues, only if there really is invalid data.

Ok. should I update the fix in #3855 ?

ChristophWurst:
> Ok. should I update the fix in #3855 ?

That would be great!

PR #4453, opened 2021-02-02
Avoids using iconv with '//IGNORE' flag if possible

StCyr:
As requested in https://github.com/nextcloud/mail/pull/4442#issuecomment-771666563


ChristophWurst (approved):
Awesome!

PR #4465, opened 2021-02-03
Fix filter expression

rullzer:
!a OR (a OR b) is always true ;)
Now it is also actually doing what the comments say it does.


kesselb (approved):

ChristophWurst (approved):
Potato potato

PR #4472, opened 2021-02-03
Sieve

kesselb:
**Todo:**

- [x] Add Managesieve as dependency
- [x] DB fields to store sieve configuration
- [x] Provisioning support
- [x] Account Form
- [x] Text-Editor for current sieve script
- [x] Custom Logger for ManageSieve (because Horde_Imap_Client_Base_Debug do not work)
- [x] Migration to add `sieveEnabled = false` to existing provisioning configurations
- [x] Unit-Test for SieveController
- [x] Test upgrade from 1.8.0 (with and without provisioning)

**Nice to have:**

- [x] Validation for sieve script (is done by managesieve itself)
- [ ] Detect sieve/managesieve via capabilities
- [ ] Move SieveController.updateAccount to service

**CI:** 

~~Build fails because the sieve integration server uses self signed certificates. 
We need https://github.com/horde/ManageSieve/pull/1 or apply the patch on CI.~~

**Screenshots:**

image
image
image

### DIFF
--- a/lib/Controller/SieveController.php
+++ b/lib/Controller/SieveController.php
@@ -1,0 +1,219 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @author Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Mail\Controller;
+
+use Horde\ManageSieve\Exception as ManagesieveException;
+use OCA\Mail\AppInfo\Application;
+use OCA\Mail\Db\MailAccountMapper;
+use OCA\Mail\Exception\ClientException;
+use OCA\Mail\Exception\CouldNotConnectException;
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Sieve\SieveClientFactory;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Http\JSONResponse;
+use OCP\IRequest;
+use OCP\Security\ICrypto;
+
+class SieveController extends Controller {
+
+	/** @var AccountService */
+	private $accountService;
+
+	/** @var MailAccountMapper */
+	private $mailAccountMapper;
+
+	/** @var SieveClientFactory */
+	private $sieveClientFactory;
+
+	/** @var string */
+	private $currentUserId;
+
+	/** @var ICrypto */
+	private $crypto;
+
+	/**
+	 * AccountsController constructor.
+	 *
+	 * @param IRequest $request
+	 * @param string $UserId
+	 * @param AccountService $accountService
+	 * @param MailAccountMapper $mailAccountMapper
+	 * @param SieveClientFactory $sieveClientFactory
+	 * @param ICrypto $crypto
+	 */
+	public function __construct(IRequest $request,
+								string $UserId,
+								AccountService $accountService,
+								MailAccountMapper $mailAccountMapper,
+								SieveClientFactory $sieveClientFactory,
+								ICrypto $crypto
+	) {
+		parent::__construct(Application::APP_ID, $request);
+		$this->currentUserId = $UserId;
+		$this->accountService = $accountService;
+		$this->mailAccountMapper = $mailAccountMapper;
+		$this->sieveClientFactory = $sieveClientFactory;
+		$this->crypto = $crypto;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param int $id account id
+	 *
+	 * @return JSONResponse
+	 *
+	 * @throws CouldNotConnectException
+	 * @throws ClientException
+	 */
+	public function getActiveScript(int $id): JSONResponse {
+		$sieve = $this->getClient($id);
+
+		$scriptName = $sieve->getActive();
+		if ($scriptName === null) {
+			$script = '';
+		} else {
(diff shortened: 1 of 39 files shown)


lib/Service/SetupService.php
```diff
@@ -124,12 +141,43 @@ public function createNewAccount($accountName, $emailAddress, $imapHost, $imapPo
...
+		$transport = $this->smtpClientFactory->create($account);
+		if ($transport instanceof Horde_Mail_Transport_Smtphorde) {
+			try {
+				$transport->getSMTPObject();
+			} catch (Horde_Mail_Exception $e) {
+				throw CouldNotConnectException::create($e, 'SMTP', $mailAccount->getOutboundHost(), $mailAccount->getOutboundPort());
+			}
+		}
+
+		if ($mailAccount->isSieveEnabled()) {
```
kesselb, line 175: @kesselb no check for sieve connectivity here.

src/errors/convert.js
```diff
@@ -54,5 +56,5 @@ export const convertAxiosError = (axiosError) => {
 		return axiosError
 	}
 
-	return new map[response.data.data.type]()
+	return new map[response.data.data.type](response.data.message)
```
kesselb, line 59: @ChristophWurst is it okay to forward the response message to the exception?

ChristophWurst: sure!

src/errors/convert.js
```diff
@@ -24,13 +24,17 @@ import MailboxNotCachedError from './MailboxNotCachedError'
...
+import ManageSieveError from './ManageSieveError'
 
 const map = {
 	'OCA\\Mail\\Exception\\DraftsMailboxNotSetException': NoDraftsMailboxConfiguredError,
 	'OCA\\Mail\\Exception\\MailboxLockedException': MailboxLockedError,
 	'OCA\\Mail\\Exception\\MailboxNotCachedException': MailboxNotCachedError,
 	'OCA\\Mail\\Exception\\SentMailboxNotSetException': NoSentMailboxConfiguredError,
 	'OCA\\Mail\\Exception\\TrashMailboxNotSetException': NoTrashMailboxConfiguredError,
+	'OCA\\Mail\\Exception\\CouldNotConnectException': CouldNotConnectError,
+	'Horde\\ManageSieve\\Exception': ManageSieveError,
```
kesselb, line 37: @ChristophWurst should we wrap `Horde\ManageSieve\Exception`?

ChristophWurst: Using the one from horde is also fine by me

kesselb:
Did some testing today for the update path.

Test instance with 3 users and a provisioning configuration. In addition configured a normal email account. Migration added the new fields (to db and provisioning config). 

Update gone through without issues. State after the update es expected (sieve disabled for all accounts).

src/components/SieveFilterForm.vue
```diff
@@ -0,0 +1,81 @@
...
+.section {
+	display: block;
+	padding: 0;
+	margin-bottom: 23px;
+}
+
+textarea {
+	width: 100%;
+}
+
```
miaulalala, line 71:
```suggestion
#sieve-text-area {
    font-size: 14px;
    line-height: 22px;
    font-family: monospace;
}
```

lib/Controller/SieveController.php
```diff
@@ -0,0 +1,219 @@
...
+	 *
+	 * @param IRequest $request
+	 * @param string $UserId
+	 * @param AccountService $accountService
+	 * @param MailAccountMapper $mailAccountMapper
+	 * @param SieveClientFactory $sieveClientFactory
+	 * @param ICrypto $crypto
+	 */
+	public function __construct(IRequest $request,
+								string $UserId,
```
miaulalala, line 67:
```suggestion
								string $userId,
```

lib/Controller/SieveController.php
```diff
@@ -0,0 +1,219 @@
...
+								AccountService $accountService,
+								MailAccountMapper $mailAccountMapper,
+								SieveClientFactory $sieveClientFactory,
+								ICrypto $crypto
+	) {
+		parent::__construct(Application::APP_ID, $request);
+		$this->currentUserId = $UserId;
```
miaulalala, line 74:
```suggestion
		$this->currentUserId = $userId;
```

lib/Service/Provisioning/Manager.php
```diff
@@ -131,6 +141,19 @@ private function updateAccount(IUser $user, MailAccount $account, Config $config
...
+			$account->setSieveSslMode($config->getSieveSslMode());
+		} else {
+			$account->setSieveUser(null);
+			$account->setSieveHost(null);
+			$account->setSievePort(null);
+			$account->setSieveSslMode(null);
+		}
 
 		return $account;
 	}
```
miaulalala, line 159:
```suggestion
		if ($config->getSieveEnabled()) {
			$account->setSieveUser($config->buildSieveUser($user));
			$account->setSieveHost($config->getSieveHost());
			$account->setSievePort($config->getSievePort());
			$account->setSieveSslMode($config->getSieveSslMode());
			
			return $account;
		}
		
		$account->setSieveUser(null);
		$account->setSieveHost(null);
		$account->setSievePort(null);
		$account->setSieveSslMode(null);

		return $account;
	}
```

No need to wait to return here, if the condition is fulfilled, we can safely return the set values.

lib/Controller/SieveController.php
ChristophWurst: I think this one needs the weird first uppercase U because on the DI container the parameter is registered as `UserId`

miaulalala (commented):
It works fine, no issues.

lib/Controller/SieveController.php
miaulalala: ah ok :)

ChristophWurst (approved):
Works very well! :shipit:

dumblob:
Looks really good and minimal to me. Well done! Thanks a lot.

Dumb question (I didn't apply this patch yet) - has this been tested with multi-account setups?

ChristophWurst:
> Dumb question (I didn't apply this patch yet) - has this been tested with multi-account setups?

Yes. There is a setting per account.

lib/Service/Provisioning/Manager.php
ChristophWurst: yes and no :wink: 

I think this is one of those cases where the if and the else branch have equal importance. The early return would (to me) signal that it's some error condition or anything very specific.

I would keep the explicit if-else here.

But again, this formatting doesn't change any logic, nor does it make anything faster. I'm fine with either :)

src/components/SieveFilterForm.vue
ChristophWurst: we will evaluate the usage of a nicer editor with syntax highlighting anyway, so let's address this later

Issue #2794, opened 2020-03-26
Clicking on current folder again shows "No message selected"


jancborchardt:
1. Go to any folder
2. The first mail shows in the detail view
3. Click the folder again
4. The mail vanishes and the detail view shows "No message selected"

Instead, clicking the folder again should not do anything.

Issue #2956, opened 2020-04-20
Emptycontent message flashes before loading spinner shows


jancborchardt:
Happens on current master:
Whenever you switch folders, you see the "No message selected" flash in the mail detail view for a second, only then the spinner appears.

Instead, the spinner should appear directly as a message will be loaded.

StCyr:
I wonder if we couldn't get rid of this NoMessageSelected component @ChristophWurst 

Especially after https://github.com/nextcloud/mail/pull/4510

PR #4510, opened 2021-02-10
Reloads the envelope list when user clicks on the folder they are currently in

StCyr:
Fixes https://github.com/nextcloud/mail/issues/2794

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -143,7 +143,7 @@ export default {
 				const first = this.envelopes[0]
 				if (typeof first !== 'undefined') {
 					logger.debug('refreshing mailbox')
-					if (first.flags.draft) {
+					if (this.$route.params.mailboxId === this.account.draftsMailboxId) {
 						this.$router.replace({
 							name: 'message',
 							params: {
@@ -236,7 +236,7 @@ export default {
 					// Keep the selected account-mailbox combination, but navigate to the message
 					// (it's not a bug that we don't use first.accountId and first.mailboxId here)
 					logger.debug('showing the first message of mailbox ' + this.$route.params.mailboxId)
-					if (first.flags.draft) {
+					if (this.$route.params.mailboxId === this.account.draftsMailboxId) {
 						this.$router.replace({
 							name: 'message',
 							params: {
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -137,6 +137,35 @@ export default {
 		},
 	},
 	watch: {
+		// Force the envelope list to re-render when user clicks on the folder it's currently in
+		$route(to) {
+			if (to.params.threadId === undefined) {
+				const first = this.envelopes[0]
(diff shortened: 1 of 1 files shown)


GretaD:
2 things i noticed while testing
1. The "No message" it is shown for 1 second before the message is loaded. 
2. The first message disappears for 1 second and then shown again

ezgif com-video-to-gif(3)

StCyr:
> 2 things i noticed while testing
> 
>     1. The "No message" it is shown for 1 second before the message is loaded.
> 
>     2. The first message disappears for 1 second and then shown again

The first issue is related to https://github.com/nextcloud/mail/issues/2956 and I wonder if we could not simply get rid of the NoMessageSelected  component

The second one, I cannot reproduce

GretaD:
> 
> The first issue is related to #2956 and I wonder if we could not simply get rid of the NoMessageSelected component
> 
oh, then its not because of your PR.

> The second one, I cannot reproduce

oh, interesting, this happens only with drafts, but not because of your PR.

src/components/Mailbox.vue
```diff
@@ -137,6 +137,15 @@ export default {
 		},
 	},
 	watch: {
+		// Reloads the envelope list when user clicks on the folder it's currently in
+		$route(to) {
+			if (to.params.threadId === undefined) {
+				this.loadEnvelopes()
```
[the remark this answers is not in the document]
StCyr: Yes, you're right. I'll fix that

src/components/Mailbox.vue
```diff
@@ -137,6 +137,35 @@ export default {
 		},
 	},
 	watch: {
+		// Force the envelope list to re-render when user clicks on the folder it's currently in
+		$route(to) {
+			if (to.params.threadId === undefined) {
+				const first = this.envelopes[0]
+				if (typeof first !== 'undefined') {
+					logger.debug('refreshing mailbox')
+					if (first.flags.draft) {
```
ChristophWurst, line 146: I have mixed feelings about this. can we take the account's `draftsMailboxId` for a more reliable decision?

StCyr: yes, but in this case we should probably also change the test at line 210 (in master):

https://github.com/nextcloud/mail/blob/f3c5de02b387081270bdbc6103c2fe4472e7dc12/src/components/Mailbox.vue#L210

miaulalala:
I just tested this but it doesn't seem to work. I still get the "No Message Selected" component shown when clicking on the inbox the second time. I had a look at the click event on the AppNavigationItem, and that doesn't trigger anything.

StCyr:
> I just tested this but it doesn't seem to work. I still get the "No Message Selected" component shown when clicking on the inbox the second time. I had a look at the click event on the AppNavigationItem, and that doesn't trigger anything.

Strange, it works here.

miaulalala:
Works fine for me now, looks very smooth @StCyr!

ChristophWurst (approved):

Issue #4415, opened 2021-01-27
Trusting a full domain and not just a sender


ChristophWurst:
### Feature Request

Related to https://github.com/nextcloud/mail/issues/505 it can be nice to trust not just one specific sender but the whole domain. E.g. when usera@domain.tld sends me an email I want any future emails from any domain.tld email to show images.

#### Summary

Trusted senders on a domain basis, not just the full email address.

PR #4513, opened 2021-02-11
Add trusted domain

GretaD:
Fixes https://github.com/nextcloud/mail/issues/4415

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -3,9 +3,16 @@
 		<MdnRequest :message="message" />
 		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content" style="color: #000000">
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
-			<button @click="onShowBlockedContent">
-				{{ t('mail', 'Show images from this sender') }}
-			</button>
+			<Actions default-icon="icon-toggle">
+				<ActionButton icon="icon-toggle"
+					@click="onShowBlockedContent">
+					{{ t('mail', 'Always show images from {sender}', {sender: message.from[0].email}) }}
+				</ActionButton>
+				<ActionButton icon="icon-toggle"
+					@click="onShowBlockedContentForDomain">
+					{{ t('mail', 'Always show images from {domain}', {domain: getDomain()}) }}
+				</ActionButton>
+			</Actions>
 		</div>
 		<div v-if="loading" class="icon-loading" />
 		<div id="message-container" :class="{hidden: loading, scroll: !fullHeight}">
@@ -23,6 +30,8 @@
 import { iframeResizer } from 'iframe-resizer'
 import PrintScout from 'printscout'
 import { trustSender } from '../service/TrustedSenderService'
+import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+import Actions from '@nextcloud/vue/dist/Components/Actions'
 
 import logger from '../logger'
 import MdnRequest from './MdnRequest'
@@ -30,7 +39,11 @@ const scout = new PrintScout()
 
 export default {
 	name: 'MessageHTMLBody',
-	components: { MdnRequest },
+	components: {
+		MdnRequest,
+		Actions,
+		ActionButton,
+	},
 	props: {
 		url: {
 			type: String,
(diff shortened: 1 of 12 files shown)


src/components/MessageHTMLBody.vue
```diff
@@ -23,14 +30,21 @@
 import { iframeResizer } from 'iframe-resizer'
 import PrintScout from 'printscout'
 import { trustSender } from '../service/TrustedSenderService'
+import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+import Actions from '@nextcloud/vue/dist/Components/Actions'
 
 import logger from '../logger'
 import MdnRequest from './MdnRequest'
 const scout = new PrintScout()
+const parser = require('tld-extract');
```
ChristophWurst, line 39: :fire:

src/components/MessageHTMLBody.vue
```diff
@@ -3,9 +3,16 @@
...
-				{{ t('mail', 'Show images from this sender') }}
-			</button>
+			<Actions default-icon="icon-toggle">
+				<ActionButton icon="icon-toggle"
+					@click="onShowBlockedContent">
+					{{ t('mail', 'Show images from {sender}', {sender: message.from[0].email}) }}
+				</ActionButton>
+				<ActionButton icon="icon-toggle"
+					@click="onShowBlockedDomain">
+					{{ t('mail', 'Show images from this {domain}', {domain: message.from[0].email.split('@').pop()}) }}
```
ChristophWurst, line 13: can we move the email to domain conversion into a method and use it here and in the other method below?

ChristophWurst (changes requested):

src/components/MessageHTMLBody.vue
```diff
@@ -3,9 +3,16 @@
 		<MdnRequest :message="message" />
 		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content" style="color: #000000">
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
-			<button @click="onShowBlockedContent">
-				{{ t('mail', 'Show images from this sender') }}
-			</button>
+			<Actions default-icon="icon-toggle">
+				<ActionButton icon="icon-toggle"
+					@click="onShowBlockedContent">
+					{{ t('mail', 'Show images from {sender}', {sender: message.from[0].email}) }}
```
ChristophWurst, line 9: what if we rename to *Always show …* to put some emphasis on the fact that this is a permanent setting?

src/components/MessageHTMLBody.vue
```diff
@@ -99,8 +112,27 @@ export default {
...
 			this.hasBlockedContent = false
-			await trustSender(this.message.from[0].email, true)
+			await trustSender(this.message.from[0].email, 'individual', true)
+
+		},
+		getDomain() {
+			const domain = this.message.from[0].email.split('@').pop()
+			return domain
+		},
+		async onShowBlockedDomain() {
```
ChristophWurst, line 122: `onShowBlockedContent` now exists twice

src/components/TrustedSenders.vue
```diff
@@ -24,6 +24,7 @@
 		<div v-for="sender in sortedSenders"
 			:key="sender.email">
 			{{ sender.email }}
+			{{ sender.type }}
```
ChristophWurst, line 27: we should translate this

ChristophWurst (changes requested):

src/components/MessageHTMLBody.vue
```diff
@@ -110,25 +110,18 @@ export default {
 			iframeDoc
 				.querySelectorAll('[data-original-style]')
 				.forEach((node) => node.setAttribute('style', node.getAttribute('data-original-style')))
-
+		},
+		async onShowBlockedContent() {
+			this.displayIframe()
 			this.hasBlockedContent = false
```
ChristophWurst, line 116: this one can also move into the `displayIframe` method as it's used here and below :)

src/components/TrustedSenders.vue
```diff
@@ -24,7 +24,7 @@
 		<div v-for="sender in sortedSenders"
 			:key="sender.email">
 			{{ sender.email }}
-
+			{{ t('mail', '{type}', {type: sender.type}) }}
```
ChristophWurst, line 27: this doesn't translate anything. the translation string only contains a placeholder but nothing else to translate.

you can do the following: add a method for `senderType(type)` with a `switch` statement. for individual and domain you can return a translated string. for anything else you fall back to `type`.

ChristophWurst (approved):
Code looks good!

PR #4516, opened 2021-02-12
Bump devDependencies

GretaD:

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,16 +79,16 @@
     "extends @nextcloud/browserslist-config"
   ],
   "devDependencies": {
-    "@babel/core": "^7.12.10",
+    "@babel/core": "^7.12.13",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
-    "@babel/preset-env": "^7.12.11",
+    "@babel/preset-env": "^7.12.13",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,22 +73,22 @@
     "vue-tabs-component": "^1.5.0",
     "vuex": "^3.6.2",
     "vuex-router-sync": "^5.0.0",
-    "webdav": "^3.6.2"
+    "webdav": "^4.0.0"
```
ChristophWurst, line 76: not a dev dependecy

ChristophWurst (changes requested):
why do you update all in one btw?

GretaD:
> why do you update all in one btw?

not a specific reason. I did it because was only one dependency and the rest dev dependency. Should i keep only the devdep?

ChristophWurst:
> not a specific reason. I did it because was only one dependency and the rest dev dependency. Should i keep only the devdep?

let's please do the bump separately if possible so we don't accidentally update anything that could cause breakage. also if we have separate bumps then we can bisect the critical update if there are any regressions later :)

GretaD:
close this and merge the bump individually

PR #4524, opened 2021-02-15
Allow adding senders to the address book

### LINKED ISSUE #4458: Allow adding senders to the address book
### Feature Request

As a user I want to be able to create a new contact form a sender of an email. Moreover I might want to add recipients of an email that I just sent to be added as well.

#### Summary

n/a

BroCanDo:
closes #4458

### DIFF
--- a/lib/Controller/ContactIntegrationController.php
+++ b/lib/Controller/ContactIntegrationController.php
@@ -1,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @author Kristian Lebold <kristian@lebold.info>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Mail\Controller;
+
+use OCA\Mail\Service\ContactIntegration\ContactIntegrationService;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\JSONResponse;
+use OCP\IRequest;
+
+class ContactIntegrationController extends Controller {
+
+	/** @var ContactIntegrationService */
+	private $service;
+
+	public function __construct(string $appName,
+								IRequest $request,
+								ContactIntegrationService $service) {
+		parent::__construct($appName, $request);
+
+		$this->service = $service;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param string $mail
+	 * @return JSONResponse
+	 */
+	public function match(string $mail): JSONResponse {
+		return new JSONResponse($this->service->findMatches($mail));
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param string $uid
+	 * @param string $mail
+	 * @return JSONResponse
+	 */
+	public function addMail(string $uid = null, string $mail = null): JSONResponse {
+		$res = $this->service->addEMailToContact($uid, $mail);
+		if ($res === null) {
+			return new JSONResponse([], Http::STATUS_NOT_FOUND);
+		}
+		return new JSONResponse($res);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param string $name
+	 * @param string $mail
+	 * @return JSONResponse
+	 */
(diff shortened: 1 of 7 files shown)


ChristophWurst:
This looks really good so far. Thanks a lot @BroCanDo! I'll try to test and review soon :v:

GretaD:
I tested it and all looks good, but design wise maybe we dont need button here: 
contact_button

Looks a bit crowded, i would say just text and icon would be fine.

BroCanDo:
@GretaD something like this?
image

GretaD:
Looks great. My last nitpick, text when its not hovered, should have the opacity .7, when hovered it should be 1. So its the same everywhere.

GretaD (approved):

lib/Controller/ContactIntegrationController.php
```diff
@@ -0,0 +1,101 @@
...
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param string $uid
+	 * @param string $mail
+	 * @return JSONResponse
+	 */
+	public function addMail(string $uid = null, string $mail = null): JSONResponse {
+		$res = $this->service->addEMailToContact($uid, $mail);
+		if ($res == null) {
```
miaulalala, line 66: please use strict comparisons

lib/Controller/ContactIntegrationController.php
```diff
@@ -0,0 +1,101 @@
...
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param string $name
+	 * @param string $mail
+	 * @return JSONResponse
+	 */
+	public function newContact(string $contactName = null, string $mail = null): JSONResponse {
+		$res = $this->service->newContact($contactName, $mail);
+		if ($res == null) {
```
miaulalala, line 82: please use strict comparisons

lib/Service/ContactsIntegration.php
```diff
@@ -118,4 +118,113 @@ private function getPhotoUri(string $raw) {
...
+		} else {
+			if (count($email) === 1 && $email[0]['value'] === '') {
+				$email = [];
+			}
+		}
+		$email[] = [ 
+			'type' => $type, 
+			'value' => $mailAddr 
+		];
+		$match['EMAIL'] = $email;
```
miaulalala: If you limit your search result to 1 by adding 'limit' => 1 to the search options, you can work without all the ifs.

Something like:

```
		$result = $this->contactsManager->search($uid, ['UID'], ['types' => true, 'limit' => 1]);
		$email = $result[0]['EMAIL'] ?? [];
		if( empty($email) ) {
			return null;
		}
```
should do

BroCanDo: @miaulalala First of all, thank you for your review! I can see why you dislike the ifs. Tbh, I think this is sub-optimal too.
To check for an acceptable output, I wrote a few test cases. Long story short: I tried my best to write something more straight forward. There are still a couple of ifs, but maybe its clearer now. I also hope that the test cases make sense in some way or another and the generated output is okay.
```
$result1 = [];
$result2 = [
			[
				'UID' => 3,
				'FN' => 'Johann Strauss II',
			]
		];
$result3 = [
			[
				'UID' => 3,
				'FN' => 'Johann Strauss II',
				'EMAIL' => [
				    'type' => 'HOME',
				    'value' => 'jonathan@frakes.com'
				    ]
			]
		];
$result4 = [
			[
				'UID' => 3,
				'FN' => 'Johann Strauss II',
				'EMAIL' => [[
				    'type' => 'HOME',
				    'value' => 'jonathan@frakes.com'
				    ]]
			]
		];
$result5 = [
			[
				'UID' => 3,
				'FN' => 'Johann Strauss II',
				'EMAIL' => []
			]
		];
$result6 = [
	[
		'UID' => 3,
		'FN' => 'Johann Strauss II',
		'EMAIL' => [[]]
	]
];
$result7 = [
	[
		'UID' => 3,
		'FN' => 'Johann Strauss II',
		'EMAIL' => [[
				    'type' => 'HOME',
				    'value' => ''
		    ]]
	]
];
$result8 = [
	[
		'UID' => 3,
		'FN' => 'Johann Strauss II',
		'EMAIL' => [
				    'type' => 'HOME',
				    'value' => ''
		    ]
	]
];


$tests = [$result1, $result2, $result3, $result4, $result5, $result6, $result7, $result8];

foreach ($tests as $result) {
    print("test case:\n");
    print_r($result);
    if (count($result) !== 1) {
        print("no match\n");
        continue;
    }
    
    $newEntry = [
        'type' => "HOME",
        'value' => "test@mail"
    ];
    
    $email = $result[0]['EMAIL'] ?? [];
    print("email entry\n");
    print_r($email);
    if(!empty($email) && !is_array($email[0])) {
        $email = [$email];
    }
    $email[] = $newEntry;
    print("after adding \n");
    print_r($email);
}
```

miaulalala: @BroCanDo Thanks for your response! Are those test cases all based on data from the contacts search result or is this you being super thorough and testing very well?

BroCanDo: @miaulalala They are partially (most of them) based on observed search results. Additionally I added a couple of edges cases which seemed plausible to me. Since I did not look at the implementation for the search, I thought that I rather handle a couple more edge cases.

miaulalala: Ok, that's fine,, thanks for being so thorough!

ChristophWurst (approved):
Works very well, code looks great :)

Issue #4532, opened 2021-02-17
Blank message with php8


ChristophWurst:
### Expected behavior

The app works with php8.

### Actual behavior

Any newly added messages are "blank" with php8 because Horde gives us null values for all the header fields we request due to an error on IMAP.

### Mail app

**Mail app version:**

v1.8.x or later

cc @miaulalala 

Upstream bug https://bugs.horde.org/ticket/15050

miaulalala:
Current workaround is to replace line 2932 in `Horde_Imap_Client_Socket` with `$cmd = ( is_string( $key ) === true && ( $key == 0 ) === false  )`. This will allow commands to be built with

ChristophWurst:
Fixed with #4534

PR #4534, opened 2021-02-17
Avoid usage of Horde header query as it causes issues with php8

ChristophWurst:
Ref #4532 

Todo
- [x] Migrate message body fetch
- [x] Migrate mdn header fetch


kesselb (commented):
Sending an MDN on PHP 7.4 still works.

kesselb (approved):

PR #4556, opened 2021-02-18
Add delete command to OCC

miaulalala:

To Do:

- [x] Add Command - exclude provisioned accounts
- [x] Test

### DIFF
--- a/lib/Command/DeleteAccount.php
+++ b/lib/Command/DeleteAccount.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @author Anna Larch <anna.larch@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Mail\Command;
+
+use OCA\Mail\Account;
+use Psr\Log\LoggerInterface;
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Exception\ClientException;
+use OCP\AppFramework\Db\DoesNotExistException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DeleteAccount extends Command {
+	public const ARGUMENT_ACCOUNT_ID = 'account-id';
+
+	/** @var AccountService */
+	private $accountService;
+
+	/** @var LoggerInterface */
+	private $logger;
+
+	public function __construct(AccountService $service,
+								LoggerInterface $logger) {
+		parent::__construct();
+
+		$this->accountService = $service;
+		$this->logger = $logger;
+	}
+
+	/**
+	 * @return void
+	 */
+	protected function configure() {
+		$this->setName('mail:account:delete');
+		$this->setDescription('Delete an IMAP account');
+		$this->addArgument(self::ARGUMENT_ACCOUNT_ID, InputArgument::REQUIRED);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$accountId = (int)$input->getArgument(self::ARGUMENT_ACCOUNT_ID);
+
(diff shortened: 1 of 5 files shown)


lib/Command/DeleteAccount.php
```diff
@@ -0,0 +1,101 @@
...
+		$consoleLogger = new ConsoleLoggerDecorator(
+			$this->logger,
+			$output
+		);
+
+		$accountId = (int)$input->getArgument(self::ARGUMENT_ACCOUNT_ID);
+
+		try {
+			$account = $this->accountService->findById($accountId);
+		} catch (Exception $e) {
```
ChristophWurst, line 75: this will catch a lot of things and not just the `DoesNotExistException` when the account isn't found. Let's do

```php
try {
    ...
  $consoleLogger->error("Account does not exist");
  return 1;
}
```

then any other unexpected error will just bubble up and fail hard.

lib/Command/DeleteAccount.php
```diff
@@ -0,0 +1,101 @@
...
+			throw new DoesNotExistException('This account does not exist');
+		}
+		$consoleLogger->info("Found account with email: " . $account->getEmail());
+
+		if ($account->getMailAccount()->getProvisioned() === true) {
+			throw new Exception('This is a provisioned account which can not be deleted from CLI. Use the Provisioning UI instead.');
```
ChristophWurst, line 81: let's also use `$consoleLogger->error` and return 2 or some other non-0 code.

lib/Command/DeleteAccount.php
```diff
@@ -0,0 +1,101 @@
...
+		}
+		$consoleLogger->info("Deleting " . $account->getEmail());
+		$this->delete($account, $consoleLogger);
+
+		$mbs = (int)(memory_get_peak_usage() / 1024 / 1024);
```
ChristophWurst, line 86: we can also do without this info. we have this in the sync command because it can take a lot of memory. the account deletion should not have any memory issues

ChristophWurst (changes requested):

lib/Command/DeleteAccount.php
miaulalala: Done

lib/Command/DeleteAccount.php
miaulalala: for all the error messages? just so it's consistent?

lib/Command/DeleteAccount.php
miaulalala: Done

lib/Command/DeleteAccount.php
ChristophWurst: yes, any error should return a non-zero exit code

tests/Unit/Command/DeleteAccountTest.php
```diff
@@ -0,0 +1,64 @@
...
+	private $logger;
+	private $command;
+	private $args = [
+		'account-id',
+	];
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->accountService = $this->getMockBuilder('\OCA\Mail\Service\AccountService')
```
kesselb, line 38:
```suggestion
		$this->accountService = $this->getMockBuilder(\OCA\Mail\Service\AccountService::class)
```

tests/Unit/Command/DeleteAccountTest.php
```diff
@@ -0,0 +1,64 @@
+<?php
```
kesselb, line 1:
```suggestion
<?php

declare(strict_types=1);
```

lib/Command/DeleteAccount.php
```diff
@@ -0,0 +1,98 @@
...
+			$account = $this->accountService->findById($accountId);
+		} catch (DoesNotExistException $e) {
+			$consoleLogger->error('This account does not exist');
+			return 2;
+		}
+		$consoleLogger->info("Found account with email: " . $account->getEmail());
+
+		if ($account->getMailAccount()->getProvisioned() === true) {
+			$consoleLogger->error('This is a provisioned account which can not be deleted from CLI. Use the Provisioning UI instead.');
+			return 2;
```
kesselb, line 81: Nice to have: A different exit code so admin's can write bash scripts and react to the different exit reasons. No account vs provisioned account.

lib/Command/DeleteAccount.php
```diff
@@ -0,0 +1,98 @@
...
+	 * @return void
+	 */
+	protected function configure() {
+		$this->setName('mail:account:delete');
+		$this->setDescription('Delete an IMAP account');
+		$this->addArgument(self::ARGUMENT_ACCOUNT_ID, InputArgument::REQUIRED);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$consoleLogger = new ConsoleLoggerDecorator(
```
kesselb, line 64: I'm unsure about `ConsoleLoggerDecorator` here because the output is written to logfile and stdout :thinking:

kesselb (approved):
Tested and works.

lib/Command/DeleteAccount.php
miaulalala: Good idea!

lib/Command/DeleteAccount.php
miaulalala: True, I'll change it to just a regular $output->writeLn

ChristophWurst (approved):

PR #4562, opened 2021-02-19
l10n: Change of letters

Valdnet:


ChristophWurst:
It seems very inconsistent, sometimes they use all upper case, then mixed case :shrug: 

https://en.wikipedia.org/wiki/ZIP_(file_format)#Standardization

Valdnet:
What we do? Do we change or leave it as it is? The name `Zip` is used throughout the NC translation. In my opinion, it must be regulated in some way.

ChristophWurst:
:+1: for consistency

ChristophWurst (approved):

Valdnet:
@ChristophWurst I'd like a merge, please 😉.

Valdnet:
@ChristophWurst Why was the PR closed if you accepted the changes?

ChristophWurst:
Wrong button, sorry

PR #4574, opened 2021-02-22
Add methods for important and favorite flag

GretaD:
Regression of #4359


ChristophWurst (approved):

ChristophWurst:
the regression is only on master, hence we don't need a backport

st3iny (approved):

Issue #4576, opened 2021-02-22
Cannot read property 'getPatcher' of undefined in FileService.js.


miaulalala:
Importing webdav as a whole doesn't work any more, instead `getClient` and `getPatcher` need to be imported separately in `FileService.js.` Otherwise the "add attachments from Files" option on an email is broken.

ChristophWurst:
regression of https://github.com/nextcloud/mail/pull/4501

PR #4577, opened 2021-02-22
Fix import for webdav

miaulalala:
Fixes https://github.com/nextcloud/mail/issues/4576


miaulalala:
See #4576

kesselb (approved):
Tested and works :+1:

PR #4579, opened 2021-02-22
Add app navigation to account settings

GretaD:
SInce now we have more than 3 sections, we need the navigation to make it easier to navigate.
the width is fixed on nc/vue
Before
beforee
After

afteree


ChristophWurst (approved):
Just one line changed and eslint is not happy :D

ChristophWurst (approved):

Issue #4466, opened 2021-02-03
Possible indexes


rullzer:
We noticed some slow queries (10+ seconds). After these indexes the slow query log stays very empty:

```sql
CREATE INDEX mail_messages_id_flags ON oc_mail_messages (mailbox_id, flag_important, flag_deleted, flag_seen);
```

```sql
CREATE INDEX mail_messages_id_flags ON oc_mail_messages (mailbox_id, flag_important, flag_deleted, flag_seen);
```

The names are not ideal and the queries might also not be. But it is a start to continue on.

ChristophWurst:
![Bildschirmfoto von 2021-02-03 19-24-59](https://user-images.githubusercontent.com/1374172/106791700-8d12b980-6655-11eb-944d-3fc4d0ed3aff.png)

rullzer:
![rxzNBzk](https://user-images.githubusercontent.com/45821/106806064-c2c09e00-6667-11eb-9ae5-2c2ba439f73c.gif)

rullzer:
Also foudna nother query that is not using proper indexes:

```sql
SELECT * FROM `oc_mail_messages` WHERE (`mailbox_id` IN (SELECT `id` FROM `oc_mail_mailboxes` WHERE `account_id` = 33)) AND ((`id` = 1730972) OR ((`thread_root_id` IS NOT NULL) AND (`thread_root_id` IN (SELECT `thread_root_id` FROM `oc_mail_messages` WHERE `id` = 1730972)))) ORDER BY `sent_at` desc;
```

ChristophWurst:
Using a join instead of subquery :rocket: 

```sql
SELECT m2.* FROM oc_mail_messages m1 LEFT JOIN oc_mail_messages m2 ON m1.thread_root_id = m2.thread_root_id WHERE m1.id = 229382 AND m1.mailbox_id IN (SELECT `id`    FROM `oc_mail_mailboxes`    WHERE `account_id` = 1248);
```

miaulalala:
Plus it needs an index on `thread_root_id`

ChristophWurst:
Little big detail :wink:

rullzer:
> Plus it needs an index on `thread_root_id`

Probably makes sense anyway

ChristophWurst:
Nope, without restructuring at least Mysql won't pick up the index

miaulalala:
[...]

PR #4583, opened 2021-02-23
Rework fetchThread Query

miaulalala:
Rework query to use a left join.

Still needs an index on `thread_root_id` but even without it, improves performance by about 50%.

see #4466


ChristophWurst (approved):
Works :shipit:

PR #4586, opened 2021-02-24
Hide mail server section if account is provisioned

kesselb:
| Before | After |
| --- | --- |
| Screenshot from 2021-02-24 15-05-57 | Screenshot from 2021-02-24 15-05-29 |


ChristophWurst (approved):
:rocket:

PR #4618, opened 2021-02-28
l10n: Change to a capital letter

Valdnet:


ChristophWurst:
but why? is there any rule for this? is it because the label is a title?

Valdnet:
> is there any rule for this?

@ChristophWurst The rule is, if you start with a capitalized word in other text strings in given options, then normalize it and use an uppercase letter as well. Examples below:

Sieve Host
Sieve Security
SSL/TLS
STARTTLS
Sieve Credentials
IMAP credentials
Custom
Sieve User
Sieve Password

Doesn't "IMAP credentials" look weird on this list?

ChristophWurst (approved):

PR #4624, opened 2021-03-01
Don't open new window on ZIP

miaulalala:
Fix from #4573 for issue #552

### DIFF
--- a/src/components/MessageAttachments.vue
+++ b/src/components/MessageAttachments.vue
@@ -118,6 +118,7 @@ export default {
 		downloadZip() {
 			window.open(this.zipUrl)
 			window.focus()
+			window.location = this.zipUrl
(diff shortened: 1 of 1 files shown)


src/components/MessageAttachments.vue
```diff
@@ -118,6 +118,7 @@ export default {
 		downloadZip() {
 			window.open(this.zipUrl)
 			window.focus()
+			window.location = this.url
```
ChristophWurst, line 121:
```suggestion
			window.location = this.zipUrl
```

right?

miaulalala: yes of course. ctrl+z did me dirty :laughing:

ChristophWurst (approved):

Issue #3869, opened 2020-10-22
Cronjob exhausts memory because of apps/mail/lib/IMAP/Threading/Container.php 


vasyugan:
### Expected behavior
cron should finish without error. The memory_limit in /etc/php/7.4/cli/php.ini is 4096m, so this should be plenty.

### Actual behavior
Two times per day, the cronjob gets killed because of exhausting memory. Before I set a memory_limit, I just saw that php was getting killed by the kernel and that nextcloud was down for several minutes. Then, I did set a memory limit and so I started getting the following message twice a day:
```
PHP Fatal error:  Allowed memory size of 4294967296 bytes exhausted (tried to allocate 262144 bytes) in /var/www/****/cloud.****/apps/mail/lib/IMAP/Threading/Container.php on line 111
PHP Fatal error:  Allowed memory size of 4294967296 bytes exhausted (tried to allocate 262144 bytes) in Unknown on line 0

```

### Mail app

**Mail app version:** 1.5.0

**Mailserver or service:** self-hosted dovecot

### Server configuration

**Operating system**: Ubuntu 20.04

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.4

**Nextcloud Version:** 20.0.0

#### Client configuration

(doesn't apply here)
**Operating system:** (e.g. Arch Linux)

ChristophWurst:
Please run ``occ mail:account:diagnose <accountId>`` for the account ID of that cron job.

ChristophWurst:
And because it's related to threading you could export the account's threading data with ``occ mail:account:export-threads --redact <accountId>`` if you want to share that. All the info is redacted. You would only leak your number of emails and how those are related on an abstract level without any real IDs or anything.

vasyugan:
https://gist.github.com/vasyugan/5b2200d8d9cd554d8f787814ba72a061

vasyugan:
> Please run `occ mail:account:diagnose <accountId>` for the account ID of that cron job.
```

IMAP capabilities:
- AUTH
- ENABLE
- ID
- IDLE
- IMAP4REV1
- LITERAL+
- LOGIN-REFERRALS
- SASL-IR

Horde error occurred: IMAP error reported by server.. See nextcloud.log for more details.

```

ChristophWurst:
> https://gist.github.com/vasyugan/5b2200d8d9cd554d8f787814ba72a061

I can build the threads just fine with only 24MB

```
[...]

PR #4628, opened 2021-03-01
Do now allow messages to be their own parents in threading

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3869

### How to test
* Store the json from https://github.com/nextcloud/mail/issues/3869#issuecomment-788171500 into a file
* Run `occ mail:thread path/to/file.json`

On master: :boom: 
Here: smooth sailing :boat:


szaimen (approved):
I guess the fix makes sense but I didn't check the test.

miaulalala (approved):
Works fine

Issue #4341, opened 2021-01-10
Mailbox unable to sync if previous sync died between message insertion and setting the sync token


mgriego:
### Expected behavior
Mailbox sync should be able to recover from prior failures.

### Actual behavior
If a previous mailbox sync fails (ie out of memory, execution time runs out, etc) in between the time that the messages begin being inserted in into the database and the time that the sync token is saved to the mailbox, then future syncs may fail resulting in a mailbox appearing stuck in a "Loading Messages" state.  This can happen because the sync token stores the most recently-retrieved UID to pass to the server on the next fetch.  If it is not updated due to the failure, then the Mail app will try to re-insert previously-downloaded messages, resulting in an SQL primary key constraint violation.  This will throw an exception that is not accounted for, resulting in a 500 response.  Note that, when this happens, subsequent sync calls from the browser will result in 409 responses as the sync locks do not get cleaned up when the exception is thrown.

Here is a real life example of this happening to me:
* My current sync token for my Inbox is: `VTI1NDg3OSxWMSxINjM3OTAx`.  When you `base64_decode` this, you see that the previous UID is `254879`.
* In my `oc_mail_messages` table, the three most recent UID values for the mailbox with the sync token shown above are `254878`, `254880`, and `254881`.

You can see from here that the UID referenced in the sync token has been removed from the mailbox, but two more messages (254880 and 254881) have been previously synced.  Since the token thinks the most recently-synced message is 254879, it uses that when retrieving messages, resulting in Mail attempting to re-insert messages 254880 and 254881.  Of course, the constraint violation happens as soon as Mail attempts to insert 254880, resulting in this error:

```
Sync failed for 1:INBOX: An exception occurred while executing 'INSERT INTO `oc_mail_messages` (`uid`, `message_id`, `references`, `in_reply_to`, `thread_root_id`, `mailbox_id`, `subject`, `sent_at`, `flag_answered`, `flag_deleted`, `flag_draft`, `flag_flagged`, `flag_seen`, `flag_forwarded`, `flag_junk`, `flag_notjunk`, `flag_important`) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params [254880, "REDACTED", "[]", null, "REDACTED", 58, "REDACTED", 1609722005, false, false, false, false, false, false, false, false, false]:

[...]

Issue #4095, opened 2020-11-18
Inbox content not visible, stuck at loading messages


patzm:
### Expected behavior
The inbox loads the list of emails in at most a few seconds.

### Actual behavior
The spinning "Loading messages" dialog does not go away.
![Peek 2020-11-18 10-35](https://user-images.githubusercontent.com/5219726/99512492-cf4b2980-2989-11eb-91f8-f1fe25c3889d.gif)

### Mail app

**Mail app version:** v1.7
**Mailserver or service:** one.com
I have multiple folders. The main inbox for instance has ~2k emails. Is that already too much?

### Server configuration
Everything is running inside a docker :whale: container. The base is [crazy-max/docker-nextcloud](https://github.com/crazy-max/docker-nextcloud), and this is [my configuration](https://github.com/patzm/dockerfiles/tree/master/nextcloud).
**Operating system**: Alpine s6:3.12
**Web server:** Nginx?
**Database:** MariaDB 10.2
**PHP version:** (e.g. 7.0)
**Nextcloud Version:** 20.0.1

#### Client configuration
**Browser:** Firefox 83
**Operating system:** Manjaro Linux, 5.8.18-1 kernel

ChristophWurst:
> I have multiple folders. The main inbox for instance has ~2k emails. Is that already too much?

Nope that should work fine.

I think it's related to https://github.com/nextcloud/mail/issues/3999.

Anything in your nextcloud.log?

patzm:
```
{"reqId":"Nvkp45virIf2ThxSxdW4","level":0,"time":"2020-11-18 13:53:51","remoteAddr":"172.29.0.3","user":"martin","app":"mail","method":"GET","url":"/apps/mail/api/messages?mailboxId=29&limit=20","message":"loading messages of folder <29>","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:83.0) Gecko/20100101 Firefox/83.0
","version":"20.0.1.1"}
```
This is the only relevant line that metnionds the `"app":"mail"`.

You are right, the issue you linked looks very similar.

This is my email account diagnosis
```
occ $ mail:account:diagnose 2
IMAP capabilities:
- AUTH
- ENABLE
- ID
- IDLE
- IMAP4REV1
- LITERAL+
- LOGIN-REFERRALS
- SASL-IR
Account has 17246 messages in 19 mailboxes
```

I then tried to sync:
```
occ $ mail:account:sync 2
[...]

PR #4644, opened 2021-03-03
Add logging to sync

miaulalala:
Add some more logging for issues like #4341 and #4095


lib/Service/Sync/ImapToDbSynchronizer.php
```diff
@@ -117,7 +117,7 @@ public function syncAccount(Account $account,
 				$logger->debug("Skipping mailbox sync for " . $mailbox->getName());
 				continue;
 			}
-
+			$logger->debug("Syncing " . $mailbox->getName());
```
ChristophWurst, line 120: for consistency I'd say let's use the id here as well instead of the name

lib/Service/Sync/ImapToDbSynchronizer.php
```diff
@@ -215,19 +215,23 @@ public function sync(Account $account,
...
+				$logger->debug("Running initial sync for " . $mailbox->getId());
 				$this->runInitialSync($account, $mailbox, $logger);
 			} else {
 				try {
+					$logger->debug("Running partial sync for " . $mailbox->getId());
 					$this->runPartialSync($account, $mailbox, $logger, $criteria, $knownUids);
 				} catch (UidValidityChangedException $e) {
 					$logger->warning('Mailbox UID validity changed. Wiping cache and performing full sync.');
 					$this->resetCache($account, $mailbox);
+					$logger->debug("Running initial sync for " . $mailbox->getId() . " after cache reset");
```
ChristophWurst, line 227: let's add the ID to the warning above

lib/Service/Sync/ImapToDbSynchronizer.php
```diff
@@ -215,19 +215,23 @@ public function sync(Account $account,
...
 					$this->runInitialSync($account, $mailbox, $logger);
 				} catch (MailboxDoesNotSupportModSequencesException $e) {
 					$logger->warning('Mailbox does not support mod-sequences error occured. Wiping cache and performing full sync.', [
 						'exception' => $e,
 					]);
 					$this->resetCache($account, $mailbox);
+					$logger->debug("Running initial sync for " . $mailbox->getId() . " after cache reset - no mod-sequences error");
```
ChristophWurst, line 234: let's add the ID to the warning above

ChristophWurst (changes requested):
Warning are logged with debug as log level, so we don't need two entries for the same paths

lib/Service/Sync/ImapToDbSynchronizer.php
ChristophWurst: it still uses the name :wink:

ChristophWurst (approved):

Issue #4412, opened 2021-01-27
Email shows unrelated email address as sender and recipients


graue70:
### Expected behavior
The mail UI shows the sender of each email.

### Actual behavior
A particular email address sent me two emails. The UI shows `no-reply@zoom.us` with the zoom logo as an additional address (besides the actual sender and me, the receiver). The UI apparently thinks that zoom is the sender because the list of emails in the middle panel shows the zoom logo (see screenshot below).

If I search the source of the email for the word 'zoom', there is no search result. If I configure the email address in thunderbird, there is no trace of 'zoom' anywhere when viewing the email, particularly not as the sender. The email is not related to zoom whatsoever.

See screenshot:
![nextcloud-mail-zoom-bug-censored](https://user-images.githubusercontent.com/23035329/105921903-d94c7100-6039-11eb-9bdd-87c91d01d370.png)
In the right panel underneath the first subject line, you see Zoom on the left side. Next to it, there is the true sender and then my own address, the receiver (both censored by me).

This might be related to #4198 but I (as in my NC user account) personally have not ever deleted an email from zoom. The subject contains `#` and `[]` and the UI shows it correctly and completely.

**Update**: The UI looks the same for a different NC user account using the same email address.

### Mail app

- **Mail app version:** 1.8.0
- **Mailserver or service:** netcup.de

### Server configuration

- **Operating system**: unknown (hosted)
- **Web server:** unknown (hosted)
- **Database:** mysql
- **PHP version:** 7.3.26
- **Nextcloud Version:** 20.0.6

### Client configuration

- **Browser:** Firefox 84.0.2
- **Operating system:** Linux Mint 19.1 and probably Windows 10

graue70:
**Update that changes the issue a lot**: I let a different user account using the same email address take the same screenshot that I took. It looks exactly the same.

ChristophWurst:
> **Update that changes the issue a lot**: I let a different user account using the same email address take the same screenshot that I took. It looks exactly the same.

I'm not sure I can follow. Is this an issue if more than one user have an account configured with an identical email address?!

The issue is kind of known, but it shows very rarely. So far we couldn't identify the pattern for when this occurs.

graue70:
> I'm not sure I can follow. Is this an issue if more than one user have an account configured with an identical email address?!

I don't know whether that is a prerequisite for the issue, but in my case that is true. Two users have the same email address configured in their NC. The issue shows for both. I'm gonna find out whether it shows for other people as well.

[...]

PR #4663, opened 2021-03-04
Delete recipient records when deleting email

miaulalala:
Should help with issues such as #4412 where orphaned records and InnoDB bugs combine to show wrong recipients.

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -406,6 +406,17 @@ class MessageMapper extends QBMapper {
 	}
 
 	public function deleteAll(Mailbox $mailbox): void {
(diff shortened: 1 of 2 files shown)


StCyr:
I don't know what this PR is expected to do 🤔

ChristophWurst:
> I don't know what this PR is expected to do thinking

Delete recipients of a message right away. On master they are expected to be cleaned up by the background clean-up. But as we found out in the ticket the database might insert new messages with the ID of the deleted messages before the cleanup happens.

st3iny (approved):
Just tested both methods and they work as intended. The failing static code check seems to be unrelated.

ChristophWurst (approved):
Code looks good!

PR #4665, opened 2021-03-04
Add tagging to messages

miaulalala:
As discussed in #25

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -41,7 +41,6 @@ use OCA\Mail\Exception\ClientException;
 use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\Http\AttachmentDownloadResponse;
 use OCA\Mail\Http\HtmlResponse;
-use OCA\Mail\Model\IMAPMessage;
 use OCA\Mail\Service\AccountService;
 use OCA\Mail\Service\ItineraryService;
 use OCP\AppFramework\Controller;
@@ -653,6 +652,68 @@ class MessagesController extends Controller {
 	 * @NoAdminRequired
 	 * @TrapError
 	 *
+	 * @param int $id
+	 * @param string $imapLabel
+	 *
+	 * @return JSONResponse
+	 *
+	 * @throws ClientException
+	 * @throws ServiceException
+	 */
+	public function setTag(int $id, string $imapLabel): JSONResponse {
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $id);
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		try {
+			$tag = $this->mailManager->getTagByImapLabel($imapLabel, $this->currentUserId);
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		$this->mailManager->tagMessage($account, $mailbox->getName(), $message, $tag, true);
+		return new JSONResponse();
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param int $id
+	 * @param string $imapLabel
+	 *
+	 * @return JSONResponse
+	 *
+	 * @throws ClientException
+	 * @throws ServiceException
+	 */
+	public function removeTag(int $id, string $imapLabel): JSONResponse {
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $id);
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		try {
+			$tag = $this->mailManager->getTagByImapLabel($imapLabel, $this->currentUserId);
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		$this->mailManager->tagMessage($account, $mailbox->getName(), $message, $tag, false);
+		return new JSONResponse();
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
 	 * @param int $accountId
 	 * @param string $folderId
 	 * @param int $id
@@ -705,9 +766,9 @@ class MessagesController extends Controller {
 	}
 
 	/**
-	 * @param array $attachment
-	 *
 	 * Determines if the content of this attachment is an image
+	 *
+	 * @param array $attachment
 	 *
 	 * @return boolean
 	 */
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -86,8 +86,9 @@ class Message extends Entity implements JsonSerializable {
 		'forwarded',
 		'junk',
 		'notjunk',
-		'important',
-		'mdnsent'
+		'mdnsent',
+		Tag::LABEL_IMPORTANT,
+		'$important' // @todo remove this when we have removed all references on IMAP to $important @link https://github.com/nextcloud/mail/issues/25
 	];
 
 	protected $uid;
@@ -124,6 +125,9 @@ class Message extends Entity implements JsonSerializable {
 
 	/** @var AddressList */
 	private $bcc;
+
+	/** @var Tag[] */
+	private $tags = [];
 
 	public function __construct() {
 		$this->from = new AddressList([]);
@@ -200,6 +204,20 @@ class Message extends Entity implements JsonSerializable {
 	}
 
 	/**
+	 * @return Tag[]
+	 */
+	public function getTags(): array {
+		return $this->tags;
+	}
+
+	/**
+	 * @param array $tags
+	 */
+	public function setTags(array $tags): void {
+		$this->tags = $tags;
+	}
+
+	/**
 	 * @return AddressList
 	 */
 	public function getCc(): AddressList {
@@ -232,14 +250,26 @@ class Message extends Entity implements JsonSerializable {
 			// Ignore
 			return;
 		}
-
-		$this->setter(
-			$this->columnToProperty("flag_$flag"),
-			[$value]
-		);
+		if ($flag === Tag::LABEL_IMPORTANT) {
+			$this->setFlagImportant($value);
+		} else {
+			$this->setter(
+				$this->columnToProperty("flag_$flag"),
+				[$value]
+			);
+		}
 	}
 
 	public function jsonSerialize() {
+		$tags = $this->getTags();
+		$indexed = array_combine(
+			array_map(
+				function (Tag $tag) {
+					return $tag->getImapLabel();
+				}, $tags),
+			$tags
+		);
+
 		return [
 			'databaseId' => $this->getId(),
 			'uid' => $this->getUid(),
@@ -253,10 +283,10 @@ class Message extends Entity implements JsonSerializable {
 				'draft' => $this->getFlagDraft(),
 				'forwarded' => $this->getFlagForwarded(),
 				'hasAttachments' => $this->getFlagAttachments() ?? false,
-				'important' => $this->getFlagImportant(),
 				'junk' => $this->getFlagJunk(),
 				'mdnsent' => $this->getFlagMdnsent(),
 			],
+			'tags' => $indexed,
 			'from' => $this->getFrom()->jsonSerialize(),
 			'to' => $this->getTo()->jsonSerialize(),
 			'cc' => $this->getCc()->jsonSerialize(),
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -25,26 +25,28 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Db;
 
+use OCP\IUser;
+use function ltrim;
 use OCA\Mail\Account;
 use OCA\Mail\Address;
-use OCA\Mail\AddressList;
-use OCA\Mail\IMAP\Threading\DatabaseMessage;
-use OCA\Mail\Service\Search\Flag;
-use OCA\Mail\Service\Search\FlagExpression;
-use OCA\Mail\Service\Search\SearchQuery;
-use OCP\AppFramework\Db\DoesNotExistException;
-use OCP\AppFramework\Db\QBMapper;
-use OCP\AppFramework\Utility\ITimeFactory;
-use OCP\DB\QueryBuilder\IQueryBuilder;
-use OCP\IDBConnection;
-use OCP\IUser;
 use RuntimeException;
-use function array_combine;
-use function array_keys;
+use OCP\IDBConnection;
 use function array_map;
 use function get_class;
-use function ltrim;
 use function mb_substr;
+use function array_keys;
+use function array_combine;
+use function array_udiff;
+use OCA\Mail\AddressList;
+use OCA\Mail\Service\Search\Flag;
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCA\Mail\Service\Search\SearchQuery;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCA\Mail\Service\Search\FlagExpression;
+use OCA\Mail\IMAP\Threading\DatabaseMessage;
+use OCA\Mail\Support\PerformanceLogger;
+use OCP\AppFramework\Db\DoesNotExistException;
 
 /**
  * @template-extends QBMapper<Message>
@@ -54,10 +56,21 @@ class MessageMapper extends QBMapper {
 	/** @var ITimeFactory */
 	private $timeFactory;
 
+	/** @var TagMapper */
+	private $tagMapper;
+
+	/** @var PerformanceLogger */
+	private $performanceLogger;
+
 	public function __construct(IDBConnection $db,
-								ITimeFactory $timeFactory) {
+								ITimeFactory $timeFactory,
+								TagMapper $tagMapper,
+								PerformanceLogger $performanceLogger
+								) {
 		parent::__construct($db, 'mail_messages');
 		$this->timeFactory = $timeFactory;
+		$this->tagMapper = $tagMapper;
+		$this->performanceLogger = $performanceLogger;
 	}
 
 	/**
@@ -119,7 +132,7 @@ class MessageMapper extends QBMapper {
 				$query->expr()->eq('m.id', $query->createNamedParameter($id, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
 			);
 
-		$results = $this->findRecipients($this->findEntities($query));
+		$results = $this->findRelatedData($this->findEntities($query));
 		if (empty($results)) {
 			throw new DoesNotExistException("Message $id does not exist");
 		}
@@ -229,7 +242,11 @@ class MessageMapper extends QBMapper {
 		$this->db->commit();
 	}
 
-	public function insertBulk(Message ...$messages): void {
+	/**
+	 * @param Message ...$messages
+	 * @return void
+	 */
+	public function insertBulk(Account $account, Message ...$messages): void {
 		$this->db->beginTransaction();
 
 		$qb1 = $this->db->getQueryBuilder();
@@ -309,18 +326,25 @@ class MessageMapper extends QBMapper {
 					$qb2->execute();
 				}
 			}
+			foreach ($message->getTags() as $tag) {
+				$this->tagMapper->tagMessage($tag, $message->getMessageId(), $account->getUserId());
+			}
 		}
 
 		$this->db->commit();
 	}
 
 	/**
-	 * @param Message ...$messages
-	 *
+	 * @param Account $account
+	 * @param Message[] $messages
 	 * @return Message[]
 	 */
-	public function updateBulk(Message ...$messages): array {
+	public function updateBulk(Account $account, Message ...$messages): array {
 		$this->db->beginTransaction();
+
+		$perf = $this->performanceLogger->start(
+			'partial sync ' . $account->getId() . ':' . $account->getName()
+		);
 
 		$query = $this->db->getQueryBuilder();
 		$query->update($this->getTableName())
@@ -339,7 +363,9 @@ class MessageMapper extends QBMapper {
 				$query->expr()->eq('uid', $query->createParameter('uid')),
 				$query->expr()->eq('mailbox_id', $query->createParameter('mailbox_id'))
 			));
-
+		// get all tags before the loop and create a mapping [message_id => [tag,...]]
+		$tags = $this->tagMapper->getAllTagsForMessages($messages);
+		$perf->step("Selected Tags for all messages");
 		foreach ($messages as $message) {
 			if (empty($message->getUpdatedFields())) {
 				// Micro optimization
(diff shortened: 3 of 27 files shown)


lib/Db/Tag.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
```
ChristophWurst, line 6: It wasn't me

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,188 @@
...
+					break;
+				case $i === 5:
+					$tag->setDisplayName('Later');
+					$tag->setColor('#800080');
+					$tag->setIsSystemFlag(1);
+					break;
+			}
+			try {
+				$this->insert($tag);
+			} catch (\Doctrine\DBAL\Exception\UniqueConstraintViolationException $e) {
```
ChristophWurst, line 182: I don't think you can actually catch the Doctrine exception. Nextcloud will wrap that exception with its own. We can catch the \OCP\Db\Exception instead and check the reason

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,188 @@
...
+				$this->logger->debug($e->getMessage());
```
ChristophWurst, line 183: when logging an exception pass the object to the logger, so the trace is included in the log line.

```suggestion
				$this->logger->debug($e->getMessage(), ['exception' => $e]);
```

lib/Migration/Version1100Date20210304143008.php
```diff
@@ -0,0 +1,115 @@
...
+
+use Closure;
+use OCA\Mail\Db\MailAccountMapper;
+use OCA\Mail\Db\TagMapper;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step.
```
ChristophWurst, line 15: you can drop this note

src/store/actions.js
```diff
@@ -556,9 +558,6 @@ export default {
 		} finally {
 			showNewMessagesNotification(newMessages)
 		}
-	},
-	toggleEnvelopeFlagged({ commit, getters }, envelope) {
```
ChristophWurst, line 560: I think those three lines should be restored

lib/Db/Tag.php
miaulalala: That's copypasta for you ;)

lib/Db/TagMapper.php
miaulalala: I tried the exception but the UniqueConstraintViolationException doesn't seem to get wrapped.

src/store/actions.js
miaulalala: 😬

lib/Db/TagMapper.php
```diff
@@ -130,10 +142,37 @@ public function untagMessage(Tag $tag, string $messageId) {
 		$qb->execute();
 	}
 
+	/**
+	 * @param array $messages
+	 * @return array [[<jshdkfjhdsjhfdskjhfdkjshjkdshfjkdsh@test.com>] => [1, 2]]
```
miaulalala, line 147: no longer accurate

ChristophWurst:
- [ ] Bump the version in appinfo/info.xml from alpha.1 to alpha.2. This will then trigger Nextcloud to check for new migration files and devs pulling master after we merge this PR will be migrated.

lib/Service/SetupService.php
```diff
@@ -139,6 +150,8 @@ public function createNewAccount($accountName, $emailAddress, $imapHost, $imapPo
 		$this->accountService->save($newAccount);
 		$this->logger->debug("account created " . $newAccount->getId());
 
+		$this->tagMapper->createDefaultTags($newAccount->getMailAccount());
```
ChristophWurst, line 153:
```suggestion
		$this->tagMapper->createDefaultTags($newAccount);
```

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @link https://github.com/nextcloud/mail/issues/25
+	 */
+	public function createDefaultTags(MailAccount $account) {
+		for ($i = 1; $i < 6; $i++) {
+			$tag = new \OCA\Mail\Db\Tag();
+			$tag->setImapLabel('$label' . $i);
+			$tag->setUserId($account->getUserId());
+			switch ($i) {
+				case $i === 1:
+					$tag->setDisplayName('Important');
```
ChristophWurst, line 190: We should localize the labels here

ChristophWurst (dismissed):
Some little quirks I found. Otherwise reading the tags from Thunderbird into our DB works as expected :)

I noticed this as well:

Bildschirmfoto von 2021-03-16 09-51-09

(accidentally added the same account multiple times)

Is that a problem? Checking before insert is likely expensive, so we might accept the cost of duplication.

ChristophWurst:
I'm not a fan of the naming *system tags* because they have an existing meaning in Nextcloud for file tagging. How about we rename to *default tags*?

ChristophWurst:
- [ ] BUG when marking messages as important it should add the tag `$label1` now

instead I currently see this being inserted 
Bildschirmfoto von 2021-03-16 10-14-49

miaulalala:
> I'm not a fan of the naming _system tags_ because they have an existing meaning in Nextcloud for file tagging. How about we rename to _default tags_?

I agree, done


> Some little quirks I found. Otherwise reading the tags from Thunderbird into our DB works as expected :)
> 
> I noticed this as well:
> 
> Bildschirmfoto von 2021-03-16 09-51-09
> 
> (accidentally added the same account multiple times)
> 
> Is that a problem? Checking before insert is likely expensive, so we might accept the cost of duplication.

That's one issue with cleanup - not deleting the records when removing the account will duplicate the entries on the initial sync when readding. Duplicates are fine though. When removing the tags it should clean up all orphaned records with the same tag, as well. If this is an account for an existing user, the tags aren't newly generated in `oc_mail_tags` so they should have the same database id for the user.

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,233 @@
...
+	 * @return Entity
+	 *
+	 * @throws DoesNotExistException
+	 */
+	public function getTagByImapLabel(string $imapLabel, string $userId): Entity {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->like('imap_label', $qb->createNamedParameter($imapLabel)),
```
ChristophWurst, line 67:
```suggestion
				$qb->expr()->eq('imap_label', $qb->createNamedParameter($imapLabel)),
```

right?

lib/Model/IMAPMessage.php
```diff
@@ -673,7 +674,13 @@ public function setInReplyTo(string $message) {
...
 	}
 
-	public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
+	/**
+	 * Cast all values from an IMAP message into the correct DB format
+	 *
+	 * @param integer $mailboxId
+	 * @return \OCA\Mail\Db\Message
+	 */
+	public function toDbMessage(int $mailboxId, MailAccount $account): \OCA\Mail\Db\Message {
```
ChristophWurst, line 683: import the `Message` class

lib/Model/IMAPMessage.php
```diff
@@ -707,10 +714,57 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
...
+		$allowed = [
+			Horde_Imap_Client::FLAG_SEEN,
+			Horde_Imap_Client::FLAG_ANSWERED,
+			Horde_Imap_Client::FLAG_FLAGGED,
+			Horde_Imap_Client::FLAG_DELETED,
+			Horde_Imap_Client::FLAG_DRAFT,
+			Horde_Imap_Client::FLAG_RECENT,
+			Horde_Imap_Client::FLAG_JUNK,
+			'junk',
+			'$junk',
```
ChristophWurst, line 731: this is identical to `Horde_Imap_Client::FLAG_JUNK`

lib/Model/IMAPMessage.php
```diff
@@ -707,10 +714,57 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
...
+			'$seen',
```
ChristophWurst, line 732: is this really an IMAP standard flag?

lib/Model/IMAPMessage.php
```diff
@@ -707,10 +714,57 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
...
+	 *
+	 * @link https://github.com/nextcloud/mail/issues/25
+	 *
+	 * @param array $tags
+	 * @return array
+	 */
+	private function generateTagEntites(array $tags, string $userId): array {
+		$t = [];
+		foreach ($tags as $keyword) {
+			$tag = new \OCA\Mail\Db\Tag();
```
ChristophWurst, line 762: import

lib/Service/MailManager.php
```diff
@@ -419,6 +427,50 @@ public function flagMessage(Account $account, string $mailbox, int $uid, string
...
+				}
+			} catch (Horde_Imap_Client_Exception $e) {
+				throw new ServiceException(
+					"Could not set message keyword on IMAP: " . $e->getMessage(),
+					(int) $e->getCode(),
+					$e
+				);
+			}
+		}
+		if ($value === true) {
```
ChristophWurst, line 467:
```suggestion
		if ($value) {
```

it's a bool as per typehint

ChristophWurst (changes requested):

lib/Model/IMAPMessage.php
miaulalala: no but it's gmail's $seen flag - do we want to make a tag out of it?

lib/Db/TagMapper.php
ChristophWurst: https://github.com/nextcloud/server/pull/26182

ChristophWurst:
CI is partly failing because of https://github.com/codecov/codecov-action/issues/234

lib/Controller/MessagesController.php
```diff
@@ -132,7 +131,8 @@ public function __construct(string $appName,
 								IURLGenerator $urlGenerator,
 								ContentSecurityPolicyNonceManager $nonceManager,
 								ITrustedSenderService $trustedSenderService,
-								IMailTransmission $mailTransmission) {
+								IMailTransmission $mailTransmission,
```
ChristophWurst, line 134: let's revert those two lines. the trailing comma is problematic on php7.2 (an old bug installed new Mail in the past, hence we try to keep php7.2 for a little longer)

lib/Db/Message.php
```diff
@@ -199,6 +203,20 @@ public function setTo(AddressList $to): void {
...
 
+	/**
+	 * @return array
+	 */
+	public function getTags(): array {
+		return $this->tags;
+	}
+
+	/**
+	 * @param array $tags
```
ChristophWurst, line 214: more specific type please

lib/Db/Message.php
```diff
@@ -199,6 +203,20 @@ public function setTo(AddressList $to): void {
 		$this->to = $to;
 	}
 
+	/**
+	 * @return array
```
ChristophWurst, line 207: more specific type please

lib/Db/Message.php
```diff
@@ -232,11 +250,14 @@ public function setFlag(string $flag, bool $value = true) {
 			// Ignore
 			return;
 		}
-
-		$this->setter(
-			$this->columnToProperty("flag_$flag"),
-			[$value]
-		);
+		if ($flag === '$label1') {
```
ChristophWurst, line 253: how about a `\OCA\Mail\Db\Tag::LABEL_IMPORTANT` for better readability and lower chances of typos?

lib/Db/MessageMapper.php
```diff
@@ -309,19 +326,26 @@ public function insertBulk(Message ...$messages): void {
...
 		$this->db->commit();
 	}
 
 	/**
-	 * @param Message ...$messages
-	 *
-	 * @return Message[]
+	 * @param Account $account
+	 * @param Message[] $messages
+	 * @return array
```
ChristophWurst, line 340:
```suggestion
	 * @return Message[]
```

as it was before

lib/Db/MessageMapper.php
```diff
@@ -360,13 +386,59 @@ public function updateBulk(Message ...$messages): array {
...
+				continue;
+			}
+
+			$toRemove = array_udiff($dbTags, $imapTags, function (Tag $a, Tag $b) {
+				return strcmp($a->getImapLabel(), $b->getImapLabel());
+			});
+			foreach ($toRemove as $tag) {
+				//add logging
+				$this->tagMapper->untagMessage($tag, $message->getMessageId());
+				$perf->step("Untagged message");
```
ChristophWurst, line 420: I'd move that outside the loop so it only gets logged once

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return Entity
+	 *
+	 * @throws DoesNotExistException
+	 */
+	public function getTag(int $id): Entity {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)));
+		$result = $this->findEntity($qb);
```
ChristophWurst, line 107: inline `$result`

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,234 @@
...
+				$qb->expr()->in('imap_message_id', $qb->createNamedParameter($ids, IQueryBuilder::PARAM_STR_ARRAY))
+			);
+		$qb = $qb->execute();
+		$queryResult = $qb->fetchAll();
+		if (empty($queryResult)) {
+			return [];
+		}
+
+		foreach ($queryResult as $qr) {
+			$result[$qr['imap_message_id']][] = $this->getTag((int)$qr['tag_id']);
```
ChristophWurst, line 174: `$result` is uninitalized and so is the index `$qr['imap_message_id']` when you first access it

lib/Service/MailManager.php
```diff
@@ -23,34 +23,36 @@
...
 use OCA\Mail\Db\Mailbox;
-use OCA\Mail\Db\MailboxMapper;
 use OCA\Mail\Db\Message;
-use OCA\Mail\Db\MessageMapper as DbMessageMapper;
-use OCA\Mail\Events\BeforeMessageDeletedEvent;
+use function array_values;
+use OCA\Mail\Db\TagMapper;
+use Psr\Log\LoggerInterface;
+use Horde_Imap_Client_Socket;
+use OCA\Mail\Db\MailboxMapper;
```
ChristophWurst, line 37: can we keep the imports sorted alphabetically?

lib/Service/Search/Flag.php
```diff
@@ -34,7 +34,7 @@ class Flag {
 	public const ANSWERED = Horde_Imap_Client::FLAG_ANSWERED;
 	public const SEEN = Horde_Imap_Client::FLAG_SEEN;
 	public const FLAGGED = Horde_Imap_Client::FLAG_FLAGGED;
-	public const IMPORTANT = '\\important';
+	public const IMPORTANT = '$label1';
```
ChristophWurst, line 37: mark as deprecated now that we know *important* is rather a *tag* than a *flag*

lib/Service/SetupService.php
```diff
@@ -56,28 +57,36 @@ class SetupService {
...
+
 	public function __construct(AutoConfig $autoConfig,
 								AccountService $accountService,
 								ICrypto $crypto,
 								SmtpClientFactory $smtpClientFactory,
 								IMAPClientFactory $imapClientFactory,
-								LoggerInterface $logger) {
+								LoggerInterface $logger,
+								TagMapper $tagMapper
+								) {
```
ChristophWurst, line 73: closing bracket should move  to previous line

ChristophWurst (changes requested):

lib/Service/MailManager.php
miaulalala: I'm not sure why my IDE is doing that, I'm sorry! I'll see which extension is causing this.

ChristophWurst:
- [ ] BUG message tags are always on the front-end

in `\OCA\Mail\Db\Message::jsonSerialize` you use `\OCA\Mail\Db\Message::getTags` but the tags prop is only set when when read messages from imap.

We need something similar to `\OCA\Mail\Db\MessageMapper::findRecipients` in the message mapper that loads each message's tags.

ChristophWurst:
> miaulalala force-pushed the enhancement/tagging-messages branch

unless there is a change on master that we need here I would recommend to only do fixups and wait with the rebase until we have positive review & CI result. That will make it easier for me and others to check out the branch and pull the latest changes.

ChristophWurst:
- [ ] BUG our old `$important` tag in inserted as a new tag, but we should just map it to `$label1`

Bildschirmfoto von 2021-03-18 10-30-17

ChristophWurst:
- [ ] BUG I could have happened because of the way I reset my db before I re-migrated to your new schema changes here but this shouldn't happen:

Bildschirmfoto von 2021-03-22 12-10-50

ChristophWurst:
- [ ] BUG `message.tags` is not an indexed object, but used as one

Bildschirmfoto von 2021-03-22 12-13-07

we have code that does `envelope.tags.$label1` but this only works it `envelope.tags` is an object with a key `$label1`. when we load messages/envelopes they have an *array* of tags, however. So we have to use something like a `envelope.tags.some(t => t.keyword === '$label1')` with a O(n) or we store the tags as a simpler object of keywords, like `{ '$label1': { id: 123, keyword: '$label1', displayName: 'Important' }` and then we can (again) use `envelope.tags.$label1` with O(1).

To sum up we should index tags by their keyword.

lib/Db/Tag.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @return array
+	 */
+	public function jsonSerialize() {
+		return [
+			'databaseId' => $this->getId(),
+			'userId' => $this->getUserId(),
+			'displayName' => $this->getDisplayName(),
+			'imapLabel' => $this->getImapLabel(),
+			'color' => $this->getColor(),
+			'isDefaultTag' => $this->getisDefaultTag(),
```
ChristophWurst, line 65:
```suggestion
			'isDefaultTag' => $this->getIsDefaultTag(),
```

ChristophWurst (changes requested):

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+	 * @param array $messages
+	 * @return array
+	 */
+	public function getAllTagsForMessages(array $messages):array {
+		$ids = array_map(function (Message $message) {
+			return $message->getMessageId();
+		}, $messages);
+
+		$qb = $this->db->getQueryBuilder();
+		$ids = $qb->select('mt.*')
```
ChristophWurst, line 175: `$ids` is never used

ChristophWurst: did you mean to call `$ids->execute()` below? let's rename to `selectIdsQuery`

miaulalala: Line 178

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+					$qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)),
+					$qb->expr()->eq('user_id', $qb->createNamedParameter($userId))
+				);
+		return $this->findEntity($qb);
+	}
+
+	/**
+	 * @param integer $id
+	 * @param string $userId
+	 * @return array
```
ChristophWurst, line 94:
```suggestion
	 * @return Tag[]
```

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+			->where(
+					$qb->expr()->eq('user_id', $qb->createNamedParameter($userId))
+				);
+		return $this->findEntities($qb);
+	}
+
+
+	/**
+	 * @param integer $id
+	 * @param string $userId
```
ChristophWurst, line 111: does not exist

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+			->from($this->getTableName())
+			->where(
+					$qb->expr()->eq('user_id', $qb->createNamedParameter($userId))
+				);
+		return $this->findEntities($qb);
+	}
+
+
+	/**
+	 * @param integer $id
```
ChristophWurst, line 110:
```suggestion
	 * @param int $id
```

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+
+	/**
+	 * Tag a message in the DB
+	 *
+	 * To tag (flag) a message on IMAP, @see OCA\Mail\Service\MailManager::tagMessage
+	 *
+	 * @param Tag $tag
+	 * @param string $messageId
+	 * @param string $userId
+	 * @return void
```
ChristophWurst, line 132: use the php `: void` notation instead

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+		$qb->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)));
+		return $this->findEntity($qb);
+	}
+
+	/**
+	 * Tag a message in the DB
+	 *
+	 * To tag (flag) a message on IMAP, @see OCA\Mail\Service\MailManager::tagMessage
```
ChristophWurst, line 127:
```suggestion
	 * To tag (flag) a message on IMAP, @see \OCA\Mail\Service\MailManager::tagMessage
```

then IDEs like phpstorm will let us jump directly to the method

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+		$qb->setValue('tag_id', $qb->createNamedParameter($tag->getId(), IQueryBuilder::PARAM_INT));
+		$qb->execute();
+	}
+
+	/**
+	 * Remove a tag from a DB message
+	 *
+	 * @param Tag $tag
+	 * @param string $messageId
+	 * @return void
```
ChristophWurst, line 155: `: void` on the method

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete('mail_message_tags')
+			->where($qb->expr()->eq('imap_message_id', $qb->createNamedParameter($messageId)))
+			->where($qb->expr()->eq('tag_id', $qb->createNamedParameter($tag->getId())));
+		$qb->execute();
+	}
+
+	/**
+	 * @param array $messages
+	 * @return array
```
ChristophWurst, line 167:
```suggestion
	 * @return Tag[][]
```

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+	public function untagMessage(Tag $tag, string $messageId) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete('mail_message_tags')
+			->where($qb->expr()->eq('imap_message_id', $qb->createNamedParameter($messageId)))
+			->where($qb->expr()->eq('tag_id', $qb->createNamedParameter($tag->getId())));
+		$qb->execute();
+	}
+
+	/**
+	 * @param array $messages
```
ChristophWurst, line 166:
```suggestion
	 * @param Message[] $messages
```

lib/Db/TagMapper.php
```diff
@@ -0,0 +1,250 @@
...
+	 * Create some default system tags
+	 *
+	 * This is designed to be similar to Thunderbird's email tags
+	 * $label1 to $label5 with the according states and colours
+	 *
+	 * <i>The array_udiff can be removed and the insert warpped in
+	 * an exception as soon as NC20 is not supported any more</i>
+	 *
+	 * @param MailAccount $account
+	 * @return void
```
ChristophWurst, line 203: move to the method signature

lib/Migration/Version1100Date20210304143008.php
```diff
@@ -0,0 +1,121 @@
...
+		return $schema;
+	}
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		// repair step!
```
ChristophWurst, line 115: I think you can drop this, we have a ticket for the todo

lib/Model/IMAPMessage.php
```diff
@@ -707,10 +716,58 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
...
+	/**
+	 * Build tag entities from keywords sent by IMAP
+	 *
+	 * Will use IMAP keyword '$xxx' to create a value for
+	 * display_name like 'xxx'
+	 *
+	 * @link https://github.com/nextcloud/mail/issues/25
+	 *
+	 * @param array $tags
+	 * @return array
```
ChristophWurst, line 756:
```suggestion
	 * @return Tag[]
```

lib/Model/IMAPMessage.php
```diff
@@ -707,10 +716,58 @@ public function toDbMessage(int $mailboxId): \OCA\Mail\Db\Message {
...
+
+	/**
+	 * Build tag entities from keywords sent by IMAP
+	 *
+	 * Will use IMAP keyword '$xxx' to create a value for
+	 * display_name like 'xxx'
+	 *
+	 * @link https://github.com/nextcloud/mail/issues/25
+	 *
+	 * @param array $tags
```
ChristophWurst, line 755:
```suggestion
	 * @param string[] $tags
```

ChristophWurst (changes requested):
Proper types and their documentation help developers, IDEs and static analysis :)

lib/Db/TagMapper.php
miaulalala: ?

lib/Db/TagMapper.php
miaulalala: Nah it's fine, it was just a superfluous assignment.

ChristophWurst (approved):
Tested and works :shipit: 

Now you may do the final rebase & squash

Issue #4656, opened 2021-03-04
Error 500 when forwarding message with image inside


grubshka:
Hello,

If I try to forward a message containing an image.
I send me a message in a mail client, inserting an image from my computer in the body.
The image is sent as an attachment.
In Nextcloud mail app, I try to forward the message.

### Expected behavior
Message should be sent

### Actual behavior
Error 500.

Here's the log from the nextcloud log file : 

[...]

PR #4667, opened 2021-03-05
Fix to error forward with inline image

StCyr:
fixes https://github.com/nextcloud/mail/issues/4656

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -463,7 +463,9 @@ export default {
 
 		// Add attachments in case of forward
 		if (this.forwardFrom?.attachments !== undefined) {
-			this.forwardFrom.attachments.forEach(att => {
+			// We filter attachments with contentDisposition header set to 'inline' as theyr are rendered
+			// in the HTML body
+			this.forwardFrom.attachments.filter(att => att.disposition !== 'inline').forEach(att => {
 				this.attachments.push({
 					fileName: att.fileName,
 					displayName: trimStart('/', att.fileName),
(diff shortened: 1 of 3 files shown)


src/components/Composer.vue
```diff
@@ -463,7 +463,7 @@ export default {
 
 		// Add attachments in case of forward
```
ChristophWurst, line 464: please add a note about why we exclude inline attachments here

src/components/MessageAttachments.vue
```diff
@@ -22,17 +22,18 @@
...
-				:id="attachment.id"
-				:key="attachment.id"
-				:file-name="attachment.fileName"
-				:size="attachment.size"
-				:url="attachment.downloadUrl"
-				:is-image="attachment.isImage"
-				:is-calendar-event="attachment.isCalendarEvent"
-				:mime="attachment.mime"
-				:mime-url="attachment.mimeUrl" />
+			<div v-for="attachment in attachments" :key="attachment.id">
```
ChristophWurst, line 25: is this really necessary?

ChristophWurst (changes requested):

src/components/Composer.vue
StCyr: done

src/components/MessageAttachments.vue
StCyr: No indeed.

StCyr: changed

ChristophWurst (approved):
Please squash :)

PR #4692, opened 2021-03-07
Do explict security scan of psalm

rullzer:


ChristophWurst (changes requested):
Mail doesn't work with composer v2 yet because of Horde

PR #4703, opened 2021-03-08
Update npm lock for v7

ChristophWurst:
Npm 7 comes with a new lock file format. This upgrades the package-lock.json file.

This format is backwards compatible to v6: https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,20543 @@
 {
   "name": "nextcloud-mail",
   "version": "1.9.0",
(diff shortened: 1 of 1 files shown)


StCyr:
Funny, it fails with the same error as in https://github.com/nextcloud/mail/pull/4590

ChristophWurst:
obsolete

PR #4706, opened 2021-03-08
Migrate from davclient.js to webdav

ChristophWurst:
The davclient.js lib isn't so well supported anymore and the webdav
package is also used for file DAV things. This patch makes the DAV
client a bit more general purpose and moves the calendar listing code
over to this client.

How to test (that this doesn't introduce a regression)
* Find a sample .ics file (or write one yourself)
* Send it to you via email
* Open that email in Mail
* Click the + icon next to the attachment -> a list of writable calendars should show (and no birthday calendars or similar)
* By clicking one of the calendars, the event in the .ics is written to that selected calendar

This makes https://github.com/nextcloud/mail/pull/4703 obsolete as my npm just updated the lock file when I dropped davclient.js.

### DIFF
--- a/src/dav/client.js
+++ b/src/dav/client.js
@@ -1,7 +1,7 @@
 /*
- * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
  *
- * @author 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @author 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
  *
  * @license GNU AGPL version 3 or any later version
  *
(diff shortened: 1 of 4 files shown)


src/dav/client.js
```diff
@@ -0,0 +1,39 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as webdav from 'webdav'
+import axios from '@nextcloud/axios'
+import memoize from 'lodash/fp/memoize'
```
skjnldsv, line 24: still using lodash? :grin:

ChristophWurst: why not?

ChristophWurst: the fp stuff is actually decent

Issue #3608, opened 2020-09-18
Too long email address breaks sync


ronnyjohn:
The initial sync encountered an error and i was stuck with the loading spinner in the app. The problematic mail was a spam mail that had a really long sender address:

[...]

PR #4708, opened 2021-03-08
Truncate email addresses that are more than 244 characters long

StCyr:
Fixes https://github.com/nextcloud/mail/issues/3608

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -341,7 +341,7 @@ class MessageMapper extends QBMapper {
 						$qb2->setParameter('message_id', $messageId, IQueryBuilder::PARAM_INT);
 						$qb2->setParameter('type', $type, IQueryBuilder::PARAM_INT);
 						$qb2->setParameter('label', mb_strcut($recipient->getLabel(), 0, 255), IQueryBuilder::PARAM_STR);
-						$qb2->setParameter('email', $recipient->getEmail(), IQueryBuilder::PARAM_STR);
+						$qb2->setParameter('email', mb_strcut($recipient->getEmail(), 0, 255), IQueryBuilder::PARAM_STR);
 
 						$qb2->executeStatement();
 					}


miaulalala (commented):
Code is fine - but what's the consequences for replying in the in the UI?

StCyr:
> 
> 
> Code is fine - but what's the consequences for replying in the in the UI?

Yes, that was one of my concerns: Replying just won't work.

I've added the "<truncated>" substring to make it more clear to the user that the email address is not usable anymore. 

But, let's be honest, emails coming from such addresses are most probably spam

miaulalala:
> > Code is fine - but what's the consequences for replying in the in the UI?
> 
> Yes, that was one of my concerns: Replying just won't work.
> 
> I've added the "" substring to make it more clear to the user that the email address is not usable anymore.
> 
> But, let's be honest, emails coming from such addresses are most probably spam

True. Is the "" substring clearly identifiable by the user?

StCyr:
> True. Is the "" substring clearly identifiable by the user?

(fun "<truncated>" ("< t r u n c a t ed >") seems to have been truncated by Github)

Not much indeed but we could use it to show it differently in the frontend (in red? with an exclamation mark?)

miaulalala:
The time for a migration would be now.

I like the idea of having an error state for mails, this could be very useful.

ChristophWurst:
- [ ] Blocked by https://github.com/nextcloud/mail/issues/4735

lib/Db/MessageMapper.php
```diff
@@ -341,7 +341,7 @@ public function insertBulk(Account $account, Message ...$messages): void {
 						$qb2->setParameter('message_id', $messageId, IQueryBuilder::PARAM_INT);
 						$qb2->setParameter('type', $type, IQueryBuilder::PARAM_INT);
 						$qb2->setParameter('label', mb_strcut($recipient->getLabel(), 0, 255), IQueryBuilder::PARAM_STR);
-						$qb2->setParameter('email', $recipient->getEmail(), IQueryBuilder::PARAM_STR);
+						$qb2->setParameter('email', mb_substr($recipient->getEmail(),0 , 255), IQueryBuilder::PARAM_STR);
```
ChristophWurst, line 344: adjusted to drop the `<truncated>` because it would be tricky with translations

ChristophWurst (approved):

Issue #1660, opened 2019-04-09
Allow switching account&signature


ChristophWurst:
### Feature Request

With the implementation at https://github.com/nextcloud/mail/pull/1657, signatures are added to the prefilled message body. Hence a change of the alias/account won't update the signature.

#### Summary

n/a

dumblob:
Raising my hand for this as otherwise having multiple email accounts with corresponding signatures doesn't work and effectively only one signature (of the default/top-most account?) can be used.

ChristophWurst:
@dumblob please don't comment on all the tickets but leave your 👍 on the original request

PR #4715, opened 2021-03-09
Add ckeditor plugin for signature element

kesselb:
**Questions**

- [x] Append `--` as divider by default (user has no choice, rainloop or roundcube does not enforce a divider)
- [x] To much events / event structure
- [x] Where to store the ckeditor plugin
- [x] On account switch a new draft is created (we may delete it, might be complicate)
- [x] We do not update the route on account switch (internal state and active state for menu, we may remove the possibility to switch account in compose view for now)
- [x] SetAlias only set's the editor mode if the previous alias is -1 (out of scope for this issue)

**Todo**

- [x] alias is undefined on first account switch
- [x] something weird is happening the first time an account is switched (v-model = `:value` + `@input`, turn's out that `@select` is executed before the value is updated. now the value is updated by the event listener).
- [x] remove ckeditor inspector
- [x] don't append empty signature (null)
- [x] quoting of text in reply does not work anymore because ready event triggers before two way binding.

Fixes https://github.com/nextcloud/mail/issues/1660

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@ckeditor/ckeditor5-basic-styles": "~26.0.0",
         "@ckeditor/ckeditor5-block-quote": "~26.0.0",
         "@ckeditor/ckeditor5-build-balloon": "~26.0.0",
+        "@ckeditor/ckeditor5-core": "^26.0.0",
         "@ckeditor/ckeditor5-dev-utils": "24.4.1",
         "@ckeditor/ckeditor5-dev-webpack-plugin": "^24.4.1",
         "@ckeditor/ckeditor5-editor-balloon": "~26.0.0",
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "@ckeditor/ckeditor5-basic-styles": "~26.0.0",
     "@ckeditor/ckeditor5-block-quote": "~26.0.0",
     "@ckeditor/ckeditor5-build-balloon": "~26.0.0",
+    "@ckeditor/ckeditor5-core": "^26.0.0",
     "@ckeditor/ckeditor5-dev-utils": "24.4.1",
     "@ckeditor/ckeditor5-dev-webpack-plugin": "^24.4.1",
     "@ckeditor/ckeditor5-editor-balloon": "~26.0.0",
--- a/src/ckeditor/signature/InsertSignatureCommand.js
+++ b/src/ckeditor/signature/InsertSignatureCommand.js
@@ -1,0 +1,82 @@
+/**
+ * @author Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+import Command from '@ckeditor/ckeditor5-core/src/command'
+
+export default class InsertSignatureCommand extends Command {
+
+	removeSignatureElement(editor, writer) {
+		// Create a range spanning over the entire root content:
+		const range = editor.model.createRangeIn(
+			editor.model.document.getRoot()
+		)
+
+		// Iterate over all items in this range:
+		for (const value of range.getWalker({ shallow: true })) {
+			if (value.item.is('element') && value.item.name === 'signature') {
+				writer.remove(value.item)
+			}
+		}
+	}
+
+	insertSignatureElement(editor, writer, value) {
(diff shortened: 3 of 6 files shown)


ChristophWurst:
> Append `--` as divider by default (user has no choice, rainloop or roundcube does not enforce a divider)

I'd say yes. Don't we do that already, at least for plain text messages?

> To much events / event structure

Is that a question? I think the structure is fine.

> Where to store the ckeditor plugin

`src/ckeditor`? we have other dedicated directories like `src/imap` and `src/dav`.

> On account switch a new draft is created (we may delete it, might be complicate)

Known https://github.com/nextcloud/mail/issues/2511 and can be addressed separately.

>  We do not update the route on account switch (internal state and active state for menu, we may remove the possibility to switch account in compose view for now)

The URL is something like `/apps/mail/box/priority/thread/new`, so it contains the *current mailbox* (`priority` in this case). Where's the part of the account that you want to update?

kesselb:
> > Append `--` as divider by default (user has no choice, rainloop or roundcube does not enforce a divider)
> 
> I'd say yes. Don't we do that already, at least for plain text messages?

Fine by me. I just wanted to point out that other webmail's giving the user a choice. 

> 
> > Where to store the ckeditor plugin
> 
> `src/ckeditor`? we have other dedicated directories like `src/imap` and `src/dav`.

`src/ckeditor-signature` right now. Is that okay? Or do you prefer `src/ckeditor/signature`? We don't have other plugins yet so `src/ckeditor` may work too. 

> > We do not update the route on account switch (internal state and active state for menu, we may remove the possibility to switch account in compose view for now)
> 
> The URL is something like `/apps/mail/box/priority/thread/new`, so it contains the _current mailbox_ (`priority` in this case). Where's the part of the account that you want to update?

OK. I had in mind to replace the mailbox id in the route.

ChristophWurst:
> `src/ckeditor-signature` right now. Is that okay? Or do you prefer `src/ckeditor/signature`? We don't have other plugins yet so `src/ckeditor` may work too.

I would give it its own directory just to not *pollute* `src` too much with this editor-specific module :)

src/ckeditor-signature/InsertSignatureCommand.js
```diff
@@ -0,0 +1,74 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+import Command from '@ckeditor/ckeditor5-core/src/command'
```
ChristophWurst, line 20: do we have to install `@ckeditor/ckeditor5-core` explicitly?

st3iny (commented):
Null gets appended when replying to one/all and no signature is configured.

ckeditor-sig-error

src/ckeditor-signature/InsertSignatureCommand.js
kesselb: I don't think so. `@ckeditor/ckeditor5-editor-balloon` has `@ckeditor/ckeditor5-core` as dependency. But eslint does complain about it.

ChristophWurst: I think eslint is not wrong. npm resolves a deep hierarchy of dependencies, not a flat one. we should depend on any import explicitly.

kesselb: Added @ckeditor/ckeditor5-core as dependency

ChristophWurst:
- [ ] BUG starting the composer for a new message with no signature set for the accounts inserts two line breaks as far as I can tell, hence the placeholder does not show.

Expected:
Bildschirmfoto von 2021-03-18 10-54-05

Actual:
Bildschirmfoto von 2021-03-18 10-54-14

ChristophWurst:
- [ ] BUG if I add an alias I can no longer select the account's main email address as sender

ChristophWurst:
- [ ] BUG switching from one accoutn with a signature to another without causes an error and the UI doesn't update

Bildschirmfoto von 2021-03-18 11-08-11

ChristophWurst:
:heavy_check_mark: works very well if two accounts have a signature set :)

kesselb:
> BUG starting the composer for a new message with no signature set for the accounts inserts two line breaks as far as I can tell, hence the placeholder does not show.

I don't know how to fix it. We need an initial value for the bodyVal to trigger the onInputEvent by ckeditor. The editor ready event is to early to run our signature logic. An event between editor ready and on input to signal a working two way binding does not exist. If we set the signature on ready before the two way binding a previous message (draft) is not shown.

kesselb:
> BUG if I add an alias I can no longer select the account's main email address as sender

Mind to share a short screencast?

ChristophWurst:
> Mind to share a short screencast?

Nevermind, this actually works. I was confused because your current choice is not in the list. Hence I somehow thought the main email is missing, but it's not.

ChristophWurst:
- [ ] BUG (but likely also on master): writing a new message after adding a (secondary) account without a page reloaded causes an error because the account does not have any aliases initialized

Bildschirmfoto von 2021-03-22 19-34-42

ChristophWurst:
now if I have to accounts with signatures it doesn't actually update the signature anymore when I switch between accounts?!

ChristophWurst (approved):
Tested and works perfectly :)

PR #4716, opened 2021-03-10
Generate an IMAP message ID when it's missing

ChristophWurst:
There should be one in theory https://stackoverflow.com/a/8513707/2239067

But as we've seen in the wild there are cases where we don't get one. Other IMAP clients like Evolution also generate their own ID if it's missing from IMAP.

I chose to have a distinct prefix that expresses that the ID was generated by us. This could be helpful for debugging in the future.

There is also a repair step that fixes messages during upgrade or whenever `occ maintenance:repair` is run by admins.

Bildschirmfoto von 2021-03-10 10-00-15

(my dev instance had LOTS of machine-generated emails without a message-id. production instances do not show such high numbers according to my tests)

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -38,6 +38,7 @@
 	</dependencies>
 	<repair-steps>
 		<post-migration>
+			<step>OCA\Mail\Migration\AddMissingMessageIds</step>
 			<step>OCA\Mail\Migration\FixCollectedAddresses</step>
 			<step>OCA\Mail\Migration\FixBackgroundJobs</step>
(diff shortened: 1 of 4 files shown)


kesselb:
Looks good to me. It's nearly impossible to have a message without message ID but already assigned to a thread right?

ChristophWurst:
> Looks good to me. It's nearly impossible to have a message without message ID but already assigned to a thread right?

Yes. I can only think about the scenario where the subject alone is taken for a thread. But inside this already rare edge case of a missing message ID this sounds even more unlikely.

Issue #2974, opened 2020-04-23
Next page sometimes loads the full mailbox


ChristophWurst:
This is possibly related to https://github.com/nextcloud/mail/pull/2648.

### Expected behavior
The next page is always 20 entries.

### Actual behavior
*Sometimes* it's all of the remaining messages. Which is already quite something to process for the browser in the case of a few hundred messages. But this will cause lockups for larger ones.

I *think* this does not happen when you use the app "normally". But when you have the app open and are scrolled down to at least the second page and reload then, the *load next page* logic gets triggered before the first page is loaded. Hence there is no cursor/limit yet and the back-end just returns everything.

### Mail app

**Mail app version:**

https://github.com/nextcloud/mail/pull/2648 but possibly also on current master.

ChristophWurst:
This also happens after the initial sync.

ChristophWurst:
We've just seen this out in the wild:

<details>

```json
{
  "reqId": "S4U4aE5GFeZeUkvPKxF2",
  "level": 3,
  "time": "2021-02-12T14:43:35+00:00",
  "remoteAddr": "79.148.98.98",
  "user": "marco",
  "app": "core",
  "method": "POST",
  "url": "/apps/mail/api/mailboxes/2412/sync",
  "message": {
    "Exception": "Doctrine\\DBAL\\Query\\QueryException",
    "Message": "More than 1000 expressions in a list are not allowed on Oracle.",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/lib/public/AppFramework/Db/QBMapper.php",
        "line": 320,
        "function": "execute",
        "class": "OC\\DB\\QueryBuilder\\QueryBuilder",
        "type": "->"
      },
      {
        "file": "/var/www/html/apps/mail/lib/Db/MessageMapper.php",
        "line": 568,
        "function": "findEntities",
        "class": "OCP\\AppFramework\\Db\\QBMapper",
        "type": "->"
      },
      {
        "file": "/var/www/html/apps/mail/lib/Service/Sync/SyncService.php",
        "line": 156,
        "function": "findIdsByQuery",
        "class": "OCA\\Mail\\Db\\MessageMapper",
        "type": "->"
      },
      {
        "file": "/var/www/html/apps/mail/lib/Service/Sync/SyncService.php",
        "line": 126,
        "function": "getDatabaseSyncChanges",
        "class": "OCA\\Mail\\Service\\Sync\\SyncService",
        "type": "->"
      },
[...]

PR #4731, opened 2021-03-10
Do not attempt to retry an (initial) sync if mailbox is locked

ChristophWurst:
When the app loads a mailbox, we try to fetch the 20 first messages. The
server will respond with a client error of HTTP400 when the mailbox
hasn't been cached yet. The client then first attempts an *intial sync*,
followed by re-fetching of the 20 first messages.

So far for the conflict-free sequence of commands. In the priority inbox
we have three virtual mailboxes that all try to get their data at the
same time. For initialized mailboxes the server will give away the 20
first messages for the respective query. For the very first access,
however, the mailboxes battle for the initial sync. Each of the
mailboxes will run into the HTTP400 for the mailbox that isn't cached,
then they will all three try to do the initial sync. One of them wins.
The other ones previous *retried* the intial sync. And that is where
things blew up. The attempted initial sync resulted in a regular sync.
And since the client requested a sync (diff) with no known message IDs,
the server assumed that all messages of the mailbox are "new". Hence
there was a huge server load and response, followed by a frozen browser
that tried to render hundrets or thousands of messages at once,
depending of the inbox size.

With this patch the initial sync doesn't actually attempt to retry when
it detects a locked mailbox. That is becaues it doesn't make much sense.
If the mailbox is locked then because some other process already takes
care of syncing. We don't have to do it ourselves. Instead we should try
to actually fetch the 20 newest messages. If that works then the process
is done.

Fixes https://github.com/nextcloud/mail/issues/2974

This patch also adds a horse load of logging, just to make any sync issues easier to debug in the future. Previously it wasn't trivial to tell similar sync/fetch processes apart when they didn't contain mailbox ID or the query argument.

### How to test

0. Enable debug mode
1. Configure an account with a fairly big inbox
2. Open the Priority Inbox
3. Click *Clear cache* on the actions menu of the inbox
4. See the page refresh
5. See the PI sections load/sync/fight

On master the browser hangs after a few secs. The logs shows something like 
Bildschirmfoto von 2021-03-10 17-01-38

… indicating that that many *new messages* are a problem. The browser will try to render them.

On this PR: only 20 messages max are fetched per account inbox. The browser stays reactive, your notebook fans stay calm.

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -143,6 +143,7 @@ export default {
 		mailbox() {
 			this.loadEnvelopes()
 				.then(() => {
+					logger.debug(`syncing mailbox ${this.mailbox.databaseId} (${this.searchQuery}) after mailbox change`)
 					this.sync(false)
 				})
 		},
@@ -159,6 +160,7 @@ export default {
 	async mounted() {
 		this.loadEnvelopes()
 			.then(() => {
+				logger.debug(`syncing mailbox ${this.mailbox.databaseId} (${this.searchQuery}) after mount`)
 				this.sync(false)
 			})
 	},
@@ -173,6 +175,7 @@ export default {
 			this.loadingCacheInitialization = true
 			this.error = false
 
+			logger.debug(`syncing mailbox ${this.mailbox.databaseId} (${this.searchQuery}) during cache initalization`)
 			this.sync(true)
 				.then(() => {
 					this.loadingCacheInitialization = false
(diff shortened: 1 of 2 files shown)


ChristophWurst:
eslint failure is unrelated

GretaD:
The logs are shown correct, but while refreshing from clear cache the first message doesnt open and it loads forever. And this error is shown:
```
ConsoleLogger.js:54 [ERROR] mail: could not load envelope thread {app: "mail", uid: "greta", threadId: 14521, error: Error: Request failed with status code 403
    at t.exports (http://localhost/nextcloud/apps/mail/j…}
```

ChristophWurst:
> while refreshing from clear cache the first message doesnt open and it loads forever

Because the URL points to a message that is wiped from the cache. That is a side effect but nothing critical. Wiping cache is an operation that we only offer for development instances. You can click the Mail icon from the apps menu and your inbox should look again

GretaD (approved):

PR #4739, opened 2021-03-11
Scope the cs-fixer and its dependencies

ChristophWurst:
Sometimes the dev dependencies of Mail cause a conflict with other apps
or Nextcloud server. Most of the Symfony dev deps are pulled in via the
php-cs-fixer, while that binary is actually independent and it doesn't
need to load any of its classes into a Nextcloud process. Therefore we
can use the composer bin plugin to move the bin into its own directory.

`composer run cs:fix` works just like before.


miaulalala (approved):

ChristophWurst:
This needs an arguably ugly Psalm baseline update. The issue is that prior to https://github.com/nextcloud/server/pull/25560/Nextcloud 22 the event dispatcher relied on the Symfony components, Psalm knows that and complains that the event is not there (in isolation, when Nextcloud's autoloader isn't involved).

This has been fixed, but as soon as we still target Nextcloud 20-21, the error might show up for any new event.

PR #4740, opened 2021-03-11
Scope phpunit and friends

ChristophWurst:
Same same as https://github.com/nextcloud/mail/pull/4739 but different.

Running tests works just like before, but anything phpunit pulls in won't be loadable by ordinary Nextcloud processes anymore during development. Yay for independence.


ChristophWurst:
everything passes and no production code changes -> merge

PR #4743, opened 2021-03-11
Add show images temporarily

GretaD:
Quoting Christoph - " I want to show images of an email because it renders badly without. but that doesn't mean I want to always have their images loaded"

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -5,6 +5,10 @@
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
 			<Actions default-icon="icon-toggle" :menu-title="t('mail', 'Show')">
 				<ActionButton icon="icon-toggle"
+					@click="displayIframe">
+					{{ t('mail', 'Show images temporarily') }}
+				</ActionButton>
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> Quoting Christoph - " I want to show images of an email because it renders badly without. but that doesn't mean I want to always have their images loaded"

what a quote :pray:

src/components/MessageHTMLBody.vue
```diff
@@ -4,6 +4,10 @@
 		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content" style="color: #000000">
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
 			<Actions default-icon="icon-toggle" :menu-title="t('mail', 'Show')">
+				<ActionButton icon="icon-toggle"
+					@click="displayIframe">
+					{{ t('mail', 'Show images temporarily from {sender}', {sender: message.from[0].email}) }}
```
ChristophWurst, line 9:
```suggestion
					{{ t('mail', 'Show images temporarily}) }}
```

No need to repeat who the message is from

ChristophWurst (approved):

PR #4750, opened 2021-03-12
Fix button style

GretaD:
- [x] Fix primary button style

**The problem**
icon_button


GretaD:
closing this as this revert fixes the problem: https://github.com/nextcloud/server/pull/26283

PR #4765, opened 2021-03-15
Lock the full mailbox when doing a (forced) initial sync

ChristophWurst:
I honestly can't tell if this solves and of the existing sync/locking issues or creates new ones. But while reading the code and stepping it through with a debugger I noticed that a forced sync of one mailbox does not actually lock it.

Would anyone else mind taking a closer look?

### DIFF
--- a/lib/Service/Sync/ImapToDbSynchronizer.php
+++ b/lib/Service/Sync/ImapToDbSynchronizer.php
@@ -197,15 +197,15 @@ class ImapToDbSynchronizer {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@kesselb @miaulalala @st3iny @StCyr what do you think? In https://github.com/nextcloud/mail/issues/4853#issuecomment-809450549 I saw that we use the forced/initial sync to get the UID of the saved draft. I'm not completely certain but it looks like the missing locks cause concurrent inserts and therefore some requests fail.

miaulalala (approved):
:+1:

Issue #4456, opened 2021-02-03
Add a manual update/refresh button


ChristophWurst:
### Feature Request

As a user I want to have an option force a refresh of the current mailbox.

#### Summary

Like the `r` shortcut, but via a button.

GretaD:
@jancborchardt Sup? Can you please suggest where we can add this button, so i can start working on it?

jancborchardt:
Yep, this requires the design change as specified at https://github.com/nextcloud/server/issues/16934#issuecomment-542201187 (for both Mail and Contacts ideally, in the component, while we are at it)

Then a refresh icon can be shown on the same row of the "New message" button, aligned to the right.

ChristophWurst:
I raised my personal concerns with this design for Mail before because IMO there is semantic/logic to a *+ Add contact* button in Contacts where you actually add a new item to that very list but in Mail you don't. You don't add a message to that *list* (inbox).

Another issues is that we are already fighting to get more information displayed for priority inbox because there we have the three section titles that take up some vertical space and therefore fewer actual messages are shown. If we have that big button on top we'll hide at least one more message from the list.

So I'm not feeling very confident with this change.

Also looking at other sophisticated solutions I don't know any that has the *New message* button there, like neither Geary nor Gmail.

ya-d:
Thank you for taking the time to consider and discuss these changes. I don't think either of this is an issue:
- Right now, the "+ New Message" button sits above the list of your mailboxes. Semantically it would even be an improvement to have it on top of your list of conversations (even though they are incoming).
- You have to place the hamburger icon somewhere which currently overlaps content. So the vertical space is already lost to that. There is also die "i" button which right now wildly floats around with content. (I don't think vertical space is the most urgent problem of the current design but rather [#16934](https://github.com/nextcloud/server/issues/16934#issuecomment-542201187) and #3940 / #4123.)
- It might be an option to move the create new / refresh buttons to the top right of the message view but that won't get rid of the hamburger icon.

Regarding the topic in this issue: **pull to refresh** is a rather common metaphor these days and might be an option .

Thank you so much for pushing Nextcloud forward. Keep it up!

<img width="279" alt="image" src="https://user-images.githubusercontent.com/30781876/107774761-ba3f2600-6d3f-11eb-951c-2220fa79f912.png">

ChristophWurst:
> **pull to refresh** is a rather common metaphor these days and might be an option

[...]

PR #4767, opened 2021-03-15
Add refresh button for current mailbox

GretaD:
Fixes https://github.com/nextcloud/mail/issues/4456

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -27,6 +27,10 @@
 			button-class="icon-add"
 			role="complementary"
 			@click="onNewMessage" />
+		<button v-if="currentMailbox"
+			class="button icon-history"
+			:disabled="refreshing"
+			@click="refreshMailbox" />
 		<template #list>
 			<ul id="accounts-list">
 				<template v-for="group in menu">
@@ -97,6 +101,11 @@ export default {
 		NavigationAccountExpandCollapse,
 		NavigationMailbox,
 	},
+	data() {
+		return {
+			refreshing: false,
+		}
+	},
 	computed: {
 		menu() {
 			return this.$store.getters.accounts.map((account) => {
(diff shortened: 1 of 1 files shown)


src/components/Navigation.vue
```diff
@@ -27,6 +27,9 @@
 			button-class="icon-add"
 			role="complementary"
 			@click="onNewMessage" />
+		<button v-if="currentMailbox"
```
ChristophWurst, line 30: can we also visually disable the button during a sync?

src/components/Navigation.vue
```diff
@@ -166,9 +180,46 @@ export default {
...
+				this.refreshing = false
+			}
+		},
 	},
 }
 </script>
 
-<style scoped>
+<style lang="scss" scoped>
+::v-deep .button {
```
ChristophWurst, line 203: are you sure v-deep is required here?

ChristophWurst (approved):
Code looks good!

src/components/Navigation.vue
GretaD: of course not

StCyr:
Could you add a little animation?

cyrille@vm-cyrille-debian:/var/www/nextcloud/apps/mail$ git diff src/
diff --git a/src/components/Navigation.vue b/src/components/Navigation.vue
index 4a75baa39..b41eda99a 100644
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -218,9 +218,20 @@ export default {
        &:disabled {
                cursor: not-allowed;
                opacity: .5;
+               animation: rotation infinite 2s linear;
        }
 }
 ::v-deep .app-navigation-new button {
        width: 250px !important;
 }
+
+@keyframes rotation{
+  from{
+    transform:rotate(-0deg);
+  }
+
+  to{
+    transform:rotate(-360deg);
+  }
+}
 </style>

StCyr (approved):
Let's merge!

Issue #4758, opened 2021-03-14
Attachments not visible for some mails.


simon1964:
### Expected behavior
An paperclip icon showing their is an attachment visible in the list of mail as well as in the opened mail.

### Actual behavior
The paperclip is only visible in the list of mail, not in the opened mail.
This behavior is true for all mail form a specific sender. Other senders are OK.
Changes since the last time I know it worked:
-mail upgrade to 1.9.3
-nextcloud upgrade from 20 to 21
-Debian upgrade from 10 (buster) to 11 (Bullseye)

### Mail app

**Mail app version: 1.93

**Mailserver or service: dovecot

### Server configuration

**Operating system**: Debian 11

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.4

**Nextcloud Version: 21

#### Client configuration
**Browser:** Firefox / Chrome / Edge

**Operating system:** Linux / Windows

ChristophWurst:
would it be possible to share the source of that email if it doesn't contain anything sensitive? I'd like to inspect the headers of the attachments, especially their content types.

Michael2222:
Hello,

I have the same problem: MS-Teams invitaion, but no displayed the *.ics file and so no button to copy in my calendar.
If I open the mail in another mail client program, the *.ics file is here.

Regards
   Michael

ChristophWurst:
https://github.com/nextcloud/mail/issues/4463 for the teams issue.

@simon1964 if you are also referring to an email from teams then we can close this ticket as duplicate.

StCyr:
might also be linked to https://github.com/nextcloud/mail/pull/4667

simon1964:
Hope this helps. Tried to copy the whole maildir file in, but that didn't work.

--000000000000325c2d05bd935466--
--000000000000325c2f05bd935468
Content-Type: application/pdf; name="NVC Needs vs. Strategies.pdf"
Content-Disposition: attachment; filename="NVC Needs vs. Strategies.pdf"
Content-Transfer-Encoding: base64
Content-ID: <f_kmampzbx0>
X-Attachment-Id: f_kmampzbx0

mfavoretto:
I also have the same problem

The paperclip appears only the general view of emails

StCyr:
> I also have the same problem
> 
> The paperclip appears only the general view of emails

If you could provide me with the source of such a message I could investigate (I might be the responsible of this bug)

mfavoretto:
> 
> 
[...]

PR #4776, opened 2021-03-17
Improve handling of attachment and inline parts.

StCyr:
Fixes https://github.com/nextcloud/mail/issues/4758

Tested here on my dev box. Did not see any adversial effect

### DIFF
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -107,6 +107,7 @@ class IMAPMessage implements IMessage, JsonSerializable {
 	public $htmlMessage = '';
 	public $plainMessage = '';
 	public $attachments = [];
+	public $inlineAttachments = [];
 	private $loadHtmlMessage = false;
 	private $hasHtmlMessage = false;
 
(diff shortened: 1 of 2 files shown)


lib/Model/IMAPMessage.php
```diff
@@ -327,9 +327,6 @@ private function hasAttachments($part) {
 			/** @var Horde_Mime_Part $p */
 			$filename = $p->getName();
 
-			if ($p->getContentId() !== null) {
```
ChristophWurst, line 330: what if we instead check the content-disposition header?

StCyr: why not indeed?
It's what I did in the frontend in https://github.com/nextcloud/mail/pull/4667/files

StCyr: This would cause an issue when the message's body is of type text/plain: in such case, the inline attachment won't be shown

StCyr: Don't know if it's that important though...

Also, maybe we should simply replace this function by Horde_Mime_Part->isAttachments() ? As what's done in MessageMapper...

StCyr: This becomes too much of a change. I'll create another PR to provide a quick fix to #4758.

Here I'll try to make various changes to improve how we handle attachment and inline mime parts.

Is it possible to cancel the backport of this PR?

ChristophWurst (approved):
embedded messages need special care, the rest seems fine

ChristophWurst (approved):
I fixed the special handling of inline messages

Tested with an embedded image – the image is not forwarded anymore. That is the expected behavior, right?

StCyr:
> 
> 
> I fixed the special handling of inline messages
> 
> Tested with an embedded image – the image is not forwarded anymore. That is the expected behavior, right?

Should not be forwarded as attachment. But, idealy, they should be embedded in the forwarded message I guess

ChristophWurst:
> But, idealy, they should be embedded in the forwarded message I guess

True, that would be the idea. But for that we need inline image support in our editor :)

PR #4777, opened 2021-03-17
This reverts a part of 17d61fc7c961a08cf2f6adb378d3e332e28144d8 causing issue.

StCyr:
Fixes https://github.com/nextcloud/mail/issues/4758

There's a more subtle issue with the way how we handle attachments though, which I aim to solve in https://github.com/nextcloud/mail/pull/4776


StCyr:
maybe we could implement #4776 directly in the end...

ChristophWurst (approved):

PR #4781, opened 2021-03-18
Change to pull_request_target

miaulalala:
See https://github.com/hmarr/auto-approve-action/issues/183 on why and how to avoid running into write only tokens for auto approve actions.


ChristophWurst (approved):
APPROVED

Issue #4659, opened 2021-03-04
Placement of signature in E-Mail when replying


Metagit:
### Feature Request

Currently the signature is added to the bottom of replies. If replying to some message, the quoted message will be between the  reply and the signature which is quite impractical and uncommon. Also see: https://help.nextcloud.com/t/placing-off-signature/87201

#### Summary

It would be nice to change the standard position to right below the reply.
On the long run it would probably make sense to implement a possibility in the UI to change the standard positioning.

thstyl2000:
The reason for this positioning of signature comes from the fact that many email clients cut-off everything below "--" when adding quotes, so that no signatures would show up in replies. Putting the signature block above the quoted text would then imply that you would not see the first message as a quote if the recipient replies again. 

Maybe it makes sense to be able to add a second (third, fourth,....) signature, one directly above the quoted text without "--" and the main signature block in the end of the message:

`Example message`
Dear Alice,

thank you for your reply!

`first signature`
Best,
Bob

Am 2021-03-17 12:09, schrieb Alice:
> Hello Bob!
> 
> It looks good!
> 
> Best,
> Alice

`main signature block`
\--
Mr. Spongebob Squarepants
124 Conch St., Bikini Bottom

ChristophWurst:
PR at https://github.com/nextcloud/mail/pull/4802

ChristophWurst:
It's done https://github.com/nextcloud/mail/pull/4971

PR #4802, opened 2021-03-22
Add option to place signature above quoted text

zimmersi:
For me it is more common the place the signature above the quoted text. Thats why I have implemented this new feature.

Fixes https://github.com/nextcloud/mail/issues/4659


ChristophWurst:
We did that with v1.9.0. The code is at https://github.com/nextcloud/mail/pull/4281.

zimmersi:
Hi @ChristophWurst ,
this is something else.
Of course, I want to have my answer on top/ above the quoted text, but also I want to have my signature on top/ above the quoted text. So that my answer as well as my signature is on top and NOT splitted by the quoted text.

ChristophWurst:
Got it, sorry for the confusion. Let's wait for https://github.com/nextcloud/mail/pull/4715, which brings some very important changes to the signature handling.

zimmersi:
:thumbsup:
all right!

zimmersi:
ok, now it is adapted to the changes from #4715

kesselb:
Hi @zimmersi :wave: 

Thanks for your pull request. I'm looking forward to add this feature to mail. Unfortunately there are some edge cases to consider.


https://user-images.githubusercontent.com/3902676/113028523-a9316380-918b-11eb-8c34-c65e5c4b8987.mp4

Using position 1 to place the signature above the reply does not work on account switch. I see two ways to tackle most edge cases: 

A) We added the signature plugin to have more control about the data passed to ckeditor. We could add another plugin "reply" plugin to add a custom element for the reply to ckeditor. Then we are able to search for those reply nodes in ckeditor and place the signature before or after the reply element. We may use `src/ckeditor/signature/SignaturePlugin.js` as template and change `signature` to `reply`. A command is not necessary. Next update `buildReplyBody` and wrap the reply with a `<div class="reply"></div>`. InsertSignatureCommand should then search for a reply element inside the nodes.

B) Another approach: `removeSignatureElement` is looking for a signature element. If we found one we could remember the current position and reuse it in `insertSignatureElement`. The idea is that a reply is usually an empty email. Then we can use position 1 to insert the signature the first time. If this email is loaded (e.g. from drafts) we find the signature position and use it instead 1. 

A or B should fix the above edge case. Not sure which way is more stable.

zimmersi:
hi @kesselb ,
thanks for your feedback!
sorry for the delay. I had some days off.

I guess it is better to implement approach A. B will not work if someone has accounts with different settings. e.g. one with signature above quote and one with signature below quote.

zimmersi:
omg, sorry for the chaos...

I tried to pass some checks and followed the instructions for DCO
`git rebase HEAD~15 --signoff`
`git push --force-with-lease origin signatureAboveQuote`

what is the best practise to restart my work?
delete my branch, create a new one, doing my stuff, create a new pull request?

zimmersi:
again, sorry for the chaos.
please head over to the new PR #4909

PR #4803, opened 2021-03-22
Report coverage less often

ChristophWurst:
Because https://github.com/codecov/codecov-action/issues/234 is getting more and more annoying.


st3iny (approved):
Very annoying indeed. Another workaround is to just restart all failed actions manually and wait for them to pass.

PR #4811, opened 2021-03-23
Add repair step for default tags

### LINKED ISSUE #4780: Add a repair step that adds missing default tags
### Feature Request

With https://github.com/nextcloud/mail/pull/4665 the app will support tags and during the migration of existing installation as well as for new accounts we create some default tags.

It would be cool if there will be a repair step that can check for missing default tags and create them, just in case admins fiddle with the database or something else drops these tags for a user.

#### Summary

Create missing default tags in a repair step.

miaulalala:
Fixes #4780

### DIFF
--- a/lib/Migration/AddMissingDefaultTags.php
+++ b/lib/Migration/AddMissingDefaultTags.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2021 Anna Larch <anna@nextcloud.com>
+ *
+ * @author 2021 Anna Larch <anna@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


lib/Migration/AddMissingDefaultTags.php
```diff
@@ -0,0 +1,82 @@
...
+
+	public function getName() {
+		return 'Restore default tags that are missing';
+	}
+
+	public function run(IOutput $output) {
+		$output->info('Looking up default tags');
+
+		/**
+		 * During the upgrade to v1.9.2/v1.10.0 the old version of the
```
ChristophWurst, line 61: I don't think that will happen because in v1.9.x there was no *tag mapper*. hence the upgrade will load the full class just for the repair step.

lib/Migration/Version1100Date20210304143008.php
```diff
@@ -105,16 +105,4 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
 	}
-
-	/**
-	 * @param IOutput $output
-	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
-	 * @param array $options
-	 */
-	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
-		$accounts = $this->mailAccountMapper->getAllUserIdsWithAccounts();
-		foreach ($accounts as $account) {
```
ChristophWurst, line 116: bot injected mappers will be unused -> :fire:

ChristophWurst (changes requested):

ChristophWurst (approved):

PR #4819, opened 2021-03-24
Fix database locks

miaulalala:
See #4812

Added the timestamp to the hasLocks method as to avoid only relying on tokens but not checking if they are still valid.

### DIFF
--- a/lib/Db/MailboxMapper.php
+++ b/lib/Db/MailboxMapper.php
@@ -148,7 +148,7 @@ class MailboxMapper extends QBMapper {
 		$now = $this->timeFactory->getTime();
 
 		if ($lock !== null
-			&& $lock > ($now - 5 * 60)) {
+			&& $lock > ($now - Mailbox::LOCK_TIMEOUT)) {
 			// Another process is syncing
 			throw MailboxLockedException::from($mailbox);
 		}
--- a/lib/Db/Mailbox.php
+++ b/lib/Db/Mailbox.php
@@ -82,6 +82,12 @@ class Mailbox extends Entity implements JsonSerializable {
 	protected $specialUse;
 	protected $syncInBackground;
 
+	/**
+	 * @var int
+	 * Lock timeout for sync (5 minutes)
+	 */
+	public const LOCK_TIMEOUT = 300;
+
 	public function __construct() {
 		$this->addType('accountId', 'integer');
 		$this->addType('messages', 'integer');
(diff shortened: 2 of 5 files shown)


lib/Db/MailboxMapper.php
```diff
@@ -31,16 +31,20 @@
...
 use OCP\IDBConnection;
-use function array_map;
 
 class MailboxMapper extends QBMapper {
 
+	/**
+	 * @var int
+	 * Lock timeout for sync (5 minutes)
+	 */
+	public const LOCK = 300;
```
ChristophWurst, line 46: I think the Mailbox entity is a better fit

GretaD:
While testing i realized that things are a better but with the combined mailboxes(in my case 2) the problem still exist if the initial sync is in progress.

lib/Db/MailboxMapper.php
```diff
@@ -119,7 +115,6 @@ public function findById(int $id): Mailbox {
 	 *
 	 * @return Mailbox
 	 *
-	 * @throws DoesNotExistException
```
ChristophWurst, line 122: please revert

lib/Db/MailboxMapper.php
```diff
@@ -93,7 +90,6 @@ public function find(Account $account, string $name): Mailbox {
...
```
ChristophWurst, line 96: revert

lib/Db/MailboxMapper.php
```diff
@@ -66,7 +64,6 @@ public function findAll(Account $account): array {
 	}
 
 	/**
-	 * @throws DoesNotExistException
```
ChristophWurst, line 69: revert

lib/Db/MailboxMapper.php
```diff
@@ -31,13 +31,11 @@
 use OCA\Mail\Account;
 use OCA\Mail\Exception\MailboxLockedException;
 use OCA\Mail\Exception\ServiceException;
-use OCP\AppFramework\Db\DoesNotExistException;
```
ChristophWurst, line 34: revert

lib/Db/MailboxMapper.php
```diff
@@ -31,13 +31,11 @@
...
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\AppFramework\Db\QBMapper;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
-use function array_map;
```
ChristophWurst, line 40: still used -> revert

ChristophWurst (approved):
Looks good

PR #4822, opened 2021-03-24
Rework getAllTags Query

### LINKED ISSUE #4820: TagMapper::getAllTagsForMessages should use a join
### Expected behavior

``\OCA\Mail\Db\TagMapper::getAllTagsForMessages`` loops over a query result to load related data and therefore creating the *n+1 problem*.

### Actual behavior

The data can be loaded in one go with the help of a join.

### Mail app

**Mail app version:** master

miaulalala:
Fixes #4820

### DIFF
--- a/lib/Db/TagMapper.php
+++ b/lib/Db/TagMapper.php
@@ -167,22 +167,31 @@ class TagMapper extends QBMapper {
 		}, $messages);
 
 		$qb = $this->db->getQueryBuilder();
-		$idsQuery = $qb->select('mt.*')
-			->from('mail_message_tags', 'mt')
+		$tagsQuery = $qb->select('t.*', 'mt.imap_message_id')
+			->from($this->getTableName(), 't')
+			->join('t', 'mail_message_tags', 'mt', $qb->expr()->eq('t.id', 'mt.tag_id', IQueryBuilder::PARAM_INT))
 			->where(
-				$qb->expr()->in('imap_message_id', $qb->createNamedParameter($ids, IQueryBuilder::PARAM_STR_ARRAY))
+				$qb->expr()->in('mt.imap_message_id', $qb->createNamedParameter($ids, IQueryBuilder::PARAM_STR_ARRAY))
 			);
-		$idsQuery = $idsQuery->execute();
-		$queryResult = $idsQuery->fetchAll();
-		if (empty($queryResult)) {
-			return [];
+		$queryResult = $tagsQuery->execute();
+		$tags = [];
+		while (($row = $queryResult->fetch()) !== false) {
+			$messageId = $row['imap_message_id'];
+			if (!isset($tags[$messageId])) {
+				$tags[$messageId] = [];
+			}
+
+			// Construct a Tag instance but omit any other joined columns
+			$tags[$messageId][] = Tag::fromRow(array_filter(
+				$row,
+				function (string $key) {
+					return $key !== 'imap_message_id';
+				},
+				ARRAY_FILTER_USE_KEY
+			));
 		}
-		$result = [];
-		foreach ($queryResult as $qr) {
-			$result[] = $qr['imap_message_id'];
-			$result[$qr['imap_message_id']][] = $this->getTag((int)$qr['tag_id']);
-		}
-		return $result;
+		$queryResult->closeCursor();
+		return $tags;
 	}
 
 	/**


lib/Db/TagMapper.php
```diff
@@ -145,20 +145,22 @@ public function getAllTagsForMessages(array $messages): array {
 
 		$qb = $this->db->getQueryBuilder();
 		$idsQuery = $qb->select('mt.*')
+			->addSelect('t.*')
 			->from('mail_message_tags', 'mt')
+			->join('mt', $this->getTableName(), 't', $qb->expr()->eq('mt.tag_id','t.id'))
```
ChristophWurst, line 150: for oracle compat we should always specify the `eq` type parameter when the comparison doesn't compare strings

lib/Db/TagMapper.php
```diff
@@ -145,20 +145,22 @@ public function getAllTagsForMessages(array $messages): array {
...
 		if (empty($queryResult)) {
 			return [];
 		}
-		$result = [];
-		foreach ($queryResult as $qr) {
-			$result[] = $qr['imap_message_id'];
-			$result[$qr['imap_message_id']][] = $this->getTag((int)$qr['tag_id']);
-		};
+		$result = array_map(function ($qr) {
+			$transformed[] = $qr['imap_message_id'];
```
ChristophWurst, line 160: this code confused me also in the previous version

so here you push a new item to the end of the array (at the numeric index or whatever php assigns)

lib/Db/TagMapper.php
```diff
@@ -145,20 +145,22 @@ public function getAllTagsForMessages(array $messages): array {
...
+			$transformed[$qr['imap_message_id']] = $qr;
```
ChristophWurst, line 161: and here you write to a specific index. Could you elaborate on that logic?

lib/Db/TagMapper.php
```diff
@@ -145,20 +145,22 @@ public function getAllTagsForMessages(array $messages): array {
...
+			return $transformed;
+		}, $queryResult);
 		return $result;
```
ChristophWurst, line 164: this method is supposed to return a Tag[][] but as far as I can see it's a generic array[][] built from row arrays

ChristophWurst (commented):
the `getTag` method will be unused now, we can :fire: it :)

lib/Db/TagMapper.php
```diff
@@ -145,20 +145,22 @@ public function getAllTagsForMessages(array $messages): array {
...
 		$idsQuery = $qb->select('mt.*')
+			->addSelect('t.*')
 			->from('mail_message_tags', 'mt')
+			->join('mt', $this->getTableName(), 't', $qb->expr()->eq('mt.tag_id','t.id'))
 			->where(
-				$qb->expr()->in('imap_message_id', $qb->createNamedParameter($ids, IQueryBuilder::PARAM_STR_ARRAY))
+				$qb->expr()->in('mt.imap_message_id', $qb->createNamedParameter($ids, IQueryBuilder::PARAM_STR_ARRAY))
 			);
 		$idsQuery = $idsQuery->execute();
 		$queryResult = $idsQuery->fetchAll();
```
ChristophWurst, line 155: not necessarily related but another microoptimization to get the memory footprint down is to not use `fetchAll` but `fetch` in a loop until you get no more results

lib/Db/TagMapper.php
miaulalala: Yeah the join doesn't return the entity apperntly. I don't know enough about the query builder to change that. Any suggestions?

ChristophWurst: if you pass the select query into `\OCP\AppFramework\Db\QBMapper::findEntities` you should get an array of entities back. you'll then have to group them per message.

lib/Db/TagMapper.php
miaulalala: because i want the array to be indexed by the imap_message_id so I can access it in the message mapper via: 			`$dbTags = $tags[$message->getMessageId()] ?? [];` - otherwise I will have to loop through the `$tags` array to find those entries.

lib/Db/TagMapper.php
miaulalala: see as above. I realise now that i need to add an additional [] accessor to this otherwise it will only ever show one tag

lib/Db/TagMapper.php
miaulalala: won't work because I don't have a mapper for the message_tags

miaulalala: that's actually why I didn't use a join in the first place, because I need the enttity to further proceed in the code.

ChristophWurst: you were right, we can't use `findEntities` directly but we can still build the entities in a loop if we omit the columns that are not meant for the entitiy. see my latest changes :)

lib/Db/TagMapper.php
ChristophWurst: done

lib/Db/TagMapper.php
ChristophWurst: done

ChristophWurst (approved):

PR #4830, opened 2021-03-25
Fix important icon on envelope

GretaD:
Lint fixed also 2 other files missing comma

### DIFF
--- a/src/store/mutations.js
+++ b/src/store/mutations.js
@@ -179,7 +179,7 @@ export default {
 		envelope.flags[flag] = value
 	},
 	tagEnvelope(state, { envelope, tag, value }) {
-		envelope.tags[tag] = value
+		Vue.set(envelope.tags, tag, value)
 	},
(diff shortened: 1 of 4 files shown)


src/store/mutations.js
```diff
@@ -179,7 +179,7 @@ export default {
 		envelope.flags[flag] = value
 	},
 	tagEnvelope(state, { envelope, tag, value }) {
```
ChristophWurst, line 181: we have to check all usages of `tagEnvelope`. in contrast to `flagEnvelope` the `value` here isn't a boolean but an object, that represents the tag. Alternatively we could also normalize the data like we do with the rest of accounts/mailboxes/messages. In the long run and especially as soon as tags can be renamed and their color changed, we'll want to have the data normalized so the update of one tag is shown everywhere instantly.

GretaD:
closing as it will be included here https://github.com/nextcloud/mail/issues/4871

PR #4838, opened 2021-03-26
Confirm before sending to many recipients in to/cc

### LINKED ISSUE #4505: Confirm before sending to many recipients in `To` or `Cc`
### Feature Request

Sending mails to groups works great but users should be warned if they try to send an email with many recipients/groups or groups with many members in `To` or `Cc`. This should prevent leaking email addresses.

myrho:
Fixes #4505

I'm not sure about hard-coding the limit of to/cc recipients. Is there a better, configurable way?

### DIFF
--- a/lib/Exception/ManyRecipientsException.php
+++ b/lib/Exception/ManyRecipientsException.php
@@ -3,8 +3,6 @@
 declare(strict_types=1);
 
 /*
- * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
- *
  * @author 2021 Matthias Rella <mrella@pisys.eu>
  *
  * @license GNU AGPL version 3 or any later version
--- a/lib/Exception/ManyRecipientsException.php
+++ b/lib/Exception/ManyRecipientsException.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2021 Matthias Rella <mrella@pisys.eu>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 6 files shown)


lib/Exception/ManyRecipientsException.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
```
ChristophWurst, line 6: it wasn't me

tests/Unit/Controller/AccountsControllerTest.php
```diff
@@ -438,6 +439,26 @@ public function testSendingError() {
...
+	}
+
+	public function testSendingManyRecipientsCcError() {
+		$this->expectException(ManyRecipientsException::class);
+		$recipients = [];
+		for ($i = 0; $i < 11; $i++) {
+			$recipients[] = "$i@x.com";
+		}
+		$recipients = implode(',', $recipients);
+		$this->controller->send(13, 'sub', 'bod', '', $recipients, '');
```
ChristophWurst, line 459: please add an empty line before this

most of our tests follow the AAA pattern: arrange, act, assert. The first few lines are where you arrange things, the final one is where you act.

src/components/Composer.vue
```diff
@@ -226,6 +226,18 @@
 			{{ t('mail', 'Retry') }}
 		</button>
 	</div>
+	<div v-else-if="state === STATES.WARNING" class="emptycontent" role="alert">
+		<h2>{{ t('mail', 'Warning sending your message') }}</h2>
+		<p v-if="errorText">
+			{{ errorText }}
```
ChristophWurst, line 232: this won't be translated. can we check on the error type and have a special case for ``ManyRecipientsError``?

src/components/Composer.vue
```diff
@@ -278,6 +292,7 @@ const STATES = Object.seal({
 	SENDING: 2,
 	ERROR: 3,
 	FINISHED: 4,
+	WARNING: 5,
```
ChristophWurst, line 295: I think that should be moved up and FINISHED should stay last

ChristophWurst (changes requested):

src/components/Composer.vue
myrho: I did this by copy pasting and adapting L217-228. Matching for `ManyRecipientsError` happens in L740.

src/components/Composer.vue
myrho: So `WARNING: 4` and `FINISHED: 5`?

ChristophWurst: yep :)

src/components/Composer.vue
ChristophWurst: my bad.

miaulalala (approved):
Looks good!

myrho:
Ah sorry, I see. My bad!

Issue #4825, opened 2021-03-25
Delete all message_tags entries if a user deletes the associated tag


miaulalala:
### Feature Request

Currently, we leave rows orphaned in `message_tags` if a tag is deleted. Rows that have invalid entries should be cleaned up two ways - remove all associated entries at the time of deleting a tag, and also add a deleteOrphans() to the Cleanup Job.

ChristophWurst:
Ref #4806 because as long as we don't have user defined tags there is nothing to delete or clean up :)

miaulalala:
> Ref #4806 because as long as we don't have user defined tags there is nothing to delete or clean up :)

Not quite true, some tags might be set/unset by users in another mail client and still show up in our app :)

Issue #4827, opened 2021-03-25
1071 Specified key was too long; max key length is 3072 bytes


StCyr:
### Expected behavior
After pulling the latest changes on master (7bc1cb1f64294236dcab0f24d4757399539c7524), I should be able to build, update, and run Mail

### Actual behavior
The update step fails with the following log:

[...]

PR #4840, opened 2021-03-27
Rework duplicate entry handling for tags

### LINKED ISSUE #4833: Rework how duplicate entries in message_tags are handled
## Summary

- [x]  Remove the unique index constraint from message_tags. The current implementation doesn't work for a few different reasons and has caused a need for some hacky solutions like #4821.
- [x] Add a cleanup job for duplicate entries  
- [x] Add a cleanup job for orphans.
- [x] The join query in `getAllTagsForMessages()` needs to handle selecting unique values only.
- [x] Additionally, to speed up the join query, a shortened index like seen in [nextcloud/server#26070](https://github.com/nextcloud/server/pull/26070) could prove useful.

miaulalala:
Fixes #4833, #4825

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -65,8 +65,7 @@ class MessageMapper extends QBMapper {
 	public function __construct(IDBConnection $db,
 								ITimeFactory $timeFactory,
 								TagMapper $tagMapper,
-								PerformanceLogger $performanceLogger
-								) {
+								PerformanceLogger $performanceLogger) {
 		parent::__construct($db, 'mail_messages');
 		$this->timeFactory = $timeFactory;
 		$this->tagMapper = $tagMapper;
@@ -120,7 +119,7 @@ class MessageMapper extends QBMapper {
 		return $max;
 	}
 
-	public function findByUserId(string $uid, int $id): Message {
+	public function findByUserId(string $userId, int $id): Message {
 		$query = $this->db->getQueryBuilder();
 
 		$query->select('m.*')
@@ -128,11 +127,11 @@ class MessageMapper extends QBMapper {
 			->join('m', 'mail_mailboxes', 'mb', $query->expr()->eq('m.mailbox_id', 'mb.id', IQueryBuilder::PARAM_INT))
 			->join('m', 'mail_accounts', 'a', $query->expr()->eq('mb.account_id', 'a.id', IQueryBuilder::PARAM_INT))
 			->where(
-				$query->expr()->eq('a.user_id', $query->createNamedParameter($uid)),
+				$query->expr()->eq('a.user_id', $query->createNamedParameter($userId)),
 				$query->expr()->eq('m.id', $query->createNamedParameter($id, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
 			);
 
-		$results = $this->findRelatedData($this->findEntities($query));
+		$results = $this->findRelatedData($this->findEntities($query), $userId);
 		if (empty($results)) {
 			throw new DoesNotExistException("Message $id does not exist");
 		}
@@ -364,7 +363,7 @@ class MessageMapper extends QBMapper {
 				$query->expr()->eq('mailbox_id', $query->createParameter('mailbox_id'))
 			));
 		// get all tags before the loop and create a mapping [message_id => [tag,...]]
-		$tags = $this->tagMapper->getAllTagsForMessages($messages);
+		$tags = $this->tagMapper->getAllTagsForMessages($messages, $account->getUserId());
 		$perf->step("Selected Tags for all messages");
 		foreach ($messages as $message) {
 			if (empty($message->getUpdatedFields())) {
@@ -552,7 +551,7 @@ class MessageMapper extends QBMapper {
 				$qb->expr()->isNotNull('m1.thread_root_id')
 			)
 			->orderBy('sent_at', 'desc');
-		return $this->findRelatedData($this->findEntities($selectMessages));
+		return $this->findRelatedData($this->findEntities($selectMessages), $account->getUserId());
 	}
 
 	/**
@@ -808,8 +807,7 @@ class MessageMapper extends QBMapper {
 				$qb->expr()->in('uid', $qb->createNamedParameter($uids, IQueryBuilder::PARAM_INT_ARRAY))
 			)
 			->orderBy('sent_at', 'desc');
-
-		return $this->findRelatedData($this->findEntities($select));
+		return $this->findRecipients($this->findEntities($select));
 	}
 
 	/**
@@ -817,7 +815,7 @@ class MessageMapper extends QBMapper {
 	 *
 	 * @return Message[]
 	 */
-	public function findByIds(array $ids): array {
+	public function findByIds(string $userId, array $ids): array {
 		if (empty($ids)) {
 			return [];
 		}
@@ -831,7 +829,7 @@ class MessageMapper extends QBMapper {
 			)
 			->orderBy('sent_at', 'desc');
 
-		return $this->findRelatedData($this->findEntities($select));
+		return $this->findRelatedData($this->findEntities($select), $userId);
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


lib/Migration/Version1100Date20210326103929.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
```
ChristophWurst, line 13: :fire:

lib/Db/TagMapper.php
```diff
@@ -248,4 +225,26 @@ public function createDefaultTags(MailAccount $account): void {
...
+	public function deleteOrphans() {
+		$qb = $this->db->getQueryBuilder();
+		$subQb1 = $this->db->getQueryBuilder();
+		$dqb = $this->db->getQueryBuilder();
+
+		// get lowest ids
+		$groupByQuery = $subQb1->select($qb->func()->min('id'))
+			->from('mail_message_tags')
+			->groupBy('imap_message_id','tag_id')
+			->having('COUNT(id)>1');
```
ChristophWurst, line 238: use ``\OCP\DB\QueryBuilder\IQueryBuilder::func`` and ``\OCP\DB\QueryBuilder\IFunctionBuilder::count`` to generate the COUNT(id) and the expression builder for the >1

lib/Db/TagMapper.php
```diff
@@ -248,4 +225,26 @@ public function createDefaultTags(MailAccount $account): void {
...
+
+		// select those for deleting
+		$select = $qb->select('*')
+			->from($qb->createFunction('(' . $groupByQuery->getSql() .')'), 'temp');
```
ChristophWurst, line 242: is a select from a sub query supported by all of our supported databases?

ChristophWurst: on a second thought, does that query do anything useful? you select * from another query. can't you just use that other query?

lib/Migration/Version1100Date20210326103929.php
```diff
@@ -0,0 +1,37 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('mail_message_tags')) {
+			$table = $schema->getTable('mail_message_tags');
+			if ($table->hasIndex('mail_msg_tag_id_idx')) {
+				$table->dropIndex('mail_msg_tag_id_idx');
+			}
+			if (!$table->hasIndex('mail_msg_imap_id_idx')) {
+				$table->addIndex(['imap_message_id'], 'mail_msg_imap_id_idx', [], ["lengths" => [128]]);
```
kesselb, line 32:
```suggestion
				$table->addIndex(['imap_message_id'], 'mail_msg_imap_id_idx', [], ['lengths' => [128]]);
```

kesselb:
```
An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1 no such table: oc_ 
```

You once again found a bug ;) `$qb->delete` is usually called with table name and alias. Null is also accepted but `QueryBuilder.getTableName` does not handle it correctly. 

image

image

As workaround for mail we can pass the table name to delete directly and remove the from call. For server we should fix prefixTableName to return null.

kesselb (changes requested):

st3iny (commented):
Just some minor code style issues :nerd_face:

miaulalala:
Fixes #4825

ChristophWurst:
> Fixes #4825

FYI you have to put this into the PR description on top, it won't work with a comment AFAIK

lib/Db/TagMapper.php
```diff
@@ -248,4 +225,31 @@ public function createDefaultTags(MailAccount $account): void {
 			$this->insert($entity);
 		}
 	}
+
+	public function deleteDuplicates(): void {
+		$selectQb = $this->db->getQueryBuilder();
```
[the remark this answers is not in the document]
miaulalala: Unfortunately this won't work in MySQL and probably others, because you can't update a table that is used in a select. I'll have to loop through the results of the subquery.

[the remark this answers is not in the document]
kesselb: OK. Executing the subselect, fetching the results and use it for the delete operation should also work. There is a check if more than 1000 arguments passed to a query. Need to limit the number of mail_message_tags in one delete operation to 1000.

lib/Db/MessageMapper.php
```diff
@@ -937,8 +938,8 @@ public function findChanged(Mailbox $mailbox, int $since): array {
 				$qb->expr()->eq('mailbox_id', $qb->createNamedParameter($mailbox->getId(), IQueryBuilder::PARAM_INT)),
 				$qb->expr()->gt('updated_at', $qb->createNamedParameter($since, IQueryBuilder::PARAM_INT))
 			);
-
-		return $this->findRelatedData($this->findEntities($select));
+		// TODO: change this to findRelatedData
```
ChristophWurst, line 941: obsolete?

lib/Db/TagMapper.php
```diff
@@ -159,21 +126,24 @@ public function untagMessage(Tag $tag, string $messageId): void {
...
 		$ids = array_map(function (Message $message) {
 			return $message->getMessageId();
 		}, $messages);
 
 		$qb = $this->db->getQueryBuilder();
-		$tagsQuery = $qb->select('t.*', 'mt.imap_message_id')
+		$tagsQuery = $qb->selectDistinct(['t.*', 'mt.imap_message_id'])
 			->from($this->getTableName(), 't')
-			->join('t', 'mail_message_tags', 'mt', $qb->expr()->eq('t.id', 'mt.tag_id', IQueryBuilder::PARAM_INT))
+			->join('t','mail_message_tags', 'mt',$qb->expr()->eq('t.id', 'mt.tag_id'))
```
ChristophWurst, line 140: spaces and parameter argument can be reverted to the old version, right?

lib/Db/TagMapper.php
```diff
@@ -240,12 +210,59 @@ public function createDefaultTags(MailAccount $account): void {
...
 			return strcmp($a->getImapLabel(), $b->getImapLabel());
 		});
 		foreach ($toInsert as $entity) {
 			$this->insert($entity);
 		}
 	}
+
+	public function deleteDuplicates(): void {
+		$qb = $this->db->getQueryBuilder();
+		$dqb = $this->db->getQueryBuilder();
```
ChristophWurst, line 224: nitpick: I'd appreciate more descriptive variables like ``$deleteQueryBuilder`` or whatever the `d` stands for :)

and you can also move the initialization of the query builder down to where it's actually used. we don't need it before ~L240

lib/Db/TagMapper.php
```diff
@@ -240,12 +210,59 @@ public function createDefaultTags(MailAccount $account): void {
...
+		$qb->select('mt2.id')
+			->from('mail_message_tags', 'mt2')
+			->join('mt2','mail_message_tags', 'mt1', $qb->expr()->andX(
+					$qb->expr()->gt('mt1.id', 'mt2.id'),
+					$qb->expr()->eq('mt1.imap_message_id', 'mt2.imap_message_id'),
+					$qb->expr()->eq('mt1.tag_id', 'mt2.tag_id')
+				)
+			);
+		$result = $qb->execute();
+		$rows = $result->fetchAll();
```
ChristophWurst, line 234: needs a `$result->closeCurser();` after the fetch

lib/Db/TagMapper.php
```diff
@@ -240,12 +210,59 @@ public function createDefaultTags(MailAccount $account): void {
...
+	public function deleteOrphans(): void {
+		$qb = $this->db->getQueryBuilder();
+		$dqb = $this->db->getQueryBuilder();
+
+		$qb->select('mt.id')
+			->from('mail_message_tags', 'mt')
+			->leftJoin('mt', $this->getTableName(), 't', $qb->expr()->eq('t.id', 'mt.tag_id'))
+			->where($qb->expr()->isNull('t.id'));
+		$result = $qb->execute();
+		$rows = $result->fetchAll();
```
ChristophWurst, line 256: close the cursor

ChristophWurst (commented):
Looks good otherwise!

lib/Db/MessageMapper.php
miaulalala: yes

lib/Db/TagMapper.php
miaulalala: yes

StCyr (approved):
Does solve https://github.com/nextcloud/mail/issues/4827 indeed. Thanks 🥳

PR #4859, opened 2021-03-29
Bump CKEditor to 27

GretaD:


ChristophWurst:
what version of `npm` do you use?

what is left before review?

GretaD:
> what version of `npm` do you use?
> 
6.14.4

> what is left before review?
>

me making sure that -23.657 removals is alright i didnt do smth wrong

ChristophWurst:
I think the big diff is from npm 6 because the npm lockfile was previously written with v7 and they have switched the format

ChristophWurst (approved):
Looks good, didn't test

ChristophWurst (changes requested):
https://github.com/nextcloud/mail/commit/0a16e28b6f63857d11483627d4a3c65fc05a736a was better

ChristophWurst (approved):
:+1:

Issue #4805, opened 2021-03-23
Show tags in envelope list and message details


ChristophWurst:
I'm splitting this from the generic ticket https://github.com/nextcloud/mail/issues/25 so we can discuss the design better.

## Envelope list (second column)

The Nextcloud Mail back-end now support tags and you can access them via `envelope.tags`. The one for *important* email is already visible to the user via the importance icon top left of the envelope avatar. The tag can be toggled via the envelope's action menu.

Now the question is: how do we display the rest of our current five [default tags](https://github.com/nextcloud/mail/blob/c8d2059c2ef25e4fd6ab472e88294c23cd7c5593/lib/Db/TagMapper.php#L183-L207) *work*, *personal*, *to do* and *later*, but also any other user-defined tags once we have a UI for the user to define their own additional tags?

With #4123 we'll have more vertical space.

## Thread details (third column)

Just like in envelopes we should show the tags in the thread view, because each of the messages in a thread could have tags attached.

---

## Brainstorming from Talk

![Bildschirmfoto von 2021-03-23 14-33-21](https://user-images.githubusercontent.com/1374172/112154757-de1a4500-8be4-11eb-9e23-87559c753cc4.png)
![Bildschirmfoto von 2021-03-23 14-33-43](https://user-images.githubusercontent.com/1374172/112154756-dd81ae80-8be4-11eb-8091-2fc6a85f8449.png)

---

cc @nimisha-vijay @GretaD who I've we had a discussion about this recently in our dev channel

thstyl2000:
What about let the user configure the lists in the priority inbox to group together tagged messages? For example instead of

- Important and Unread
- Favorites
- Other

one may choose to configure the PI as follows:

- Important
- Unread
- [Work] && [To Do] && [Unread]
- [Personal] && [Read]

BTW Thank you for your great great great work!!!!

ChristophWurst:
This could be configurable. Could you open a separate feature request?

thstyl2000:
Done: #4816. Thanks!

nimisha-vijay:
Something like this could work?

![image](https://user-images.githubusercontent.com/52440189/112480073-a73a5f80-8d9b-11eb-95e0-723319e7299f.png)

Multiple tags can just wrap to the next line, and if the envelope list is very narrow then the tags could be small like on the right. 
Not sure about the colours yet, may want to use similar colours as the tags on the Deck app.

GretaD:
[...]

Issue #4806, opened 2021-03-23
Tag management UI


ChristophWurst:
This is a more detailed ticket about the UI parts of https://github.com/nextcloud/mail/issues/25 for users to set up their tags.

## User story

As a user I want to be able to add, edit and possibly delete the tags that are used in my email accounts. The system will create five default tags for me, just like Thunderbird does. I want to be able to change their color as well as their *display name*.

![Bildschirmfoto von 2021-03-23 10-21-31](https://user-images.githubusercontent.com/1374172/112123528-a1892200-8bc1-11eb-876c-730e333eff2f.png)

## Technical background

We don't store tags on IMAP directly because of encoding and length limitations. Hence we use *keywords* aka placeholders that are sent to IMAP and mapped in the application. For example *important* is saved as *$label1* on IMAP.

Every tag has a
* keyword for imap
* display name for the UI
* color

Just like in Thunderbird I would only allow the users to influence display name and color, the keyword should be assigned by the system (e.g. incrementing `$labelXYZ`).

![Bildschirmfoto von 2021-03-23 10-22-15](https://user-images.githubusercontent.com/1374172/112123683-c41b3b00-8bc1-11eb-8ea0-d995f47d8841.png)

As of #4665 we have the back-end support to manage tags inside Nextcloud. For the most part we need an UI, but possibly a few controller methods to create and modify user tags.

As opposed to many other settings tags are not to be configured per email account but per Nextcloud account. Hence they should probably not show in the account settings modal but in a more generic settings modal. The *trusted senders* are actually similar to this, we could move those two into a new settings modal.

- [x] Requires a restructured Vuex store for central tag mutations https://github.com/nextcloud/mail/issues/4871

cc @nimisha-vijay for design input :)

sutidor:
Cool feature. How can I remove or edit (color, name) a tag?

ChristophWurst:
> Cool feature. How can I remove or edit (color, name) a tag?

https://help.nextcloud.com/c/apps/mail/35 is the place for questions :v:

PR #4894, opened 2021-04-06
Show tags

GretaD:
Fixes https://github.com/nextcloud/mail/issues/4805

- [x] Show the labels if theres labels added by another mail client
- [x] Show the right color to each tag
- [x] Fix the styling
- [x] Use ListItem component instead of divs
- [x] Improve styling on small screen. Its okey-ish how its now but can be better

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@nextcloud/logger": "^2.0.0",
     "@nextcloud/moment": "^1.1.1",
     "@nextcloud/router": "^2.0.0",
-    "@nextcloud/vue": "^3.9.0",
+    "@nextcloud/vue": "^4.0.0-alpha.4",
     "@nextcloud/vue-dashboard": "^2",
     "@riophae/vue-treeselect": "^0.4.0",
     "@vue/babel-preset-app": "^4.5.13",
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -1,5 +1,5 @@
 <template>
-	<router-link
+	<ListItem
 		v-draggable-envelope="{
 			accountId: data.accountId ? data.accountId : mailbox.accountId,
 			mailboxId: data.mailboxId,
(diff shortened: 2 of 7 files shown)


ChristophWurst:
you can rebase now :)

src/components/Envelope.vue
```diff
@@ -314,9 +325,25 @@ export default {
...
+	}
 
 	.app-content-list-item-details {
 		padding-right: 7px;
 	}
 }
+.tagLabel {
+	z-index: 1;
+	position: relative;
+	border: 1px solid #FFC300;
```
GretaD, line 339: not the right styling, i did it just to see where my tags are, will fix it

GretaD:
taglabel

ChristophWurst:
> The state right now

@nimisha-vijay :eyes: :)

GretaD:
look at this beauty

taggin

ma12-co (commented):
Nice one :) 
Just one question, don't you think that those tags are a bit to prominent? Both in terms of size and vibrancy of the colors? 
cc @jancborchardt and @nimisha-vijay

nimisha-vijay:
This looks so cool! :D I agree that the colours are a tiny bit too intense. Maybe we can do something similar to Deck?
image

Or have it with a full opacity border but lower opacity fill? (dunno if this is used in any other apps)
image

What do you guys think? @ma12-co @ChristophWurst @jancborchardt

ma12-co:
I like deck solution but I also like that in deck the font-size of the tag is smaller than the title. Looking at Greta's screenshot, the tags are much more prominent than sender's name and subject, which is not good. So on top of making colors lighter, I'd shrink them a bit too.

nimisha-vijay:
Yes correct, decreasing text size sounds good. Maybe also decreasing the padding in the tag by a couple of pixels might make it appear less prominent. 
And maybe align the left of the tags list with the text? So that there is nothing vertically in between the avatars. Does that make sense?

ma12-co:
> Yes correct, decreasing text size sounds good. Maybe also decreasing the padding in the tag by a couple of pixels might make it appear less prominent.

Agree with reducing padding too!

> And maybe align the left of the tags list with the text? So that there is nothing vertically in between the avatars. Does that make sense?

Not sure about this one, might create a weird empty spot to the right, I'd have to see that in the a list to see how that would work..

src/components/Envelope.vue
```diff
@@ -223,100 +343,133 @@ export default {
...
+	padding-right: 7px;
+	}
+	}
+	.tagLabel {
+	margin: 0 10px;
+	}
+	.tagGroup {
+	display: inline-block;
+	border: 1px solid transparent;
+	border-radius: 35px;
```
ma12-co, line 448:
```suggestion
	border-radius: var(--border-radius-pill);
```

src/components/Envelope.vue
```diff
@@ -223,100 +343,133 @@ export default {
...
 
 	.app-content-list-item-details {
-		padding-right: 7px;
+	padding-right: 7px;
+	}
+	}
+	.tagLabel {
+	margin: 0 10px;
+	}
+	.tagGroup {
```
ma12-co, line 445:
```suggestion
	.tagGroup {
	font-size: calc(var(--default-font-size) * 0.8)
```

ma12-co (changes requested):

GretaD:
> Or have it with a full opacity border but lower opacity fill? (dunno if this is used in any other apps)
> image
> 


@nimisha-vijay what are the codes for these colour? background and border. Jan agreed to have this background, but not border. But i need the border colour to add it to the text.

nimisha-vijay:
image

Important: #FF7A66
Todo: #317CCC
Finished: #31CC7C
Personal: #A85BF7
All at 100% opacity for text, 15% for background. 
Later: #DFCC54 at 15% opacity for bg, and #7E6D00 at 100% opacity for text

cc @jancborchardt looks good?

ChristophWurst:
> Important: #FF7A66
> Todo: #317CCC
> Finished: #31CC7C
> Personal: #A85BF7
> All at 100% opacity for text, 15% for background.
> Later: #DFCC54 at 15% opacity for bg, and #7E6D00 at 100% opacity for text

We can work with some wisely picked defaults for the five *default tags*. But after #4806 the users will be able to assign *any* color to their tags. Will the opacity for text/background work with that?

nimisha-vijay:
As long as there is enough contrast between text and bg, it should be fine. Lighter colours will probably not work because the text would not be visible (like the yellow tag, later, used a different text and bg colour because otherwise it would not be visible). 
If we are using a colour picker palette, maybe we can explicitly specify values for text and bg colour for each colour in the palette to make sure there are no visiblity/contrast issues?

ChristophWurst:
> [ @GretaD ](/GretaD)
> ` [fixup! Merge branch 'master' into enhancement/show_tag](/nextcloud/mail/pull/4894/commits/d6e5e5798948a61f6edd23d86da41248ad4d3c1f)`

:warning: is this really a fixup of the *merge*? Please double check. The rebase might drop your changes otherwise.

package.json
```diff
@@ -39,7 +39,7 @@
     "@nextcloud/logger": "^1.1.2",
     "@nextcloud/moment": "^1.1.1",
     "@nextcloud/router": "^1.2.0",
-    "@nextcloud/vue": "^3.9.0",
+    "@nextcloud/vue": "^4.0.0-alpha.3",
```
ChristophWurst, line 42: v4 will pull in `@nextcloud/router` v2 and that will drop some of the browsers that we still support or which are still supported for Nextcloud 20-21

ChristophWurst (changes requested):
Making sure we don't merge this lightly

package.json
ChristophWurst: We decided to drop older browsers.

ChristophWurst:
- [ ] BUG color indicator should extend to the full height

Bildschirmfoto von 2021-05-19 13-36-51

ChristophWurst:
- [ ] BUG important messages have a large invisible area at which they become clickable. If there is a message blow, you can't select that one

Bildschirmfoto vom 2021-05-19 13-37-47

ChristophWurst (approved):
Works :v:

PR #4896, opened 2021-04-06
Avoid double testing with pushes

skjnldsv:


kesselb (approved):
Looks good to me :+1:

PR #4898, opened 2021-04-06
Signatures for aliases

### LINKED ISSUE #4457: Allow a signature per alias
### Feature Request

Currently I can pick one signature per account, but I'd like to also have the ability to select one per alias.

#### Summary

n/a

- [x] Requires https://github.com/nextcloud/mail/issues/1660

kesselb:
Fix #4457 

| > 1 identity = select | 1 identity = no select |
| --- | --- |
| Screenshot from 2021-04-06 21-30-32 | Screenshot from 2021-04-06 21-34-08 |

**Todo**

- [x] Edge case: Update currently open draft with new signature
- [x] Unit-Tests

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -11,7 +11,6 @@
 				label="name"
 				track-by="selectId"
 				:searchable="false"
-				:hide-selected="true"
 				:custom-label="formatAliases"
 				:placeholder="t('mail', 'Select account')"
 				:clear-on-select="false"
@@ -412,7 +411,7 @@ export default {
 						aliasId: alias.id,
 						selectId: cnt++,
 						editorMode: account.editorMode,
-						signature: account.signature,
+						signature: alias.signature,
 						name: alias.name,
 						emailAddress: alias.alias,
 					}
@@ -461,6 +460,21 @@ export default {
 		},
 		allRecipients() {
 			this.checkRecipientsKeys()
+		},
+		aliases(newAliases) {
+			console.debug('aliases changed')
+			if (this.selectedAlias === NO_ALIAS_SET) {
+				return
+			}
+
+			const newAlias = newAliases.find(alias => alias.id === this.selectedAlias.id && alias.aliasId === this.selectedAlias.aliasId)
+			if (newAlias === undefined) {
+				// selected alias does not exist anymore.
+				this.onAliasChange(newAliases[0])
+			} else {
+				// update the selected alias
+				this.onAliasChange(newAlias)
+			}
 		},
 	},
 	async beforeMount() {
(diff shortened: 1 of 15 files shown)


src/service/AliasService.js
```diff
@@ -26,3 +23,19 @@ export const deleteAlias = async(account, alias) => {
 
 	return axios.delete(url).then((resp) => resp.data)
 }
+
+class AliasService {
```
kesselb, line 27: @ChristophWurst can we use classes?

kesselb:
>  Edge case: Update currently open draft with new signature

Is more complex then expected. I tried to write a watcher for the signature property. The property is deeply nested in state.accounts.

src/service/AliasService.js
ChristophWurst: possible but typically not required. what would be the benefit here?

kesselb: it's much nicer ;)

src/components/Composer.vue
```diff
@@ -462,6 +461,21 @@ export default {
...
+		aliases(newAliases) {
+			console.debug('aliases changed')
+			if (this.selectedAlias === NO_ALIAS_SET) {
+				return
+			}
+
+			const newAlias = newAliases.find(alias => alias.id === this.selectedAlias.id && alias.aliasId === this.selectedAlias.aliasId)
+			if (newAlias === undefined) {
+				// selected alias does not exist anymore.
+				this.onAliasChange(newAliases[0])
```
kesselb, line 473: I wonder if newAliases could be empty. We have the default "alias" from the email address.

ChristophWurst: we can defensively add a check

src/components/Composer.vue
```diff
@@ -462,6 +461,21 @@ export default {
 		allRecipients() {
 			this.checkRecipientsKeys()
 		},
+		aliases(newAliases) {
+			console.debug('aliases changed')
+			if (this.selectedAlias === NO_ALIAS_SET) {
+				return
+			}
+
+			const newAlias = newAliases.find(alias => alias.id === this.selectedAlias.id && alias.aliasId === this.selectedAlias.aliasId)
```
ChristophWurst, line 470: :thinking: :thinking: :thinking: 

so ``alias.id`` is the *account id* and ``alias.aliasId`` is the *alias id*?

kesselb: image

Yes

src/components/Composer.vue
kesselb: should be fine without. the account is always there.

src/components/Composer.vue
ChristophWurst: could we flip that so that ``alias.id`` is the *alias id?

and for the comparison it should be sufficient to compare the alias IDs as those are unique, right?

ChristophWurst (approved):
Smoooooooooooooth! Works super well :shipit: 

Let's do the account id vs alias id in another PR, ok?

PR #4899, opened 2021-04-06
Cleanup internal alias api

kesselb:
Follow up to https://github.com/nextcloud/mail/pull/4864

A small cleanup as I messed up our internal api a bit :see_no_evil: 

The AliasService should not return an client exception because they are "bound" to http requests but forward the original exception to the caller. As the ErrorMiddleware already knows DoesNotExistException no need to catch the exception in AliasController :broom: 

Internal changes => No backport required

### DIFF
--- a/lib/Controller/AliasesController.php
+++ b/lib/Controller/AliasesController.php
@@ -23,10 +23,10 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Controller;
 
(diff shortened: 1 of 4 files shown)


miaulalala (approved):

ChristophWurst:
the cs fixer is not happy ;)

PR #4903, opened 2021-04-07
Change `mail_accounts.provisioned` column to allow null

### LINKED ISSUE #4897: Column "oc_mail_accounts"."provisioned" is type Bool and also NotNull, so it can not store "false".
Installs have been failing because ~~the column `provisioned` in `mail_accounts` is not correctly created.~~ a few columns of different tables are oracle incompatible.

This bug is related to: https://github.com/nextcloud/server/pull/24055

Create a new migration that rewrites ~~the provisioned column to allow null.~~ these columns.

miaulalala:
Fixes #4897

### DIFF
--- a/lib/Db/MailAccount.php
+++ b/lib/Db/MailAccount.php
@@ -63,11 +63,11 @@ use OCP\AppFramework\Db\Entity;
  * @method void setLastMailboxSync(int $time)
  * @method string getEditorMode()
  * @method void setEditorMode(string $editorMode)
- * @method bool getProvisioned()
+ * @method bool|null getProvisioned()
  * @method void setProvisioned(bool $provisioned)
  * @method int getOrder()
  * @method void setOrder(int $order)
- * @method bool getShowSubscribedOnly()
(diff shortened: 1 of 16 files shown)


ChristophWurst (approved):

lib/Db/MailAccount.php
```diff
@@ -186,7 +186,7 @@ public function __construct(array $params = []) {
 		$this->addType('inboundPort', 'integer');
 		$this->addType('outboundPort', 'integer');
 		$this->addType('lastMailboxSync', 'integer');
-		$this->addType('provisioned', 'bool');
+		$this->addType('provisioned', 'boolean');
```
kesselb, line 189: fyi `bool` and `boolean` should work: https://www.php.net/manual/en/function.settype.php

ChristophWurst: Yes, we discussed this and concluded some consistency would be nice :)

kesselb: fine by me but weird to use boolean because for method signatures we need bool ;)

PR #4909, opened 2021-04-08
add option to place signature above quoted text

zimmersi:
Signed-off-by: Silvio Zimmer <silvio.zimmer@die-kinderwelt.com>


lib/Migration/Version1100Date20210408091323.php
```diff
@@ -0,0 +1,34 @@
...
+	 * @return ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$accountsTable = $schema->getTable('mail_accounts');
+		
+		$accountsTable->addColumn('signature_above_quote', 'boolean', [
+			'notnull' => true,
```
kesselb, line 28:
```suggestion
			'notnull' => false,
```

notnull for bool on oracle has some issues.

src/ckeditor/signature/InsertSignatureCommand.js
```diff
@@ -54,24 +61,47 @@ export default class InsertSignatureCommand extends Command {
...
 		)
+
+		// Iterate over all items in this range:
+		for (const value of range.getWalker({ shallow: true })) {
+			if (value.item.is('element') && value.item.name === 'quote') {
+				return writer.createPositionBefore(value.item)
+			}
+		}
+
+		return writer.createPositionAt(editor.model.document.getRoot(), 1)
```
kesselb, line 95:
```suggestion
		return writer.createPositionAt(editor.model.document.getRoot(), 'end')
```

If we reach this point there is no quote right? We should append the signature to the end then.

kesselb (commented):
Thanks again for this great contribution :+1: 

We have some conflicts again with https://github.com/nextcloud/mail/pull/4898. I hope to finish the signature for aliases feature this week. Then it's safe to continue with this feature.

image

An option to place the signature above the quoted text looks more like a configuration for all accounts. Similar to put my text to the bottom of a reply instead on top of it. 

Is the position of the signature something you change per account?

lib/Migration/Version1100Date20210408091323.php
zimmersi: thanks for pointing this out 👍

src/ckeditor/signature/InsertSignatureCommand.js
zimmersi: yes, sounds reasonable

zimmersi:
> An option to place the signature above the quoted text looks more like a configuration for all accounts. Similar to put my text to the bottom of a reply instead on top of it.
> 
> Is the position of the signature something you change per account?

well, for now it is per account. could be possible that someone has different requirements for private/business accounts...I am not sure how realistic this is

kesselb:
> well, for now it is per account. could be possible that someone has different requirements for private/business accounts...I am not sure how realistic this is

@miaulalala, @ChristophWurst and I talked about this today. An option per account is good. 

I've sent you an invite to the nextcloud organization. One advantage is to create branches with our repositories. However it's totally fine to continue developing on your fork. I forgot to signoff the code suggestions. You may rebase the branch and squash the code suggestion into your commit. No need to keep my name or email on it ;)

I will ping you after we merged the signature for aliases feature. I'm still thinking about the watcher. We may use a similar approach as https://github.com/nextcloud/mail/pull/4715/files#diff-77b33550ce11bdb5cd04cb76105da60c54610d3c204ab3c20f513e2336a66d51. `:value` and `@input` to directly change the state.

zimmersi:
allright, thks for the invitation! 👍 ❤️

kesselb:
Hi,

could you rebase your feature branch with master?

zimmersi:
ok, done.
now we have all the commits in the history of this request. is this normal after rebase?

kesselb:
> now we have all the commits in the history of this request. is this normal after rebase?

Hmm. Did you force push after the rebase? I usually do 
```
git rebase -i master
git push --force 
```

zimmersi:
hmm, no, I did the following:
git pull origin place-signature-above-quoted-text
git fetch upstream 
git rebase upstream/master 
then I tried git push origin place-signature-above-quoted-text but this failed due to error: failed to push some refs
so I did again git pull origin place-signature-above-quoted-text
then commit and push....

sry, I m a git newbie, but I can create a new branch & pull request if needed

kesselb:
> sry, I m a git newbie, but I can create a new branch & pull request if needed

No problem. I will apply the diff (https://patch-diff.githubusercontent.com/raw/nextcloud/mail/pull/4909.diff) later the day and give you some feedback on the watcher. 

I tried to fix the commit history locally but not idea how :see_no_evil:

src/components/SignatureSettings.vue
```diff
@@ -59,32 +80,80 @@ export default {
 		return {
 			loading: false,
 			bus: new Vue(),
+			identity: null,
+			signature: '',
+			signatureAboveQuoteToggle: this.account.signatureAboveQuote,
```
kesselb, line 85: let's call it signatureAboveQuote

src/components/SignatureSettings.vue
```diff
@@ -59,32 +80,80 @@ export default {
...
+				id: -1,
+				label: this.account.name + ' (' + this.account.emailAddress + ')',
+				signature: this.account.signature,
+			})
+
+			return identities
+		},
+	},
+	watch: {
+		signatureAboveQuoteToggle(val, oldVal) {
```
kesselb, line 108:
```
		async signatureAboveQuote(val, oldVal) {
			try {
				await this.$store.dispatch('patchAccount', {
					account: this.account,
					data: {
						signatureAboveQuote: val,
					},
				})
				logger.debug('signature above quoted updated to ' + val)
			} catch (e) {
				logger.error('could not update signature above quote', { e })
				this.signatureAboveQuote = oldVal
			}
		},
```

we can use async/await inside the watcher. that's a bit nicer to read.

kesselb (commented):
Thanks :+1: 

Works for me. Some smaller nitpicks. Please send a new pull request with a clean git history.

zimmersi:
Thanks!
I have already committed your proposed changes here. 
Will create a new pull request now.

zimmersi:
here is the new one #4971

PR #4910, opened 2021-04-08
Split migrations for backport

miaulalala:
Follow up to #4903


lib/Migration/Version1095Date20210407150016.php
```diff
@@ -0,0 +1,109 @@
...
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
```
ChristophWurst, line 13: :fire:

lib/Migration/Version1100Date20210304143008.php
```diff
@@ -78,13 +78,6 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 				'length' => 4,
 			]);
 			$tagsMessageTable->setPrimaryKey(['id']);
-			$tagsMessageTable->addUniqueIndex(
```
ChristophWurst, line 81: unrelated change, right?

ChristophWurst (changes requested):

lib/Migration/Version1095Date20210407150016.php
miaulalala: every time

lib/Migration/Version1100Date20210304143008.php
miaulalala: that should have been resolved by the rebase. I'm sorry - shall I leave it? It's in the rework tagging branch as well

ChristophWurst: right, that was also in https://github.com/nextcloud/mail/pull/4840. so actually this PR doesn't drop anything anymore as the code is already gone. rebase after your fixup and it should be all good :)

lib/Migration/Version1100Date20210406144257.php
```diff
@@ -21,81 +21,13 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
-			'notnull' => false,
-			'default' => false
-		]);
-		$table->changeColumn('structure_analyzed', [
-			'notnull' => false,
-			'default' => false
-		]);
+		if ($schema->hasTable('mail_accounts')) {
+			$table = $schema->getTable('mail_accounts');
+			$table->changeColumn('sieve_enabled', [
```
ChristophWurst, line 26: but that column already existed in v1.9. Shouldn't this be moved to the other migration and instead here you change the mail_tags.is_default_tag?

miaulalala: ah yes, sieve was released with 1.9, i forgot.
Do I even need the is_default_tag? Since this is in the 1.10 migration which wasn't released yet it should be fine to just change the create table statement.

st3iny (approved):
Checking out stable1.9 and adding `lib/Migration/Version1095Date20210407150016.php` manually will fix installing on nextcloud server master. 

Nit: Rename `lib/Migration/Version1095Date20210407150016.php` to `lib/Migration/Version1096Date20210407150016.php` because v1.9.5 is already released.

lib/Migration/Version1095Date20210407150016.php
```diff
@@ -0,0 +1,114 @@
...
+				'default' => false
+			]);
+			$table->changeColumn('structure_analyzed', [
+				'notnull' => false,
+				'default' => false
+			]);
+		}
+
+		if ($schema->hasTable('mail_accounts')) {
+			$table = $schema->getTable('mail_accounts');
```
ChristophWurst, line 105: move that one up to the other mail_accounts block?

ChristophWurst (commented):
Let's rename to Version1096 indeed to take some of the potential confusion around a backported migration :)

PR #4912, opened 2021-04-08
Add an array_flat_map function

ChristophWurst:
More functional and more expressive.

This is my fancy code moment of 2021-04-08.

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -47,6 +47,7 @@ use OCA\Mail\Service\Search\FlagExpression;
 use OCA\Mail\IMAP\Threading\DatabaseMessage;
 use OCA\Mail\Support\PerformanceLogger;
 use OCP\AppFramework\Db\DoesNotExistException;
+use function OCA\Mail\array_flat_map;
 
 /**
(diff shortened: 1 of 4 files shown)


miaulalala (approved):
Fancy fancy

lib/Db/MessageMapper.php
```diff
@@ -178,15 +179,10 @@ public function findUidsForIds(Mailbox $mailbox, array $ids) {
...
-		$results = [];
-		foreach ($chunks as $chunk) {
+		return array_flat_map(function(array $chunk) use ($query) {
 			$query->setParameter('ids', $chunk, IQueryBuilder::PARAM_INT_ARRAY);
-			$results[] = $this->findUids($query);
-		}
-
-		return array_merge(...$results);
+			return $this->findUids($query);
+		}, array_chunk($ids, 1000));
```
nickvergessen, line 185: move array_chunk also inside?

ChristophWurst: inside the new function? Nah, that is too specific for this use case

rullzer:
Look at mr fancy pants over here

PR #4928, opened 2021-04-09
Run dependabot-approve-merge only for dependabot branches.

kesselb:
Screenshot from 2021-04-09 13-01-02

Was looking for some CI results and noticed that the auto approve action run's for every branch. I think we can limit the workflow to branches starting with dependabot.

Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

### DIFF
--- a/.github/workflows/dependabot-approve-merge.yml
+++ b/.github/workflows/dependabot-approve-merge.yml
@@ -1,5 +1,8 @@
 name: Dependabot
-on: pull_request_target
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

kesselb:
fyi @skjnldsv

kesselb:
Does not work :(

skjnldsv:
careful, with `pull_request_target`

PR #4932, opened 2021-04-09
Move recipient delete to separate query

rullzer:
The sub query did not use proper indexes. So now we just fetch the data
and then issue a separate delete statement that can use all the right
indexes. Resulting esp on large setups on a lot better performance.

Now truth be told I did not test this.
Lets see if it is covered by tests :P

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -482,14 +482,22 @@ class MessageMapper extends QBMapper {
 
 	public function deleteAll(Mailbox $mailbox): void {
 		$messageIdQuery = $this->db->getQueryBuilder();
-		$deleteRecipientsQuery = $this->db->getQueryBuilder();
 		$messageIdQuery->select('id')
 			->from($this->getTableName())
-			->where($messageIdQuery->expr()->eq('mailbox_id', $deleteRecipientsQuery->createNamedParameter($mailbox->getId())));
+			->where($messageIdQuery->expr()->eq('mailbox_id', $messageIdQuery->createNamedParameter($mailbox->getId())));
+
+		$cursor = $messageIdQuery->execute();
+		$messageIds = $cursor->fetchAll();
+		$cursor->closeCursor();
+
+		$messageIds = array_map(function (array $row) {
+			return (int)$row['id'];
+		}, $messageIds);
 
 		// delete all related recipient entries
+		$deleteRecipientsQuery = $this->db->getQueryBuilder();
 		$deleteRecipientsQuery->delete('mail_recipients')
-			->where($deleteRecipientsQuery->expr()->in('message_id', $deleteRecipientsQuery->createFunction($messageIdQuery->getSQL()), IQueryBuilder::PARAM_INT_ARRAY));
+			->where($deleteRecipientsQuery->expr()->in('message_id', $deleteRecipientsQuery->createNamedParameter($messageIds, IQueryBuilder::PARAM_INT_ARRAY)));
 		$deleteRecipientsQuery->execute();
 
 		$query = $this->db->getQueryBuilder();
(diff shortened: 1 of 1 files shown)


lib/Db/MessageMapper.php
```diff
@@ -487,9 +487,17 @@ public function deleteAll(Mailbox $mailbox): void {
 			->from($this->getTableName())
 			->where($messageIdQuery->expr()->eq('mailbox_id', $deleteRecipientsQuery->createNamedParameter($mailbox->getId())));
 
+		$cursor = $messageIdQuery->execute();
+		$messageIds = $cursor->fetchAll();
```
kesselb, line 491:
```suggestion
		$messageIds = $cursor->fetchFirstColumn();
```

maybe? no array_map then.

ChristophWurst: that didn't exist with Nextcloud 20 but only after the migration of dbal in 21

lib/Db/MessageMapper.php
```diff
@@ -487,9 +487,17 @@ public function deleteAll(Mailbox $mailbox): void {
 			->from($this->getTableName())
 			->where($messageIdQuery->expr()->eq('mailbox_id', $deleteRecipientsQuery->createNamedParameter($mailbox->getId())));
```
ChristophWurst, line 488:
```suggestion
			->where($messageIdQuery->expr()->eq('mailbox_id', $messageIdQuery->createNamedParameter($mailbox->getId())));
```

I'm pretty sure the parameter won't work otherwise. You always have to use the one that you use to execute the query.

lib/Db/MessageMapper.php
```diff
@@ -510,9 +518,18 @@ public function deleteByUid(Mailbox $mailbox, int ...$uids): void {
 				$messageIdQuery->expr()->eq('mailbox_id', $deleteRecipientsQuery->createNamedParameter($mailbox->getId())),
```
ChristophWurst, line 518: same remark as above

ChristophWurst (changes requested):
In theory we have to be carefuly with IN and literals but I guess more than 1000 recipients per messages is rather unlikely

rullzer:
More reviews

lib/Db/MessageMapper.php
```diff
@@ -482,14 +482,22 @@ public function updatePreviewDataBulk(Message ...$messages): array {
...
+
+		$messageIds = array_map(function (array $row) {
+			return (int)$row['id'];
+		}, $messageIds);
 
 		// delete all related recipient entries
+		$deleteRecipientsQuery = $this->db->getQueryBuilder();
 		$deleteRecipientsQuery->delete('mail_recipients')
-			->where($deleteRecipientsQuery->expr()->in('message_id', $deleteRecipientsQuery->createFunction($messageIdQuery->getSQL()), IQueryBuilder::PARAM_INT_ARRAY));
+			->where($deleteRecipientsQuery->expr()->in('message_id', $messageIds, IQueryBuilder::PARAM_INT_ARRAY));
```
kesselb, line 500:
```suggestion
			->where($deleteRecipientsQuery->expr()->in('message_id', $deleteRecipientsQuery->createNamedParameter($messageIds, IQueryBuilder::PARAM_INT_ARRAY)));
```

lib/Db/MessageMapper.php
```diff
@@ -502,17 +510,25 @@ public function deleteAll(Mailbox $mailbox): void {
...
+
+		$messageIds = array_map(function (array $row) {
+			return (int)$row['id'];
+		}, $messageIds);
+
 		// delete all related recipient entries
+		$deleteRecipientsQuery = $this->db->getQueryBuilder();
 		$deleteRecipientsQuery->delete('mail_recipients')
-			->where($deleteRecipientsQuery->expr()->in('message_id', $messageIdQuery->createFunction($messageIdQuery->getSQL()), IQueryBuilder::PARAM_INT_ARRAY));
+			->where($deleteRecipientsQuery->expr()->in('message_id', $messageIds, IQueryBuilder::PARAM_INT_ARRAY));
```
kesselb, line 531:
```suggestion
			->where($deleteRecipientsQuery->expr()->in('message_id', $deleteRecipientsQuery->createNamedParameter($messageIds, IQueryBuilder::PARAM_INT_ARRAY)));
```

kesselb (changes requested):

miaulalala (approved):

PR #4933, opened 2021-04-09
Fixes 2 warning  `<template>` require directive  vue/no-lone-template lint warnings

StCyr:


st3iny (approved):
Code looks good. Recipient bubbles and mailbox context menus still work correctly.

According to the [eslint-plugin-vue documentation](https://eslint.vuejs.org/rules/no-lone-template.html) we can drop those template tags safely: `In Vue.js 2.x, the <template> elements that have no specific directives have no effect.`

PR #4946, opened 2021-04-13
Backport #4910 to 1.9

miaulalala:

### DIFF
--- a/lib/Db/MailAccount.php
+++ b/lib/Db/MailAccount.php
@@ -63,11 +63,11 @@ use OCP\AppFramework\Db\Entity;
  * @method void setLastMailboxSync(int $time)
  * @method string getEditorMode()
  * @method void setEditorMode(string $editorMode)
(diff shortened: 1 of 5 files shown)


ChristophWurst:
looks good!

should we possibly also backport the change of the documented entity property types so that they are clearly nullable?

ChristophWurst (changes requested):
Code looks fine but as we learned today the nullable bollean columns actually work on Oracle. I assume we'll revert that change in server and then this becomes obsolete.

ChristophWurst:
> Code looks fine but as we learned today the nullable bollean columns actually work on Oracle. I assume we'll revert that change in server and then this becomes obsolete.

Looks like we won't revert.

So I guess this is the way to go. The only concern I have is that we still migrate a potentially large database. I can't judge if database will perform the column type quickly.

ChristophWurst (approved):
Code looks still good

PR #4952, opened 2021-04-15
Fix/filter forwarded and nonjunk tag

### LINKED ISSUE #4951: nonjunk and forwarded should not show as tags
### Expected behavior
It is shown as flag and not tag

### Actual behavior
database

### Mail app

**Mail app version:** master

cc @miaulalala

miaulalala:
Thunderbird sets a non- standard flag `nonjunk` on messages that are not junk. Additionally, some standard flags like `$forwarded` are missing.

Fixes #4951

### DIFF
--- a/lib/Model/IMAPMessage.php
+++ b/lib/Model/IMAPMessage.php
@@ -717,22 +717,26 @@ class IMAPMessage implements IMessage, JsonSerializable {
 			in_array(Horde_Imap_Client::FLAG_JUNK, $flags, true) ||
 			in_array('junk', $flags, true)
 		);
-		$msg->setFlagNotjunk(in_array(Horde_Imap_Client::FLAG_NOTJUNK, $flags, true));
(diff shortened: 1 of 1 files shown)


GretaD (approved):
tested it and it works just good

lib/Model/IMAPMessage.php
```diff
@@ -761,7 +765,7 @@ private function generateTagEntites(array $tags, string $userId): array {
 		$t = [];
 		foreach ($tags as $keyword) {
 			// Map the old $important to $label1 until we have caught them all
-			if ($keyword === '$important') {
+			if ($keyword === '$important' || $keyword === 'important') {
```
ChristophWurst, line 768: didn't we want to leave that one out for now until we know *important* is just a remnant from an earlier test?

miaulalala: I understood it as the other way around!

ChristophWurst (approved):

PR #4953, opened 2021-04-15
Remove Jans test (for now)

miaulalala:
Due to the test failing constantly currently, let's remove it and revert this commit at a future date.


ChristophWurst (approved):
Was nice knowing you little test. You served us well.

PR #4954, opened 2021-04-15
Revert "Remove Jans test (for now)"

miaulalala:
Reverts nextcloud/mail#4953

Merge this when we have a way to avoid running into the same issue again (being blocked by the Gravatar API)


ChristophWurst (approved):
Can be merged as soon as it's passing again

PR #4947, opened 2021-04-13
Chunking `in` statements for Oracle Compliance

miaulalala:
Oracle can't deal with more than 1000 values in an expression list for `in` queries.

Chunk all `in` queries so they'll use a max of 1000 values.

Fixes #4941

PR #4970, opened 2021-04-20
Only show each thread once per message list

### LINKED ISSUE #3652: Only show each thread once per message list
Right now we still show the actual mailbox' contents and list messages. Threads are only shown when you open an individual message.

We should change the list to show *threads* in a sense that messages of the same thread are only shown once and at the position of the latest message.

kesselb:
Fixes #3652 

**Todo**

- [x] Endpoint to delete / move a thread
- [x] Delete in envelope list should delete all messages (excl. trash)
- [x] Move in envelope list should move all messages (excl. trash)
- [x] Delete the latest message in thread view should add the second latest message to the envelope list<sup>1</sup>
- [x] Move the latest message in thread view should add the second latest message to the envelope list<sup>1</sup>
- [x] Select multiple threads/messages in envelope to delete/move must still work
- [x] Keyboard shortcuts
- [x] Rename delete to delete thread in envelope list
- [x] Rename move to move thread in envelope list

<sup>1</sup> Is done by the existing sync backend request

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,7 +71,7 @@
       },
       "devDependencies": {
         "@babel/core": "^7.14.3",
-        "@babel/eslint-parser": "^7.14.3",
+        "@babel/eslint-parser": "^7.14.4",
         "@babel/plugin-syntax-dynamic-import": "^7.8.3",
         "@babel/preset-env": "^7.14.2",
         "@nextcloud/browserslist-config": "^2.1.0",
@@ -1179,9 +1179,9 @@
       }
     },
     "node_modules/@babel/eslint-parser": {
-      "version": "7.14.3",
-      "resolved": "https://registry.npmjs.org/@babel/eslint-parser/-/eslint-parser-7.14.3.tgz",
-      "integrity": "sha512-IfJXKEVRV/Gisvgmih/+05gkBzzg4Dy0gcxkZ84iFiLK8+O+fI1HLnGJv3UrUMPpsMmmThNa69v+UnF80XP+kA==",
+      "version": "7.14.4",
+      "resolved": "https://registry.npmjs.org/@babel/eslint-parser/-/eslint-parser-7.14.4.tgz",
+      "integrity": "sha512-7CTckFLPBGEfCKqlrnJq2PIId3UmJ5hW+D4dsv/VvuA5DapgqyZFCttq+8oeRIJMZQizFIe5gel3xm2SbrqlYA==",
       "dev": true,
       "dependencies": {
         "eslint-scope": "^5.1.0",
@@ -23453,9 +23453,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.14.3",
-      "resolved": "https://registry.npmjs.org/@babel/eslint-parser/-/eslint-parser-7.14.3.tgz",
-      "integrity": "sha512-IfJXKEVRV/Gisvgmih/+05gkBzzg4Dy0gcxkZ84iFiLK8+O+fI1HLnGJv3UrUMPpsMmmThNa69v+UnF80XP+kA==",
+      "version": "7.14.4",
+      "resolved": "https://registry.npmjs.org/@babel/eslint-parser/-/eslint-parser-7.14.4.tgz",
+      "integrity": "sha512-7CTckFLPBGEfCKqlrnJq2PIId3UmJ5hW+D4dsv/VvuA5DapgqyZFCttq+8oeRIJMZQizFIe5gel3xm2SbrqlYA==",
       "dev": true,
       "requires": {
         "eslint-scope": "^5.1.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   ],
   "devDependencies": {
     "@babel/core": "^7.14.3",
-    "@babel/eslint-parser": "^7.14.3",
+    "@babel/eslint-parser": "^7.14.4",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@babel/preset-env": "^7.14.2",
     "@nextcloud/browserslist-config": "^2.1.0",


lib/Db/MessageMapper.php
```diff
@@ -664,9 +665,9 @@ public function findIdsGloballyByQuery(IUser $user, SearchQuery $query, ?int $li
 		$qbMailboxes = $this->db->getQueryBuilder();
 
 		$select = $qb
-			->selectDistinct('m.id')
-			->addSelect('m.sent_at')
-			->from($this->getTableName(), 'm');
+			->selectDistinct(['m.id', 'm.sent_at'])
+			->from($this->getTableName(), 'm')
+			->leftJoin('m', $this->getTableName(), 'm2', 'm.mailbox_id = m2.mailbox_id and m.thread_root_id = m2.thread_root_id and m.id < m2.id');
```
ChristophWurst, line 670: shouldn't the join condition compare sent_at instead of the id?

ChristophWurst:
> Delete the latest message in thread view should add the second latest message to the envelope list

As discussed last week those operations *might* not actually need special handling. The existing sync logic should be able to see that a new message is "added" to the list.

What we might have to change is the sync logic for *new messages*. Right now we consider all messages with a higher UID than highest known UID as new. What we can do instead is consider all messages as new whose UID is higher than the *lowest* known UID and not in the list of known UIDs already. Then you can also detect "new" messages that pop up in the middle of the previous message list.

lib/Db/MessageMapper.php
```diff
@@ -912,7 +914,7 @@ public function findNewIds(Mailbox $mailbox, array $ids): array {
 		$sub = $this->db->getQueryBuilder();
 
 		$subSelect = $sub
-			->select($sub->func()->max('uid'))
+			->select($sub->func()->max('sent_at'))
```
kesselb, line 917: Note to myself: A message moved from A to B get's a new uid.

Delete the last message from a thread and delete the second last message afterwards. In the trash the second last appears before the last because the uid is higher.

kesselb:
https://user-images.githubusercontent.com/3902676/117020085-63ba0600-acf6-11eb-8ae2-516eb98b75c2.mp4

**Case 1**: Delete the latest message in a thread. The message is removed from the envelope list but none of the other messages show up. 

I'm not sure yet if changing the findNewIds method is possible. Right now we take the known id's, find the max uid (on master) / max sent_at (this branch) and query for all messages > max uid/sent_at. 

As suggest we may change max to min uid/sent_at of the known id's. This will not work for my example case as the known id's are a few. However if a user has more than 20 messages it should work. 

In addition we could lookup the envelopes store on message delete if there is another message of the same thread and add it to visible envelopes list.

kesselb:
**View**
Inbox (and other mailboxes but not trash): Show messages in the thread view but hide messages in trash. 
Trash: Show messages in thread view that are in the trash. Hide messages from other mailboxes.
Hiding messages is done on client. 

**Delete**
Inbox (and other mailboxes but not trash): Delete a thread should move all messages (in inbox and other mailboxes but not trash) to trash mailbox.
Trash: Delete a thread should delete the deleted messages of a thread that are in trash.

**Move**
Inbox (and other mailboxes but not trash): Move a thread should move all messages (in inbox and other mailboxes but not trash) to target mailbox.
Trash: Move a thread should move the deleted messages of a thread that are in trash.

src/store/getters.js
```diff
@@ -64,10 +64,26 @@ export const getters = {
 	getMessage: (state) => (id) => {
 		return state.messages[id]
 	},
-	getEnvelopeThread: (state) => (id) => {
-		const thread = state.envelopes[id]?.thread ?? []
-		const envelopes = thread.map(id => state.envelopes[id])
-		return sortBy(prop('dateInt'), envelopes)
+	getEnvelopeThread: (state, getters) => (id) => {
+		if (!state.envelopes[id]) {
```
kesselb, line 68: - [ ] Add test `src/tests/unit/store/getters.spec.js`

tests/Integration/Db/ThreadMapperTest.php
```diff
@@ -0,0 +1,80 @@
...
+		$this->threadMapper = new ThreadMapper(
+			\OC::$server->get(IDBConnection::class)
+		);
+
+		$this->queryLogger = \OC::$server->get(IQueryLogger::class);
+	}
+
+	public function dataFindMessageUidsAndMailboxNamesByAccountAndThreadRoot(): array {
+		return [
+			'inbox' => [false, 'SELECT "messages"."uid", "mailboxes"."name" FROM "oc_mail_messages" "messages" INNER JOIN "oc_mail_mailboxes" "mailboxes" ON messages.mailbox_id = mailboxes.id WHERE ("messages"."thread_root_id" = :dcValue1) AND ("mailboxes"."account_id" = :dcValue2) AND ("mailboxes"."id" <> :dcValue3)'],
```
kesselb, line 53: - [ ] Computer says no ;)

src/components/MoveModal.vue
```diff
@@ -41,18 +45,20 @@ export default {
...
-				if (envelopeIds.length === 0) {
+				if (envelopes.length === 0) {
 					return
 				}
 
-				await Promise.all(envelopeIds.map((id) => this.$store.dispatch('moveMessage', {
-					id,
-					destMailboxId: this.destMailboxId,
-				})))
+				for (const envelope of envelopes) {
```
kesselb, line 55: The change `Promise.all` to `for...of` is not necessary for the threading improvements. `Promise.all` will dispatch a moveMessage / moveThread action for each envelope passed to the MoveModal. Ideally would be a "loading" state to signal that mail is working on something.

lib/Db/MessageMapper.php
```diff
@@ -954,28 +945,33 @@ public function findRelatedData(array $messages, string $userId): array {
...
 			->where(
-				$sub->expr()->eq('mailbox_id', $qb->createNamedParameter($mailbox->getId(), IQueryBuilder::PARAM_INT)),
-				$sub->expr()->in('id', $qb->createParameter('ids'))
+				$subSelect->expr()->eq('mailbox_id', $select->createNamedParameter($mailbox->getId(), IQueryBuilder::PARAM_INT)),
+				$subSelect->expr()->in('id', $knowIdsParameter)
 			);
 
-		$qb->select('id')
-			->from($this->getTableName())
+		$select
```
kesselb, line 961: - [ ] https://github.com/nextcloud/mail/pull/4947 introduced chunking for in expressions. Check the findNewIds query to make sure it still works es expected.

st3iny (commented):
I just tested this PR on my local instance and here are some things I noticed:
* Deleting the most recent message from a thread in the thread envelope view will make the thread disappear form the list. Triggering a sync will make the thread reappear.
* Deleting an older message from a thread manually and then restoring it from the trash bin won't put the envelope back in the thread but show it as a separate entry in the list. (This might be related to how threads are built internally). Honestly, I wouldn't bother fixing this because it shouldn't happen often.
* A thread is still shown after moving it to another folder, switching to a third folder and then back again. It will be shown in both mailboxes a the same time. Moving the thread back to the original folder will show it twice in the list. A page reload will fix the duplication.
* Marking a thread important using the context menu of an envelope in the list will not change the label of the button (`Mark important` -> `Mark unimportant`).

I only looked at the code briefly.

PS: I tested f97fc51a776ddcce12a7b6782f23e5621895f17e. The most recent commit on this branch will trigger an internal server error upon visiting the mail app.

lib/Db/ThreadMapper.php
```diff
@@ -0,0 +1,74 @@
...
+	 * @return array<array-key, array{mailboxName: string, messageUid: int}>
+	 */
+	public function findMessageUidsAndMailboxNamesByAccountAndThreadRoot(MailAccount $mailAccount, string $threadRootId, bool $trash): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('messages.uid', 'mailboxes.name')
+			->from($this->tableName, 'messages')
+			->join('messages', 'mail_mailboxes', 'mailboxes', 'messages.mailbox_id = mailboxes.id')
+			->where(
+				$qb->expr()->eq('messages.thread_root_id', $qb->createNamedParameter($threadRootId, IQueryBuilder::PARAM_STR)),
+				$qb->expr()->eq('mailboxes.account_id', $qb->createNamedParameter($mailAccount->getId(), IQueryBuilder::PARAM_STR))
```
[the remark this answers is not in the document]
kesselb: Thanks!

PR #4971, opened 2021-04-21
add option to place signature above quoted text

zimmersi:
Signed-off-by: Silvio Zimmer <silvio.zimmer@die-kinderwelt.com>

### DIFF
--- a/src/components/SignatureSettings.vue
+++ b/src/components/SignatureSettings.vue
@@ -21,6 +21,16 @@
 
 <template>
 	<div class="section">
+		<div>
+			<input
+				id="signature-above-quote-toggle"
+				v-model="signatureAboveQuote"
+				type="checkbox"
+				class="checkbox">
+			<label for="signature-above-quote-toggle">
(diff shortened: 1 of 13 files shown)


src/components/SignatureSettings.vue
```diff
@@ -93,6 +104,27 @@ export default {
...
+	watch: {
+		async signatureAboveQuote(val, oldVal) {
+			await this.$store
+				.dispatch('patchAccount', {
+					account: this.account,
+					data: {
+						signatureAboveQuote: val,
+					},
+				})
+				.then(() => {
```
ChristophWurst, line 116: watch out, you are mixing async/await with promises.

if you await the dispatch then wrap it with a try-catch for the error handling. you can then use the code as if it was synchronous

zimmersi: omg, thx for pointing this out

kesselb (approved):

kesselb:
Thanks :+1:

zimmersi:
Thanks! 👍

PR #5020, opened 2021-05-11
Changed Horde link to https

thibaultmol:
it's a stupidly small change, but noticed that (for some reason) the link to Horde here didn't use https even though their website seems to work just fine with https


ChristophWurst (approved):
Thanks :)

PR #4829, opened 2021-03-25
Add UI for tag assignment

GretaD:
Fixes https://github.com/nextcloud/mail/issues/4807

- [x] Add a tag 
- [x] Remove a tag
- [x] Polish

PR #5073, opened 2021-05-26
Load accounts and their settings via initial state

ChristophWurst:
We'll also need initial state loading in https://github.com/nextcloud/mail/pull/4829.

There is no behavioral change in the app, we just offload some code to a provided server instead of doing the serialization ourselves.


GretaD (approved):
tested and it works (or at least it doesnt break anything) :)

PR #5079, opened 2021-05-28
Fix the page controller test after #5075

ChristophWurst:
With that darn integration test failling I didn't notice there was a unit test regression in https://github.com/nextcloud/mail/pull/5075. This fixes it again.


ChristophWurst:
I'll merge as soon as the tests pass.

PR #5080, opened 2021-05-28
Adds support for text/calendar attachment

### LINKED ISSUE #4463: MSTeams meatings request attachement are not shown
### Expected behavior
When a mail is meeting request issued by Microsoft Teams, the request should be visible as an attachment so we could add it in our calendar.

### Actual behavior
The HTML text of the meeting request is visible as the mail body with no attachment.

The mail itself is a multipart MIME with 2 base64 encoded parts : 

* Message in text format (ignored by Mail app)
```
Content-Type: text/plain; charset="utf-8"
```
* Same message in HTML format (the one Mail app displays)
```
Content-Type: text/html; charset="utf-8"
```
* The calendar invitation which should be processed as an attachement (ignored by Mail app)
```
Content-Type: text/calendar; charset="utf-8"; method=REQUEST
```

### Mail app

**Mail app version:** 1.8.1

**Mailserver or service:** MS Teams (+ MS Exchange ?)

### Server configuration

**Operating system**: Debian 9

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.3

**Nextcloud Version:** 20.0.7

#### Client configuration
**Browser:** Firefox 85

**Operating system:** Ubuntu

StCyr:
TODO:

- [x] Support adding MSTeam appointments in NC calendar
- [x] Support downloading MSTeam appointments for import in Outlook

Closes #4463

### DIFF
--- a/lib/Attachment.php
+++ b/lib/Attachment.php
@@ -79,10 +79,6 @@ class Attachment {
 
 		$this->mimePart = new \Horde_Mime_Part();
 
-		// To prevent potential problems with the SOP we serve all files with the
-		// MIME type "application/octet-stream"
-		$this->mimePart->setType('application/octet-stream');
-
 		// Serve all files with a content-disposition of "attachment" to prevent Cross-Site Scripting
 		$this->mimePart->setDisposition('attachment');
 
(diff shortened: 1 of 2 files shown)


lib/Attachment.php
```diff
@@ -24,6 +24,7 @@
 
 namespace OCA\Mail;
 
+use function str_starts_with;
```
ChristophWurst, line 27: https://www.php.net/manual/en/function.str-starts-with.php only exists in php8. we're not there yet.

ChristophWurst (changes requested):

lib/Attachment.php
StCyr: ah that's why static analysis was complaining

GretaD (approved):
tested and it works

ChristophWurst (approved):
I'm still a bit afraid that there will be yet another edge case where the handling isn't correct but let's do this. Eventually the logic from IMAPMessage has to be moved somewhere else so we can finally have unit tests to cover all the weird cases automatically and detect regressions.

PR #5104, opened 2021-06-01
Refactor create and update tag

kesselb:


lib/Controller/MessagesController.php
```diff
@@ -671,7 +671,7 @@ public function setTag(int $id, string $imapLabel): JSONResponse {
 
 		try {
 			$tag = $this->mailManager->getTagByImapLabel($imapLabel, $this->currentUserId);
-		} catch (DoesNotExistException $e) {
+		} catch (ClientException $e) {
```
kesselb, line 674: Alternative: Not catching the `ClientException` and let `ErrorMiddleware` handle it. I needed to change the code because the wrong exception was documented.

ChristophWurst (approved):

ChristophWurst:
the linter wants to have a word with you :)

kesselb:
> the linter wants to have a word with you :)

:see_no_evil:

st3iny (approved):
Looks good and linter is happy now.

PR #5105, opened 2021-06-01
Update dependencies

GretaD:


ChristophWurst:
Why all in one and not just merge the Dependabot PRs? We don't need to combine the bumps unless it's ckeditor

GretaD:
> Why all in one and not just merge the Dependabot PRs? We don't need to combine the bumps unless it's ckeditor

when i do it one by one, couple of these show the TypeError: t is null,  when i open the new message or when i start writing smth. And here it is not.

GretaD:
the dependencies were merged individually

PR #5107, opened 2021-06-01
Fix mark (un)important action on envelope list

kesselb:
Reported by @st3iny while reviewing the threading pr. 

1) removeEnvelopeTag was called with tag but expects imapLabel
2) ~important is not a flag and not available as data.flags.important~ say hello again to data.flags.important
3) change find to some to return a boolean if $label1 is true for a given message


ChristophWurst:
> important is not a flag and not available as data.flags.important

Actually it should be. *important* is both a custom flag in our app and a tag that we store as $label1 on IMAP.

ChristophWurst (approved):

PR #5124, opened 2021-06-07
Allow renaming tags

### LINKED ISSUE #5085: Edit tag display name
### Feature Request

As user I would like to change the display name of a tag.

GretaD:
Fixes #5085 #5143

- [x] Exclude default tags from renaming
- [x] Polish

### DIFF
--- a/src/components/TagModal.vue
+++ b/src/components/TagModal.vue
@@ -23,7 +23,7 @@
 	<Modal size="large" @close="onClose">
 		<div class="modal-content">
 			<h2 class="tag-title">
-				{{ t('mail', 'Edit tags') }}
+				{{ t('mail', 'Add default tags') }}
 			</h2>
 			<div v-for="tag in tags" :key="tag.id" class="tag-group">
 				<button class="tag-group__label"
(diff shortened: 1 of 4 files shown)


ChristophWurst:
eslint has found some remarks :P

src/components/TagModal.vue
```diff
@@ -155,6 +177,34 @@ export default {
...
 		},
+		openEditTag() {
+			this.renameTagLabel = false
+			this.renameTagInput = true
+			this.showSaving = false
+
+		},
+		async renameTag(tag, event) {
+			this.renameTagInput = false
+			this.showSaving = false
```
ChristophWurst, line 188: shouldn't this switch to `true` here …

GretaD: Technically yes, but that will add 1 second more to whole action to be saved. The saving its so fast that you cannot read the sting anyway. I think its unnecessary.

src/components/TagModal.vue
```diff
@@ -155,6 +177,34 @@ export default {
...
+				})
+				this.renameTagLabel = true
+				this.renameTagInput = false
+				this.showSaving = false
+			} catch (error) {
+				showInfo(t('mail', 'An error occurred, unable to rename the tag.'))
+				console.error(error)
+				this.renameTagLabel = false
+				this.renameTagInput = false
+				this.showSaving = true
```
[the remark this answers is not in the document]
GretaD: i set this to true, because when we have an error( that popups on top-right of the screen), the saving message " Saving tag …" will be shown and the user will not be able to click save more than once if they miss the popup error.  So this is true only if we have an error

src/components/TagModal.vue
ChristophWurst: doesn't saving depend on the connection to the server and how fast the server is able to respond?

GretaD: Yes, but showSaving - shows the popup that says "Saving..." not showing this popup saves 1-2 seconds. I dont have a strong opinion on this, i can show it if you think its helpful. We do the same on rename mailbox btw.

GretaD:
> @GretaD anything specific to review design-wise at this point btw? :) cc @nimisha-vijay

No, i applied all what you told me during the call. There are some other things to improve but they'll be done separately. We open tickets for each one of them. This will be merged when Christoph is back, we just want to give it a little bit of time after the release since we changed the namings

ChristophWurst (approved):
Tested, works :+1:

PR #5126, opened 2021-06-07
Revert "Bump vue and vue-template-compiler"

ChristophWurst:
Reverts nextcloud/mail#5114 due to UI regressions. The envelope list doesn't render correctly anymore.


ChristophWurst:
With 2.6.14 it will work. Let's wait for that bump.

ChristophWurst:
> Let's wait for that bump.

https://github.com/nextcloud/mail/pull/5137

PR #5135, opened 2021-06-07
Show tags inside threads

### LINKED ISSUE #5123: Show message tags inside the conversation
### Feature Request

With the upcoming release this app will support (custom) tags. We currently show them in the envelope/thread list, but not in the right column when you open an actual conversation.

#### Summary

Show tags for messages of a thread

GretaD:
tagingthread
fixes #5123

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -51,6 +51,14 @@
 						{{ envelope.subject }}
 					</span>
 				</div>
+				<div v-for="tag in tags"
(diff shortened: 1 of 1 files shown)


GretaD:
yes, there is a problem with vue right now, so the screenshot doesnt show subject, but should be fine after the fixup merge

ChristophWurst (approved):
Code looks good!

ChristophWurst (approved):
Tested and works as expected :sparkles:

PR #5137, opened 2021-06-08
Update Vue and vue-template-compiler to 2.6.14

ChristophWurst:
This unbreaks current master rendering as discussed in the front-end channel yesterday. 2.6.13 had some regressions with slots.


st3iny (approved):
List renders again

PR #5174, opened 2021-06-14
Tag messages in classifier

miaulalala:
Fixes last open item in #25

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -33,6 +33,7 @@ use OCA\Mail\Service\Search\Flag;
 use OCA\Mail\Service\Search\FlagExpression;
 use OCA\Mail\Service\Search\SearchQuery;
 use OCA\Mail\Support\PerformanceLogger;
+use OCA\Mail\Support\PerformanceLoggerTask;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\QBMapper;
 use OCP\AppFramework\Utility\ITimeFactory;
@@ -339,10 +340,11 @@ class MessageMapper extends QBMapper {
 
 	/**
 	 * @param Account $account
+	 * @param bool $permflagsEnabled
 	 * @param Message[] $messages
 	 * @return Message[]
 	 */
-	public function updateBulk(Account $account, Message ...$messages): array {
+	public function updateBulk(Account $account, bool $permflagsEnabled, Message ...$messages): array {
 		$this->db->beginTransaction();
 
 		$perf = $this->performanceLogger->start(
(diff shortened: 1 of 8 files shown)


lib/Listener/NewMessageClassificationListener.php
```diff
@@ -67,11 +74,18 @@ public function handle(Event $event): void {
...
-			return ($message->getFlagImportant() === false);
+		$messages = $event->getMessages();
+
+		// if this is a message that's been flagged / tagged as important before, we don't want to reclassify it again.
+		$doNotReclassify = $this->tagMapper->getTaggedMessageIdsForMessages($event->getMessages(), $event->getAccount()->getUserId(), Tag::LABEL_IMPORTANT);
+		$messages = array_filter($messages, static function ($message) use ($doNotReclassify) {
+			return ($message->getFlagImportant() === false || in_array($message->getMessageId(), $doNotReclassify, true));
 		});
 
+		// we can manually build the tag here, because we do the DB lookup in the tagMessage method
```
ChristophWurst, line 85: related to the comment above I'm not a fan of this. Making assumptions about the implementation over there is prone to bugs in the future as code is changed for other features.

Can we just load `$label1` tag instance here and assign the actual entity instance that has the correct ID?

lib/Listener/NewMessageClassificationListener.php
```diff
@@ -82,6 +96,7 @@ public function handle(Event $event): void {
 			foreach ($event->getMessages() as $message) {
 				if ($predictions[$message->getUid()] ?? false) {
 					$message->setFlagImportant(true);
+					$message->setTags([$important]);
```
ChristophWurst, line 99: :warning:  this will overwrite any other existing tags

add a `addTag` that appends

ChristophWurst (changes requested):

lib/Listener/NewMessageClassificationListener.php
miaulalala: Sure thing, but that would use the `getTagByImapLabel` method again, which is indeed a bottleneck. We do need to check if the tag exists, which means checking every tag on every message against the DB to either insert it or set the right `id` - because if we don't we get an Unique Constraint violation on the `imap_label - user_id` combination IIRC.

I'll open another PR for the bulkUpdate / insert method to speed it up.

lib/Db/TagMapper.php
```diff
@@ -94,10 +94,8 @@ public function getAllTagsForUser(string $userId): array {
 	 * To tag (flag) a message on IMAP, @see \OCA\Mail\Service\MailManager::tagMessage
 	 */
 	public function tagMessage(Tag $tag, string $messageId, string $userId): void {
-		/** @var Tag $exists */
 		try {
-			$exists = $this->getTagByImapLabel($tag->getImapLabel(), $userId);
-			$tag->setId($exists->getId());
+			$tag = $this->getTagByImapLabel($tag->getImapLabel(), $userId);
```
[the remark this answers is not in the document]
miaulalala: We did not?

ChristophWurst (dismissed):
Code looks good, though I couldn't test (obviously :wink:)

lib/Db/TagMapper.php
```diff
@@ -94,10 +94,8 @@ public function getAllTagsForUser(string $userId): array {
 	 * To tag (flag) a message on IMAP, @see \OCA\Mail\Service\MailManager::tagMessage
 	 */
 	public function tagMessage(Tag $tag, string $messageId, string $userId): void {
-		/** @var Tag $exists */
 		try {
-			$exists = $this->getTagByImapLabel($tag->getImapLabel(), $userId);
-			$tag->setId($exists->getId());
```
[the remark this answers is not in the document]
miaulalala: because I'm a potato and sometimes don't think the NC way still.

lib/Db/Message.php
```diff
@@ -217,6 +217,13 @@ public function setTags(array $tags): void {
 		$this->tags = $tags;
 	}
 
+	/**
+	 * @param Tag $tag
+	 */
+	public function addTag(Tag $tag): void {
```
ChristophWurst, line 223: this one can be removed again

lib/Db/MessageMapper.php
```diff
@@ -366,62 +368,37 @@ public function updateBulk(Account $account, Message ...$messages): array {
...
-				return strcmp($a->getImapLabel(), $b->getImapLabel());
-			});
-			foreach ($toAdd as $tag) {
-				// add logging
-				$this->tagMapper->tagMessage($tag, $message->getMessageId(), $account->getUserId());
-			}
-			$perf->step("Tagged messages");
+		// get all tags before the loop and create a mapping [message_id => [tag,...]] but only if permflags are enabled
+		$tags = [];
+		if ($permflagsEnabled === true) {
```
ChristophWurst, line 374:
```suggestion
		if ($permflagsEnabled) {
```

lib/Db/MessageMapper.php
```diff
@@ -366,62 +368,37 @@ public function updateBulk(Account $account, Message ...$messages): array {
...
 			}
 
-			$toRemove = array_udiff($dbTags, $imapTags, function (Tag $a, Tag $b) {
-				return strcmp($a->getImapLabel(), $b->getImapLabel());
-			});
-			foreach ($toRemove as $tag) {
-				//add logging
-				$this->tagMapper->untagMessage($tag, $message->getMessageId());
+			// check permflags and only go through the tagging logic if they're enabled
+			if ($permflagsEnabled === true) {
```
ChristophWurst, line 399:
```suggestion
			if ($permflagsEnabled) {
```

lib/Db/MessageMapper.php
```diff
@@ -430,6 +407,39 @@ public function updateBulk(Account $account, Message ...$messages): array {
 		return $messages;
 	}
 
+	private function updateTags(Account $account, Message $message, array $tags, PerformanceLoggerTask $perf): void {
```
ChristophWurst, line 410: add a phpdoc for `$tags`

lib/Db/MessageMapper.php
```diff
@@ -430,6 +407,39 @@ public function updateBulk(Account $account, Message ...$messages): array {
...
+		$imapTags = $message->getTags();
+		$dbTags = $tags[$message->getMessageId()] ?? [];
+
+		if (empty($imapTags) === true && empty($dbTags) === true) {
```
ChristophWurst, line 414:
```suggestion
		if (empty($imapTags) && empty($dbTags)) {
```

lib/Db/MessageMapper.php
```diff
@@ -430,6 +407,39 @@ public function updateBulk(Account $account, Message ...$messages): array {
...
+			// neither old nor new tags
+			return;
+		}
+
+		$toAdd = array_udiff($imapTags, $dbTags, static function (Tag $a, Tag $b) {
+			return strcmp($a->getImapLabel(), $b->getImapLabel());
+		});
+		foreach ($toAdd as $tag) {
+			// add logging
```
ChristophWurst, line 423: logging?

lib/Db/MessageMapper.php
```diff
@@ -430,6 +407,39 @@ public function updateBulk(Account $account, Message ...$messages): array {
...
+			$this->tagMapper->tagMessage($tag, $message->getMessageId(), $account->getUserId());
+		}
+		$perf->step("Tagged messages");
+
+		if (empty($dbTags) === true) {
```
ChristophWurst, line 428:
```suggestion
		if (empty($dbTags)) {
```

lib/Db/TagMapper.php
```diff
@@ -166,6 +164,40 @@ function (string $key) {
...
+			->from($this->getTableName(), 't')
+			->join('t', 'mail_message_tags', 'mt', $qb->expr()->eq('t.id', 'mt.tag_id', IQueryBuilder::PARAM_INT))
+			->where(
+				$qb->expr()->in('mt.imap_message_id', $qb->createParameter('ids'), IQueryBuilder::PARAM_STR_ARRAY),
+				$qb->expr()->eq('t.user_id', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)),
+				$qb->expr()->eq('t.imap_label', $qb->createNamedParameter($imapLabel, IQueryBuilder::PARAM_STR))
+			);
+
+		$messageIds = [];
+		foreach (array_chunk($ids, 1000) as $chunk) {
```
ChristophWurst, line 189: nit: we could define a constant somewhere for the oracle chunking size to make this less of a magical number. It will help future devs understand the reason for the chunking more easily

lib/Listener/NewMessageClassificationListener.php
```diff
@@ -66,11 +80,24 @@ public function handle(Event $event): void {
 			}
 		}
 
-		// if the message is already flagged as important, we won't classify it again.
-		$messages = array_filter($event->getMessages(), function ($message) {
-			return ($message->getFlagImportant() === false);
+		$messages = $event->getMessages();
+
+		// if this is a message that's been flagged / tagged as important before, we don't want to reclassify it again.
+		$doNotReclassify = $this->tagMapper->getTaggedMessageIdsForMessages($event->getMessages(), $event->getAccount()->getUserId(), Tag::LABEL_IMPORTANT);
```
ChristophWurst, line 86:
```suggestion
		$doNotReclassify = $this->tagMapper->getTaggedMessageIdsForMessages(
				$event->getMessages(),
				$event->getAccount()->getUserId(),
				Tag::LABEL_IMPORTANT
		);
```

makes long lines readable

ChristophWurst (approved):

lib/Db/MessageMapper.php
miaulalala: left over from previous commits, removed it

lib/Db/TagMapper.php
miaulalala: I'll open a new ticket for it

miaulalala: Here: #5214

Issue #3064, opened 2020-05-05
Create event or task from email


ChristophWurst:
### Feature Request

As a user I want to create a calendar event or task from an email like other email solutions offer. The subject will be the event name. Ideally there will be a component from calendar that allows the user to edit props directly instead of having default values sent directly to the user's calendar. As an interim solution we can just create the event and deep link to calendar for edits if the app is available.

#### Summary

Add an action to envelopes/messages that allows the user to create an event or task.

This is not available in all clients, hence not tagging as *feature parity*.

ChristophWurst:
> Ideally there will be a component from calendar that allows the user to edit props directly instead of having default values sent directly to the user's calendar. As an interim solution we can just create the event and deep link to calendar for edits if the app is available.

cc @georgehrke for brainstorming :)

georgehrke:
The interim solution will definitely work.

I like the idea of a calendar-component. We should probably think about a general way for inter-app-components. In the calendar app, there are also cases where we could use a "compose email" component. (https://github.com/nextcloud/calendar/issues/1529)

I'm just worried that those components could get really big. The calendar one will have to bundle both calendar-js and cdav-library.

jakobroehrl:
It would be nice to create a deck card of a mail, too.

agentlibre:
> It would be nice to create a deck card of a mail, too.

And, conversely, to link a deck card to an email. This would be so great!
 
Thanks to all the contributors for the great work, here and elsewhere! 🙏🏻

shtrom:
> > It would be nice to create a deck card of a mail, too.
> And, conversely, to link a deck card to an email. This would be so great!

And add that link back to the email in the description when creating the task/event/deck card.

hilakata:
For clarification purposes: is it now possible to create a task or deck card from email?
Or is this feature still open?

Thanks!

ChristophWurst:
Please use https://help.nextcloud.com/c/apps/mail/35 for support questions or browse the existing feature requests.

Octothorn:
I can see email<->Deck card linking being a handy feature here.  Improving integration and utility in the ecosystem.

> > > It would be nice to create a deck card of a mail, too.
> > > And, conversely, to link a deck card to an email. This would be so great!
> 
> And add that link back to the email in the description when creating the task/event/deck card.

PR #5178, opened 2021-06-14
Create an event from an email

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3064

Todo
- [x] New envelope action
- [x] New message action
- [x] New event form
- [x] Load a user's calendars
- [x] Save the event
  - [x] Requires https://github.com/nextcloud/calendar-js/issues/273
- [x] Error handling
- [x] Fix all day error
- [x] Requires `@nextcloud/vue` https://github.com/nextcloud/mail/pull/5477

Bildschirmfoto von 2021-06-14 15-40-47


ChristophWurst:
Unexpected road block: I'll have to handle fun stuff like time zones here. And then I'm more or less duplicating the simplified editor of Calendar. *Maybe* it's worth making the component reusable, export it in some way and make use if it here as well.

jakobroehrl:
@ChristophWurst 
Would love to use it here https://github.com/nextcloud/calendar/pull/3260 und maybe in the dashboard widget to directly create an event

nimisha-vijay:
image
- ideally ```from``` datetime can be gotten from the contents of the email, but if that's not possible it can be rounded to the nearest hour from current time
- ```to``` datetime default can be ```from``` datetime + 1 hour

cc @GretaD for the feedback

jancborchardt:
Great mockup @nimisha-vijay! :)

Detail for the event title I only just noticed:
- We could cut any "Re: ", "Fwd: " from the title, there is already a list of these prefixes at https://github.com/nextcloud/mail/blob/5a2c76f9b2f62e7d3ebab1ccfd84dcc3394a6854/src/ReplyBuilder.js#L131-L189
- Maybe additionally we can cut content inbetween [Brackets] as that is usually mailing list titles or smth which is less relevant for a short and succinct event title?

GretaD:
Thank you Nimisha, I really like the mockup, but its different from what we have now on calendar. @jancborchardt shouldnt we have it the same as everywhere? What is the benefit of moving the "selected calendar" in the end of the section, apart from aesthetically?

nimisha-vijay:
That's right actually, but maybe we can try to move away from the old calendar dropdown in general? cc @jancborchardt

GretaD:
> That's right actually, but maybe we can try to move away from the old calendar dropdown in general? cc @jancborchardt

I am fine with that, but in general i am a fan of consistency and if we change the design, it should be taken in consideration how easy/hard is to change the old design everywhere. We have "trained" people to choose the calendar as a second step, and now its the last step. I am gonna start working on this mockup, but just wanted to point that out.

ChristophWurst:
Bildschirmfoto von 2021-07-28 09-41-59

timezones are now handled properly

ChristophWurst:
> image

Why do we move off standard with the calendar picker? Would it not look more consistent with the controls above if it also has its default border?

src/components/EventModal.vue
```diff
@@ -6,16 +6,24 @@
 				<input v-model="eventTitle" type="text">
 			</div>
 			<div class="dateTimePicker">
-				<DatetimePicker type="datetime" :show-timezone-select="true" :timezone-id="startTimezoneId" />
-				<DatetimePicker type="datetime" :show-timezone-select="true" :timezone-id="endTimezoneId" />
+				<DatetimePicker :format="'YYYY-MM-DD HH:mm'"
```
skjnldsv, line 9: The `allDay` is working. Your issue is about the format. It will keep remembering the hour. But if you change the date after clicking `allDay`, then it will be reset to midnight.

skjnldsv: But You're missing two v-mode or `value` bindings, so you cannot update the picker and therefore you cannot see the change.

skjnldsv: Peek 11-08-2021 08-46

skjnldsv:
```diff
diff --git a/src/components/EventModal.vue b/src/components/EventModal.vue
index 734733e15..96357284a 100644
--- a/src/components/EventModal.vue
+++ b/src/components/EventModal.vue
@@ -6,14 +6,18 @@
                                <input v-model="eventTitle" type="text">
                        </div>
                        <div class="dateTimePicker">
-                               <DatetimePicker :format="'YYYY-MM-DD HH:mm'"
+                               <DatetimePicker
+                                       v-model="startDate"
                                        :clearable="false"
+                                       :format="dateFormat"
                                        :minute-step="5"
                                        :show-second="false"
                                        :type="datePickerType"
                                        :show-timezone-select="true"
                                        :timezone-id="startTimezoneId" />
-                               <DatetimePicker :format="'YYYY-MM-DD HH:mm'"
+                               <DatetimePicker
+                                       v-model="endDate"
+                                       :format="dateFormat"
                                        :type="datePickerType"
                                        :show-timezone-select="true"
                                        :timezone-id="endTimezoneId" />
@@ -93,6 +97,8 @@ export default {
                        endTimezoneId: defaultTimezoneId,
                        saving: false,
                        selectedCalendar: undefined,
+                       startDate: new Date(),
+                       endDate: new Date(),
                }
        },
        computed: {
@@ -100,6 +106,9 @@ export default {
                        console.info('DATE PICKER', this.isAllDay ? 'date' : 'datetime')
                        return this.isAllDay ? 'date' : 'datetime'
                },
+               dateFormat() {
+                       return this.isAllDay ? 'YYYY-MM-DD' : 'YYYY-MM-DD HH:mm'
+               },

```

GretaD:
All looks good, apart from the event is saved as "Untitled event" on the calendar.

ChristophWurst:
> the event is saved as "Untitled event" on the calendar

I didn't have that. Could you check the browser logs? Before the calendar is written the data is printed.

GretaD:
> > the event is saved as "Untitled event" on the calendar
> 
> I didn't have that. Could you check the browser logs? Before the calendar is written the data is printed.

only this info:
```
[INFO] mail: importing event component 
{app: "mail", uid: "greta", component: "BEGIN:VCALENDAR\r\nPRODID:-//Nextcloud Mail\r\nBEGIN:V…Berlin:20210817T154204\r\nEND:VEVENT\r\nEND:VCALENDAR"}
app: "mail"
component: "BEGIN:VCALENDAR\r\nPRODID:-//Nextcloud Mail\r\nBEGIN:VTIMEZONE\r\nTZID:Europe/Berlin\r\nBEGIN:DAYLIGHT\r\nTZOFFSETFROM:+0100\r\nTZOFFSETTO:+0200\r\nTZNAME:CEST\r\nDTSTART:19700329T020000\r\nRRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU\r\nEND:DAYLIGHT\r\nBEGIN:STANDARD\r\nTZOFFSETFROM:+0200\r\nTZOFFSETTO:+0100\r\nTZNAME:CET\r\nDTSTART:19701025T030000\r\nRRULE:FREQ=YEARLY;BYMONTH=10;BYDAY=-1SU\r\nEND:STANDARD\r\nEND:VTIMEZONE\r\nBEGIN:VEVENT\r\nCREATED:20210817T134209Z\r\nDTSTAMP:20210817T134209Z\r\nLAST-MODIFIED:20210817T134209Z\r\nSEQUENCE:1\r\nUID:fcb07c0c-77c8-4a43-92ae-53e6bf0cf3fa\r\nDTSTART;TZID=Europe/Berlin:20210817T154204\r\nDTEND;TZID=Europe/Berlin:20210817T154204\r\nEND:VEVENT\r\nEND:VCALENDAR"
uid: "greta"
```

PR #5183, opened 2021-06-15
Document MailboxSync::persist arguments

ChristophWurst:
Makes IDEs and static analysis happier.

### DIFF
--- a/lib/IMAP/MailboxSync.php
+++ b/lib/IMAP/MailboxSync.php
@@ -153,6 +153,11 @@ class MailboxSync {
 		$this->mailboxMapper->update($mailbox);
 	}
 
+	/**
+	 * @param Account $account
+	 * @param Folder[] $folders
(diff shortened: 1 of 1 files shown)


kesselb (approved):

lib/IMAP/MailboxSync.php
```diff
@@ -31,6 +31,7 @@
 use Horde_Imap_Client_Namespace_List;
 use OCA\Mail\Events\MailboxesSynchronizedEvent;
 use OCA\Mail\Exception\ServiceException;
+use OCP\DB\Exception;
```
kesselb, line 34: Psalm: Does not exist for Nextcloud 20 :thinking:

ChristophWurst: Then let's do without the exception. That wasn't my focus anyway :)

PR #5189, opened 2021-06-16
Sanitize urls in css style sheets

st3iny:
Currently, CSS style sheets are not sanitized and allow tracker scripts to bypass our trusted senders feature. Furthermore, our style attribute sanitizer was not strict enough so I improved it to replace all `url(...)` values (not just those of `background-*` rules).

I added a new PHP dependency for parsing CSS style sheets. Using regex is not sufficient to parse style sheets and could be exploited by a future attacker.

## Example
```css
ul { list-style-image: url(https://my.tracker.com/some-image.png); }
div { background-image: url(https://my.tracker.com/some-image.png); }
```

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -92,6 +92,7 @@ export default {
 			this.hasBlockedContent
 				= iframeDoc.querySelectorAll('[data-original-src]').length > 0
 				|| iframeDoc.querySelectorAll('[data-original-style]').length > 0
+				|| iframeDoc.querySelectorAll('style[data-original-content]').length > 0
 
 			this.loading = false
 			if (this.isSenderTrusted) {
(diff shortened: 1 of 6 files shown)


src/components/MessageHTMLBody.vue
```diff
@@ -114,6 +115,11 @@ export default {
 			iframeDoc
 				.querySelectorAll('[data-original-style]')
 				.forEach((node) => node.setAttribute('style', node.getAttribute('data-original-style')))
+			iframeDoc
+				.querySelectorAll('style[data-original-content]')
+				.forEach((node) => {
+					node.innerHTML = node.getAttribute('data-original-content')
```
LukasReschke, line 121: I haven't tested this, but could I use this to bypass any other kind of sanitization we may be running on the HTML?

As in: what if someone sends an email that has something like:

```
<span data-original-content="<img src=&quot;doesnotexist&quot; onerror=&quot;alert(1)&quot;">
```

Or don't we have any kind of sanitization anymore regardless as we sandbox the iframe?

st3iny: The selector only selects `style` elements with the attribute `data-original-content` so your example won't have any effect. 

Nevertheless, an attacker could send a forged style tag containing some malicious content but our sanitation extracts, merges and then sanitizes all style sheets at once so theoretically this shouldn't be exploitable. The rendered html email will only have one `style` element right at the beginning.

kesselb (approved):

PR #5201, opened 2021-06-17
Implement resizable columns

ChristophWurst:
My version of https://github.com/nextcloud/mail/pull/4123. Tested and works.

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -1,6 +1,6 @@
 <template>
 	<AppContent pane-config-key="mail">
-		<div id="app-content-wrapper">
+		<template slot="list">
 			<AppContentList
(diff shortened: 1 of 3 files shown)


miaulalala:
Just testing - if I pull the column as far to the right as possible, I seem to have to click pretty close to the edge of the column to let go of the resize state. If I click too far right, it still will resize the column if I move the mouse to the left again.

ChristophWurst:
I'd say if there are any quirks with the resize handling that they are most likely from the nc/vue lib. Let's report them there. You can also try https://github.com/nextcloud/contacts/pull/2246 for reference. If my assumption is right then you should experience the same there.

miaulalala (approved):

szaimen:
🎉

PR #5216, opened 2021-06-22
Fix reply icon on answered messages

GretaD:
For Christoph's eyes only

replydark
replylight

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -49,6 +49,7 @@
 			</div>
 		</template>
 		<template #subtitle>
+			<span v-if="data.flags.answered" class="icon-reply" />
 			<span v-if="data.flags.hasAttachments === true" class="icon-public icon-attachment" />
 			<span v-if="draft" class="draft">
 				<em>{{ t('mail', 'Draft: ') }}</em>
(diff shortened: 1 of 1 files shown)


src/components/Envelope.vue
```diff
@@ -455,7 +456,6 @@ list-item-style.draft .app-content-list-item-line-two {
 }
 
 .icon-reply {
-	background-image: url('../../img/reply.svg');
 	-ms-filter: 'progid:DXImageTransform.Microsoft.Alpha(Opacity=50)';
 	opacity: 0.5;
```
GretaD, line 460: should i change the opacity .25 same as attachment?

ChristophWurst: Whatever makes the look the same :wink:

src/components/Envelope.vue
```diff
@@ -455,9 +456,8 @@ list-item-style.draft .app-content-list-item-line-two {
 }
 
 .icon-reply {
-	background-image: url('../../img/reply.svg');
```
ChristophWurst, line 458: where's the icon loaded from if this is dropped? :thinking:

GretaD: web console says its from mail.css

ChristophWurst (approved):

PR #5222, opened 2021-06-23
Anti spam report feature

### LINKED ISSUE #4416: Spam reporting
### Feature Request

As an admin I want to allow my users to report spam emails easily, e.g. to adjust filter rules on the company spam filter. There should be a system-wide reporting email configurable. When the user reports a message, an email (with the spam message attached) will be sent there. The user doesn't need to know who a message is sent to. That message should also not be stored on IMAP.

If no such reporting address is configured, the menu option shall be hidden.

#### Summary

Allow users to report spam email to their admin.

miaulalala:
Fixes #4416

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -28,6 +28,7 @@ namespace OCA\Mail\Controller;
 use OCA\Mail\AppInfo\Application;
 use OCA\Mail\Exception\ValidationException;
 use OCA\Mail\Http\JsonResponse as HttpJsonResponse;
+use OCA\Mail\Service\AntiSpamService;
 use OCA\Mail\Service\Provisioning\Manager as ProvisioningManager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\JSONResponse;
@@ -39,10 +40,15 @@ class SettingsController extends Controller {
 	/** @var ProvisioningManager */
 	private $provisioningManager;
 
+	/** @var AntiSpamService */
+	private $antiSpamService;
+
 	public function __construct(IRequest $request,
-								ProvisioningManager $provisioningManager) {
+								ProvisioningManager $provisioningManager,
+								AntiSpamService $antiSpamService) {
 		parent::__construct(Application::APP_ID, $request);
 		$this->provisioningManager = $provisioningManager;
+		$this->antiSpamService = $antiSpamService;
 	}
 
 	public function index(): JSONResponse {
@@ -86,6 +92,26 @@ class SettingsController extends Controller {
 			$this->provisioningManager->deprovision($provisioning);
 		}
 
+		return new JSONResponse([]);
+	}
+
+	/**
+	 *
+	 * @return JSONResponse
+	 */
+	public function setAntiSpamEmail(string $spam, string $ham): JSONResponse {
+		$this->antiSpamService->setSpamEmail($spam);
+		$this->antiSpamService->setHamEmail($ham);
+		return new JSONResponse([]);
+	}
+
+	/**
+	 * Store the credentials used for SMTP in the config
+	 *
+	 * @return JSONResponse
+	 */
+	public function deleteAntiSpamEmail(): JSONResponse {
+		$this->antiSpamService->deleteConfig();
 		return new JSONResponse([]);
 	}
 }
--- a/lib/Listener/HamReportListener.php
+++ b/lib/Listener/HamReportListener.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2021 Anna Larch <anna@nextcloud.com>
+ *
+ * @author Anna Larch <anna@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Mail\Listener;
+
+use OCA\Mail\Events\MessageFlaggedEvent;
+use OCA\Mail\Service\AntiSpamService;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use Psr\Log\LoggerInterface;
+
+class HamReportListener implements IEventListener {
+
+	/** @var LoggerInterface */
+	private $logger;
+
+	/** @var AntiSpamService */
+	private $antiSpamService;
+
+	public function __construct(LoggerInterface $logger,
+								AntiSpamService $antiSpamService) {
+		$this->logger = $logger;
+		$this->antiSpamService = $antiSpamService;
+	}
+
+	public function handle(Event $event): void {
+		if (!$event instanceof MessageFlaggedEvent || $event->getFlag() !== '$notjunk') {
+			return;
+		}
+
+		if (!$event->isSet()) {
+			return;
+		}
+
+		// Send message to reporting service
+		try {
+			$this->antiSpamService->sendReportEmail($event->getAccount(), $event->getMailbox(), $event->getUid(), $event->getFlag());
+		} catch (\Throwable $e) {
+			$this->logger->error("Could not send spam report: " . $e->getMessage(), ['exception' => $e]);
+		}
+	}
+}
--- a/lib/Listener/SpamReportListener.php
+++ b/lib/Listener/SpamReportListener.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2021 Anna Larch <anna@nextcloud.com>
+ *
+ * @author Anna Larch <anna@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Mail\Listener;
+
+use OCA\Mail\Events\MessageFlaggedEvent;
+use OCA\Mail\Service\AntiSpamService;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use Psr\Log\LoggerInterface;
+
+class SpamReportListener implements IEventListener {
+
+	/** @var LoggerInterface */
+	private $logger;
+
+	/** @var AntiSpamService */
+	private $antiSpamService;
+
+	public function __construct(LoggerInterface $logger,
+								AntiSpamService $antiSpamService) {
+		$this->logger = $logger;
+		$this->antiSpamService = $antiSpamService;
+	}
+
+	public function handle(Event $event): void {
+		if (!$event instanceof MessageFlaggedEvent || $event->getFlag() !== '$junk') {
+			return;
+		}
+
+		if (!$event->isSet()) {
+			return;
+		}
+
+		// Send message to reporting service
+		try {
+			$this->antiSpamService->sendReportEmail($event->getAccount(), $event->getMailbox(), $event->getUid(), $event->getFlag());
+		} catch (\Throwable $e) {
+			$this->logger->error("Could not send spam report: " . $e->getMessage(), ['exception' => $e]);
+		}
+	}
+}
--- a/lib/Model/IMessage.php
+++ b/lib/Model/IMessage.php
@@ -127,9 +127,17 @@ interface IMessage {
 
 	/**
 	 * @param string $name
-	 * @param string $content
+	 * @param string $content attached with mime type 'application/octet-stream'
 	 */
 	public function addRawAttachment(string $name, string $content): void;
+
+	/**
+	 * @param string $name
+	 * @param string $content attached with mime type 'message/rfc822'
+	 *
+	 * @return void
+	 */
+	public function addEmbeddedMessageAttachment(string $name, string $content): void;
 
 	/**
 	 * @param File $file
--- a/lib/Service/AntiSpamService.php
+++ b/lib/Service/AntiSpamService.php
@@ -1,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2021 Anna Larch <anna@nextcloud.com>
+ *
+ * @author Anna Larch <anna@nextcloud.com>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Mail\Service;
+
+use OCA\Mail\Account;
+use OCA\Mail\Contracts\IMailTransmission;
+use OCA\Mail\Db\Mailbox;
+use OCA\Mail\Db\MessageMapper;
+use OCA\Mail\Exception\SentMailboxNotSetException;
+use OCA\Mail\Exception\ServiceException;
+use OCA\Mail\Model\NewMessageData;
+use OCP\IConfig;
+
+class AntiSpamService {
+	private const NAME = 'antispam_reporting';
+	private const MESSAGE_TYPE = 'message/rfc822';
+
+	/** @var IConfig */
+	private $config;
+
+	/** @var MessageMapper */
+	private $messageMapper;
+
+	/** @var IMailTransmission */
+	private $transmission;
+
+	public function __construct(IConfig $config,
+								MessageMapper $messageMapper,
+								IMailTransmission $transmission) {
+		$this->config = $config;
+		$this->messageMapper = $messageMapper;
+		$this->transmission = $transmission;
+	}
+
+	public function getSpamEmail(): string {
+		return $this->config->getAppValue('mail', self::NAME . '_spam');
+	}
+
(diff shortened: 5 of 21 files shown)


lib/Service/AntiSpamService.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @author Tahaa Karim <tahaalibra@gmail.com>
```
ChristophWurst, line 6: :eye:

miaulalala: copy paste strikes again...

lib/Listener/AntiSpamReportListener.php
```diff
@@ -63,17 +52,18 @@ public function handle(Event $event): void {
...
 
 			//Send message to reporting service
-			$this->service->sendSpamReport(
-				$event
-			);
+			try {
+				$message = $this->service->createSpamReportMessageData($event->getAccount(), $event->getMailbox(), $event->getUid());
+				$this->service->sendSpamReport($message);
+			} catch (ServiceException $e) {
+				$this->logger->error($e->getMessage(), [$e]);
```
ChristophWurst, line 65: needs the `exception` as key

miaulalala: Like so? `[ServiceException::class => $e]`

ChristophWurst: > If an Exception object is passed in the context data, it MUST be in the 'exception' key. Logging exceptions is a common pattern and this allows implementors to extract a stack trace from the exception when the log backend supports it. Implementors MUST still verify that the 'exception' key is actually an Exception before using it as such, as it MAY contain anything.

https://www.php-fig.org/psr/psr-3/

``['exception' => $e]``

miaulalala: ah! thanks :)

tests/Integration/Service/AntiSpamServiceIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+/**
+ * @copyright 2017 Christoph Wurst <christoph@winzerhof-wurst.at>
```
ChristophWurst, line 4: it wasn't me :notes:

tests/Integration/Service/AntiSpamServiceIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+class AntiSpamServiceIntegrationTest extends TestCase {
+	use ImapTest,
+		ImapTestAccount;
+
+	/** @var AntiSpamService */
+	private $service;
+
+	public function setUp():void {
+		parent::setUp();
+		$this->service = new AntiSpamService(
```
ChristophWurst, line 50: you could also fetch the whole `AntiSpamService` from the container and thereby indirectly test if the container can build the service. This has the nice side effect that new arguments to the constructor don't have to be added here.

tests/Unit/Service/AntiSpamServiceNotActiveTest.php
```diff
@@ -0,0 +1,74 @@
...
+		$this->transmission = $this->createMock(IMailTransmission::class);
+
+		$this->service = new AntiSpamService(
+			$this->config,
+			$this->messageMapper,
+			$this->transmission
+		);
+	}
+
+	public function testSendSpamReport() {
```
ChristophWurst, line 65:
```suggestion
	public function testSendSpamReport(): void {
```

tests/Unit/Service/AntiSpamServiceNotActiveTest.php
```diff
@@ -0,0 +1,74 @@
...
+		$account = $this->createMock(Account::class);
+		$mailbox = $this->createMock(Mailbox::class);
+		$this->messageMapper->expects($this->never())
+			->method('getIdForUid');
+		$this->transmission->expects($this->never())
+			->method('sendMessage');
+		$this->service->sendSpamReport($account, $mailbox, 123);
```
ChristophWurst, line 72:
```suggestion

		$this->service->sendSpamReport($account, $mailbox, 123);
```

this remark was brought to you by the [*Arrange-Act-Assert gang*](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/)

tests/Unit/Service/AntiSpamServiceTest.php
```diff
@@ -0,0 +1,131 @@
...
+		$this->transmission = $this->createMock(IMailTransmission::class);
+
+		$this->service = new AntiSpamService(
+			$this->config,
+			$this->messageMapper,
+			$this->transmission
+		);
+	}
+
+	public function testSendSpamReportNoMessageFound() {
```
ChristophWurst, line 67:
```suggestion
	public function testSendSpamReportNoMessageFound(): void {
```

also for the other test methods :)

lib/Model/IMessage.php
```diff
@@ -131,6 +131,14 @@ public function getAttachments(): array;
 	 */
 	public function addRawAttachment(string $name, string $content): void;
 
+	/**
+	 * @param string $name
+	 * @param string $content
```
ChristophWurst, line 136: might be worth noting the expected rfc822 format for the string

lib/Listener/AntiSpamReportListener.php
```diff
@@ -0,0 +1,68 @@
...
+	private $service;
+
+	public function __construct(LoggerInterface $logger,
+								AntiSpamService $service) {
+		$this->logger = $logger;
+		$this->service = $service;
+	}
+
+	public function handle(Event $event): void {
+		if ($event instanceof MessageFlaggedEvent && $event->getFlag() === 'junk') {
```
ChristophWurst, line 50: flip expression and make this an early return

lib/Listener/AntiSpamReportListener.php
```diff
@@ -0,0 +1,68 @@
...
+			// No anti spam config found
+			if (empty($this->service->getReportEmail())) {
+				return;
+			}
+
+			//Send message to reporting service
+			try {
+				$this->service->sendSpamReport($event->getAccount(), $event->getMailbox(), $event->getUid());
+			} catch (ServiceException $e) {
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
```
ChristophWurst, line 64:
```suggestion
				$this->logger->error("Could not send spam report: " . $e->getMessage(), ['exception' => $e]);
```

then admins encountering those entries will more likely know what the error is about

ChristophWurst (changes requested):
Some more nitpicks

lib/Model/IMessage.php
miaulalala: Good idea

ChristophWurst (approved):
Code looks good!

miaulalala:
@ChristophWurst I had a discussion with @kesselb 

- We're not moving messages to the spam mailbox when marking it spam (but AFAIK neither does Thunderbird unless you set it in options) PR here: #5326
- AFAIK we're not flagging a message as spam when it is moved to the spam folder but the question is if this is intentional? Rainloop et al aren't flagging them.
- We're missing a way to report a message as "not junk"/ ham  / false positive
- the feature should be better documented

lib/Service/AntiSpamService.php
```diff
@@ -0,0 +1,112 @@
...
+	public function __construct(IConfig $config,
+								MessageMapper $messageMapper,
+								IMailTransmission $transmission) {
+		$this->config = $config;
+		$this->messageMapper = $messageMapper;
+		$this->transmission = $transmission;
+	}
+
+	/**
+	 * @codeCoverageIgnore
```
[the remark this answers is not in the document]
miaulalala: https://phpunit.readthedocs.io/en/9.5/annotations.html

kesselb:
> the feature should be better documented

This sounds so harsh ;) I had in mind to add some more explanation to the admin ui and explain what exactly the feature does. As each company / mail provider may setup this process a bit different. 

Example: If a spam / ham email address is configured a message to the given address is sent with the email as rfc xyz attachment when the user click mark as spam / mark not as spam.

lib/Listener/AntiSpamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+		if (!$event instanceof MessageFlaggedEvent || $event->getFlag() !== 'junk') {
+			return;
+		}
+
+		if (!$event->isSet()) {
+			return;
+		}
+
+		// No anti spam config found
+		if (empty($this->service->getReportEmail())) {
```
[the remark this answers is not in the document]
miaulalala: I removed it there actually.

lib/Listener/HamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+		if (!$event->isSet()) {
+			return;
+		}
+
+		// No ham config found
+		if (empty($this->antiSpamService->getHamEmail())) {
+			return;
+		}
+
+		//Send message to reporting service
```
ChristophWurst, line 63:
```suggestion
		// Send message to reporting service
```

lib/Listener/HamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+		try {
+			$this->antiSpamService->sendReportEmail($event->getAccount(), $event->getMailbox(), $event->getUid(), $this->antiSpamService->getHamEmail(), $this->antiSpamService->getHamSubject());
```
ChristophWurst, line 65: I find it a bit strange that the email and subject for this method call are acquired from the same service. Could you omit this here and just call the getters internally?

lib/Service/AntiSpamService.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public function getSpamSubject(): string {
+		return 'Learn as Junk';
+	}
+
+	/**
+	 * @codeCoverageIgnore
```
ChristophWurst, line 80: y tho

tests/Unit/Listener/HamReportListenerTest.php
```diff
@@ -0,0 +1,198 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
```
ChristophWurst, line 6: it wasn't me

lib/Listener/HamReportListener.php
miaulalala: sure thing. I don't even remember why I did it this way.

lib/Service/AntiSpamService.php
miaulalala: So codecov won't be giving me minus points?

lib/Listener/HamReportListener.php
miaulalala: Ah, now I remember. But it's ok. I will change it.

lib/Service/AntiSpamService.php
ChristophWurst: Either we except the lower rating or we write a test that covers this line :)

tests/Unit/Listener/HamReportListenerTest.php
ChristophWurst: :thinking:

src/components/settings/AdminSettings.vue
```diff
@@ -111,13 +111,35 @@
...
+						t(
+							'mail',
+							'You can set up an anti spam service email address here.'
+						)
+					}}
+					<br>
+					{{
+						t(
+							'mail',
+							'Any email that is marked as junk will not only be flagged as such but additionally sent to the anti spam service.'
```
kesselb, line 128:
```suggestion
							'Any email that is marked as spam will not only be flagged as such but additionally sent to the anti spam service.'
```

lib/Listener/SpamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+	private $antiSpamService;
+
+	public function __construct(LoggerInterface $logger,
+								AntiSpamService $antiSpamService) {
+		$this->logger = $logger;
+		$this->antiSpamService = $antiSpamService;
+	}
+
+	public function handle(Event $event): void {
+		if (!$event instanceof MessageFlaggedEvent || $event->getFlag() !== \Horde_Imap_Client::FLAG_JUNK) {
```
kesselb, line 50: image

image

Hmm. Is this broken on my end? Or the check might be wrong as `$junk` is imap but `junk` our internal value (passed from the frontend to the backend and forwarded to the event).

lib/Service/AntiSpamService.php
```diff
@@ -0,0 +1,116 @@
...
+		$this->config->deleteAppValue('mail', self::NAME . '_spam');
+		$this->config->deleteAppValue('mail', self::NAME . '_ham');
+	}
+
+	/**
+	 * @param MessageFlaggedEvent $event
+	 * @param int $uid
+	 * @throws ServiceException
+	 */
+	public function sendReportEmail(MessageFlaggedEvent $event, int $uid): void {
```
kesselb, line 91: I find it a bit weird to pass the event and uid but the uid is already part of the event.

lib/Listener/HamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+	private $antiSpamService;
+
+	public function __construct(LoggerInterface $logger,
+								AntiSpamService $antiSpamService) {
+		$this->logger = $logger;
+		$this->antiSpamService = $antiSpamService;
+	}
+
+	public function handle(Event $event): void {
+		if (!$event instanceof MessageFlaggedEvent || $event->getFlag() !== \Horde_Imap_Client::FLAG_NOTJUNK) {
```
kesselb, line 50: image

lib/Service/AntiSpamService.php
ChristophWurst: guess we shouldn't actually pass the event. that isn't a concern of the service, only the listener

lib/Listener/HamReportListener.php
miaulalala: Damn it, I wanted to use a CONST instead of writing it out. Now i made a mess.

ChristophWurst (changes requested):
Doesn't work due to inconsistencies with flags. `$junk` and `$notjunk` are the ones per RFC, right?

lib/Controller/SettingsController.php
```diff
@@ -88,4 +94,24 @@ public function deprovision(int $id): JSONResponse {
 
 		return new JSONResponse([]);
 	}
+
+	/**
+	 *
+	 * @return JSONResponse
+	 */
+	public function setAntiSpamEmail(array $data): JSONResponse {
```
ChristophWurst, line 102:
```suggestion
	public function setAntiSpamEmail(string $spam, string $ham): JSONResponse {
```

you should be able to spread to type-hinted arguments

lib/Listener/HamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+		if (!$event instanceof MessageFlaggedEvent || $event->getFlag() !== '$notjunk') {
+			return;
+		}
+
+		if (!$event->isSet()) {
+			return;
+		}
+
+		// No ham config found
+		if (empty($this->antiSpamService->getHamEmail())) {
```
ChristophWurst, line 59: I would still argue that it's better to not have this check here but let `sendReportEmail` become a NOP if no email is set for the report type. Otherwise one could theoretically use `sendReportEmail` without this check and the code will do :boom:

lib/Listener/HamReportListener.php
```diff
@@ -0,0 +1,70 @@
...
+			return;
+		}
+
+		// Send message to reporting service
+		try {
+			$this->antiSpamService->sendReportEmail($event->getAccount(), $event->getMailbox(), $event->getUid(), $event->getFlag());
+		} catch (ServiceException $e) {
```
ChristophWurst, line 66: I would widen this to \Throwable because anything that goes wrong here is not essential and therefore shouldn't kill the flagging request

tests/Unit/Listener/HamReportListenerTest.php
```diff
@@ -0,0 +1,192 @@
...
+	private $listener;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->serviceMock = $this->createServiceMock(HamReportListener::class);
+		$this->listener = $this->serviceMock->getService();
+	}
+
+	public function testHandleUnrelated() {
```
ChristophWurst, line 54:
```suggestion
	public function testHandleUnrelated(): void {
```

same for all the other new test methods

tests/Unit/Listener/HamReportListenerTest.php
```diff
@@ -0,0 +1,192 @@
...
+		$this->serviceMock->getParameter('antiSpamService')
+			->expects($this->once())
+			->method('getHamEmail')
+			->willReturn('SpammitySpam@WonderfulSpam.egg');
+		$this->serviceMock->getParameter('antiSpamService')
+			->expects($this->once())
+			->method('sendReportEmail')
+			->with($account, $mailbox, 123, '$notjunk');
+		$this->serviceMock->getParameter('logger')
+			->expects($this->never())
```
ChristophWurst, line 187:
```suggestion
			->expects(self::never())
```

tests/Unit/Listener/HamReportListenerTest.php
```diff
@@ -0,0 +1,192 @@
...
+		$event = new MessageFlaggedEvent(
+			$account,
+			$mailbox,
+			123,
+			'$notjunk',
+			true
+		);
+
+		$this->serviceMock->getParameter('antiSpamService')
+			->expects($this->once())
```
ChristophWurst, line 179:
```suggestion
			->expects(self::once())
```

lib/Service/AntiSpamService.php
```diff
@@ -0,0 +1,120 @@
...
+use OCA\Mail\Contracts\IMailTransmission;
+use OCA\Mail\Db\Mailbox;
+use OCA\Mail\Db\MessageMapper;
+use OCA\Mail\Exception\SentMailboxNotSetException;
+use OCA\Mail\Exception\ServiceException;
+use OCA\Mail\Model\NewMessageData;
+use OCP\IConfig;
+
+class AntiSpamService {
+	public const NAME = 'antispam_reporting';
```
ChristophWurst, line 38: IMO should be private

lib/Service/AntiSpamService.php
```diff
@@ -0,0 +1,120 @@
...
+	public const MESSAGE_TYPE = 'message/rfc822';
```
ChristophWurst, line 39: same, only used internally

tests/Unit/Service/AntiSpamServiceTest.php
```diff
@@ -0,0 +1,185 @@
...
+	public function testSendReportEmailNoMessageFound(): void {
+		$event = $this->createConfiguredMock(MessageFlaggedEvent::class, [
+			'getAccount' => $this->createMock(Account::class),
+			'getMailbox' => $this->createMock(Mailbox::class),
+			'getFlag' => 'junk'
+		]);
+
+		$this->config->expects($this->once())
+			->method('getAppValue')
+			->with('mail', AntiSpamService::NAME . '_spam')
```
ChristophWurst, line 75:
```suggestion
			->with('mail', 'antispam_reporting_spam')
```

duplication is fine in testing code

ChristophWurst (dismissed):
Works as expected. Some code remarks :)

lib/Settings/AdminSettings.php
```diff
@@ -53,6 +59,16 @@ public function getForm() {
 			$this->provisioningManager->getConfigs()
 		);
 
+		$this->initialStateService->provideInitialState(
+			Application::APP_ID,
+			'antispam_setting',
+			[
+				$this->antiSpamService->getSpamEmail(),
```
ChristophWurst, line 66:
```suggestion
				'spam' => $this->antiSpamService->getSpamEmail(),
```

lib/Settings/AdminSettings.php
```diff
@@ -53,6 +59,16 @@ public function getForm() {
...
+				$this->antiSpamService->getHamEmail(),
```
ChristophWurst, line 67:
```suggestion
				'ham' => $this->antiSpamService->getHamEmail(),
```

ChristophWurst (changes requested):
Otherwise the email addresses are blank after the next page load

ChristophWurst (approved):
:shipit:

PR #5229, opened 2021-06-24
Test against nc 23

GretaD:
thats all i need to change, right?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         php-versions: [7.3, 7.4]
-        nextcloud-versions: ['stable20', 'stable21', 'master']
+        nextcloud-versions: ['stable20', 'stable21', 'stable22', 'master']
         include:
             - php: 8.0
               nextcloud-version: master


.github/workflows/static-analysis.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       runs-on: ubuntu-latest
       strategy:
           matrix:
-              ocp-version: [ 'dev-master', 'v20.0.4' ]
+              ocp-version: [ 'v22.0.0', 'v20.0.4' ]
```
ChristophWurst, line 15: let's leave dev-master and *add* v22.0.0-RC1

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         php-versions: [7.3, 7.4]
-        nextcloud-versions: ['stable20', 'stable21', 'master']
+        nextcloud-versions: ['stable20', 'stable21', 'stable22']
```
ChristophWurst, line 16: we should test both stable22 and master

.github/workflows/test.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       strategy:
           matrix:
               php-versions: [7.4, 8.0]
-              nextcloud-versions: ['master']
+              nextcloud-versions: ['stable22']
```
ChristophWurst, line 62: this can stay on master

ChristophWurst:
info.xml doesn't allow 23 yet

GretaD:
> info.xml doesn't allow 23 yet

because i need to change this: https://github.com/nextcloud/mail/blob/master/appinfo/info.xml#L37
to 23, right?

ChristophWurst:
> because i need to change this: https://github.com/nextcloud/mail/blob/master/appinfo/info.xml#L37
> to 23, right?

yes. That will bite us as soon as they've updated https://github.com/nextcloud/server/blob/4e9727037220a8acddf7a3478a047c0bc6fc4bae/version.php#L33 to say 23

GretaD:
i see Joas pushed #5232  and #5230

ChristophWurst:
> i see Joas pushed #5232 and #5230

Rebase to see what is left. I think your changes on the CI matrix are still required :)

ChristophWurst (approved):

PR #5251, opened 2021-06-28
Change workflows to exclude NC23

miaulalala:
Since we don't support NC23 for now, exclude it from tests and StatAna


ChristophWurst:
> miaulalala wants to merge 1 commit into master

:warning: you have to branch off `stable1.10` and do your PR back against stable1.10. Whatever is in `master` has no influence on the CI run for other branches

PR #5263, opened 2021-06-30
Fix infinite scoll loading, take II

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/5234

Incorporates https://github.com/ChristophWurst/vue-infinite-scroll/pull/1

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -58,7 +58,7 @@ jobs:
     - name: Set up Node
       uses: actions/setup-node@v2.2.0
       with:
-        node-version: 12.x
+        node-version: 14.x
     - name: npm install
       run: npm ci
     - name: eslint
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -1,6 +1,11 @@
 name: Packaging
 
-on: [push, pull_request]
+on:
+    pull_request:
+    push:
+        branches:
+            - master
+            - stable*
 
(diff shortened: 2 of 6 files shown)


ChristophWurst:
CI says no becaues npm pulls through ssh

miaulalala (approved):

ChristophWurst:
https://github.com/npm/cli/issues/2610#issuecomment-851347583 looks like a promising fix :crossed_fingers:

ChristophWurst:
Now we vendor the lib. Not great but also not terrible for this workaround :shipit:

GretaD (approved):
it works. On inbox view all looks good. While on PI, i need to scroll twice to make it work.

raimund-schluessler:
> 
> 
> Now we vendor the lib. Not great but also not terrible for this workaround :shipit:

@ChristophWurst Is there something wrong with loading libs from github like so?
```
"vue-infinite-scroll": "github:ChristophWurst/vue-infinite-scroll#748261ea9411bdc2e5c7509d9638e02007c2db50",
```

I just ask, because we do it like this in the Tasks repository (see https://github.com/nextcloud/tasks/blob/master/package.json#L43), and I now wonder if this is not a good idea for some reason?

ChristophWurst:
If CI passes then it's fine ;)

For us CI failed because of https://github.com/npm/cli/issues/2610. This was our only blocker.

PR #5292, opened 2021-07-07
Move userId to controller to fix tagging despite option disabled

### LINKED ISSUE #5290: Mails are getting automatically classified even if classification is disabled
### Expected behavior
After #5270 was merged, I was expecting that deselecting that new setting would stop Nextcloud from making virtually all new Mails show up as Important in Thunderbird.

### Actual behavior
Most new mails are still getting Tagged as important, despite the setting being disabled.

### Mail app

**Mail app version:** 1.10.0

**Mailserver or service:** Dovecot

### Server configuration

**Operating system**: Gentoo

**Web server:** Apache

**Database:** PostgreSQL

**PHP version:** 7.4

**Nextcloud Version:** 22.0.0

#### Client configuration
**Browser:** Edge 91

**Operating system:** Windows

miaulalala:
Cron runs did still tag messages because the user ID injection does not work for cron.

Fixes #5290

### DIFF
--- a/lib/Contracts/IUserPreferences.php
+++ b/lib/Contracts/IUserPreferences.php
@@ -29,15 +29,17 @@ namespace OCA\Mail\Contracts;
 interface IUserPreferences {
 
 	/**
+	 * @param string $userId
 	 * @param string $key
 	 * @param mixed $value
 	 * @return mixed new value
 	 */
-	public function setPreference($key, $value);
+	public function setPreference(string $userId, $key, $value);
 
 	/**
+	 * @param string $userId
 	 * @param string $key
 	 * @param mixed|null $default
 	 */
-	public function getPreference($key, $default = null);
+	public function getPreference(string $userId, $key, $default = null);
 }
--- a/lib/Controller/PreferencesController.php
+++ b/lib/Controller/PreferencesController.php
@@ -37,14 +37,19 @@ class PreferencesController extends Controller {
 	/** @var IUserPreferences */
 	private $userPreference;
 
+	/** @var string */
+	private $userId;
+
 	/**
 	 * @param IRequest $request
 	 * @param IUserPreferences $userPreference
+	 * @param string $UserId
 	 */
-	public function __construct(IRequest $request, IUserPreferences $userPreference) {
+	public function __construct(IRequest $request, IUserPreferences $userPreference, string $UserId) {
 		parent::__construct('mail', $request);
 
 		$this->userPreference = $userPreference;
+		$this->userId = $UserId;
 	}
 
 	/**
@@ -56,7 +61,7 @@ class PreferencesController extends Controller {
 	 */
 	public function show(string $id): JSONResponse {
 		return new JSONResponse([
-			'value' => $this->userPreference->getPreference($id)
+			'value' => $this->userPreference->getPreference($this->userId, $id)
 		]);
 	}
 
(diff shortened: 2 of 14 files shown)


lib/Contracts/IUserPreferences.php
```diff
@@ -29,15 +29,18 @@
...
 
 	/**
+	 * @param string $userId
 	 * @param string $key
 	 * @param mixed $value
 	 * @return mixed new value
+	 * @throws \OCP\PreConditionNotMetException
 	 */
-	public function setPreference($key, $value);
+	public function setPreference($userId, $key, $value);
```
ChristophWurst, line 38:
```suggestion
	public function setPreference(string $userId, $key, $value);
```

lib/Contracts/IUserPreferences.php
```diff
@@ -29,15 +29,18 @@
...
 
 	/**
+	 * @param string $userId
 	 * @param string $key
 	 * @param mixed|null $default
 	 */
-	public function getPreference($key, $default = null);
+	public function getPreference($userId, $key, $default = null);
```
ChristophWurst, line 45:
```suggestion
	public function getPreference(string $userId, $key, $default = null);
```

lib/Controller/PreferencesController.php
```diff
@@ -31,20 +31,26 @@
...
 use OCP\IRequest;
+use OCP\PreConditionNotMetException;
 
 class PreferencesController extends Controller {
 
 	/** @var IUserPreferences */
 	private $userPreference;
 
+	/** @var string */
+	private $UserId;
```
ChristophWurst, line 42:
```suggestion
	private $userId;
```

capitalization on the first letter is only required for the argument on the constructor. The internal member we can case correctly :)

ChristophWurst (approved):

lib/Controller/PreferencesController.php
miaulalala: Ah good to know! Will change :)

lib/Contracts/IUserPreferences.php
```diff
@@ -29,15 +29,18 @@
 interface IUserPreferences {
 
 	/**
+	 * @param string $userId
 	 * @param string $key
 	 * @param mixed $value
 	 * @return mixed new value
+	 * @throws \OCP\PreConditionNotMetException
```
kesselb, line 36: We don't pass $preCondition to setUserValue so the method will not throw a PreConditionNotMetException. I would prefer to not list in phpdocs.

kesselb:
Needs an update:

- https://github.com/nextcloud/mail/blob/267f964aa37c292dc1551d6e7687942e5eec2749/lib/Listener/AddressCollectionListener.php#L59
- https://github.com/nextcloud/mail/blob/0a1e1ad8bc46bb4d4b0846b840cd6b8a4ff3c408/lib/Service/AvatarService.php#L81

lib/Controller/PreferencesController.php
```diff
@@ -74,7 +80,11 @@ public function update($key, $value): JSONResponse {
 			throw new ClientException('key or value missing');
 		}
 
-		$newValue = $this->userPreference->setPreference($key, $value);
+		try {
+			$newValue = $this->userPreference->setPreference($this->userId, $key, $value);
+		} catch (PreConditionNotMetException $e) {
+			throw new ClientException('key or value invalid');
```
kesselb, line 86: Please remove try-catch here. The method will not throw a PreConditionNotMetException for us.

lib/Controller/PreferencesController.php
```diff
@@ -31,20 +31,26 @@
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\IRequest;
+use OCP\PreConditionNotMetException;
```
kesselb, line 34:
```suggestion
```

ChristophWurst:
please rebase and squash

ChristophWurst (approved):

Issue #3006, opened 2020-04-27
Give more insights into setup errors


ChristophWurst:
### Feature Request

During setup we use very generic error messages for when the IMAP or SMTP connection fails. We should be able to differentiate between some of them by the error code and thus show more helpful information to the user.

#### Summary

n/a

sunjam:
In my case on Mail 1.3.4 on v19.0.0 beta5, it would be helpful to auto-fill the `imap server` and `smtp server` with my domain information automatically.

`email@example.com`
`imap server: example.com`
`smtp server: example.com`

I think it is also helpful to have a simple drop-down menu to select common email ports, such as imap `995` or `993` or `465`in the automatic configuration menu.  Thanks!

ChristophWurst:
But how? This is a very limiting assumption. For many users this will autofill incorrect suggestions.

> I think it is also helpful to have a simple drop-down menu to select common email ports, such as imap `995` or `993` or `465`in the automatic configuration menu. Thanks!

We already do that. If you select the security mode we pick the corresponding port. A drop down would limit the set of possible values and there are always some special cases where the user needs the ability to override defaults and use custom combinations.

sunjam:
> But how? This is a very limiting assumption. For many users this will autofill incorrect suggestions.
> 
> > I think it is also helpful to have a simple drop-down menu to select common email ports, such as imap `995` or `993` or `465`in the automatic configuration menu. Thanks!
> 
> We already do that. If you select the security mode we pick the corresponding port. A drop down would limit the set of possible values and there are always some special cases where the user needs the ability to override defaults and use custom combinations.

My limited assumption is from testing the app.  I see no drop-down method for selecting ports in auto or manual. Everything must be written in manually, or I have this auto menu.  Perhaps this feedback is useless; keep up the good work.
![Screenshot_20200428_094557](https://user-images.githubusercontent.com/1787238/80514609-97b95900-8935-11ea-8f1d-8907912ecf1e.png)

adefaria:
I absolutely hate error messages like "Something went wrong" or "Error while doing <x>". It's lazy unprofessional at best. You leave the user with the very natural questions of "what went wrong" and "why did it fail" providing absolutely nothing to answer either of those questions. What I always say is that the programmer, when encountering an error is the best person in the whole wide world to explain what was going on and how something failed as he has the most knowledge of what he was trying to do and how he was trying to do it.

[...]

PR #5317, opened 2021-07-12
Fix error reporting during mail setup

CarlSchwan:
Previously only the generic (and not useful) error message was
displayed when a more useful message was available and could have
been displayed instead.

The error message is located in error.data.message and not error.message

image

### DIFF
--- a/src/views/Setup.vue
+++ b/src/views/Setup.vue
@@ -58,8 +58,8 @@ export default {
 				.catch((error) => {
 					logger.error('Could not create account', { error })
 
-					if (error.message) {
-						this.error = error.message
+					if (error?.data?.message) {
+						this.error = error.data.message
 					} else {
 						this.error = t('mail', 'Unexpected error during account creation')
 					}


ChristophWurst:
Thanks *a lot*.

I guess this then (at least partially) solves https://github.com/nextcloud/mail/issues/3006.

src/views/Setup.vue
```diff
@@ -58,8 +58,8 @@ export default {
 				.catch((error) => {
 					logger.error('Could not create account', { error })
 
-					if (error.message) {
-						this.error = error.message
+					if (error.data.message) {
```
ChristophWurst, line 61:
```suggestion
					if (error?.data.message) {
```

just in case a non-axios error is thrown, those don't have a `data` prop

ChristophWurst:
Now i remember that the message isn't always set. Did you try this with debug mode on or off? I think message is omitted in production mode.

ognarb:
> Now i remember that the message isn't always set. Did you try this with debug mode on or off? I think message is omitted in production mode.

I just tested with "debug" => false in my config.php and this particular message is still displayed.

GretaD:
On master, when you add wrong data you always get: `Unexpected error during account creation`

On this branch: 
When i add the host but wrong pasword: `Creating account failed:Connection to IMAP at mx.nextcloud.com:993 failed. Authentication failed.`

when i "forget" to add the host: `Creating account failed:Connection to IMAP at :993 failed. Error connecting to mail server.`

So this is a great improvement already :)

GretaD:
You need to squash and sign your pr and its good to go

CarlSchwan:
> On master, when you add wrong data you always get: `Unexpected error during account creation`
> 
> On this branch:
> When i add the host but wrong pasword: `Creating account failed:Connection to IMAP at mx.nextcloud.com:993 failed. Authentication failed.`
> 
> when i "forget" to add the host: `Creating account failed:Connection to IMAP at :993 failed. Error connecting to mail server.`
> 
> So this is a great improvement already :)

Thanks :)

Some wording could probably be improved but that's probably material for another MR :)

I squashed both commit and signed the MR.

ChristophWurst (approved):

PR #5319, opened 2021-07-12
Make spacing consistent with chat app for list item

CarlSchwan:
image

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -445,6 +445,10 @@ export default {
 list-item-style.draft .app-content-list-item-line-two {
 	font-style: italic;
 }
+.list-item-style {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@ognarb with subtle design changes it's best to provide a before and after screenshot in the PR, then it's easier for us to spot the difference :)

GretaD:
Hey @ognarb, thank you for your first contribution on mail :)

Do you know that the second column here on mail is re-sizable? I am asking because this change would be applicable only on size that is exactly as wide as the Talk chat one. Otherwise people who decide to have it wider or smaller, this doesnt change anything.

GretaD:
this pr its not relevant anymore, because of the resizing, but also because on mail we have the account color vertical line, and this pr breaks that.

PR #5328, opened 2021-07-14
Fix missing slash in sharing file generated url

mepsteinj:
Fixes https://github.com/nextcloud/mail/issues/5307


st3iny (approved):
Thanks for the fix :)

CI failure seems unrelated.

PR #5393, opened 2021-08-04
Add parent domain to imap/smtp check

### LINKED ISSUE #5372: Improve auto configuration for mail accounts
### Feature Request

* [x] Remove connectivity check for port 585. According to Wikipedia port 585 was intended for IMAP over TLS/SSL but is deregistered in favour of port 993.

* [x] Hosting providers may split the services. For example the mx record for a domain test.com might be mx.provider.com. But the imap service is reachable at imap.provider.com. Mail already has a check for [imap.mx.provider.com](https://github.com/nextcloud/mail/blob/6b5e0b9db3d0338a8e7adffe70d8128fa8e93419/lib/Service/AutoConfig/ImapConnectivityTester.php#L75). We should consider to check also the parent domain imap.provider.com.

miaulalala:
Fixes #5372

### DIFF
--- a/lib/Service/AutoConfig/MxRecord.php
+++ b/lib/Service/AutoConfig/MxRecord.php
@@ -56,4 +56,17 @@ class MxRecord {
 		// TODO: sort by weight
 		return $mxRecords;
 	}
+
+	private function stripSubdomain(string $domain): string {
+		$labels = explode('.', $domain);
+
+		$top = count($labels) >= 2 ? array_pop($labels) : '';
+		$second = array_pop($labels);
+
+		return $second . '.' . $top;
+	}
+
+	public function getSanitizedRecords(array $mxHosts): array {
+		return array_unique(array_merge($mxHosts, array_map([$this, 'stripSubdomain'], $mxHosts)));
+	}
 }
--- a/tests/Unit/Service/Autoconfig/MxRecordTest.php
+++ b/tests/Unit/Service/Autoconfig/MxRecordTest.php
@@ -40,16 +40,36 @@ class MxRecordTest extends TestCase {
 		$this->record = new MxRecord($logger);
 	}
 
-	public function testQuery() {
+	public function testQuery(): void {
 		$records = $this->record->query('nextcloud.com');
 
 		$this->assertIsArray($records);
 		$this->assertNotEmpty($records);
 	}
 
-	public function testQueryNoRecord() {
+	public function testQueryNoRecord(): void {
 		$records = $this->record->query('example.com');
 
 		$this->assertFalse($records);
+	}
+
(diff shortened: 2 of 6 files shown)


ChristophWurst (commented):
This looks complicated. Could we have tests? :)

lib/Service/AutoConfig/MxRecord.php
```diff
@@ -56,4 +56,17 @@ public function query(string $host) {
...
+	public function stripSubdomain(string $domain): string {
+		$labels = explode('.', $domain);
+
+		$top = count($labels) >= 2 ? array_pop($labels) : '';
+		$second = array_pop($labels);
+
+		return $second . '.' . $top;
+	}
+
+	public function getAllRecords(array $mxHosts): array {
```
kesselb, line 69: I would say the method is not getting all records but sanitize a list of records. Can we find a better name for it?

lib/Service/AutoConfig/MxRecord.php
```diff
@@ -56,4 +56,17 @@ public function query(string $host) {
 		// TODO: sort by weight
 		return $mxRecords;
 	}
+
+	public function stripSubdomain(string $domain): string {
```
kesselb, line 60: I suggest to make the method private. We can use `$this->invokePrivate($object, 'methodName');` in a test case to execute private methods.

lib/Service/AutoConfig/MxRecord.php
miaulalala: sure, getSanitizedRecords?

tests/Unit/Service/Autoconfig/MxRecordTest.php
```diff
@@ -52,4 +52,29 @@ public function testQueryNoRecord() {
 
 		$this->assertFalse($records);
 	}
+
+	public function testStripSubdomain() {
```
ChristophWurst, line 56:
```suggestion
	public function testStripSubdomain(): void {
```

tests/Unit/Service/Autoconfig/MxRecordTest.php
```diff
@@ -52,4 +52,29 @@ public function testQueryNoRecord() {
...
+		$record = $this->record->stripSubdomain('mx.google.com');
+		$this->assertEquals('google.com', $record);
+
+		$record = $this->record->stripSubdomain('shgfghd.fhdjfhsdf.gggg.google.com');
+		$this->assertEquals('google.com', $record);
+	}
+
+	public function testGetAllRecords() {
```
ChristophWurst:
```suggestion
	public function testGetAllRecords(): void {
```

ChristophWurst (approved):
It aint much, but it's honest typing

PR #5397, opened 2021-08-06
fix: handle invalid imap message id

rti:
When IMAPMessage.toDbMessage() encountered an invalid messageId it was
creating a DB Message without messageId. This leads for example to
TagMapper::tagMessage() to fail because it must not be called with a
null messageId.

This patch leverages the functionality to create a valid messageId in
case the messageId was missing completely. It now uses it as well for
invalid messageIds not accepted by Horde_Mail_Rfc822_Identification and
therefore not set by Message.setMessageId().

An example for an invalid messageId I encountered in my inbox is
Message-ID: <359f166ea1b402e793c0801834a9722a-support@aruba-studios.de
(The closing > bracket is missing).

PR #5398, opened 2021-08-06
fix: handle invalid imap message id

miaulalala:
When IMAPMessage.toDbMessage() encountered an invalid messageId it was
creating a DB Message without messageId. This leads for example to
TagMapper::tagMessage() to fail because it must not be called with a
null messageId.

This patch leverages the functionality to create a valid messageId in
case the messageId was missing completely. It now uses it as well for
invalid messageIds not accepted by Horde_Mail_Rfc822_Identification and
therefore not set by Message.setMessageId().

An example for an invalid messageId I encountered in my inbox is
Message-ID: <359f166ea1b402e793c0801834a9722a-support@aruba-studios.de
(The closing > bracket is missing).

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -36,7 +36,7 @@ use function json_encode;
 /**
  * @method void setUid(int $uid)
  * @method int getUid()
- * @method string getMessageId()
+ * @method string|null getMessageId()
  * @method void setReferences(string $references)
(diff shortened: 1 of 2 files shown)


miaulalala:
See #5397 for the original PR by @rti - thanks for finding this tricky bug and your commit!

ChristophWurst (approved):

PR #5427, opened 2021-08-11
Do not trim null message IDs

ChristophWurst:
Regression of https://github.com/nextcloud/mail/pull/5398.


kesselb (approved):
Sure. Cannot trim null :see_no_evil:

Issue #5108, opened 2021-06-02
An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '19376-8' for key 'mail_msg_mb_uid_idx'


Megachip:
### Expected behavior
Working Mail Add-on

### Actual behavior
Loads forever 

### Mail app

**Mail app version:** 1.9.5

**Mailserver or service:** own

### Server configuration

**Operating system**: Linux ... 4.9.0-12-amd64 #1 SMP Debian 4.9.210-1+deb9u1 (2020-06-07) x86_64 GNU/Linux

**Web server:** nginx

**Database:** mysql 5.7.33

**PHP version:** 7.4.18

**Nextcloud Version:** 21.0.2

#### Error
```
[mail] Error: OCA\Mail\Exception\ServiceException: Sync failed for 1:INBOX: An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '19376-8' for key 'mail_msg_mb_uid_idx' at <<closure>>

0. /.../apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php line 128
   OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 8}, OC\AppFramework\ScopedPsrLogger {}, 42, null, false, true)
1. /.../apps/mail/lib/BackgroundJob/SyncJob.php line 92
   OCA\Mail\Service\Sync\ImapToDbSynchronizer->syncAccount(OCA\Mail\Account {}, OC\AppFramework\ScopedPsrLogger {})
2. /.../lib/public/BackgroundJob/Job.php line 80
   OCA\Mail\BackgroundJob\SyncJob->run({accountId: 1})
3. /.../lib/public/BackgroundJob/TimedJob.php line 61
   OCP\BackgroundJob\Job->execute(OC\BackgroundJob\JobList {}, OC\Log {})
4. /.../cron.php line 150
   OCP\BackgroundJob\TimedJob->execute(OC\BackgroundJob\JobList {}, OC\Log {})

GET /cron.php
from ... at 2021-06-02T19:18:53+00:00
```

Megachip:
Maybe duplicate of #5019

ChristophWurst:
indeed a duplicate

Issue #5019, opened 2021-05-08
Gmail inbox stopped working


dakontiva:
### Expected behavior
Inboxes of all email accounts should work

### Actual behavior
Gmail inbox stopped working. Also, even though other email accounts are working fine when I select each inbox individually, the all inboxes view is stuck at loading messages ever since Gmail inbox stopped working.

I checked the logs and found this error:

```
[mail] Error: OCA\Mail\Exception\ServiceException: Sync failed for 2:INBOX: An exception occurred while executing 'INSERT INTO `oc_mail_messages` (`uid`, `message_id`, `references`, `in_reply_to`, `thread_root_id`, `mailbox_id`, `subject`, `sent_at`, `flag_answered`, `flag_deleted`, `flag_draft`, `flag_flagged`, `flag_seen`, `flag_forwarded`, `flag_junk`, `flag_notjunk`, `flag_important`, `flag_mdnsent`) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params [86499, "<000000000000882e5d05c1c5ca95@google.com>", "[]", null, "<000000000000882e5d05c1c5ca95@google.com>", 12, "Your Google data is ready to download", 1620430134, false, false, false, false, false, false, false, false, false, false]:

SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '86499-12' for key 'mail_msg_mb_uid_idx' at <<closure>>

0. /data/www/anirudh.owncube.com/apps/mail/lib/Service/Sync/SyncService.php line 121
   OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 12}, OC\AppFramework\ScopedPsrLogger {}, 42, [86218,86230,862 ... "], false)
1. /data/www/anirudh.owncube.com/apps/mail/lib/Controller/MailboxesController.php line 161
   OCA\Mail\Service\Sync\SyncService->syncMailbox(OCA\Mail\Account {}, OCA\Mail\Db\Mailbox {id: 12}, 42, [15829,15828,158 ... "], true, null)
2. /data/www/anirudh.owncube.com/lib/private/AppFramework/Http/Dispatcher.php line 169
   OCA\Mail\Controller\MailboxesController->sync(12, [15829,15828,158 ... "], false, null)
[...]

PR #5440, opened 2021-08-12
Ignore already known UIDs when looking for new ones to satisfy mail_msg_mb_uid_idx

ChristophWurst:
Fixes issues like https://github.com/nextcloud/mail/issues/5108, https://github.com/nextcloud/mail/issues/5019 and https://github.com/nextcloud/mail/issues/4341

Basically if one sync attempt fails mid way after it had already inserted messages of one chunk (our level of atomicity for transacted and rolled back inserts) then the next sync will fail until the eternity. This patch discards duplicates when we determine new UIDs. Therefore the duplicates are avoided.

### DIFF
--- a/lib/Service/Sync/ImapToDbSynchronizer.php
+++ b/lib/Service/Sync/ImapToDbSynchronizer.php
@@ -51,6 +51,7 @@ use OCP\EventDispatcher\IEventDispatcher;
 use Psr\Log\LoggerInterface;
 use Throwable;
 use function array_chunk;
+use function array_filter;
 use function array_map;
 
 class ImapToDbSynchronizer {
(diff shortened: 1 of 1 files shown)


st3iny (commented):
Code looks good. How could I test this?

ChristophWurst:
> Code looks good. How could I test this?

Throw an arbitrary exception after https://github.com/nextcloud/mail/blob/06edd02616adde377ea2488a8395ff4d8acdc9c4/lib/Db/MessageMapper.php#L347. Then send a message to your inbox and sync again. Your error will show up, but the new message is stored in the DB. Then remove the exception and try again. On master you are in an unrecoverable state. On this branch the code will be able to continue.

ChristophWurst:
FYI deplyed this on my personal instance that was stuck before and it revived my inbox from the dead :raised_hands:

miaulalala (approved):
LGTM

ramezrafla:
@ChristophWurst 
Thanks so much for this change - a mailbox that was locked also is released now and functional.
Well done!

Issue #5322, opened 2021-07-13
thread list grouping not working 100%


kinimodmeyer:
How are you grouping the thread list?

In the screenshoot you find 2 spam mails "Rosie Burrows".
The mail app threaded it correctly but in the list it is seperated:

![Bildschirmfoto von 2021-07-13 08-17-32](https://user-images.githubusercontent.com/3705274/125400949-d0b2e180-e3b2-11eb-9cc9-e14700a7443f.png)

header first e-mail: [1.txt](https://github.com/nextcloud/mail/files/6806334/1.txt)
header second e-mail: [2.txt](https://github.com/nextcloud/mail/files/6806333/2.txt)
(SPAM / no confidential infromation included)

I have other exmaples where threading not works but step by step :)

ChristophWurst:
:eyes: @kesselb this seems to be the issue I experienced as well yesterday

I can't reliably reproduce this. sometimes a page reload correctly groups the messages again in the list.

> The mail app threaded it correctly but in the list it is seperated

That's also my findings. I think the threading in the back-end is correct. This is merely an issue of how we group the threads for the thread list.

PR #5106, opened 2021-06-01
Only show each thread once per message list

kesselb:
Fixes #3652 

**Todo**

- [x] Endpoint to delete / move a thread
- [x] Delete in envelope list should delete all messages (excl. trash)
- [x] Move in envelope list should move all messages (excl. trash)
- [x] Delete the latest message in thread view should add the second latest message to the envelope list<sup>1</sup>
- [x] Move the latest message in thread view should add the second latest message to the envelope list<sup>1</sup>
- [x] Select multiple threads/messages in envelope to delete/move must still work
- [x] Keyboard shortcuts
- [x] Rename delete to delete thread in envelope list
- [x] Rename move to move thread in envelope list

<sup>1</sup> Is done by the existing sync backend request 

_Technical details and design decisions: https://github.com/nextcloud/mail/pull/4970._

PR #5453, opened 2021-08-18
Add store for threads

kesselb:
**Story** 

Update front end to handle message threads better. After https://github.com/nextcloud/mail/pull/5106 the back end provide messages in a threaded way for the message list. For each thread the back end sent one message. Yet there are some (known) edge cases:

- #5322 when a new message is available for an existing/already shown thread
- Delete the latest message in thread view should add the second latest message to the envelope list
- Move the latest message in thread view should add the second latest message to the envelope list

**Todo**
- [x] A new message for the same thread replace the old message in the envelope list 
- [ ] Update envelope / thread view (right now it blanks after a while due a removeEnvelope action)

Fixes https://github.com/nextcloud/mail/issues/5322

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -23,17 +23,65 @@ import Vue from 'vue'
 import Vuex from 'vuex'
 
 import {
+	PRIORITY_INBOX_ID,
 	UNIFIED_ACCOUNT_ID,
 	UNIFIED_INBOX_ID,
-	PRIORITY_INBOX_ID,
 } from './constants'
 import actions from './actions'
 import { getters } from './getters'
 import mutations from './mutations'
-
+import { normalizedEnvelopeListId } from './normalization'
+import orderBy from 'lodash/fp/orderBy'
 Vue.use(Vuex)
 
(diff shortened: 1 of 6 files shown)


src/store/index.js
```diff
@@ -23,17 +23,65 @@ import Vue from 'vue'
...
+	},
+	mutations: {
+		addAccount(state, account) {
+			const mailboxes = account.mailboxes || []
+			mailboxes.map(mailboxId => Vue.set(state.lists, mailboxId, {}))
+		},
+		addMailbox(state, { mailbox }) {
+			Vue.set(state.lists, mailbox.databaseId, {})
+		},
+		addEnvelope(state, { query, envelope }) {
```
kesselb, line 52: When a vuex module is not namespaced it reacts to the existing mutations. A commit for addEnvelope trigger the existing mutation and our new one.

ChristophWurst:
conflicts. is this still relevant?

ChristophWurst:
close?

PR #5490, opened 2021-08-30
Show reply settings loading spinner correctly

ChristophWurst:
* Shows when reply settings are toggled
* Does not show when avatar settings are toggled


kesselb (approved):
:see_no_evil:

GretaD (approved):

Issue #679, opened 2017-12-17
Add possibility to change sort attribute


Friptick:
As a user I want to  change how my emails are sorted

## Todo

- [x] Change sort order
- [ ] Change sort attribute

---

Original report

<details>

There does not seem to be an option to change the mail sort order.

True?

Some users like to read mail in the order it was received.

**Mail app version:** 0.7.6

</details>

ChristophWurst:
True!

jancborchardt:
> Some users like to read mail in the order it was received.

That’s the way mails are sorted, no? And the great thing is that if you delete the current message, it goes to the next newest one.

As soon as there’s threading #2125, you can also read threads in order.

Hence closing because we de facto have it, and also cause of lots of overhead vs little interest during 2 years since the issue is open.

Friptick:
> > Some users like to read mail in the order it was received.
> 
> That’s the way mails are sorted, no? 

NO. it is absolutely *not* the way they are sorted. They are sorted anti-chronologically, like a blog. Not chronologically, AKA "in the order in which they were received".

Meaning that if you deal with incoming mail *chronologically* (which is logical and polite), you are constantly reading from bottom to top. Ridiculous. And why I still do not use Nextcloud Mail.

> As soon as there’s threading #2125, you can also read threads in order.

That will be great, but it is absolutely not the same thing.

> Hence closing because we de facto have it

But we do not.

> and also cause of lots of overhead vs little interest during 2 years since the issue is open.

This is a better reason. Unfortunately it is a deal-breaker for me, but I guess that's my problem. Seemed like an easy fix. Oh well.

ChristophWurst:
Re-opening because I see how this is an important feature for some people.

Historically it was a close to impossible to implement as we had to rely on server sorting (often not there because of missing IMAP capabilities) or doing client-side sorting, which is very inefficient as we have to fetch the full mailbox, sort it, and just take a small chunk.

As of #2064 we have the messages and their important attributes locally. We can thus quickly and efficiently change the sort order. Of course right now lots of logic just assumes that we have sort by date, but that should be okay to change. Then we can allow sorting by more fields and also changing descending/ascending.

alexanderdd:
can someone change the title of this issue to "sort by date asc/desc, read/unread, starred/unstarred, has/no attachment"
Cause I think this is what is needed in the long run, and it includes the original request. 
If not, I will open a new issue with that title.

[...]

PR #5495, opened 2021-09-01
Allow users to see oldest messages first

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/679

Tasks
* [x] Add UI to toggle between newest first or oldest first
* [x] Send search preference to back-end with each query
* [x] Adjust front-end data storage to search preference
* [x] Fix sync for new messages when latest page is reached
* [x] Adjust `needsFetch` comparisons to differently ordered timestamps

Follow-ups
* Possibly allow switching the sort order without a full page reload


lib/Service/Sync/SyncService.php
```diff
@@ -151,24 +158,35 @@ public function syncMailbox(Account $account,
 	private function getDatabaseSyncChanges(Account $account,
 											Mailbox $mailbox,
 											array $knownIds,
+											?int $lastMessageTimestamp,
+											string $sortOrder,
 											?SearchQuery $query): Response {
-		if (empty($knownIds)) {
+		if (empty($knownIds) || $lastMessageTimestamp === null) {
```
ChristophWurst, line 164:
```suggestion
		if (empty($knownIds) && $lastMessageTimestamp === null) {
```

but not sure

kesselb:
Priority inbox / All inboxes

Screenshot from 2021-09-20 21-02-41

I have two mail accounts configured. On scroll down the next messages are loaded from both accounts. We see the oldest messages from mail account A and B next to each other.

src/store/actions.js
```diff
@@ -94,7 +95,13 @@ const findIndividualMailboxes = curry((getMailboxes, specialRole) =>
 	)
 )
 
-const combineEnvelopeLists = pipe(flatten, orderBy(prop('dateInt'), 'desc'))
+const combineEnvelopeLists = (sortOrder) => {
+	if (sortOrder === 'oldest-first') {
+		return pipe(flatten, orderBy(prop('dateInt'), 'desc'))
```
kesselb, line 100: This looks wrong but it seems to work locally with desc :man_shrugging:

src/store/mutations.js
```diff
@@ -170,29 +174,53 @@ export default {
...
 			})
+			return
+		}
+
+		envelopes.forEach((envelope) => {
+			const existing = mailbox.envelopeLists[listId] || []
+			normalizeTags(state, envelope)
+			Vue.set(state.envelopes, envelope.databaseId, Object.assign({}, state.envelopes[envelope.databaseId] || {}, envelope))
+			Vue.set(envelope, 'accountId', mailbox.accountId)
+			Vue.set(mailbox.envelopeLists, listId, uniq(orderByDateInt(existing.concat([envelope.databaseId]))))
```
kesselb, line 209: Could we update the existing envelope list in one go for all envelopes?

ChristophWurst:
> I have two mail accounts configured. On scroll down the next messages are loaded from both accounts. We see the oldest messages from mail account A and B next to each other.

My observation: the first page is correct, then, when you load more, it actually doesn't just add messages to the end of the list but also in between. This is why the previously last element also jumps.

src/store/actions.js
ChristophWurst: This is wrong and caused https://github.com/nextcloud/mail/pull/5495#issuecomment-929907448. After I changed to asc that quirk goes away. Thanks :hugs:

ChristophWurst:
> I have two mail accounts configured. On scroll down the next messages are loaded from both accounts. We see the oldest messages from mail account A and B next to each other.

I can't reproduce this. Could you try with the latest version?

GretaD:
I tested this one and i noticed that when i click show oldest first, the first active message its not open, so the third column is empty. Same when i change it to newest.

lib/Db/MessageMapper.php
```diff
@@ -151,13 +153,32 @@ public function findAllUids(Mailbox $mailbox): array {
...
 
+		if ($limit !== null) {
+			$query->setMaxResults($limit);
+		}
+		if ($sortOrder === IMailSearch::ORDER_NEWEST_FIRST) {
+			$query->orderBy('sent_at', 'desc');
+		} elseif ($sortOrder === IMailSearch::ORDER_OLDEST_FIRST) {
+			$query->orderBy('sent_at', 'asc');
+		} elseif ($sortOrder !== null) {
+			throw new InvalidArgumentException('Invalid sort orderd');
```
jancborchardt, line 179:
```suggestion
			throw new InvalidArgumentException('Invalid sort order');
```

ChristophWurst:
Stale and terribly conflicting

PR #5496, opened 2021-09-01
Remove subject from same thread

### LINKED ISSUE #5481: Remove thread subject of individual message unless the subject changed within the thread (excluding FWD/RE prefixes)

GretaD:
fixes #5481

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -46,10 +46,10 @@
 				:class="{seen: envelope.flags.seen}"
 				@click.native.prevent="$emit('toggleExpand', $event)">
 				<span class="sender">{{ envelope.from && envelope.from[0] ? envelope.from[0].label : '' }}</span>
-				<div class="subject">
+				<div v-if="hasChangedSubject" class="subject">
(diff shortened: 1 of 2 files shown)


src/components/ThreadEnvelope.vue
```diff
@@ -206,6 +210,16 @@ export default {
 		tags() {
 			return this.$store.getters.getEnvelopeTags(this.envelope.databaseId).filter((tag) => tag.imapLabel !== '$label1')
 		},
+		hasChangedSubject() {
+			return this.cleanSubject !== this.threadSubject
+		},
+		cleanSubject() {
+			let cleanSubject = this.envelope.subject
+			while (cleanSubject.startsWith('Re')) {
```
ChristophWurst, line 218: you can use a regex (possibly the one from below) with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test and test for all possible prefixes

ChristophWurst (approved):

GretaD:
need to fix the tagging design break caused by this, then we can merge

PR #5513, opened 2021-09-08
Move message composer into a modal

### LINKED ISSUE #5483: Move new message composer into a modal

GretaD:
fixes #5483 

- [x] Add modal
- [x] Add a dedicated Discard draft button
~~- [x] Add a dedicated Save draft button~~
- [x] The route /apps/mail/box/priority/thread/new still gives me the old composer, so there is some leftover code to remove, that URL shouldn't exist/work anymore
- [x] Make height 60%
- [x] Draft button and save should be shown all the time because it will be used as close button as well
~~- [ ] Move discard button on the bottom-left~~
- [x] Possibly add some padding so the text doesn’t stick so close to edges

Follow up:
- [ ] Let's not make the modal closeable
- [ ] Improve the way attachment look

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -150,6 +150,12 @@
 				:upload-size-limit="attachmentSizeLimit"
 				@upload="onAttachmentsUploading" />
 			<div class="composer-actions-right">
+				<button v-if="savingDraft === false"
+					class="button"
+					:title="t('mail', 'Discard & close draft')"
+					@click="discardDraft">
+					{{ t('mail', 'Discard draft') }}
+				</button>
 				<p class="composer-actions-draft">
 					<span v-if="!canSaveDraft" id="draft-status">{{ t('mail', 'Cannot save draft because this account does not have a drafts mailbox configured.') }}</span>
 					<span v-else-if="savingDraft === true" id="draft-status">{{ t('mail', 'Saving draft …') }}</span>
@@ -212,7 +218,14 @@
 		</div>
 	</div>
 	<Loading v-else-if="state === STATES.UPLOADING" :hint="t('mail', 'Uploading attachments …')" role="alert" />
-	<Loading v-else-if="state === STATES.SENDING" :hint="t('mail', 'Sending …')" role="alert" />
+	<Loading v-else-if="state === STATES.SENDING"
+		:hint="t('mail', 'Sending …')"
+		role="alert"
+		class="sending-hint" />
+	<Loading v-else-if="state === STATES.DISCARDING" :hint="t('mail', 'Discarding …')" class="emptycontent" />
+	<EmptyContent v-else-if="state === STATES.DISCARDED" icon="icon-mail">
+		<h2>{{ t('mail', 'Draft was discarded!') }}</h2>
+	</EmptyContent>
 	<div v-else-if="state === STATES.ERROR" class="emptycontent" role="alert">
 		<h2>{{ t('mail', 'Error sending your message') }}</h2>
 		<p v-if="errorText">
@@ -237,12 +250,9 @@
 			{{ t('mail', 'Send anyway') }}
 		</button>
 	</div>
-	<div v-else class="emptycontent">
+	<EmptyContent v-else icon="icon-checkmark">
 		<h2>{{ t('mail', 'Message sent!') }}</h2>
-		<button v-if="!isReply" class="button primary" @click="reset">
-			{{ t('mail', 'Write another message') }}
-		</button>
-	</div>
+	</EmptyContent>
 </template>
 
 <script>
@@ -256,6 +266,7 @@ import Actions from '@nextcloud/vue/dist/Components/Actions'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
 import ActionCheckbox from '@nextcloud/vue/dist/Components/ActionCheckbox'
 import ActionLink from '@nextcloud/vue/dist/Components/ActionLink'
+import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
 import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
 import { showError } from '@nextcloud/dialogs'
 import { translate as t } from '@nextcloud/l10n'
@@ -292,6 +303,8 @@ const STATES = Object.seal({
 	ERROR: 3,
 	WARNING: 4,
 	FINISHED: 5,
+	DISCARDING: 6,
+	DISCARDED: 7,
 })
 
 export default {
@@ -306,6 +319,7 @@ export default {
 		Loading,
 		Multiselect,
 		TextEditor,
+		EmptyContent,
 	},
 	props: {
 		fromAccount: {
@@ -349,6 +363,11 @@ export default {
 			type: Object,
 			required: false,
 			default: () => undefined,
+		},
+		forwardedMessages: {
+			type: Array,
+			required: false,
+			default: () => [],
 		},
 	},
 	data() {
@@ -490,18 +509,6 @@ export default {
 			this.$refs.toLabel.$el.focus()
 		}
 
-		// event is triggered when user clicks 'new message' in navigation
-		this.$root.$on('newMessage', () => {
-			this.draftsPromise
-				.then(() => {
-					return this.saveDraft(this.getMessageData)
-				})
-				.then(() => {
-					// wait for the draft to be saved before resetting the message content
-					this.reset()
-				})
-		})
-
 		// Add attachments in case of forward
 		if (this.forwardFrom?.attachments !== undefined) {
 			this.forwardFrom.attachments.forEach(att => {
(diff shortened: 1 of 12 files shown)


src/components/MailboxThread.vue
```diff
@@ -60,7 +60,7 @@
 			</AppContentList>
 		</template>
 		<NewMessageDetail v-if="newMessage" />
```
ChristophWurst, line 62: is `NewMessageDetail` dead now?

st3iny (commented):
- [ ] Looks nice but I found a bug:

1. Open a thread.
2. Click reply.
3. While editing the reply, create a new message.
4. The new modal won't open.

src/components/MailboxThread.vue
GretaD: no, how dare you?

GretaD:
> Looks nice but I found a bug:
> 
> 1. Open a thread.
> 2. Click reply.
> 3. While editing the reply, create a new message.
> 4. The new modal won't open.

Did you try to turn it on and off again?

ChristophWurst (changes requested):
* Let's not make the modal closeable
  + Add a dedicated *Discard draft* button
  + Add a dedicated *Save draft* button
* The route `/apps/mail/box/priority/thread/new` still gives me the old composer, so there is some leftover code to remove, that URL shouldn't exist/work anymore

src/components/Composer.vue
```diff
@@ -200,6 +200,13 @@
 						}}
 					</ActionLink>
 				</Actions>
+				<button v-if="savingDraft === undefined"
```
ChristophWurst, line 203: now you see *Discard draft* only when no draft has been saved yet, as soon as there is one the button vanishes, don't you?

jancborchardt:
@GretaD then I’d say let’s put "Discard draft" into the 3-dot menu. Should be last entry in there, and when clicked, the composer can close automatically. That’s better than the little delete icon :)

GretaD:
Now it looks like this. Not to confuse people with _**draft saved**_ and _**save draft**_(now this look unnecessary, but we need it) should we rename "save and close draft" instead?

modalbutton

jancborchardt:
> Why don't we use a stateful button for saving a draft? When there is no ongoing save it is enabled and has the title "Save draft". If there is a save in progress (either manually or automatically in the background) it is disabled and shows a title like "Saving draft ...".

Sounds good, yep.

GretaD:
> > Why don't we use a stateful button for saving a draft? When there is no ongoing save it is enabled and has the title "Save draft". If there is a save in progress (either manually or automatically in the background) it is disabled and shows a title like "Saving draft ...".
> 
> Sounds good, yep.

ok, then how do we close the modal after saving? The whole idea of this button is to close the modal after saving. We dont need a save button as a disabled state after we save the draft, because the next step(in the background) is closing the modal. So if i click save, the draft is saved and the modal is closed. @jancborchardt can you please elaborate in steps how should i do it?

ChristophWurst:
Is it intended that we still open replies at the old editor position instead of the modal?

ChristophWurst:
- [ ] BUG *Draft saved* stays for second message if previous one was sent

* Open the modal for a new message
* Wait for a draft to be saved
* Send the message
* Click button to composer another message
* *Draft saved* is still shown

It looks like the internal component state is not properly reset.

GretaD:
> Is it intended that we still open replies at the old editor position instead of the modal?
> 

We didnt discuss anything related to _forward_ or _reply_. The scope of this PR was only Moving _New message_ into a modal. Gmail uses _modal_(whatever is their popup called) only for _New message._

ChristophWurst:
> > Is it intended that we still open replies at the old editor position instead of the modal?
> 
> We didnt discuss anything related to _forward_ or _reply_. The scope of this PR was only Moving _New message_ into a modal. Gmail uses _modal_(whatever is their popup called) only for _New message._

Okay, fine with me if this is intentional.

src/components/Composer.vue
```diff
@@ -808,6 +808,8 @@ export default {
 			this.newRecipients = []
 			this.requestMdn = false
 			this.appendSignature = true
+			this.savingDraft = undefined
+			this.saveDraftDebounced = debounce(10 * 1000, this.saveDraft)
```
ChristophWurst, line 812: the function doesn't need to be reset

jancborchardt:
@GretaD do you have updated screenshots for easier design review? cc @nimishavijay

GretaD:
> @GretaD do you have updated screenshots for easier design review? cc @nimishavijay

The UI hasnt changed since the last discussion we had. I will update the screenshots when the PR is 100% done, but technically shouldnt have anything to review design wise.

GretaD:
newmessagemodal

src/components/NewMessageDetail.vue
```diff
@@ -132,24 +132,7 @@ export default {
...
-				}
-
-				return {
-					accountId,
-					to: this.stringToRecipients(this.$route.query.to),
-					cc: this.stringToRecipients(this.$route.query.cc),
-					subject: this.$route.query.subject || '',
-					body: this.$route.query.body ? detect(this.$route.query.body) : html(''),
-				}
+				throw new Error('new message details can only be for replies and forwards')
```
ChristophWurst, line 135: `Error` conflicts with the imported error component

what you want to throw is a new instance of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

throwing a new constructor instance causes an error

> [Vue warn]: Error in render: "TypeError: _Error__WEBPACK_IMPORTED_MODULE_7__.default is not a constructor"

Import the error component under a different name. then the global javascript `Error` will be unhidden and work as expected.

You can trigger the error if you open `/apps/mail/box/priority/thread/new`

ChristophWurst (changes requested):
There are still lots of remnant usages of the old *new message* route. Search for `threadId: 'new',`. They all have to be fixed so that the modal opens when we previously triggered this routing.

`mailto:` links and so on are currently broken

src/components/NewMessageModal.vue
```diff
@@ -0,0 +1,199 @@
...
+		max-width: 80%;
+	}
+}
+::v-deep .modal-container {
+	width: 80%;
+	min-height: 60%;
+}
+::v-deep .modal-wrapper .modal-container {
+	overflow-y: auto;
+	overflow-x: auto;
```
st3iny, line 197:
```suggestion
	overflow-y: auto !important;
	overflow-x: auto !important;
```

Modal isn't scrollable otherwise.

nimishavijay:
@GretaD would you be posting screenshots when you are done? The previous one you posted looks almost finished already so there wouldn't too much drastic feedback, but just for a final check :)

GretaD:
> @GretaD would you be posting screenshots when you are done? The previous one you posted looks almost finished already so there wouldn't too much drastic feedback, but just for a final check :)

Yes i will, of course.

GretaD:
- [ ] Add padding on top of from. (adding more on left and right brings troubles to the borders of from, to cc@nimisha)
- [ ] Remove "Write another message"
- [ ] Add icons on discarded and sent feedback

discard
discarddraft
msgsent

nimishavijay:
Looks good! Only 3 minor points about icons
- can we make it bigger (around 44*44 px)
- can we use a lighter colour like `--color-text-maxcontrast`
- for 'Message sent' can we use `icon-checkmark` instead

Ideally it looks similar to this
image

GretaD:
sent
dis
discarddraft

ChristophWurst:
Bildschirmfoto von 2021-12-01 11-26-28

minor design quirk: the modal scrolls when it might not have to

ChristophWurst (approved):
Works :shipit:

src/components/AppSettingsMenu.vue
```diff
@@ -175,6 +175,7 @@ export default {
 			}
 		},
 		registerProtocolHandler() {
+			console.debug('MAILTO IS WORKING')
```
ChristophWurst, line 178: oh nonononono

ChristophWurst (changes requested):

ChristophWurst (changes requested):
mailto handling is dead and doesn't open any modal

ChristophWurst:
- [ ] BUG still tons of remants of the old `new` thread routing that showed the old composer

```
$ rg "'new'" src/
src/components/Envelope.vue
259:                                            threadId: 'new',

src/components/RecipientBubble.vue
175:                                    threadId: 'new',

src/components/Mailbox.vue
152:                                                            threadId: 'new',
244:                                                            threadId: 'new',

src/components/MenuEnvelope.vue
254:                                    threadId: 'new',

src/components/MailboxThread.vue
150:                            this.$route.params.threadId === 'new'

src/views/Home.vue
78:                                     threadId: 'new',
97:                                     threadId: 'new',
```

ChristophWurst:
- [ ] BUG mailto doesn't work

ChristophWurst:
- [ ] BUG recipient bubble *Reply* doesn't work

GretaD:
> mailto handling is dead and doesn't open any modal

this is broken on master for me

src/components/AppSettingsMenu.vue
GretaD: im sure you mean nananana

ChristophWurst:
> this is broken on master for me

Works fine for me with the examples at https://www.rapidtables.com/web/html/mailto.html. `/apps/mail/box/priority/thread/new?to=name%40rapidtables.com` opens the composer and puts in that email address as recipient.

GretaD:
> * [x]  BUG still tons of remants of the old `new` thread routing that showed the old composer
> 
We removed the ones that are directly connect to _New message_, _edit as new_ and _forward as attachment._ The ones left are for _reply_ and _forward_. The rest will be removed when we fix the reply and forward in a separate PR.

st3iny (approved):
Tested and works :tada:

PR #5532, opened 2021-09-12
Handle itinerary times with and without timezones

vkrause:
Event reservations and calendar event generation will need similar fixes, I'll look into that once my local setup allows me to test those properly.


ChristophWurst (approved):
Sure

kesselb:
@vkrause thanks :+1: 

Could you have a look at the eslint warnings for the code style?

vkrause:
> Could you have a look at the eslint warnings for the code style?

Should be taken care of now, sorry about that.

PR #5533, opened 2021-09-13
Use the same IMAP query for the HTML body part as for attachments

vkrause:
This not only removes some code but also makes this actually work, fixing
itinerary extraction from HTML bodies.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -433,21 +433,7 @@ class MessageMapper {
 			// No HTML part
 			return null;
 		}
-		$partsQuery = new Horde_Imap_Client_Fetch_Query();
-		$partsQuery->fullText();
-		foreach ($structure->partIterator() as $structurePart) {
-			/** @var Horde_Mime_Part $structurePart */
-			$partsQuery->bodyPart($structurePart->getMimeId(), [
-				'decode' => true,
-				'peek' => true,
-			]);
-			$partsQuery->bodyPartSize($structurePart->getMimeId());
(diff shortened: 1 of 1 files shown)


kesselb:
Thanks :+1: 

Is there a way to test it?

vkrause:
I've added a sample mail showing this here: https://invent.kde.org/pim/kitinerary/-/blob/master/autotests/extractordata/synthetic/flight-booking.eml - with this change you now see the flight information. Common real-world examples would be mails from Eventbrite, Meetup, booking.com and various airlines.

kesselb:
LGTM :+1: 

https://invent.kde.org/pim/kitinerary/-/blob/master/autotests/extractordata/synthetic/flight-booking.eml does work on master and this branch. I guess that's the expected behavior? 

A recent hotel confirmation from booking.com does not work (on master and this branch) but the source of the email does not contain any schema org markup.

kesselb (approved):

vkrause:
> LGTM +1
> 
> https://invent.kde.org/pim/kitinerary/-/blob/master/autotests/extractordata/synthetic/flight-booking.eml does work on master and this branch. I guess that's the expected behavior?

Interesting, that one didn't work for me on master (against Cyrus IMAP), but it does work now.

> A recent hotel confirmation from booking.com does not work (on master and this branch) but the source of the email does not contain any schema org markup.

booking.com doesn't always include schema.org markup unfortunately indeed.

Also, hotel reservations aren't displayed yet, so you only see that in the JSON response so far (I'll look into adding that after fixing the issues I see with the existing reservation types).

PR #5535, opened 2021-09-14
Show attachments inline

### LINKED ISSUE #4181: Attachments should show inline below the message with more suitable icon
When reading a mail in the current version of the mail app, the only clue that there is a file attached is the chain icon below the message.
This easily slips away from the attention, as it has no context or label.

First - I'd argue that the Icon is not the best choice, as in the rest of Nextcloud's GUI it is used to indicate a folder being shared (Files App - File List), or stands for the URL behind a share (Share Sidebar). The well known Icon for an Attachment is the Paper-Clip. I'd propose to change to that.
An Icon could also be useful at the Avatar Circle of the mail in the content frame and maybe even attached to the avatar in the Mail List.

In addition to that the Filename should be visible directly below the Message with a proper Icon and a possibility to save the file to files app or HDD. Images should have a preview.

This was also discussed during last contributor week in Mail design review.

GretaD:
fixes #4181 
- [x] Fix 3dot menu position
- [x] Fix menu elements icons
- [x] The popup its not positioned properly in some long mails, but hard to reproduce

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -38,23 +38,13 @@
 			:body="message.body"
 			:signature="message.signature"
 			:message="message" />
-		<Popover v-if="message.attachments[0]" class="attachment-popover">
-			<Actions slot="trigger">
-				<ActionButton icon="icon-public icon-attachment">
-					{{ t('mail', 'Attachments') }}
-				</ActionButton>
-			</Actions>
-			<MessageAttachments v-close-popover="true" :attachments="message.attachments" :envelope="envelope" />
-		</Popover>
+		<MessageAttachments :attachments="message.attachments" :envelope="envelope" />
 		<div id="reply-composer" />
 	</div>
 </template>
 
 <script>
-import Actions from '@nextcloud/vue/dist/Components/Actions'
-import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
 import { generateUrl } from '@nextcloud/router'
-import Popover from '@nextcloud/vue/dist/Components/Popover'
 
 import { html, plain } from '../util/text'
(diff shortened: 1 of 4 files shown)


GretaD:
it looks like this in different scenarios

attachemnt2
2.

calendarattachment
3.

attachinline1

GretaD:
@st3iny i need your help with the popup thingy. Maybe im overlooking, but i cannot find a way to fix it and i dont know why it happens. So when the message its long till the end of the screen the popup of the attachments jumps up and its not positioned properly. Please have a look if you have any idea.

ChristophWurst:
> @st3iny i need your help with the popup thingy. Maybe im overlooking, but i cannot find a way to fix it and i dont know why it happens. So when the message its long till the end of the screen the popup of the attachments jumps up and its not positioned properly. Please have a look if you have any idea.

this reminds me of the jumping actions menu (in the navigation?). is that related @marcoambrosini?

st3iny:
@GretaD I was able to reproduce the bug and fixed it. I'll push a new commit here.

st3iny (commented):
The `Download Attachment` button is not working. I guess it's missing a `@click` handler.

GretaD:
> The `Download Attachment` button is not working. I guess it's missing a `@click` handler.

done, can you have a general look and accept so we can merge it :)

st3iny (approved):
I tested locally and it worked fine.

I would like the individual attachments to have a bit more horizontal space. Their names are truncated frequently even if there is enough vertical space. We can fix that in a followup.

ChristophWurst:
> We can fix that in a followup.

@GretaD please create a ticket for this :)

PR #5557, opened 2021-09-20
Remove unread counter from trash mailbox

### LINKED ISSUE #5272: Unread counter on Trash is unnecessarily distracting
There is an unread counter for every folder, which does make sense everywhere – except Trash. If you trashed a message unread, then you don’t want to be bothered by it.

The solution for that would be to not show the unread counter on the Trash folder. Gmail does it like that as well:

Nextcloud Mail with unread trash counter | Gmail without unread trash counter
-|-
mail navigation|image

GretaD:
fixes #5272

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -111,7 +111,7 @@
 				{{ t('mail', 'Delete mailbox') }}
 			</ActionButton>
 		</template>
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

jancborchardt:
Btw thank youuu @GretaD! :)

PR #5558, opened 2021-09-20
Improve right actions on the right section of thread

### LINKED ISSUE #5528: Move the right elements in an envelope more to the right
There's a lot of space on the left that can me used
rightmeny

GretaD:
fixes #5528 

after
rightafter

before
rightmeny


ChristophWurst (approved):
:rocket:

PR #5566, opened 2021-09-23
Add vertical whitespace between accounts

### LINKED ISSUE #5271: Whitespace missing between accounts
There used to be whitespace above the account names for visual separation in the navigation, now there is none and it’s all just one list:
Mail nav before with whitespace | Mail nav now, no vertical whitespace
-|-
mail nav before | mail navigation

GretaD:
after
afterwhitespace
before
beforewhitespace
fixes #5271

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -234,4 +234,7 @@ to {
 		transform: rotate(-360deg);
 	}
 }
+.app-navigation-spacer {
+	order: 0 !important;
+}
 </style>


ChristophWurst (approved):
~~Would have expected that there was already code/css for this and we just broke the selector on a nc/vue upgrade or similar~~ reread the ticket

jancborchardt:
Thank you @GretaD! :)

PR #5567, opened 2021-09-23
Calculate height for iframe by taggedElement

### LINKED ISSUE #5527: HTML message is clipped
### Expected behavior

The sender is in the white list to display images.

The mail should be in a taller box, with a scrollbar if needed

### Actual behavior

image

Mail can't be read.

### Mail app

**Mail app version:** 	1.10.3

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...): gmx.com

### Server configuration

**Operating system**: (e.g. Debian 8)  Raspbian (base Debian 10)

**Web server:** nginx

**Database:** Mariadb

**PHP version:** 7.3

**Nextcloud Version:** 21.0.4

#### Client configuration
**Browser:** (e.g. Firefox 48) Firefox 92, Opera 79

**Operating system:** (e.g. Arch Linux) Linux Mint 20.2

kesselb:
Fix #5527  

We disabled CSSTidy on purpose because some CSS rules were rewritten in a weird way: https://github.com/nextcloud/mail/pull/5424

For some reason the iframe height calculation does not work anymore. document.body.offsetHeight is 150.

taggedElement as heightCalculationMethod finds the bottom of the lowest element with a data-iframe-height attribute.

As reference the documentation for heightcalculationmethod: https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/parent_page/options.md#heightcalculationmethod

bodyScroll for example also works but the recommendation is to use lowestElement or taggedElement when the default does not work. Please let me know if one of you know a better dummy element for the data-iframe-height attribute.

### DIFF
--- a/lib/Http/HtmlResponse.php
+++ b/lib/Http/HtmlResponse.php
@@ -83,6 +83,6 @@ class HtmlResponse extends Response {
 			return $this->content;
 		}
 
-		return '<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><script nonce="' . $this->nonce . '" src="' . $this->scriptUrl . '"></script></head><body>' . $this->content . '</body></html>';
+		return '<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><script nonce="' . $this->nonce . '" src="' . $this->scriptUrl . '"></script></head><body>' . $this->content . '<div data-iframe-height></div></body></html>';
 	}
 }
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -75,7 +75,7 @@ export default {
 		scout.on('afterprint', this.onAfterPrint)
 	},
 	mounted() {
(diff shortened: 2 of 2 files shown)


lib/Http/HtmlResponse.php
```diff
@@ -83,6 +83,6 @@ public function render(): string {
 			return $this->content;
 		}
 
-		return '<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><script nonce="' . $this->nonce . '" src="' . $this->scriptUrl . '"></script></head><body>' . $this->content . '</body></html>';
+		return '<!DOCTYPE html><html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><script nonce="' . $this->nonce . '" src="' . $this->scriptUrl . '"></script></head><body>' . $this->content . '<script type="text/plain" data-iframe-height></script></body></html>';
```
ChristophWurst, line 86: could we use a `div` instead of a `script` tag?

st3iny (approved):
Beautiful

PR #5571, opened 2021-09-24
Consider itinerary timezones when importing into the calendar

vkrause:
Signed-off-by: Volker Krause <vkrause@kde.org>

PR #5572, opened 2021-09-24
Use itineraryDateTime in EventReservation

kesselb:
```
{
    "@type": "QDateTime",
    "@value": "2021-10-14T18:00:00+02:00",
    "timezone": "Europe/Berlin"
}
```

To show the proper start and end date when data.reservationFor is something like above.

### DIFF
--- a/src/components/itinerary/EventReservation.vue
+++ b/src/components/itinerary/EventReservation.vue
@@ -74,13 +74,13 @@ export default {
 			if (!('startDate' in this.data.reservationFor)) {
 				return
 			}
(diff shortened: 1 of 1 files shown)


kesselb:
fyi @vkrause

vkrause:
> fyi @vkrause

Looks good to me. The ical part eventually will also need the timezone support from https://github.com/nextcloud/mail/pull/5571 I think, but this is definitely an improvement as-is already.

ChristophWurst (approved):

Issue #1037, opened 2018-08-16
`mb_convert_encoding` does not support `windows-1250`.


ChristophWurst:
### Expected behavior
The app should be able to read popular encodings.

### Actual behavior
Some encodings (e.g. `windows-1250` and Arabic) cause warnings/errors. Apparently that's not supported by the build-in function: https://secure.php.net/manual/de/mbstring.supported-encodings.php. However, some clients (Outlook, I suppose) use this encoding.

### Mail app

**Mail app version:**
0.9.0

### Server configuration

**PHP version:**
7.0

**Nextcloud Version:**
14 Beta 4

Ref https://sentry.io/share/issue/303496811cbc42899c531fbe51a3e37f/.

whirm:
I'm not seeing .ics files attached to e-mails (works just fine with other webmails) and this error shows every time I load an e-mail:

```
mb_convert_encoding(): Illegal character encoding specified at /var/www/html/custom_apps/mail/lib/Model/IMAPMessage.php#553
```

I'm guessing it's the same issue? 

I'm running 16.0.1 and Mail 0.15.1 on PHP 7.3.6

ChristophWurst:
> I'm guessing it's the same issue?

would you be able to attach debugger or add some logs to verify?

HLFH:
For windows-1256 / cp1256 support, could be interesting to look into this code: https://github.com/khaled-alshamaa/ar-php/blob/master/src/Arabic.php

At the section * Convert Arabic string into glyph joining in UTF-8 hexadecimals stream*.

HLFH:
ìconv` does not support the following 27 charsets:

> UTF7-IMAP
ISO-2022-JP-MS
CP51932
SJIS-mac (alias: MacJapanese)
SJIS-Mobile#DOCOMO (alias: SJIS-DOCOMO)
SJIS-Mobile#KDDI (alias: SJIS-KDDI)
SJIS-Mobile#SOFTBANK (alias: SJIS-SOFTBANK)
UTF-8-Mobile#DOCOMO (alias: UTF-8-DOCOMO)
UTF-8-Mobile#KDDI-A
UTF-8-Mobile#KDDI-B (alias: UTF-8-KDDI)
UTF-8-Mobile#SOFTBANK (alias: UTF-8-SOFTBANK)
ISO-2022-JP-MOBILE#KDDI (alias: ISO-2022-JP-KDDI)
JIS
JIS-ms
CP50220
CP50220raw
CP50221
[...]

PR #5593, opened 2021-09-28
Support windows-1250 and windows-1256 charsets

HLFH:
Fixes https://github.com/nextcloud/mail/issues/1037.

The workaround commit https://github.com/nextcloud/mail/commit/e76f9ce4cd273a105430e048397c50fb8f7472aa linked to PR https://github.com/nextcloud/mail/pull/309 removed the support of the windows-1250 and windows-1256 charsets.

If I understand correctly, in my new code, the try/catch block will not be thrown in prior versions (prior to PHP 8), which makes the try/catch block a harmless no-op in older PHP versions.

I am not sure how much you didn't like iconv (https://github.com/nextcloud/mail/pull/309) but AFAIK, this commit is fixing the new issue we encountered.
Happy to update the commit if you have any advice on that.

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -84,8 +84,8 @@ class Message extends Entity implements JsonSerializable {
 		'flagged',
 		'seen',
 		'forwarded',
-		'junk',
-		'notjunk',
+		'$junk',
+		'$notjunk',
 		'mdnsent',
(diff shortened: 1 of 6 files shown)


HLFH:
cc @ChristophWurst

brzd:
What's the status with this?

PR #5600, opened 2021-09-30
Prevent irrelevant envelopes entering the unified mailboxes

ChristophWurst:
Unified mailboxes are a combination of individual mailboxes. For the
first unified page we have to fetch each individual page, then combine
them and take the 20 most recent messages.

When fetching the individual pages we already added all envelopes to the
unified mailboxes through the addEnvelope mutation. That means you
already had 40 messages in the unified mailbox before they were combined
and only the latest messages were picked.

If you have two accounts with somewhat similarly old messages this bug
doesn't always show. However, if you have one inbox with many recent
messages and another inbox with only significantly older messages then
this bug showed in a way that also the newest messages of the older
inbox were part of the 40 first messages. Now after the fix you see the
20 new messages from the newer inbox, then 20 more of the same inbox and
only when the message timestamps of the second inbox align with the first
inbox you will start to see messages from the second inbox.

In order to test this, try to following
1) Do you see more than 20 messages on the first page of *All inboxes*? Compare master vs this branch
2) As you scroll down in *All inboxes*, does the last message before the next page is loaded jump when that page comes in or does it stay at its position? Compare master vs this branch


GretaD:
All looks good, the message doesnt jump, but after i scroll twice, it loads but not new messages are shown. This warning is shown:
`actions.js:537 envelope list is not defined, can't fetch next envelopes 248`

ChristophWurst:
@GretaD another observation: when you scroll down, you can see that the logic to load the next page is already triggered. Then you scroll again and it triggers another time. Then things somehow go wrong … :duck:

ChristophWurst:
@GretaD I think you can test again with your special accounts. The messages should now be in the expected order without the gap :crossed_fingers:

ChristophWurst:
The account colors are super misleading btw. Just because everything is shown in pink doesn't mean it's from one of the accounts. I fixed that separately via https://github.com/nextcloud/mail/pull/5601 :)

PR #5674, opened 2021-10-28
move threads not messages in drag-and-drop

StCyr:
Currently, the drag-and-drop function moves messages rather than threads. 

It results that when you move a thread, drag-and-drop actualy only moves the message corresponding to the selected envelope. This results in some visual glitch: first the envelope is removed from the envelope list resulting in the whole thread disappearing from the envelope list, then the envelope list is refreshed and displays back the thread by showing the envelope of another thread's message, one that hasn't been moved yet.

Moving threads rather than messages in the drag-and-drop function makes sense both from a logical perspective (ie: when the user moves a thread's envelope he actually wants to move the whole corresponding thread and not just the corresponding message) and a ux perspective (ie: avoiding this visual glitch).

This PR implements moving threads rather than messages in drag-and-drop.

<----------- EDIT START ------------>
It seems the limitations previously mentioned hereunder are not real:

1.Mail doesn't navigate after moving a thread => This is true but is already the case today, without this PR
2.After the thread is moved, the messages appear as individual messages, not as a single thread, in the destination mailbox => Cannot reproduce it. Was probably just caused by me forgeting to make dev-setup.

So, for me the PR is ready for merge
<----------- EDIT END ----------->

Some limitations of the PR currently are:
1. Mail doesn't navigate after moving a thread
2. After the thread is moved, the messages appear as individual messages, not as a single thread, in the destination mailbox => We probably need to re-compute thread information on the destination mailbox after the move.

Fixes https://github.com/nextcloud/mail/issues/8310

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -3,7 +3,7 @@
 		v-draggable-envelope="{
 			accountId: data.accountId ? data.accountId : mailbox.accountId,
 			mailboxId: data.mailboxId,
-			envelopeId: data.databaseId,
+			databaseId: data.databaseId,
 			draggableLabel,
 			selectedEnvelopes,
 			isDraggable,
--- a/src/directives/drag-and-drop/draggable-envelope/draggable-envelope.js
+++ b/src/directives/drag-and-drop/draggable-envelope/draggable-envelope.js
@@ -45,14 +45,14 @@ export class DraggableEnvelope {
 				envelopes.push({
 					accountId,
 					mailboxId,
-					envelopeId: envelope.databaseId,
+					databaseId: envelope.databaseId,
 					draggableLabel: `${envelope.subject} (${envelope.from[0].label})`,
 				})
 			})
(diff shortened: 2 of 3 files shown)


StCyr:
The issue addressed by this PR is still there today

ChristophWurst:
> After the thread is moved, the messages appear as individual messages, not as a single thread, in the destination mailbox => Cannot reproduce it. Was probably just caused by me forgeting to make dev-setup

I can image this happens when the messages are moved in the local cache but not via IMAP. As soon as a sync is triggered, the threads are rebuilt and messages should show up combined.

GretaD:
> > After the thread is moved, the messages appear as individual messages, not as a single thread, in the destination mailbox => Cannot reproduce it. Was probably just caused by me forgeting to make dev-setup
> 
I can reproduce it with submailbox. So, when i move a thread on a mailbox the thread is moved as a one. When i move it on the  submailbox its splited.

PR #5711, opened 2021-11-08
Correctly center envelop header content using flexbox

CarlSchwan:
This change was done using the webinspector and I didn't test it locally
    since I encountered ssl errors when setting the mail app up and I was
    expecting this patch to be a 5min job :)
    
Please test locally when reviewing

Before: 
image

After:
image

Signed-off-by: Carl Schwan <carl@carlschwan.eu>

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -295,6 +295,7 @@ export default {
 		justify-content: flex-end;
 		margin-left: 10px;
 		height: 44px;
+		align-self: center;
 
 		.app-content-list-item-menu {
 			margin-left: 4px;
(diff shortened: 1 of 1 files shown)


GretaD:
> this doesnt work for me, even though it seems correct from the code. I will check the code again soon and try some other classes

i take this back, this works just fine on thread, we need to fix it for envelope as well, i tested this on envelope

GretaD:
@CarlSchwan after implementation of tags, this change looks like this: 
threadcenter


first, it breaks the imporant and favorite icon, also, the avatar shouldnt be in the center of the content but inline with address and reply button. Are you ok, if i push a fixup?

GretaD:
on the latest master the content is properly aligned.

PR #5719, opened 2021-11-11
Update npm and node engines versions

nextcloud-command:
Automated update of the npm and node engines versions


ChristophWurst (changes requested):
The app also works with current versions of node/npm, we don't have to be restrictive

PR #5730, opened 2021-11-18
Fix composer regression

### LINKED ISSUE #5708: New message draft - content isn't editable
### Expected behavior
I can type a new message

### Actual behavior
Nothing

### Mail app

Looks like the `<p data -placeholder>` tag is missing, although I am unsure whether that's CKEditor or the Composer.vue component that's removed this tag.

A reply to an old message works fine.

GretaD:
On plain text, we cannot write a message. This is caused by this pr: https://github.com/nextcloud/mail/pull/5516

the reason why Kesselb removed it was that ParagraphPlugin was already loaded by another component, this doesn't happen anymore. I add the plugin, and the packages are updated fine.

fixes #5708


ChristophWurst (approved):
Fine by me if `plugincollection-plugin-name-conflict` doesn't show and both HTML and plain text editors work

GretaD:
Yes, lets have a final test from Anna and we can merge it :)

miaulalala (approved):
LGTM

image

PR #5744, opened 2021-11-23
Fix iframe after printing

### LINKED ISSUE #5667: Print email
### Feature Request

Unfortunately, the printouts are terrible.

#### Summary

When printing an e-mail (CTRL-P) the result is almost always unusable. Text is cut off, formatting is not really present.
So I wish me in the 3-point menu the entry "print", which then provides a usable result.

GretaD:
fixes #5667


ChristophWurst (approved):

st3iny (approved):
Tested and works. Please squash fixups and we are ready to go.

PR #5748, opened 2021-11-25
Fixes use mailbox.accountId as fallback when accountId isn't defined in Envelope.vue

StCyr:
reported in https://github.com/nextcloud/mail/pull/4288#issuecomment-972831255

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -249,7 +249,7 @@ export default {
 			return this.data.flags.draft
 		},
 		account() {
-			const accountId = this.data.accountId ?? this.data.accountId
+			const accountId = this.data.accountId
 			return this.$store.getters.getAccount(accountId)
 		},
 		link() {


src/components/Envelope.vue
```diff
@@ -249,7 +249,7 @@ export default {
 			return this.data.flags.draft
 		},
 		account() {
-			const accountId = this.data.accountId ?? this.data.accountId
```
ChristophWurst, line 252: I don't think the accountId is missing still on envelopes. That was only the case for messages loaded in a paged result. After https://github.com/nextcloud/mail/pull/5601 this shouldn't happen anymore

ChristophWurst (changes requested):

src/components/Envelope.vue
ChristophWurst: that is true. it should just be `const accountId = this.data.accountId` :)

GretaD (approved):
lets get this in

ChristophWurst (approved):

PR #5769, opened 2021-12-02
Improve envelope alignment

GretaD:
before
alignenvbefore
after
alignenvelope

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -1,156 +1,154 @@
 <template>
-	<ListItem
-		v-draggable-envelope="{
-			accountId: data.accountId ? data.accountId : mailbox.accountId,
-			mailboxId: data.mailboxId,
-			envelopeId: data.databaseId,
-			draggableLabel,
-			selectedEnvelopes,
-		}"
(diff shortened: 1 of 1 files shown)


marcoambrosini (commented):
there still seems to be more space below in the screenshot, or am I hallucinating? :)

GretaD:
> there still seems to be more space below in the screenshot, or am I hallucinating? :)

there's definitively some more space below. Either this was the result when we worked on it and we didnt notice on the small screen, or i missed something when i rebased. Is there anything missing on css? Can you tell?

src/components/Envelope.vue
```diff
@@ -547,6 +547,9 @@ list-item-style.draft .app-content-list-item-line-two {
 .app-content-list-item-star.icon-starred {
 	display: block;
 }
+.app-content-list-item-icon {
+	display: flex;
```
marcoambrosini, line 551: Maybe this?

```suggestion
	display: flex;
	align-items: center;
```

GretaD: this doesnt do anything, its already added on the wrapper

PR #5772, opened 2021-12-02
Drop Nextcloud 20 support

### LINKED ISSUE #5758: Drop Nextcloud 20 support
Ref https://github.com/nextcloud/groupware/issues/7

GretaD:
fixes #5758

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


ChristophWurst (changes requested):
Nextcloud 21 requries php7.3. So we can drop 7.2

GretaD:
> Nextcloud 21 requries php7.3. So we can drop 7.2

you mean from lint and ci, because on these 2 files it starts from 7.3

ChristophWurst:
yes :)

ChristophWurst (approved):

PR #5777, opened 2021-12-03
Fix trusted domain regression

### LINKED ISSUE #5773: Cannot add domain to trusted senders
### Expected behavior
Click on "Always show images from domain.tld", add domain.tld to trusted senders
Re-open mail from that domain, images should be shown

### Actual behavior
Images are blocked again and I can choose again what I want to allow

### Mail app

**Mail app version:**  1.11.0

**Mailserver or service:** alfahosting

### Server configuration

**Operating system**: official Docker

**Web server:** 

**Database:** Postgres 13

**PHP version:** 

**Nextcloud Version:** 23.0

#### Client configuration
**Browser:** latest Firefox

**Operating system:** openSUSE Tumbleweed

Heres an error from the browser console after clicking the button:

```
TypeError: this.getDomain is not a function
    onShowBlockedContentForDomain MessageHTMLBody.vue:137
    VueJS 4
    onClick ActionButton.js:43
    VueJS 3
vue.runtime.esm.js:1897:12
    VueJS 8
    onClick ActionButton.js:43
    VueJS 3
```

GretaD:
fixes #5773
regression of https://github.com/nextcloud/mail/commit/29dddbfc38b63226f0e3af5fb5612ae3782098aa#diff-ca2bac59d6118981505ea366fa8cf3b743e0bdcdd146289a4f5974c33b82162dL129

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -134,7 +134,7 @@ export default {
 		async onShowBlockedContentForDomain() {
 			this.displayIframe()
 			// TODO: there might be more than one @ in an email address
-			await trustSender(this.getDomain(), 'domain', true)
+			await trustSender(this.domain, 'domain', true)
 		},
 	},
 }


miaulalala (approved):

src/components/MessageHTMLBody.vue
```diff
@@ -134,7 +134,7 @@ export default {
 		async onShowBlockedContentForDomain() {
 			this.displayIframe()
 			// TODO: there might be more than one @ in an email address
-			await trustSender(this.getDomain(), 'domain', true)
+			await trustSender(this.domain(), 'domain', true)
```
ChristophWurst, line 137: does this work? :thinking:

GretaD: yes, no error is shown and the action is saved correctly. _getDomain_ was removed here: https://github.com/nextcloud/mail/commit/29dddbfc38b63226f0e3af5fb5612ae3782098aa and _domain_ was added but not replaced in this method.

ChristophWurst: https://vuejs.org/v2/guide/computed.html#Computed-Caching-vs-Methods while it is allowed and correct to use computed properties as method I would drop the `()` like we do in other code

ChristophWurst (approved):

PR #5779, opened 2021-12-03
Add anti abuse detection

ChristophWurst:
Some admins like to see alerts when mail accounts are used in suspicious ways, e.g. to disable accounts that send spam.

* Trigger alert for number of recipients of a single message
* Trigger alert for number of messages per time period (15m, 1h, 1d)

Todo
- [x] Implement
- [x] Test
- [x] Document

### DIFF
--- a/lib/Service/AntiAbuseService.php
+++ b/lib/Service/AntiAbuseService.php
@@ -1,0 +1,151 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2021 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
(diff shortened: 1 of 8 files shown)


lib/Service/AntiAbuseService.php
```diff
@@ -0,0 +1,153 @@
...
+			return;
+		}
+
+		$now = $this->timeFactory->getTime();
+
+		// Build blocks of periods per period size
+		$periodStart = ((int)($now / $period)) * $period;
+		$cacheKey = implode('_', ['counter', $id, $periodStart]);
+		$cache->add($cacheKey, 0);
+		$counter = $cache->inc($cacheKey);
```
ChristophWurst, line 142: The increment should possibly be the number of recipients. Because if you send one message to five people then you cause five messages :thinking:

miaulalala (changes requested):
Looks good but I agree with you self review, the number of recipients should be considered in the counter

miaulalala (approved):
LGTM

PR #5798, opened 2021-12-09
Fix initial sync stopping prematurely and endlessly

ChristophWurst:
The initial sync is broken up into chunks. These chunks are estimated
because we only know the lowest and highest UID and there are likely
gaps in this range. One of the last chunks might query a range that has
more messages than the allowed chunk size. We slice it down. This means
the actually fetched max UID is lower than the theoretical max UID for this
chunk.

If the stop condition is to stop at the theoretical max UID then we
occasionally stop too early. Where we actually have to stop is the
actually fetched highest UID.

Extracted from https://github.com/nextcloud/mail/pull/5721 to be easier to review.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -40,9 +40,13 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use Psr\Log\LoggerInterface;
 use function array_filter;
 use function array_map;
+use function count;
 use function in_array;
 use function iterator_to_array;
+use function max;
+use function min;
 use function reset;
+use function sprintf;
 
 class MessageMapper {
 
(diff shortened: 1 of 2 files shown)


miaulalala (approved):

lib/IMAP/MessageMapper.php
```diff
@@ -107,11 +111,11 @@ public function findAll(Horde_Imap_Client_Socket $client,
 			]
 		);
 		/** @var int $min */
-		$min = $metaResults['min'];
+		$min = (int) $metaResults['min'];
 		/** @var int $max */
-		$max = $metaResults['max'];
+		$max = (int) $metaResults['max'];
```
ChristophWurst, line 116: Bildschirmfoto von 2021-12-09 12-24-36
Bildschirmfoto von 2021-12-09 12-24-13

:facepalm:

PR #5807, opened 2021-12-09
New database table for reset

snevas:
As version 1.11.0 introduces a new table, we need to include that table in the cleanup procedure as well in the docs.


miaulalala:
Can you please add some context to this PR in the description? Thanks :)

ChristophWurst (approved):

PR #5828, opened 2021-12-10
Check for empty string in message id and reply to

miaulalala:


ChristophWurst (approved):
Eliminates botched initial sync/thread data from my dev env

Production instances might need a cleared cache.

ChristophWurst (approved):

PR #5840, opened 2021-12-13
Remove command to fix threads and make it a repair step instead.

miaulalala:


lib/Migration/RepairMailTheads.php
```diff
@@ -39,25 +38,18 @@ class RepairMailTheads extends Command {
...
 
-	protected function execute(InputInterface $input, OutputInterface $output): int {
+	public function run(IOutput $output) {
 		$count = $this->mapper->resetInReplyTo();
 		$this->logger->info('Repairing Mail Threading, ' . $count . ' messages updated');
-		$output->writeln('');
-		$output->writeln('Repaired threads, ' . $count . ' messages updated');
-		return 0;
+		$output->info(sprintf('Repaired threads, %s messages updated', $count));
+		$output->finishProgress();
```
ChristophWurst, line 53: only used if you call `\OCP\Migration\IOutput::startProgress` before

ChristophWurst (approved):

PR #5846, opened 2021-12-14
Enforce issue templates, add links

ChristophWurst:
Makes it harder for people to open blank issues.


st3iny (approved):
Hard to test but LGTM!

PR #5856, opened 2021-12-15
Catch unknown method during upgrade repair step

ChristophWurst:
It's our friend the upgrade error.

How it blows up
* The Nextcloud process doing the upgrade loads all application code. Mail app code as well
* The migration is done
* Post migration steps are run
  + The new repair step is loaded, but it uses a class that existed before. That class isn't loaded another time. Kaboom.

The same trick is also found in `\OCA\Mail\Migration\AddMissingMessageIds::run`.

### DIFF
--- a/lib/Migration/RepairMailTheads.php
+++ b/lib/Migration/RepairMailTheads.php
@@ -27,6 +27,7 @@ use OCA\Mail\Db\MessageMapper;
 use OCP\Migration\IOutput;
 use OCP\Migration\IRepairStep;
 use Psr\Log\LoggerInterface;
+use function method_exists;
 
 class RepairMailTheads implements IRepairStep {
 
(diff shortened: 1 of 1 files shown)


lib/Migration/RepairMailTheads.php
```diff
@@ -47,6 +48,16 @@ public function getName(): string {
 	}
 
 	public function run(IOutput $output): void {
+		/**
+		 * During the upgrade to v1.11.3 and later the old version of the
+		 * mapper is loaded before this new repair step is performed. Hence even after
+		 * the program code got replaced, the class doesn't have the new method. We
+		 */
+		if (!method_exists($this->mapper, 'resetInReplyTo')) {
+			$output->warning('New Mail code hasn\'t been loaded yes, skipping message ID repair. Please run `occ maintenance:repair` after the upgrade.');
```
nickvergessen, line 57:
```suggestion
			$output->warning('New Mail code hasn\'t been loaded yet, skipping message ID repair. Please run `occ maintenance:repair` after the upgrade.');
```

ChristophWurst: yet you are right

miaulalala (approved):

lib/Migration/RepairMailTheads.php
```diff
@@ -47,6 +48,16 @@ public function getName(): string {
...
+		 * During the upgrade to v1.11.3 and later the old version of the
+		 * mapper is loaded before this new repair step is performed. Hence even after
+		 * the program code got replaced, the class doesn't have the new method. We
+		 */
+		if (!method_exists($this->mapper, 'resetInReplyTo')) {
+			$output->warning('New Mail code hasn\'t been loaded yet, skipping message ID repair. Please run `occ maintenance:repair` after the upgrade.');
+			return;
+		}
+
 		$count = $this->mapper->resetInReplyTo();
```
raimund-schluessler, line 61: @ChristophWurst @miaulalala Sorry, I am late here. But couldn't this method be located directly in the migration step, e.g. in the `postSchemaChange`? The method is only used in the repair step and you wouldn't need to run `occ maintenance:repair` afterwards manually.

ChristophWurst: that would be possible as a workaround.

personally I would rather push for finally fixing repair steps than creating even more workaround everywhere

PR #5864, opened 2021-12-16
Fix column for account name too narrow

ChristophWurst:
The old length was 64. There were instances where this was too little.
Now we use 255.

Fixes

```
An exception occurred while executing a query: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'name' at row 1
```


st3iny (commented):
Please also adjust the length of `oc_mail_aliases.name` beause it's also initialized to the display name of the user on provisioning.

https://github.com/nextcloud/mail/blob/132259b821e2dae4624c8e1851953f2cb61a2305/lib/Service/Provisioning/Manager.php#L146

miaulalala (approved):

GretaD (approved):

PR #5883, opened 2021-12-20
Set BCC for email recipients query

miaulalala:
Previously, BCC was not set at all, and reset `from` again.


ChristophWurst (approved):
:facepalm:

PR #5890, opened 2021-12-21
[stable1.11] Fix mark as spam

GretaD:
backport of #5843


st3iny (approved):
Diffs match. Tested and works.

PR #5935, opened 2022-01-10
Add indices on various columns to increase performance

miaulalala:
This should speed up different queries, namely the threading query, considerably.

### DIFF
--- a/lib/Migration/Version1116Date20220110153658.php
+++ b/lib/Migration/Version1116Date20220110153658.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version1116Date20220110153658 extends SimpleMigrationStep {
+
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('mail_messages');
+		if (!$table->hasIndex('mail_messages_id_flags')) {
+			$table->addIndex(['mailbox_id', 'flag_important', 'flag_deleted', 'flag_seen'], 'mail_messages_id_flags');
+		}
+		if (!$table->hasIndex('mail_messages_id_flags2')) {
+			$table->addIndex(['mailbox_id', 'flag_deleted', 'flag_flagged'], 'mail_messages_id_flags2');
+		}
+
+		// Postgres doesn't allow for shortened indices, so let's skip the last index.
+		if ($schema->getDatabasePlatform() instanceof PostgreSQL94Platform) {
+			return $schema;
+		}
+
+		if (!$table->hasIndex('mail_msg_thrd_root_snt_idx')) {
+			$table->addIndex(['mailbox_id', 'thread_root_id', 'sent_at'], 'mail_msg_thrd_root_snt_idx', [], ['lengths' => [null, 64, null]]);
+		}
+		return $schema;
+	}
+}
--- a/tests/psalm-baseline.xml
+++ b/tests/psalm-baseline.xml
@@ -66,11 +66,6 @@
       <code>Command</code>
     </UndefinedClass>
   </file>
-  <file src="lib/Command/RepairMailTheads.php">
-    <UndefinedClass occurrences="1">
-      <code>Command</code>
-    </UndefinedClass>
-  </file>
   <file src="lib/Command/SyncAccount.php">
     <UndefinedClass occurrences="1">
       <code>Command</code>
@@ -88,7 +83,7 @@
   </file>
   <file src="lib/Db/AliasMapper.php">
     <ImplicitToStringCast occurrences="1">
-      <code>$qb2-&gt;createNamedParameter($chunk, IQueryBuilder::PARAM_INT_ARRAY)</code>
+      <code>$qb2-&gt;createParameter('ids')</code>
     </ImplicitToStringCast>
   </file>
   <file src="lib/Db/CollectedAddressMapper.php">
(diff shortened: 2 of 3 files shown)


lib/Migration/Version1116Date20220110153658.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 * @throws \Doctrine\DBAL\Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('mail_messages')) {
```
nickvergessen, line 27: this is a prerequirement, so should not be checked but instead break hard.

miaulalala: whaddaya mean? Exception?

GretaD (approved):

tests/psalm-baseline.xml
```diff
@@ -290,6 +285,18 @@
       <code>$qb-&gt;createNamedParameter($accountIds, IQueryBuilder::PARAM_INT_ARRAY)</code>
     </ImplicitToStringCast>
   </file>
+  <file src="lib/Migration/Version1116Date20220110153658.php">
+    <UndefinedClass occurrences="1">
+      <code>PostgreSQL94Platform</code>
+    </UndefinedClass>
+    <UndefinedDocblockClass occurrences="2">
+		<code>\Doctrine\DBAL\Exception</code>
+		<code>\Doctrine\DBAL\Platforms\AbstractPlatform</code>
```
miaulalala, line 294: @nickvergessen

lib/Migration/Version1116Date20220110153658.php
```diff
@@ -0,0 +1,44 @@
...
+			return null;
+		}
+
+		$table = $schema->getTable('mail_messages');
+		if (!$table->hasIndex('mail_messages_id_flags')) {
+			$table->addIndex(['mailbox_id', 'flag_important', 'flag_deleted', 'flag_seen'], 'mail_messages_id_flags');
+		}
+		if (!$table->hasIndex('mail_messages_id_flags2')) {
+			$table->addIndex(['mailbox_id', 'flag_deleted', 'flag_flagged'], 'mail_messages_id_flags2');
+		}
```
nickvergessen, line 38: those two indexes can be added just fine in postgres, only the one with the length should be skipped.

miaulalala: Right :see_no_evil:

nickvergessen (approved):

miaulalala:
@nickvergessen I can't get Static Analysis to accept the migration. Any chance you could take a look?

lib/Migration/Version1116Date20220110153658.php
```diff
@@ -0,0 +1,46 @@
...
+
+		// Postgres doesn't allow for shortened indices, so let's skip the last index.
+		if ($schema->getDatabasePlatform() instanceof PostgreSQL94Platform) {
```
ChristophWurst, line 37: open a tech dept ticket. this code is prone to explode with any changes to the implementation of the db abstraction. we had the same issue when IDbConnection (or was it the builder?) was assumed to directly implement the Doctrine classes. But since we added a new layer of abstraction that was never the case -> kaboom.

suggestion for a fix: propose an OCP API change with a new accessor that gives us the db type, e.g. as string or number that you can compare to a constant.

lib/Migration/Version1116Date20220110153658.php
```diff
@@ -0,0 +1,46 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('mail_messages');
+		if (!$table->hasIndex('mail_messages_id_flags')) {
+			$table->addIndex(['mailbox_id', 'flag_important', 'flag_deleted', 'flag_seen'], 'mail_messages_id_flags');
+		}
+		if (!$table->hasIndex('mail_messages_id_flags2')) {
+			$table->addIndex(['mailbox_id', 'flag_deleted', 'flag_flagged'], 'mail_messages_id_flags2');
```
ChristophWurst, line 33: same potential performance problems here. maybe this should all be optional indices?

lib/Migration/Version1116Date20220110153658.php
miaulalala: API change on the ISchemaWrapper class?

lib/Migration/Version1116Date20220110153658.php
```diff
@@ -0,0 +1,46 @@
...
+	 * @return null|ISchemaWrapper
+	 * @throws Exception
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('mail_messages');
+		if (!$table->hasIndex('mail_messages_id_flags')) {
+			$table->addIndex(['mailbox_id', 'flag_important', 'flag_deleted', 'flag_seen'], 'mail_messages_id_flags');
```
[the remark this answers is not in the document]
miaulalala: I haven't but these indices are from c.nc.com. They must have been applied last year when we did an evaluation. It was suggested to a customer to use them and it worked fine for them and seemed to help with performance.

If you tell me how to, I will do it :)

lib/Migration/Version1116Date20220110153658.php
miaulalala: Sure thing, we could add an occ command for it for example?

lib/Migration/Version1116Date20220110153658.php
miaulalala: https://github.com/nextcloud/server/issues/30877

lib/Migration/Version1116Date20220110153658.php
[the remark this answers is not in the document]
ChristophWurst: I don't have an exact threshold but any migration should run in a matter of seconds. The reason is that many installation use the web updater and there is a default timeout of 30s. Assuming that this app might not be the only component that needs time for a migration, we only have a fraction of those 30s to finish the job.

Anything that takes longer should be optional, if possible.

lib/Migration/Version1116Date20220110153658.php
ChristophWurst: yes, albeit it would be hard for admins to ever discover those as we do not have app-provided setup checks. https://github.com/nextcloud/server/pull/25471 is probably the more integrated approach.

miaulalala:
This is scheduled to be applied with the cache reset.

miaulalala:
Redo as discussed

PR #5943, opened 2022-01-12
Add outbox mailbox

GretaD:
this is not a solution yet, just stuff I'm testing to learn more about how to approach outbox


GretaD:
closing as Richard took over

PR #5946, opened 2022-01-14
Add more logging to IMAP to DB conversion

miaulalala:


ChristophWurst:
what's the status here? finish or discard?

PR #5976, opened 2022-01-24
Fix message account sync error if uidCandidates list is empty

pravussum:
The mail app in my NC installation fails the initital account sync with the following error.
`"Message":"Undefined array key -1 at /var/www/html/custom_apps/mail/lib/IMAP/MessageMapper.php#186",`

The reason is that the `$uidCandidates` array is empty after the array_filter step, although the `$fetchResult` isn't.
To prevent an endless loop I added the shortcut and return an empty result.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -142,6 +142,15 @@ class MessageMapper {
 			$max,
 			$lower + $estimatedPageSize
 		);
+		if ($lower > $upper) {
+			$this->logger->debug("Range for findAll did not find any (not already known) messages and all messages of mailbox $mailbox have been fetched.");
+			return [
+				'messages' => [],
+				'all' => true,
+				'total' => 0,
+			];
+		}
+
 		$this->logger->debug("Built range for findAll: min=$min max=$max total=$total totalRange=$totalRange estimatedPageSize=$estimatedPageSize lower=$lower upper=$upper highestKnownUid=$highestKnownUid");
 
 		$query = new Horde_Imap_Client_Fetch_Query();
--- a/tests/Unit/IMAP/MessageMapperTest.php
+++ b/tests/Unit/IMAP/MessageMapperTest.php
@@ -322,6 +322,44 @@ class MessageMapperTest extends TestCase {
 		self::assertFalse($result['all']);
 	}
 
+	public function testFindAllNoUidCandidates(): void {
+		/** @var Horde_Imap_Client_Socket|MockObject $client */
+		$client = $this->createMock(Horde_Imap_Client_Socket::class);
+		$mailbox = 'inbox';
+		$client->expects(self::once())
+			->method('search')
+			->with(
+				$mailbox,
+				null,
+				[
+					'results' => [
+						Horde_Imap_Client::SEARCH_RESULTS_MIN,
+						Horde_Imap_Client::SEARCH_RESULTS_MAX,
+						Horde_Imap_Client::SEARCH_RESULTS_COUNT,
+					]
+				]
+			)
+			->willReturn([
+				'min' => 1,
+				'max' => 35791,
+				'count' => 32122,
+			]);
+		$query = new Horde_Imap_Client_Fetch_Query();
+		$query->uid();
+		$client->expects(self::never())
+			->method('fetch');
+
+		$result = $this->mapper->findAll(
+			$client,
+			$mailbox,
+			5000,
+			99999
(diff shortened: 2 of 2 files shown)


lib/IMAP/MessageMapper.php
```diff
@@ -178,6 +178,20 @@ function (int $uid) use ($highestKnownUid) {
...
 		);
+		if(count($uidCandidates) === 0) {
+			if($upper < $max) {
+				$this->logger->debug("Range for findAll did not find any (not already known) messages. Trying again with a succeeding range... $upper < $max");
+				return $this->findAll($client, $mailbox, $maxResults, $upper);
+			} else {
+				$this->logger->debug("Range for findAll did not find any (not already known) messages and all messages of mailbox $mailbox have been fetched.");
+				return [
+					'messages' => [],
+					'all' => true,
```
miaulalala, line 189: If this happens in the middle of a sync, the sync would stop here and never fetch candidates with higher UIDs due to all being true.

pravussum: But this is in the branch where $upper already reached or exceeded $max, which I understand is the MAX UID returned by the IMAP server for this query - thus the last range of this query is empty.

The first branch on the other hand continues with the next range.

miaulalala: Right, ignore my comment :sweat_smile:

I'm currently working on the same piece of code, so it might take some time to merge this just to be sure that there are no unintended side effects :slightly_smiling_face:

ChristophWurst:
> Thanks for opening your first pull request in this repository! v

This is a very impressive change and insight for a first pull request. Great work, @pravussum!

lib/IMAP/MessageMapper.php
```diff
@@ -178,6 +178,20 @@ function (int $uid) use ($highestKnownUid) {
 				return $uid > $highestKnownUid;
 			}
 		);
+		if(count($uidCandidates) === 0) {
```
ChristophWurst, line 181: I would like to understand how we can possibly end up here.

1) The mailbox is not empty, or we'll already exit at the `$total === 0` check
2) There was likely an earlier sync chunk that did not reach `$upper === $max`

Does that mean that the *highest known UID* info is possibly wrong when we fetch if from the database? Do you use any kind of fancy database cluster or is it a simple single node database?

pravussum: I reproduced the error by removing my fix and enhancing the log output in the error case (`count($uidCandidates) === 0`) - this is the output: 

`Built range for findAll: min=1 max=35791 total=32122 totalRange=35791 estimatedPageSize=5572 lower=35792 upper=35791 highestKnownUid=35791 count(fetchResult)=1`

And another one:

`Built range for findAll: min=1083 max=1583 total=8 totalRange=501 estimatedPageSize=313126 lower=1584 upper=1583 highestKnownUid=1583 count(fetchResult)=1`

So in both cases the `$upper` is less than `$lower`, while the IMAP query still returns 1 result. I think the IMAP server is smart enough to correct the query by flipping the upper and lower bound. The resulting query can only return the UID equal to highestKnownUid (the other candidate is already > max). The single result then gets filtered out by the code starting in line 168, leaving the `$uidCandidates` empty. 

As an alternative to my solution one could also shortcut the whole thing - in line 145 if `$upper < $lower`.

influjensbahr:
Hi, just wanted to add that I have this error too! Let me know if there is anything I can do to help find the source. I also opened a thread on the nextcloud forum, where I'll post any updates as well: https://help.nextcloud.com/t/mail-loads-forever-floods-with-error-warnings/133474

influjensbahr:
I pulled these changes as a test. The logs are quiet now, but messages still don't show. Let me know if I can provide any info or try something out

pravussum:
The first issue looks really unrelated. Guess its trying to get the favicon of one of the senders domains and fails to do so.

The second one looks more like invalid credentials or wrong message account settings (auth mode correct?).

miaulalala:
@influjensbahr can you open a ticket for the first one and log out and back in again to see if this fixes the issue for the second one?

lib/IMAP/MessageMapper.php
ChristophWurst: I think I finally understood it. That only took a couple of weeks :speak_no_evil: 

In any case, I've added another unit test for your real world example and adjusted the code for it. I liked your idea with the shortcut as it saves one IMAP roundtrip :v:

Issue #5747, opened 2021-11-25
Argument 2 passed to OCA\\Mail\\Service\\MailManager::getThread() must be of the type string, null given


BlackLight:
### Expected behavior

Email bodies should be shown when clicking on an email on the list.

### Actual behavior

The body is not shown and `/index.php/apps/mail/api/messages/<id>/thread` returns `500` with the following payload:

```json
{
  "status": "error",
  "message": "Argument 2 passed to OCA\\Mail\\Service\\MailManager::getThread() must be of the type string, null given, called in /srv/http/nextcloud/apps/mail/lib/Controller/MessagesController.php on line 310",
  "data": {
    "debug": true,
    "type": "Exception",
    "message": "Argument 2 passed to OCA\\Mail\\Service\\MailManager::getThread() must be of the type string, null given, called in /srv/http/nextcloud/apps/mail/lib/Controller/MessagesController.php on line 310",
    "code": 0,
    "trace": [
      {
        "file": "/srv/http/nextcloud/lib/private/AppFramework/App.php",
        "line": 157,
        "function": "dispatch",
        "class": "OC\\AppFramework\\Http\\Dispatcher"
      },
      {
        "file": "/srv/http/nextcloud/lib/private/Route/Router.php",
        "line": 303,
        "function": "main",
        "class": "OC\\AppFramework\\App"
      },
      {
        "file": "/srv/http/nextcloud/lib/base.php",
        "line": 993,
        "function": "match",
        "class": "OC\\Route\\Router"
      },
      {
        "file": "/srv/http/nextcloud/index.php",
        "line": 37,
        "function": "handleRequest",
        "class": "OC"
      }
    ],
    "previous": {
      "type": "TypeError",
      "message": "Argument 2 passed to OCA\\Mail\\Service\\MailManager::getThread() must be of the type string, null given, called in /srv/http/nextcloud/apps/mail/lib/Controller/MessagesController.php on line 310",
      "code": 0,
      "trace": [
        {
          "file": "/srv/http/nextcloud/apps/mail/lib/Controller/MessagesController.php",
          "line": 310,
          "function": "getThread",
          "class": "OCA\\Mail\\Service\\MailManager"
        },
        {
          "file": "/srv/http/nextcloud/lib/private/AppFramework/Http/Dispatcher.php",
          "line": 218,
          "function": "getThread",
          "class": "OCA\\Mail\\Controller\\MessagesController"
        },
        {
[...]

PR #5981, opened 2022-01-25
Handle empty thread root id

### LINKED ISSUE #5774: Unexpected behaviour in Mail Rendering
### Expected behavior
This is how the mail looks like in Rainloop (inside Nextcloud) and Fairemail on my Android Phone
image
image

### Actual behavior
This is the mail in NC Mail with the "Termin buchen" Link not visible, but if you know where it is, you can click it at least 🗡️ 
image

### Mail app
**Mail app version:** 1.11
**Mailserver or service:** IMAP

### Server configuration
**Operating system**: Debian 10
**Web server:** Apache
**Database:** MariaDB
**PHP version:** 7.3
**Nextcloud Version:** 22.2.2
#### Client configuration
**Browser:** FF 94 & Edge 96.0
**Operating system:** Windows 10

HTML Code of the Mail can be shared if requested.

miaulalala:
Fixes #5774

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -134,7 +134,6 @@ class MessagesController extends Controller {
 								ITrustedSenderService $trustedSenderService,
 								IMailTransmission $mailTransmission) {
 		parent::__construct($appName, $request);
-
 		$this->accountService = $accountService;
 		$this->mailManager = $mailManager;
 		$this->mailSearch = $mailSearch;
@@ -326,6 +325,10 @@ class MessagesController extends Controller {
 			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
 		} catch (DoesNotExistException $e) {
 			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		if (empty($message->getThreadRootId())) {
+			return new JSONResponse([], Http::STATUS_NOT_FOUND);
 		}
 
 		return new JSONResponse($this->mailManager->getThread($account, $message->getThreadRootId()));
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -174,7 +174,9 @@ class Message extends Entity implements JsonSerializable {
 	}
 
 	public function setThreadRootId(?string $messageId): void {
-		$this->setMessageIdFieldIfNotEmpty('threadRootId', $messageId);
+		$threadRootId = (!empty($messageId)) ? '<' . rtrim(ltrim($messageId, '<'), '>') . '>' : $this->getMessageId();
+		$parsed = new Horde_Mail_Rfc822_Identification($threadRootId);
+		$this->setter('threadRootId', [$parsed->ids[0] ?? $this->getMessageId()]);
 	}
 
 	private function setMessageIdFieldIfNotEmpty(string $field, ?string $id): void {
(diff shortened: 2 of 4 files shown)


miaulalala:
I don't know if the frontend will handle the new 404 response correctly. Waiting for an email to test this.

lib/Controller/MessagesController.php
```diff
@@ -307,6 +306,10 @@ public function getThread(int $id): JSONResponse {
 			return new JSONResponse([], Http::STATUS_FORBIDDEN);
 		}
 
+		if (empty($message->getThreadRootId())) {
```
ChristophWurst, line 309: in what scenario do we allow the thread root ID to be null? shouldn't it always have a value, worst case it's own message ID, no?

miaulalala: Is it?  didn't know. Seems like the issue is cropping up somewhere else then since the related ticket is having issues with exactly that.

ChristophWurst: yeah. because if it's NULL, what thread does it belong to?

ref https://github.com/nextcloud/mail/issues/5747 for one of the more recent incidents where the ID was erroneously null.

miaulalala: Right, programatically, we're currently supporting null values for the threadRootId: https://github.com/nextcloud/mail/blob/8aabd4dede1ff2f7021b018e72a9aee35595e268/lib/Db/Message.php#L175

We could change that line to use the `messageId` in case the `threadRootId` is empty? Would prevent future issues like this.

And as for fixing the old ones, a repair step that encapsulates https://github.com/nextcloud/mail/issues/5747#issuecomment-979483114 ?

ChristophWurst: Let's do that!

miaulalala:
@ChristophWurst is the repair step still needed for the empty thread root id's? Something's vaguely stick in my brain that we do update empty thread root ids already when we build the threads?

lib/Db/Message.php
```diff
@@ -174,7 +174,9 @@ public function setInReplyTo(?string $inReplyTo): void {
 	}
 
 	public function setThreadRootId(?string $messageId): void {
-		$this->setMessageIdFieldIfNotEmpty('threadRootId', $messageId);
+		$threadRootId = (!empty($messageId)) ? '<' . rtrim(ltrim($messageId, '<'), '>') . '>' : $this->getMessageId();
+		$parsed = new Horde_Mail_Rfc822_Identification($threadRootId);
+		$this->threadRootId = ($parsed->ids[0] ?? null);
```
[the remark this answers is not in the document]
miaulalala: yeah but it needs to run through the horde parser to properly format the id

ChristophWurst, line 179:
```suggestion
		if (!empty($parsed->ids[0]) {
			$this->setter('threadRootId', $parsed->ids[0]);
		}
```

1) Does *not* set the thread root id to `null`
2) Works with the entity change detection. If you assign `$this->threadRootId` directly you bypass the call to `\OCP\AppFramework\Db\Entity::markFieldUpdated`.

ChristophWurst (approved):

ChristophWurst:
> @ChristophWurst is the repair step still needed for the empty thread root id's? Something's vaguely stick in my brain that we do update empty thread root ids already when we build the threads?

I would say the repair step still makes sense because all messages with invalid headers sync'ed before this fix can potentially have a null thread root id.

ChristophWurst:
I would leave https://github.com/nextcloud/mail/blob/main/lib/Migration/AddMissingMessageIds.php in. If all messages are correct then the job doesn't touch anything.

PR #6006, opened 2022-02-01
Remove the autocomplete text after selecting recipient

GretaD:
Right now, when you write a recipient on the to/cc/bcc field, the autocomplete opens, then you select a recipient but the text you already added doesnt go away. That means that you should delete the text before adding some other recipient. This pr fixes that


miaulalala (changes requested):
Auto suggestion selection works great, but I have a few suggestions:

1) there is no way to remove the a reciever that is "in the middle" of the others because deletion only works via backspace. So I would have to delete every entry after the one I want to delete and then reenter them:

Say I want to delete Anna Banana:
image

I have to delete miau_lalala first.

An (x) that removes that entry would be super.

2) Removing an entry with backspace is too easy. A single backspace will remove the entire recipient - possibly make that a double backspace, or make it so the entry is edited instead of removed completely - do so it isn't that easy to remove a recipient accidentally.

3) Google Chrome Autofill keeps messing up the view and takes the focus from the dropdown:

image

`autocomplete="false"` on the input might work if it doesn't break the our autocompletion.

ChristophWurst:
Legit suggestions @miaulalala. @GretaD if this is achievable with a config flag on the Vue multiselect then go for it. Otherwise I would suggest to offload this to a ticket and address only the non-closing autcomplete in this PR.

GretaD:
> Legit suggestions @miaulalala. @GretaD if this is achievable with a config flag on the Vue multiselect then go for it. Otherwise I would suggest to offload this to a ticket and address only the non-closing autcomplete in this PR.

Yes, i think the best way would be to change the component and to use EntityPicker instead of changing the Multiselect and adding stuff that EntityPicker already has. 
So, @miaulalala please approve and your suggestions will be included when we do improve the recipients as a whole

miaulalala (approved):

PR #6028, opened 2022-02-10
Fix missing account aliases for new accounts

### LINKED ISSUE #5907: No `account.aliases` for new accounts
## How to reproduce

1. Delete any existing account
2. Add a new account
3. Open the new message composer

### Expected behavior

Composer works.

### Actual behavior

Error is thrown because `account.aliases` is undefind in `Compser.vue`. After a page reload the error goes away. This indicates that we set `aliases` from the deseralized data on page load, but do not set a default value for accounts added after the page load.

This is bad for the first use experience. Hence high priority.

### Mail app

**Mail app version:** v1.12

JuliaKirschenheuter:
fixes #5907

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -43,6 +43,7 @@ export default new Vuex.Store({
 				accountId: UNIFIED_ACCOUNT_ID,
 				isUnified: true,
 				mailboxes: [PRIORITY_INBOX_ID, UNIFIED_INBOX_ID],
+				aliases: [],
 				collapsed: false,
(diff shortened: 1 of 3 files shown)


GretaD (approved):

ChristophWurst:
front end tests are complaining. you need to adjust some tests. you can run them with `npm run test`

ChristophWurst (approved):
Please squash the commits into one so that code and test code changes are in one atomic commit

Issue #5885, opened 2021-12-20
Create database structure and REST API for an outbox


ChristophWurst:
1. `mail_local_mailbox`
  * `type`: int, allows us to distinguish between outboxes and local draft mailboxes
  * `account_id`: foreign key to `mail_accounts`
  * `send_at`: nullable timestamp
  * `text`: the email as RFC822 message (or separate columns for to, cc, bcc, subject, contents and attachments as another 1:n table)
2. QB mapper for the table(s)
3. A service for handles access the outbox
4. REST resource at `/apps/mail/api/outbox`
  * `POST /apps/mail/api/outbox` to enqueue a message
  * `GET /apps/mail/api/outbox` returns the outbox contents
  * `GET /apps/mail/api/outbox/{id}` returns the outbox message by id
  * `DELETE /apps/mail/api/outbox/{id}` deletes a specific message
  * `POST /apps/mail/api/outbox/{id}` sends a specific message
  * `PUT /apps/mail/api/outbox/{id}` updates the message

Whenever a Mail account is deleted, we'll have to wipe the outbox messages that are still to be sent. Otherwise the sending will run into a null account.

miaulalala:
Should I also add a type here, as we've discussed in December re: Draft Handling?

ChristophWurst:
Done via https://github.com/nextcloud/mail/pull/6052 and https://github.com/nextcloud/mail/pull/6030

PR #6030, opened 2022-02-11
Add database schema and layer for local mailboxes

miaulalala:
This adds the first part of https://github.com/nextcloud/mail/issues/5885.

### DIFF
--- a/lib/Db/LocalAttachmentMapper.php
+++ b/lib/Db/LocalAttachmentMapper.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Luc Calaresu <dev@calaresu.com>
@@ -26,6 +28,7 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\QBMapper;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
+use Throwable;
 
 /**
  * @template-extends QBMapper<LocalAttachment>
@@ -40,10 +43,33 @@ class LocalAttachmentMapper extends QBMapper {
 	}
 
 	/**
+	 * @return LocalAttachment[]
+	 */
+	public function findByLocalMessageId(int $localMessageId): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->eq('local_message_id', $qb->createNamedParameter($localMessageId, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
+			);
+		return $this->findEntities($qb);
+	}
+
+	/**
+	 * @return LocalAttachment[]
+	 */
+	public function findByLocalMessageIds(array $localMessageIds): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->in('local_message_id', $qb->createNamedParameter($localMessageIds, IQueryBuilder::PARAM_INT_ARRAY), IQueryBuilder::PARAM_INT_ARRAY)
+			);
+		return $this->findEntities($qb);
+	}
+
+	/**
 	 * @throws DoesNotExistException
-	 *
-	 * @param string $userId
-	 * @param int $id
 	 */
 	public function find(string $userId, int $id): LocalAttachment {
 		$qb = $this->db->getQueryBuilder();
@@ -54,5 +80,19 @@ class LocalAttachmentMapper extends QBMapper {
 			->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT));
 
 		return $this->findEntity($query);
+	}
+
+	public function deleteForLocalMailbox(int $localMessageId): void {
+		$this->db->beginTransaction();
+		try {
+			$qb = $this->db->getQueryBuilder();
+			$qb->delete($this->getTableName())
+				->where($qb->expr()->eq('local_message_id', $qb->createNamedParameter($localMessageId), IQueryBuilder::PARAM_INT));
+			$qb->execute();
+			$this->db->commit();
+		} catch (Throwable $e) {
+			$this->db->rollBack();
+			throw $e;
+		}
 	}
 }
--- a/lib/Db/Recipient.php
+++ b/lib/Db/Recipient.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Mail App
+ *
+ * @copyright 2022 Anna Larch <anna.larch@gmx.net>
+ *
+ * @author Anna Larch <anna.larch@gmx.net>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Mail\Db;
+
+use JsonSerializable;
+use OCP\AppFramework\Db\Entity;
+
+/**
+ * @method int|null getMessageId()
+ * @method void setMessageId(int $messageId)
+ * @method int|null getLocalMessageId()
+ * @method void setLocalMessageId(int $localMessageId)
+ * @method int getType()
+ * @method void setType(int $type)
+ * @method string getLabel()
+ * @method void setLabel(string $label)
+ * @method string getEmail()
+ * @method void setEmail(string $email)
+ */
+class Recipient extends Entity implements JsonSerializable {
+	public const TYPE_FROM = 0;
+	public const TYPE_TO = 1;
+	public const TYPE_CC = 2;
+	public const TYPE_BCC = 3;
+
+	/** @var int|null */
+	protected $messageId;
+
+	/**
+	 * @var int
+	 * @psalm-var self::TYPE_*
+	 */
+	protected $type;
+
+	/** @var int|null */
+	protected $localMessageId;
+
+	/** @var string */
+	protected $label;
+
+	/** @var string */
+	protected $email;
+
+	public function __construct() {
+		$this->addType('messageId', 'integer');
+		$this->addType('localMessageId', 'integer');
+		$this->addType('type', 'integer');
+		$this->addType('mailboxType', 'integer');
+	}
+
+	public function jsonSerialize(): array {
+		return [
+			'id' => $this->getId(),
+			'messageId' => $this->getMessageId(),
+			'localMessageId' => $this->getLocalMessageId(),
+			'type' => $this->getType(),
+			'label' => $this->getLabel(),
+			'email' => $this->getEmail()
+		];
+	}
+}
--- a/lib/Db/RecipientMapper.php
+++ b/lib/Db/RecipientMapper.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Mail App
+ *
+ * @copyright 2022 Anna Larch <anna.larch@gmx.net>
+ *
+ * @author Anna Larch <anna.larch@gmx.net>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Mail\Db;
+
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+
+/**
+ * @template-extends QBMapper<Recipient>
+ */
+class RecipientMapper extends QBMapper {
+	public function __construct(IDBConnection $db) {
+		parent::__construct($db, 'mail_recipients');
+	}
+
+	/**
+	 * @returns Recipient[]
+	 */
+	public function findByLocalMessageId(int $localMessageId): array {
+		$qb = $this->db->getQueryBuilder();
+
+		$query = $qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->eq('local_message_id', $qb->createNamedParameter($localMessageId, IQueryBuilder::PARAM_INT))
+			);
+
+		return $this->findEntities($query);
+	}
+
+	/**
+	 *  @return Recipient[]
+	 */
+	public function findByLocalMessageIds(array $localMessageIds): array {
+		$qb = $this->db->getQueryBuilder();
+
+		$query = $qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->in('local_message_id', $qb->createNamedParameter($localMessageIds, IQueryBuilder::PARAM_INT_ARRAY), IQueryBuilder::PARAM_INT_ARRAY)
+			);
+
+		return $this->findEntities($query);
+	}
+
+	public function deleteForLocalMailbox(int $localMessageId): void {
+		$qb = $this->db->getQueryBuilder();
+
+		$qb->delete($this->getTableName())
+			->where(
+				$qb->expr()->eq('local_message_id', $qb->createNamedParameter($localMessageId, IQueryBuilder::PARAM_INT))
+			);
+		$qb->execute();
+	}
+
+	/**
+	 * @param int $localMessageId
+	 * @param Recipient[] $recipients
+	 * @param int $type
+	 * @psalm-param Recipient::TYPE_* $type
+	 */
+	public function saveRecipients(int $localMessageId, array $recipients, int $type): void {
+		if (empty($recipients)) {
+			return;
+		}
+
+		$qb = $this->db->getQueryBuilder();
+		$qb->insert($this->getTableName());
+		$qb->setValue('local_message_id', $qb->createParameter('local_message_id'));
+		$qb->setValue('type', $qb->createParameter('type'));
+		$qb->setValue('label', $qb->createParameter('label'));
+		$qb->setValue('email', $qb->createParameter('email'));
+
+		foreach ($recipients as $recipient) {
+			$qb->setParameter('local_message_id', $localMessageId, IQueryBuilder::PARAM_INT);
+			$qb->setParameter('type', $type, IQueryBuilder::PARAM_INT);
+			$qb->setParameter('label', $recipient->getLabel() ?? $recipient->getEmail(), IQueryBuilder::PARAM_STR);
+			$qb->setParameter('email', $recipient->getEmail(), IQueryBuilder::PARAM_STR);
+			$qb->execute();
+		}
+	}
+}
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -12,7 +12,7 @@
 - **🙈 We’re not reinventing the wheel!** Based on the great [Horde](https://horde.org) libraries.
 - **📬 Want to host your own mail server?** We don’t have to reimplement this as you could set up [Mail-in-a-Box](https://mailinabox.email)!
 	]]></description>
-	<version>1.12.0-alpha.2</version>
+	<version>1.12.0-alpha.3</version>
 	<licence>agpl</licence>
 	<author>Greta Doçi</author>
 	<author homepage="https://github.com/nextcloud/groupware">Nextcloud Groupware Team</author>
--- a/lib/Db/LocalAttachment.php
+++ b/lib/Db/LocalAttachment.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Luc Calaresu <dev@calaresu.com>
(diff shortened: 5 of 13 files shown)


ChristophWurst:
Could you please also include the code for the DB layer here? Like the entity and mapper?

lib/Db/LocalAttachmentMapper.php
```diff
@@ -23,36 +23,135 @@
...
 
 /**
  * @template-extends QBMapper<LocalAttachment>
  */
 class LocalAttachmentMapper extends QBMapper {
+	/** @var ITimeFactory */
+	private $timeFactory;
+
+	/**
+	 * @throws DoesNotExistException|Exception|MultipleObjectsReturnedException
```
ChristophWurst, line 41: only the first is to be expected, right?

lib/Db/LocalAttachmentMapper.php
```diff
@@ -23,36 +23,135 @@
...
+			->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT));
+
+		return $this->findEntity($query);
+	}
 
 	/**
 	 * @param IDBConnection $db
```
ChristophWurst, line 57: drop this outdated parameter documentation

lib/Db/LocalAttachmentMapper.php
```diff
@@ -23,36 +23,135 @@
...
+		$qb2 = $this->db->getQueryBuilder();
+		$qb2->insert('mail_lcl_mbx_attchmts')
+			->setValue('local_message_id', $qb2->createNamedParameter($localMessageId))
+			->setValue('attachment_id', $qb2->createNamedParameter($attachmentId));
+		$result = $qb2->execute();
+		$result->closeCursor();
+	}
+
+	/**
+	 * @throws Exception
```
ChristophWurst, line 147: do you expect this to be thrown?

lib/Db/LocalAttachmentMapper.php
```diff
@@ -23,36 +23,135 @@
...
+			return $row['attachment_id'];
+		}, $result->fetchAll());
+
+		$result->closeCursor();
+
+		if (empty($attachmentIds)) {
+			return [];
+		}
+
+		return array_map(function ($attachmentId) use ($userId) {
```
miaulalala, line 84: make this a join too

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,127 @@
...
+								LocalAttachmentMapper $attachmentMapper,
+								RecipientMapper $recipientMapper) {
+		parent::__construct($db, 'mail_local_mailbox');
+		$this->recipientMapper = $recipientMapper;
+		$this->attachmentMapper = $attachmentMapper;
+	}
+
+	/**
+	 * @param string $userId
+	 * @return LocalMailboxMessage[]
```
ChristophWurst, line 55: it does not

lib/Db/LocalAttachmentMapper.php
```diff
@@ -23,36 +23,135 @@
...
+		$result->closeCursor();
+		return $rows;
+	}
 
-		return $this->findEntity($query);
+	/**
+	 * @throws Exception
+	 */
+	public function deleteForLocalMailbox(int $localMessageId, string $userId): void {
+		$attachments = $this->findForLocalMailbox($localMessageId, $userId);
```
miaulalala, line 109: change to subquery

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,127 @@
...
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->in('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
+			);
+		return $this->findEntity($qb);
+	}
+
+	public function getRelatedData(int $id, string $userId): array {
```
ChristophWurst, line 86: will we need this as public method or is related data loaded automatically in `find`/`getAllForUser` and similar?

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,127 @@
...
+			->where(
+				$qb->expr()->eq('a.user_id', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR), IQueryBuilder::PARAM_STR),
+				$qb->expr()->eq('m.local_message_id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
+			);
+		$rows = $qb->execute();
+		$result = $rows->fetchAll();
+		$rows->closeCursor();
+		$related = [];
+		$related['attachments'] = $result;
+		$related['recipients'] = $this->recipientMapper->findRecipients($id, Recipient::MAILBOX_TYPE_OUTBOX);
```
ChristophWurst, line 100: let's avoid these stringly typed arrays and attache the data to the LocalMailbox instances instead

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,127 @@
...
+		array_map(function ($recipient) use ($message) {
+			$this->recipientMapper->createForLocalMailbox($message->getId(), $recipient['label'] ?? $recipient['email'], $recipient['email']);
+		}, array_merge($to, $cc, $bcc));
+		foreach ($attachmentIds as $attachmentId) {
+			$this->attachmentMapper->linkAttachmentToMessage($message->getId(), $attachmentId);
+		}
+	}
+
+	/**
+	 * @throws DBException
```
ChristophWurst, line 120: is it expected to be thrown? should callers react to it?

lib/Db/Recipient.php
```diff
@@ -0,0 +1,75 @@
...
+class Recipient extends Entity implements JsonSerializable {
+	public const MAILBOX_TYPE_INBOX = 0;
+	public const MAILBOX_TYPE_OUTBOX = 1;
+
+	public const TYPE_FROM = 0;
+	public const TYPE_TO = 1;
+	public const TYPE_CC = 2;
+	public const TYPE_BCC = 3;
+
+	protected $messageId;
```
ChristophWurst, line 53: type?

lib/Db/RecipientMapper.php
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+namespace OCA\Mail\Db;
+
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+
+/**
+ * @template-extends QBMapper<Message>
```
ChristophWurst, line 34: nope

lib/Db/RecipientMapper.php
```diff
@@ -0,0 +1,81 @@
...
+ */
+class RecipientMapper extends QBMapper {
+	public function __construct(IDBConnection $db) {
+		parent::__construct($db, 'mail_recipients');
+	}
+
+	/**
+	 * @returns Recipient[]
+	 * @throws \OCP\DB\Exception
```
ChristophWurst, line 43: too broad

tests/Integration/Db/LocalMailboxMessageTest.php
```diff
@@ -0,0 +1,66 @@
...
+class LocalMailboxMessageTest extends TestCase {
+
+	/** @var ITimeFactory|MockObject  */
+	private $timeFactory;
+
+	protected function setUp(): void {
+		$this->timeFactory = $this->createMock(ITimeFactory::class);
+	}
+
+	public function testObject(): void {
```
ChristophWurst, line 42:
```suggestion
	public function testSettersGetters(): void {
```

ChristophWurst (changes requested):

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,127 @@
...
+	 */
+	public function saveWithRelatedData(LocalMailboxMessage $message, array $to, array $cc, array $bcc, array $attachmentIds = []): void {
+		$this->insert($message);
+		// @TODO
+		// not that easy - this needs work to actually create the right recipient type in the DB!
+		array_map(function ($recipient) use ($message) {
+			$this->recipientMapper->createForLocalMailbox($message->getId(), $recipient['label'] ?? $recipient['email'], $recipient['email']);
+		}, array_merge($to, $cc, $bcc));
```
miaulalala, line 113: NONO!

lib/Db/LocalMailboxMessageMapper.php
miaulalala: i can attach all related data as a join

lib/Db/RecipientMapper.php
```diff
@@ -0,0 +1,81 @@
...
+				$qb->expr()->eq('mailbox_type', $qb->createNamedParameter(Recipient::MAILBOX_TYPE_OUTBOX, IQueryBuilder::PARAM_INT))
+			);
+
+		$result = $qb->execute();
+		$result->closeCursor();
+	}
+
+	public function createForLocalMailbox(int $messageId, string $label, string $email): Recipient {
+		$recipient = new Recipient();
+		$recipient->setType(Recipient::TYPE_TO);
```
miaulalala, line 73: get type passed to create instead of statically setting it here

lib/Db/LocalAttachmentMapper.php
```diff
@@ -55,4 +88,73 @@ public function find(string $userId, int $id): LocalAttachment {
...
+		$result = $qb->execute();
+		$attachmentId = $qb->getLastInsertId();
+		$result->closeCursor();
+
+		$qb2 = $this->db->getQueryBuilder();
+		$qb2->insert('mail_lcl_mbx_attchmts')
+			->setValue('local_message_id', $qb2->createNamedParameter($localMessageId))
+			->setValue('attachment_id', $qb2->createNamedParameter($attachmentId));
+		$result = $qb2->execute();
+		$result->closeCursor();
```
ChristophWurst, line 144: wrap in transaction

lib/Db/LocalMailboxMessage.php
```diff
@@ -0,0 +1,150 @@
...
+	/** @var int|null */
+	protected $draftId;
+
+	/** @var array|null */
+	protected $attachments;
+
+	/** @var array|null */
+	protected $recipients;
+
+	public const OUTGOING = 0;
```
ChristophWurst, line 86:
```suggestion
	public const TYPE_OUTGOING = 0;
```

?

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,135 @@
...
+		while (($row = $rows->fetch()) !== false) {
+			$results[] = $this->mapRowToEntity($row);
+			$ids[] = $row['id'];
+		}
+		$rows->closeCursor();
+
+		$attachments = $this->attachmentMapper->findAllForLocalMailbox($ids, $userId);
+		$recipients = $this->recipientMapper->findAllRecipients($ids, Recipient::MAILBOX_TYPE_OUTBOX);
+
+		return array_map(static function($entity) use ($attachments, $recipients) {
```
ChristophWurst, line 79: O(n*m), where n is the number of messages, m number of recipients. You have a loop inside a loop.

I would suggest to iterate over recipients only once and fill a hash map of the form `messageId -> recipient[]`. Then iterate over the messages and get the recipients in O(1) and O(n) in total.

Same applies to attachments.

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,135 @@
...
+					return Recipient::fromRow(($recipient));
+				}, array_filter($recipients, static function ($recipient) use ($entity){
+					return $entity->getId() === $recipient['message_id'];
+				}))
+			);
+			return $entity;
+		}, $results);
+	}
+
+	public function find(int $id, string $userId): LocalMailboxMessage {
```
ChristophWurst, line 102: misses `@throws DoesNotExistException`

ChristophWurst: rename to `findById`

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,135 @@
...
+		array_map(function ($recipient) use ($message) {
+			$this->recipientMapper->createForLocalMailbox($message->getId(), Recipient::TYPE_TO, $recipient['label'] ?? $recipient['email'], $recipient['email']);
+		}, $to);
+		array_map(function ($recipient) use ($message) {
+			$this->recipientMapper->createForLocalMailbox($message->getId(), Recipient::TYPE_CC, $recipient['label'] ?? $recipient['email'], $recipient['email']);
+		}, $cc);
+		array_map(function ($recipient) use ($message) {
+			$this->recipientMapper->createForLocalMailbox($message->getId(), Recipient::TYPE_BCC, $recipient['label'] ?? $recipient['email'], $recipient['email']);
+		}, $bcc);
+		$this->attachmentMapper->linkAttachmentToMessage($message->getId(), $attachmentIds);
```
ChristophWurst, line 127: wrap in transaction for atomicity and better performance

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	public function deleteWithRelated(LocalMailboxMessage $message): void {
+		$this->attachmentMapper->deleteForLocalMailbox($message->getId());
+		$this->recipientMapper->deleteForLocalMailbox($message->getId());
+		$this->delete($message);
```
ChristophWurst, line 133: wrap in transaction

lib/Db/Recipient.php
```diff
@@ -0,0 +1,84 @@
...
+
+	public const TYPE_FROM = 0;
+	public const TYPE_TO = 1;
+	public const TYPE_CC = 2;
+	public const TYPE_BCC = 3;
+
+	/** @var int */
+	protected $messageId;
+
+	/** @var int */
```
ChristophWurst, line 56: psalm-annotate

lib/Db/LocalAttachmentMapper.php
```diff
@@ -55,4 +88,73 @@ public function find(string $userId, int $id): LocalAttachment {
...
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('attachment_id')
+			->from('mail_lcl_mbx_attchmts')
+			->where($qb->expr()->eq('local_message_id', $qb->createNamedParameter($localMessageId), IQueryBuilder::PARAM_INT));
+
+		$qb1 = $this->db->getQueryBuilder();
+		$qb1->delete($this->getTableName())
+			->where($qb1->expr()->in('id', $qb1->createFunction($qb->getSQL()), IQueryBuilder::PARAM_INT_ARRAY), IQueryBuilder::PARAM_INT_ARRAY);
+		$result = $qb1->execute();
+		$result->closeCursor();
```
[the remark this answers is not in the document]
miaulalala: yep my bad

lib/Db/RecipientMapper.php
```diff
@@ -0,0 +1,102 @@
...
+
+		$qb->delete($this->getTableName())
+			->where(
+				$qb->expr()->eq('message_id', $qb->createNamedParameter($messageId, IQueryBuilder::PARAM_INT)),
+				$qb->expr()->eq('mailbox_type', $qb->createNamedParameter(Recipient::MAILBOX_TYPE_OUTBOX, IQueryBuilder::PARAM_INT))
+			);
+		$qb->execute();
+	}
+
+	public function saveRecipients(int $messageId, array $recipients,int $type, int $mailboxType = Recipient::MAILBOX_TYPE_INBOX): void {
```
miaulalala, line 84: this is erroring out for me with `PDOException: SQLSTATE[HY000]: General error: 1364 Field 'message_id' doesn't have a default value`.

I can alter the recipients table but I'm not sure that is a good decision. @ChristophWurst your input? :)

lib/Db/LocalAttachmentMapper.php
```diff
@@ -55,4 +88,83 @@ public function find(string $userId, int $id): LocalAttachment {
 
 		return $this->findEntity($query);
 	}
+
+	/**
+	 * @throws Exception
+	 */
+	public function findRows(string $userId, int $id): array {
```
ChristophWurst, line 95: unused?

lib/Migration/Version2000Date20220104144742.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCA\Mail\Migration;
```
ChristophWurst, line 5: license header

lib/Migration/Version2000Date20220104144742.php
```diff
@@ -0,0 +1,91 @@
...
+			'notnull' => false,
+			'length' => 4,
+		]);
+		$localMailboxTable->addColumn('draft_id', 'integer', [
+			'notnull' => false,
+			'length' => 4,
+		]);
+		$localMailboxTable->setPrimaryKey(['id']);
+
+		$attachmentsTable = $schema->createTable('mail_lcl_mbx_attchmts');
```
ChristophWurst, line 73: isn't `mail_local_mb_attachments` still short enough?

ChristophWurst (commented):
No more remarks for now :)

lib/Db/RecipientMapper.php
ChristophWurst: locally? do you have the oc_mail_recipients.message_id column?

miaulalala: yep I do :)

ChristophWurst: Phew. No clue. Could you step through this code and acquire the effective SQL query that will be sent to the DB?

lib/Migration/Version2000Date20220104144742.php
miaulalala: image

no sorry :( but mb instead of mbx is nicer

miaulalala: Compormise on `mail_local_mb_attchmts` ? :grin:

lib/Db/LocalAttachmentMapper.php
```diff
@@ -126,6 +127,7 @@ public function deleteForLocalMailbox(int $localMessageId): void {
 			$this->db->rollBack();
 			throw $e;
 		}
+		$this->db->commit();
```
st3iny, line 130: This has to be investigated further. Transaction have to be committed manually and there are many instances were it was forgotten.

st3iny: I fixed some instances I found during my testing of the frontend.

ChristophWurst: would it make sense to have a trait or function that does the repetitive pattern for us?

something like a

```php

function doTransactional(callable $fn, IDBConnection $db) {
  $db->beginTransaction();
  try {
    $result = $fn();
    $db->commit();
    return $result;
  } catch (Throwable $e) {
    $db->rollback();
    throw $e;
}
```

in https://github.com/nextcloud/mail/blob/main/lib/functions.php?

lib/Db/RecipientMapper.php
st3iny: I fixed it in my commit. The error is kind of hard to spot :D 

Compare both lines:
```php
		$qb->setValue('message_id, ', $qb->createParameter('message_id'));
		$qb->setValue('message_id', $qb->createParameter('message_id'));
```

A clue: Look at the position of the comma and the string quotes.

lib/Db/LocalAttachmentMapper.php
st3iny: Great idea!

lib/Db/RecipientMapper.php
miaulalala: thanks Richard, you're the real MVP!

lib/Db/LocalAttachmentMapper.php
ChristophWurst: https://github.com/nextcloud/server/pull/31205 I've also submit this upstream as a proposal. We wouldn't be able to use this as new APIs are not backported. As a short term solution we need this construct in our app as well.

miaulalala:
- [ ] add local_message_id on recipients

- [ ] add foreign key constraint

- [ ] rebuild recipientsmapper to use either message_id for IMAP or local_message_id for outbox messages (with FK constraints on each ;)

- [ ] look through deleteOrphans job to avoid recipients being deleted

lib/Db/LocalAttachmentMapper.php
```diff
@@ -24,28 +26,57 @@
...
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
+use Throwable;
 
 /**
  * @template-extends QBMapper<LocalAttachment>
  */
 class LocalAttachmentMapper extends QBMapper {
+	/** @var ITimeFactory */
+	private $timeFactory;
```
ChristophWurst, line 40: unused?

lib/Db/LocalMailboxMessageMapper.php
```diff
@@ -0,0 +1,142 @@
...
+use Throwable;
+use function array_map;
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+
+/**
+ * @template-extends QBMapper<LocalMailboxMessage>
+ */
+class LocalMailboxMessageMapper extends QBMapper {
```
ChristophWurst, line 38:
```suggestion
class LocalMessageMapper extends QBMapper {
```

lib/Migration/Version2000Date20220104144742.php
```diff
@@ -0,0 +1,110 @@
...
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		$localMailboxTable = $schema->createTable('mail_local_mailbox');
```
ChristophWurst, line 46:
```suggestion
		$localMailboxTable = $schema->createTable('mail_local_messages');
```

* plural is our convention
* this matches your entity name
* if we every store local mailboxes into the database, the name is available

ChristophWurst (changes requested):

miaulalala:
Migration might need a change in Version since it's Version2000Date20220104144742?

miaulalala:
- [ ] Tests

ChristophWurst:
> Migration might need a change in Version since it's Version2000Date20220104144742?

The plan is still to release the outbox as part of v1.12. Scheduled/undo sending will follow in another release.

lib/Migration/Version2000Date20220104144742.php
```diff
@@ -0,0 +1,108 @@
...
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use Doctrine\DBAL\Schema\Table;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version2000Date20220104144742 extends SimpleMigrationStep {
```
ChristophWurst, line 35: v1.12 is the target

miaulalala:
Removed FK constraints exception check from tests as SQLite doesn't support FK constraints by default. Also something we should be aware of and possibly recommend to users. More info here https://www.sqlite.org/foreignkeys.html

miaulalala:
StatAna isn't liking the PHP version:

 [InvalidArgumentException]                                                                                             
  Package christophwurst/nextcloud at version dev-master has a PHP requirement incompatible with your PHP version (7.3)

ChristophWurst:
> StatAna isn't liking the PHP version:
> 
> [InvalidArgumentException] Package christophwurst/nextcloud at version dev-master has a PHP requirement incompatible with your PHP version (7.3)

We fixed that with https://github.com/nextcloud/mail/pull/6066.

https://github.com/nextcloud/mail/pull/6132 updates the feature branch to latest main.

ChristophWurst (approved):

PR #6031, opened 2022-02-11
Add outbox backend implementation

miaulalala:
Closes https://github.com/nextcloud/mail/issues/6139

- [x] Adapt to https://github.com/nextcloud/mail/pull/6052
- [x] OutboxService::getMessages
- [x] OutboxService::getMessage
- [x] OutboxService::sendMessage
- [x] OutboxService::deleteMessage
- [x] LocalAttachmentMapperTest (Integration)
- [x] RecipientMapperTest (Integration)
- [x] LocalMessageMapperTest (Integration)
- [x] OutboxServiceIntegrationTest (Integration)
- [x] Fix OutboxControllerTest (Unit)
- [x] Fix OutboxServiceTest (Unit)
- [x] AttachmentServiceTest (Unit)
- [x] MailTransmissionUnitTest
- [x] RepiledMessageData in Transmission::sendLocalMessage
     - [x] Pass Message ID only and let the SentMessageEvent handle the mark replied
 - [x] change the Event Order to delete a draft immediately on scheduling send.
- [x] MailTransmissionIntegrationTest - cover sendLocalMessage
- [x] Debug `fwrite` error
- [x] Attachments: add an array of attachments and process them per type

### DIFF
--- a/src/store/outbox/getters.js
+++ b/src/store/outbox/getters.js
@@ -21,6 +21,6 @@
  */
 
 export default {
-	getAllMessages: state => Object.values(state.messages),
+	getAllMessages: state => state.messageList.map(id => state.messages[id]),
 	getMessage: state => id => state.messages[id],
 }
--- a/src/store/outbox/mutations.js
+++ b/src/store/outbox/mutations.js
@@ -26,8 +26,13 @@ export default {
 	addMessage(state, { message }) {
 		const existing = state.messages[message.id] ?? {}
 		Vue.set(state.messages, message.id, Object.assign({}, existing, message))
+		// Add the message only if it's new
+		if (state.messageList.indexOf(message.id) === -1) {
+			state.messageList.unshift(message.id)
+		}
 	},
 	deleteMessage(state, { id }) {
+		state.messageList = state.messageList.filter(i => i !== id)
 		Vue.delete(state.messages, id)
 	},
 	updateMessage(state, { message }) {
--- a/src/store/outbox/state.js
+++ b/src/store/outbox/state.js
@@ -21,5 +21,6 @@
  */
 
 export default {
+	messageList: [],
 	messages: {},
 }


miaulalala:
Don't forget to change the Event Order to delete a draft immediately on scheduling send.

lib/Controller/OutboxController.php
```diff
@@ -77,15 +77,9 @@ public function index(): JsonResponse {
...
-			$message = $this->service->getMessage($id, $this->userId);
-			$this->accountService->find($this->userId, $message->getAccountId());
-			return JsonResponse::success($message);
-		} catch (ClientException $e) {
-			return JsonResponse::fail($e->getMessage(), $e->getHttpCode());
-		} catch (DoesNotExistException $e) {
-			return JsonResponse::fail(null, Http::STATUS_NOT_FOUND);
-		}
+		$message = $this->service->getMessage($id, $this->userId);
+		$this->accountService->find($this->userId, $message->getAccountId());
```
ChristophWurst, line 79: doesn't the combined query for ID and user ID above already catch non authorized access?

miaulalala: Nice, another line of code gone :D

lib/Controller/OutboxController.php
```diff
@@ -124,70 +106,97 @@ public function create(
 		array $bcc = [],
 		array $attachmentIds = [],
 		?int $aliasId = null,
```
miaulalala, line 108: Possibly add the draft id here so the draft deletion can be triggered from here?

`$context->registerEventListener(MessageSentEvent::class, DeleteDraftListener::class);`

Is currently the event registered, changing that to a SaveOutboxMessageEvent?

ChristophWurst: OutboxMessageCreatedEvent

miaulalala: Damn it, so close. I had `OutboxMessageSavedEvent` :grin:

ChristophWurst:
CI *should* run but doesn't. I hope this is just a Github hiccup. This PR doesn't even change/touch the workflow files.

lib/Controller/OutboxController.php
```diff
@@ -110,84 +92,116 @@ public function show(int $id): JSONResponse {
...
+	 * @param int|null $aliasId
+	 * @param string|null $inReplyToMessageId
+	 * @return JsonResponse
 	 */
 	public function create(
 		int $accountId,
 		string $subject,
 		string $body,
-		bool $isHtml,
+		bool $isHtml
```
st3iny, line 104:
```suggestion
		bool $isHtml,
```

miaulalala:
StatAna isn't liking the event dispatcher, otherwise it looks ok

ChristophWurst:
> StatAna isn't liking the event dispatcher, otherwise it looks ok

User the interface instead of the implementation

lib/Db/LocalMessageMapper.php
```diff
@@ -117,25 +121,49 @@ public function findById(int $id, string $userId): LocalMessage {
...
 		$this->db->beginTransaction();
 		try {
 			$message = $this->insert($message);
-			$this->recipientMapper->saveRecipients($message->getId(), $to, Recipient::TYPE_TO);
-			$this->recipientMapper->saveRecipients($message->getId(), $cc, Recipient::TYPE_CC);
-			$this->recipientMapper->saveRecipients($message->getId(), $bcc, Recipient::TYPE_BCC);
+			$this->recipientMapper->saveRecipients($message->getId(), $to);
+			$this->recipientMapper->saveRecipients($message->getId(), $cc);
+			$this->recipientMapper->saveRecipients($message->getId(), $bcc);
+			//@todo
```
ChristophWurst, line 131: todo?

lib/Service/AutoConfig/ConnectivityTester.php
```diff
@@ -26,7 +26,7 @@
 use Psr\Log\LoggerInterface;
 
 class ConnectivityTester {
-	public const CONNECTION_TIMEOUT = 5;
+	public const CONNECTION_TIMEOUT = 20;
```
ChristophWurst, line 29: revert

lib/Service/MailTransmission.php
```diff
@@ -212,6 +219,53 @@ public function sendMessage(NewMessageData $messageData,
...
+			}
+		}
+
+		try {
+			$this->sendMessage($messageData, $replyData ?? null, $alias ?? null);
+		} catch (SentMailboxNotSetException $e) {
+			throw new ClientException('Could not send message' . $e->getMessage(), (int)$e->getCode(), $e);
+		}
+
+		// @todo mark message as replied to???? or does the send handle that?
```
ChristophWurst, line 266: todo?

lib/Service/OutboxService.php
```diff
@@ -0,0 +1,132 @@
...
+								IMailManager $mailManager) {
+		$this->transmission = $transmission;
+		$this->mapper = $mapper;
+		$this->attachmentService = $attachmentService;
+		$this->eventDispatcher = $eventDispatcher;
+		$this->mailManager = $mailManager;
+	}
+
+	/**
+	 * @param array $recipients
```
ChristophWurst, line 71:
```suggestion
	 * @param Recipient[] $recipients
```

miaulalala: it's not.

lib/Db/MessageMapper.php
```diff
@@ -615,6 +614,25 @@ public function findThread(Account $account, string $threadRootId): array {
...
+	 *
+	 * @return Message[]
+	 */
+	public function findByMessageId(Account $account, string $messageId): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('messages.*')
+			->from($this->getTableName(), 'messages')
+			->join('messages', 'mail_mailboxes', 'mailboxes', $qb->expr()->eq('messages.mailbox_id', 'mailboxes.id', IQueryBuilder::PARAM_INT))
+			->where(
+				$qb->expr()->eq('mailboxes.account_id', $qb->createNamedParameter($account->getId(), IQueryBuilder::PARAM_INT)),
```
ChristophWurst, line 629:
```suggestion
				$qb->expr()->eq('mailboxes.account_id', $qb->createNamedParameter($account->getId(), IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT),
```

lib/Service/Attachment/AttachmentService.php
```diff
@@ -102,4 +103,59 @@ public function deleteAttachment(string $userId, int $id) {
...
+		// delete storage
+		foreach ($attachments as $attachment) {
+			$this->storage->delete($userId, $attachment->getId());
+		}
+	}
+
+	/**
+	 * @return LocalAttachment[]
+	 */
+	public function saveLocalMessageAttachments(int $messageId, array $attachmentIds): array {
```
ChristophWurst, line 121: document the array to be int[]

lib/Service/MailManager.php
```diff
@@ -746,4 +746,8 @@ public function deleteThread(Account $account, Mailbox $mailbox, string $threadR
 			);
 		}
 	}
+
+	public function getByMessageId(Account $account, string $messageId): array {
```
ChristophWurst, line 750: phpdoc for the returned array

miaulalala:
A few notes:

The test setup for the embedded IMAP attachment `\OCA\Mail\Tests\Integration\Service\OutboxServiceIntegrationTest::testSaveAndGetMessageWithMessageAttachment` is with 99% certainty a false positive at the moment (unless the mail server on Github is running better than the one locally).

Why?

Horde does not allow appending a message with an attachment via `Horde_Mime_Mail`. When using an attachment part via `$mail->addAttachment`, the part will be ignored and not written to IMAP. Same goes for manually adding a new `Horde_Mime_Part`.

My research lead me to use a string which is a `multipart/mixed` message in the IMAP test so I could run some Integration tests on the attachment handling:

https://github.com/nextcloud/mail/blob/2165b484945bb6c13a3cd70afaf5445edc2bc300/tests/Integration/Framework/ImapTest.php#L157-L189

The message can't be decoded by `\OCA\Mail\IMAP\MessageMapper::getRawAttachments` and leads to a mime type of `plain/x-empty` (?) and an empty content for the attachment when downloading the attachment from IMAP.

I am unsure if this is because of the way I created the message, because of a limitation with horde's append function or a limitation of IMAP (doubtful, base64 encoded attachments should be fine). This needs to be investigated more but I am at the end of my knowledge.

I also couldn't test the frontend as I have issues with npm again that I need some expert advice to fix.

I would be grateful if you could test this @ChristophWurst!

ChristophWurst:
That code looks very scary. If there is no straight forward way to do this then I'd rather not have a test than one that builds a MIME part by hand.

ChristophWurst:
What versions of node/npm do you use?

ChristophWurst:
> When using an attachment part via `$mail->addAttachment`, the part will be ignored and not written to IMAP. Same goes for manually adding a new `Horde_Mime_Part`.

How did you use ``\Horde_Mime_Mail::addAttachment``?

miaulalala:
> > When using an attachment part via `$mail->addAttachment`, the part will be ignored and not written to IMAP. Same goes for manually adding a new `Horde_Mime_Part`.
> 
> How did you use `\Horde_Mime_Mail::addAttachment`?

```php
$mail->addAttachment('thisIsMyFileContentAsAString', 'test.txt', 'attachment');
```

ChristophWurst:
- [ ] BUG

Bildschirmfoto von 2022-04-06 11-42-26
Bildschirmfoto von 2022-04-06 11-42-35

ChristophWurst:
> ```
> $mail->addAttachment('thisIsMyFileContentAsAString', 'test.txt', 'attachment');
> ```

The first argument should be a path to your file, not the contents as string

miaulalala:
> > ```
> > $mail->addAttachment('thisIsMyFileContentAsAString', 'test.txt', 'attachment');
> > ```
> 
> The first argument should be a path to your file, not the contents as string

I did try this for sure, I'm not Jason who's 19 and can't read ;) 

But it doesn't matter anyway:

```php
    public function getRaw($stream = true)
    {
        if ($stream) {
            $hdr = new Horde_Stream();
            $hdr->add($this->_headers->toString(), true);
            return Horde_Stream_Wrapper_Combine::getStream(
                array($hdr->stream,
                      $this->getBasePart()->toString(
                        array('stream' => true, 'encode' => Horde_Mime_Part::ENCODE_7BIT | Horde_Mime_Part::ENCODE_8BIT | Horde_Mime_Part::ENCODE_BINARY))
                )
            );
        }

        return $this->_headers->toString() . $this->getBasePart()->toString();
    }

```

There is no way to get the additional body parts from a raw message.

miaulalala:
> * [ ]  BUG
> 
> Bildschirmfoto von 2022-04-06 11-42-26 Bildschirmfoto von 2022-04-06 11-42-35

should I unset the param in the backend?

ChristophWurst:
Bildschirmfoto von 2022-04-06 20-25-16

Sending messages with attachments fails successfully

ChristophWurst:
Bildschirmfoto von 2022-04-06 20-27-48

lib/Controller/AccountsController.php
```diff
@@ -421,14 +420,13 @@ public function send(int $id,
...
 		$messageData = NewMessageData::fromRequest($account, $expandedTo, $expandedCc, $expandedBcc, $subject, $body, $attachments, $isHtml, $requestMdn);
-		$repliedMessageData = null;
 		if ($messageId !== null) {
 			try {
 				$repliedMessage = $this->mailManager->getMessage($this->currentUserId, $messageId);
+				$repliedMessageId = $repliedMessage->getMessageId();
 			} catch (ClientException $e) {
 				$this->logger->info("Message in reply " . $messageId . " could not be loaded: " . $e->getMessage());
 			}
-			$repliedMessageData = new RepliedMessageData($account, $repliedMessage);
```
ChristophWurst, line 431: what the old data structure and identification allowed was to receive a message with account A, reply with account B and still get the flag updates on account A. This will break now.

lib/Service/OutboxService.php
ChristophWurst: Ideally this should be pushed up to the controller where we do the conversion of stringly typed HTTP into the PHP world

ChristophWurst (approved):
Works

Code looks good but couldn't grasp all the changes due to the size of the patch

lib/Controller/AccountsController.php
ChristophWurst: Raised as https://github.com/nextcloud/mail/issues/6189

Issue #5835, opened 2021-12-13
Performance improvements


ChristophWurst:
### Feature Request

The app performance can be improved. This is an overview. Tickets are sorted from highest impact to lowest.

* [x] https://github.com/nextcloud/mail/issues/4466
* [x] https://github.com/nextcloud/mail/issues/6305
* [x] https://github.com/nextcloud/mail/issues/6307
* [x] https://github.com/nextcloud/mail/issues/4823
* [x] https://github.com/nextcloud/mail/issues/4983
* [x] https://github.com/nextcloud/mail/issues/6308
* [x] https://github.com/nextcloud/mail/issues/4664
* [x] https://github.com/nextcloud/mail/issues/5294

Removed from the short term plan
* https://github.com/nextcloud/mail/issues/5250

Related changes
* https://github.com/nextcloud/mail/issues/4735

Improvements that do not show any performance benefits right now
* https://github.com/nextcloud/mail/issues/4824

#### Summary

Make the app faster and more usable.

brendan-pike:
Our small organisation has just starting using mail for its main communication and file sharing with clients. We have 4 employees sharing a single email account,  to achieve this we simply added the imap account for each user.  Loading an email can take 10 to 20 second to load a single email. We have a VPS with lots of resources 4 CPU/8GB low load and has been highly tuned as per nextcloud documentation.

Performance is our top issue right now.  So I see this ticket is very important.

ChristophWurst:
You judged too early. The tags just tell us how quickly we need to pick something up. The linked tickets will be addressed shortly.

brendan-pike:
Thanks for the feedback Christoph, that is encouraging to hear.

My team has had me install afterlogic and rainloop and have been testing 
them and while of course they are faster none of them properly integrate 
in the way that nextcloud mail does.  Nextlcoud mail will always be 
everyone's first choice, but as a tool used all day long the speed issue 
is significant.

Also, I'm new to this community but I've been a dedicated opensource 
user and advocate for decades. I would strongly encourage you to not 
hide comments that you find negative, unless they are offensive or a 
personal attack. Censorship quickly disengages a community where as an 
open response if respected even if it strongly disagrees with the first 
person.

Thank you for everything you and the Nextcloud team do!

Cheers
Brendan

PR #6049, opened 2022-02-18
Log slow MySQL queries on CI

ChristophWurst:
To help us with https://github.com/nextcloud/mail/issues/5835

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,6 +128,12 @@ jobs:
             run: |
                 php -f nextcloud/occ config:system:set debug --type bool --value true
                 php -f nextcloud/occ config:system:set app.mail.verify-tls-peer --type bool --value false
+          - name: Enable slow mysql query logs
+            if: ${{ matrix.db == 'mysql' }}
+            run: |
+              echo "SET GLOBAL log_queries_not_using_indexes = 1;" | mysql -h 127.0.0.1 -u root -pmy-secret-pw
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -128,6 +128,12 @@ jobs:
             run: |
                 php -f nextcloud/occ config:system:set debug --type bool --value true
                 php -f nextcloud/occ config:system:set app.mail.verify-tls-peer --type bool --value false
+          - name: Enable slow mysql query logs
+            if: ${{ matrix.db == 'mysql' }}
+            run: |
+              echo "SET GLOBAL log_queries_not_using_indexes = 1;" | mysql -h 127.0.0.1 -u root -pmy-secret-pw
```
ChristophWurst, line 134: on CI queries aren't slow enough to trigger a *slow query* as per https://mariadb.com/kb/en/slow-query-log-overview/, so I've added this line to show anything that doesn't use an index.

nickvergessen (commented):
Clever 👍
Hopefully you don't get too much false positives

miaulalala (approved):
Great idea! :shipit:

.github/workflows/test.yml
nickvergessen: Any clue if that shows queries where an index exists but is not used because it matches 90% of the rows anyway and stuff like this?

ChristophWurst: That is indeed the case. So if your tests clear a table with `DELETE FROM xyz` it will show up.

PR #6066, opened 2022-02-22
Fix Psalm checks against server master with PHP7.3

ChristophWurst:
The OCP package doesn't allow for PHP7.3 but our composer.json
overwrites it so that we do not update dependencies to versions with too
restrictive minimum PHP versions (looking at you Dependabot). Hence we
ignore the platform dependency here. Once PHP7.3 got dropped we will
update composer.json and the platform requirement can be enforced again.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -25,7 +25,11 @@ jobs:
                 coverage: none
           - name: Install dependencies
             run: composer i
-          - name: Install dependencies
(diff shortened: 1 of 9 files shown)


miaulalala (approved):

ChristophWurst:
https://github.com/nextcloud/mail/pull/6059 accidentally updated the psr container to a version that requires PHP7.4. I downgraded to revive PHP7.3 compat.

PR #6092, opened 2022-02-27
Triming IMAP and SMTP hosts on setup.

Ad5001:
Often when you try to add new mail accounts from manual setup, you copy the IMAP and SMTP hosts from your provider. However, more than often, copying and pasting like that may lead to additional unwanted whitespaces on the hosts fields.

This pull requests adds trimming to both IMAP and SMTP hosts on setup in the Accounts controller.

It doesn't cost much but is a great quality of life feature because the mail app tries, unsuccessfully, to connect to hosts with whitespaces and doesn't give any meaningful message related to it.

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -332,9 +332,14 @@ export default {
 					...this.autoConfig,
 				})
 			} else {
+				// Removing additional whitespaces from manual configuration hosts
+				// In order to avoid issues when copy pasting imap & smtp hosts from providers documentations,
+				// which may have whitespaces.
 				return this.save({
 					autoDetect: false,
 					...this.manualConfig,
+					imapHost: this.manualConfig.imapHost.trim(),
+					smtpHost: this.manualConfig.smtpHost.trim(),
 				})
 			}
 		},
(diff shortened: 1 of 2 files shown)


ChristophWurst:
I'm fine with trimming. It's a nice enhancement. But I'd rather put this responsibility to the frontend because that is where we deal with user input. The controller method on the backend should expect clean input data.

Are you familiar with javascript development to move the logic?

Ad5001:
While I'm not familiar with Vue.js, I'm familiar enough with javascript to move it there. I'll look into the format of the manualConfig variable when I can.

src/components/AccountForm.vue
```diff
@@ -332,6 +332,11 @@ export default {
...
 				})
 			} else {
+				// Removing additional whitespaces from manual configuration hosts
+				// In order to avoid issues when copy pasting imap & smtp hosts from providers documentations,
+				// which may have whitespaces.
+				this.manualConfig.imapHost = this.manualConfig.imapHost.trim()
+				this.manualConfig.smtpHost = this.manualConfig.smtpHost.trim()
 				return this.save({
 					autoDetect: false,
 					...this.manualConfig,
```
ChristophWurst, line 342:
```suggestion
				return this.save({
					autoDetect: false,
					...this.manualConfig,
					imapHost: this.manualConfig.imapHost.trim(),
					smtpHost: this.manualConfig.smtpHost.trim(),
```

same result but without the side effects

Ad5001: I didn't know that syntax was possible. Added to the changes.

ChristophWurst (approved):

ChristophWurst:
@Ad5001 DCO is not happy. Could you squash your commits into one and sign off that final commit? Thanks :v:

PR #6099, opened 2022-02-28
Improve account setup error reporting

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3006

Todo
- [x] Improve backend error reporting
- [x] Handle errors on front-end

---

### Autoconfig fails

HTTP400

```json
{
  "status": "fail",
  "data": {
    "error": "AUTOCONFIG_FAILED",
    "message": "Auto detect failed. Please try manual mode." // this is translated. we can also use a translated string on the frontend
  }
}
```

### Manual config fails if server is not reachable

HTTP400

```json
{
  "status": "fail",
  "data": {
    "error": "CONNECTION_ERROR",
    "service": "IMAP", // or "SMTP"
    "host": "mail.domain.tld",
    "port": 587,
  }
}
```

### Manual config fails if server is username or password is wrong

HTTP400

```json
{
  "status": "fail",
  "data": {
    "error": "AUTHENTICATION",
    "service": "IMAP", // or "SMTP"
    "host": "mail.domain.tld",
    "port": 587,
  }
}
```

### Anything other handled error

HTTP400

```json
{
  "status": "fail",
  "data": {
    "error": "OTHER"
  }
}
```

### Any unexpected error

HTTP500

No specified format. Content doesn't matter.

## Forntend improvements 

- Frontend validation for the email was added below the email field. Validation will take place when the focus will be left. 

image

- Connect button is not available, if required fields are not filled out or other frontend validation failed.
- User will see detailed backend errors.
- Connect button was improved.
image

### DIFF
--- a/src/views/Setup.vue
+++ b/src/views/Setup.vue
@@ -1,14 +1,17 @@
 <template>
 	<Content app-name="mail">
 		<Navigation v-if="hasAccounts" />
-		<div id="emptycontent">
-			<div class="icon-mail" />
-			<h2>{{ t('mail', 'Connect your mail account') }}</h2>
-			<AccountForm :display-name="displayName" :email="email" :save="onSave">
-				<template v-if="error" #feedback class="warning">
-					{{ error }}
+		<div class="mail-empty-content">
+			<EmptyContent icon="icon-mail">
+				<h2>{{ t('mail', 'Connect your mail account') }}</h2>
+				<template #desc>
+					<AccountForm :display-name="displayName" :email="email" :save="onSave">
+						<template v-if="error" #feedback class="warning">
+							{{ error }}
+						</template>
+					</AccountForm>
 				</template>
-			</AccountForm>
+			</EmptyContent>
 		</div>
 	</Content>
 </template>
@@ -19,6 +22,7 @@ import { loadState } from '@nextcloud/initial-state'
 import { translate as t } from '@nextcloud/l10n'
 
 import AccountForm from '../components/AccountForm'
+import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
 import Navigation from '../components/Navigation'
 import logger from '../logger'
 
@@ -27,6 +31,7 @@ export default {
 	components: {
 		AccountForm,
 		Content,
+		EmptyContent,
 		Navigation,
 	},
 	data() {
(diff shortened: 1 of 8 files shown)


JuliaKirschenheuter:
@jancborchardt @nimishavijay 
this manual setting is actually too long. Should we deal with it somehow?
image

ChristophWurst:
@JuliaKirschenheuter let's focus on the error handling in this change, otherwise we'll find more and more things to change and this won't be reviewable nor does it ever reach an end :)

Regarding the design I seem to remember that @jancborchardt once proposed something like a tabbed interface. I know we don't want to go with a modal but the settings modal we use for accounts would help us structure this.

But again, something for another ticket and PR :)

src/views/Setup.vue
```diff
@@ -1,32 +1,31 @@
 <template>
-	<Content app-name="mail">
-		<Navigation v-if="hasAccounts" />
-		<div id="emptycontent">
-			<div class="icon-mail" />
+	<EmptyContent icon="icon-mail">
```
skjnldsv, line 2: I find it odd that you can drop the Content, please double check this is intentional :)

JuliaKirschenheuter: Thank you! I have corrected it

JuliaKirschenheuter:
@nimishavijay @jancborchardt The appearance of the whole setup is a bit different, because i have used this component: https://nextcloud-vue-components.netlify.app/#/Components/EmptyContent. The empty content will be consistent over whole app.

jancborchardt:
@JuliaKirschenheuter that looks very nice, good on using the EmptyContent component! :)

------

Btw if it is easier, we don’t have to stick to the tabbed Auto / Manual interface. We can also go with an expander between password and button like so. But probably also something for another pull request:
image


> Regarding the design I seem to remember that @jancborchardt once proposed something like a tabbed interface. I know we don't want to go with a modal but the settings modal we use for accounts would help us structure this.

@ChristophWurst just for the record, we’re not against a modal per se here :) @nimishavijay and I talked about it and our verdict was:
- For initial setup, it looks best for the setup to show in the content. Showing in a modal would be strange as there is nothing behind anyway
- For adding a new account when you already have existing ones, it would actually look better to have it in a modal
- Especially because as you said we have the account settings in a modal for existing accounts already as well

But as you said → something for a separate PR. :)

jancborchardt (approved):
Looks good design-wise – all feedback from the comment is for separate PRs :)

src/views/Setup.vue
```diff
@@ -58,10 +63,22 @@ export default {
 				.catch((error) => {
 					logger.error('Could not create account', { error })
 
-					if (error?.data?.message) {
-						this.error = error.data.message
+					if (error.data.error === 'AUTOCONFIG_FAILED') {
```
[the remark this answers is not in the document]
ChristophWurst: Fixed.

JuliaKirschenheuter (approved):

GretaD (approved):
:v:

PR #6132, opened 2022-03-18
Merge main into feature/outbox

ChristophWurst:
https://github.com/nextcloud/mail/tree/feature/outbox is 25d old and misses some important fixes from main.


miaulalala (approved):
:shipit:

PR #6150, opened 2022-03-24
Fix index of next message to delete

### LINKED ISSUE #6042: Deleting topmost message in a list does not move to open next message
### Steps to reproduce

1. Open any folder, e.g. your inbox
2. Go to the topmost message
3. Delete it, either via the menu or the Delete key

### Expected behavior

Move to the next mail below and open it. It does work if you do it with any other mail in the list, moving to the next newest. Of course there needs to be 1 special case, which is to open the next older mail if there is no next newest mail. :)

(This used to work at some point in the past which is why it is a regression.)

### Actual behavior

The envelope view on the right turns blank and no message is selected.

### Mail app version

Current on cloud.nextcloud.com

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

JuliaKirschenheuter:
For selected messages: navigation after all messages have been deleted
For selected messages: Fix index of next message to delete

fixes #6042

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -341,54 +341,63 @@ export default {
 			this.unselectAll()
 		},
 		async deleteAllSelected() {
-			for (const envelope of this.selectedEnvelopes) {
-				// Navigate if the message being deleted is the one currently viewed
-				// Shouldn't we simply use $emit here?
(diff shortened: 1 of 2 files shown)


GretaD (approved):
tested and it works perfectly :+1:

PR #6158, opened 2022-03-24
Merge main into feature/outbox (again)

ChristophWurst:
Same as https://github.com/nextcloud/mail/pull/6132 but a week later.


GretaD (approved):
didnt test it but the vue part at least looks good

PR #6172, opened 2022-03-31
Check mail for missing attachment

max65482:
This PR sketches a solution to awkward problems that I sometimes encounter before my first coffee: I misspell the name of the email recipient in the salutation (Dear XXX) or I mention an attachment, but forget to attach something.

Before sending an email, the app now searches for these accidents and warns the user.

If there is interest, I can refactor and give the code presentable quality. Otherwise, feel free to close this.

grafik

grafik

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -30,20 +30,27 @@
 		<Loading v-else-if="sending"
 			:hint="t('mail', 'Sending …')"
 			role="alert" />
-		<EmptyContent v-else-if="warning" :title="t('mail', 'Warning sending your message')" role="alert">
-			<p>{{ warning }}</p>
-			<ButtonVue
-				type="tertiary"
-				:aria-label="t('mail', 'Go back')"
-				@click="warning = undefined">
-				{{ t('mail', 'Go back') }}
-			</ButtonVue>
-			<ButtonVue
-				type="tertiary"
-				:aria-label="t('mail', 'Send anyway')"
-				@click="onForceSend">
-				{{ t('mail', 'Send anyway') }}
-			</ButtonVue>
+		<EmptyContent v-else-if="warning"
+			:title="t('mail', 'Warning sending your message')"
+			class="centered-content"
+			role="alert">
+			<template #description>
+				{{ warning }}
+			</template>
+			<template #action>
+				<ButtonVue
+					type="tertiary"
+					:aria-label="t('mail', 'Go back')"
+					@click="warning = undefined">
+					{{ t('mail', 'Go back') }}
+				</ButtonVue>
+				<ButtonVue
+					type="tertiary"
+					:aria-label="t('mail', 'Send anyway')"
+					@click="onForceSend">
+					{{ t('mail', 'Send anyway') }}
+				</ButtonVue>
+			</template>
 		</EmptyContent>
 		<template v-else>
 			<NcActions class="minimize-button">
@@ -114,6 +121,7 @@ import { UNDO_DELAY } from '../store/constants'
 import { matchError } from '../errors/match'
 import NoSentMailboxConfiguredError from '../errors/NoSentMailboxConfiguredError'
 import ManyRecipientsError from '../errors/ManyRecipientsError'
+import AttachmentMissingError from '../errors/AttachmentMissingError.js'
 import Loading from './Loading'
 import { mapGetters } from 'vuex'
 import MinimizeIcon from 'vue-material-design-icons/Minus.vue'
@@ -290,7 +298,7 @@ export default {
 				.then(() => logger.debug('attachments uploaded'))
 				.catch((error) => logger.error('could not upload attachments', { error }))
 		},
-		async onSend(data) {
+		async onSend(data, force = false) {
 			logger.debug('sending message', { data })
 
 			await this.attachmentsPromise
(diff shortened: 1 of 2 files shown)


src/components/NewMessageModal.vue
```diff
@@ -131,11 +132,79 @@ export default {
...
+						attachmentMentioned = true
+						break
+					}
+				}
+
+				if (attachmentMentioned) {
+					if (warnings !== "") {
+						warnings += ", "
+					}
+					warnings += "missing attachments"
```
ChristophWurst, line 175: this needs translation

ChristophWurst (commented):
I like the feature, e.g. to warn for the possibly missing attachments.

Translation will be tricky, because there are one or multiple words in each language (and variants of them, think: singular vs plural). For HTML messages we need to saveguard that a translation doesn't match a HTML tag or attribute, because that would show false warnings.

The editor is based on CKEditor. We could check if there is anything existing, like a module built in or a third party module that we could add.

src/components/NewMessageModal.vue
ChristophWurst: @max65482  could you address this please?

max65482: I will be looking at this PR again soon (with warnings only about missing attachments, not salutations for now).
(If you are referring to the translation: At this time I feel unable to address this.)

max65482:
> Translation will be tricky, because there are one or multiple words in each language (and variants of them, think: singular vs plural).

For simplicity, we should assume that the singular word is always part of the plural word.

> For HTML messages we need to saveguard that a translation doesn't match a HTML tag or attribute, because that would show false warnings.

`toPlain(data.body)` should be enough to get rid of HTML tags etc., right?

max65482:
@ChristophWurst Can you give it another look?

src/components/NewMessageModal.vue
```diff
@@ -131,6 +132,27 @@ export default {
...
+				var attachmentMissing = false
+				var lines = toPlain(data.body).value.toLowerCase().split('\n')
+				var wordAttachment = t('mail', 'attachment').toLowerCase()
+				var wordAttached = t('mail', 'attached').toLowerCase()
+				for (let line of lines) {
+					if (line.startsWith('>') || line.startsWith('--')) {
+						break
+					}
+					if (line.includes(wordAttachment) || line.includes(wordAttached)) {
+						attachmentMissing = true
```
ChristophWurst, line 146: remove the variable and just throw here

src/components/NewMessageModal.vue
```diff
@@ -131,6 +132,27 @@ export default {
 				...data,
 				body: data.isHtml ? data.body.value : toPlain(data.body).value,
 			}
+
+			if (!data.force && data.attachments.length === 0) {
+				var attachmentMissing = false
+				var lines = toPlain(data.body).value.toLowerCase().split('\n')
+				var wordAttachment = t('mail', 'attachment').toLowerCase()
+				var wordAttached = t('mail', 'attached').toLowerCase()
```
ChristophWurst, line 140: replace `var` with `const`

ChristophWurst (commented):
Looks good otherwise codewise

Are users able to send despite the error?

max65482:
@ChristophWurst I rebased and stumbled across changes that came with [this](https://github.com/nextcloud/mail/commit/b07457ad86cf05d6aad2bca511bcb476c991293f#diff-77b33550ce11bdb5cd04cb76105da60c54610d3c204ab3c20f513e2336a66d51) commit.

https://github.com/nextcloud/mail/blob/41b17c7d7cb204a7692a91a0a1cac5810dc42114/src/components/NewMessageModal.vue#L6-L32

In particular, `onForceSend` does not exist in this context. Could you help out?

ChristophWurst:
We had to move some code around for a new feature. Sorry for the confusion.

max65482:
I still cannot find `onForceSend` in this modal although it is referenced. This method would be required for the extension. In my understanding this is technical debt.
As I don't feel proficient enough to fix this on my own, I opened #8617. I will continue once this is fixed.

max65482:
Thanks to #8625 `onForceSend` exists now.

Now I stumbled across another issue. NewMessageModal seems to use states that aren't defined in the modal:
https://github.com/nextcloud/mail/blob/8f7fd93f4081b2a8bb07624ab06402214777976b/src/components/NewMessageModal.vue#L18

Can someone help out? @GretaD?

GretaD:
> Thanks to #8625 `onForceSend` exists now.
> 
> Now I stumbled across another issue. NewMessageModal seems to use states that aren't defined in the modal:
> 
> https://github.com/nextcloud/mail/blob/8f7fd93f4081b2a8bb07624ab06402214777976b/src/components/NewMessageModal.vue#L18
> 
> Can someone help out? @GretaD?

i am having a look if its a leftover or needs to be declared

GretaD:
> Thanks to #8625 `onForceSend` exists now.
> 
> Now I stumbled across another issue. NewMessageModal seems to use states that aren't defined in the modal:
> 
> https://github.com/nextcloud/mail/blob/8f7fd93f4081b2a8bb07624ab06402214777976b/src/components/NewMessageModal.vue#L18
> 
> Can someone help out? @GretaD?

the pr to remove the leftovers is merged. You can continue now :)

max65482:
Thanks @GretaD!
PR is ready for review now.

GretaD:
it works as described, but instead of warning it shows an error. I remember seeing a ticket related to implementing warning, but couldnt find it
Screenshot from 2023-08-07 16-33-07

max65482:
Yes, as of now there is no distinction between error and warning. E.g. `ManyRecipientsError` would also show up as an error although it is more of hint. I agree that (ideally) there would be a separate mechanism for warnings.

Speaking about `ManyRecipientsError`, this might be an opportunity for fixing #6461 as well.

GretaD (approved):
we can get this in, when we have the warning part done, we can change this into a warning instead of an error.

src/components/NewMessageModal.vue
```diff
@@ -362,7 +390,7 @@ export default {
 			}
 		},
 		async onForceSend() {
-			await this.onSend(null, true)
+			await this.onSend(this.cookedComposerData, true)
```
ChristophWurst, line 393: :cook: ?

ChristophWurst: Just realized the prop is really called like that. nevermind.

GretaD:
after rebasing, this pr has issues, having a look why

GretaD:
running npm run lint:fix changed a lot of things. I hope i didnt broke something..now looks good

PR #6174, opened 2022-04-04
Handle IMAP disconnects more explicitly

ChristophWurst:
This changes the IMAP client usages to be used as a *resource* that is
freed after finished use. Previously we just memoized connections to the
same account to lower the number of connections, but that has still
shown to cause too many open connections during tests but possibly also
in production.

* [x] Move IMAP client usage into try-(catch-)finally patterns
* [x] Drop the IMAP client factory memoization
* [x] Open follow-up ticket to diagnose if we have potential reconnects because of this change. If we previously used the memoized connection in two sections of a code there is now a little overhead. It might not be worth the optimization but should be checked nevertheless. https://github.com/nextcloud/mail/issues/6175

### DIFF
--- a/lib/Command/DiagnoseAccount.php
+++ b/lib/Command/DiagnoseAccount.php
@@ -84,7 +84,6 @@ class DiagnoseAccount extends Command {
 			$output->writeln('<error>No IMAP passwort set. The user might have to log into their account to set it.</error>');
 		}
 		$imapClient = $this->clientFactory->getClient($account);
-
 		try {
 			$this->printCapabilitiesStats($output, $imapClient);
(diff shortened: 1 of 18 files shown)


miaulalala:
https://github.com/nextcloud/mail/blob/689ef95dd7d36142fc1037eb5f2c9bfdf90d8a41/lib/Account.php#L185-L190

Is still using the deprecated getImapConnection method.

Refctor?

ChristophWurst:
``\OCA\Mail\Account::getMailbox`` is deprecated itself. I wouldn't refactor this anymore but try to get rid of those usages.

miaulalala (approved):

PR #6177, opened 2022-04-04
Feature/undo message sending

GretaD:


GretaD:
pushed on master, should have pushed on feature/outbox

Issue #5438, opened 2021-08-12
Undo sending an email


nimishavijay:
## Feature Request

Many emails are sent because you accidentally hit the "Send" button prematurely. In Gmail, Apple mail, Outlook and Mailbird there is an option to undo sending an email for that's available for a few seconds after you hit "send". We can have something like that in Mail as well.

Here's how Gmail does it (pretty much everyone else does it the same way too).
After clicking on send, there's a little pop up in the bottom left corner of the screen which says "Sending..." which turns into this popup once the message has been sent. Here you can either undo the sending, which turns it back into a draft and opens the message composer, or view the message, or dismiss the popup.
![image](https://user-images.githubusercontent.com/52440189/129203871-c6b15713-9a8e-470f-9e64-f906e3b00a32.png)

The duration for showing this popup is configurable in the settings, but even just having this option with a set duration of 15 seconds or so would be super useful.

## Implementation

* [x] #5884
* [x] Utilize the outbox
  + Push the message to the outbox when fake sending it
  + Send me message after x seconds
  + Drop the message off the outbox if sending is undone

---

cc @jancborchardt and @karlitschek

karlitschek:
very cool!!! 👍

GretaD:
Steps how this will work:\
1. I click send
2. This  feedback will  be shown and look more or less like this: 
![undo](https://user-images.githubusercontent.com/12728974/160658273-b71487ed-d697-42a9-bee7-b2a096c78f99.png)
3. After i click the undo button it will take me to the modal message in an editing state of the message i am sending i order to do the changes i need before sending again

@jancborchardt @nimishavijay

ChristophWurst:
The alternative would have been to stay in the existing composer modal and where we already say *Message sent* there could be an *Undo* button underneath that goes back to the editor. That would avoid the closing and re-opening of the modal.

@jancborchardt @nimishavijay your call.

nimishavijay:
> The alternative would have been to stay in the existing composer modal and where we already say

This sounds good. We could have an empty content component with "Message sent" as the text and a button that says "Undo". The modal could disappear automatically after 5 seconds or 7 seconds, or dismissed using a close button on the top right corner.

nimishavijay:
Yep, an undo button below the "Message sent". And we should automatically close the modal after a few seconds (however long the undo action is available -- around 5 seconds would work)

GretaD:
the second proposal will look like this

[...]

PR #6178, opened 2022-04-04
Add undo sending for new messages

GretaD:
To do:

- [x] Show the undo dialog that undo sending of a message
- [x] Hide the composer modal while the undo dialog pop us is shown
- [x] Clicking the undo should show the last state of the message before sending
- [x] Move modal composer into a vuex store to use it globally #6181 
- [x] Start a timer when the message is fake-sent and stop it if the user clicks *Undo*
- [x] Requires https://github.com/nextcloud/mail/pull/6216

Fixes https://github.com/nextcloud/mail/issues/5438


src/components/NewMessageModal.vue
```diff
@@ -33,6 +35,7 @@ export default {
 		return {
 			original: undefined,
 			originalBody: undefined,
+			hideComposer: false,
```
ChristophWurst, line 38: unused

ChristophWurst (approved):
Works for me both for new messages and the ones that are already in the outbox

src/components/NewMessageModal.vue
```diff
@@ -23,19 +24,17 @@ import logger from '../logger'
 import { toPlain } from '../util/text'
 import { saveDraft } from '../service/MessageService'
 import Composer from './Composer'
+import { showError, showUndo } from '@nextcloud/dialogs'
 import { translate as t } from '@nextcloud/l10n'
 
+const UNDO_DELAY = 7 * 1000
```
skjnldsv, line 30: TODO: could be exposed from the capabilities
Maybe a followup for cleaner code :)

ChristophWurst: what capabilities?

skjnldsv: That UNDO is supported and the timeout which the backend will use before sending the message?
Or is it being sen straight from the front only?

ChristophWurst: Straight outta frontend. The backend only sends the message if the tab/browser is closed

ChristophWurst: Nevertheless I forgot to add the delay to the message's `sendAt`, so the backend could have picked up the message before the elapsed undo timer :see_no_evil: 

Fixed.

JuliaKirschenheuter:
Test case:
1. Create  a new message
2. Send this message
3. Undo
4. Close a composer-dialog
5. Click to ... in a message-menu -> "Send now"
6. Expect message to be sent, but it will not be sent, outbox counter is still old one

ChristophWurst:
Right. That is broken.

ChristophWurst:
- [ ] BUG: somehow the message has an `aliasId` equal to the `accountId` and the lookup fails on the server, hence the message can't be sent. We have to debug how that happens.

ChristophWurst:
> somehow the message has an `aliasId` equal to the `accountId`

I wonder how that happened @miaulalala? https://github.com/nextcloud/mail/blob/8627531f617f2674f2ccd27c678aa62ba8d8a9c5/lib/Db/LocalMessage.php#L103 :see_no_evil:

JuliaKirschenheuter (approved):

PR #6186, opened 2022-04-06
Pipe all composer data through the Vuex store

ChristophWurst:
This is a follow-up to https://github.com/nextcloud/mail/pull/6181. This unifies the data handling of the composer to work for both IMAP and local (outbox) messages.

The logic of fetching an original message for the *Edit as new message* is now handled centrally by the store.

What I've tested
* Compose a new message
  + Draft is still saved
  + Message can be enqueued
* Handle a *mailto* link
  + To, cc and subject worked
* Open and edit an outbox message
  + Changes are sent to the server
  + Sending is borked but it was already before. I only added a comment to the relevant line.


src/components/NewMessageModal.vue
```diff
@@ -36,158 +33,117 @@ export default {
...
+					sendAt: Math.floor(now / 1000), // JS timestamp is in milliseconds
+					subject: data.subject,
+					body: data.isHtml ? data.body.value : toPlain(data.body).value,
+					isHtml: data.isHtml,
+					isMdn: false,
+					inReplyToMessageId: '',
+					to: data.to,
+					cc: data.cc,
+					bcc: data.bcc,
+					attachmentIds: [],
```
miaulalala, line 114: this should be attachments I think

src/components/NewMessageModal.vue
```diff
@@ -36,158 +33,117 @@ export default {
...
```
miaulalala, line 135: same here

src/components/NewMessageModal.vue
ChristophWurst: yes but no. The old ``\OCA\Mail\Controller\AccountsController::send`` took a nested array `attachments`. ``\OCA\Mail\Controller\OutboxController::create`` takes an int array `attachmentIds`. I don't know how to bridge those.

JuliaKirschenheuter (approved):

src/components/NewMessageModal.vue
```diff
@@ -36,158 +33,117 @@ export default {
...
+					})
+					data.draftId = this.draft.databaseId
+				}
+				const dataForServer = {
+					...data,
+					body: data.isHtml ? data.body.value : toPlain(data.body).value,
+				}
+				await this.$store.dispatch('outbox/updateMessage', { message: dataForServer, id: this.composerData.id })
+			} else {
+				if (data.draftId === undefined && this.draft) {
```
ChristophWurst, line 72: @GretaD @st3iny this code didn't make any sense but I just left it as is. There was no `this.draft` in the old component either. This will always be undefined.

PR #6207, opened 2022-04-07
Fix outbox API from/to/cc/bcc properties

ChristophWurst:
They are expected as four properties but were normalized into one array.

I missed this in https://github.com/nextcloud/mail/pull/6031.

On the target branch all outbox messages show a *?* for the avatar. With this patch it corresponds to the first recipient again.

Moreover if you continue editing an outbox message it previously lost all recipients.


miaulalala:
:see_no_evil:

miaulalala (approved):

Issue #6189, opened 2022-04-07
Outbox doesn't set \\answered to cross-account replies


ChristophWurst:
### Steps to reproduce

1. Configure two accounts
2. Open a message with account 1
3. Click reply
4. Select account 2 as sender
5. Send the email

### Expected behavior

The original email in account 1 gets marked as replied.

### Actual behavior

The original email remains unmarked.

### Mail app version

1.12

### Additional info

Regression of https://github.com/nextcloud/mail/pull/6031

ChristophWurst:
Please check if the flag is set at all @miaulalala. I saw that in Composer.vue::getMessageData we put the replied to message database ID into `messageId` and in NewMessageModal.vue::sendMessage `inReplyToMessageId` is always null.

miaulalala:
Database isn't marked, only IMAP:

miaulalala:
Right, tested this on a new branch off the feature/outbox branch. the reply isn't going through the outbox at all, and uses the old Accounts enpoint. Also fails to send as the Accounts endpoint expects a string for the recipient, but an array is provided. 

I am not sure what to do about this inconsistent state of editors and endpoints. I have added a PR to fix the "mark as \\answered" part of things on the DB level.

Re: message id vs database id: it is certainly possible to refactor to use the DB id instead of the message id for the outbox.

ChristophWurst:
As per yesterday's discussion with @jancborchardt I'll streamline the editors so everything goes through the same modal and the outbox.

> Ok, since for reply you can also not see the thread, it makes more sense to put it in a modal. And then since drafts would be the last embedded one, it would be better to bring it in line too with a modal. So all modals is then technically easier, and makes it look the same and more expected too.

miaulalala:
Let me know what you want to do about the databse id vs the mssage id

ChristophWurst:
We definitely still want to use the message ID to be independent of the cache

PR #6211, opened 2022-04-07
Mark database message as "\\answered"

miaulalala:
In addition to marking the IMAP message, also mark the database message entry.

Partial fix for #6189

### DIFF
--- a/lib/Listener/FlagRepliedMessageListener.php
+++ b/lib/Listener/FlagRepliedMessageListener.php
@@ -97,6 +97,9 @@ class FlagRepliedMessageListener implements IEventListener {
 						'exception' => $e,
 					]);
 				}
+
+				$message->setFlagAnswered(true);
+				$this->dbMessageMapper->update($message);
 			}
 		} finally {
 			$client->logout();


lib/Listener/FlagRepliedMessageListener.php
```diff
@@ -97,6 +97,9 @@ public function handle(Event $event): void {
 						'exception' => $e,
 					]);
 				}
+
+				$message->setFlagAnswered(true);
```
ChristophWurst, line 101: :warning: you are outside the loop. `$message` is undefined if you have 0 messages or you access the *last* one in the `$messages` array.

lib/Listener/FlagRepliedMessageListener.php
```diff
@@ -97,6 +97,9 @@ public function handle(Event $event): void {
...
+				$this->dbMessageMapper->update($message);
 			}
```
miaulalala, line 103: that's the loop end bracket?

lib/Listener/FlagRepliedMessageListener.php
miaulalala: Line 103 is the loop end bracket tho?

ChristophWurst: Bildschirmfoto vom 2022-04-08 17-55-03

stupid wonky github indentation :see_no_evil:

ChristophWurst (approved):

PR #6214, opened 2022-04-08
Updating appstore-build-publish.yml workflow from template

nextcloud-command:
Automated update of the appstore-build-publish.yml workflow from https://github.com/nextcloud/.github


.github/workflows/appstore-build-publish.yml
```diff
@@ -64,7 +69,6 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ env.PHP_VERSION }}
-          tools: composer:v1
```
ChristophWurst, line 67: this will break the release

ChristophWurst (changes requested):

PR #6233, opened 2022-04-13
Fix sending message workflow

JuliaKirschenheuter:
Undo sending will appear if "send now" will be chosen

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -117,10 +117,10 @@ export default {
 				attachments: data.attachments,
 				aliasId: data.aliasId,
 				inReplyToMessageId: null,
-				sendAt: data.sendAt ? data.sendAt : Math.floor(now / 1000),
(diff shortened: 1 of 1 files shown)


src/components/NewMessageModal.vue
```diff
@@ -135,7 +135,7 @@ export default {
 				})
 			}
 
-			if (!data.sendAt) {
+			if (!data.sendAt || data.sendAt < Math.floor(now / 1000)) {
```
miaulalala, line 138: I think this should be a comparison with the UNDO_SEND constant, no? So 30 seconds or whatever?

JuliaKirschenheuter: Why do you think so?i don't think so. Here i compare scheduled time (timestamp) from store with timestamp of now. And if it is somehow scheduled, "undo sending" will not appear/

miaulalala (approved):

PR #6301, opened 2022-04-23
Avoid calling Mailbox.vue::loadEnvelopes twice when switching between two accounts

### LINKED ISSUE #3746: Switching from one mailbox to another triggers two get requests to the /apps/mail/api/messages endpoint
### Expected behavior
When you switch from one mailbox to another, then only one get requests to the /apps/mail/api/messages endpoint is triggered

### Actual behavior
When you switch from one mailbox to another, then two get requests to the /apps/mail/api/messages endpoint are triggered

Mail app
Mail app version: v1.5.0

Mailserver or service: Strato

Server configuration
Operating system: Ubuntu 20.04

Web server: Apache

Database: MariaDB

PHP version: 7.4

Nextcloud Version: 20.0.0

Client configuration
Browser: Chrome 81

Operating system: Ubuntu 20.04

IchbinkeinReh:
Fix #3746

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -60,6 +60,7 @@ import EmptyMailboxSection from './EmptyMailboxSection'
 import { showError } from '@nextcloud/dialogs'
 import NoTrashMailboxConfiguredError
 	from '../errors/NoTrashMailboxConfiguredError'
+import _ from 'lodash'
 
 export default {
 	name: 'Mailbox',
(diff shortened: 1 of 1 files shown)


ChristophWurst:
We likely trigger that method from one of the three watchers, or the `mounted` hook. Could you add some logging to find out?

Let's try to find out why there is a double-trigger and fix the issue there :)

IchbinkeinReh:
Just had a look at this old and still existing issue. Found out that one load comes from https://github.com/nextcloud/mail/blob/main/src/components/Mailbox.vue#L167 and the other from https://github.com/nextcloud/mail/blob/main/src/components/Mailbox.vue#L170

GretaD:
done here https://github.com/nextcloud/mail/pull/8944

PR #6302, opened 2022-04-25
Fix discard message on outbox and improve the discard new message

### LINKED ISSUE #6268: Move modal when discarding a message into a toast

GretaD:
fixes #6268

needs #6180

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1001,9 +1001,10 @@ export default {
 
 			return `${alias.name} <${alias.emailAddress}>`
 		},
-		discardDraft() {
+		async discardDraft() {
 			console.debug('discaaaaard')
-			this.$emit('discardDraft')
+			const id = await this.draftsPromise
+			this.$emit('discardDraft', id)
 		},
 		/**
 		 * Whether the date is acceptable
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -302,10 +302,6 @@
 		:hint="t('mail', 'Sending …')"
 		role="alert"
 		class="sending-hint" />
-	<Loading v-else-if="state === STATES.DISCARDING" :hint="t('mail', 'Discarding …')" class="emptycontent" />
-	<EmptyContent v-else-if="state === STATES.DISCARDED" icon="icon-mail">
-		<h2>{{ t('mail', 'Draft was discarded!') }}</h2>
-	</EmptyContent>
 	<div v-else-if="state === STATES.ERROR" class="emptycontent" role="alert">
 		<h2>{{ t('mail', 'Error sending your message') }}</h2>
 		<p v-if="errorText">
@@ -350,7 +346,7 @@ import ActionLink from '@nextcloud/vue/dist/Components/ActionLink'
 import ActionRadio from '@nextcloud/vue/dist/Components/ActionRadio'
 import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
 import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
-import { showError } from '@nextcloud/dialogs'
+import { showError, showSuccess } from '@nextcloud/dialogs'
 import { translate as t, getCanonicalLocale, getFirstDay, getLocale } from '@nextcloud/l10n'
 import Vue from 'vue'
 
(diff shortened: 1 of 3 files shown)


src/components/Composer.vue
```diff
@@ -1008,10 +1002,9 @@ export default {
 			return `${alias.name} <${alias.emailAddress}>`
 		},
 		async discardDraft() {
-			this.state = STATES.DISCARDING
 			const id = await this.draftsPromise
 			await this.$store.dispatch('deleteMessage', { id })
```
st3iny, line 1006: Please add a try catch around this dispatch and call `showError(...)` in the catch block. Otherwise, the user won't be notified in case of failure.

st3iny (commented):
I just noticed that discarding messages in outbox is broken. Clicking on the discard button won't do anything. This is also broken on master, only that the modal will load indefinitely over there.

image

Peek 2022-04-26 11-50

src/components/Composer.vue
GretaD: yep. fixed the try..catch, i tested it with new message and it works, but discarding while in an outbox message, it doesnt work. Will fix it in this pr

JuliaKirschenheuter:
I have tested a behavior with discarding. It is a bit oddly, that i can click 2+ times on the "Discard draft". As a conclusion the request to delete a message ("DELETE") will be sent 2+ times with a same data. 
image

And it will cause 2+ toasters:

image

There is no timeout. But deleting takes some time. 
Well, i guess it would be an UX improvement, if this 2 lines will be moved on the top of `discardDraft()`

`showSuccess(t('mail', 'Message discarded'))`
`this.$emit('close')`

src/components/Composer.vue
```diff
@@ -1008,11 +1002,9 @@ export default {
 			return `${alias.name} <${alias.emailAddress}>`
 		},
 		async discardDraft() {
-			this.state = STATES.DISCARDING
+			console.debug('discaaaaard')
 			const id = await this.draftsPromise
-			await this.$store.dispatch('deleteMessage', { id })
-			this.state = STATES.DISCARDED
-			this.$emit('close')
+			this.$emit('discardDraft', id)
```
st3iny, line 1007:
```suggestion
			this.$emit('discard-draft', id)
```

src/components/NewMessageModal.vue
```diff
@@ -17,6 +17,7 @@
 			:draft="saveDraft"
 			:send="sendMessage"
 			:forwarded-messages="forwardedMessages"
+			@discardDraft="discardDraft"
```
st3iny, line 20:
```suggestion
			@discard-draft="discardDraft"
```

src/components/Composer.vue
GretaD: actually, i used to have it like that and the button was not interacting, then Christoph sent me this: https://v2.vuejs.org/v2/guide/components-custom-events.html#Event-Names

and according to that we must write it @ discardDraft

st3iny: But the page you linked says exactly that:

> Unlike components and props, event names don’t provide any automatic case transformation. Instead, the name of an emitted event must exactly match the name used to listen to that event.

src/components/NewMessageModal.vue
```diff
@@ -186,6 +188,25 @@ export default {
 				return `"${recipient.label}" <${recipient.email}>`
 			}
 		},
+		async discardDraft(id) {
+			console.debug('discarding working?', id)
```
st3iny, line 192: :boom:

src/components/Composer.vue
```diff
@@ -1008,11 +1002,9 @@ export default {
 			return `${alias.name} <${alias.emailAddress}>`
 		},
 		async discardDraft() {
-			this.state = STATES.DISCARDING
+			console.debug('discaaaaard')
```
st3iny, line 1005: :boom:

src/components/Composer.vue
GretaD: ah, this yes :) sorry

GretaD:
closing on behalf #6338

PR #6310, opened 2022-04-26
Modify attachements view at thread envelope

sazanof:
Hello! I don't know whether to open a PR or ISSUE, but as a suggestion, I decided to experiment a little with the visual arrangement of attachments in the message.
I don't know if my PR will be useful, but if there is an opportunity, please read it.
attachements-env-thread

I'll try to attach a GIF to fully show the logic of the work. In this drop-down menu, it may be worth moving the preview as well

Peek 2022-04-26 19-33

As an option - it would be possible to use vertical scrolling of two rows ...
Thank you!

Fixes https://github.com/nextcloud/mail/issues/7120

### DIFF
--- a/src/components/MessageAttachment.vue
+++ b/src/components/MessageAttachment.vue
@@ -21,14 +21,18 @@
 
 <template>
 	<div class="attachment" :class="{'message-attachment--can-preview': canPreview }" @click="$emit('click', $event)">
-		<img v-if="isImage"
-			class="mail-attached-image"
-			:src="url">
-		<img class="attachment-icon" :src="mimeUrl">
-		<span class="attachment-name"
-			:title="label">{{ name }}
-			<span class="attachment-size">({{ humanReadable(size) }})</span>
-		</span>
+		<div class="mail-attachment-img--wrapper">
+			<img v-if="isImage"
+				class="mail-attached-image"
+				:src="url">
+			<img v-else class="attachment-icon" :src="mimeUrl">
+		</div>
+		<div class="mail-attached--content">
+			<span class="attachment-name"
+				:title="label">{{ name }}
+			</span>
+			<span class="attachment-size">{{ humanReadable(size) }}</span>
+		</div>
 		<Actions :boundaries-element="boundariesElement">
 			<ActionButton
 				v-if="isCalendarEvent"
@@ -233,11 +237,21 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+
 .attachment {
-	position: relative;
-	display: inline-block;
-	width: calc(100% - 32px);
-	padding: 16px;
+	height: auto;
+    display: inline-flex;
+    flex-wrap: wrap;
+    justify-content: space-between;
+    width: calc(33.3334% - 4px);
+    margin: 2px;
+	padding: 5px;
+    position: relative;
+    align-items: center;
+
+	&:hover {
+		border-radius: 6px;
+	}
 }
 
 .attachment:hover,
@@ -247,6 +261,33 @@ export default {
 	&.message-attachment--can-preview * {
 		cursor: pointer;
 	}
+}
+
+.mail-attachment-img--wrapper {
+	height: 44px;
+	width: 44px;
+	overflow: hidden;
+	display:flex;
+	justify-content: center;
+	position: relative;
+	border-radius: 6px;
+
+	img {
+		transition: 0.3s;
+		opacity: 1;
+		width: 44px;
+		height: 44px;
+	}
+
+	.mail-attached-image {
+		width: 100px;
+	}
+}
+
+.mail-attached--content {
+	width: calc(100% - 100px);
+	display: flex;
+    flex-direction: column;
 }
 
 .mail-attached-image {
@@ -263,18 +304,19 @@ export default {
 }
 .attachment-name {
 	display: inline-block;
-	width: calc(100% - 72px);
+	width: 100%;
 	white-space: nowrap;
 	overflow: hidden;
 	text-overflow: ellipsis;
 	vertical-align: middle;
-	margin-bottom: 20px;
 }
 
 /* show attachment size less prominent */
 .attachment-size {
 	-ms-filter: 'progid:DXImageTransform.Microsoft.Alpha(Opacity=50)';
 	opacity: 0.5;
+	font-size: 12px;
+	line-height: 14px;
 }
 
 .attachment-icon {
@@ -283,8 +325,7 @@ export default {
 	margin-bottom: 20px;
 }
 .action-item {
-	display: inline-block !important;
-	position: relative !important;
+	transition: 0.4s;
 }
 .mail-message-attachments {
 	overflow-x: auto;
--- a/src/components/MessageAttachments.vue
+++ b/src/components/MessageAttachments.vue
@@ -20,52 +20,57 @@
   -->
 
 <template>
-	<div class="mail-message-attachments">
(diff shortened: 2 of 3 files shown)


ChristophWurst:
@jancborchardt @nimishavijay please have a look. We can create a ticket if this needs any kind of discussion :)

jancborchardt:
@sazanof very cool! Did you see the mockup by @nimishavijay at https://github.com/nextcloud/mail/issues/5768#issuecomment-990920989 about attachments when writing a message?

We could use the same style here, and yours is mostly there, I’m mainly seeing 2 things:
- The menu icon should be on the right and displayed for all items, not only on hover (so it is discoverable and also works on mobile)
- Might be good to collapse it at a certain point if there are a lot of attachments

sazanof:
@jancborchardt, @ChristophWurst  hello! Yes, I saw this comment. I really liked the implementation. I came across it after (=)) how I implemented this functionality. So far I haven't figured out how to render previews. And so, I have already opened a branch in parallel to implement exactly the idea you referred to. However, there will be much more changes than those that I made earlier.

Perhaps as a temporary measure.

Why did I move the additional actions button to the icon - we mainly use laptops 1920x1080 15.6" 
It seemed that there was not quite enough space both in the modal onk when creating a message and when viewing it. I decided to make it more compact. But here the user may suffer, he will not immediately understand where the button is =)

> Might be good to collapse it at a certain point if there are a lot of attachments

maybe. I just looked at roundcube and a couple more email clients and made a display of all attachments at once.

sazanof:
Do you think it might be worth moving the block with attachments above the body of the letter?

sazanof:
I have received a lot of feedback from users that attachments on top are more convenient. I also looked at how it is implemented by other clients. In the vast majority - attachments are located on top. Therefore, I suggest you consider this revision. Perhaps, in terms of convenience, this is indeed the case. Thanks!

image
multiline

image
_oneline_

st3iny:
@jancborchard @nimishavijay Could we get another round of feedback for this (especially regarding showing on top vs bottom)? Both bullet points from the last review have been implemented.

sazanof:
@st3iny thank you for you review. I corrected the attachment padding and `click` on the image.

st3iny (approved):
Tested and works. My previous feedback was addressed.

I'd suggest to keep the attachments on the top and open an issue for discussion. This can be changed easily later.

st3iny:
Please squash your commits into one.

nimishavijay:
It actually looks really good! The design itself looks great, only one question: Is the name of the file in bold? We should change it to regular.

I would actually expect the attachments to be below the body text. Gmail and Apple Mail show it below. Outlook shows it above. Do you know what other clients show it above?
I agree that in case there a lot of quoted text the attachments are easily accessible from above, but I think showing it above introduces additional complexity, for eg. requiring a collapse/expand action in case are there many attachments. I would suggest we show it below, always expanded to show all the attachments :)

JuliaKirschenheuter (approved):
The attachments on the top looks great, and works well (on mobile too)

nimishavijay:
Thanks for the examples! I'm leaning towards showing it at the bottom because we try to avoid expanding/collapsing interactions, but since there are many clients which show on top this sounds alright as well :)
If we are going with showing at the top, we need a better indication of if there are more attachments, right now it's not so clear that there are 7 attachments because only 3 are shown by default. I would suggest having another smaller row at the bottom which says "View 4 more attachments" or something similar, and clicking on it would show the rest of the attachments. 

cc @jancborchardt about the placement

jancborchardt:
Agree with @nimishavijay (and the initial design) that attachments should be shown on the bottom. :)
- The text of the message is most likely more important than the attachments
- We already have an attachment indicator icon in the message list to communicate there are attachments
- Even if it's shown at the bottom and the message exceeds the viewport, they can be sticky at the bottom of the viewport for visibility
- Often there are useless or technical attachments which are visualized or used otherwise, like a PGP signature, or ics file which should be shown via iMIP, etc
- Sidenote: The most popular and well-designed mail apps (Google Mail and Apple Mail) show it at the bottom as well. So it's not about number of apps who dilo it, but more number of users / popularity. :)

Hope that clarifies it

st3iny (changes requested):
Don't merge yet due to what Jan and Nimisha said.

jancborchardt:
Yes @sazanof, beautiful! :)

src/components/MessageAttachments.vue
```diff
@@ -20,37 +20,52 @@
...
+					@close="showPreview = false" />
+			</div>
+		</div>
+		<div v-if="hasNextLine"
+			class="show-more-attachments"
+			@click="isToggled = !isToggled">
+			<ChevronDown v-if="isToggled" :size="24" />
+			<ChevronUp v-if="!isToggled" :size="24" />
+			<span v-if="isToggled">
+				{{ n('mail', 'View more {count} attachment', 'View more {count} attachments', (attachments.length - visible), { count: attachments.length - visible }) }}
```
ChristophWurst, line 50:
```suggestion
				{{ n('mail', 'View {count} more attachment', 'View {count} more attachments', (attachments.length - visible), { count: attachments.length - visible }) }}
```

Or simpler: *View all attachments*

src/components/MessageAttachment.vue
```diff
@@ -209,17 +213,49 @@ export default {
...
+
+	.mail-attached-image {
+		width: 100px;
+	}
 }
 
-.attachment:hover,
-.attachment span:hover {
-	background-color: var(--color-background-hover);
-	cursor: pointer;
```
ChristophWurst, line 222: this got lost. images should appear clickable if clicking triggers a preview. this will conflict with https://github.com/nextcloud/mail/pull/7066

src/components/MessageAttachments.vue
```diff
@@ -20,37 +20,52 @@
...
+			</span>
+			<span v-else>
+				{{ t('mail', 'View less attachments') }}
```
ChristophWurst, line 53:
```suggestion
				{{ t('mail', 'View fewer attachments') }}
```

*Less* is for anything you can't count, e.g. water
*Fewer* is for anything you can count, e.g. pineapples

sazanof: Fixed

src/components/MessageAttachment.vue
sazanof: Restored.

_I didn't quite understand what exactly was lost. I didn't see the difference. Unless there was no style for `.attachment span:hover.`_

src/components/MessageAttachments.vue
sazanof: Fixed, thanks!

ChristophWurst (approved):
Code and UI look good

This will likely conflict with https://github.com/nextcloud/mail/pull/7066, which is planned to be merged today. I'll see if I can resolve the conflicts later on.

sazanof:
> This will likely conflict with #7066, which is planned to be merged today. I'll see if I can resolve the co

@ChristophWurst 
Do I understand correctly that today "this" is normal behavior when clicking on an image?  I updated main, made `npm i ` && `npm run watch` and this error came out (I just couldn't understand why I couldn't fix the conflicts)
image

ChristophWurst:
It is not expected. @GretaD ran into the same bug. Could you possibly provide the info from https://github.com/nextcloud/mail/pull/7066#issuecomment-1234303810?

ChristophWurst (approved):
Tested & works

PR #6314, opened 2022-04-27
Update styles (sender position, margins, etc.) for envelope in thread

sazanof:
Hello. Slightly corrected the styles. You can familiarize yourself with PR, in case it turns out to be useful. Thanks!
some-style-mod

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -329,7 +329,7 @@ export default {
 }
 
 #mail-content {
-	margin: 10px 38px 0 59px;
+	margin: 10px 38px 0 63px;
 }
 
 #mail-content iframe {
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -298,7 +298,7 @@ export default {
 #mail-thread-header-fields {
 	// initial width
 	width: 0;
-	padding-left: 70px;
+	padding-left: 16px;
 	// grow and try to fill 100%
 	flex: 1 1 auto;
 	h2,
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -29,7 +29,7 @@
 				:email="envelope.from[0].email"
 				:display-name="envelope.from[0].label"
 				:disable-tooltip="true"
-				:size="40" />
+				:size="44" />
 			<div
 				v-if="isImportant"
 				class="app-content-list-item-star icon-important"
@@ -441,6 +441,7 @@ export default {
 <style lang="scss" scoped>
 	.sender {
 		margin-left: 8px;
+		font-weight: bold;
 	}
 
 	.right {
@@ -591,7 +592,7 @@ export default {
 		left: 4px;
 	}
 	.envelope--header.list-item-style {
-		border-radius: 16px;
+		border-radius: 16px 16px 0 0;
 	}
 	.junk-favorite-position-with-tag-subline {
 		margin-bottom: 14px !important;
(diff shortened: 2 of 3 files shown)


ChristophWurst:
Could you rebase and post before/after screenshots? I'm not able to see the difference from the two posted screenshots

There is a small conflict as the `hasChangedSubject` was replaced with the message preview.

sazanof:
> Could you rebase and post before/after screenshots? I'm not able to see the difference from the two posted screenshots
> 
> There is a small conflict as the `hasChangedSubject` was replaced with the message preview.

Thanks for review!

Just small changes:

_before_
image

_after_
image

ChristophWurst:
> Just small changes:

@jancborchardt @nimishavijay wdyt?

sazanof:
BEFORE
image

AFTER
image

src/components/ThreadEnvelope.vue
```diff
@@ -29,7 +29,7 @@
 				:email="envelope.from[0].email"
 				:display-name="envelope.from[0].label"
 				:disable-tooltip="true"
-				:size="40" />
+				:size="44" />
```
jancborchardt, line 32: Note that increasing the avatar size also needs more text indent of the mail text, so that sender name and mail body is aligned. :)

src/components/Thread.vue
```diff
@@ -298,7 +298,7 @@ export default {
 #mail-thread-header-fields {
 	// initial width
 	width: 0;
-	padding-left: 70px;
+	padding-left: 16px;
```
jancborchardt, line 301: I'd say I prefer the indented header, but I don't have a strong opinion. :) @ChristophWurst @nimishavijay @GretaD any preference?

src/components/ThreadEnvelope.vue
```diff
@@ -441,6 +441,7 @@ export default {
 <style lang="scss" scoped>
 	.sender {
 		margin-left: 8px;
+		font-weight: bold;
```
jancborchardt, line 444: Sure let's go for it! Just a detail, code indent seems wrong here?

src/components/ThreadEnvelope.vue
```diff
@@ -591,7 +592,7 @@ export default {
 		left: 4px;
 	}
 	.envelope--header.list-item-style {
-		border-radius: 16px;
+		border-radius: 16px 16px 0 0;
```
jancborchardt, line 595: Doesn't this look a bit off when hovering/focusing the heading? Or you prefer it like that?

jancborchardt (commented):
Some comments :)

src/components/ThreadEnvelope.vue
sazanof: Yes, earlier, when the NC Mail design was different, this fix would have been useful. But now there is no background in this element. I'll clean it up.

src/components/ThreadEnvelope.vue
sazanof: image

Like this?

src/components/ThreadEnvelope.vue
sazanof: Linter doesn't swear

src/components/Thread.vue
ChristophWurst: Indented also looks better to me

src/components/ThreadEnvelope.vue
ChristophWurst: We use bold for unread messages. I think this style conflicts.

ChristophWurst:
Closing because the three proposed changes don't seem to fit our design

Thanks for the ideas @sazanof

Issue #5136, opened 2021-06-07
Migrate from composer v1 to composer v2


LukasReschke:
On current master (e7b92945281eaeab9d9d9d1e60a19d1fda359183):

```
lukasreschke@Lukass-Air mail % make dev-setup
composer install -o

                                                          
  [RuntimeException]                                      
  The PEAR repository has been removed from Composer 2.0  
                                                          

make: *** [install-composer-deps-dev] Error 1
```
---

Migrate Mail from composer v1 to composer v2, because v1 has been deprecated and will stop working soon.

- [ ] Help migrate the Horde packages from PEAR to packagist upstream
- [ ] Check if any other packages require migration

ChristophWurst:
yeah … you need composer v1. There are some attempts at the Horde project to bring their packages over but I think those stalled. There is nothing we can do here at the moment :/

szaimen:
I have the same problem

ChristophWurst:
tracked in https://github.com/orgs/nextcloud/projects/45#card-67313209 and will be taken in consideration for the next planning

HLFH:
Link to composer v1: https://getcomposer.org/download/1.10.22/composer.phar

ChristophWurst:
Ref https://github.com/composer/composer/issues/8778#issuecomment-717050280

szaimen:
🎉

PR #6317, opened 2022-04-27
Migrate PEAR Horde packages to Bytestream packages

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/5136

Once the Horde packages are migrated to packagist we can jump back to the "original". Let's send patches to both versions for now.

To do
- [x] https://github.com/nextcloud/mail/pull/6342

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report_form.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report_form.yml
@@ -56,7 +56,6 @@ body:
           Select PHP engine version serving Nextcloud Server.
           _Describe in the "Additional info" section if you chose "Other"._
       options:
-          - "PHP 7.3"
           - "PHP 7.4"
           - "PHP 8.0"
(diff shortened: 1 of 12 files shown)


ChristophWurst:
@bytestream thanks _a lot_ for your packages. The migration experience was super smooth. I'll send patches for the deprecation notices soon :v:

st3iny (approved):
Looks good and works so far. When can we release/merge this? Did NC drop PHP 7.3 already?

ChristophWurst:
> Did NC drop PHP 7.3 already?

Only with Nextcloud 24+

bytestream:
For what it's worth, I don't _think_ I actually introduced any PHP 7.4 specific code. You could send PR to https://packagist.org/?query=bytestream%20horde which loosens the PHP requirement to `^7.3 || ^8.0` and updates the GitHub workflow

PR #6318, opened 2022-04-27
Speed up bulk inserts

ChristophWurst:
Doctrine has to convert parameters if they are an array or the
parameters are named. If we replace the key with a position then the
bulk insert speeds up quite a bit.

Todo
- [ ] Use constants instead of numbers
- [ ] Apply the same trick for the bulk update

Bildschirmfoto vom 2022-04-27 20-21-52


ChristophWurst:
This feels a bit too fragile. Let's leave this aside for now.

ChristophWurst:
https://github.com/nextcloud/mail/issues/6476

Issue #6308, opened 2022-04-26
Prevent provisioned account updates that have no effect


ChristophWurst:
### Is your feature request related to a problem? Please describe.

It looks like we often update provisoined Mail accounts in the before controller middleware hook even if there are no changes. This slows down every Mail request a tiny bit.

![Bildschirmfoto vom 2022-04-26 18-20-17](https://user-images.githubusercontent.com/1374172/165346982-71d3edaf-c315-4f64-b9d1-a251a87e1b4f.png)

SQL: ``update oc_mail_accounts set provisioning_id = ? where id = ?``

### Describe the solution you'd like

Only update the account if necessary.

PR #6324, opened 2022-04-27
 Check account for currentness before reprovisioning it

miaulalala:
If it's ugly but it works is it still ugly? Strings are interned in PHP so the string comparison should be fast,

To Do:

- [ ] Run a performance evaluation against it

### DIFF
--- a/lib/Service/Provisioning/Manager.php
+++ b/lib/Service/Provisioning/Manager.php
@@ -174,12 +174,12 @@ class Manager {
 		}
 
 		try {
-			// TODO: match by UID only, catch multiple objects returned below and delete all those accounts
 			$mailAccount = $this->mailAccountMapper->findProvisionedAccount($user);
-
-			$mailAccount = $this->mailAccountMapper->update(
-				$this->updateAccount($user, $mailAccount, $provisioning)
-			);
+			if (!$this->isUpToDate($user, $mailAccount, $provisioning)) {
+				$mailAccount = $this->mailAccountMapper->update(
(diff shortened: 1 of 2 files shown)


lib/Service/Provisioning/Manager.php
```diff
@@ -274,6 +274,21 @@ private function updateAccount(IUser $user, MailAccount $account, Provisioning $
 		return $account;
 	}
 
+	private function isUpToDate(IUser $user, MailAccount $account, Provisioning $provisioning): bool {
```
ChristophWurst, line 277: instead of a custom check, how about we use `\OCP\AppFramework\Db\Entity::getUpdatedFields` after the account was updated with the new/current values. If all properties have the same value as before, the mapper will not have to update anything.

ChristophWurst (changes requested):

lib/Service/Provisioning/Manager.php
ChristophWurst: I'm actually thinking that I had that in mind. If the account has no changes, the mapper does a noop. Could you debug why this isn't the case anymore? From the blackfire.io session I remember that the UPDATE query assigned a new provisioning_id, right?

ChristophWurst: *Your* bug https://github.com/nextcloud/mail/blame/27fe2975aac226aad344d7a54b0858d136565d91/lib/Db/MailAccount.php#L199

kesselb: Ah. The good old snake_case / camelCase mixup ;)

Issue #4983, opened 2021-04-26
Cache provisioning configs


miaulalala:
### Feature Request

Provisioning middleware is called each time an account interacts with the mail app. This means, every time there is the possibility of a DB query.

#### Summary

Cache the provisioning config to avoid DB reads.

PR #6325, opened 2022-04-27
Cache provisioning configs

miaulalala:
Fixes https://github.com/nextcloud/mail/issues/4983


ChristophWurst:
Does this actually work?

1) ICache is not injectable. You have to use the factory to acquire a local or distributed cache
2) ICache::get gives you a string, not an array of deserialized objects

ChristophWurst (approved):
Tested, reviewed and profiled

Bildschirmfoto vom 2022-04-29 15-44-10
Bildschirmfoto vom 2022-04-29 15-44-56

ChristophWurst:
@miaulalala please make your code compatible with PHP7.3. We will migrate the code to PHP7.4 once 7.3 is gone

PR #6338, opened 2022-05-01
Fix discard message on outbox and improve the discard new message

GretaD:
Fixes https://github.com/nextcloud/mail/issues/6268

just like #6302 but without a failed rebase

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -317,10 +317,6 @@
 		:hint="t('mail', 'Sending …')"
 		role="alert"
 		class="sending-hint" />
-	<Loading v-else-if="state === STATES.DISCARDING" :hint="t('mail', 'Discarding …')" class="emptycontent" />
-	<EmptyContent v-else-if="state === STATES.DISCARDED" icon="icon-mail">
-		<h2>{{ t('mail', 'Draft was discarded!') }}</h2>
-	</EmptyContent>
 	<div v-else-if="state === STATES.ERROR" class="emptycontent" role="alert">
 		<h2>{{ t('mail', 'Error sending your message') }}</h2>
 		<p v-if="errorText">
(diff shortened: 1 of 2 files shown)


src/components/NewMessageModal.vue
```diff
@@ -187,6 +188,25 @@ export default {
...
+			const isOutbox = this.composerMessage.type === 'outbox'
+			if (isOutbox) {
+				id = this.composerMessage.data.id
+			}
+			this.$emit('close')
+			try {
+				if (isOutbox) {
+					await this.$store.dispatch('outbox/deleteMessage', { id })
+				} else {
+					await this.$store.dispatch('deleteMessage', { id })
```
ChristophWurst, line 202: id is always undefind

GretaD: Im not sure what am i supposed to do

src/components/Composer.vue
```diff
@@ -1041,16 +1037,8 @@ export default {
...
-			try {
-				id = await this.draftsPromise
-				await this.$store.dispatch('deleteMessage', { id })
-			} catch (err) {
-				logger.error('Could not delete message with id ' + id)
-			}
-			this.state = STATES.DISCARDED
-			this.$emit('close')
+			const id = await this.draftsPromise
+			this.$emit('discard-draft', id)
```
ChristophWurst, line 1041: here you emit an id

src/components/NewMessageModal.vue
```diff
@@ -187,6 +188,25 @@ export default {
 				return `"${recipient.label}" <${recipient.email}>`
 			}
 		},
+		async discardDraft() {
```
ChristophWurst, line 191: but you don't use the argument value

ChristophWurst (approved):
Works

PR #6345, opened 2022-05-02
Fix design for bottom action on composer

GretaD:
- [x] Aligned "discard draft" with the starting of the text
- [x] All the bottom actions should be sticky
boto-action


GretaD:
css needs some refactoring but for now i guess its fine

JuliaKirschenheuter (approved):

PR #6346, opened 2022-05-02
Fix position of composer actions

JuliaKirschenheuter:
Fix a regression with composer actions


GretaD:
duplication of https://github.com/nextcloud/mail/pull/6345

PR #6356, opened 2022-05-03
Show account color indicator only if several mailboxId's are present

### LINKED ISSUE #6191: Account color indicator shows in priority inbox even if only 1 account is configured
### Steps to reproduce

1. Have only 1 account in Mail
2. Check the Priority inbox
3. Notice a color line on the left (and the color line also shows for the "load more" action entry which seems off)

### Expected behavior

No color line on the left since it’s only 1 account.

### Actual behavior

This color line is shown on the left:
image

cc @ChristophWurst @karlitschek as we saw in the call. :)

### Mail app version

Current on cloud.nc.com

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

JuliaKirschenheuter:
fixes #6191

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -16,7 +16,7 @@
 		@click="onClick">
 		<template #icon>
 			<div
-				v-if="mailbox.isUnified"
+				v-if="mailbox.isUnified && hasMultipleAccounts"
 				class="mail-message-account-color"
 				:style="{'background-color': accountColor}" />
 			<div
@@ -145,7 +145,7 @@
 		</template>
 		<template #extra>
 			<div
-				v-if="mailbox.isUnified"
+				v-if="mailbox.isUnified && hasMultipleAccounts"
 				class="mail-message-account-color"
 				:style="{'background-color': accountColor}" />
 			<div v-for="tag in tags"
@@ -235,6 +235,10 @@ export default {
 			type: Array,
 			required: false,
 			default: () => [],
+		},
+		hasMultipleAccounts: {
+			type: Boolean,
+			default: false,
 		},
 	},
 	data() {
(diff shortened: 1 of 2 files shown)


src/components/Envelope.vue
```diff
@@ -16,7 +16,7 @@
 		@click="onClick">
 		<template #icon>
 			<div
-				v-if="mailbox.isUnified"
```
ChristophWurst, line 19: now it will also show in individual mailboxes like *Draft* and *Sent*. This should only be there in any *unified* mailbox like *All inboxes*, *Priority inbox* and *Outbox*.

src/components/Envelope.vue
```diff
@@ -236,6 +236,10 @@ export default {
 			required: false,
 			default: () => [],
 		},
+		isSeveralMailboxIdsPresent: {
```
ChristophWurst, line 239: `hasMultipleAccounts`?

ChristophWurst (changes requested):

src/components/EnvelopeList.vue
```diff
@@ -272,6 +273,10 @@ export default {
 		selectedEnvelopes() {
 			return this.envelopes.filter((env) => this.selection.includes(env.databaseId))
 		},
+		hasMultipleAccounts() {
+			const mailboxIds = this.envelopes.map(envelope => envelope.mailboxId)
+			return Array.from(new Set(mailboxIds)).length > 1
```
ChristophWurst, line 278: It would hurt to comment such code. I get that the conversion from array to set to array is to remove duplicates but it's not obvious :)

ChristophWurst (approved):
Looks good

PR #6369, opened 2022-05-05
Fix modal size in different size of the screen

### LINKED ISSUE #6273: Modal composer buttons on bottom are not sticky to bottom on 4k screen

GretaD:
fixes #6273 

width: 80% its overwritten by modal.vue style, 
min-height:60% is causing the modal to extend if the screen is too high, but doesnt do anything else

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -193,10 +193,6 @@ export default {
 		max-width: 80%;
 	}
 }
-::v-deep .modal-container {
-	width: 80%;
-	min-height: 60%;
-}
 ::v-deep .modal-wrapper .modal-container {
 	overflow-y: auto !important;
 	overflow-x: auto !important;


GretaD:
@miaulalala  please test it with the same screen where you could reproduce it <3

JuliaKirschenheuter (approved):

miaulalala (approved):
Looks good! Tested it and the modal is now neat and tidy.

szaimen:
> sure, is it breaking anything, or is it deadcode?

It should not break anything and make the modal size work better on mobile :)

szaimen:
Shall I open a PR for this?

ChristophWurst:
> @ChristophWurst why did you tag 1.13 instead of 1.12.1?

Because this PR and its commit(s) only go into 1.13. You assigned the wrong milestone: https://github.com/nextcloud/mail/pull/6369#event-6557343609

PR #6372, opened 2022-05-06
Fix automatic email signature on the composer

### LINKED ISSUE #6361: The new compose modal does not seem to add the email signature
### Steps to reproduce

1. Set a mail signature in the mail account settings
2. Click on "new message" or on "reply" in a received mail

### Expected behavior

The signature should be shown at the bottom of the message.

### Actual behavior

No signature appear. **However**, selecting another account (if you have multiple set) will change the signature to the one of the corresponding account. Switching back to the original account also put the correct signature.

### Mail app version

1.12.0

### Mailserver or service

Not relevant

### Operating system

Fedora

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

JuliaKirschenheuter:
Fixes #6361

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -183,7 +183,7 @@
 				:placeholder="t('mail', 'Write message …')"
 				:focus="isReply"
 				:bus="bus"
-				@input="callSaveDraft(true, getMessageData)" />
+				@input="onEditorInputText" />
 			<TextEditor
 				v-else-if="!encrypt && !editorPlainText"
 				key="editor-rich"
@@ -194,7 +194,7 @@
 				:placeholder="t('mail', 'Write message …')"
 				:focus="isReply"
 				:bus="bus"
-				@input="onEditorRichInputText" />
+				@input="onEditorInputText" />
 			<MailvelopeEditor
 				v-else
 				ref="mailvelopeEditor"
@@ -510,6 +510,10 @@ export default {
 			type: Object,
 			default: () => html(''),
 		},
+		editorBody: {
+			type: String,
+			default: '',
+		},
 		draftId: {
 			type: Number,
 			default: undefined,
@@ -551,7 +555,7 @@ export default {
 		},
 	},
 	data() {
-		let bodyVal = toHtml(this.body).value
+		let bodyVal = this.editorBody
 		if (bodyVal.length === 0) {
 			// an empty body (e.g "") does not trigger an onInput event.
 			// but to append the signature a onInput event is required.
@@ -610,7 +614,7 @@ export default {
 				},
 			},
 			autoLimit: true,
-			editorRichInputTextReady: false,
+			editorInputTextReady: false,
 		}
 	},
 	computed: {
@@ -748,6 +752,15 @@ export default {
 				// update the selected alias
 				this.onAliasChange(newAlias)
 			}
+		},
+		editorMode() {
+			this.appendSignature = true
+		},
+		bodyVal() {
+			if (this.body.value === '') {
+				this.appendSignature = true
+			}
+			this.handleAppendSignature()
 		},
 	},
 	async beforeMount() {
@@ -924,8 +937,7 @@ export default {
 		onSave() {
 			this.callSaveDraft(false, this.getMessageData)
 		},
-		onInputChanged() {
-			this.callSaveDraft(true, this.getMessageData)
+		handleAppendSignature() {
 			if (this.appendSignature) {
 				const signatureValue = toHtml(detect(this.selectedAlias.signature)).value
 				this.bus.$emit('insertSignature', signatureValue, this.selectedAlias.signatureAboveQuote)
@@ -935,11 +947,11 @@ export default {
 		// needs to bypass an input event of the first initialisation, because of:
 		// an empty body (e.g "") does not trigger an onInput event.
 		// but to append the signature a onInput event is required.
-		onEditorRichInputText() {
-			if (this.editorRichInputTextReady) {
+		onEditorInputText() {
+			if (this.editorInputTextReady) {
 				this.callSaveDraft(true, this.getMessageData)
 			}
-			this.editorRichInputTextReady = true
+			this.editorInputTextReady = true
 		},
 		onChangeSendLater(value) {
 			this.sendAtVal = value ? Number.parseInt(value, 10) : undefined
@@ -957,7 +969,7 @@ export default {
 			logger.debug('changed alias', { alias })
 			this.selectedAlias = alias
 			this.appendSignature = true
-			this.onInputChanged()
+			this.handleAppendSignature()
 		},
 		onAddLocalAttachment() {
 			this.bus.$emit('onAddLocalAttachment')
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -11,6 +11,7 @@
 			:subject="composerData.subject"
 			:attachments-data="composerData.attachments"
 			:body="composerData.body"
+			:editor-body="convertEditorBody(composerData)"
 			:in-reply-to-message-id="composerData.inReplyToMessageId"
 			:reply-to="composerData.replyTo"
 			:forward-from="composerData.forwardFrom"
@@ -25,7 +26,7 @@
 <script>
 import Modal from '@nextcloud/vue/dist/Components/Modal'
 import logger from '../logger'
-import { toPlain } from '../util/text'
+import { toPlain, toHtml, plain } from '../util/text'
 import { saveDraft } from '../service/MessageService'
 import Composer from './Composer'
 import { showError, showSuccess } from '@nextcloud/dialogs'
@@ -70,6 +71,8 @@ export default {
 		},
 	},
 	methods: {
+		toHtml,
+		plain,
 		async saveDraft(data) {
 			if (!this.composerMessage) {
 				logger.info('Ignoring draft because there is no message anymore', { data })
@@ -101,6 +104,7 @@ export default {
 				accountId: data.account,
 				subject: data.subject,
 				body: data.isHtml ? data.body.value : toPlain(data.body).value,
+				editorBody: data.body.value,
 				isHtml: data.isHtml,
 				to: serializeRecipients ? data.to.map(this.recipientToRfc822).join(', ') : data.to,
 				cc: serializeRecipients ? data.cc.map(this.recipientToRfc822).join(', ') : data.cc,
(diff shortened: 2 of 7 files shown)


st3iny (commented):
It works fine when composing a new mail. However, when editing messages in outbox there are 2 problems.

## HTML-Editor
The signature "jumps" when editing existing messages.

Peek 2022-05-06 10-29

## Plain-Editor
The Signature is added multiple times. Each time I edit the message a new copy is appended. That is possibly unrelated because it is caused by the forced selection of the HTML-Editor by default (ref https://github.com/nextcloud/mail/issues/6264).

Peek 2022-05-06 10-35

JuliaKirschenheuter:
@st3iny @ChristophWurst @nimishavijay I've tested actual behavior. First of all: if signature should be in a message, then it needs a "place" in a composer body, like ``` <p></p>```. This was already implemented. 
Signature needs ``` <p></p>``` to be placed in a message. If signature is there - body of composer is not empty, and saving process will be triggered. Right now i don't know how to exclude a signature from other things in message body. 

Now it works like: if there is no signature, and all other stuff is empty or was not changed - saving process will not be triggered. If there is a signature - saving process will be triggered.

ChristophWurst:
I'm not sure I can follow.

Does the signature work now? Is there anything left to do?

nimishavijay:
Ideally if there is no other text other than signature, it should not be saved, otherwise there would be a lot of drafts saved with just the signature. But if it is not possible, then no worries :)

ChristophWurst:
* [ ] BUG

1) Configure one account
2) Set a signature
3) Write a new message
4) Click *Undo*

Expected: signature is there like before
Actual: signature is there twice

Bildschirmfoto vom 2022-05-20 08-55-02

ChristophWurst:
This still happens.

Bildschirmfoto vom 2022-05-20 13-19-22

I'm using plain text mode.

Bildschirmfoto vom 2022-05-20 13-19-36

JuliaKirschenheuter:
> This still happens.
> I'm using plain text mode.

Yes, this problem is only in a plain text mode. We are trying to solve this problem

ChristophWurst:
Right now the front-end converts the editor input to html/plain text before it is sent to the server. How about we always store the HTML and let the backend convert to plain if necessary? Would that help with the signature detection @kesselb?

kesselb:
> Would that help with the signature detection @kesselb?

Yep. If we keep the message in html the existing logic works (that should be already the case when you switch the html). If possible I would always store html and the version to sent. If the users open the message again from the outbox we use the stored html for the editor and update the version to sent.

ChristophWurst:
Then let's add a migration for this new column and attribute, shall we?

ChristophWurst:
To update myself on the current status: the signature is still inserted twice.

Let's add the column then.

ChristophWurst:
Outbox messages now have a nullable `editorBody`.

New todo
1) Fill it with the editor contents when saving an outbox message
2) Load the contents for the editor. Beware of the possible null value. You have to use `text` as a fallback if `bodyText` is not there (upgraded from an older app version).

ChristophWurst:
Info.xml needs a bump so the migration runs

ChristophWurst:
@GretaD the bump is here. Go to main. Run `occ upgrade`. Then switch to this branch and run upgrade again. Then your `oc_loca_messages` will have a `editor_body` column

GretaD (approved):

src/components/NewMessageModal.vue
```diff
@@ -11,6 +11,7 @@
 			:subject="composerData.subject"
 			:attachments-data="composerData.attachments"
 			:body="composerData.body"
+			:editor-body="composerData.editorBody"
```
ChristophWurst, line 14: what happens if `editorBody` is null? Every installation with existing outbox messages will run into this.

ChristophWurst: https://github.com/nextcloud/mail/pull/6372#issuecomment-1143325891

kesselb:
> @GretaD the bump is here. Go to main. Run `occ upgrade`. Then switch to this branch and run upgrade again. Then your `oc_loca_messages` will have a `editor_body` column

I don't see the migration anymore.

ChristophWurst:
And this, friends, is why we don't force push.

ChristophWurst:
I've pushed the previous state to `fix/fix-6361-composer-email-signature-restored`.

kesselb:
- [ ] **Losing formatting when clicking outside the modal**

- Write a **plain** text message (more than one word, multiple lines, newlines, ...)
- Select send later
- Open the message again
- Click outside the composer window (=> this will close the message
- Open the message again
- Newlines are gone

https://user-images.githubusercontent.com/3902676/171648216-c6ab7dd0-48ff-429b-8b42-376e1a047f84.mp4

src/components/NewMessageModal.vue
JuliaKirschenheuter: i would say this is solved

JuliaKirschenheuter:
> - [ ] **Losing formatting when clicking outside the modal**

@ChristophWurst @kesselb should we fix this in a separate bug? it is still very important, but would be great to fix this signatures first

src/components/NewMessageModal.vue
```diff
@@ -99,6 +102,7 @@ export default {
 				accountId: data.account,
 				subject: data.subject,
 				body: data.isHtml ? data.body.value : toPlain(data.body).value,
+				editorBody: data.body.value,
```
ChristophWurst, line 105: unused

ChristophWurst (approved):
Does the trick

ChristophWurst:
Be careful with your final rebase. You have "fixups" that aren't following the git fixup conventions. Git rebase will *not* be able to reorder your commits and squash them automatically. You will have to order them yourself. Right now the second from last commit will probably be a fixup of the merge and therefore potentially dropped.

If the IDE can't do it please do fixup commits via CLI and use the `--fixup=<sha1>` switch.

src/components/NewMessageModal.vue
JuliaKirschenheuter: this one is needed, because action "updateMessage" needs editorBody. If i would remove this, it would be impossible to open again the undo-sendet-message

src/components/NewMessageModal.vue
```diff
@@ -181,6 +185,19 @@ export default {
...
 		},
+		convertEditorBody(composerData) {
+			if (composerData.editorBody !== null) {
+				return composerData.editorBody
+			}
+			if (composerData.body === null) {
+				return ''
+			}
+			if (composerData.body.format !== 'html') {
+				return toHtml(composerData.body).value
```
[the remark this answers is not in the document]
ChristophWurst: https://github.com/nextcloud/mail/blob/213c25884ed2162e8d25e8ca3b6a6c6255170ea1/src/util/text.js#L199-L201 returns the identical object.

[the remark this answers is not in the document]
JuliaKirschenheuter: ah, thank you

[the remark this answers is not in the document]
JuliaKirschenheuter: But, i still cannot remove it. This is still needed, if backend give a plain text, this should be explicitly converted to html. Otherwise we would run into problem described here https://github.com/nextcloud/mail/pull/6372#issuecomment-1144910893

src/components/NewMessageModal.vue
```diff
@@ -181,6 +185,19 @@ export default {
...
+			} else {
+				 return composerData.body.value
+			}
```
ChristophWurst, line 199:
```suggestion
			toHtml(composerData.body).value
```

This is the change I suggested.

Leave it as is if it doesn't make sense. It's just a nitpick.

src/components/NewMessageModal.vue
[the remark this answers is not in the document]
ChristophWurst: I think my intent was not clear. I did not suggest dropping the conversion, only your if-else condition.

```js
			if (composerData.body.format !== 'html') {
				return toHtml(composerData.body).value
			} else {
				 return composerData.body.value
			}
```

should be

```js
			return toHtml(composerData.body).value
```

If this doesn't make sense then leave the code as is.

ChristophWurst (dismissed):
Works

I've pushed a fixup for the now outdated version in the migration and allow the migration to run more than once.

ChristophWurst:
> Losing formatting when clicking outside the modal

The problem already occurs when the modal is *opened*, when we "stop" the message. The `showMessageComposer` action uses `data.body.value` for `editorBody`, the plain text version.

src/store/actions.js
```diff
@@ -368,6 +368,7 @@ export default {
 			originalSendAt = data.sendAt
 			const message = {
 				...data,
+				editorBody: data.body.value,
```
ChristophWurst, line 371: drop this line to fix https://github.com/nextcloud/mail/pull/6372#issuecomment-1144910893

`data` has a `editorBody` property with the correct value, the HTML markup

In its current state the plain text is saved as editor body:
Bildschirmfoto vom 2022-06-08 12-44-26

src/store/actions.js
```diff
@@ -391,6 +392,7 @@ export default {
 				message: {
 					...message.data,
 					body: message.data.isHtml ? body.value : toPlain(body).value,
+					editorBody: body.value,
```
ChristophWurst, line 395: this is also wrong and causes https://github.com/nextcloud/mail/pull/6372#issuecomment-1144910893

src/store/actions.js
JuliaKirschenheuter: Thanks, fixed.

src/components/OutboxMessageListItem.vue
```diff
@@ -145,6 +145,7 @@ export default {
 			const message = {
 				...this.message,
 				failed: false,
+				editorBody: this.message.body.value,
```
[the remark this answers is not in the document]
ChristophWurst: Delete the line to fix *Send now* :)

src/components/Composer.vue
```diff
@@ -926,6 +942,9 @@ export default {
 		},
 		onInputChanged() {
 			this.callSaveDraft(true, this.getMessageData)
+			this.editorPlainTextInputReady = true
```
ChristophWurst, line 945: this attribute does not exist

ChristophWurst:
:warning: I will push fixups for the linter and test failure

src/components/Composer.vue
JuliaKirschenheuter: removed onInputChanged

ChristophWurst (approved):
Works

The only thing I noticed is that if you have two accounts, editing an outbox message always has the first account selected in the alias picker. So even if your message is sent from the second alias, the first one is shown. To switch to the first you need to click the second and then the first. This is very likely an issue with the outbox/composer, unrelated to this change. I'll create a ticket.

Issue #6400, opened 2022-05-10
Memory leak during (initial) synchronization of large mailbox


ChristophWurst:
### Steps to reproduce

1. Add an account with an inbox of roughly 200k messages or more
2. Run the initial synchronization via `occ:mail:account:sync`
3. Wait

### Expected behavior

The synchronization runs in batches but eventually finishes.

### Actual behavior

```
PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 8388616 bytes) in /nextcloud/mail/lib/Db/MessageMapper.php on line 87
PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 327680 bytes) in /nextcloud/apps/mail/lib/Cache/Cache.php on line 128
```

### Mail app version

v1.12

### Additional info

If I comment out the cache in the IMAP client factory, memory usage increases very slowly with each batch run.

ChristophWurst:
Finding: explictly `unset`ing some heavy variables does _not_ increase the memory consumption. So the PHP garbage collector does an excellent job already.

ChristophWurst:
Memory usage jumps when ``\OCA\Mail\IMAP\MessageMapper::findAll`` calls ``\OCA\Mail\IMAP\MessageMapper::findByIds``. Afterwards the usage doesn't to down much, so I suspect the leak start there.

ChristophWurst:
I can see quite some memory usage before/when we do the bulk insert

```
[debug] Running initial sync for 1317
[debug] Initial sync 2880:INBOX - start took 0s. 8/9MB mem used
[debug] Initial sync 2880:INBOX - findHighestUid took 0s. 8/9MB mem used
[debug] Initial sync 2880:INBOX - getClient took 0s. 8/9MB mem used
[debug] Initial sync 2880:INBOX - meta search took 2s. 8/9MB mem used
[debug] Initial sync 2880:INBOX - fetch UIDs took 0s. 13/14MB mem used
[debug] Initial sync 2880:INBOX - uids to fetch took 0s. 12/14MB mem used
[debug] Initial sync 2880:INBOX - findByIds took 5s. 53/53MB mem used
[debug] Initial sync 2880:INBOX - fetch all messages from IMAP took 0s. 50/53MB mem used
[debug] Initial sync 2880:INBOX - persist chunk of messages in database took 0s. 52/55MB mem used
[...]

PR #6408, opened 2022-05-11
Fix memory leak when accessing Horde headers

ChristophWurst:
Horde offers access to headers either unparsed, as stream or parsed into
an object. Before the patch we used the latter. That has the current bug
that it opens a PHP memory stream, parses into an object but leaves the
stream open. If this is done thousands of times, there is a considerate
amount of memory that doesn't get freed.

With the patch we request just the stream, trigger the parsing ourselves
and close the stream afterwards. This frees the memory.

This contributes to https://github.com/nextcloud/mail/issues/6400 but doesn't fully fix it.

Upstream PR https://github.com/horde/Imap_Client/pull/18
Fork PR https://github.com/bytestream/Imap_Client/pull/2


miaulalala (approved):
LGTM

PR #6409, opened 2022-05-11
Close streams after reading message parts

ChristophWurst:
We ask Horde for a stream of the message parts but those streams are not
closed. Closing the stream frees memory.

Discovered while I work on https://github.com/nextcloud/mail/issues/6400. This applies the pattern of https://github.com/nextcloud/mail/pull/6408 to other areas. However, this doesn't contribute to the sync memory leak.


miaulalala (approved):
LGTM. Fingers crossed for no side effects!

kesselb (approved):
:+1:

PR #6410, opened 2022-05-11
Turn off cache for initial sync to prevent a memory leak

ChristophWurst:
The cache implementation doesn't persist its values until the shutdown
of the process. For each iteration of the initial synchronization we
have to create one instance of the cache. Those two facts combined mean
that for each iteration Horde builds a new set of cache data for the
mailbox and therefore creates a memory leak.

Turning off cache isn't great but the data we need for the synchronization has to be fetched anyway.

Ref https://github.com/nextcloud/mail/issues/6400#issuecomment-1123773725

### DIFF
--- a/lib/IMAP/IMAPClientFactory.php
+++ b/lib/IMAP/IMAPClientFactory.php
@@ -60,9 +60,10 @@ class IMAPClientFactory {
 	 * (and stale) connections at a minimum.
 	 *
 	 * @param Account $account
+	 * @param bool $useCache
 	 * @return Horde_Imap_Client_Socket
 	 */
-	public function getClient(Account $account): Horde_Imap_Client_Socket {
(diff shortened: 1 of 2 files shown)


miaulalala (commented):
With the current initial value for `$useCache`, the cache is always on still. Are you adding the param in the initial sync later?

kesselb:
LGTM. It's inconvenient that Horde_Imap_Client_Socket does not expose a way to access the wrapped cache object. I guess turning off the cache is our best option here.

kesselb (approved):

ChristophWurst:
> With the current initial value for `$useCache`, the cache is always on still. Are you adding the param in the initial sync later?

Correct. I have lots of local changes and forgot to add one line with `git add -p`. Should be good now.

miaulalala (approved):

PR #6411, opened 2022-05-11
Hide subscribed / unsubscribed checkbox for Infox, Favorites and Sent

### LINKED ISSUE #6296: Unsubscribe option for "Favourites" folder should be hidden
### Steps to reproduce

1. Try unsubscribing from mailbox "Favourites"
2. Frontend gets confused and throws an error

### Expected behavior

No unsubscribe option is offered as it's not a real mailbox.

### Actual behavior

The unsubscribe option is offered and errors out.

### Mail app version

1.11.6

### Mailserver or service

no info

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

_No response_

JuliaKirschenheuter:
fixes #6296

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -260,7 +260,7 @@ export default {
 			return this.mailbox.name.toLowerCase() !== 'inbox'
 		},
 		notVirtual() {
-			return !this.account.isUnified && this.mailbox.specialRole !== 'flagged'
+			return !this.account.isUnified && this.mailbox.specialRole !== 'flagged' && !this.filter
 		},
 		title() {
 			if (this.filter === 'starred') {


JuliaKirschenheuter:
Should we also hide subscribed / unsubscribed checkbox for "Inbox" and "Sent" folders? Because it is actually not possible to unsubscribe them, they are always "subscribed".

ChristophWurst:
> Should we also hide subscribed / unsubscribed checkbox for "Inbox" and "Sent" folders? Because it is actually not possible to unsubscribe them, they are always "subscribed".

I'd says yes.

src/components/NavigationMailbox.vue
```diff
@@ -92,7 +92,7 @@
 			</ActionButton>
 
 			<ActionCheckbox
-				v-if="notVirtual"
+				v-if="notVirtual && filter !== 'starred' "
```
ChristophWurst, line 95: you could also check if `filter` is not undefined. Then this will still work for any other kind of filtered mailbox.

ChristophWurst: and you could move the logic into `notVirtual`

The part ``this.mailbox.specialRole !== 'flagged'`` seems to actually target the case you are fixing here. Did we lose the specialRole=flagged on the favorites mailbox?

JuliaKirschenheuter: > you could also check if filter is not undefined. Then this will still work for any other kind of filtered mailbox.

unfortunately it is not working this way

> and you could move the logic into notVirtual

done

> Did we lose the specialRole=flagged on the favorites mailbox?

i think not, or what do you mean? 

new pushed version works

src/components/NavigationMailbox.vue
```diff
@@ -206,7 +206,8 @@ export default {
 			return this.mailbox.name.toLowerCase() !== 'inbox'
 		},
 		notVirtual() {
-			return !this.account.isUnified && this.mailbox.specialRole !== 'flagged'
+			return !this.account.isUnified && this.mailbox.specialRole !== 'flagged' && this.filter !== 'starred'
```
ChristophWurst, line 209:
```suggestion
			return !this.account.isUnified && this.mailbox.specialRole !== 'flagged' && !!this.filter
```

ChristophWurst (commented):
Then this condition will work with any kind of filtered mailbox

src/components/NavigationMailbox.vue
ChristophWurst: See https://github.com/nextcloud/mail/pull/6411#pullrequestreview-974926360

ChristophWurst (approved):
Works right now

PR #6416, opened 2022-05-12
Print the memory consumption of a long occ sync

ChristophWurst:
This will help debug bugs like https://github.com/nextcloud/mail/issues/6400 in production environments because it makes it visual how the memory consumption drifts off.

### DIFF
--- a/lib/Command/SyncAccount.php
+++ b/lib/Command/SyncAccount.php
@@ -40,6 +40,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use function memory_get_peak_usage;
+use function memory_get_usage;
 
 class SyncAccount extends Command {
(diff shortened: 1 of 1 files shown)


lib/Command/SyncAccount.php
```diff
@@ -113,7 +114,8 @@ private function sync(Account $account, bool $force, OutputInterface $output): v
 				throw $e;
 			}
 
-			$output->writeln("Batch of new messages sync'ed");
+			$mbs = (int)(memory_get_usage() / 1024 / 1024);
```
CarlSchwan, line 117: Maybe add a check that the function exists just in case this is blocked by a security policy

CarlSchwan (approved):
Looks good

lib/Command/SyncAccount.php
ChristophWurst: Most disabled functions I've seen were at hosters and they don't give CLI access either. Therefore I think this will show up very rarely.

PR #6417, opened 2022-05-12
Add more detailed performance logging to occ sync

ChristophWurst:
This was done for https://github.com/nextcloud/mail/issues/6400 but I think it makes sense to keep in general. We previously had the issue that some of the logging, especially from the mapper's `findAll`, was not shown on the console output and we had to match the info from the nextcloud log. Now the lines are shown with the console output and it's a lot easier to trace what happens during a sync.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -36,6 +36,7 @@ use Horde_Mime_Part;
 use OCA\Mail\Db\Mailbox;
 use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\Model\IMAPMessage;
+use OCA\Mail\Support\PerformanceLoggerTask;
 use OCP\AppFramework\Db\DoesNotExistException;
 use Psr\Log\LoggerInterface;
 use function array_filter;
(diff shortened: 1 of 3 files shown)


miaulalala (approved):

lib/IMAP/MessageMapper.php
```diff
@@ -162,6 +167,7 @@ public function findAll(Horde_Imap_Client_Socket $client,
 				'ids' => new Horde_Imap_Client_Ids($lower . ':' . $upper)
```
miaulalala, line 167: I'm still wondering whether we could directly fetch the messages themselves here, saving us another fetch / search query further down with `findByIds`

ChristophWurst: In a fragmented mailbox you will run into overfetching when the target range of UIDs is dense. This will then create backpressure on how much data we need to download from IMAP *and* create a unexpectedly large number of INSERTs unless we discard the overfetched data.

So yes it is one IMAP roundtrip more right now but it prevents the overall synchronization from running too long and hitting a timeout.

PR #6419, opened 2022-05-12
Include memory consumption in performance logging

ChristophWurst:
```
[debug] Initial sync 2880:INBOX - fetch all messages from IMAP took 6s. 45/48MB memory used
[debug] Initial sync 2880:INBOX - persist messages in database took 6s. 41/48MB memory used
```

I put safe guards around the functions because they might be disabled.

Developed for https://github.com/nextcloud/mail/issues/6400 but will help us see memory issues in other places more easily.


miaulalala (approved):
LGTM

PR #6421, opened 2022-05-12
Add forward as attachment on more action menu

### LINKED ISSUE #5655: "Forward as attachment" is hard to discover
### Expected behavior

As a user I occasionally want to forward a message as attachment as opposed to the default quoted forward. The app supports this feature. Yet it's only shown in the multiselect actions menu, not the simple envelope or message actions. This leads to confusion.

### Actual behavior

The app supports this feature. Yet it's only shown in the multiselect actions menu, not the simple envelope or message actions. This leads to confusion.

This action wasn't only done for multiselect by accident, there was some reasoning behind: https://github.com/nextcloud/mail/pull/4251#issuecomment-747436011 and following comments.

### Proposed solution

Once the nc/vue components allow something like sub actions we could have *Forward as attachment* as sub item of *Forward*. Then it's directly accessible and discoverable.

### Mail app

**Mail app version:** v1.8.0-1.10.0

---

cc @nextcloud/designers

GretaD:
fixes #5655

### DIFF
--- a/src/components/MenuEnvelope.vue
+++ b/src/components/MenuEnvelope.vue
@@ -96,6 +96,12 @@
 						{{ t('mail', 'More actions') }}
 					</template>
 				</ActionButton>
+				<ActionButton
+					icon="icon-forward"
+					:close-after-click="true"
+					@click.prevent="forwardSelectedAsAttachment">
+					{{ t('mail', 'Forward message as attachment') }}
+				</ActionButton>
 				<ActionButton icon="icon-add"
 					:close-after-click="true"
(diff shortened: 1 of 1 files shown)


GretaD:
@marcoambrosini before i mark this as ready for review, just double checking, this was your proposal, right? Just move it on more actions?

GretaD:
Nevermind, its very clear from Jan's comment > _Yes, agree with Marco that this action should go into "More actions"._

ChristophWurst (approved):
Code looks good

PR #6234, opened 2022-04-13
Performance improvements migration

miaulalala:
Add indices on mail_messages table

change column types to bigint for tables recipients and messages

Change foreign keys to named foreign keys for recipients and attachments

Fixes #4466
Fixes #5985
Fixes #4664 
Fixes #6223

PR #6434, opened 2022-05-12
Do not delete outbox recipients during migration

ChristophWurst:
Updating my local instance after the merge of https://github.com/nextcloud/mail/pull/6234 wiped recipients of scheduled messages.


kesselb (approved):
Thanks :+1: 

I run into the same error locally :see_no_evil:

PR #6436, opened 2022-05-12
Fix column type on mysql

kesselb:
Length option on mysql/maria does not affect the min/max values the column can store :(

https://stackoverflow.com/questions/5634104/what-is-the-size-of-column-of-int11-in-mysql-in-bytes

https://github.com/nextcloud/3rdparty/blob/51285f0a85f53a534201ea9a189669c2822ce9b8/doctrine/dbal/src/Schema/MySQLSchemaManager.php#L192-L200

### DIFF
--- a/lib/Migration/Version1130Date20220412111833.php
+++ b/lib/Migration/Version1130Date20220412111833.php
@@ -28,8 +28,11 @@ namespace OCA\Mail\Migration;
 
 use Closure;
 use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+use Doctrine\DBAL\Types\Type;
 use OCP\DB\ISchemaWrapper;
 use OCP\DB\QueryBuilder\IQueryBuilder;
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
:exploding_head:

ChristophWurst:
I've deleted the backport comment because the migration only exists for v1.13

lib/Migration/Version1130Date20220412111833.php
```diff
@@ -86,6 +89,9 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		/** @var ISchemaWrapper $schema */
 		$schema = $schemaClosure();
 
+		// bigint and autoincrement are not possible on sqlite: https://github.com/nextcloud/server/commit/f57e334f8395e3b5c046b6d28480d798453e4866
```
kesselb, line 92:
```suggestion
		// bigint and primary key with autoincrement is not possible on sqlite: https://github.com/nextcloud/server/commit/f57e334f8395e3b5c046b6d28480d798453e4866
```

PR #6261, opened 2022-04-16
Make user to edit reciepients list

sazanof:
**Hello!** First of all, thank you for the wonderful project and your attitude to it!

_I've been using Nextcloud for a little over a month (we moved from o365), so I'm quite green in this regard=)_

Using the Mail application, I received a question whether it is possible to somehow edit the recipient lists when writing a letter and opening a draft. Some employees write 10-20 addresses in the "to" field (yes, yes, yes =)) and except for pressing "backspace" it is not possible to delete recipients.

I ask you to consider my request to change the functionality in terms of editing the list of recipients.

I took out 2 buttons in a separate place.

- show hide advanced mode.
- show hide BCC CC

edit-rec-list
edit-rec-list2

Thanks!

PR #6439, opened 2022-05-13
Users can edit recipients list while writing a message

sazanof:
**Hello!**  Link to the discussion of the previously closed PR (my mistake) - https://github.com/nextcloud/mail/pull/6261
!Peek 2022-05-11 11-18
I tried to take into account all the comments in the old PR.
THANKS!

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -4,93 +4,149 @@
 			<label class="from-label" for="from">
 				{{ t('mail', 'From') }}
 			</label>
-			<Multiselect
-				id="from"
-				:value="selectedAlias"
-				:options="aliases"
-				label="name"
-				track-by="selectId"
-				:searchable="false"
-				:custom-label="formatAliases"
-				:placeholder="t('mail', 'Select account')"
-				:clear-on-select="false"
-				@select="onAliasChange" />
+			<div class="composer-fields--custom">
+				<Multiselect
+					id="from"
+					:value="selectedAlias"
+					:options="aliases"
+					label="name"
+					track-by="selectId"
+					:searchable="false"
+					:custom-label="formatAliases"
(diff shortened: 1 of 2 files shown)


JuliaKirschenheuter (approved):
works well, thank you =)

jancborchardt:
Just 1 feedback still design-wise, having 2 icons/actions there is confusing and we should have only 1:
- If there is anyone in cc or bcc field already, no action should hide that (so you don't forget about it)
- Expanding the recipients lists and showing cc/bcc can be toggled using the same button. Basically it's a "show me everything" vs "show me only lines with content, single line each" toggle.

And that icon/action should be in the "to" row since it relates to the recipients. It should not be in "from" as it does not relate to yourself.

Does that make sense @sazanof? :) The icon can be the expand/compact toggle you used.

jancborchardt:
@sazanof almost! :) Once opened, when you click on the collapse toggle:
- If there is no one in cc or bcc: Collapse those rows too
- If there is anyone in cc or bcc: Leave that row displayed (cc if someone is in cc, bcc of someone is in bcc, both if someone in both)

sazanof:
@jancborchardt so, the drum roll.....
Peek 2022-05-16 18-44
Like this?

GretaD:
> by the way, the BCC is not loaded again (into the modal window) after saving the draft (test on `main` branch). `bcc: []` May be there is an open issue, i dont know...

a known issues, yes, #6458

jancborchardt:
@sazanof awesome! This seems perfect indeed. :) A thing I’m wondering about, but it’s probably separate (Vue modal component?) is that everything jumps vertically since the size/height changes.

But I guess it’s not blocking this change? @skjnldsv @marcoambrosini

sazanof:
@st3iny Done. I hope I did everything right =)

PR #6456, opened 2022-05-16
Composer should not to open when we select envelopes on draft

### LINKED ISSUE #6389: Multi-selecting drafts open modal editor
### Steps to reproduce

1. Go to Drafts folder
2. Click on three dots for selecting the first message
3. Click on checkbox to select further messages
4. See how modal editor window opens

### Expected behavior

Selecting a message with the checkbox selects the message

### Actual behavior

Selecting a message with the checkbox selects the message and opens it in the modal editor

### Mail app version

1.12.0

### Mailserver or service

_No response_

### Operating system

Docker image

### PHP engine version

PHP 8.0

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

_No response_

GretaD:
fixes #6389

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -350,8 +350,8 @@ export default {
 		toggleSelected() {
 			this.$emit('update:selected', !this.selected)
 		},
-		onClick() {
-			if (this.draft) {
+		onClick(event) {
+			if (this.draft && !event.defaultPrevented) {
 				this.$store.dispatch('showMessageComposer', {
 					data: {
 						...this.data,


GretaD:
Please have a look and see if this makes sense

ChristophWurst:
Does this work for you?

GretaD:
> Does this work for you?

yes

src/components/Envelope.vue
```diff
@@ -350,15 +350,19 @@ export default {
...
-			if (this.draft) {
-				this.$store.dispatch('showMessageComposer', {
-					data: {
-						...this.data,
-						draftId: this.data.databaseId,
-					},
-					templateMessageId: this.data.databaseId,
-				})
+		onClick(event) {
+			if (event) {
```
ChristophWurst, line 354: I don't understand why `event` would be truthy only in some cases. Don't we always get an event object from Vue?

GretaD: i honestly dont know, if i do `onClick(event) {
 event.preventDefault() }`
i doesnt work, if i add the if statement, it works, maybe my approach is wrong

ChristophWurst: Multiselect works now, but you can't open a draft anymore. 

The `event` is never falsy. So `onClick` always goes into the `if` branch, never `else`.

Issue #6382, opened 2022-05-07
Sending mails to groups fails (again!)


escribana:
### Steps to reproduce

1. Create new message for group
2. Send
3.  Message lands in outbox without being sent

### Expected behavior

Message is sent to all users of the group

### Actual behavior

Message stays in outbox and cannot be sent

### Mail app version

1.12.0

### Mailserver or service

bluewin.ch & gmx.net

### Operating system

Ubuntu 20.0.4

### PHP engine version

PHP 7.4

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

Log for bluewin.ch -------------------------------------------------------------------------------------------------

OCA\Mail\Exception\ServiceException: Could not send message: need RCPT before DATA
/var/www/html/nextcloud/apps/mail/lib/Service/MailTransmission.php - line 271:
OCA\Mail\Service\MailTransmission->sendMessage()
/var/www/html/nextcloud/apps/mail/lib/Service/OutboxService.php - line 129:
OCA\Mail\Service\MailTransmission->sendLocalMessage()
/var/www/html/nextcloud/apps/mail/lib/Controller/OutboxController.php - line 191:
OCA\Mail\Service\OutboxService->sendMessage()
/var/www/html/nextcloud/lib/private/AppFramework/Http/Dispatcher.php - line 217:
OCA\Mail\Controller\OutboxController->send()
/var/www/html/nextcloud/lib/private/AppFramework/Http/Dispatcher.php - line 126:
OC\AppFramework\Http\Dispatcher->executeController()
/var/www/html/nextcloud/lib/private/AppFramework/App.php - line 157:
OC\AppFramework\Http\Dispatcher->dispatch()
/var/www/html/nextcloud/lib/private/Route/Router.php - line 302:
OC\AppFramework\App::main()
/var/www/html/nextcloud/lib/base.php - line 1008:
OC\Route\Router->match()
/var/www/html/nextcloud/index.php - line 36:
OC::handleRequest()

Caused by Horde_Mime_Exception: need RCPT before DATA
/var/www/html/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Mime/Horde/Mime/Mail.php - line 479:
Horde_Mime_Part->send()
/var/www/html/nextcloud/apps/mail/lib/Service/MailTransmission.php - line 208:
Horde_Mime_Mail->send()
/var/www/html/nextcloud/apps/mail/lib/Service/MailTransmission.php - line 271:
[...]

PR #6460, opened 2022-05-16
Fix sending messages to groups

ChristophWurst:
Groups were expanded in the accounts controller. Since moving over to
the outbox logic this feature was missing and internal group identifiers
were passed to SMTP.

With this patch groups are expanded again just before a message is sent.
This means the group memberships are read as late as possible and
editing an outbox message looks like the original message because
members have not been expanded there yet.

## How to test
* Create a few contacts and form a group
* Compose a new message, add the group as recipient
* Send the message

On main/v1.12.0: :boom: 
Here: :sunglasses: 

Regression of https://github.com/nextcloud/mail/pull/6031
Fixes https://github.com/nextcloud/mail/issues/6382

### DIFF
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -410,16 +410,12 @@ class AccountsController extends Controller {
 		$account = $this->accountService->find($this->currentUserId, $id);
 		$alias = $aliasId ? $this->aliasesService->find($aliasId, $this->currentUserId) : null;
 
-		$expandedTo = $this->groupsIntegration->expand($to);
-		$expandedCc = $this->groupsIntegration->expand($cc);
-		$expandedBcc = $this->groupsIntegration->expand($bcc);
-
-		$count = substr_count($expandedTo, ',') + substr_count($expandedCc, ',') + 1;
+		$count = substr_count($to, ',') + substr_count($cc, ',') + 1;
 		if (!$force && $count >= 10) {
 			throw new ManyRecipientsException();
 		}
 
-		$messageData = NewMessageData::fromRequest($account, $expandedTo, $expandedCc, $expandedBcc, $subject, $body, $attachments, $isHtml, $requestMdn);
+		$messageData = NewMessageData::fromRequest($account, $to, $cc, $bcc, $subject, $body, $attachments, $isHtml, $requestMdn);
 		if ($messageId !== null) {
 			try {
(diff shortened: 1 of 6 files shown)


kesselb:
LGTM :+1: 

image

Question: I have a group admin and a user admin who belongs to the group. But the administrator was to lazy to set an email address. Without the email address it will fail at https://github.com/nextcloud/mail/blob/c8148e64d320f2d16b00c5ccde6133f3164892cb/lib/Service/MailTransmission.php#L225

miaulalala (approved):

ChristophWurst:
> Question: I have a group admin and a user admin who belongs to the group. But the administrator was to lazy to set an email address. Without the email address it will fail at

Good question. But I don't know how we could/should improve this. We could validate the inputs when a message is added to the outbox. Right now we don't do any of that. You can also add obvious non-email addresses as a sender and the app will just store it.

kesselb:
> > Question: I have a group admin and a user admin who belongs to the group. But the administrator was to lazy to set an email address. Without the email address it will fail at
> 
> Good question. But I don't know how we could/should improve this. We could validate the inputs when a message is added to the outbox. Right now we don't do any of that. You can also add obvious non-email addresses as a sender and the app will just store it.

When one user has no email address sending to a group is still not possible and fails with:

 `Argument 2 passed to OCA\Mail\Address::fromRaw() must be of the type string, null given, called in /var/www/html/apps-extra/mail/lib/Service/MailTransmission.php on line 227`. 

The old code used this check to make sure we email is not empty.

image

Sounds like a good plan to add a warning to new message when some members of a group don't have a valid email address. So the sender is aware that not everyone is getting the email. Should i create a ticket for this?

ChristophWurst:
> Question: I have a group admin and a user admin who belongs to the group. But the administrator was to lazy to set an email address.

I've added a filter for member to only consider those with an email address.

ChristophWurst:
Could you file that as a new ticket @LostinSpacetime? Do you have access to the email headers/text that get rejected?

LostinSpacetime:
> Could you file that as a new ticket @LostinSpacetime? Do you have access to the email headers/text that get rejected?

I did file a new bug report here #6497 
The problem was introduced with 1.12.0 but is not related to groups. It only happens if the address/es are in BCC.

Issue #6441, opened 2022-05-13
Composer modal causes errors while editing recipients


JuliaKirschenheuter:
### Steps to reproduce

1. Open a composer modal for creating new message  / edit existing one
2. Change recipients: to, cc, bcc
3. Wait for saving a draft
4. Edit recipients again

### Expected behavior

No errors in a browser console 

### Actual behavior

permanent throwing errors: 

Error in render: "TypeError: this.message.to.map is not a function
found in
---> <OutboxMessageListItem> at src/components/OutboxMessageListItem.vue

Error in render: "TypeError: this.message.cc.map is not a function"
found in
---> <OutboxMessageListItem> at src/components/OutboxMessageListItem.vue

### Mail app version

1.12.0

### Mailserver or service

not relevant 

### Operating system

not relevant

GretaD:
im stuck in this state. and the error is 
xhr.js:210          PUT http://nextcloud.local/index.php/apps/mail/api/outbox/39 500 (Internal Server Error)

![outboxclose](https://user-images.githubusercontent.com/12728974/168278963-71375acb-97f7-464d-adb5-97e4b8f27701.png)

GretaD:
i did a quick bisect and this showed up as the breaking change, but not sure how is this related

<pre><font color="#C4A000">commit 5095e1cde6967c21f35e911db822c40b117c8548</font>
Merge: d582f1fa3 094759006
Author: Christoph Wurst &lt;ChristophWurst@users.noreply.github.com&gt;
Date:   Thu May 12 13:12:22 2022 +0200

    Merge pull request #6342 from nextcloud/chore/drop-php-73
    
    Drop PHP7.3 support
</pre>

GretaD:
i cannot reproduce this anymore after git pull and update everywhere

ChristophWurst:
The error happens because the server returns outbox message `cc` as Object instead of an array.

ChristophWurst:
![Bildschirmfoto vom 2022-05-16 17-27-49](https://user-images.githubusercontent.com/1374172/168628838-958804dd-343f-4c86-8b05-c4e3a6fb689e.png)

Looks like I can narrow it down. The *to* has array index 0, *cc* has 1. Therefore the former is considered non-indexed and the other is indexed. PHP converts that into an array and an object.

PR #6463, opened 2022-05-16
Always serialize outbox message recipients into arrays

ChristophWurst:
array_filter keeps the index of array elements prior to the filtering.
So if you have a message with one *to* and one *cc* values, the *to*
will have index 0 and *cc* has 1. json_encode will later consider the
former as a simple PHP array and map it to a json array. Since the *cc*
value has a non-0 index for its first element, it would convert the
value to a json object and use the 1 index as key.

By running the values through array_value we can re-index all recipient
arrays before they run through json_encode.

Fixes https://github.com/nextcloud/mail/issues/6441.


JuliaKirschenheuter (approved):
Thank you Christoph!
Works well, i'll merge it!

PR #6464, opened 2022-05-16
Prevent causal read for outbox message and recipient inserts

ChristophWurst:
The message and the recipients are inserted in one transaction but the
recipients are read another time outside a transaction. Read-write split
database clusters might not be in full sync mode and then reading the
recipients gives partial or no results.

The insert will assign the primary key value to the recipient entities.
Therefore we can skip reading the data.

Ref https://github.com/nextcloud/groupware/issues/24


kesselb (approved):
LGTM :+1:

Issue #6457, opened 2022-05-16
Can not close outbox modal of a scheduled message


ChristophWurst:
### Steps to reproduce

1. Compose a message
2. Schedule it for later
3. Reload the page
4. Open the same message
5. Click outside the modal

### Expected behavior

The modal closes.

### Actual behavior

The modal closes but the background overlay stays

![Bildschirmfoto vom 2022-05-16 11-25-17](https://user-images.githubusercontent.com/1374172/168561849-51d3f81d-96e3-4be3-b64b-7db64dff6a66.png)

```
XHRPUThttps://localhost/apps/mail/api/outbox/225
[HTTP/2 500 Internal Server Error 116ms]

OCA\Mail\Controller\OutboxController::update(): Argument #4 ($body) must be of type string, array given
``` 

This is sent to the server
![Bildschirmfoto vom 2022-05-16 11-32-05](https://user-images.githubusercontent.com/1374172/168563126-37f218e5-cda0-425c-854c-ad1a5c33b241.png)

### Additional info

As far as I can tell this is a regression of https://github.com/nextcloud/mail/pull/6350

GretaD:
i think this is a regression of https://github.com/nextcloud/mail/commit/a4b7ca3feec30ff0302fdbf1f6fcbdd3708f1954#diff-7f345453ff2ebcb0438364dc0c2bcdc5f1737b31e70168b49c95b4f54d2e4d8bR5

ChristophWurst:
Because it seems to be of relevance: my account is set to use *plain text* editing and not HTML.

PR #6471, opened 2022-05-17
Fix text format when closing the composer modal

ChristophWurst:
When plain editing mode is selected, rescheduling an outbox message sent
the text object to the server, while that data structure is only known
on the front-end. HTML had already been handled correctly, only plain
missed the conversion to string.

## How to test
* Set your account to *plain text* editing
* Compose a new message
* Schedule it
* Click send
* Open the outbox
* Open the message
* Click outside the modal

Main: modal vanishes but background overlay stays; js error on console shows; request to server fails and message is *not* rescheduled.
Here: modal closes and message is correctly rescheduled.

---

Leftover of https://github.com/nextcloud/mail/pull/6350. This needs a backport.
Fixes https://github.com/nextcloud/mail/issues/6457

### DIFF
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -385,7 +385,7 @@ export default {
 				id: message.data.id,
 				message: {
 					...message.data,
-					body: message.data.isHtml ? body.value : toPlain(body),
+					body: message.data.isHtml ? body.value : toPlain(body).value,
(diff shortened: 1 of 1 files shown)


st3iny (approved):
Good catch :bug:

PR #6481, opened 2022-05-17
Improve composer size on different screens

### LINKED ISSUE #6474: Unnecessary scrolling in message composer modal / ghost attachments
### Steps to reproduce

1. Compose a new message
2. Add an attachment (from Nextcloud files)

### Expected behavior

I see that there is a file attached to the message I'm currently composing

### Actual behavior

The attachment is invisible because the editor has lots of scrolling area even when empty. The attachments are only visible once I scroll down.

### Before scrolling
Bildschirmfoto vom 2022-05-17 10-43-42

### After scrolling
Bildschirmfoto vom 2022-05-17 10-43-47

### Mail app version

v1.13.0

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

GretaD:
fixes #6474

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -116,7 +116,7 @@
 				})
 			}}
 		</div>
-		<div class="composer-fields">
+		<div class="composer-fields message-editor">
 			<!--@keypress="onBodyKeyPress"-->
 			<TextEditor
 				v-if="!encrypt && editorPlainText"
@@ -1078,6 +1078,10 @@ export default {
 .message-composer {
 	margin: 0;
 	z-index: 100;
+	display: flex;
+	flex-direction: column;
+	height: 100%;
+	max-height: 100%;
 }
 
 .composer-actions {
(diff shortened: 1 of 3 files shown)


GretaD:
this is a tentative fix. having `min-height: 40vh` will force the body height to extend when the content is very long without showing the scrolling.

ChristophWurst:
what would be the proper fix?

GretaD:
Maybe this is the proper fix, i just dont know if theres a better one. That is what i was trying to say. If what we want is to remove the scrolling till the modal reaches the "end" of the screen, then this should be it

GretaD:
https://user-images.githubusercontent.com/12728974/169857561-e88a9ad3-c662-4cec-a59c-ad57e7f4bc03.mp4


https://user-images.githubusercontent.com/12728974/169857355-efa0af65-6240-40bb-a4e6-0b21e768ec90.mp4

GretaD:
i think this is not as easy as unnecessary scrolling. The body of the message "continues" outside the composer, see video one above. And the text message will not "force" scrolling while im writing a text message and i reach close to _draft saved_. I will keep writing but if i dont manually scroll, i'd know where my cursor is.

I can make the attachment sticky for now so it doesnt create confusion, as we would change it again when we add the preview, but this is not a scrolling problem for attachment only.

st3iny (approved):
Tested and works! @skjnldsv Thanks a lot for your magic :mage_man:

PR #6488, opened 2022-05-17
Reduce log messages by account auto detect

kesselb:
Account auto detect produce a lot of log messages spamming the log file. 
This pr will reduce the verbosity of the logged information.

### DIFF
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -336,7 +336,6 @@ class AccountsController extends Controller {
 	 * @param bool $autoDetect
 	 *
 	 * @return JSONResponse
-	 * @throws ClientException
 	 */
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
:broom:

lib/Controller/AccountsController.php
```diff
@@ -346,15 +345,15 @@ public function create(string $accountName, string $emailAddress, string $passwo
...
-			return \OCA\Mail\Http\JsonResponse::fail([
+			$data = [
 				'error' => $e->getReason(),
 				'service' => $e->getService(),
 				'host' => $e->getHost(),
 				'port' => $e->getPort(),
-			]);
+			];
+
+			$this->logger->info('Creating account failed: ' . $e->getMessage(), $data);
```
nickvergessen, line 355: Let's hope there will never be a conflict between `$data` and the generated json array :P

nickvergessen (approved):

PR #6501, opened 2022-05-18
Use workaround for boolean and oci

kesselb:
Need the same workaround as https://github.com/nextcloud/mail/pull/4903 to make boolean and oci work. 
For some reason false is stored as null.


ChristophWurst (approved):
:smiling_face_with_tear:

PR #6505, opened 2022-05-19
Migrate secondary message/envelope actions to material icons

ChristophWurst:
Ref https://github.com/nextcloud/groupware/issues/38

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -85,39 +85,62 @@
 				</ActionButton>
 			</EnvelopePrimaryActions>
 			<template v-if="!moreActionsOpen">
-				<ActionButton icon="icon-junk"
-					:close-after-click="true"
+				<ActionButton :close-after-click="true"
 					@click.prevent="onToggleJunk">
+					<template #icon>
+						<AlertOctagonIcon
+							:title="data.flags.$junk ? t('mail', 'Mark not spam') : t('mail', 'Mark as spam')"
+							:size="20" />
+					</template>
 					{{
 						data.flags.$junk ? t('mail', 'Mark not spam') : t('mail', 'Mark as spam')
 					}}
 				</ActionButton>
-				<ActionButton icon="icon-checkmark"
-					:close-after-click="true"
+				<ActionButton :close-after-click="true"
 					@click.prevent="toggleSelected">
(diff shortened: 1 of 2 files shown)


JuliaKirschenheuter:
Thank you for fixing!

Should we also change icons in actions inside of EnvelopePrimaryActions?

ChristophWurst:
Yes. They just need some styling fixes.

src/components/Envelope.vue
```diff
@@ -85,39 +85,62 @@
 				</ActionButton>
 			</EnvelopePrimaryActions>
 			<template v-if="!moreActionsOpen">
-				<ActionButton icon="icon-junk"
-					:close-after-click="true"
+				<ActionButton :close-after-click="true"
 					@click.prevent="onToggleJunk">
+					<template #icon>
+						<AlertOctagonIcon
```
GretaD, line 91: this doesnt work, when i mark as spam, the icon is not shown

GretaD:
Please notice on the top right the small arrow showing up when i click select or mark as spam
https://user-images.githubusercontent.com/12728974/169501814-8c302aa3-dd1d-4dd7-bdfe-0df7bfee5606.mp4

src/components/Envelope.vue
ChristophWurst: The action icon  renders for spam and non-spam message. Clicking toggles the state. Then I can toggle back. This works in the envelope list and in the message view.

What exactly does not show for you?

ChristophWurst:
> Please notice on the top left the small arrow showing up when i click select or mark as spam

Have you compared this to main? I think it's caused by the popover menu.

src/components/Envelope.vue
JuliaKirschenheuter: for me works, besides icons inside of EnvelopePrimaryActions

GretaD: Sorry, i was testing this wrong, as i said on my first comment, when i toggle mark as span, the icon doesnt show on the avatar (not related to this pr). 
What you have done here is changing the left icon on the menu.

GretaD (approved):
all the icons are shown correctly

JuliaKirschenheuter (approved):

PR #6515, opened 2022-05-20
Update column definition for oci compatibility

kesselb:
I met our old friend `Column "oc_mailbox.selectable" is type Bool and also NotNull, so it can not store "false"` when installing Mail 1.11.7 on Nextcloud 23 :see_no_evil:


ChristophWurst (approved):

kesselb:
Rebased for a fresh CI run :broom:

PR #6413, opened 2022-05-11
Use UID ranges for IMAP search instead of unique IDs

miaulalala:
Fixes #6038

I would like some performance evaluation for this too.

PR #6529, opened 2022-05-23
Use UID ranges for IMAP fetch operations during sync

ChristophWurst:
This does not show any measurable performance improvement but prevents
the issue that the IMAP command can be too long for some IMAP servers
otherwise.

This is similar to https://github.com/nextcloud/mail/pull/6413 but for another section of our code.

Ref https://github.com/nextcloud/mail/commit/e22742731adc1c1c03c394029948c7fa67e31b74


lib/IMAP/MessageMapper.php
```diff
@@ -242,14 +243,14 @@ public function findByIds(Horde_Imap_Client_Base $client,
...
+			'ids' => $ids,
 		]), false);
 
 		if (empty($fetchResults)) {
 			$this->logger->debug("findByIds in $mailbox got " . count($ids) . " UIDs but found none");
 		} else {
-			$minRequested = $ids[0];
-			$maxRequested = $ids[count($ids) - 1];
+			$minRequested = $ids->ids[0];
+			$maxRequested = $ids->ids[count($ids) - 1];
```
tcitworld, line 253: Possibly?

```suggestion
			$maxRequested = $ids->count() - 1;
```

ChristophWurst: Let's meet in the middle with `$ids->ids[$ids->count() - 1];`? count-1 would not give me the value of the last element.

kesselb:
```
 * @property-read string $max  The largest ID (@since 2.20.0).
 * @property-read string $min  The smallest ID (@since 2.20.0).
 * @property-read string $range_string  Generates a range string consisting of
 *                                      all messages between begin and end of
 *                                      ID list.
```

It seems that Horde_Imap_Client_Ids has max, min and even a range_string out of the box ;)

ChristophWurst: How many engineers does it take to …

PR #6544, opened 2022-05-24
Add attachments objects on message forward

kesselb:
~Blocked by https://github.com/nextcloud/mail/pull/6543~

On message forward the attachment objects are missing. 

Draft because I'm unhappy how we mix Nextcloud with IMAP ids (attachment.databaseId) to reference the attachment. Still looking for a better approach.

### DIFF
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -70,7 +70,8 @@ import {
 	removeEnvelopeTag,
 	setEnvelopeFlag,
 	setEnvelopeTag,
-	syncEnvelopes, updateEnvelopeTag,
+	syncEnvelopes,
+	updateEnvelopeTag,
 } from '../service/MessageService'
 import * as AliasService from '../service/AliasService'
 import logger from '../logger'
@@ -331,6 +332,13 @@ export default {
 						body: data.body,
 						originalBody: data.body,
 						forwardFrom: reply.data,
+						attachments: original.attachments.map(attachment => ({
+							...attachment,
+							mailboxId: original.mailboxId,
+							// messageId for attachments is actually the uid
+							uid: attachment.messageId,
+							type: 'message-attachment',
+						})),
 					},
 				})
 				return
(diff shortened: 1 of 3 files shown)


ChristophWurst:
We should definitely send IMAP message-ids to the backend. The reasoning behind it is that the cache should remain volatile so we can clear it at any time. That would reassign new database IDs to a cached messge. The IMAP message-id, however, would always point to the same message.

ChristophWurst:
``\OCA\Mail\Service\Attachment\AttachmentService::handleForwardedAttachment`` should work with those IMAP message-ids instead of DB ids.

src/store/actions.js
```diff
@@ -331,6 +331,13 @@ export default {
 						body: data.body,
 						originalBody: data.body,
 						forwardFrom: reply.data,
+						attachments: original.attachments.map(attachment => ({
+							fileName: attachment.fileName,
+							id: attachment.id,
+							mailboxId: original.mailboxId,
+							uid: original.uid,
+							type: 'message-attachment',
+						})),
```
ChristophWurst, line 340:
```suggestion
						attachments: original.attachments.map(attachment => ({
							...attachment,
							mailboxId: original.mailboxId,
							type: 'message-attachment',
						})),
```

src/store/actions.js
```diff
@@ -20,8 +20,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import flatMapDeep from 'lodash/fp/flatMapDeep'
-import orderBy from 'lodash/fp/orderBy'
+import flatMapDeep from 'lodash/fp/flatMapDeep';
```
kesselb, line 23: Something seems wrong with my local eslint :see_no_evil:

src/store/actions.js
```diff
@@ -331,6 +332,12 @@ export default {
 						body: data.body,
 						originalBody: data.body,
 						forwardFrom: reply.data,
+						attachments: original.attachments.map(attachment => ({
+							...attachment,
+							mailboxId: original.mailboxId,
+							uid: original.messageId,
```
kesselb, line 338: Does not work ... because the original.messageId is the real message it but attachment.messageId is actually the uid :confused: :confused: :confused:

ChristophWurst:
backport?

PR #6549, opened 2022-05-25
Fix modal width

GretaD:


GretaD:
fyi, all modal on mobile have square corners

ChristophWurst (approved):

PR #6551, opened 2022-05-25
Move encrypted label to icon

sazanof:

Hello! 
I ask you to consider transferring the "encrypted" label. There is more free space if there are no other labels. When hovering - a hint. Maybe there are other suggestions. Maybe it will be useful.
Thanks!

it was

image

became

image
====

it was

image

became

image

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -50,6 +50,11 @@
 			</div>
 		</template>
 		<template #subtitle>
+			<span v-if="isEncrypted"
+				class="icon-encrypted"
+				:title="t('mail','Message encrypted.')">
+				<LockCheckOutline :size="20" fill-color="#98b379" />
+			</span>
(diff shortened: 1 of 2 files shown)


ChristophWurst:
> _I probably should have asked in a public group first before sending a pr_
> Do you think it's worth closing PR?

The group channel is great for ad hoc help and quick questions but I think this enhancement should be discussed in a feature request ticket here on github. Could you please open one at https://github.com/nextcloud/mail/issues/new?assignees=&labels=enhancement%2C0.+to+triage&template=feature_request.yml?

jancborchardt:
I’d say we don’t need to do anything special here, except filter out the tag as per https://github.com/nextcloud/mail/issues/6561? Ideally, the experience whether a mail is encrypted or not should be the same, and if you have Mailvelope installed, the mail will just be shown.

PR #6558, opened 2022-05-25
Add PHP8.2 to CI matrix

ChristophWurst:
Proof of concept for https://github.com/nextcloud/groupware/issues/40.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,8 @@ jobs:
               nextcloud-versions: stable24
             - php-versions: 8.1
               nextcloud-versions: master
+            - php-versions: 8.2
+              nextcloud-versions: master
     name: Nextcloud ${{ matrix.nextcloud-versions }} php${{ matrix.php-versions }} unit tests
     steps:
     - name: Set up php${{ matrix.php-versions }}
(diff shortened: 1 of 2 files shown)


ChristophWurst:
I didn't expect this to "just work" but let's go

kesselb:
Pushed a commit to patch the version check before nextcloud setup.

.github/workflows/test.yml
```diff
@@ -128,6 +136,9 @@ jobs:
                 coverage: xdebug
           - name: Checkout Nextcloud
             run: git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b ${{ matrix.nextcloud-versions }} nextcloud
+          - name: Patch version check for nightly PHP
+                  if: ${{ matrix.php-versions == '8.2' }}
```
kesselb, line 140: That looks off ;)

ChristophWurst: While fixing it squash the commits right away. I think CI should pass now

ChristophWurst:
> App "Mail" cannot be installed because the following dependencies are not fulfilled: PHP with a version lower than 8.1 is required.

we have to patch info.xml

kesselb:
> App "Mail" cannot be installed because the following dependencies are not fulfilled: PHP with a version lower than 8.1 is required.

Added another commit to patch max-version in appinfo.

Issue #6540, opened 2022-05-24
E-mail is re-sent again on each background job update


snyssen:
### Steps to reproduce

Unfortunately this issue seems to have spawned randomly as I use the mail client everyday and it has just happened once without me being able to pinpoint the actual cause.

1. Send an e-mail normally
2. Close the web page maybe too soon ? It might be the only thing that changed from what I usually do.

### Expected behavior

The e-mail is sent only once.

### Actual behavior

The same e-mail seems to get re-sent on every Nextcloud background job. It was sent precisely each 5 minutes, which corresponds to the frequency of the cron job I set to manage NextCloud.

### Mail app version

1.12.0 (I think, I can't look right now)

### Mailserver or service

mailfence

### Operating system

Fedora

### Database

PostgreSQL

### Additional info

I had to shutdown my Nextcloud instance for now as it was spamming my recipient. Is there any way I can cancel an "in transit" e-mail ? I am guessing it gets sent to a queue somewhere but isn't removed correctly afterwards, which would explain the loop ?

ChristophWurst:
As a user you should be able to see the message in your *Outbox* and delete it from there.

snyssen:
Okay here is the log:
```json
{
    "reqId": "iYOLdVI0oCnd6sll2719",
    "level": 2,
    "time": "2022-05-24T09:10:17+00:00",
    "remoteAddr": "",
    "user": "snyssen",
    "app": "mail",
    "method": "",
    "url": "--",
    "message": "Could not send outbox message 13: Could not send message: Error when communicating with the server.",
    "userAgent": "--",
    "version": "22.2.6.2",
    "exception": {
        "Exception": "OCA\\Mail\\Exception\\ServiceException",
        "Message": "Could not send message: Error when communicating with the server.",
        "Code": 2,
        "Trace": [
            {
                "file": "/var/www/html/custom_apps/mail/lib/Service/MailTransmission.php",
                "line": 271,
                "function": "sendMessage",
                "class": "OCA\\Mail\\Service\\MailTransmission",
                "type": "->",
                "args": [
                    {
                        "__class__": "OCA\\Mail\\Model\\NewMessageData"
                    },
                    null,
                    null
                ]
            },
            {
                "file": "/var/www/html/custom_apps/mail/lib/Service/OutboxService.php",
                "line": 129,
                "function": "sendLocalMessage",
                "class": "OCA\\Mail\\Service\\MailTransmission",
                "type": "->",
                "args": [
                    {
[...]

PR #6593, opened 2022-05-31
Fix collecting addresses in the background

ChristophWurst:
Nextcloud can inject the userId but only for web processes. Now that
messages are sent also in the background and not just from a web request,
the user ID can be null and make the address collection log an
exception.

Following the pattern that user ID should only be injected into a
controller, the user ID is now passed as argument of the collector,
making it suitable for background jobs too.

Ref https://github.com/nextcloud/mail/issues/6540#issuecomment-1141180344 for the original report.


nickvergessen (approved):
Looks like it could fix it

PR #4411, opened 2021-01-26
Load more message(s) on delete

StCyr:
fixes https://github.com/nextcloud/mail/issues/863

PR #6637, opened 2022-06-03
Migrate from Mocha to jest

ChristophWurst:
This will hopefully open the possibility for more Vue component tests.

Ref https://v2.vuejs.org/v2/guide/testing.html
Ref https://cli.vuejs.org/core-plugins/unit-jest.html#transform-dependencies-from-node-modules
Ref https://github.com/nextcloud/groupware/issues/29

To do
- [x] Make things work
- [x] Drop sinon leftovers
- [x] Drop Chai leftovers

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -196,10 +196,10 @@ jobs:
           - name: Set up Node
             uses: actions/setup-node@v3
             with:
-                node-version: 12.x
+                node-version: 14.x
           - name: npm install
             run: npm install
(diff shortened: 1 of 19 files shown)


ChristophWurst:
Ironically we had some tests that were close to a NOP due to the indirection of `fetchNextEnvelopePage` to `fetchNextEnvelopes` from https://github.com/nextcloud/mail/pull/4411.

JuliaKirschenheuter (approved):

Issue #6622, opened 2022-06-01
Replies and forwards not quoted anymore


ChristophWurst:
### Steps to reproduce

1. Open a message
2. Reply/forward

### Expected behavior

The original body is quoted

### Actual behavior

The original body is not quoted

### Mail app version

1.14

### Additional info

I can _not_ trigger this with v1.13.0 RC1

ChristophWurst:
@GretaD could you have a look please as you also debugged https://github.com/nextcloud/mail/issues/6564?

GretaD:
I cannot reproduce it on latest anymore. Any special condition on your local env?

ChristophWurst:
I can trigger it for replies on plain text and HTML message, with plain or HTML editing mode as default.

ChristophWurst:
I can reproduce with HTML and plain text emails on v1.13.0

ChristophWurst:
So this must have been introduced somewhere between RC1 and .0. Ref https://github.com/nextcloud/mail/commits/stable1.13

ChristophWurst:
Replying to an email with formatting also drops the formatting.

kenpyfin:
Any follow-up on this issue. We have to stop using the mail app because of this issue. It's major enough that it makes the mail app not usable.

ChristophWurst:
Fix is at https://github.com/nextcloud/mail/pull/6663.

kenpyfin:
Thank you for the quick fix!! Appreciate everyone's effort.

PR #6663, opened 2022-06-07
Fix missing Composer.vue::initBody call after mount

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6622

Regression of https://github.com/nextcloud/mail/pull/6596 due to an incorrect revert. 

13896384e2ffd86600f2ce984036f8d660d72b8f moved the method call into a deferred callback, the revert https://github.com/nextcloud/mail/pull/6596/commits/db2b08c29d9bd37d94eac7145933de068c8ddbc9 fully removed the method call. Therefore we never initialized the body.

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -752,6 +752,7 @@ export default {
 	},
 	async beforeMount() {
 		this.setAlias()
+		this.initBody()
 		await this.onMailvelopeLoaded(await getMailvelope())
 	},
 	mounted() {


kesselb:
Reply works.

Forward also works but it's not taking the right path. `this.forwardFrom` sems undefind when forwarding a message and the note with datetime is missing.

GretaD (approved):

ChristophWurst:
> Forward also works but it's not taking the right path. `this.forwardFrom` sems undefind when forwarding a message and the note with datetime is missing.

Right. Investigating.

ChristophWurst:
Fixed with https://github.com/nextcloud/mail/pull/6665.

ChristophWurst:
That is fine because https://github.com/nextcloud/mail/pull/6596 hasn't been backported to stable1.12.

PR #6665, opened 2022-06-07
Fix quoting forwarded messages

ChristophWurst:
Ref https://github.com/nextcloud/mail/pull/6663#issuecomment-1148630058.

Regression of https://github.com/nextcloud/mail/pull/2854.

Composer.vue received a new prop `forwardFrom` that never got filled. Therefore neither reply nor forward quoting was used in Composer.vue::initBody.


GretaD (approved):
tested and it works

ChristophWurst:
/backport 4063d19bad9d7c08144df098cf4118c4178e95cc to stable1.13

PR #6671, opened 2022-06-07
Migrate reply and reply all icon to material design

GretaD:
ref https://github.com/nextcloud/groupware/issues/38

before
buttonold

after
afterbutton

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -72,16 +72,18 @@
 			</router-link>
 			<div class="right">
 				<Moment class="timestamp" :timestamp="envelope.dateInt" />
-				<button
-					:class="{
-						'icon-reply-all-white': hasMultipleRecipients,
-						'icon-reply-white': !hasMultipleRecipients,
-						primary: expanded,
-					}"
-					class="button"
+				<Button
+					:class="{ primary: expanded}"
+					type="primary"
 					@click="onReply">
+					<template #icon>
+						<ReplyAllIcon v-if="hasMultipleRecipients"
+							:size="20" />
+						<ReplyIcon v-else
+							:size="20" />
+					</template>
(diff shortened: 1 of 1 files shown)


ChristophWurst:
FYI https://github.com/nextcloud/server/pull/25774 requires https://github.com/nextcloud/server/pull/25774. This means we have a hard dependency on Nextcloud 23 with buttons from nextcloud-vue. If I'm not mistaken they will not render correctly with 22. Give it a try and if it works then proceed, otherwise we need to wait for the drop of Nextcloud 22.

GretaD:
> Give it a try and if it works then proceed, otherwise we need to wait for the drop of Nextcloud 22.

You mean to give it a try with Vue `<Button>` ?

ChristophWurst:
Is the button component a hard requirement for icons withing buttons?

GretaD:
> Is the button component a hard requirement for icons within buttons?

to answer your question, no its not a hard requirement, but in this case im stupid, i imported the Button with the wrong path and assumed its not released yet.

src/components/ThreadEnvelope.vue
```diff
@@ -71,16 +71,18 @@
...
 			<div class="right">
 				<Moment class="timestamp" :timestamp="envelope.dateInt" />
-				<button
-					:class="{
-						'icon-reply-all-white': hasMultipleRecipients,
-						'icon-reply-white': !hasMultipleRecipients,
-						primary: expanded,
-					}"
-					class="button"
+				<Button
```
ChristophWurst, line 74: did you make a screenshot of your Nextcloud 22 test by any chance? I wonder how the button renders

GretaD: yes, check the description

ChristophWurst:
Please squash

Issue #4580, opened 2021-02-22
wrong avatars on the widget "Wichtige E-Mail"


kinimodmeyer:
### Expected behavior

the displayed avatar on the email in the widget should be correct.

### Actual behavior
on the widget "Wichtige E-Mail" i see wrong avatars which are not related to the email/thread.
when i inspect the html i  can see the following in the src of the img: "https://../remote.php/dav/addressbooks/system/system/system/Database:XYZ.vcf?photo"
(XYZ is not part on this e-mail)

pressing F5 on the dashboard also shows different not related avatars on the same email.
seams like a random display behavior.

### Mail app

**Mail app version:** 1.8.3

ChristophWurst:
Dup of https://github.com/nextcloud/mail/issues/4412?

kinimodmeyer:
hm i don´t saw this wrong avatar handling on the email lists or inside a email.

ChristophWurst:
The other technical reason I could think of is that Vue tries to reuse unrelated avatars. But in theory we have a key on all of our envelopes … :thinking:

AutomCoding:
I have the same issue (favicon and text censored):
![screenshot_nextcloud](https://user-images.githubusercontent.com/40275136/152966282-fe2112e7-f49d-410d-bfa5-07d08333cbfb.png)

Messages 4, 5 and 7 (T. S.) are from the same person, but number 5 uses the favicon of number 6 (M. R.). Number 6 then uses the default avatar from their initials.

My gut feeling says it likely has something to do with consecutive duplicates. I can confirm that it behaves as intended in the Mail app itself.

PR #6686, opened 2022-06-08
Fix contacts photo integration picking wrong PHOTO entry

sazanof:
Hello everyone! =)

I found in 2 places that avatars are loaded incorrectly.

1) ~~In the list of recipients in the letter
**How to reproduce**: add 1 contact with the address all@example.com (in my case, this is a mailing group) and сall@example.com and set avatars for them
Write to all@example.com from your email
Open this email and see that the wrong avatar is in the recipient list. Click on "details" and see that in the list of "identical contacts" all@ and call@~~ 
Move to https://github.com/nextcloud/mail/pull/6714

2) In a message and in the list of message
**How to reproduce**: There should be 2 contacts with the same e-mail (one of them with an avatar). I have this bug when creating my adressbook + recent contacts (which appeared themselves earlier than my address book)
The problem is that if an array > 1 element can fall into the $result variable, then the first element is returned. if the photo is in the second element, the first one will be returned WITHOUT the photo.

Thank you!

It may also occur in https://github.com/nextcloud/mail/issues/4580

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -107,11 +107,13 @@ class ContactsIntegration {
 	 */
 	public function getPhoto(string $email) {
 		$result = $this->contactsManager->search($email, ['EMAIL']);
-		if (count($result) > 0) {
-			if (isset($result[0]['PHOTO'])) {
-				return $this->getPhotoUri($result[0]['PHOTO']);
+		foreach ($result as $contact) {
+			if (!isset($contact['PHOTO']) || empty($contact['PHOTO'])) {
+				continue;
 			}
+			return $this->getPhotoUri($contact['PHOTO']);
 		}
+
 		return null;
 	}
 


lib/Service/ContactsIntegration.php
```diff
@@ -106,10 +106,12 @@ public function getMatchingRecipient(string $term): array {
 	 * @return false|null|string
 	 */
 	public function getPhoto(string $email) {
-		$result = $this->contactsManager->search($email, ['EMAIL']);
+		$result = $this->contactsManager->search($email, ['EMAIL'], ['escape_like_param' => false]);
 		if (count($result) > 0) {
```
ChristophWurst, line 110: You can remove the `if` and just leave the `foreach`. This it will "skip" the loop if the array is empty.

lib/Service/ContactsIntegration.php
```diff
@@ -106,10 +106,12 @@ public function getMatchingRecipient(string $term): array {
...
-			if (isset($result[0]['PHOTO'])) {
-				return $this->getPhotoUri($result[0]['PHOTO']);
+			foreach ($result as $contact){
+				if (isset($contact['PHOTO']) && !empty($contact['PHOTO'])) {
```
ChristophWurst, line 112: Negate the condition and `continue` in the `if`, then continue with the early `return` after the `if` block.

ChristophWurst, line 112: Negate the condition and `continue` in the `if`, then continue with the early `return` after the `if` block.

lib/Service/ContactsIntegration.php
```diff
@@ -106,10 +106,12 @@ public function getMatchingRecipient(string $term): array {
 	 * @return false|null|string
 	 */
 	public function getPhoto(string $email) {
-		$result = $this->contactsManager->search($email, ['EMAIL']);
+		$result = $this->contactsManager->search($email, ['EMAIL'], ['escape_like_param' => false]);
```
ChristophWurst, line 109: Could you submit this change as a separate PR? I would like to have a closer look at this. The other change can go in earlier :)

ChristophWurst, line 109: Could you submit this change as a separate PR? I would like to have a closer look at this. The other change can go in earlier :)

ChristophWurst (changes requested):

lib/Service/ContactsIntegration.php
sazanof: I hope I got it right

lib/Service/ContactsIntegration.php
```diff
@@ -106,12 +106,14 @@ public function getMatchingRecipient(string $term): array {
 	 * @return false|null|string
 	 */
 	public function getPhoto(string $email) {
-		$result = $this->contactsManager->search($email, ['EMAIL']);
-		if (count($result) > 0) {
-			if (isset($result[0]['PHOTO'])) {
-				return $this->getPhotoUri($result[0]['PHOTO']);
+		$result = $this->contactsManager->search($email, ['EMAIL'], ['escape_like_param' => false]);
```
ChristophWurst, line 109: shouldn't this line be part of https://github.com/nextcloud/mail/pull/6714?

sazanof: I then clarified in the chat about the separation of pr (`getPhoto` In one PR, `doSearch` in another PR).

 Maybe we didn't understand each other? You probably wanted to "have a closer look at" `['escape_like_param' => false]` parameter ?

ChristophWurst: > Maybe we didn't understand each other? You probably wanted to "have a closer look at" `['escape_like_param' => false]` parameter ?

Exactly. I want to be sure the turned off escaping doesn't have unexpected side effects

ChristophWurst (approved):

lib/Service/ContactsIntegration.php
```diff
@@ -107,11 +107,13 @@ public function getMatchingRecipient(string $term): array {
 	 */
 	public function getPhoto(string $email) {
 		$result = $this->contactsManager->search($email, ['EMAIL']);
-		if (count($result) > 0) {
-			if (isset($result[0]['PHOTO'])) {
-				return $this->getPhotoUri($result[0]['PHOTO']);
+		foreach ($result as $contact){
```
ChristophWurst, line 110:
```suggestion
		foreach ($result as $contact) {
```

sazanof: Fixed. Space added.

ChristophWurst (approved):

PR #6689, opened 2022-06-08
Change sender position to be centered with the avatar and timestamps

### LINKED ISSUE #6621: Sender of thread message not vertically centered
### Steps to reproduce

1. Open a thread (can be one message only)
2. Look at the avatar/sender

### Expected behavior

Avatar and sender are on one line

### Actual behavior

Sender floats above the avatar base line.

Bildschirmfoto vom 2022-06-01 09-14-43

### Mail app version

1.13

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

@GretaD you worked on the thread subject logic. Can you tell why the sender isn't centered (anymore)?

GretaD:
**before**
before-nosubject-notag

**after**
no subjet no tag
nosubjetc

the last 2 are the same as before, this pr fixes only when theres no subject or tags the sender should be centered.
subject+tag
subject-tag-
tag, no subject
nosubject tag

fixes #6621

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -51,7 +51,8 @@
 				class="left"
 				:class="{seen: envelope.flags.seen}"
 				@click.native.prevent="$emit('toggleExpand', $event)">
-				<div class="sender">
+				<div class="sender"
+					:class="{ 'centered-sender': centeredSender }">
 					{{ envelope.from && envelope.from[0] ? envelope.from[0].label : '' }}
 				</div>
 				<div v-if="hasChangedSubject" class="subject">
@@ -198,6 +199,9 @@ export default {
 		cleanSubject() {
 			return this.envelope.subject.replace(/((?:[\t ]*(?:R|RE|F|FW|FWD):[\t ]*)*)/i, '')
 		},
+		centeredSender() {
+			  return (!this.hasChangedSubject || this.cleanSubject.length === 0) && this.tags.length === 0
+		},
 	},
 	watch: {
 		expanded(expanded) {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
1) I *think* that the avatar and sender should be on one line if there is no changing subject.
2) If the subject changes (and is shown) *or* there are tags assigned to a message, I would not move the sender down anymore.

In other words, I see the sender/subject/tags as one *block*. and that block should be aligned with the avatar. Does that make sense?

GretaD:
yeah, sounds good, also i think this is not a bug, its the same design as envelope list, we just removed the subject for threads and didn't change the position.

ChristophWurst:
could you please add before/after screenshots

ChristophWurst (approved):
Linter should complain about the indentation

src/components/ThreadEnvelope.vue
```diff
@@ -198,6 +199,9 @@ export default {
 		cleanSubject() {
 			return this.envelope.subject.replace(/((?:[\t ]*(?:R|RE|F|FW|FWD):[\t ]*)*)/i, '')
 		},
+		centeredSender() {
+			  return (!this.hasChangedSubject || this.cleanSubject.length === 0) && this.tags.length === 0
+				 },
```
ChristophWurst, line 204: something looks odd here

GretaD: yes? i dont see it :(

this was the only way how i could make it work for all the scenarios. Am i missing something, or the way how i wrote the conditions is wrong?

ChristophWurst: The indentation

ChristophWurst (approved):

PR #6714, opened 2022-06-09
Fix recipients in users-bubble

sazanof:
The second part of "Avatar" trylogy. Named "The Avatar: Way of lost recipients in user bubble"

Linked with https://github.com/nextcloud/mail/pull/6686

image

In the list of recipients in the letter
**How to reproduce**: add 1 contact with the address all@example.com (in my case, this is a mailing group) and сall@example.com and set avatars for them
Write to all@example.com from your email
Open this email and see that the wrong avatar is in the recipient list. Click on "details" and see that in the list of "identical contacts" all@ and call@

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -112,7 +112,6 @@ class ContactsIntegration {
 			}
 			return $this->getPhotoUri($contact['PHOTO']);
 		}
-
 		return null;
 	}
 
@@ -199,10 +198,12 @@ class ContactsIntegration {
 	/**
 	 * @param string[] $fields
 	 */
-	private function doSearch(string $term, array $fields): array {
+	private function doSearch(string $term, array $fields, bool $strictSearch): array {
 		$allowSystemUsers = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
 
-		$result = $this->contactsManager->search($term, $fields);
+		$result = $this->contactsManager->search($term, $fields, [
+			'strict_search' => $strictSearch
+		]);
 		$matches = [];
 		foreach ($result as $r) {
 			if (!$allowSystemUsers && isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
@@ -225,7 +226,7 @@ class ContactsIntegration {
 	 * @return array
 	 */
 	public function getContactsWithMail(string $mailAddr) {
-		return $this->doSearch($mailAddr, ['EMAIL']);
+		return $this->doSearch($mailAddr, ['EMAIL'], true);
 	}
 
 	/**
@@ -235,6 +236,6 @@ class ContactsIntegration {
 	 * @return array
 	 */
 	public function getContactsWithName(string $name) {
-		return $this->doSearch($name, ['FN']);
+		return $this->doSearch($name, ['FN'], false);
 	}
 }


sazanof:
Hello! Can I find out if PR is relevant?

Thanks!

lib/Service/ContactsIntegration.php
```diff
@@ -106,14 +106,12 @@ public function getMatchingRecipient(string $term): array {
 	 * @return false|null|string
 	 */
 	public function getPhoto(string $email) {
-		$result = $this->contactsManager->search($email, ['EMAIL']);
-		foreach ($result as $contact) {
-			if (!isset($contact['PHOTO']) || empty($contact['PHOTO'])) {
-				continue;
+		$result = $this->contactsManager->search($email, ['EMAIL'], ['escape_like_param' => false]);
```
ChristophWurst, line 109: there is a better way: `'strict_search' => true,`

we don't need nor want wildcard matches for the email address

lib/Service/ContactsIntegration.php
```diff
@@ -203,7 +201,7 @@ public function newContact(string $name, string $mailAddr, string $type = 'HOME'
 	private function doSearch(string $term, array $fields): array {
 		$allowSystemUsers = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
 
-		$result = $this->contactsManager->search($term, $fields);
+		$result = $this->contactsManager->search($term, $fields, ['escape_like_param' => false]);
```
ChristophWurst, line 204: this seems partly unnecessary

```patch
diff --git a/lib/Service/ContactsIntegration.php b/lib/Service/ContactsIntegration.php
index fa744e723..c88be5f70 100644
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -199,10 +199,12 @@ class ContactsIntegration {
        /**
         * @param string[] $fields
         */
-       private function doSearch(string $term, array $fields): array {
+       private function doSearch(string $term, array $fields, bool $strictSearch): array {
                $allowSystemUsers = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
 
-               $result = $this->contactsManager->search($term, $fields);
+               $result = $this->contactsManager->search($term, $fields, [
+                       'strict_search' => $strictSearch,
+               ]);
                $matches = [];
                foreach ($result as $r) {
                        if (!$allowSystemUsers && isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
@@ -225,7 +227,7 @@ class ContactsIntegration {
         * @return array
         */
        public function getContactsWithMail(string $mailAddr) {
-               return $this->doSearch($mailAddr, ['EMAIL']);
+               return $this->doSearch($mailAddr, ['EMAIL'], true);
        }
 
        /**
@@ -235,6 +237,6 @@ class ContactsIntegration {
         * @return array
         */
        public function getContactsWithName(string $name) {
-               return $this->doSearch($name, ['FN']);
+               return $this->doSearch($name, ['FN'], false);
        }
 }

```

ChristophWurst (changes requested):

lib/Service/ContactsIntegration.php
sazanof: yep, since  v 23.0 =)

lib/Service/ContactsIntegration.php
```diff
@@ -105,14 +105,14 @@ public function getMatchingRecipient(string $term): array {
...
-		foreach ($result as $contact) {
-			if (!isset($contact['PHOTO']) || empty($contact['PHOTO'])) {
-				continue;
+		$result = $this->contactsManager->search($email, ['EMAIL'], ['strict_search' => true]);
+		if (count($result) > 0) {
+			foreach ($result as $contact) {
+				if (isset($contact['PHOTO'])) {
+					return $this->getPhotoUri($contact['PHOTO']);
+				}
 			}
```
sazanof, line 114: I added this, because while testing note that if you create 2 identical contacts and the first one without a photo, and the second one with a photo, then the avatar is not loaded. 

I would still like to be able to find an avatar if there are duplicates of contacts and show it to the user. what do you think?

sazanof: @ChristophWurst

lib/Service/ContactsIntegration.php
sazanof: I updated the code

lib/Service/ContactsIntegration.php
ChristophWurst: The `if` is unecessary. Remove it. If there are no results the loop won't loop and `null` is returned.

The logic was and is that results with out a PHOTO are skipped. I don't see how you are changing the logic.

sazanof: Previously, a photo of the first element of the array ot null was return. 

thus, if there are >1 elements in the array and there is a photo in the second (third... etc.... ) element, we will still get null. 

I changed it, added a loop in which we go through all the elements exactly and return a photo if there is one, and null if no photo is found after iterating through the array

sazanof: I don't understand anything, for some reason I have an old version of the file. everything is already correct in the new MAIN )

ChristophWurst: I believe this was changed in one of your other PRs :)

sazanof: Exactly 😊💯
The confusion arose due to the fact that I split one PR into 3

sazanof: [d05f6a3](https://github.com/nextcloud/mail/pull/6714/commits/d05f6a3442f827b0962fee5f8b84cbc234ee2832)

ChristophWurst (approved):

sazanof:
Hooray!

Issue #6589, opened 2022-05-30
Composer is too narrow / small


marcoambrosini:
### Steps to reproduce

1. Compose

### Expected behavior

The composer is reasonably wide

### Actual behavior

The composer is unreasonably narrow 

<img width="1726" alt="Screenshot 2022-05-30 at 16 53 59" src="https://user-images.githubusercontent.com/26852655/171021029-795598c8-af1d-4b99-9388-cd84cd24eeaf.png">

 😅

marcoambrosini:
Leaving a couple of references here: gmail and Protonmail. They're maybe too big but I think that when composing, most of the screen should be occupied by the composer

<img width="1728" alt="Screenshot 2022-05-30 at 17 16 31" src="https://user-images.githubusercontent.com/26852655/171021571-c0c6a45b-9201-487a-878a-902c2e321b7b.png">
<img width="1728" alt="Screenshot 2022-05-30 at 17 17 27" src="https://user-images.githubusercontent.com/26852655/171021666-2f2a6f5e-4a8d-4314-8e5e-b752a08cebc8.png">

ChristophWurst:
@nimishavijay @jancborchardt what width shall we use?

szaimen:
Maybe the large modal size?

Bibo-Joshi:
Another comment in a similar direction is that the width of received mails is rather narrow and doesn't get wider even if you collapse the inbox column to a minimum. See e.g. the dead space around the mail body in the following picture:

![uiae](https://user-images.githubusercontent.com/22366557/174728095-c585d6e0-455a-49ba-922b-981d526ac7a4.png)

I would very much like to have the mail use the full available width (minus some padding) or at least have the possibility to adjust the width like for the inbox column.

I'm commenting this here as a "which fixed width should we use" discussion is apparently ongoing in the linked pr #6717. If I should open a new issue instead, please tell :)

jdaviescoates:
Yes, the ridiculously small pop-up compose window of the current version of Mail is why I cannot event begin to try it out.  It never used to be this bad (as you can still see from the screenshot used on https://apps.nextcloud.com/apps/mail ), but I find this current iteration to be completely unusable.

nimishavijay:
I agree, if it's not possible to interact with the background then it should be bigger. AFAIK the modal composer is a first step towards a popup composer like in Gmail which allows you to browse your mails when you are composing, there it makes sense to have a smaller composer. 
[...]

PR #6717, opened 2022-06-10
change the composer modal size to large

szaimen:
Close https://github.com/nextcloud/mail/issues/6589

| Before | After |
|---|---|
| image | image |

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -1,6 +1,6 @@
 <template>
 	<Modal
-		size="normal"
+		size="large"
 		:title="modalTitle"
 		@close="$emit('close', { restoreOriginalSendAt: true })">
 		<Composer


GretaD:
I would wait for Jan's or Nimisha's approval :)

jancborchardt (changes requested):
This would widen the composer window line length from an already wide ~80 characters per line to a huge ~125 characters per line.

The recommended line length is between 45 and 75 characters per line, see:
- https://en.wikipedia.org/wiki/Line_length
- https://baymard.com/blog/line-length-readability

So the composer modal should stay at normal size.

Remember our design is not governed by how much screen space there is, but by human cognition. ;)

marcoambrosini:
@jancborchardt keep in mind that our `large` modal size is not that large. If you look at the composer on a not zoomed in page you'll see it's really really tiny. Gmail's smaller composer is tiny too but they keep the rest of the page interactive while that's deployed, while we're blocking everything else.

To give a point of reference, a size large is the same as the app settings dialog. To me this would be pretty reasonable for Mail's composer. We could also bump the size of the text in the composer to 16 to make it more readable.

reference of a "large" modal window on a 16 inch screen with recommended scaling and no zoom in the web page  
Screenshot 2022-06-17 at 13 16 44

Bibo-Joshi:
> To give a point of reference, a size large is the same as the app settings dialog. To me this would be pretty reasonable for Mail's composer.

Personally I'm used to writing mails in full-page view such that fixing a window size seems undesirable to me. I'd like to have it at least adjustable by the user or even better provide the possibility to edit it on the right hand side as it used to be the case (preferably the default behavior for this would be a setting somewhere).

miaulalala:
I personally have to agree with Marco and Bibo-Joshi. The current modal window is super small. Can we keep the line lenght and increase font size while making the modal bigger, and possibly add an overlay? It feels fiddly.

marcoambrosini:
Oh btw @jancborchardt I just noticed that the composer uses 13px as font size. Is there any reason for that? 
If we bump it to 15px and use the large modal size I think we'd have a reasonable line length.

nimishavijay:
Also agreed, if the composer was a popup composer on the right like in Gmail then it makes sense to keep it small, but seems to me that there is no use of keeping a modal-type focused composer which is so small. If you add attachments that's even lesser of the body text that's not visible. I think it's a good idea to increase the size, and possibly the font size as well :)

szaimen:
I would but where do I adjust that?

ChristophWurst:
Finish or discard?

szaimen:
> Finish or discard?

I asked a question in the comment above :P

ChristophWurst:
@jancborchardt

ChristophWurst:
@jancborchardt please have a look. I think this still makes sense but I'll let you decide.

jancborchardt:
@ChristophWurst The feedback I left in https://github.com/nextcloud/mail/pull/6717#issuecomment-1234559544 still has to be addressed though?

ChristophWurst:
@GretaD 

1) Change ckeditor font size to 15px
2) Enter 80 wide characters and see if the line breaks with the current modal size
3) if not: done
4) if it breaks: increase modal size

GretaD:
i did 2 scenarios, which one should we consider?

Screenshot from 2023-09-29 17-30-12

miaulalala:
This is the big gmail window:

image

The main benefits here are:

1) all the content is hidden by a light grey overlay plus the size of the modal. My eyes do not get caught up on the number of unread emails in my inboxes or any notifications, leaving me free to focus on the task that I want to do **right now** - composing an email.

2) the area I can (accidentally) click on to close the modal is small but still there, making the whole UI feel more stable and less "accidentally closing my email even though I wasn't done" -  prone even though the mechanism is the same as ours.

3) The line lenght takes the whole page but it doesn't matter as I will use line breaks and new paraghraphs to structure my email anyway. For **writing**, the max amount of characters is thus irrelevant.

jancborchardt:
@GretaD you can test with a variety of 80-character lines (not only 2) and increase the modal size/width accordingly, as @ChristophWurst said?

@miaulalala again, the Gmail large window is not standard, we are going in circles here. If we would make it so wide by default (and as the only option) it would severely hamper legibility.

GretaD:
> @GretaD you can test with a variety of 80-character lines (not only 2) and increase the modal size/width accordingly, as @ChristophWurst said?
> 
the text size is 15px.
see:
Screenshot from 2023-10-05 10-30-44

I used: https://charactercounter.de/ to count characters. Spaces are considered characters too in the example above.

miaulalala:
> @miaulalala again, the Gmail large window is not standard, we are going in circles here. If we would make it so wide by default (and as the only option) it would severely hamper legibility.

I disagree that it's not standard. GMX has a massive input field, if oldschool. So does Yahoo Mail. Outlook online is smaller but still more than 80 chars in a line. Clients like Thunderbird do not limit line length either. 

The fiddlyness of the modal we currently have is seriously annoying. And again, legibility is not really an issue in a window that will be structured by my own input - unlike a window that has the structure determined from the other side.

ChristophWurst:
Closing because we will do https://github.com/nextcloud/mail/issues/9100 instead.

PR #6723, opened 2022-06-10
improve the attachment viewer

szaimen:
Signed-off-by: szaimen <szaimen@e.mail.de>


src/components/AttachmentImageViewer.vue
```diff
@@ -28,25 +26,16 @@ export default {
...
-
-	&--full {
-		.modal-container,
-		.modal__attachment {
-			max-width: 95vw !important;
-			max-height: 95vh !important;
-		}
+	.modal-container {
+		background-color: unset !important;
+		box-shadow: none !important;
```
szaimen, line 31: We could also decide to keep the background...

szaimen: Maybe makes sense espwcially for mobile...

GretaD: yes, we should keep the background otherwise logos with just text, it will e hard to tell.

GretaD:
This looks good, on mobile the image is top aligned instead of center

szaimen:
> on mobile the image is top aligned instead of center

Should it be centered on mobile?

GretaD:
> > on mobile the image is top aligned instead of center
> 
> Should it be centered on mobile?

yes, i would say so. It should looks as much similar to the file viewer as it can.

szaimen:
> yes, i would say so. It should looks as much similar to the file viewer as it can.

all right! I'll have a look at this!

szaimen:
This is now ready from my side. We will now only show the backgrond directly behind the image

src/components/AttachmentImageViewer.vue
```diff
@@ -28,25 +26,19 @@ export default {
...
```
szaimen, line 31: I decided to keep this since it looks better for a image viewer, imo.

szaimen:
@GretaD does this look good now? :)

miaulalala:
Can you maybe add some screenshots @szaimen ?

st3iny (approved):
Tested and works.

PR #6736, opened 2022-06-13
Work around breaking server change and QB expressions

ChristophWurst:
Expressions are now "query function" objects instead of strings.

Ref https://github.com/nextcloud/server/pull/31966


kesselb:
image

:thinking:

kesselb (approved):

miaulalala (approved):
Just what I need to get my PR working :+1:

PR #6750, opened 2022-06-16
Fix infinite scrolling on mobile

GretaD:
Using max-height as in contacts: https://github.com/nextcloud/contacts/pull/2382
looks like fixes the problem of loading in mobile. Tested in brave and FF with webconsole


src/components/MailboxThread.vue
```diff
@@ -272,4 +273,8 @@ export default {
 ::v-deep .button-vue--vue-secondary {
 	box-shadow: none;
 }
+.envelope-list {
+	max-height: calc(100vh - var(--header-height));
+	overflow: auto;
```
skjnldsv, line 278:
```suggestion
	overflow-y: auto;
```

ChristophWurst:
backport?

PR #6756, opened 2022-06-17
Migrate avatar icons for envelope list avatar and thread envelope avatar to material design

GretaD:


src/components/ThreadEnvelope.vue
```diff
@@ -412,20 +419,26 @@ export default {
...
+	.app-content-list-item-star.favorite-icon-style {
 		display: inline-block;
 		position: absolute;
-		margin-top: -2px;
-		margin-left: 27px;
+		margin-top: -24px;
+		margin-left: 28px;
 		cursor: pointer;
+		&:hover {
+			opacity: .4;
```
st3iny, line 429:
```suggestion
			opacity: .5;
```

To make it consistent with the important icon.

GretaD:
so, there is a problem when the message has tags and 'no subjet' and clean subject. It breaks the styling and i tried different combinations and its not working. So have to look into that

src/components/Envelope.vue
```diff
@@ -19,9 +19,11 @@
 				v-if="mailbox.isUnified && hasMultipleAccounts"
 				class="mail-message-account-color"
 				:style="{'background-color': accountColor}" />
-			<div
+			<Star
 				v-if="data.flags.flagged"
-				class="app-content-list-item-star icon-starred"
+				fill-color="#f9cf3d"
```
ChristophWurst, line 24: https://github.com/nextcloud/mail/pull/6633 let's do without

GretaD: then should we change also the important icon? This is also over the avatar so the theme doesnt influence the visibility

ChristophWurst: if only the color is used to distinguish between the two states then yes

GretaD: not really, on the avatar its only "orange", the other state is removing the favorite or important icon. 
On the menu it changed color when it was not favorite. 

So, you're proposing these 2 icons to be dark grey now?
favoriteimportant

ChristophWurst: I'm sorry, I thought the PR touched the envelope menu. The overlay icon should probably keep its color :+1:

ChristophWurst:
- [ ] BUG junk and star overlap

Bildschirmfoto vom 2022-06-28 08-50-48

GretaD:
> * [ ]  BUG junk and star overlap
> 
> Bildschirmfoto vom 2022-06-28 08-50-48

it has always been like that and we have done it on purpose IIRC because the scenario that a message is favorite and spam is very rare. Where should we position the spam if we want to change it?

ChristophWurst (approved):

PR #6770, opened 2022-06-21
Collapse quoted text in HTML message

sazanof:
**Hello everyone!**

> I will be glad if there is time to discuss my PR and if it will be some useful.

Based on https://github.com/nextcloud/mail/issues/3640

**In fact, I have encountered several (many =)) problems. It would be great if you could tell me something.**

A little preface: perhaps this is not the most successful implementation, but I have not found another way yet.

### Problems:

- "`OC not found`". I need to connect the translation in `new Vue()` instance and upload avatars functions. And when importing l10n or fetchAvatarUrlMemoized, they just don't work. Tell me how to connect them correctly?
- CSS syles & variables. Variables and styles are not loaded into the iframe. And importing from `core/css` gives an error, Tell me how I can include these styles correctly.
- Different clients always send quotes in different ways. I would like to beautifully get the mail, name, date from the line. But this is practically impossible.

I would like to hear opinions about the implementation, because although it is not urgent, it is a very pleasant improvement. I will be grateful for any help and glad for any advice! 

Thanks!

>  _PS: I mainly tested sending between accounts via NC. I tested a little through other email clients. It seems to be (underlined) working. But this is not accurate =)_

Well, where to without screenshots:

image
image

### DIFF
--- a/src/components/HtmlBlockQuote.vue
+++ b/src/components/HtmlBlockQuote.vue
@@ -1,0 +1,143 @@
+<template>
+	<div class="quotes">
+		<a class="btn" @click="showAllQuotes()">
+			<CommentQuote :size="18" />
+			Toggle {{ total }} quotes
+		</a>
+		<HtmlBlockQuoteItem :quote="quotes" @toggleQuote="toggleQuote(quotes)" />
+	</div>
+</template>
+
+<script>
+
+import HtmlBlockQuoteItem from './HtmlBlockQuoteItem'
+import CommentQuote from 'vue-material-design-icons/CommentQuote'
+
+const REGEXP_EMAIL = /([!#-'*+/-9=?A-Z^-~-]+(\.[!#-'*+/-9=?A-Z^-~-]+)*|(\[\]!#-[^-~ \t]|(\\[\t -~]))+")@([!#-'*+/-9=?A-Z^-~-]+(\.[!#-'*+/-9=?A-Z^-~-]+)*|\[[\t -Z^-~]*])/
+const REGEXP_LABEL = /"(.*)"/
+
+export default {
+	name: 'HtmlBlockQuote',
+	components: {
+		HtmlBlockQuoteItem,
+		CommentQuote,
+	},
+	props: {
+		quote: {
+			type: HTMLElement,
+			required: true,
+		},
+		blockquote: {
+			type: HTMLElement,
+			required: true,
+		},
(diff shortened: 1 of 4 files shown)


ChristophWurst:
> Different clients always send quotes in different ways. I would like to beautifully get the mail, name, date from the line. But this is practically impossible.

I think that will always be a bit of catch-up. I suggest we abstract the parsing/detection a bit and put it into a testable function. That function can be unit tested against a few common formattings of popular email clients.

src/components/HtmlBlockQuoteItem.vue
```diff
@@ -0,0 +1,135 @@
...
+	methods: {
+		toggleQuote(quote) {
+			quote.hide = !quote.hide
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+	.quote {
```
ChristophWurst, line 69: nit: restructure to use BEM https://scalablecss.com/bem-blocks-within-blocks/

src/components/HtmlBlockQuote.vue
```diff
@@ -0,0 +1,143 @@
...
+		<HtmlBlockQuoteItem :quote="quotes" @toggleQuote="toggleQuote(quotes)" />
+	</div>
+</template>
+
+<script>
+
+import HtmlBlockQuoteItem from './HtmlBlockQuoteItem'
+import CommentQuote from 'vue-material-design-icons/CommentQuote'
+
+const REGEXP_EMAIL = /([!#-'*+/-9=?A-Z^-~-]+(\.[!#-'*+/-9=?A-Z^-~-]+)*|(\[\]!#-[^-~ \t]|(\\[\t -~]))+")@([!#-'*+/-9=?A-Z^-~-]+(\.[!#-'*+/-9=?A-Z^-~-]+)*|\[[\t -Z^-~]*])/
```
ChristophWurst, line 16: this is the part I'd like to see abstracted and tested :laughing:

ChristophWurst:
friendly ping @sazanof

sazanof:
4 packets received, 0 packets lost.
Is this PR actual? What do you think?

sazanof:
Well, there will be something to do =)

ChristophWurst:
I'm going to close this PR due to inactivity.

This can be resumed at any time as the previous changes are available here.

Thanks a lot @sazanof

Issue #6459, opened 2022-05-16
Wrong initial value (NaN) for prop "checked" for scheduled sending radio button


JuliaKirschenheuter:
### Steps to reproduce

1. Click to "New message"
2. Choose "Send later"
3. See the wrong initial value for radio button

### Expected behavior

No error, Boolean should be given to scheduled sending radio button

### Actual behavior

Invalid prop: type check failed for prop "checked". Expected Boolean, got Number with value NaN.

### Mail app version

1.12.0

### Mailserver or service

not relevant 

### Operating system

not relevant

ChristophWurst:
This still pops up when I click *Send later*. Could you please have a look?

PR #6799, opened 2022-06-27
Fix wrong initial value (NaN) for prop "checked"

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/mail/issues/6459


st3iny (approved):
Tested and works. The error message is gone.

PR #6768, opened 2022-06-20
Revert "Users can clear their mailboxes"

miaulalala:
Reverts nextcloud/mail#6286

Still waiting for feedback from @ChristophWurst to see if the bypassing of the trashbin is ok (Context: https://github.com/nextcloud/mail/pull/6286#pullrequestreview-1010395565)

If it is, we can close this PR.

PR #6800, opened 2022-06-28
Users can clear their mailboxes (Updated)

sazanof:
**Hello!** 
Now when user clear mailbox and if this mailbox not equals Trash, messages move to trash firstly

Referenced https://github.com/nextcloud/mail/pull/6286 , https://github.com/nextcloud/mail/pull/6768#issuecomment-1167040130

Fixes https://github.com/nextcloud/mail/issues/5858

### DIFF
--- a/lib/Service/MailManager.php
+++ b/lib/Service/MailManager.php
@@ -584,6 +584,39 @@ class MailManager implements IMailManager {
 	}
 
 	/**
+	 * Clear messages in folder
+	 *
+	 * @param Account $account
+	 * @param Mailbox $mailbox
+	 *
+	 * @throws DoesNotExistException
+	 * @throws Horde_Imap_Client_Exception
+	 * @throws Horde_Imap_Client_Exception_NoSupportExtension
+	 * @throws ServiceException
+	 */
+	public function clearMailbox(Account $account,
+								  Mailbox $mailbox): void {
+		$client = $this->imapClientFactory->getClient($account);
+		$trashMailboxId = $account->getMailAccount()->getTrashMailboxId();
+		$currentMailboxId = $mailbox->getId();
+		try {
+			if (($currentMailboxId !== $trashMailboxId) && !is_null($trashMailboxId)) {
+				$trash = $this->mailboxMapper->findById($trashMailboxId);
+				$client->copy($mailbox->getName(), $trash->getName(), [
+					'move' => true
+				]);
+			} else {
+				$client->expunge($mailbox->getName(), [
+					'delete' => true
+				]);
+			}
+			$this->dbMessageMapper->deleteAll($mailbox);
+		} finally {
+			$client->logout();
+		}
+	}
+
+	/**
 	 * @param Account $account
 	 * @param Mailbox $mailbox
 	 * @param Message $message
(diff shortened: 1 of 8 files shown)


lib/Db/MessageMapper.php
```diff
@@ -1220,4 +1220,16 @@ public function resetInReplyTo(): int {
 			);
 		return $update->execute();
 	}
+
+	public function changeMailbox(int $currentId, int $destinationId): int {
+		$qb = $this->db->getQueryBuilder();
+
+		$update = $qb->update($this->tableName)
```
ChristophWurst, line 1227: this won't work. each IMAP message has a `UID` assigned, but that UID is only unique within the mailbox. If a message is moved from one mailbox to another, a new UID is assigned.

If you only update the mailbox_id, the message UIDs will point to other messages.

It is not trivial to figure out the new UIDs. Therefore I would suggest that we move the messages on IMAP and delete them from the source messages cache. Synchronization will re-add the messages in the trash bin with their new UIDs.

lib/Service/MailManager.php
```diff
@@ -585,6 +585,40 @@ public function deleteMailbox(Account $account,
...
+	 *
+	 * @throws DoesNotExistException
+	 * @throws Horde_Imap_Client_Exception
+	 * @throws Horde_Imap_Client_Exception_NoSupportExtension
+	 * @throws ServiceException
+	 */
+	public function clearMailbox(Account $account,
+								  Mailbox $mailbox): void {
+		$client = $this->imapClientFactory->getClient($account);
+		$trashMailboxId = $account->getMailAccount()->getTrashMailboxId();
```
ChristophWurst, line 602: ``$trashMailboxId`` can be `null`

ChristophWurst (changes requested):

lib/Service/MailManager.php
sazanof: So if null, then just delete permanently, right?

lib/Db/MessageMapper.php
sazanof: OK, just do after that `$this->dbMessageMapper->deleteAll($mailbox);`

lib/Service/MailManager.php
ChristophWurst: Yes :)

sazanof: This method is deleted

src/components/NavigationMailbox.vue
```diff
@@ -507,6 +519,29 @@ export default {
 				this.clearCache = false
 			}
 		},
+		clearMailbox() {
+			const id = this.mailbox.databaseId
+			OC.dialogs.confirmDestructive(
+				t('mail', 'All messages in folder will be deleted.'),
```
ChristophWurst, line 525:
```suggestion
				t('mail', 'All messages in mailbox will be deleted.'),
```

sazanof:
Good afternoon! =) I made changes. Do you have any other requests? 
If you have time, can you double-check the PR? Thanks!

sazanof:
@ChristophWurst hello! 
I fixed the conflicts! =) can you see when there will be time? Thank you and have a nice day!

ChristophWurst (approved):
Tested, works

ChristophWurst:
```
/home/runner/work/mail/mail/src/service/MailboxService.js
Error:   58:34  error  Missing space before function parentheses  space-before-function-paren
```

PR #6812, opened 2022-06-30
Add imip processing in the background

### LINKED ISSUE #6802: Process iMip responses

See #160 

Backend:
- [x] Provide backend parsing for body parts with  content type "text/calendar" in getBodyStructureData (REPLY, CANCEL)
- [x] Create database columns on oc_mail_messages that tells a BG job that there is a caldav response to process (REPLY, CANCEL) and one that is set to true if the scheduling has been processed already
- [x] refactor mail processing to also be done when syncing and not only on loading the envelope list
- [x] Create BG job that allow processing of reply and cancel emails
- [x] In this BG job, access the pulic calendar API and get all calendars - search via UID for the event
- [x] Check all security implications such as the recipient being the ORGANIZER etc as specified in the RFC and the write to the calendar

miaulalala:
Fixes #6802 

- [x] Requires https://github.com/nextcloud/server/pull/33001

Questions:

what happens when I have a very large mailbox where I never trigger the preview for messages below the fold?

### DIFF
--- a/lib/BackgroundJob/IMipMessageJob.php
+++ b/lib/BackgroundJob/IMipMessageJob.php
@@ -2,10 +2,10 @@
 
 declare(strict_types=1);
 
-/**
- * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+/*
+ * @copyright 2022 Anna Larch <anna.larch@gmx.net>
  *
- * @author 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @author 2022 Anna Larch <anna.larch@gmx.net>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -23,35 +23,25 @@ declare(strict_types=1);
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-namespace OCA\Mail\IMAP;
+namespace OCA\Mail\BackgroundJob;
 
-class MessageStructureData {
-	/** @var bool */
-	private $hasAttachments;
+use OCA\Mail\Service\IMipService;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\TimedJob;
 
-	/** @var string */
-	private $previewText;
+class IMipMessageJob extends TimedJob {
+	private IMipService $iMipService;
 
-	/** @var bool */
-	private $isImipMessage;
+	public function __construct(ITimeFactory $time,
+								IMipService $iMipService) {
+		parent::__construct($time);
 
-	public function __construct(bool $hasAttachments,
-								string $previewText,
-								bool $isImipMessage) {
-		$this->hasAttachments = $hasAttachments;
-		$this->previewText = $previewText;
-		$this->isImipMessage = $isImipMessage;
+		// Run once per hour
+		$this->setInterval(60 * 60);
+		$this->iMipService = $iMipService;
 	}
 
-	public function hasAttachments(): bool {
-		return $this->hasAttachments;
-	}
-
-	public function getPreviewText(): string {
-		return $this->previewText;
-	}
-
-	public function isImipMessage(): bool {
-		return $this->isImipMessage;
+	protected function run($argument): void {
+		$this->iMipService->process();
 	}
 }
--- a/lib/BackgroundJob/PreviewEnhancementProcessingJob.php
+++ b/lib/BackgroundJob/PreviewEnhancementProcessingJob.php
@@ -1,0 +1,96 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Anna Larch <anna.larch@gmx.net>
+ *
+ * @author Anna Larch <anna.larch@gmx.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Mail\BackgroundJob;
+
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\PreprocessingService;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\IJobList;
+use OCP\BackgroundJob\TimedJob;
+use OCP\IUserManager;
+use Psr\Log\LoggerInterface;
+use function sprintf;
+
+class PreviewEnhancementProcessingJob extends TimedJob {
+	private IUserManager $userManager;
+	private AccountService $accountService;
+	private LoggerInterface $logger;
+	private IJobList $jobList;
+	private PreprocessingService $preprocessingService;
+
+	public function __construct(ITimeFactory $time,
+								IUserManager $userManager,
+								AccountService $accountService,
+								PreprocessingService $preprocessingService,
+								LoggerInterface $logger,
+								IJobList $jobList) {
+		parent::__construct($time);
+
+		$this->userManager = $userManager;
+		$this->accountService = $accountService;
+		$this->logger = $logger;
+		$this->jobList = $jobList;
+		$this->preprocessingService = $preprocessingService;
+
+		$this->setInterval(3600);
+		$this->setTimeSensitivity(self::TIME_SENSITIVE);
+	}
+
+	/**
+	 * @return void
+	 */
+	public function run($argument) {
+		$accountId = (int)$argument['accountId'];
+
+		try {
+			$account = $this->accountService->findById($accountId);
+		} catch (DoesNotExistException $e) {
+			$this->logger->debug('Could not find account <' . $accountId . '> removing from jobs');
+			$this->jobList->remove(self::class, $argument);
+			return;
+		}
+
+		$user = $this->userManager->get($account->getUserId());
+		if ($user === null || !$user->isEnabled()) {
+			$this->logger->debug(sprintf(
+				'Account %d of user %s could not be found or was disabled, skipping preprocessing of messages',
+				$account->getId(),
+				$account->getUserId()
+			));
+			return;
+		}
+
+		$dbAccount = $account->getMailAccount();
+		if (!is_null($dbAccount->getProvisioningId()) && $dbAccount->getInboundPassword() === null) {
+			$this->logger->info("Ignoring preprocessing job for provisioned account that has no password set yet");
+			return;
+		}
+
+		$limitTimestamp = $this->time->getTime() - (60 * 60 * 24 * 14); // Two weeks into the past
+		$this->preprocessingService->process($limitTimestamp, $account);
+	}
+}
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -76,6 +76,12 @@ use function json_encode;
  * @method null|string getPreviewText()
  * @method void setUpdatedAt(int $time)
  * @method int getUpdatedAt()
+ * @method bool isImipMessage()
+ * @method void setImipMessage(bool $imipMessage)
+ * @method bool isImipProcessed()
+ * @method void setImipProcessed(bool $imipProcessed)
+ * @method bool isImipError()
+ * @method void setImipError(bool $imipError)
  */
 class Message extends Entity implements JsonSerializable {
 	private const MUTABLE_FLAGS = [
@@ -114,6 +120,9 @@ class Message extends Entity implements JsonSerializable {
 	protected $flagImportant = false;
 	protected $flagMdnsent;
 	protected $previewText;
+	protected $imipMessage = false;
+	protected $imipProcessed = false;
+	protected $imipError = false;
 
 	/** @var AddressList */
 	private $from;
@@ -152,6 +161,9 @@ class Message extends Entity implements JsonSerializable {
 		$this->addType('flagImportant', 'boolean');
 		$this->addType('flagMdnsent', 'boolean');
 		$this->addType('updatedAt', 'integer');
+		$this->addType('imipMessage', 'boolean');
+		$this->addType('imipProcessed', 'boolean');
+		$this->addType('imipError', 'boolean');
 	}
 
 	/**
@@ -316,6 +328,7 @@ class Message extends Entity implements JsonSerializable {
 			'inReplyTo' => $this->getInReplyTo(),
 			'references' => empty($this->getReferences()) ? null: json_decode($this->getReferences(), true),
 			'threadRootId' => $this->getThreadRootId(),
+			'imipMessage' => $this->isImipMessage(),
 			'previewText' => $this->getPreviewText(),
 		];
 	}
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -312,7 +312,6 @@ class MessageMapper extends QBMapper {
 				$qb1->setParameter('flag_notjunk', $message->getFlagNotjunk(), IQueryBuilder::PARAM_BOOL);
 				$qb1->setParameter('flag_important', $message->getFlagImportant(), IQueryBuilder::PARAM_BOOL);
 				$qb1->setParameter('flag_mdnsent', $message->getFlagMdnsent(), IQueryBuilder::PARAM_BOOL);
-
 				$qb1->execute();
 
 				$messageId = $qb1->getLastInsertId();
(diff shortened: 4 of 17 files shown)


lib/Migration/Version1140Date20220630113356.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2022 Your name <your@email.com>
+ *
+ * @author Your name <your@email.com>
```
ChristophWurst, line 8: hello your name, I'm christoph

lib/Migration/Version1140Date20220630113356.php
```diff
@@ -0,0 +1,61 @@
...
+			'default' => false,
+		]);
+		$messagesTable->addColumn('imip_processed', 'boolean', [
+			'notnull' => false,
+			'default' => false,
+		]);
+		$messagesTable->addColumn('imip_error', 'boolean', [
+			'notnull' => false,
+			'default' => false,
+		]);
```
ChristophWurst, line 58: something to be written down somewhere: write migrations in a way they can be run more than once.

Imagine adding the second column fails. Then oc_migrations *won't* have this migration and nextcloud will try to run it another time. Then you get an error because `imip_message` already exists. That is because DDL are not transactional.

So write it like
```
if (!$table->hasColumn(xyz) {
  $table->addColumn(...);
}
```

lib/Migration/Version1140Date20220630113356.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$messagesTable = $schema->getTable('mail_messages');
+		$messagesTable->addColumn('imip_message', 'boolean', [
+			'notnull' => false,
+			'default' => false,
```
ChristophWurst, line 49: does this require a post migration logic to reset `structure_analyzed` to trigger re-processing all messages and set `imip_message`?

lib/Db/Message.php
```diff
@@ -76,6 +76,12 @@
  * @method null|string getPreviewText()
  * @method void setUpdatedAt(int $time)
  * @method int getUpdatedAt()
+ * @method bool isImipMessage();
```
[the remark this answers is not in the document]
miaulalala: that would explain a few things

lib/Migration/Version1140Date20220630113356.php
ChristophWurst: optimization [brain blast](https://www.youtube.com/watch?v=k3q1UTRKow0): we can also optimize the backgroudn processing to only process messages of 2w of age max and similar. then we don't need to pre-process historic emails that don't make a difference anyway. something for a later stage of the PR.

lib/Migration/Version1140Date20220630113356.php
miaulalala: Good point, I will add this!

miaulalala:
The way we currently only load envelope data means we can never really process all messages as we don't know whether they contain an iMIP body part.

Suggestion 1: only process messages that are opened (where the loading of the body data means we would update the is_imip_field in the db and later pick it up in the background job)

Benefits: efficient, no overhead
Drawbacks: we can never be sure we have processed all iMIP messages in the mailboxes and might miss important ones. Since the processing itself is done hidden in a background job it could confuse users as to why certain messages were processed and others weren't.

Suggestion 2: get all new messages in a mailbox and  additionallyload their body content too in a background job (whether we extend an exisiting job or add a new one could be debatable - I would tend towards option 1). 

Benefit: we would definitely catch all messages
Drawback: loading data from IMAP is slow and could hit us straight into the performance.

I would prefer option 2. 

For the preformance issue I propose the following solution (thanks to @st3iny who helped me on this):

We could extend the horde cache and implement our own horde cache class that lets us cache body content, or we could use ICache. It is more work but it would enable us to cache IMAP messages making Mail a lot faster. Since IMAP messages are immutable, we could cache them indefinitely or until they are deleted - or we could set a cache limit of a couple of weeks or even a month, it would still help us.

I would not cache message bodies in the database.

It would also help with full text search for mail, etc...

kevincox:
I'm pretty strongly against 1 because it doesn't work well if you aren't using the nextcloud webmail. Users of other webmail, other IMAP clients or even just people who rarely check their email won't get these updates handled. For example if I am on my phone I don't use the nextcloud mail UI and I may want to check RSVPs to an event and have an up-to-date number without checking my mail.

So I still think 2 is best. However even that isn't perfect. For example if I get a message on my phone and quickly delete it IIUC this may not be processed since it wasn't in the inbox for any 5min background job interval. It is probably a good "baseline" option but maybe there could be an advanced option where the mailserver could post new messages to nextcloud for processing? This could be a simple filter rule to ensure that nothing is missed and that imip messages are processed in real time. However that is probably a later improvement.

miaulalala:
> I'm pretty strongly against 1 because it doesn't work well if you aren't using the nextcloud webmail. Users of other webmail, other IMAP clients or even just people who rarely check their email won't get these updates handled. For example if I am on my phone I don't use the nextcloud mail UI and I may want to check RSVPs to an event and have an up-to-date number without checking my mail.
> 
> So I still think 2 is best. However even that isn't perfect. For example if I get a message on my phone and quickly delete it IIUC this may not be processed since it wasn't in the inbox for any 5min background job interval. It is probably a good "baseline" option but maybe there could be an advanced option where the mailserver could post new messages to nextcloud for processing? This could be a simple filter rule to ensure that nothing is missed and that imip messages are processed in real time. However that is probably a later improvement.

Thanks for your input! I do tend to think option 2 is a good solution, if imperfect. But we're also additionally processing any opened messages anyway (that part already works) - it's just the unopened, undownloaded messages we need a way to deal with.

Good point on the delete, if you open an iMIP message, it might be a great idea to process it immediately... at least for a REPLY. Something to think about.

Unfortunately, mail servers are not capable of pushing changes to external clients. That is not a part of the IMAP standard, so the only way to work with an IMAP server is to ask the server for changes.

kevincox:
> Unfortunately, mail servers are not capable of pushing changes to external clients.

There is IDLE support, but IDK how well that work work for nextcloud. Definitely doesn't fit the current architecture with no long-running processes.

Also many support filtering and other tools that can send requests to other services. Not part of IMAP, but still a useful option. As I said it would be a nice future feature and the base polling serves as a common ground that works for all IMAP servers with no extra setup.

miaulalala:
> > Unfortunately, mail servers are not capable of pushing changes to external clients.
> 
> There is IDLE support, but IDK how well that work work for nextcloud. Definitely doesn't fit the current architecture with no long-running processes.
> 
> Also many support filtering and other tools that can send requests to other services. Not part of IMAP, but still a useful option. As I said it would be a nice future feature and the base polling serves as a common ground that works for all IMAP servers with no extra setup.

taking the discussion to the linked ticket as this is supposed to be for code reviews only ;)

miaulalala:
add another BG job that processes all messages until 2 weeks into the past for structure Analysis.

(should make mail app faster)

lib/Service/IMipService.php
```diff
@@ -0,0 +1,169 @@
...
+
+			// Check for accounts or mailboxes that no longer exist,
+			// no processing for drafts, sent items, junk or trash
+			if ($account === null
+				|| $mailbox->getSpecialUse() === '["sent"]'
+				|| $mailbox->getSpecialUse() === '["archive"]'
+				|| $mailbox->getSpecialUse() === '["junk"]'
+				|| $mailbox->getSpecialUse() === '["drafts"]'
+				|| $mailbox->getSpecialUse() === '["trash"]'
+			) {
```
ChristophWurst, line 121: the column is json encoded and if you have multiple uses on one mailbox your change would fail

HOWEVER

we don't use the IMAP special use for the sent/trash/drafts preference. we use account.*_mailbox_id. on IMAP there can be n mailboxes with the same use. There is no guaranteed correlation between the imap flags and our preferences other than we try to detect ours from what is found on imap.

You can see `\OCA\Mail\Service\Classification\ImportanceClassifier::getOutgoingMailboxes` for reference

miaulalala: kk, thanks for the info

miaulalala: Chnaged it to the mailbox id methods, plus one use case for archive.

lib/IMAP/MessageMapper.php
```diff
@@ -674,19 +678,26 @@ public function getBodyStructureData(Horde_Imap_Client_Socket $client,
...
 			$structure = $fetchData->getStructure();
-			foreach ($structure as $part) {
-				if ($part instanceof Horde_Mime_Part && $part->isAttachment()) {
+			foreach ($structure->getParts() as $part) {
+				if ($part->isAttachment()) {
 					$hasAttachments = true;
-					break;
+				}
+				$bodyParts = $part->getParts();
+				foreach ($bodyParts as $bodyPart) {
```
kesselb, line 689: Please add `/** @var Horde_Mime_Part $bodyPart */`

lib/IMAP/MessageMapper.php
```diff
@@ -674,19 +678,26 @@ public function getBodyStructureData(Horde_Imap_Client_Socket $client,
 		return array_map(function (Horde_Imap_Client_Data_Fetch $fetchData) use ($mailbox, $client) {
 			$hasAttachments = false;
 			$text = '';
+			$isImipMessage = false;
 
 			$structure = $fetchData->getStructure();
-			foreach ($structure as $part) {
-				if ($part instanceof Horde_Mime_Part && $part->isAttachment()) {
+			foreach ($structure->getParts() as $part) {
```
kesselb, line 684: Please add `/** @var Horde_Mime_Part $part */`

lib/Service/IMipService.php
```diff
@@ -0,0 +1,168 @@
...
+					$message->setImipProcessed(true);
+				}, $filteredMessages); // Silently drop from passing to DAV and mark as processed, so we won't run into these messages again.
+				$this->messageMapper->updateImipData(...$filteredMessages);
+				continue;
+			}
+
+			try {
+				$imapMessages = $this->mailManager->getImapMessagesForScheduleProcessing($account, $mailbox, array_map(function ($message) {
+					return $message->getUid();
+				}, $filteredMessages), true);
```
kesselb, line 131:
```suggestion
				}, $filteredMessages));
```

lib/Db/MessageMapper.php
```diff
@@ -501,6 +503,7 @@ public function updatePreviewDataBulk(Message ...$messages): array {
 					$message->getPreviewText() === null ? null : mb_strcut($message->getPreviewText(), 0, 255),
```
kesselb, line 503: This email [mail_5322.txt](https://github.com/nextcloud/mail/files/9354885/mail_5322.txt) uses a weird encoding. It breaks the updatePreviewDataBulk process for me and rollback all changes. 

The following code made it work. I'm not 100% sure if that's not a problem with my dev setup somehow. 
```
$previewText = $message->getPreviewText();
if ($previewText !== null) {
    $previewText = mb_strcut(mb_convert_encoding($previewText, 'UTF-8', 'UTF-8'), 0, 255);
}
```

kesselb (commented):
Don't forget to bump the version number to install the background job.

lib/Db/MessageMapper.php
miaulalala: I think this is a general issue with the encoding that we've been having for a while now, see #5834 or #6913 for example.

lib/Service/IMipService.php
```diff
@@ -0,0 +1,168 @@
...
+
+			// Check for accounts or mailboxes that no longer exist,
+			// no processing for drafts, sent items, junk or trash
+			if ($account === null
+				|| $account->getMailAccount()->getTrashMailboxId() === $mailbox->getId()
+				|| $account->getMailAccount()->getSentMailboxId() === $mailbox->getId()
+				|| $account->getMailAccount()->getDraftsMailboxId() === $mailbox->getId()
+				|| $mailbox->isSpecialUse(\Horde_Imap_Client::SPECIALUSE_ARCHIVE)
+			) {
+				array_map(function ($message) {
```
kesselb, line 121: That's a bit weird. It works because objects are passed by reference. I would prefer to store the new array or merge the line with `$this->messageMapper->updateImipData`

miaulalala:
just waiting for server tests now

miaulalala:
stable22 test failing is not surprising as this isn't backported

miaulalala:
@ChristophWurst I added a check on the IManager to see if the method `handleImipReply` exists so we can avoid the whole test suite exploding for the next however many releases.

Shamelessly stolen from https://github.com/nextcloud/calendar/blob/main/tests/php/unit/Service/Appointments/BookingServiceTest.php

lib/BackgroundJob/PreviewEnhancementProcessingJob.php
```diff
@@ -0,0 +1,101 @@
...
+		if (!is_null($dbAccount->getProvisioningId()) && $dbAccount->getInboundPassword() === null) {
+			$this->logger->info("Ignoring preprocessing job for provisioned account that has no password set yet");
+			return;
+		}
+
+		try {
+			$limitTimestamp = $this->time->getTime() - (60 * 60 * 24 * 14); // Two weeks into the past
+			$this->preprocessingService->process($limitTimestamp, $account);
+		} catch (Exception $e) {
+			$this->logger->error('Error processing messages', ['exception' => $e]);
```
ChristophWurst, line 98: misleading

lib/Db/MessageMapper.php
```diff
@@ -1232,4 +1277,45 @@ public function resetInReplyTo(): int {
...
+
+		return $this->findEntities($select);
+	}
+
+	/**
+	 * @return Message[]
+	 *
+	 * @throws \OCP\DB\Exception
+	 */
+	public function getUnanalysed(int $lastRun, array $mailboxIds): array {
```
ChristophWurst, line 1308:
```suggestion
	public function getUnanalyzed(int $lastRun, array $mailboxIds): array {
```

for consistency with the rest of the code

lib/Migration/Version1140Date20220630113356.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version1140Date20220630113356 extends SimpleMigrationStep {
```
ChristophWurst, line 34: this is only going into v2

lib/Service/MailManager.php
```diff
@@ -193,6 +194,35 @@ public function getImapMessage(Account $account,
...
+		try {
+			return $this->imapMessageMapper->findByIds(
+				$client,
+				$mailbox->getName(),
+				new Horde_Imap_Client_Ids($uids),
+				true
+			);
+		} catch (Horde_Imap_Client_Exception $e) {
+			throw new ServiceException(
+				'Could not load message',
```
ChristophWurst, line 217: plural

ChristophWurst: and include `$e->getMessage()`

lib/Service/PreprocessingService.php
```diff
@@ -0,0 +1,88 @@
...
+			return;
+		}
+		$mailboxIds = array_unique(array_map(function (Mailbox $mailbox) {
+			return $mailbox->getId();
+		}, $mailboxes));
+
+
+		$messages = $this->messageMapper->getUnanalysed($limitTimestamp, $mailboxIds);
+		if (empty($messages)) {
+			$this->logger->info('No structure data to analyse.');
```
ChristophWurst, line 71: this will show up a lot. lower to debugging or drop

lib/Service/PreprocessingService.php
```diff
@@ -0,0 +1,88 @@
...
+			$filteredMessages = array_filter($messages, function ($message) use ($mailbox) {
+				return $message->getMailboxId() === $mailbox->getId();
+			});
+
+			if (empty($filteredMessages)) {
+				continue;
+			}
+
+			$processedMessages = $this->previewEnhancer->process($account, $mailbox, $filteredMessages);
+			$this->logger->info('Processed ' . count($processedMessages) . ' messages for structure data for mailbox ' . $mailbox->getId());
```
ChristophWurst, line 85: debug

lib/BackgroundJob/PreviewEnhancementProcessingJob.php
ChristophWurst: Sure

ChristophWurst:
linter

lib/BackgroundJob/IMipMessageJob.php
```diff
@@ -0,0 +1,60 @@
...
+	private IMipService $iMipService;
+
+	public function __construct(ITimeFactory $time,
+								IMipService $iMipService) {
+		parent::__construct($time);
+
+		// Run once per hour
+		$this->setInterval(60 * 60);
+		/**
+		 * @todo remove checks with 24+
```
ChristophWurst, line 45: this can go right away or we remove it with the other jobs

lib/BackgroundJob/IMipMessageJob.php
```diff
@@ -0,0 +1,60 @@
...
+		 */
+		if (defined('\OCP\BackgroundJob\IJob::TIME_SENSITIVE') && method_exists($this, 'setTimeSensitivity')) {
+			$this->setTimeSensitivity(self::TIME_SENSITIVE);
+		}
+		$this->iMipService = $iMipService;
+	}
+
+	protected function run($argument): void {
+		// iMIP is NC25+
+		if (!method_exists(IManager::class, 'handleImipReply')) {
```
ChristophWurst, line 55: v2 is 25+ as well

ChristophWurst (approved):

PR #6830, opened 2022-07-04
Add Google OAuth support

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6454

Todo
- [x] Show *Sign in with Google*
- [x] Properly restructure setup procedure to work well with fully automatic, manual and mixed setups https://github.com/nextcloud/mail/pull/6831
- [x] Save Google's refresh and access token after successful redirection back
- [x] If necessary update access token before an IMAP client is built
- [ ] Make sure the browser doesn't block the Google popup
- [x] Check/fix account update procedure
- [x] Handle closed popup
- [ ] Handle failed/cancelled user consent
- [ ] Admin settings page UI
- [ ] Any other edge cases
- [ ] Test migration from app password to oauth
- [ ] Test migration from broken account (login password) to oauth

PR #6831, opened 2022-07-04
Split auto config and account creation

ChristophWurst:
Required to make https://github.com/nextcloud/mail/pull/6830 usable.

## Before

We send email and password off to the server and let the server do all the magic in one step. This process is expensive and slow. The user does not get *any* sense of progress despite the server doing hard work.

Technically, we need to do lots of checks that take time (IO wait) because PHP doesn't allow for concurrency.

This is what it looks like configuring a random Gmail account with wrong credentials:

[mail_auto_setup-2022-07-06_14.12.55.webm](https://user-images.githubusercontent.com/1374172/177548105-d43ca01a-97e1-4345-8823-da8ca47362ce.webm)

Note how long this takes and how useless the end result is. The user gets no insight into the actual problem.

## After

The process is split into smaller tasks that are handled from the front end. We are able to update the UI to keep the user informed about progress. This makes the loading *feel* less broken/stuck.

The auto discovery logic changed in the following ways
* If we find a Thunderbird auto config entry, then we trust it without running the user credentials against it. The database is well maintained.
* We only test *secure connections*. That is, SSL and STARTTLS ports/encryption. If the user wants to use an insecure connection then they have to use the manual configuration. These days that shouldn't happen too often. Even privately hosted email servers can get free certificates. Most people don't host their own email.

Thanks to JavaScript's concurrency we are able to test various configurations in parallel. Those are 6 tests right now. Each can take up to 5s. So it is 30s vs ~5s. A huge impact for the user.

[mail_auto_setup-2022-07-06_14.14.33.webm](https://user-images.githubusercontent.com/1374172/177548451-fadbe3d7-51f1-4a7c-8d90-399286af359b.webm)


JuliaKirschenheuter (approved):

miaulalala (commented):
Looks good, works pretty fast for my GMX account.

But the auto config dialogue changes don't propagate to the manual setup dialogue if you've clicked the Connect button once before and add more changes but then switch to the manual settings dialogue.

kesselb:
Case: The response for ispdb is `null` and mx is `[]`. It would be a nice enhancement to show a note. 

https://user-images.githubusercontent.com/3902676/177599789-b98e0be4-5fec-4364-8080-11f6bd5a73a9.mp4

kesselb:
Case: Update existing account and use a wrong password


https://user-images.githubusercontent.com/3902676/177602569-183408a5-b6ab-479d-897e-82e2d1cc643d.mp4

```
{
  "status": "fail",
  "data": {
    "message": "Updating account failed: Connection to IMAP at placeholder.com:993 failed. Authentication failed.",
    "type": "OCA\\Mail\\Exception\\ClientException"
  }
}
```

kesselb:
```
Status for the jail: dovecot
|- Filter
|  |- Currently failed:	0
|  |- Total failed:	12
|  `- File list:	/var/log/mail.log
`- Actions
   |- Currently banned:	1
   |- Total banned:	1
   `- Banned IP list:	my.ip.address

```

I wasn't sure if the fail2ban rules for dovecot are working on my server. After testing this PR I can confirm that they work :rofl:

kesselb (approved):
Awesome :+1: 

Adding a new account is so much better now :partying_face:
It's faster and we have a status indicator what's happening right now. 

Well done :muscle:

ChristophWurst:
> It would be a nice enhancement to show a note.

[mail_auto_setup-2022-07-07_10.41.48.webm](https://user-images.githubusercontent.com/1374172/177731162-c9942065-9c24-4929-85db-011a0b014715.webm)

Another tiny improvement: changing settings removes the error

[mail_auto_setup-2022-07-07_10.42.03.webm](https://user-images.githubusercontent.com/1374172/177731304-6a63a393-4d29-4e5f-8d34-0eb06728d49b.webm)

ChristophWurst:
> Update existing account and use a wrong password. Some feedback here would be nice

I think that is a leftover from https://github.com/nextcloud/mail/pull/6099. I'll unify the error handling because right now it's handled by the parent component of the form and one of them has no insightful feedback.

miaulalala:
The copying works now but I don't have any error messages displayed any more.

Pulled, ran nom ci && npm run build in Icognito Window, added the wrong password on purpose.


Peek 2022-07-07 13-41


Error log in the console:

```
setup:1 The resource http://nextcloud.local/apps/mail/js/mail.8.23f1591338ff0bdbcf36.js was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
api/accounts:1          Failed to load resource: the server responded with a status of 500 (Internal Server Error)
ConsoleLogger.js:41 [ERROR] mail: could not save account details Object
log @ ConsoleLogger.js:41
vue.runtime.esm.js:1897 TypeError: Cannot read properties of undefined (reading 'service')
    at AccountForm.vue?2cb1:515:1
    at l (mail.1.ca53df8e66ba9ae227eb.js:21:3462)
    at Generator._invoke (mail.1.ca53df8e66ba9ae227eb.js:21:3215)
    at Generator.throw (mail.1.ca53df8e66ba9ae227eb.js:21:3825)
    at _ (mail.1.ca53df8e66ba9ae227eb.js:21:8697)
    at s (mail.1.ca53df8e66ba9ae227eb.js:21:8936)
```

miaulalala (approved):
Error messages are back. Looks geat!

ChristophWurst:
> Error messages are back.

How far have we come to celebrate this

PR #6846, opened 2022-07-08
Migrate js license headers to SPDX license

ChristophWurst:
Extracted from the eslint upgrade in https://github.com/nextcloud/mail/pull/6842 for easier review.


st3iny (approved):
:broom: om:

PR #6847, opened 2022-07-08
Fix Vue custom even casing

ChristophWurst:
They should be kebab case.

For https://github.com/nextcloud/mail/pull/6842.

ESLint would otherwise complain about lots of wrong casings, like

```
mail/src/components/Composer.vue
  948:20  error  Custom event name 'insertSignature' must be kebab-case           vue/custom-event-name-casing
  980:19  error  Custom event name 'onAddLocalAttachment' must be kebab-case      vue/custom-event-name-casing
  984:19  error  Custom event name 'onAddCloudAttachment' must be kebab-case      vue/custom-event-name-casing
  988:19  error  Custom event name 'onAddCloudAttachmentLink' must be kebab-case  vue/custom-event-name-casing
```

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -486,14 +486,14 @@ export default {
 				if (!this.account) {
 					const account = await this.$store.dispatch('createAccount', data)
 					this.feedback = t('mail', 'Account created')
-					this.$emit('accountCreated', account)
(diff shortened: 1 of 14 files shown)


gary-kim (approved):
It is the recommended casing for Vue 2.

PR #6863, opened 2022-07-10
👌 IMPROVE: don't open first mail by default

### LINKED ISSUE #6526: Do not open first message by default

In my email workflow I leave messages unread if I have to take action later on. Right now the Mail app automatically opens the first message of my priority inbox, leaving that message as read even if I don't read it.

Do not open the message until I navigate to it.

N/a

### Additional context

There might be another ticket about this same request but I'm unable to find it right now.

cc @jancborchardt @nimishavijay

violoncelloCH:
fix #6526 

This removes the logic to automatically open the first message of a mailbox

In contrary to #2794, this also doesn't open the first mail, when the current folder is reloaded as this would be unexpected: If I open a mailbox, don't select any mail and click on the same mailbox again I don't want the first message to be opened at that time. (Also the current implementation would've lead to first messages being opened while navigation between mailboxes.)

There is one remaining bug: on the initial load of the priority mailbox, `this.$store.getters.getEnvelopes(this.mailbox.databaseId, this.searchQuery)` returns an empty list even though there are mails in the inbox. Therefore `hasEnvelopes` is false and the "No message selected" placeholder isn't displayed. After a few seconds (after the third or fourth `sync` request; but I'm not sure that's actually related) the `getEnvelopes` getter suddenly returns the correct answer and the view gets updated. I couldn't find out why this getter isn't working at initial load.
Any idea on how to debug this further ?

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -57,7 +57,6 @@
 						class="nameother"
 						:account="unifiedAccount"
 						:mailbox="unifiedInbox"
-						:open-first="false"
 						:search-query="appendToSearch(priorityOtherQuery)"
 						:is-priority-inbox="true"
 						:bus="bus" />
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -147,6 +147,10 @@ export default {
 			return this.$store.getters.getMailbox(UNIFIED_INBOX_ID)
 		},
 		hasEnvelopes() {
+			if (this.mailbox.isPriorityInbox) {
+				return this.$store.getters.getEnvelopes(this.mailbox.databaseId, this.appendToSearch(priorityImportantQuery)).length > 0
+					|| this.$store.getters.getEnvelopes(this.mailbox.databaseId, this.appendToSearch(priorityOtherQuery)).length > 0
+			}
 			return this.$store.getters.getEnvelopes(this.mailbox.databaseId, this.searchQuery).length > 0
 		},
 		hasImportantEnvelopes() {
@@ -167,7 +171,7 @@ export default {
 		query() {
 			if (this.$route.params.filter === 'starred') {
 				if (this.searchQuery) {
-					return this.searchQuery + ' is:starred'
+					return this.appendToSearch('is:starred')
 				}
 				return 'is:starred'
 			}
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -97,10 +97,6 @@ export default {
 			type: Number,
 			default: 20,
 		},
-		openFirst: {
-			type: Boolean,
-			default: true,
-		},
 		searchQuery: {
 			type: String,
 			required: false,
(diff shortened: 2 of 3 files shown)


ChristophWurst:
> Therefore `hasEnvelopes` is false and the "No message selected" placeholder isn't displayed. After a few seconds (after the third or fourth `sync` request; but I'm not sure that's actually related) the `getEnvelopes` getter suddenly returns the correct answer and the view gets updated. I couldn't find out why this getter isn't working at initial load.
> Any idea on how to debug this further ?

Sounds like a Vue reactivity bug I've fought before. Not sure how I solved it last time.

violoncelloCH:
Yes, indeed...
I did some further debugging and found that `getters.getMailbox(mailboxId).envelopeLists` initially only contains one entry: `"is:pi-other": [ 4, 3 ]` only later on a second one is addes `"": [ 4, 3 ]` (that's when `getEnvelopes` actually returns the list of envelopes instead of just an empty list... no idea though where that second entry comes from and why it's only added at that time and not initially

violoncelloCH:
Okay, some more results:
The problem seems to be that only the sync (with timeout 30*1000) in App.vue (this exactly matches the delay of 30 seconds I get for the "No message selected" message to appear)
https://github.com/nextcloud/mail/blob/916e98a4fbfc73b1933afb22baa4a5a8d9798e5a/src/App.vue#L41
triggers the async action `syncInboxes`
https://github.com/nextcloud/mail/blob/916e98a4fbfc73b1933afb22baa4a5a8d9798e5a/src/store/actions.js#L713-L715
which in turn dispatches `fetchEnvelopes` with `query` undefined
this then commits `addEnvelope` with query being empty/undefined:
https://github.com/nextcloud/mail/blob/916e98a4fbfc73b1933afb22baa4a5a8d9798e5a/src/store/actions.js#L467-L471
so only at that point, `mailbox.envelopeLists` actually contains an element with index being an empty string...

My question now is: Is that element of `envelopeLists` actually supposed to have an empty string as index ?
If yes, how should this be properly fixed ? 
- By doing a `this.$store.dispatch('syncInboxes')` at the initial loading of the page ?
- or just by doing `dispatch('fetchEnvelopes')` with undefined `query` parameter at initial load ?
- or is there a better way I don't see yet ?

violoncelloCH:
Thanks for the explanation !

>It is or is not displayed?

On the unified inbox, the placeholder is not displayed initially but only after 30 seconds, once there exists a list for `''`.
If I understand it correctly, the `hasEnvelope` lookup should query for `is:pi-other` instead of `''` in the case of the unified inbox. I'll check how to fix this.

src/components/MailboxThread.vue
```diff
@@ -146,6 +145,11 @@ export default {
 			return this.$store.getters.getMailbox(UNIFIED_INBOX_ID)
 		},
 		hasEnvelopes() {
+			if (this.mailbox.isPriorityInbox) {
+				return this.$store.getters.getEnvelopes(this.mailbox.databaseId, this.appendToSearch(priorityImportantQuery)).length > 0
+					|| this.$store.getters.getEnvelopes(this.mailbox.databaseId, this.appendToSearch(priorityStarredQuery)).length > 0
```
st3iny, line 150:
```suggestion
```

This is obsolete now (ref https://github.com/nextcloud/mail/pull/6967).

st3iny (commented):
Tested and works. Thanks a lot for your PR.

jancborchardt:
@violoncelloCH could you add a screenshot for easier design review? :) What would be great is an emptycontent screen like:
> mail icon
> **Welcome to Nextcloud Mail** (h2)
> Write a new message (as a button)

GretaD:
@jancborchardt  please see attached a preview of your comment and if i understood you right,
> mail icon
> **Welcome to Nextcloud Mail** (h2)
> Write a new message (as a button)


Screenshot from 2023-08-04 19-27-19

GretaD (approved):

GretaD:
Jan, sorry, it was green, and i just click the merge button. If theres still something to do here, i will push a fixup on Monday 🤦

PR #6868, opened 2022-07-13
Print newly created account ID on CLI

ChristophWurst:
This makes debugging life easier when you want to create and sync a test account only via CLI.


kesselb (approved):
:sunglasses:

PR #6871, opened 2022-07-13
Migrate composer action menu icons to material design icons + migrate to Vue buttons

GretaD:
ref https://github.com/nextcloud/groupware/issues/38

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -234,17 +234,26 @@
 			<div class="composer-actions--secondary-actions">
 				<Actions @close="isMoreActionsOpen = false">
 					<template v-if="!isMoreActionsOpen">
-						<ActionButton icon="icon-upload" @click="onAddLocalAttachment">
+						<ActionButton @click="onAddLocalAttachment">
+							<template #icon>
+								<IconUpload :size="20" />
+							</template>
 							{{
 								t('mail', 'Upload attachment')
 							}}
 						</ActionButton>
-						<ActionButton icon="icon-folder" @click="onAddCloudAttachment">
+						<ActionButton @click="onAddCloudAttachment">
+							<template #icon>
+								<IconFolder :sizse="20" />
+							</template>
 							{{
 								t('mail', 'Add attachment from Files')
 							}}
 						</ActionButton>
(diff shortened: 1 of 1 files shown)


src/components/Composer.vue
```diff
@@ -234,17 +234,26 @@
...
-						<ActionButton icon="icon-upload" @click="onAddLocalAttachment">
+						<ActionButton @click="onAddLocalAttachment">
+							<template #icon>
+								<IconUpload :size="20" />
+							</template>
 							{{
 								t('mail', 'Upload attachment')
 							}}
 						</ActionButton>
 						<ActionButton icon="icon-folder" @click="onAddCloudAttachment">
```
ChristophWurst, line 245: Icon still used

ChristophWurst (approved):

ChristophWurst (changes requested):
Why did you rename `button` to `Button`? the nc/vue button component is imported as `VButton`

Are the button changes even necessary for the icon replacement?

GretaD:
> Why did you rename `button` to `Button`? the nc/vue button component is imported as `VButton`
> 
> Are the button changes even necessary for the icon replacement?

no there are not necessary on this pr because <button> doesnt have an icon. When it has an icon then it needs <Button> I did it here because ive done it with other PRs

ChristophWurst (approved):
:+1: :elephant:

raimund-schluessler:
There is quite a bit of chaos with the buttons now, you have `button`, `Button` and `VButton`, whereas they all should be `VButton` (or how you want to import them) regardless of whether they have icons or not. `button` will only render to a standard HTML component and not profit from the accessibility features of the vue component.

`button` (standard HTML button with an icon even):
https://github.com/nextcloud/mail/blob/57ffa52197d78275ce4a94595128b9b85a92bfe7/src/components/Composer.vue#L61-L67

`Button` (probably also standard HTML with an icon):
https://github.com/nextcloud/mail/blob/57ffa52197d78275ce4a94595128b9b85a92bfe7/src/components/Composer.vue#L355-L363

`Button` without icon:
https://github.com/nextcloud/mail/blob/57ffa52197d78275ce4a94595128b9b85a92bfe7/src/components/Composer.vue#L377-L379

PR #6879, opened 2022-07-14
👌 IMPROVE: hide system tags from UI

### LINKED ISSUE #6020: Unexpected tag "Loadremoteimages" showing up in tag list
The tag "Loadremoteimages" shows up in the user-facing tag list even though it is just machine readable. Would be nicer if we prevent it from showing up there:

image

violoncelloCH:
fix #6020 

I've implemented the proposed solution with a list of keywords to filter out. 
I've also added a check on the creation of new tags to prevent people from adding tags which would not be visible afterwards.

However I'm not completely sure if we should compare `tag.displayName` or rather `tag.imapLabel` ?

Also the list of keywords might not be complete yet; I've just picked the ones from  #6020 (with exception of "Newsletter" as this seems to be a bit too generic to prohibit people from using it manually).

### DIFF
--- a/src/components/tags.js
+++ b/src/components/tags.js
@@ -1,0 +1,28 @@
+/**
+ * @copyright 2022 Jonas Sulzer <jonas@violoncello.ch>
+ *
+ * @author 2022 Jonas Sulzer <jonas@violoncello.ch>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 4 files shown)


src/components/tags.js
```diff
@@ -0,0 +1,28 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export const hiddenTags = [
```
ChristophWurst, line 22: an array search with `.includes` is O(n). If you make this an object (key=tag, value=anything) it becomes O(1)

violoncelloCH: I'm not sure if I see why this would change the performance. Wouldn't it still need to compare against all keys of the object (and therefore remain O(n))? As far as I understand, having the list of tags in a binary tree would allow for a faster lookup, but I'm not sure if that exists in javascript.

ChristophWurst: in my understanding objects are like maps or hash tables. we can assume that key access is as efficient as possible: https://stackoverflow.com/a/45267146/2239067

violoncelloCH: ahh that makes sense then; thanks for the explanation!
I'll change this accordingly.

PR #6885, opened 2022-07-15
Reduce memory footprint of loading thread data

ChristophWurst:
`fetchAll` is memory hungry. Using a loop and building the entity array iteratively lowers the peak memory consumption and therefore helps the threading process succeed with large accounts.

Very large accounts still run out of memory later on at ``/apps/mail/lib/IMAP/Threading/Container.php on line 102``. That will be the next OOM issue to fix.


kesselb (approved):

miaulalala:
Linter wasn't happy.

PR #6903, opened 2022-07-19
Fix whitespace insertion on mailbox names 

miaulalala:
which are causing issues when looking for folders.

The sync logic doesn't find the mailbox / folder in the DB (afaik MySQL/ MariaDB only) if the name was created with whitespace because of the way MySQL handles whitespace at the end of strings. (See https://github.com/nextcloud/mail/issues/5200)
This then leads to a unique constraint violation when trying to insert the mailbox.

### DIFF
--- a/lib/Db/Mailbox.php
+++ b/lib/Db/Mailbox.php
@@ -36,8 +36,6 @@ use function ltrim;
 use function strtolower;
 
 /**
- * @method string getName()
- * @method void setName(string $name)
  * @method int getAccountId()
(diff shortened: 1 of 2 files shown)


miaulalala:
This is a very debatable fix.

1) The issue will only fix the MySQL / MariaDB use case but might break DBs that handle whitespace differently.
2) it is debatable if renaming the IMAP mailbox is actually a good idea at all. Since we might not find the mailbox on IMAP by name if we ever need to.

Issue #6122, opened 2022-03-08
Improve the "Show images" message and styling


nimishavijay:
### Is your feature request related to a problem? Please describe.

If an email contains images, there is a message saying "The images have been blocked  to protect your privacy" with a "Show" button. However, the styling of this with the full sized button makes it even more present than the content of the email sometimes. 

### Describe the solution you'd like

Just a couple of small changes could make the message look better

Before:
![image](https://user-images.githubusercontent.com/52440189/157226110-8097b536-5957-4a7d-b897-2602fc86b185.png)
After:
![image](https://user-images.githubusercontent.com/52440189/157655473-71da56c6-7331-4b4d-a36b-6e14bd1f676a.png)

- text color is changed to `--color-text-maxcontrast`
- instead of a button, "Show images" has normal weight, no icon, and an underline to indicate that it is clickable.

miaulalala:
I'd like an underline tbh, just to accentuate that there is an action possible.

nimishavijay:
You're right! It was there in the specs, but not the mockup, fixed it!

marcoambrosini:
@nimishavijay @jancborchardt  this uses the actions component which is about to be overhauled by @raimund-schluessler. Can we wait till then, look at all the patterns that we want to introduce and decide if this is one of them? 
Also, should a dropdown menu open when clicking on something that's styled like a link? Could we use a `tertyary` or `tertiary-no-background` button instead?

marcoambrosini:
I hacked the vue docs a bit to preview how this would look

https://user-images.githubusercontent.com/26852655/182381149-80438c43-4d99-4c04-9031-651ac687d626.mov

.

GretaD:
> I hacked the vue docs a bit to preview how this would look
> 
using actions? so we need to change the action on vue repo, thats what you're saying, right?

PR #6947, opened 2022-08-02
Change the 'show image' styling

GretaD:
Fixes https://github.com/nextcloud/mail/issues/6122

this is how i tried to make it work with the state of actions we have now.
But i agree with Marco's comment here: https://github.com/nextcloud/mail/issues/6122#issuecomment-1202480324

before
Screenshot from 2022-08-02 15-35-35
Screenshot from 2022-08-02 15-35-22

after
hvcfgh

png

~~- [ ] Requires https://github.com/nextcloud/nextcloud-vue/pull/2911~~

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -3,19 +3,26 @@
 		<MdnRequest :message="message" />
 		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content" style="color: #000000">
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
-			<Actions default-icon="icon-toggle" :menu-title="t('mail', 'Show')">
-				<ActionButton icon="icon-toggle"
+			<Actions type="tertiary" :menu-title="t('mail', 'Show images')">
+				<ActionButton
 					@click="displayIframe">
+					<template #icon>
+						<IconImage :size="20" />
+					</template>
 					{{ t('mail', 'Show images temporarily') }}
 				</ActionButton>
 				<ActionButton v-if="sender"
-					icon="icon-toggle"
 					@click="onShowBlockedContent">
+					<template #icon>
+						<IconMail :size="20" />
+					</template>
 					{{ t('mail', 'Always show images from {sender}', { sender }) }}
 				</ActionButton>
 				<ActionButton v-if="domain"
-					icon="icon-toggle"
 					@click="onShowBlockedContentForDomain">
+					<template #icon>
+						<IconDomain :size="20" />
+					</template>
 					{{ t('mail', 'Always show images from {domain}', { domain }) }}
 				</ActionButton>
 			</Actions>
@@ -38,6 +45,9 @@ import PrintScout from 'printscout'
 import { trustSender } from '../service/TrustedSenderService'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
 import Actions from '@nextcloud/vue/dist/Components/Actions'
+import IconImage from 'vue-material-design-icons/ImageSizeSelectActual'
+import IconMail from 'vue-material-design-icons/Email'
+import IconDomain from 'vue-material-design-icons/Domain'
 
 import logger from '../logger'
 import MdnRequest from './MdnRequest'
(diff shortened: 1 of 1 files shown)


GretaD:
@jancborchardt and @nimishavijay please have a look at this solution and if you're ok with it

jancborchardt (changes requested):
- [x] "Show image" → "Show images"
- [x] The eye icon is not necessary, as per mockup
- [x] In the menu, the icons are all the same, so either we should go with no icons or:
    - Temporarily: https://fonts.google.com/icons?selected=Material%20Icons%3Aimage%3A
    - Show from this email: https://fonts.google.com/icons?selected=Material%20Icons%3Amail%3A
    - Show from this domain: https://fonts.google.com/icons?selected=Material%20Icons%3Adomain%3A
    
Regarding the comment from @marcoambrosini, either is fine with me. Indeed using the tertiary button component would be more accurate since it’s not a link which goes somewhere.

GretaD:
> * [ ]  "Show image" → "Show images"
> * [ ]  The eye icon is not necessary, as per mockup
> * [ ]  In the menu, the icons are all the same, so either we should go with no icons or:
>   
>   * Temporarily: https://fonts.google.com/icons?selected=Material%20Icons%3Aimage%3A
>   * Show from this email: https://fonts.google.com/icons?selected=Material%20Icons%3Amail%3A
>   * Show from this domain: https://fonts.google.com/icons?selected=Material%20Icons%3Adomain%3A
> 
> Regarding the comment from @marcoambrosini, either is fine with me. Indeed using the tertiary button component would be more accurate since it’s not a link which goes somewhere.

all the point can be delivered but the icon one. The action must have an icon, sorry should have commented that before. If we wait for the vue changes, this will be delivered a bit late, if we move on with my solution, i will look like this(and your comments, but with an icon)

jancborchardt:
@GretaD cool, and fine for now with the icon too, but then also use the "image" filetype icon instead of "eye".

marcoambrosini:
We can definitely add the option not to have an icon and have the menus triggered by tex-tonly buttons. I think it totally makes sense.

marcoambrosini:
...and that will automatically be the case when https://github.com/nextcloud/nextcloud-vue/issues/2768 happens

GretaD:
> We can definitely add the option not to have an icon and have the menus triggered by tex-tonly buttons. I think it totally makes sense.

i know that, but when it is going to be released? :)

marcoambrosini:
pr is there https://github.com/nextcloud/nextcloud-vue/pull/2911#issuecomment-1207891395
Let's wait for more reviewers since it's a big change. Also you can review @GretaD :) 

Once that's we can do a minor release

ChristophWurst:
How is this unblocked when it depends on https://github.com/nextcloud/nextcloud-vue/pull/2911 and this app still uses v5?

GretaD:
> How is this unblocked when it depends on [nextcloud/nextcloud-vue#2911](https://github.com/nextcloud/nextcloud-vue/pull/2911) and this app still uses v5?

ah sorry didnt see you added the component on desc. I am not using that component, that was not related to actions without icons, what Marco meant(double checked with him) is that after that is in, we can edit the actions on vue again and add a slot with no icons. 

What i did was hiding the icon with display none. 
I dont have time to spend on the component now, so we can either do this or wait till we have time.

src/components/MessageHTMLBody.vue
```diff
@@ -3,19 +3,26 @@
 		<MdnRequest :message="message" />
 		<div v-if="hasBlockedContent" id="mail-message-has-blocked-content" style="color: #000000">
 			{{ t('mail', 'The images have been blocked to protect your privacy.') }}
-			<Actions default-icon="icon-toggle" :menu-title="t('mail', 'Show')">
-				<ActionButton icon="icon-toggle"
+			<Actions type="tertiary" :menu-title="t('mail', 'Show image')">
```
jancborchardt, line 6:
```suggestion
			<Actions type="tertiary" :menu-title="t('mail', 'Show images')">
```

jancborchardt (changes requested):
Except changing it to plural it looks good to me :)

GretaD:
not ready to merge

GretaD:
> not ready to merge

ready now, please reviewers test it

ChristophWurst (approved):
Tested, works

PR #6962, opened 2022-08-08
Store message (HTML) text preview in db cache

ChristophWurst:
This contributes the backend parts of https://github.com/nextcloud/mail/issues/2579.

## How to test
1) Check out the branch
2) Run the migration
3) Open the app and wait for the messages to load

-> either inspect the network traffic, use the Vue tools or look directly into the database to find out *most* messages have a value set for `preview_text`. There is no guarantee that we find one, and sometimes it might be garbage due to the HTML to text conversation. But this is a start.

Tested with plain text email, HTML emails and multipart emails.


kesselb (approved):

kesselb:
Force pushed a fix for the linter

Issue #6628, opened 2022-06-02
Show that mailbox has sub-mailbox with unread messages


ChristophWurst:
### Is your feature request related to a problem? Please describe.

A mailbox with unread messages shows a counter. If I organize my mailboxes in a hierarchy, I do not notice that a child of a top level mailbox has unread messages if the top level mailbox is collapsed.

Example

* Notifications (0)
  * Nextcloud (17)
  * Server (3)

^ The *Notifications* mailbox would *not* indicate any unread messages in its children.

### Describe the solution you'd like

Highlight that a mailbox has unread items.

Evolution does this
![Bildschirmfoto vom 2022-06-02 09-47-22](https://user-images.githubusercontent.com/1374172/171580622-be0e22dc-22af-40e2-88a6-59fb6bada736.png)

Roundcube does this
![Bildschirmfoto vom 2022-06-02 09-49-47](https://user-images.githubusercontent.com/1374172/171581082-f64c8540-37e6-4d0e-aea0-8b1458a5b4c2.png)

PR #6963, opened 2022-08-08
Show mailbox counter if sub mailbox has unread messages

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6628

Bildschirmfoto vom 2022-08-08 23-06-52

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -174,7 +174,10 @@
 				{{ t('mail', 'Delete mailbox') }}
 			</ActionButton>
 		</template>
-		<AppNavigationCounter v-if="showUnreadCounter && mailbox.specialRole !== 'trash'" slot="counter">
+		<AppNavigationCounter v-if="showUnreadCounter && subCounter" slot="counter">
+			{{ mailbox.unread }}&nbsp;({{ subCounter }})
+		</AppNavigationCounter>
+		<AppNavigationCounter v-else-if="showUnreadCounter" slot="counter">
 			{{ mailbox.unread }}
 		</AppNavigationCounter>
 		<template slot="extra">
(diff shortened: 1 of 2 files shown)


st3iny (commented):
Tested and works. However, the design could be improved.

How it looks with a 2-digit notification counter:
image

ChristophWurst:
> Tested and works. However, the design could be improved.
> 
> How it looks with a 2-digit notification counter:

Hmm. Maybe a non-breakable whitespace?

ChristophWurst:
Bildschirmfoto vom 2022-08-09 15-23-07

nbsp looks acceptable

src/components/NavigationMailbox.vue
```diff
@@ -174,7 +174,10 @@
 				{{ t('mail', 'Delete mailbox') }}
 			</ActionButton>
 		</template>
-		<AppNavigationCounter v-if="showUnreadCounter && mailbox.specialRole !== 'trash'" slot="counter">
+		<AppNavigationCounter v-if="showUnreadCounter && subCounter" slot="counter">
```
miaulalala, line 177: wouldn't this only show the unread counter if there was an unread item in the submailbox?

JuliaKirschenheuter (changes requested):
Works! Thank you!

I would remove ```max-width: 44px;```
because of 
image
there are a lot of space

src/components/NavigationMailbox.vue
ChristophWurst: this is for unread messages in a child. the one for the parent is in the else-if

JuliaKirschenheuter (approved):
works!

GretaD (approved):
it works and looks nice

PR #6966, opened 2022-08-09
Show message preview in envelope list

ChristophWurst:
Contributes the envelope list part of https://github.com/nextcloud/mail/issues/2579

Bildschirmfoto vom 2022-08-10 14-37-27
Bildschirmfoto vom 2022-08-10 14-36-17
Bildschirmfoto vom 2022-08-10 14-36-15

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -316,6 +316,7 @@ class Message extends Entity implements JsonSerializable {
 			'inReplyTo' => $this->getInReplyTo(),
 			'references' => empty($this->getReferences()) ? null: json_decode($this->getReferences(), true),
 			'threadRootId' => $this->getThreadRootId(),
+			'previewText' => $this->getPreviewText(),
 		];
 	}
 }
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -7,7 +7,7 @@
 			draggableLabel,
 			selectedEnvelopes,
 		}"
-		class="list-item-style"
+		class="list-item-style envelope"
 		:class="{seen: data.flags.seen, draft, selected: selected}"
 		:to="link"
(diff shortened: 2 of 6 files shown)


GretaD (approved):
tested and it works

jancborchardt (changes requested):
Very nice! :)
- [x] The line-height is different between the 3 lines, there should be equal space between them. So essentially reduce the top padding/margin of the message preview a bit to match the distance between sender and subject. 
- [x] The avatar is not vertically centered with the first 2 lines (sender name and subject). And it probably looks nicer to vertically center it with the 3 lines now (sender name, subject, preview – but _not_ the tags / attachments line). Ref like the screenshot of Gmail mobile you posted: https://github.com/nextcloud/mail/issues/2579#issuecomment-582909752
- [x] The message preview gets ellipsized nicely, but the message subject doesn’t seem to (also in master though), seen by @nimishavijay 

I assume this also then goes into the thread list on the right in a separate PR?

jancborchardt:
Looks good except the avatar is not vertically centered yet (both lines on the left are same length, both lines on right as well):
mail alignment

ChristophWurst:
I think the problem is that around the actual avatar

Bildschirmfoto vom 2022-08-10 16-29-01

we have some bottom spacing

Bildschirmfoto vom 2022-08-10 16-28-58

although there is no margin or padding set. Ideas?

jancborchardt:
> although there is no margin or padding set. Ideas?

Stupid but works:
`.app-content-list-item-icon { height: 40px; }`

Not sure if this issue comes from the component? It’s not there in Contacts or Talk, but not sure what they use and it seems different.

JuliaKirschenheuter (approved):
Works well!
I would say small css polishing can be done a bit later.

Issue #3299, opened 2020-06-24
Have *Important and unread* as first section of priority inbox


ChristophWurst:
### Feature Request

The priority inbox is nice, but I feel like the important section is not ideal as it is because handled items (messages read, replied, fav'ed) remain there and take space we could use for other incoming messages.

I think we should consider doing it similar to Imbox (HEY) or Gmail and have a better workflow to triage *new* messages. Like the priority section should be your primary source for checking your email. You read, fav, mark or delete them. Or leave them unread. But the importance alone should not be the factor to keep the message here. I think anything that has been read should vanish there.

I could unmark the message. But that defeats the purpose of the flag. Just be because the message was processed doesn't mean it loses it's importance. If I want to keep it for something later, I'd star it. If I just took a peek and want it to remain there I can mark as unread.

#### Summary

Change the filter of the important section of the priority inbox to *important* AND *unread*.

---

@nextcloud/mail what do you think?

jancborchardt:
Yep, good call! :) As far as I remember this was also part of the original spec (since we wanted to do it like Gmail), might have gotten lost somewhere along the way.

ChristophWurst:
Alright, let's see if we can finally do this.

The most tricky part is the logic expression to define what goes into what sections. If I'm not entirely wrong, it should be like this

* Important and unread: `imp && ~read || (fav && ~fav)`
* Favs: `fav /\ (~imp \/ (imp /\ read))`
* Other: `~fav && (~imp || (imp && read))`

You can copy the expression into https://web.stanford.edu/class/cs103/tools/truth-table-tool/ to get truth tables.

PR #6967, opened 2022-08-09
Improve PI design

GretaD:
needs #6774


GretaD:
important_newlook

ChristophWurst (approved):
Code looks good so far

GretaD:
i will change the text color soon, its too bright

nicojx:
Apologies if I'm late to a debate on this, but it'd be great to have the "Important and Unread" PI back, or at least an option to switch or a tickbox to filter out unread. There was a discussion a while back here (that I've just come across): https://github.com/nextcloud/mail/issues/3299

PR #6998, opened 2022-08-12
Add ID on appsettingsection

GretaD:


src/components/AccountSettings.vue
```diff
@@ -155,6 +164,19 @@ export default {
 		},
 	},
 	methods: {
+		onAccountUpdated() {
```
GretaD, line 167: this is a regresssion from another old pr, and fixed it here for time reasons

src/components/AccountSettings.vue
```diff
@@ -24,9 +24,11 @@
 
 <template>
 	<AppSettingsDialog
+		id="app-settings-dialong"
```
miaulalala, line 27:
```suggestion
		id="app-settings-dialogue"
```

miaulalala (changes requested):

src/components/AccountSettings.vue
miaulalala: or dialog, whatever is better for you :)

miaulalala (approved):
:shipit:

src/components/AccountSettings.vue
kesselb: Thanks!

PR #7016, opened 2022-08-16
Fix important position

### LINKED ISSUE #6987: Important icon and spam is misalignment on the envelope avatar
### Steps to reproduce

1. Open Mail
2. Mark a message that has subject as important
3. Mark a message as important that doesn't have subject (calendar invites i.e)
4. See that the one that doesnt have a subject is misalignment 

### Expected behavior

is aligned 

### Actual behavior

subject_important

### Mail app version

latest

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

i think is a regression of https://github.com/nextcloud/mail/issues/2579

GretaD:
fixes #6987 this is fixed with the new nc/vue update but on main is broken.

_the computed prop is not right_

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -29,6 +29,7 @@
 			<div
 				v-if="isImportant"
 				class="app-content-list-item-star svg icon-important"
+				:class="{ 'important-position-subject': importantIconPosition, 'important-position-nosubject' : importantIconPositionNoSubject }"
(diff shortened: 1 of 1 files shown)


GretaD:
this bug is not happening once updated to nc/vue beta2-3, keeping it open just in case

GretaD:
fixed with the update beta3

PR #7021, opened 2022-08-17
Preview message content in the thread view

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2579

Bildschirmfoto vom 2022-08-18 09-10-27

## Todo

* [x] Go in the subline below the sender
* [x] Be in color-text-maxcontrast
* [x] Only show for collapsed mails cause for expanded ones it would duplicate otherwise
* [x] Fix width/overflow

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -37,7 +37,6 @@
 				:key="env.databaseId"
 				:envelope="env"
 				:mailbox-id="$route.params.mailboxId"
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@GretaD I can't figure out why the preview element expands beyond the parent div. Could you check it out later and have a look?

Bildschirmfoto vom 2022-08-17 13-52-12

skjnldsv (approved):

PR #7048, opened 2022-08-18
Mark avatar has immutable

CarlSchwan:
This will prevent firefox and safari generating a requests that is hitting the DB on each page load just to check if the etag changed


icewind1991 (approved):

ChristophWurst (approved):
Awesome!

miaulalala:
AvatarControlle::test fails ;)

Issue #7055, opened 2022-08-18
Prefetch first, second to last and siblings for thread messages


miaulalala:

ChristophWurst:
Moved over to a real ticket https://github.com/nextcloud/mail/issues/7267

PR #7059, opened 2022-08-19
[POC] for threading messages prefetch

miaulalala:
Still has some issues with fetching the first message - but POC so :shrug: 

The idea is to prefetch a predetermined set of thread messages in the background to speed up threading.

Prefetch logic:

If the message is the last in the thread, get the second to last message and the first message of a thread. Reason: the user might want to look at the previous message to see what someone else said before them, or go back to the beginning to see what the thread is about.

If the message is in the middle of a thread, fetch the previous and following message. Reason: the user might search for a message that is somewhere in the middle and could concievably click one message further up or down to find what they are looking for.

The pre- fetched messages won't always be 100% what is needed, but it might still speed up rendering considerably.

Fixes https://github.com/nextcloud/mail/issues/7055

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -334,6 +334,20 @@ export default {
 				await this.fetchItineraries()
 			}
 		},
+		async fetchPartialThread() {
+			this.loading = true
+
+			logger.debug(`fetching partial thread messages ${this.envelope.databaseId}`)
+
+			try {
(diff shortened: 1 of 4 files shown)


ChristophWurst (commented):
This doesn't work, does it?

miaulalala:
> This doesn't work, does it?

parts of it do, but as this was never meant to be production ready code we can close this PR. This was a POC that came from a discussion during the performance hackweek.

PR #6995, opened 2022-08-12
Migrate icons to material design icons

GretaD:
Ref https://github.com/nextcloud/groupware/issues/38
To do
- [x] All the Loading icons
- [x] some leftovers (icons on attachments)
- [x] change button to buttonvue - for the icon parts
- [x] Requires https://github.com/nextcloud/mail/pull/7060
- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/3092 because https://github.com/nextcloud/nextcloud-vue/pull/3066

there are some bugs because of nc/vue update but not connected to this pr, those bugs can be found here: https://github.com/nextcloud/mail/pull/7071

test carefully composer(send email, drafts and stuff) and loading envelopes

PR #7060, opened 2022-08-19
Upgrade to nc/vue6beta3

GretaD:
Messages open in a new tab on click, this is fixed on vue but not released yet.
I have an open pr to fix loading icons and some other small design issues, if you notice smth here, i will fix it in there - https://github.com/nextcloud/mail/pull/6995

This pr is needed so i can move on with this one: https://github.com/nextcloud/mail/pull/6995


ChristophWurst:
What is done and what is left to do?

ChristophWurst (approved):

PR #7078, opened 2022-08-24
Fix active state on envelope

GretaD:
Fixes https://github.com/nextcloud/mail/issues/7098

for some reasons, the active state on envelope is not handled by vue router, i dont know why.
Adding the :active on listitem on evelope makes it work again(aka using the style from nc/vue component without forcing style on mail side)

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -12,6 +12,7 @@
 		:to="link"
 		:data-envelope-id="data.databaseId"
 		:title="addresses"
+		:active="isActive"
 		:details="formatted()"
 		@click="onClick">
 		<template #icon>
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> for some reasons, the active state on envelope is not handled by vue router, i dont know why.

@JuliaKirschenheuter @st3iny can one of you have a look? I strongly believe that the active state should be handled by the Vue router already. Maybe our styling just uses the wrong active state. If anything, this should be fixed in nc/vue.

ChristophWurst:
https://github.com/nextcloud/nextcloud-vue/pull/3147 and https://github.com/nextcloud/mail/pull/7128 fix the problem at its root.

PR #7080, opened 2022-08-25
Remove unneccessary account check

miaulalala:
Stumbled upon this while working on the Drafts, as discussed with @juliushaertl before.

As we're already joining on the user accounts table in `findById`, the account doesn't need to be queried again for the Controller's `show`.

This is a mini performance optimistation.

The complete account data could also be returned with the JOIN queries as a second
performance improvement, instead of querying the entity fromt he DB again. The update method does just that at the moment, which could be optimised for further performance improvements.


ChristophWurst (approved):
Awesome

PR #7081, opened 2022-08-25
Migrate composer attachment classes to BEM

ChristophWurst:
This prevents possible overlaps with a generic `.error` selector.


GretaD (approved):
there's still one class called list-item on  <style>

PR #6660, opened 2022-06-06
Improve the way attachment look

sazanof:
Hello! **I will be glad if my PR will be useful in some way**. Based on https://github.com/nextcloud/mail/issues/5768
Attachments_new2
Of the possible nuances of implementation:
- It is not possible to generate previews of downloaded files (except images)
-  If at the time of downloading some large file you try to download the same file in parallel - the previous download stops. However, if you move `Vue.set(this, 'uploads', {})` in `complete()`, then due to the fact that it is impossible to correctly identify the necessary object from two identical ones, the progress bar is buggy
- There is no possibility (yet) to show the beginning of the selection of cloud files ...

There is a parallel PR https://github.com/nextcloud/mail/pull/6310 which is related to this. It may be worth using the same component to display the attachment.

**If you have the opportunity, please take a look and give your comments. Thanks!**

BTW the previous version was a bit in a different style =) :
image

_And yes, maybe I overdid it a bit in the code_ (he shrugs his shoulders)
Signed-off-by: Mikhail Sazanov <m@sazanof.ru>

PR #7066, opened 2022-08-22
Open image, video, audio and PDF attachments with viewer

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6983

- [x] Requires https://github.com/nextcloud/viewer/pull/1338
- [x] Requires https://github.com/nextcloud/files_pdfviewer/pull/646
- [x] Requires https://github.com/nextcloud/viewer/pull/1350
- [x] Keep allow list of mimes we know work, because there are a handful of Viewer handlers that don't work at the moment
- [x] Requires https://github.com/nextcloud/viewer/pull/1348 to close the Viewer again
- [x] Requires https://github.com/nextcloud/viewer/pull/1349 for a working *Download* action inside the Viewer
- [x] Requires https://github.com/nextcloud/files_pdfviewer/pull/648 for the PDF viewer to work
- [x] Version check to make this only available with 25+ (older Viewer would error out)

## Tested media

* PDFs
* JPEG
* GIF
* OGG
* MP4

## How to test

1) Create an account
2) Send an email to yourself with various attachments
3) Open the email
4) Click the attachment - it will open if its supported

PR #7082, opened 2022-08-25
Make it clear that not all attachments will be sent

ChristophWurst:
If you upload a handful of attachments, the list collapses due to https://github.com/nextcloud/mail/pull/6660 and upload errors are not noticeable visibly anymore. This changes the color of the collapse handle to reflect errors.

Bildschirmfoto vom 2022-08-25 14-25-58

Bildschirmfoto vom 2022-08-25 14-26-12

I believe the flow and error handling has still a lot of potential for improvement. E.g. to disable the send button until all errors are handled (user needs to remove the faulty attachment or we never add it) but that requires larger restructuring.

### DIFF
--- a/src/components/ComposerAttachments.vue
+++ b/src/components/ComposerAttachments.vue
@@ -22,14 +22,21 @@
 
 <template>
 	<div class="new-message-attachments">
-		<div v-if="hasNextLine" class="message-attachments-count" @click="isToggle = !isToggle">
+		<div v-if="hasNextLine"
+			class="new-message-attachments--counter"
+			:class="{ 'new-message-attachments--counter--with-errors': hasAttachmentErrors }"
+			@click="isToggle = !isToggle">
 			<span>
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@GretaD @JuliaKirschenheuter :ping_pong: for review

src/components/ComposerAttachments.vue
```diff
@@ -422,7 +432,19 @@ export default {
...
+		padding: 10px 20px;
+		cursor:pointer;
+		display:flex;
+		align-items: center;
+
+		&--with-errors {
+			color:red;
+		}
+	}
+
```
JuliaKirschenheuter, line 446: please place ```cursor:pointer;``` to all child elements
`	&--counter * {
		cursor:pointer;
	}
`

JuliaKirschenheuter (commented):
works well

src/components/ComposerAttachments.vue
ChristophWurst: Done in https://github.com/nextcloud/mail/pull/7066/files#diff-13360562f049327b0c3be2784d54548428fbe349d2a0a43a04a5278104998722R247

PR #7084, opened 2022-08-25
Respect share enumeration constraints on autocomplete

### LINKED ISSUE #6017: This app does not respect the "Allow sharing with groups"-setting
### Expected behavior

Given the following settings

Screenshot from 2022-02-08 14-23-03

When you write a new email, you shouldn't be proposed autocompletion emails from users from groups you are not in.

### Actual behavior

It does propose you all the emails of the instance.

### Steps to reproduce

1. configure the sharing settings like in the screenshot
2. create a new email
3. start to type in the to field and get autocompletion proposition from users outside your groups.

### Mail app version

1.11.6

### Mailserver or service

Not relevant.

### Operating system

Not relevant.

### PHP engine version

Other

### Web server

Other

### Database

Other

### Additional info

Relates to this bug:
https://github.com/nextcloud/server/issues/25390

On shared instance, this is a privacy issue.

### LINKED ISSUE #6148: Mail App ignores autocomplete settings
### Steps to reproduce

1. Set "Allow sharing with groups" + "Restrict users to only share with users in their groups"
2. Set "Allow username autocompletion in share dialog" + "Allow username autocompletion to users within the same group"
3. Mail app autocompletes users from other groups and other groups
image
image

### Expected behavior

Isolate groups from others for privacy reasons. Still allow sharing and autocompletion within group

### Actual behavior

Mail app autocompletes other groups on the same instance and users from other groups, despite the admin sharing settings

### Mail app version

1.11.7

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

PHP 8.0

### Web server

Apache (supported)

### Database

MySQL

### Additional info

_No response_

hrenard:
Fixes #6017
Fixes #6148

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -26,39 +26,79 @@ namespace OCA\Mail\Service;
 
 use OCP\Contacts\IManager;
 use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\IUserManager;
 
 class ContactsIntegration {
 	/** @var IManager */
 	private $contactsManager;
 
+	/** @var IGroupManager */
+	private $groupManager;
+
+	/** @var IUserManager */
+	private $userManager;
+
 	/** @var IConfig */
 	private $config;
 
-	public function __construct(IManager $contactsManager, IConfig $config) {
+	public function __construct(IManager $contactsManager,
+								IGroupManager $groupManager,
+								IUserManager $userManager,
+								IConfig $config) {
 		$this->contactsManager = $contactsManager;
+		$this->groupManager = $groupManager;
+		$this->userManager = $userManager;
 		$this->config = $config;
 	}
 
 	/**
 	 * Extracts all matching contacts with email address and name
 	 *
+	 * @param string $userId
 	 * @param string $term
 	 * @return array
 	 */
-	public function getMatchingRecipient(string $term): array {
+	public function getMatchingRecipient(string $userId, string $term): array {
 		if (!$this->contactsManager->isEnabled()) {
 			return [];
 		}
 
 		// If 'Allow username autocompletion in share dialog' is disabled in the admin sharing settings, then we must not
 		// auto-complete system users
-		$allowSystemUsers = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
+		$shareeEnumeration = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
+		$shareeEnumerationInGroupOnly = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_group', 'no') === 'yes';
+		$shareeEnumerationFullMatch = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match', 'yes') === 'yes';
+		$shareeEnumerationFullMatchUserId = $shareeEnumerationFullMatch && $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_userid', 'yes') === 'yes';
+		$shareeEnumerationFullMatchEmail = $shareeEnumerationFullMatch && $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_email', 'yes') === 'yes';
 
-		$result = $this->contactsManager->search($term, ['FN', 'EMAIL']);
+		$result = $this->contactsManager->search($term, ['UID', 'FN', 'EMAIL'], ['enumeration' => $shareeEnumeration]);
+		if (empty($result)) {
+			return [];
+		}
 		$receivers = [];
+
+		if ($shareeEnumeration && $shareeEnumerationInGroupOnly) {
+			$user = $this->userManager->get($userId);
+			if ($user === null) {
+				return [];
+			}
+			$userGroups = $this->groupManager->getUserGroupIds($user);
+		}
+
 		foreach ($result as $r) {
-			if (!$allowSystemUsers && isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
-				continue;
+			$isSystemUser = isset($r['isLocalSystemBook']) && $r['isLocalSystemBook'];
+			$isInSameGroup = false;
+			if ($isSystemUser && $shareeEnumerationInGroupOnly) {
+				foreach ($userGroups as $userGroup) {
+					if ($this->groupManager->isInGroup($r['UID'], $userGroup)) {
+						$isInSameGroup = true;
+						break;
+					}
+				}
+				if (!$shareeEnumerationFullMatch && !$isInSameGroup) {
+					continue;
+				}
 			}
 
 			$id = $r['UID'];
@@ -77,18 +117,15 @@ class ContactsIntegration {
 				if ($e === '') {
 					continue;
 				}
-				// Show full name if possible or fall back to email
-				if ($fn !== null) {
+				$lowerTerm = strtolower($term);
+				if (!$isSystemUser || $isInSameGroup || ($lowerTerm !== '' && (
+					($shareeEnumerationFullMatch && !empty($fn) && $lowerTerm === strtolower($fn)) ||
+					($shareeEnumerationFullMatchUserId && $lowerTerm === strtolower($id)) ||
+					($shareeEnumerationFullMatchEmail && $lowerTerm === strtolower($e))))) {
 					$receivers[] = [
 						'id' => $id,
-						'label' => "$fn ($e)",
-						'email' => $e,
-						'photo' => $photo,
-					];
-				} else {
-					$receivers[] = [
-						'id' => $id,
-						'label' => $e,
+						// Show full name if possible or fall back to email
+						'label' => (empty($fn) ? $e : "$fn ($e)"),
 						'email' => $e,
 						'photo' => $photo,
 					];
--- a/lib/Service/AutoCompletion/AutoCompleteService.php
+++ b/lib/Service/AutoCompletion/AutoCompleteService.php
@@ -45,7 +45,7 @@ class AutoCompleteService {
 	}
 
 	public function findMatches(string $userId, string $term): array {
-		$recipientsFromContacts = $this->contactsIntegration->getMatchingRecipient($term);
+		$recipientsFromContacts = $this->contactsIntegration->getMatchingRecipient($userId, $term);
 		$recipientGroups = $this->groupsIntegration->getMatchingGroups($term);
 		$fromCollector = $this->addressCollector->searchAddress($userId, $term);
 
--- a/tests/Unit/Service/Autocompletion/AutoCompleteServiceTest.php
+++ b/tests/Unit/Service/Autocompletion/AutoCompleteServiceTest.php
@@ -69,7 +69,7 @@ class AutoCompleteServiceTest extends TestCase {
 
 		$this->contactsIntegration->expects($this->once())
 			->method('getMatchingRecipient')
-			->with($term)
+			->with('testuser', $term)
 			->willReturn($contactsResult);
 		$this->groupsIntegration->expects($this->once())
 			->method('getMatchingGroups')
--- a/tests/Unit/Service/ContactsIntegrationTest.php
+++ b/tests/Unit/Service/ContactsIntegrationTest.php
@@ -25,10 +25,19 @@ use ChristophWurst\Nextcloud\Testing\TestCase;
 use OCA\Mail\Service\ContactsIntegration;
 use OCP\Contacts\IManager;
 use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\IUserManager;
+use OCP\IUser;
 
 class ContactsIntegrationTest extends TestCase {
 	/** @var IManager */
 	private $contactsManager;
+
+	/** @var IGroupManager */
+	private $groupManager;
+
+	/** @var IUserManager */
+	private $userManager;
 
 	/** @var IConfig */
 	private $config;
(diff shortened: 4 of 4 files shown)


lib/Service/ContactsIntegration.php
```diff
@@ -54,12 +67,28 @@ public function getMatchingRecipient(string $term): array {
...
 		$receivers = [];
 		foreach ($result as $r) {
-			if (!$allowSystemUsers && isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
-				continue;
+			if (isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
+				if (!$allowSystemUsers) {
+					continue;
+				}
+				if ($allowSystemUsersInGroupOnly) {
+					$userGroups = $this->groupManager->getUserGroupIds($this->userSession->getUser());
```
ChristophWurst, line 80: Nitpick: the user should be an input argument of the method.

Background: that makes the method independent of the context. E.g. some processes run in the background. They will then still work and not produce unpredictable results.

lib/Service/ContactsIntegration.php
```diff
@@ -26,40 +26,64 @@
...
 		$receivers = [];
 		foreach ($result as $r) {
-			if (!$allowSystemUsers && isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
-				continue;
+			if (isset($r['isLocalSystemBook']) && $r['isLocalSystemBook']) {
+				if (!$allowSystemUsers) {
+					continue;
+				}
+				if ($allowSystemUsersInGroupOnly) {
+					$userGroups = $this->groupManager->getUserIdGroupIds($userId);
```
kesselb, line 75: I recommend to move this code before the result foreach. 

```
$allowSystemUsersInGroupOnly = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_group', 'no') === 'yes';
if ($allowSystemUsersInGroupOnly) {
    $userGroups = $this->groupManager->getUserIdGroupIds($userId);
}
```

hrenard:
I want to add a test, but I'm no php developer.
```php
$this->groupManager->expects($this->once())
	->method('isInGroup')
	->will($this->returnValue(false));
$this->groupManager->expects($this->once())
	->method('isInGroup')
	->with('auser', 'agroup')
	->will($this->returnValue(true));
```
Does this work  ?

lib/Service/ContactsIntegration.php
```diff
@@ -26,40 +26,67 @@
...
 		if (!$this->contactsManager->isEnabled()) {
 			return [];
 		}
 
 		// If 'Allow username autocompletion in share dialog' is disabled in the admin sharing settings, then we must not
 		// auto-complete system users
 		$allowSystemUsers = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
+		$allowSystemUsersInGroupOnly = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_group', 'no') === 'yes';
+		if ($allowSystemUsersInGroupOnly) {
+			$userGroups = $this->groupManager->getUserIdGroupIds($userId);
```
miaulalala, line 67: This method is private.

Inject the `IUserManager` into the constructor and get the user object:

```suggestion
		$user = $this->userManager->get($userId);
		if($user === null) {
			return [];
		}
			$userGroups = $this->groupManager->getUserGroupIds($user);
```

miaulalala (changes requested):

hrenard:
Thanks for the feedback ! I added `testGetMatchingRecipientOnlyGroupUsers` for this use case, but I've not tried it.

miaulalala:
> Thanks for the feedback ! I added `testGetMatchingRecipientOnlyGroupUsers` for this use case, but I've not tried it.

You can run a singular unit test with `composer run test:unit:dev ./tests/Unit/Service/ContactsIntegrationTest.php `

hrenard:
Okay. The tests are working.

hrenard:
And the coding standards... Sorry for that.

miaulalala:
Postgres tests are failing because of an incompatibility. Can you rebase your brach onto the latest mail main? The fix for this has been merged :)

hrenard:
> Postgres tests are failing because of an incompatibility. Can you rebase your brach onto the latest mail main? The fix for this has been merged :)

Done.

miaulalala (changes requested):
Hi @hrenard , thanks for your hard work so far!

After some feedback I found out there are more conditions to be checked for:

https://github.com/nextcloud/server/blob/88b0aea2b0f7cc30ab0de7d13914474d0711e5cd/lib/private/Collaboration/Collaborators/MailPlugin.php#L88-L93

Would you be up for implementing these too?

If that's too overwhelming, I can take over and build upon your excellent work.

Thanks again!

hrenard:
> Hi @hrenard , thanks for your hard work so far!
> 
> After some feedback I found out there are more conditions to be checked for:
> 
> https://github.com/nextcloud/server/blob/88b0aea2b0f7cc30ab0de7d13914474d0711e5cd/lib/private/Collaboration/Collaborators/MailPlugin.php#L88-L93
> 
> Would you be up for implementing these too?
> 
> If that's too overwhelming, I can take over and build upon your excellent work.
> 
> Thanks again!

Yep, will do.

Only `shareapi_restrict_user_enumeration_full_match` & `shareapi_restrict_user_enumeration_full_match_email` are missing, right ?

miaulalala:
> > Hi @hrenard , thanks for your hard work so far!
> > After some feedback I found out there are more conditions to be checked for:
> > https://github.com/nextcloud/server/blob/88b0aea2b0f7cc30ab0de7d13914474d0711e5cd/lib/private/Collaboration/Collaborators/MailPlugin.php#L88-L93
> > Would you be up for implementing these too?
> > If that's too overwhelming, I can take over and build upon your excellent work.
> > Thanks again!
> 
> Yep, will do.
> 
> Only `shareapi_restrict_user_enumeration_full_match` & `shareapi_restrict_user_enumeration_full_match_email` are missing, right ?

And the `shareapi_restrict_user_enumeration_to_phone` I think. Thanks so much!

hrenard:
As we only search by `FN` or `EMAIL`, can we omit the phone ?

`shareapi_restrict_user_enumeration_full_match` is supposed to match the username. Do we consider the search result's `UID` to be the username or do we just skip system users if `shareapi_restrict_user_enumeration_full_match_email` is not set ?

Maybe we should also hide group recipients according to `shareapi_allow_group_sharing` and when other restrictions are set ?
If I'm to be in the same group or to know the full email of someone to send an email, it's weird that I can send emails to random groups thus random peoples. And if someone replies, I get the list of this group members.

Or should we introduce config specific to this app ?

miaulalala:
> `shareapi_restrict_user_enumeration_full_match` is supposed to match the username. Do we consider the search result's `UID` to be the username or do we just skip system users if `shareapi_restrict_user_enumeration_full_match_email` is not set ?

You could pass the `fullmatch` option to the search if the `shareapi_restrict_user_enumeration_full_match_email` and let the `IManager` handle it if that does what I think you want to get done? Have a look at this method:

https://github.com/nextcloud/server/blob/master/lib/private/ContactsManager.php#L52

So you could do something like:

```php
$shareapi_restrict_user_enumeration_full_match_email = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_email', 'no') === 'yes';

$result = ($shareapi_restrict_user_enumeration_full_match_email 
                             ? $this->contactsManager->search($term, ['FN', 'EMAIL'], ['fullmatch' => true])
                             : $this->contactsManager->search($term, ['FN', 'EMAIL']));
```

I think the phone match is fine to leave out, who searches for a users phone no when they want to send an email anyway :sweat_smile:

hrenard:
Interesting, I'll look into it.

What behavior is intended when both `shareapi_restrict_user_enumeration_to_group` and `shareapi_restrict_user_enumeration_full_match_email` are true ?
1. Only match system user with full email.
2. Only match system user in the same group.
3. Match system user with full email or in the same group.

I found the 3 more logical, but I'm not sure of the initial design.

hrenard:
I think `shareapi_restrict_user_enumeration_full_match_email` is not used anymore.

@miaulalala I pushed a proposal for option 3 (previous comment). Tell me what you think. I'll update tests when we agree on the behavior.

miaulalala:
> I think `shareapi_restrict_user_enumeration_full_match_email` is not used anymore.

As far as I can see it is still used in the \OC\Collaboration\Collaborators\UserPlugin::search

If I'm not mistaken, it is done in the way you suggested in 3:

```php
$lowerSearch = strtolower($search);
if (
    $this->shareeEnumerationFullMatch && $lowerSearch !== '' && (strtolower($uid) === $lowerSearch ||
    strtolower($userDisplayName) === $lowerSearch ||
    ($this->shareeEnumerationFullMatchIgnoreSecondDisplayName && trim(strtolower(preg_replace('/ \(.*\)$/', '', $userDisplayName))) === $lowerSearch) ||
    ($this->shareeEnumerationFullMatchEmail && strtolower($userEmail ?? '') === $lowerSearch))
) {}

```

If I have my ANDs and ORs right, we can proceed with your suggestion :)

hrenard:
I can't find the `shareapi_restrict_user_enumeration_full_match_email` setting in the UI of the latest NC which led me to assume it's deprecated. But, I haven't searched that much, so I'm probably wrong.

(I'll try to finish this next week.)

lib/Service/ContactsIntegration.php
```diff
@@ -26,39 +26,74 @@
...
 
 		// If 'Allow username autocompletion in share dialog' is disabled in the admin sharing settings, then we must not
 		// auto-complete system users
-		$allowSystemUsers = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
+		$shareeEnumeration = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'no') === 'yes';
+		$shareeEnumerationInGroupOnly = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_group', 'no') === 'yes';
+		$shareeEnumerationFullMatch = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match', 'yes') === 'yes';
+		$shareeEnumerationFullMatchUserId = $shareeEnumerationFullMatch && $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_userid', 'yes') === 'yes';
+		$shareeEnumerationFullMatchEmail = $shareeEnumerationFullMatch && $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_email', 'yes') === 'yes';
+		if ($shareeEnumeration && $shareeEnumerationInGroupOnly) {
```
miaulalala, line 74: Move the contactsManager query before the `if` check
```suggestion
		$result = $this->contactsManager->search($term, ['UID', 'FN', 'EMAIL'], ['enumeration' => $shareeEnumeration]);
		if(empty($result)) {
			return [];
		}
		$receivers = [];

		if ($shareeEnumerationInGroupOnly) {
```

lib/Service/ContactsIntegration.php
```diff
@@ -77,6 +112,16 @@ public function getMatchingRecipient(string $term): array {
...
+				$lowerTerm = strtolower($term);
+				if ($isSystemUser &&
+					($shareeEnumerationInGroupOnly || $shareeEnumerationFullMatch) &&
+					!($shareeEnumerationInGroupOnly && $isInSameGroup) &&
+					!($shareeEnumerationFullMatch && $lowerTerm !== '' && $lowerTerm === strtolower($fn)) &&
+					!($shareeEnumerationFullMatchUserId && $lowerTerm === strtolower($id)) &&
+					!($shareeEnumerationFullMatchEmail && $lowerTerm === strtolower($e))
+				) {
+					continue;
+				}
```
miaulalala, line 124: Allow same group and all address books other than the system one as the share api matches should not influence one's own contacts and the limitations of the share API should not apply to them.

```suggestion
if(!$isSystemUser || $isInSameGroup) {
					$receivers[] = [
						'id' => $id,
						'label' => (empty($fn) ? $e : "$fn ($e)"),
						'email' => $e,
						'photo' => $photo,
					];
					continue;
				}

				if($shareeEnumerationFullMatch && $lowerTerm !== '' &&
					((!empty($fn) && strtolower($fn) === $lowerTerm) ||
					($shareeEnumerationFullMatchUserId && strtolower($id) === $lowerTerm )||
					($shareeEnumerationFullMatchEmail && strtolower($e) === $lowerTerm))
				){
					$receivers[] = [
						'id' => $id,
						'label' => (empty($fn) ? $e : "$fn ($e)"),
						'email' => $e,
						'photo' => $photo,
					];
				}
```

Then the function is done.

I have restructured the if condition to be permissive instead of dismissve to be a bit more clear.

lib/Service/ContactsIntegration.php
```diff
@@ -26,39 +26,74 @@
...
+			$user = $this->userManager->get($userId);
+			if ($user === null) {
+				return [];
+			}
+			$userGroups = $this->groupManager->getUserGroupIds($user);
+		}
 
-		$result = $this->contactsManager->search($term, ['FN', 'EMAIL']);
+		$result = $this->contactsManager->search($term, ['UID', 'FN', 'EMAIL'], ['enumeration' => $shareeEnumeration]);
 		$receivers = [];
```
miaulalala, line 83:
```suggestion
```

miaulalala (changes requested):
Thanks for your work!

I have a few changes, but in principle it's :man_cook: :ok_hand:

lib/Service/ContactsIntegration.php
miaulalala, line 74:
```suggestion
Move the contactsManager query before the `if` check
```suggestion
		$result = $this->contactsManager->search($term, ['UID', 'FN', 'EMAIL'], ['enumeration' => $shareeEnumeration]);
		if(empty($result)) {
			return [];
		}
		$receivers = [];

		if ($shareeEnumerationInGroupOnly) {
```
```

miaulalala (changes requested):

lib/Service/ContactsIntegration.php
hrenard: You right the permissive approach is better ! I propose a compromise though :wink:.

miaulalala: hehe I see you like your `if`s short and sweet. Well done!

miaulalala (approved):
Great work! Thanks a lot!

Issue #6142, opened 2022-03-21
Improve look of priority inbox


nimishavijay:
### Is your feature request related to a problem? Please describe.

The priority inbox is the first thing we see when we open the Mail app, some design improvements could go a long way to make the Mail app more polished.

### Describe the solution you'd like

@jancborchardt and I worked on making some design changes that (hopefully) make the experience better.

| New priority inbox | Skeleton screen | No important mails |
| --- | --- | --- |
| ![image](https://user-images.githubusercontent.com/52440189/159458613-98761ab7-0b63-4e8e-ac6f-0c101a4399e5.png) | ![image](https://user-images.githubusercontent.com/52440189/159458957-889ddf51-23fc-4c5f-8727-1224d72526f9.png) | ![image](https://user-images.githubusercontent.com/52440189/159459103-719d3139-1750-43d9-9ca6-5f9d4a6850b6.png) |

Major changes:

- [x] Remove the "Favorite" section as it can be accessed from the navigation
- [x] Move the "New message" and "Refresh" button from top of navigation to top of the mail list #6774 
- [x] Instead of "Important and unread" show only all important mails (read and unread) as you can't access the important mails from anywhere -- what do you think? Many email clients allow different configurations of the inbox so it's not clear which is the most common one. 
- [x] Show more emails in the Important section --> if possible we could calculate the maximum number of mails that can be shown from the height of the viewport? 
- [x] Show skeleton screen while loading instead of the loading spinner
- [x] Hide *Important* section when it's empty
- [x] Show skeleton for the thread details. There shall be placeholders for the title, sender etc of a typical thread and *one message*

Minor changes:

- [x] - Change refresh icon to MDI `refresh`
- [x] - Left align the message shown when a section is empty
- [x] - Also change the text so that it is more helpful when there are no important mails 
- [x] "No messages" --> "Messages marked as important will show up here"
- [x] - Add icon and change alignment of "Load more"

What do you think?
cc @jancborchardt and @karlitschek 

### Additional context

(edit: decided with Jan that the collapsing of the important section is not necessary after all, and changed the screenshots to reflect that)

thstyl2000:
You're doing a great work, but I've got to be honest: I haven't yet found a way to include 'Important' messages to my workflow....
I am personally using a tagging/folder system for my work account and a favorites/folder system for my personal email.

From this point of view, how about having no 'Imprtant' list, but only favorites, but the important emails get sorted on top?

karlitschek:
[...]

PR #7116, opened 2022-08-31
Hide important section when no important messages

GretaD:
Fixes parts of #6142

### DIFF
--- a/src/components/EmptyMailboxSection.vue
+++ b/src/components/EmptyMailboxSection.vue
@@ -1,6 +1,6 @@
 <template>
 	<div id="emptycontent">
-		<slot> {{ t('mail', 'Messages marked as important will show up here') }} </slot>
+		<slot> {{ t('mail', 'No messages') }} </slot>
 	</div>
 </template>
 
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -21,7 +21,7 @@
 					:bus="bus"
 					:open-first="mailbox.specialRole !== 'drafts'" />
 				<template v-else>
-					<div class="app-content-list-item">
+					<div v-show="hasImportantEnvelopes" class="app-content-list-item">
 						<SectionTitle class="important" :name="t('mail', 'Important')" />
 						<Popover trigger="hover focus">
 							<ButtonVue slot="trigger"
@@ -37,7 +37,7 @@
 							</p>
 						</Popover>
 					</div>
-					<Mailbox
+					<Mailbox v-show="hasImportantEnvelopes"
 						class="nameimportant"
 						:account="unifiedAccount"
 						:mailbox="unifiedInbox"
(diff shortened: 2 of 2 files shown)


src/components/MailboxThread.vue
```diff
@@ -22,9 +22,9 @@
 					:bus="bus" />
 				<template v-else>
 					<div class="app-content-list-item">
-						<SectionTitle v-if="hasImportantEnvelopes" class="important" :name="t('mail', 'Important')" />
+						<SectionTitle v-if="!hasImportantEnvelopes" class="important" :name="t('mail', 'Important')" />
```
GretaD, line 25: hmm, this is still shown :weary:

arjentz:
Just wondering, is this one actually not a draft?

GretaD:
> Just wondering, is this one actually not a draft?

hey arjen, thanks a lot for having a look. This is suppose to work, but the labels are not hidden when theres no messages on important section, and i dont know what im doing wrong. Thats why i kept it in draft.

src/components/MailboxThread.vue
```diff
@@ -280,6 +286,9 @@ export default {
 .app-content-list-item:hover {
 	background: transparent;
 }
+app-content-list-item {
```
ChristophWurst, line 289: missing a dot?

ChristophWurst:
@GretaD please test again

src/components/EmptyMailboxSection.vue
```diff
@@ -1,7 +1,5 @@
 <template>
-	<div id="emptycontent">
-		<slot> {{ t('mail', 'Messages marked as important will show up here') }} </slot>
```
ChristophWurst, line 3: I believe this file might be obsolete now

ChristophWurst: I don't think that is the case. `EmptyMailboxSection` renders when `is-priority-inbox=true`, which is only done for the two priority inbox sections in `MailboxThread`. And we no longer show the *important* mailbox when it's empty, only *Other*.

So whenever *Other* is empty, we show a display-none'd `EmptyMailboxSection`.

This means we have to undo the display none and rectify `Messages marked as important will show up here` to make sense for *Others*. It was a wrong assumption that the component would only ever show for important emails.

GretaD: when i switch between mailboxes, this is shown
Screenshot from 2022-10-17 16-18-39

GretaD: exactly, that was my problem since the beginning i cannot hide the label

ChristophWurst: Test again. I accidentally undid one of my changes when I committed. Fixed now.

GretaD: yes, all looks good now

ChristophWurst (approved):

PR #7123, opened 2022-09-01
Migrate to native datetime picker

ChristophWurst:
- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/3063

### Convert email into event

Bildschirmfoto vom 2022-09-01 13-08-13

### Send later

Bildschirmfoto vom 2022-09-01 13-05-37


ChristophWurst:
Updated the branch. Both picker still work :+1:

src/components/EventModal.vue
```diff
@@ -6,15 +6,15 @@
 				<input v-model="eventTitle" type="text">
 			</div>
 			<div class="dateTimePicker">
-				<DatetimePicker v-model="startDate"
+				<DateTimePickerNative v-model="startDate"
 					:format="dateFormat"
 					:clearable="false"
 					:minute-step="5"
 					:show-second="false"
```
JuliaKirschenheuter, line 13: this attributes should be removed / replaced. Native datetime picker does't have some of them / or have other analogs

JuliaKirschenheuter:
@ChristophWurst there are some other things to do, i can take over on Monday and finish it

ChristophWurst:
> @ChristophWurst there are some other things to do, i can take over on Monday and finish it

Let's do from scratch when we have time.

Issue #6973, opened 2022-08-11
Widen default width of AppContentList 


ChristophWurst:
### Is your feature request related to a problem? Please describe.

Second column is too narrow by default

### Describe the solution you'd like

* [x] ~250px to 300px (same as AppNavigation) https://github.com/nextcloud/nextcloud-vue/pull/3142
* [x] 4px whitespace left/right on AppContentList items toward edges, like in new AppNavigation and in Talk nav https://github.com/nextcloud/mail/pull/7122

ChristophWurst:
https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/AppContent?id=appcontent-1 has a percentual init value. We don't set anything in Mail at the moment, so it's 20%. How far should we go up @jancborchardt?

ChristophWurst:
If we assume a screen of 1080px there are 300px for the nav and 780 for the content. 300px/780px is 38%. Would it be ok to use seomthing like that as new default?
The current max width is 40%. If we go near that for the default width we should probably increase that.

jancborchardt:
20% seems fine, but we should probably have a min-width of 300px (same as AppNavigation) as well.

ChristophWurst:
Min/max are also defined as percentage. The component default is 15 as min, 40 as max.

ChristophWurst:
> min-width of 300px

The shallowest window width with three columns is 1024px. With the sidebar collapsed we are at 30% for the message list to be at ~300px. With the open sidebar it's ~40% for 280px.

How about we change the min to 25%? The only downside could be very large screens. Like on a 4k screen with 2160px width the column would be at least 540px wide.

ChristophWurst:
Alternatively we could look into changing the `@nextcloud/vue` component to also allow pixel values for the props.

ChristophWurst:
@skjnldsv what do you think? Would it make sense to offer pixels as alternative? The percentile min/max is tricky to get right for various browser widths. Any other ideas?

PR #7127, opened 2022-09-01
Bump @nextcloud/vue to v6.0.0-beta6

ChristophWurst:
Tested and works

Fixes https://github.com/nextcloud/mail/issues/6973


GretaD (approved):
works

PR #7137, opened 2022-09-02
Release v1.14.0-beta2

GretaD:
ref https://github.com/nextcloud/groupware/issues/55


ChristophWurst (approved):
Eventually we have to combine the v1.14.0-* changelogs into one. We don't do changelogs for prereleases, only for releases

GretaD:
done here: https://github.com/nextcloud/mail/pull/7139

becase we branched off

PR #7150, opened 2022-09-05
Improve text contrast and spacing of envelope list

jancborchardt:
- Subject text in regular text color instead of maxcontrast
- Slightly reduced line height (to nice 130%), so lines of a single mail belong together, while mails are more clearly separated

(Also just FYI @ChristophWurst, `--color-text-lighter` is deprecated, we only use either `--color-main-text` or `--color-text-maxcontrast`.)

Before | After
-|-
image|image

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -578,12 +578,14 @@ export default {
 		gap: 4px;
 
 		&__subject {
+			color: var(--color-main-text);
+			line-height: 130%;
 			overflow: hidden;
 			text-overflow: ellipsis;
 		}
 	}
 	&__preview-text {
-		color: var(--color-text-lighter);
+		color: var(--color-text-maxcontrast);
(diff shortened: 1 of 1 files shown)


GretaD (approved):
works

ChristophWurst:
backport?

ChristophWurst:
> Also just FYI @ChristophWurst, `--color-text-lighter` is deprecated

https://docs.nextcloud.com/server/latest/developer_manual/search.html?q=--color-text-lighter&check_keywords=yes&area=default :shrug:

jancborchardt:
@ChristophWurst yeah, it's not only deprecated, it was never advised. :D

Here are the design guidelines for devs: https://docs.nextcloud.com/server/latest/developer_manual/search.html?q=--color-text-maxcontrast&check_keywords=yes&area=default

PR #7170, opened 2022-09-06
 Give email thread some spacing at the bottom

ChristophWurst:
Long threads or small windows put the thread messages very close to the bottom. It might look better if we give them some space.

| Before | After |
|---|---|
Bildschirmfoto vom 2022-09-06 09-48-18 | Bildschirmfoto vom 2022-09-06 09-48-12 |


jancborchardt:
Even more space is good as well, like 30vh for example. We do a similar thing for Files.

jancborchardt (approved):
Noice!

PR #7178, opened 2022-09-06
Fix show-image styling

GretaD:
because of changes in nc/vue and some classes got refactor, the styling of show images aka trusted sender got broken, this fixes it

before
804-378-max

after
Screenshot from 2022-09-06 14-42-43
Screenshot from 2022-09-06 14-42-29

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -181,8 +181,7 @@ export default {
 		overflow-y: auto;
 	}
 }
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
:rocket:

ChristophWurst:
do we need a backport to stable1.14?

GretaD:
yes, most probably because the problem started i think with the ncbutton.vue component. My stable1.14 is in maintenance mode right now and i cannot test

Issue #6401, opened 2022-05-10
Retry message send on fail


dodedodo:
### Steps to reproduce

This one is not easily reproducable. You need an email server that  will error on the first try, and succeed on the 2nd.

1. Send email
2. Have email server (SMTP) return an error the first time
3. Message remain in outbox forever.

### Expected behavior

1. Send email
2. Have email server (SMTP) return an error the first time
3. Nextcloud mail detects an error and automatically tries again in 30 seconds, with a maximum of 5 retries.
4. On the second try the mail actually sends.

### Actual behavior

My personal mailserver has an odd warmup issue where it does not send the first message after x hours of inactivity. The how and why is outside of the scope of this issue. In general, mail services can be temporarily unavailable for numerous reasons.

Nextcloud mail used to tell me my mail did not send properly and offer me a button to 'retry'. With the new outbox system, this changes. I first get a toast pop-up notification that says my email is sent (but it's not. It is retained in outbox because of the 'undo send' feature). After some 30 seconds nextcloud mail shows a toast pop-up 'Could not send message'. 

At this point I might not even notice the error message because I already saw the 'Message sent' toast message and clicked away. 

The actual message will never send without manual intervention. That is; go to outbox and click 'send now'. 

### Mail app version

1.12.0

### Mailserver or service

Self-hosted mailserver (mailcow)

### Operating system

Docker

### PHP engine version

PHP 8.0

### Web server

Other

### Database

MariaDB

miaulalala:
I'd say this is an enhancement, not necessarily a bug, as manually resending the message is still possible. We've discussed this internally before - some visual enhancements were also proposed to signify a message that has failed. So +1 from me.

Thanks for opening the ticket!

dodedodo:
Hey Miaulalala, thx for your response.

I get what you're saying, but I'm not sure I completely agree. The feedback a user gets (message sent notification) is sometimes plain incorrect. The code may be working as expected. But from a user's perspective, the application / UI does not work as expected.

PR #7180, opened 2022-09-06
Retrying to send fail messages (enh)

sazanof:
**Hello everyone!**
I will be glad if you consider this improvement and give your comments. =)
Maybe it refers to https://github.com/nextcloud/mail/issues/6401

Basically there are 2.5 improvements:

- Batch sending of emails with the failed status
- - The interface for sending a single message has been slightly redesigned
- Background sending of emails with the failed status when the page loads

And here are the screenshots:

Peek 2022-09-06 16-03

Peek 2022-09-06 16-04

Perhaps another processing logic is needed here, so I would like to discuss this. 
Thanks!

### DIFF
--- a/src/components/NavigationOutbox.vue
+++ b/src/components/NavigationOutbox.vue
@@ -27,7 +27,10 @@
 		:title="t('mail', 'Outbox')"
 		:to="to">
 		<template #icon>
-			<IconOutbox
+			<IconLoading v-if="sending"
+				class="outbox-sending-icon"
+				:size="20" />
+			<IconOutbox v-else
 				class="outbox-opacity-icon"
 				:size="20" />
 		</template>
@@ -42,6 +45,10 @@
 <script>
 import { NcAppNavigationItem as AppNavigationItem, NcCounterBubble as CounterBubble } from '@nextcloud/vue'
 import IconOutbox from 'vue-material-design-icons/InboxArrowUp'
+import IconLoading from 'vue-material-design-icons/Loading'
+
+const RETRY_COUNT = 5
+const RETRY_TIMEOUT = 10000
 
 export default {
 	name: 'NavigationOutbox',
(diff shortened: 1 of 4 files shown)


jancborchardt (changes requested):
Very nice work! :) Design feedback:
- [x] The "Retry" icon button on the right might be better as just a text-only tertiary button with text "Retry".
- [x] The loading spinner doesn’t seem to be the standard one from the Vue components?
- [ ] For long subject or error subline, the text needs to be ellipsized a bit earlier to not overlap with the icon

sazanof:
> Very nice work! :) Design feedback:
> 
> * [ ]  The "Retry" icon button on the right might be better as just a text-only tertiary button with text "Retry".
> * [ ]  The loading spinner doesn’t seem to be the standard one from the Vue components?
> * [ ]  For long subject or error subline, the text needs to be ellipsized a bit earlier to not overlap with the icon

@jancborchardt  done. Please take a look.

src/components/NavigationOutbox.vue
```diff
@@ -60,11 +76,55 @@ export default {
...
+				}, RETRY_TIMEOUT)
+
+			} else {
+
+				this.retry = false
+				this.attempts = 0
+			}
+		},
+	},
+	created() {
```
ChristophWurst, line 93: this means with every page load we retry sending those messages. I'm not sure if this is how we want it.

shouldn't each  message have some kind of retry counter so it's only retried a few times? with the first iteration of the outbox feature we continuously retried sending failed messages. but that caused more trouble than expected. so this was deliberately disabled via https://github.com/nextcloud/mail/pull/6602.

also see https://github.com/nextcloud/mail/issues/6540. there are some side effects with blindly resending ever pending message in the outbox.

ChristophWurst: Let's try with a counter :+1:

ChristophWurst:
also friendly ping here

src/components/NavigationOutbox.vue
sazanof: **Hello everyone!** 
Is it better to turn on the counter when the page loads, or to attract something like a permanent storage (localstorage)? Although probably the latter has more chances for bugs?

ChristophWurst: I'd say the counter should be in the backend

ChristophWurst:
https://github.com/nextcloud/mail/pull/9364 and related handle this on the backend. The frontend trigger is obsolete

PR #7181, opened 2022-09-06
Prevent unwanted live timestamp replacement

ChristophWurst:
Regression of https://github.com/nextcloud/mail/pull/7105.

## Steps to reproduce
1) Open a message thread
2) Wait ~1m

Main: the timestamp will be replaced by a long version
Here: the timestamp stays short

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-09-06 15-13-48 | Bildschirmfoto vom 2022-09-06 15-13-24 |


kesselb (approved):
Good catch :+1:

PR #7187, opened 2022-09-06
Fix important messages showing in PI's *Other* section

ChristophWurst:
Regression of https://github.com/nextcloud/mail/pull/6967.

Before: *Other* shows important but unread messages.
After: *Other* never shows any messages marked as important.


st3iny (approved):
Tested and works

PR #7189, opened 2022-09-06
Design update of threading view

jancborchardt:
As per mockup:
- Remove the extra boxes with shadow, which look strange with the new design which already has containers
- Introduce nice slight lines (not full width) with a bit of added vertical space for separation

Before | After
-|-
Screenshot from 2022-09-06 16-12-14|Screenshot from 2022-09-06 16-16-52

Please review @ChristophWurst @GretaD

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -463,15 +463,18 @@ export default {
 	.envelope {
 		display: flex;
 		flex-direction: column;
-		box-shadow: 0 0 10px var(--color-box-shadow);
-		border-radius: 16px;
 		margin-left: 10px;
 		margin-right: 10px;
 		background-color: var(--color-main-background);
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Code looks good

GretaD:
im no designer but i would prefer a full line, somehow now it looks like the line its not loaded fully, also, maybe show the line also when the message is closed because right now it looks a bit too plain. But just an idea.
Screenshot from 2022-09-06 16-42-12

jancborchardt:
Good point @GretaD, also came up in a discussion with Frank and Christoph, I will send a new Pull request.

@GretaD also though, when there is only a single message, it doesn’t make sense to close, right? Could we prevent that somehow?

jancborchardt:
Updated pull request at https://github.com/nextcloud/mail/pull/7198

PR #7194, opened 2022-09-06
Fix double <ul> in nav

ChristophWurst:
The slot renders our provided `<li>`s into a `<ul>`. We don't have to wrap it ourselves.

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-09-06 16-59-19 | Bildschirmfoto vom 2022-09-06 16-59-02 |

Ref https://github.com/nextcloud/nextcloud-vue/issues/3184

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -22,49 +22,47 @@
 <template>
 	<AppNavigation>
 		<template #list>
-			<ul id="accounts-list">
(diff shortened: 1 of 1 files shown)


GretaD (approved):
didnt test

sazanof (approved):
Tested, works

jancborchardt (approved):
Looks great design-wise now! :)

Issue #7172, opened 2022-09-06
mail.css and mobile.css missing


kesselb:
### Steps to reproduce

1. Open mail app
2. Open network inspector
3. Reload page
4. See a 302 request for mail.css and mobile.css
5. See the files are missing

### Expected behavior

Don't request mail.css and mobile.css ;)

### Actual behavior

A request for mail.css and mobile.css

### Mail app version

2ef0f2b63f32af0bdf9de0b6fb5eff3c1f5bb799

ChristophWurst:
That is because of the scss drop in server. It is no longer compiled.

PR #7202, opened 2022-09-07
Fix inclusion of mail.scss and mobile.scss

ChristophWurst:
When server dropped the SCSS compilation those styles silently vanished. They are now moved into the bundle. So this can also seen as a micro optimization and removing two request per page load.

Fixes https://github.com/nextcloud/mail/issues/7172

### DIFF
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -304,23 +304,6 @@
 	opacity: .5;
 }
 
-/* icons */
-@include icon-black-white('inbox', 'mail', 1);
-@include icon-black-white('important', 'mail', 1);
-
-.icon-flagged {
-	@include icon-color('star', 'mail', $color-black);
-}
-
-@include icon-black-white('drafts', 'mail', 1);
-@include icon-black-white('sent', 'mail', 1);
-@include icon-black-white('archive', 'mail', 1);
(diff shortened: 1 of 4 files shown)


css/mail.scss
```diff
@@ -304,23 +304,6 @@
...
-}
-
-@include icon-black-white('drafts', 'mail', 1);
-@include icon-black-white('sent', 'mail', 1);
-@include icon-black-white('archive', 'mail', 1);
-@include icon-black-white('junk', 'mail', 1);
-@include icon-black-white('trash', 'mail', 1);
-@include icon-black-white('reply', 'mail', 1);
-@include icon-black-white('reply-all', 'mail', 1);
-@include icon-black-white('forward', 'mail', 1);
```
ChristophWurst, line 322: this is save to remove, right @GretaD?

kesselb:
Thank you :+1: 

Nice to save two requests. I guess the css rules were still needed. 

Shall we move this rules https://github.com/nextcloud/mail/pull/7103/files to the (now available again) global stylesheet instead the TextEditor component?

ChristophWurst:
I don't think we should move. I would rather move mail.scss snippets to the individual components so that styles are in place with the component (and the Vue loader can even scope them)

Issue #7204, opened 2022-09-07
Whitespace-only text node above message-container creates unwanted whitespace above body text


jancborchardt:
This is with Mail 2.0.0-beta.2

![image](https://user-images.githubusercontent.com/925062/188842253-b94c3a72-4419-44bb-91d6-9098e9838592.png)

It doesn’t happen in the version currently deployed on cloud.nc (no screenshot as I get error opening mailbox).

PR #7205, opened 2022-09-07
Fix unwanted whitespaces in mail content

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/7204


jancborchardt (approved):
Tested and works, thanks! :)

Issue #6045, opened 2022-02-16
Message loading experience is a multi-step process and feels slow and not solid as a result


jancborchardt:
**This is how message loading looks right now:**
![HTML 4 step rendering](https://user-images.githubusercontent.com/925062/154257575-f5ffea34-3173-481a-b54b-c834aa8eab16.gif)

**The HTML message viewing experience is currently a 4-step process:**
1. The envelope view goes blank first
2. A small row/envelope appears, with a loading spinner cutting through the bottom, looking broken
3. It extends to a container which looks of a more proper height, still with spinner
4. The mail content loads fully and expands (or if very little text, retracts) to full height

**To improve this, could we do the following, or some of it?**
1. Completely cut step 1 with the blank view. We do have the avatar and sender data from the message list, the rest is generic controls, so it could all be shown directly?
2. Skip this step with the small row/envelope → can we directly go to step 3 where the envelope has a reasonable default height?
3. Several things possible here:
	- Instead of a simple spinner, we can use a skeleton view here – just like we do for Talk conversations, which makes the experience feel much quicker than a generic spinner.
	- The height in this step (new default height) could be reduced a tad to be the exact height it would be if the message had no content (e.g. if only a subject), or only 1 line of text.
	- Sometimes in this step the read/unread state of the mail toggles (see animation, unbolding and bolding of sender name) and goes back again, making it look a bit off.
4. Have a very quick expand animation (100ms quick or so) to the full height once the content is fully loaded so that it looks less jarring.

ChristophWurst:
## Break down of current HTML loading order

### Thread loading
![vlcsnap-2022-11-09-11h46m54s353](https://user-images.githubusercontent.com/1374172/200810603-b23129b3-1e31-43dd-ac22-0621e1e38868.png)

### Message loading
![vlcsnap-2022-11-09-11h46m59s137](https://user-images.githubusercontent.com/1374172/200810653-7e33010c-bb1c-467a-925a-f8315cf80f6d.png)

### ???
![vlcsnap-2022-11-09-11h47m06s683](https://user-images.githubusercontent.com/1374172/200810710-f4fe8774-c11f-4353-a4d6-f134e31c5ab7.png)

### Finished loading
![vlcsnap-2022-11-09-11h47m12s550](https://user-images.githubusercontent.com/1374172/200810757-b8b1da03-eed6-41db-9c16-336b50ab9ce8.png)

ChristophWurst:
Todo

[...]

PR #7219, opened 2022-09-07
Add skeleton for envelope list and thread

### LINKED ISSUE #6142: Improve look of priority inbox

GretaD:
fixes #6142 
and some parts of #6045

to do

- [x] add the code into a component

### DIFF
--- a/src/components/LoadingSkeleton.vue
+++ b/src/components/LoadingSkeleton.vue
@@ -1,0 +1,72 @@
+<template>
+	<div>
+		<div v-for="i in numberOfLines" :key="i" class="item-list__entry">
+			<NcAvatar v-if="withAvatar" class="item-avatar" :size="44" />
+			<div class="item__details">
+				<h3>&nbsp;</h3>
+				<p class="message">
+					&nbsp;
+				</p>
+			</div>
+		</div>
+	</div>
+</template>
+
+<script>
+import NcAvatar from '@nextcloud/vue/dist/Components/NcAvatar'
+
+export default {
+	name: 'LoadingSkeleton',
+	components: {
+		NcAvatar,
+	},
+	props: {
+		numberOfLines: {
+			type: Number,
+			default: 10,
+		},
+		withAvatar: {
+			type: Boolean,
+			default: true,
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+
+/* skeleton */
+.item-list__entry {
+	display: flex;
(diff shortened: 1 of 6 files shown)


src/components/LoadingSkeleton.vue
```diff
@@ -0,0 +1,62 @@
...
+	font-size: 100%;
+	margin: 0;
+}
+.message {
+	width: 80%;
+	height: 15px;
+	margin-top: 5px;
+}
+}
+}
```
ChristophWurst, line 61: suspicous

CarlSchwan: indentation is wrong

st3iny (approved):
Tested and works. I just added the option to make the avatar optional to make it more fitting for the mail body (because it doesn't contain avatars).

[Peek 2022-09-20 11-19.webm](https://user-images.githubusercontent.com/1479486/191220389-3ee097d4-7313-4223-a373-d6f2ea3f2b77.webm)

Issue #7212, opened 2022-09-07
Clicking the actions menu of a previewable attachment previews the attachment


ChristophWurst:
### Steps to reproduce

1. Open an attachment with images or PDFs attached
2. Open the actions menu of one of the attachments

### Expected behavior

I see the actions menu

### Actual behavior

I see the attachment preview and the actions menu on top

### Mail app version

2.0.0 beta 3

PR #7220, opened 2022-09-07
Fix opening attachments actions

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/7212


sazanof (approved):
image
Now it is opening correct.

Issue #7076, opened 2022-08-24
Rework Draft Handling - Backend Part


miaulalala:
Backend part of https://github.com/nextcloud/mail/issues/4768

miaulalala:
Note: the Outbox Backend already handles drafts via an OutboxCreatedEvent that triggers the DeleteDraftListerner. Adjust this class to delete local and IMAP drafts.

miaulalala:
Note: We do need a GET for the contents of a draft (unless we load all messages with their content?) something for the fontenders to tell me :)

ChristophWurst:
> Note: We do need a GET for the contents of a draft (unless we load all messages with their content?) something for the fontenders to tell me :)

https://github.com/nextcloud/mail/issues/4768#issuecomment-1224352762. `MessageService.js::fetchMessage`.

ChristophWurst:
## Leftovers

* [ ] T6

ChristophWurst:
UC3 doesn't work apparently. Local drafts stay local draft forever, the background job does not move idle items to IMAP.

PR #7223, opened 2022-09-07
Add local drafts backend

miaulalala:
Fixes https://github.com/nextcloud/mail/issues/7076

Move the draft handling from IMAP to our local messages mailbox.

To do:
- [ ] Provide API endpoints
- [ ] Tests
- [ ] Background Job that syncs drafts without UID to IMAP
- [ ] Background Job that syncs drafts from IMAP to the local mailbox

There was some discussion about whether the drafts box should be unified or not.

- [x] ~~Clarify whether we need to split messages on a per account basis (this should be possible to handle with a few changes)~~ Neither. Local drafts are not visible to the user ever but have an account ID assigned.

### DIFF
--- a/lib/Controller/DraftsController.php
+++ b/lib/Controller/DraftsController.php
@@ -1,0 +1,214 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Mail App
+ *
+ * @copyright 2022 Anna Larch <anna.larch@gmx.net>
+ *
+ * @author Anna Larch <anna.larch@gmx.net>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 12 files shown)


lib/Controller/DraftsController.php
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param int $id
+	 * @return JsonResponse
+	 */
+	public function show(int $id): JsonResponse {
```
miaulalala, line 76: not needed

lib/Controller/DraftsController.php
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @return JsonResponse
+	 */
+	public function index(): JsonResponse {
+		return JsonResponse::success(['messages' => $this->service->getMessages($this->userId)]);
```
miaulalala, line 66: not needed

miaulalala (commented):
Closing this as this needs a major overhaul

miaulalala:
Superseded by https://github.com/nextcloud/mail/pull/7517

Issue #7210, opened 2022-09-07
Envelope draft indicator positioning


ChristophWurst:
### Steps to reproduce

1. Compose a draft
2. Close the editor
3. Open the *Drafts* mailbox

### Expected behavior

Well designed list of drafts

### Actual behavior

![Bildschirmfoto vom 2022-09-07 14-44-06](https://user-images.githubusercontent.com/1374172/188881403-a55136c0-71be-42a0-bc3d-de783560a04e.png)

### Mail app version

2.0.0 beta 3

ChristophWurst:
Introduced by https://github.com/nextcloud/mail/pull/7150

PR #7235, opened 2022-09-08
Fix drafts prefix position

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/7210

Bildschirmfoto vom 2022-09-08 17-06-11

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -635,6 +635,15 @@ export default {
 .list-item-style.selected {
 	background-color: var(--color-background-dark);
 }
+.list-item-style {
+	.draft {
+		line-height: 130%;
+
+		em {
+			font-style: italic;
+		}
+	}
+}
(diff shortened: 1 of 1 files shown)


sazanof (approved):
Tested, look nice=)

src/components/Envelope.vue
```diff
@@ -635,6 +635,15 @@ export default {
 .list-item-style.selected {
 	background-color: var(--color-background-dark);
 }
+.list-item-style {
+	.draft {
+		line-height: 130%;
+
+		em {
+			font-style: italic;
```
jancborchardt, line 643: We do not use italic anywhere in the interface for legibility, shouldn’t do it here either. :) As per [design guidelines](https://docs.nextcloud.com/server/latest/developer_manual/design/foundations.html#typography) @ChristophWurst

Issue #7165, opened 2022-09-05
Recipient multiselect jumps with focus


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Configure an account
3. Click *New message*
4. Click into *To*

### Expected behavior

Focus change but text stays where it was

### Actual behavior

| Blur | Focus |
|---|---|
| ![Bildschirmfoto vom 2022-09-05 22-30-46](https://user-images.githubusercontent.com/1374172/188510052-91164a27-ae6f-4c95-8c34-1d4f3a656cc0.png) | ![Bildschirmfoto vom 2022-09-05 22-30-50](https://user-images.githubusercontent.com/1374172/188510050-d4bb053b-1000-473c-a1b4-834577655700.png) |

### Mail app version

2.0.0 beta 1

JuliaKirschenheuter:
After searching for this odd behavior i can't reproduce this bug anymore. If this will happens again, i'll reopen this issue.

PR #7240, opened 2022-09-09
Improving the appearance of the Composer window

sazanof:
Hello. 
Related to https://github.com/nextcloud/mail/issues/7165 
An attempt to tidy up the appearance of the drop-down list.

What I never figured out how to fix - values `+2 +5 +x` - a white field remains to the right of the recipient (this has been the case since old versions).

If it will be useful, I will be glad.

I also removed `:focus-active box-shadow`

Peek 2022-09-09 14-55

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1265,15 +1265,54 @@ export default {
 	border-top: 1px solid var(--color-border);
 	align-items: flex-start;
 
+	label {
+		padding: 11px 20px 11px 0;
+	}
+
+	::v-deep(.multiselect--multiple .multiselect__tags) {
+		display: grid;
+		grid-template-columns: calc(100% - 18px) 18px 100%;
+
+		.multiselect__limit {
+			margin-right: 0;
+			margin-left: 8px
+		}
+	}
+
+	::v-deep(.multiselect__content-wrapper) {
+		border-bottom: 1px solid var(--color-border);
+		margin-top: 0;
+
+		& li > span::before {
+			display: none
+		}
+	}
+
+	::v-deep(.multiselect__input) {
+		position: relative !important;
+		top: 0;
+		grid-column-start: 1;
+		grid-column-end: 3;
+	}
+
+	::v-deep(.multiselect--active input:focus-visible) {
+		box-shadow: none;
+	}
+
+	::v-deep(.multiselect__tags) {
+		box-sizing: border-box;
+		height: auto;
+	}
+
 	&__from {
 		margin-right: 50px; /* for the modal close button */
 	}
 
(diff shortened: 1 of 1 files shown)


GretaD:
I think :`clear-on-select` should stay true, otherwise, if want to add greta and i type greta, text stays there, then i want to add Mikhail, i have to delete Greta first and then start typing the new name. I think it takes a lot of time if i have to add many people. 

The rest looks good

sazanof:
@GretaD thenk you for review! That's right. 

No problems. I understood why I put it (clear-on-select="false"): there are more than 500 people in my contacts. And if when the field is cleared, then all of them are loaded into Multiselect as options. And when I tested this,  browser freezes for a while =) again and again.
And so it is every time when cleansing. Maybe there is some way to limit the number of options? What do you think?

GretaD:
> @GretaD thank you for review! That's right.
> 
> No problems. I understood why I put it (clear-on-select="false"): there are more than 500 people in my contacts. And if when the field is cleared, then all of them are loaded into Multiselect as options. And when I tested this, browser freezes for a while =) again and again. And so it is every time when cleansing. Maybe there is some way to limit the number of options? What do you think?

I think on the popup are either the ones you recently contacted or the ones that match the text you added, not all the contacts you have, so not sure why it loads all the contacts for you

sazanof:
> I think on the popup are either the ones you recently contacted or the ones that match the text you added, not all the contacts you have, so not sure why it loads all the contacts for you

Try the following:
In the "To" field, write one letter (say "g") (`/api/autoComplete?term=g`)
Waiting for the `/api/autoComplete?term=g`
Then after a second quickly add "reta" (`/api/autoComplete?term=greta`)

You will have !one! a suitable contact. Select it. 

It will be added, and then, when the multiselect clears the value, all the contacts that were obtained after calling `/api/autoComplete?term=g` **will added to Multiselect options**

GretaD:
> > I think on the popup are either the ones you recently contacted or the ones that match the text you added, not all the contacts you have, so not sure why it loads all the contacts for you
> 
> Try the following: In the "To" field, write one letter (say "g") (`/api/autoComplete?term=g`) Waiting for the `/api/autoComplete?term=g` Then after a second quickly add "reta" (`/api/autoComplete?term=greta`)
> 
> You will have !one! a suitable contact. Select it.
> 
> It will be added, and then, when the multiselect clears the value, all the contacts that were obtained after calling `/api/autoComplete?term=g` **will added to Multiselect options**
> 
> _(Selectable recipients - taken from the first request)_

maybe we open this as an issue and see how we can improve it. I merged this one as it does deliver the improvement that claims :)

PR #7241, opened 2022-09-09
Release v1.14.0-rc.1

GretaD:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 1.14.0-rc.1 - 2022-09-09
+### Fixed
+- Overlapped composer modal close button
+- Element alignment of thread message
+- Unwanted whitespaces in mail content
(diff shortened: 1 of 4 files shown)


ChristophWurst (approved):
One last time: we don't need changelogs for each prerelease. There should be one for v1.14. Nothing more. Nothing less.

Anything that doesn't change between the last release and the next also won't go into the changelog. E.g. all the regressions that we caught before the release are not worth mentioning because they don't fix anything for the end users.

GretaD:
> One last time: we don't need changelogs for each prerelease. There should be one for v1.14. Nothing more. Nothing less.
> 
> Anything that doesn't change between the last release and the next also won't go into the changelog. E.g. all the regressions that we caught before the release are not worth mentioning because they don't fix anything for the end users.

ah, my understanding was that we dont split them into beta..rc..and such on changelog but have the general one i.e 1.14 and add everything there, like here for example: https://github.com/nextcloud/mail/pull/7242/files

PR #7245, opened 2022-09-09
Bump CKEditor to v35.1

ChristophWurst:
Tested and works.


GretaD (approved):
didnt test, let me know if i must

PR #7280, opened 2022-09-15
Fix new message button header transparency and position

meichthys:
Fixes transparency on new message button.

Current:
image

Changes:
image

### DIFF
--- a/src/components/NewMessageButtonHeader.vue
+++ b/src/components/NewMessageButtonHeader.vue
@@ -79,8 +79,10 @@ export default {
 
 <style lang="scss" scoped>
 .header {
-	display: flex;
+	display: absolute;
 	align-items: center;
+	position: inherit;
+	background-color: transparent;
 	padding: 8px 8px 8px 48px;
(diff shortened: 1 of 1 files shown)


meichthys:
@GretaD I'm on  2.0.0-beta.6 using the built in dark theme on nextcloud.
I've tried both dark and light themes in firefox without any difference.

meichthys:
> Looks like [nextcloud/richdocuments#2426](https://github.com/nextcloud/richdocuments/issues/2426)

Yes, this looks right. I noticed some office css being applied but wasn't sure if that was expected or not.

ChristophWurst:
https://github.com/nextcloud/richdocuments/pull/2432 should fix the office style leaks

ChristophWurst:
Nevertheless thanks a lot for taking care of this design bug, @meichthys! Highly appreciated :)

meichthys:
NP 👍
Sometimes I feel like i must be the only one using Mail as a daily driver, but I'm going to stick it out!

PR #7284, opened 2022-09-16
Implement new dashboard api

### LINKED ISSUE #7112: Adapt unread mail dashboard widget for mobile clients

With https://github.com/nextcloud/server/pull/33658 mobile clients can access app widget data.

Adapt the *Unread mail* widget to https://github.com/nextcloud/server/pull/33658 once it's done.

_No response_

### Additional context

_No response_

### LINKED ISSUE #7113: Adapt important mail dashboard widget for mobile clients

With https://github.com/nextcloud/server/pull/33658 mobile clients can access app widget data.

Adapt the *Important mail* widget to https://github.com/nextcloud/server/pull/33658 once it's done.

_No response_

### Additional context

_No response_

eneiluj:
closes #7112
closes #7113

Implement IAPIWidget, IIConWidget and IOptionWidget for the client oriented dashboard API.
IButtonWidget is not needed here if, like in the frontend dashboard, the widgets don't have a "see more" button.

### DIFF
--- a/lib/Dashboard/MailWidget.php
+++ b/lib/Dashboard/MailWidget.php
@@ -30,12 +30,15 @@ namespace OCA\Mail\Dashboard;
 use OCA\Mail\AppInfo\Application;
 use OCA\Mail\Service\AccountService;
 use OCP\AppFramework\Services\IInitialState;
+use OCP\Dashboard\IIconWidget;
+use OCP\Dashboard\IOptionWidget;
 use OCP\Dashboard\IWidget;
+use OCP\Dashboard\Model\WidgetOptions;
 use OCP\IL10N;
 use OCP\IURLGenerator;
 use OCP\Util;
 
-abstract class MailWidget implements IWidget {
+abstract class MailWidget implements IWidget, IIconWidget, IOptionWidget {
 	/** @var IL10N */
 	protected $l10n;
 
@@ -80,6 +83,15 @@ abstract class MailWidget implements IWidget {
 	/**
 	 * @inheritDoc
 	 */
+	public function getIconUrl(): string {
+		return $this->urlGenerator->getAbsoluteURL(
+			$this->urlGenerator->imagePath(Application::APP_ID, 'mail.svg')
+		);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
 	public function getUrl(): ?string {
 		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->linkToRoute('mail.page.index'));
 	}
@@ -94,5 +106,12 @@ abstract class MailWidget implements IWidget {
 			'mail-accounts',
 			$this->accountService->findByUserId($this->userId)
 		);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getWidgetOptions(): WidgetOptions {
+		return new WidgetOptions(true);
 	}
 }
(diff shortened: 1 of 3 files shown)


lib/Dashboard/MailWidget.php
```diff
@@ -95,4 +124,88 @@ public function load(): void {
...
+			});
+			foreach ($mailboxes as $mailbox) {
+				$searchResult = $this->mailSearch->findMessages(
+					$account,
+					$mailbox,
+					$filter,
+					null,
+					10
+				);
+				$emails = array_merge($emails, $searchResult);
```
ChristophWurst, line 159: array_merge in a loop is considered a performance anti-pattern. I think this needs to be restructured a bit so that array_merge is only run once after the nested loops

lib/Dashboard/MailWidget.php
```diff
@@ -95,4 +124,88 @@ public function load(): void {
...
+		$emails = [];
+
+		foreach ($accounts as $account) {
+			$mailboxes = $this->mailManager->getMailboxes($account);
+			/** @var Mailbox[] $mailboxes */
+			$mailboxes = array_filter($mailboxes, function (Mailbox $mailbox) {
+				return $mailbox->isInbox();
+			});
+			foreach ($mailboxes as $mailbox) {
+				$searchResult = $this->mailSearch->findMessages(
```
ChristophWurst, line 152: can't we use ``\OCA\Mail\Contracts\IMailSearch::findMessagesGlobally``? that eliminates looping over accounts and mailboxes

it sounds crazy but we have seen plenty of accounts where users have like 200 mailboxes. And if you loop over them this operation takes a lot of queries and time

lib/Dashboard/MailWidget.php
```diff
@@ -95,4 +124,88 @@ public function load(): void {
...
+		return $emails;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getItems(string $userId, ?string $since = null, int $limit = 7): array {
+		$emails = $this->getEmails($userId);
+
+		if ($since !== null) {
```
ChristophWurst, line 175: it would pass this down to the `getEmails` call so irrelevant messages are never loaded into memory

ChristophWurst (changes requested):
Some performance considerations

lib/Dashboard/MailWidget.php
eneiluj: I didn't see this method. Much better indeed.

lib/Dashboard/MailWidget.php
eneiluj: Is it possible to filter by `send_at`? If so, I didn't get how to do it by looking at `FilterStringParser::parse()`.

tobiasKaminsky:
Tested and works :clap

lib/Dashboard/MailWidget.php
```diff
@@ -95,4 +110,72 @@ public function load(): void {
...
+			return [];
+		}
+		$filter = $this->getSearchFilter();
+		$emails = $this->mailSearch->findMessagesGlobally($user, $filter, null, $limit);
+
+		if ($minTimestamp !== null) {
+			return array_filter($emails, function (Message $email) use ($minTimestamp) {
+				return $email->getSentAt() > $minTimestamp;
+			});
+		}
```
ChristophWurst, line 140:
```suggestion
		$emails = $this->mailSearch->findMessagesGlobally($user, $filter, $minTimestamp, $limit);
```

https://github.com/nextcloud/mail/blob/a40d8751d75420635a453150b100646b623391a3/lib/Db/MessageMapper.php#L764-L768

Assuming this is the comparison we want

eneiluj: Thanks but that selects all the messages that are older than a given timestamp. We want those that are more recent.

eneiluj:
Rebased on main.

PR #7303, opened 2022-09-21
Improve HTTP caching

ChristophWurst:
* More immutable cache headers
* Cache itinerary responses
* Cache quota
* Cache contacts integration lookups
* Cache recipient auto completion lookups
* Cache auto config lookups

## How to test
1) Open the app with master
2) Open a mailbox
3) Open a thread
4) Open another thread
5) Go back to the first thread -> this should feel a bit slow

Then switch to this branch and do the same steps.

It should go without saying but don't have your browser cache disabled or you won't see a difference.

### DIFF
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -402,6 +402,6 @@ class AccountsController extends Controller {
 		if ($quota === null) {
 			return MailJsonResponse::fail([], Http::STATUS_NOT_IMPLEMENTED);
 		}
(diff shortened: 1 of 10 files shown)


st3iny (approved):
Tested and the app feels quicker. I also tested the recipient search and the responses are properly cached.

The code looks good too.

CarlSchwan (approved):
Yeah for more caching!

ChristophWurst:
This has to work

PR #7306, opened 2022-09-22
Adjust testing matrix for Nextcloud 25 on main

nickvergessen:
- [x] Requires https://github.com/nextcloud/mail/pull/7307

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,9 @@ jobs:
       matrix:
         php-versions: ['7.4', '8.0', '8.1', '8.2']
         nextcloud-versions: ['master']
+      include:
+        - php-versions: 8.0
+          nextcloud-versions: stable25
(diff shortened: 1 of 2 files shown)


miaulalala:
this needs to be a bit more elaborate. Stable branches lower than 2 need to run against NC stable24. everything higher needs to run against 25+

cc @ChristophWurst

CarlSchwan (approved):

.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         php-versions: ['7.4', '8.0', '8.1', '8.2']
-        nextcloud-versions: ['master']
+        nextcloud-versions: ['master', 'stable25']
```
ChristophWurst, line 19: let's include just one job or we will go from 4 test jobs to 8. 5 should be plenty.

PR #7307, opened 2022-09-22
Add Nextcloud 26 support

nickvergessen:


ChristophWurst (changes requested):
Let's wait until v2 is branched off

ChristophWurst (approved):

PR #7308, opened 2022-09-22
Add skeleton for loading state

GretaD:
A skeleton will be shown when we initialize the mailbox.
The skeleton will not be shown when the page is loaded in the background or when we click refresh button. 
The reason for that is that after the list is already shown, when the automatic refreshing kicks in, the list will be transformed into skeleton seconds after it is shown. So, having the loading icon for that, should be fine.

Also, the skeleton doesnt replace the loadingCacheInitialization, there we should have the loading message, because it takes a while to populate the inbox/mailbox and the user should know what is happening, instead of having to look at a skeleton for a bit.

### DIFF
--- a/src/components/LoadingSkeleton.vue
+++ b/src/components/LoadingSkeleton.vue
@@ -1,0 +1,99 @@
+<template>
+	<div>
+		<div v-for="i in numberOfLines" :key="i" class="item-list__entry">
+			<div
+				class="item-avatar" />
+			<div class="item__details">
+				<h3>&nbsp;</h3>
+				<p class="message">
+					&nbsp;
+				</p>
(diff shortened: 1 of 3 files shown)


GretaD:
my inbox is very fast today and i cannot record better view of the skeleton, but change the speed to look at it better.

[Screencast from 14.10.2022 13:44:18.webm](https://user-images.githubusercontent.com/12728974/195841391-3a5a1a03-14df-4dd6-a0db-d91c4d2897f1.webm)

GretaD:
Screenshot from 2022-10-14 14-05-59

ChristophWurst (approved):
Works!

PR #7355, opened 2022-10-01
Archiving messages

### LINKED ISSUE #22: Archiving messages
Messages should be archivable by simple clicks:
- This means that the message(s) should be moved to e.g. Archives/YYYY/MM, Archives/YYYY or Archives/YYYY-MM depending on how the user configures it
- Moving in one folder only may make it one day extremely large and a fast pre-selection of messages is not possible (e.g. you know the message came August, 2016, then choose Archives/2016/08/ and start searching there)

Splitted from #15

arjentz:
Figured this couldn't be too hard. I don't know how to generate migrations, I edited the database manually and it worked as I expected.

Fixes #22


ChristophWurst:
Migrations are covered by https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/migrations.html. We can help with that part.

arjentz:
Thanks! Do you want me to generate and add the migrations to the PR? Or will you add them before a new release?

ChristophWurst:
If you can please add it :)

src/components/Envelope.vue
```diff
@@ -251,6 +261,7 @@ import IconCreateEvent from 'vue-material-design-icons/Calendar'
 import CheckIcon from 'vue-material-design-icons/Check'
 import ChevronLeft from 'vue-material-design-icons/ChevronLeft'
 import DeleteIcon from 'vue-material-design-icons/Delete'
+import ArchiveIcon from 'vue-material-design-icons/Archive'
```
GretaD, line 264: for the archive icon, we use /PackageDown

src/components/ThreadEnvelope.vue
```diff
@@ -115,6 +115,16 @@
 								:size="20" />
 						</template>
 					</ButtonVue>
+					<ButtonVue v-if="showArchiveButton"
+						:close-after-click="true"
+						type="tertiary-no-background"
+						@click.prevent="onArchive">
+						<template #icon>
+							<ArchiveIcon
+								:title="t('mail', 'Archive message')"
```
GretaD, line 124: we dont use the :title anymore, if this is a label, add it below as {{ t('mail', 'Archive message') }}

src/components/Envelope.vue
arjentz: Fixed it

src/components/ThreadEnvelope.vue
arjentz: Since this is part of the `ThreadEnvelope` it is for the mouseover event, I just copied it from the `DeleteIcon` a few lines below this

arjentz:
> If you can please add it :)

Added it, hope it is right this way (I checked the one where the other mailbox shortcuts where added)

ChristophWurst:
I have no mailbox marked for ``\\Archive`` so it picked my trash mailbox

Bildschirmfoto vom 2022-10-04 14-11-55

Bildschirmfoto vom 2022-10-04 14-10-16

ChristophWurst (approved):
:+1: Works. Thanks a lot!

matbgn:
Hi,

Any news for the deployment? I've successfully upgraded my NC instance to 25.0.1 and got NCMail v2.1.0 but no Archive button appearance. Any suggestion? Will it be postponed to 2.2.0 instead of actual tagged version here in GitHub?

PR #7356, opened 2022-10-01
Update ::v-deep to :deep and fix style issue

arjentz:
Fixes `[@vue/compiler-sfc] ::v-deep usage as a combinator has been deprecated. Use :deep(<inner-selector>) instead.`


ChristophWurst:
Thanks a lot @arjentz!

GretaD (approved):
I tested mostly of the components that Arjen changed and it works correctly and no errors on my console :partying_face:

PR #7365, opened 2022-10-01
Toggle priority inbox

### LINKED ISSUE #3265: Allow to disable "Priority Inbox" (or set another starting inbox)
### Feature Request
WIth release 1.4.0 the mail app introduces the "Priority Inbox" which is set as the default inbox when opening the mail app.

I'd like to have a setting to disable the "Priority Inbox" completely or at least set a another inbox (especially the "shared inbox" like in previous app version) as the default inbox to open.

arjentz:
Did some work on this, as I read that it's basically a front-end only issue. I think it does everything it should do (it hides the priority inbox based on a setting and organizes the "special" mailboxes accordingly). However, I do not seem to get the startMailbox set to 'unified' instead of 'priority', how to fix this?

Fixes #3265

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -153,6 +153,7 @@ class PageController extends Controller {
 				'collect-data' => $this->preferences->getPreference($this->currentUserId, 'collect-data', 'true'),
 				'start-mailbox-id' => $this->preferences->getPreference($this->currentUserId, 'start-mailbox-id'),
 				'tag-classified-messages' => $this->preferences->getPreference($this->currentUserId, 'tag-classified-messages', 'true'),
+				'show-priority-inbox' => $this->preferences->getPreference($this->currentUserId, 'show-priority-inbox', 'true'),
 			]);
 		$this->initialStateService->provideInitialState(
 			'prefill_displayName',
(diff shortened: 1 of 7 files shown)


ChristophWurst:
If PI is turned off you either have to start in unified inbox for multi account setup or the first inbox if there is only one account

arjentz:
I figured the unified mailbox would also work fine if only one account is set up? However, I do not seem to get it work (even if I change `start-mailbox-id` to `unified` in the database). Any pointers on where to go from here?

szaimen:
@arjentz can you please fix the conflicts?

arjentz:
> @arjentz can you please fix the conflicts?

Rebased it and fixed the conflicts

arjentz:
> If PI is turned off you either have to start in unified inbox for multi account setup or the first inbox if there is only one account

@ChristophWurst Do you maybe have some pointers on how to achieve this behaviour? I tried by setting `start-mailbox-id` to `unified`, but it didn't seem to work that way.

szaimen:
@arjentz can you do as advised and fix the tests? :)

arjentz:
@szaimen I haven't found the time to do so yet, but I hope to look at it some time soon

ChristophWurst:
Friendly ping. Will you have time to finish this in the near future or shall we close the PR?

arjentz:
Thanks for the ping. Unfortunately, I don't think I will find the time to finish it in the near future. So it's better to close it for now.

PR #7374, opened 2022-10-03
Fix background for default folders

GretaD:
before
Screenshot from 2022-10-03 15-49-27

after
Screenshot from 2022-10-03 15-46-05


ChristophWurst (approved):
Should this not go into `@nextcloud/vue`? Why is this fixed here?

GretaD:
> Should this not go into `@nextcloud/vue`? Why is this fixed here?

because its not a multiselect but a treeselect: https://github.com/nextcloud/mail/blob/main/src/components/MailboxInlinePicker.vue#L11

Issue #6341, opened 2022-05-02
Restructure \OCA\Mail\Db\MessageMapper::updateBulk to run in O(1)


ChristophWurst:
### Is your feature request related to a problem? Please describe.

If you look at ``\OCA\Mail\Db\MessageMapper::updateBulk`` then you can see a loop of all messages to update and one query per message.

Putting the perm flags (tags) aside, there is a fixed number of updates we do
* Set/unset *replied* flag
* Set/unset *draft* flag
* ...

This is one `UPDATE oc_mail_messages SET flag_answered = x, flag_draft = y, ...  WHERE uid = a AND mailbox = b` query per message. Hence a complexity of O(n).

### Describe the solution you'd like

Run the updates in bulk for each column so that there is a fixed but small number of queries:

* `UPDATE oc_mail_messages SET flag_answered = 0  WHERE uid in (a,b,c) AND mailbox = d`
* `UPDATE oc_mail_messages SET flag_answered = 1  WHERE uid in (e,f,g) AND mailbox = d`
* ...

We have 10 flag_* columns. That means it takes 20 queries to update all messages. That sounds a lot but the bulk update is called with 20 messages when triggered from the front-end and the full mailbox when triggered from a background sync. So the number of queries will be as high as before in the worst case and significantly lower in the best case.

### Describe alternatives you've considered

N/a

### Additional context

This is just a brain dump and I can not proof that this actually improves the sync. @miaulalala @st3iny this should be something to tinker about nevertheless :)

PR #7375, opened 2022-10-03
Add lean sync for servers without CONDSTORE

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6341

This updates flags in 20+ queries. Bigger mailboxes require more queries. But still, if you  have 150k messages the sync ran 150k updates before, now it runs 20k. If we say that only read/unread and important flags are relevant we can cut it to 600 queries.

It aint much, but it it's an honest optimization.

## Todo
- [ ] Doing
- [ ] Testing

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -25,6 +25,8 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Db;
 
+use Horde_Imap_Client;
+use InvalidArgumentException;
 use OCA\Mail\Account;
 use OCA\Mail\Address;
 use OCA\Mail\AddressList;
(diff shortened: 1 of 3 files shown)


lib/Db/MessageMapper.php
```diff
@@ -429,6 +431,25 @@ public function updateBulk(Account $account, bool $permflagsEnabled, Message ...
...
+	 */
+	public function updateFlagBulk(Mailbox $mailbox, string $flag, bool $set, array $uids): void {
+		$col = $this->hordeFlagToColumnName($flag);
+		$query = $this->db->getQueryBuilder();
+		$query->update($this->getTableName())
+			->set($col, $query->createNamedParameter($set, IQueryBuilder::PARAM_BOOL))
+			->where(
+				$query->expr()->eq('mailbox_id', $query->createNamedParameter($mailbox->getId(), IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT),
+				$query->expr()->eq($col, $query->createNamedParameter(!$set, IQueryBuilder::PARAM_BOOL), IQueryBuilder::PARAM_BOOL),
+				$query->expr()->in('uid', $query->createParameter('uids'), IQueryBuilder::PARAM_INT_ARRAY),
```
rullzer, line 445: I'm not sure if possible with all your fancy pancy supported databases

but an index like

`('mailbox_id', 'uid') INCLUDING (flag1, flag2, flag3, flag4, flag5, flag6)` would allow the filtering to happen mainly on the index level. Without going to the actual data.

Issue #5071, opened 2021-05-25
Missing visual message drop target


kesselb:
https://user-images.githubusercontent.com/3902676/119572138-99608500-bdb2-11eb-96aa-d70fe942a981.mp4

**How to reproduce**

- Open a thread
- Drag a message from conversation view to another mailbox

**Mail app version** master

GretaD:
The scope of this Issue should change, the drag and drop now works, the problem is that you cannot visually see which mailbox is selected.

[Screencast from 23.09.2022 15:03:46.webm](https://user-images.githubusercontent.com/12728974/191966647-fcb12c43-cee0-44dc-91ce-d8137d7d4493.webm)

PR #7378, opened 2022-10-04
Fix missing visual drop target on nav mailboxes

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/5071

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-10-04 15-06-58 | Bildschirmfoto vom 2022-10-04 15-05-45 |


JuliaKirschenheuter (approved):
tested and works

PR #7403, opened 2022-10-10
Bump ckeditor5 35.2

GretaD:
Please test any of the drop down tools, like for example font family or font size. I have a bug, when i click the dropdown, it only shows for a sec and then vanishes. Its also on master, but please test it here so i know it not something on my browser


ChristophWurst (changes requested):
Can confirm that this is bugged

On v2.0.0 I can see
Bildschirmfoto vom 2022-10-13 17-35-45
but here I can't.

ChristophWurst:
On latest main I can also open the sub menus

Bildschirmfoto vom 2022-10-13 17-54-07

ChristophWurst:
I did the upgrade myself from latest main. Same result. This could indicate a bug in CKEditor or a new incompatibility with one of our other libraries or components.

ChristophWurst:
[](url)

PR #7419, opened 2022-10-13
Bump @nextcloud/vue to v7

ChristophWurst:
No more prerelease. Yay.


JuliaKirschenheuter (approved):

szaimen:
I guess the backport would be good since at least this bug is not fixed without: https://github.com/nextcloud/mail/issues/7193

PR #7420, opened 2022-10-13
Allow changing an account's auth method via REST

ChristophWurst:
Extracted from https://github.com/nextcloud/mail/pull/6830 to make the PR reviewable.


miaulalala (approved):
LGTM

Issue #6591, opened 2022-05-31
Microsoft 365 changes Oct 1 2022 and the effect on this app


bwragg:
### Is your feature request related to a problem? Please describe.

As announced here:

[https://techcommunity.microsoft.com/t5/exchange-team-blog/basic-authentication-and-exchange-online-september-2021-update/ba-p/2772210](url)

& 

[https://techcommunity.microsoft.com/t5/exchange-team-blog/basic-authentication-deprecation-in-exchange-online-may-2022/ba-p/3301866](url)

Microsoft are going to turn Basic Auth for IMAP & POP from October 1 2022. No exceptions.

I've been searching for a way around this but there doesn't seem to currently be one apart from stopping using the Mail app.

### Describe the solution you'd like

Is there another way in the for connecting to 365 (Exchange Online) accounts? The documents above link to this article:

[https://docs.microsoft.com/en-gb/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth](url)

This is something that could be/is planned to be implemented? Or is the approach not suitable for Nextcloud (due to the mail app fetching via IMAP done on the server side). Is there another way it could be done?

ChristophWurst:
I'm looking into https://github.com/nextcloud/mail/issues/6454 right now. The changes sound similar. Gmail has already shut down their basic IMAP logins, for Outlook we have a bit of time.

andyxheli:
@ChristophWurst 
https://learn.microsoft.com/en-us/exchange/clients-and-mobile-in-exchange-online/deprecation-of-basic-authentication-exchange-online

Looks like MS did disable Basic authentication for Outlook, EWS, RPS, POP, IMAP, and EAS protocols in Exchange Online.

Is there any talk about supporting Exchange Active Sync with the Mail App? Since that's the standard and recommended solution for  Exchange Online account these days. Especially with 2fa being enforced on the O365 tenants + alot of Orgs disable from allowing them to create app password and tell you the truth most users wouldn't even know how if you show them too

ChristophWurst:
This app is an IMAP client. I don't think any other protocol fits the architecture at the moment.

ChristophWurst:
The changes are done: https://github.com/nextcloud/mail/pull/7722. Time for a technical review and testing by developers.

Issue #6454, opened 2022-05-14
GMail compatibility without app passwords


Hal1512:
### Is your feature request related to a problem? Please describe.

GMail considers the integration from this Mail app to be 'less secure'.   They claim on their site that they will no longer support apps that log into your mail account with only user name and password.   

https://support.google.com/accounts/answer/6010255?hl=en 

### Describe the solution you'd like

I believe that if the mail connector also asked for and presented an App Password to GMail, that qualifies as a more secure way to connect.

https://support.google.com/accounts/answer/185833

feutl:
This is an issue now!
I have 2 accounts which do not work with nextcloud mail any more. The workaround to use 2FA and a app password is not a sufficient solution.
Fairemail fixed the issue on Android and let me sync my mails without setting up 2FA.

How could something like this not being resolved in time? Gmail is sadly used by lots of NC users - I am quite sure.

ChristophWurst:
My personal account still works.

feutl:
@ChristophWurst 
Have you setup 2FA with an app specific password ? If so, yes it works.
If not, I am surprised, all 3 of my accounts without 2FA do not work any more.

ChristophWurst:
Right, I'm using 2FA with an app password.

feutl:
And this is the issue, if you have not setup 2FA (for whatever reason) there is no app password option in gmail. Therefor the authentication needs to be fixed for those accounts.
As I said, Fairemal - the android client - has fixed this already some time ago.

MrPresident2:
any news about that? I don't think we can use Gmail anymore

enekonieto:
Until someone is assigned I think we should hope no movement.

MrPresident2:
do you know about another way to connect Gmail to Nextcloud then?

enekonieto:
No, I am also stuck with this issue :(

ChristophWurst:
I causes a generic *Invalid credentials*. So we need to keep book about the token validity and do the refresh proactively.

ChristophWurst:
POC is at https://github.com/nextcloud/mail/pull/6830. Linking the Nextcloud Mail account to Gmail works. Keeping the access token updated works.

The open todos are mostly about handling all possible conditions during the setup and making sure the app stays usable with this new auth option.

ChristophWurst:
https://github.com/nextcloud/mail/pull/6830#issuecomment-1178964939 sneak preview

feutl:
[...]

PR #7421, opened 2022-10-13
Add account column/props for XOAUTH2 refresh tokens and their TTL

ChristophWurst:
Required for https://github.com/nextcloud/mail/issues/6454 and #6591.

The logic to fetch and refresh the tokens is vendor specific and will be added in separate PRs.


miaulalala (approved):
LGTM

PR #7298, opened 2022-09-21
Add ckeditor5-html-embed plugin in TextEditor.vue

Zupe007:
I added the ckeditor5-html-embed plugin, so that it is possible to use it in the signature editor.
It enables the use of html signatures.
I made this change, because working with  tables #7282  to create a copy of our company signature doesnt work.

PR #7427, opened 2022-10-14
Add ckeditor5-html-embed plugin

ChristophWurst:
Rebased version of https://github.com/nextcloud/mail/pull/7298.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "@ckeditor/ckeditor5-essentials": "~35.1.0",
         "@ckeditor/ckeditor5-font": "~35.1.0",
         "@ckeditor/ckeditor5-heading": "~35.1.0",
+        "@ckeditor/ckeditor5-html-embed": "~35.1.0",
         "@ckeditor/ckeditor5-image": "^35.1.0",
         "@ckeditor/ckeditor5-link": "~35.1.0",
         "@ckeditor/ckeditor5-list": "~35.1.0",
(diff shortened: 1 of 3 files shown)


ChristophWurst:
How it looks

Bildschirmfoto vom 2022-10-14 11-59-26

Zupe007:
@ChristophWurst 
I dont know why it is so small in the beginning. It was the same for me.
I fixed it temporarily by using the Nextcloud CSS APP with

`#select, button:not(.button-vue), input, textarea, div[contenteditable="true"], div[contenteditable="false"]{
        width: unset;
}`

not it looks like this for me.
Signatur



I am not skilled enough to fix the standard behavior.

ChristophWurst:
I will have to discuss this with our designers then. I was under the impression that the HTML plugin would allow the users to edit the sources of the full message. Instead, it allows users to insert a block of HTML.

Zupe007:
Yes it inserts a block. For my company that is exactly the feature we need and it is working quite good. 
For Example:

Creating a signature.
Signatur anlegen


Writing an email, signature is added and shown perfectly.
Signatur in Mail

Zupe007:
@ChristophWurst Hey, is there any chance my Pull request will be or maybe is accepted?

technikfrek:
Is there any progress?
Me and my company are also interested in this feature.

Thanks in advance!

ChristophWurst:
Closing for now. I need designers to sign off this change. It still doesn't look like the right way to do this, sorry.

Ref https://github.com/nextcloud/mail/pull/7427#issuecomment-1283598965

PR #7448, opened 2022-10-18
Reroute to priority mailbox when a mailbox is deleted

### LINKED ISSUE #7406: Deleting a folder / mailbox clears the entire right content area including background
### Steps to reproduce

1. Delete a mailbox / folder via the left navigation
2. Observe the interface break

### Expected behavior

When deleting a folder / deleting a mailbox, Mail could automatically jump to open the Inbox of the same account? That would be similar to how we do it for deleting mails.

### Actual behavior

The content area gets wiped including background container:
image

### Mail app version

Current on cloud.nc.com

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

GretaD:
fixes #7406

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -223,7 +223,7 @@ import EraserVariant from 'vue-material-design-icons/EraserVariant'
 import ImportantIcon from './icons/ImportantIcon'
 import IconSend from 'vue-material-design-icons/Send'
 import MoveMailboxModal from './MoveMailboxModal'
-import { UNIFIED_INBOX_ID } from '../store/constants'
+import { PRIORITY_INBOX_ID, UNIFIED_INBOX_ID } from '../store/constants'
 
 import { clearCache } from '../service/MessageService'
 import { getMailboxStatus } from '../service/MailboxService'
@@ -603,6 +603,14 @@ export default {
 							.dispatch('deleteMailbox', { mailbox: this.mailbox })
 							.then(() => {
 								logger.info(`mailbox ${id} deleted`)
+								if (parseInt(this.$route.params.mailboxId, 10) === this.mailbox.databaseId) {
+									this.$router.push({
+										name: 'mailbox',
+										params: {
+											mailboxId: PRIORITY_INBOX_ID,
+										},
+									})
+								}
 							})
 							.catch((error) => logger.error('could not delete mailbox', { error }))
 					}


src/components/NavigationMailbox.vue
```diff
@@ -579,6 +579,12 @@ export default {
 							.dispatch('deleteMailbox', { mailbox: this.mailbox })
 							.then(() => {
 								logger.info(`mailbox ${id} deleted`)
+								this.$router.push({
```
ChristophWurst, line 582: the reroute should only happen if the *current mailbox* is deleted

ChristophWurst (changes requested):

JuliaKirschenheuter (approved):
looks good, but haven't tested

src/components/NavigationMailbox.vue
```diff
@@ -579,6 +579,14 @@ export default {
 							.dispatch('deleteMailbox', { mailbox: this.mailbox })
 							.then(() => {
 								logger.info(`mailbox ${id} deleted`)
+								if (this.$route.params.mailboxId) {
```
ChristophWurst, line 582:
```suggestion
								if (this.$route.params.mailboxId == this.mailbox.databaseId) {
```

ChristophWurst (changes requested):

src/components/NavigationMailbox.vue
```diff
@@ -579,6 +579,14 @@ export default {
 							.dispatch('deleteMailbox', { mailbox: this.mailbox })
 							.then(() => {
 								logger.info(`mailbox ${id} deleted`)
+								if (this.$route.params.mailboxId === this.mailbox.databaseId) {
```
ChristophWurst, line 582: This can never be truthy. `this.$route.params.mailboxId` is a string, `this.mailbox.databaseId` is a Number.

ChristophWurst (changes requested):

src/components/NavigationMailbox.vue
ChristophWurst: It is an int before the page load. After a page load it becomes a string.

ChristophWurst:
@GretaD give this a final test but I think this is mergeable now

ChristophWurst (approved):

PR #7472, opened 2022-10-20
Add skeleton loading thread messages

GretaD:
contributes to #6045 

- [x] Add skeleton for message body
- [x] When there is a thread with more than one msg - one message container is shown(see Jans mockup below)  then the other threads show up.
- [x] Animate the whole container
- [x] Dont animate the text

We agreed with Jan that the skeleton lines to be square since we have it like that everywhere, but in future they should be round/pill style.

[Screencast from 17.11.2022 18:27:21.webm](https://user-images.githubusercontent.com/12728974/202516961-a508e332-5d1d-41a8-bcd9-f8983c90868b.webm)

### DIFF
--- a/src/components/MessageLoadingSkeleton.vue
+++ b/src/components/MessageLoadingSkeleton.vue
@@ -1,0 +1,98 @@
+<template>
+	<div class="message-loading-skeleton">
+		<div class="message-loading-skeleton__body">
+			<div class="message-loading-skeleton__body__line-one">
+					&nbsp;
+			</div>
+			<div class="message-loading-skeleton__body__line-two">
+					&nbsp;
+			</div>
+			<div class="message-loading-skeleton__body__line-three">
+					&nbsp;
+			</div>
+			<div class="message-loading-skeleton__body__line-four">
+					&nbsp;
+			</div>
+			<div class="message-loading-skeleton__body__line-five">
+					&nbsp;
+			</div>
+			<div class="message-loading-skeleton__body__line-six">
+					&nbsp;
+			</div>
+		</div>
+	</div>
+</template>
+
+<script>
+
+export default {
+	name: 'MessageLoadingSkeleton',
+}
+</script>
+
+<style lang="scss" scoped>
+
+/* skeleton */
+.message-loading-skeleton {
+	display: flex;
+	flex-grow: 1;
+	align-items: flex-start;
+	padding: 8px;
+	width: 100%;
+	&__body {
+		max-height: 500px;
+		max-width: 1200px;
+		flex: 1 1 100%;
+		display: flex;
+		flex-direction: column;
+		margin-left: 50px;
+		&__line-one,
+		&__line-two,
+		&__line-three,
+		&__line-four,
+		&__line-five,
+		&__line-six {
+			white-space: nowrap;
+			height: 13px;
+			background: linear-gradient(-45deg, #d3d3d3, #f5f5f5, #f2f3f4, #e5e4e2);
+			background-size: 400% 400%;
+			animation: gradient 3s ease-in infinite;
+		}
+		&__line-one {
+			width: 10%;
+			margin-bottom: 25px;
+		}
+		&__line-two {
+			width: 80%;
+			margin-bottom: 6px;
+		}
+		&__line-three {
+			width: 50%;
+			margin-bottom: 6px;
+		}
+		&__line-four {
+			width: 70%;
+			margin-bottom: 6px;
+		}
+		&__line-five {
+			width: 40%;
+			margin-bottom: 25px;
+		}
+		&__line-six {
+			width: 30%;
+		}
+	}
+	@keyframes gradient {
+		 0% {
+			 background-position: 0% 50%;
+		 }
+		 50% {
+			 background-position: 100% 50%;
+		 }
+		 100% {
+			 background-position: 0% 50%;
+		 }
+	 }
+}
+
+</style>
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -34,10 +34,7 @@
 					</div>
 				</div>
 			</div>
-			<!-- Show inner loading screen only if we have at least one message of the thread -->
-			<Loading v-if="loading && thread.length" :hint="t('mail', 'Loading messages')" />
 			<ThreadEnvelope v-for="env in thread"
-				v-else
 				:key="env.databaseId"
 				:envelope="env"
 				:mailbox-id="$route.params.mailboxId"
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -143,7 +143,7 @@
 				</template>
 			</div>
 		</div>
-		<Loading v-if="loading !== LOADING_DONE" />
+		<MessageLoadingSkeleton v-if="loading !== LOADING_DONE" />
 		<Message v-if="message && loading !== LOADING_MESSAGE"
 			v-show="loading === LOADING_DONE"
 			:envelope="envelope"
@@ -164,7 +164,7 @@ import Error from './Error'
 import importantSvg from '../../img/important.svg'
 import IconFavorite from 'vue-material-design-icons/Star'
 import JunkIcon from './icons/JunkIcon'
-import Loading from './Loading'
+import MessageLoadingSkeleton from './MessageLoadingSkeleton'
 import logger from '../logger'
 import Message from './Message'
 import MenuEnvelope from './MenuEnvelope'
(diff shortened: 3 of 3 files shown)


src/components/Thread.vue
```diff
@@ -236,17 +236,17 @@ export default {
 				if (thread.length === 0) {
 					logger.info('thread could not be found and is empty', { threadId })
 					this.errorMessage = getRandomMessageErrorMessage()
-					this.loading = false
+					this.loading = true
```
GretaD, line 239: for testing reasons

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+	<div>
+		<div v-for="i in numberOfLines" :key="i" class="item-list__entry">
```
ChristophWurst, line 3: this is not an item list, is it?

ChristophWurst:
This does what it advertises but IMO it just emphasizes https://github.com/nextcloud/mail/issues/6045 more because loading the thread info is relatively quick (database query), loading the message is slow (IMAP roundtrip)

src/components/LoadingSkeletonThread.vue
GretaD: yes it is a list

GretaD:
> This does what it advertises but IMO it just emphasizes #6045 more because loading the thread info is relatively quick (database query), loading the message is slow (IMAP roundtrip)

that is true. So should we fix the loading part first for the IMAP part to make it to load faster then we merge this one?

src/components/LoadingSkeletonThread.vue
ChristophWurst: there is exactly one "item"

GretaD: ah sorry, we are in threads, no its not a list in threads.

GretaD:
the quality of screen recorder is bad but you get a bit the idea of gradient and how it looks like

[Screencast from 16.11.2022 17:23:27.webm](https://user-images.githubusercontent.com/12728974/202236882-e91e4d3e-e0de-46bb-b7f3-374cf2274f8b.webm)

jancborchardt (changes requested):
Nice! :) Here is a mockup of how it should look like (animation is missing of course as it is a screenshot):
- [x] We can show the container around it because we know it will be there
- [ ] We can show the avatar because we already have it client-side (we show it in the message list as well as directly above the container already)
- [ ] We can show the sender name for the same reason
- [ ] I didnt even include the first line of text which we also have client-side already – but as far as I understand we don’t have line-breaks or anything in there so maybe best to not use it?
- [ ] The skeleton lines should be for the mail content. And for that it looks nicer to have several lines instead of 1 big block
Mail skeleton screen

GretaD:
> Nice! :) Here is a mockup of how it should look like (animation is missing of course as it is a screenshot):
> 
> * [ ]  We can show the container around it because we know it will be there
> * [ ]  We can show the avatar because we already have it client-side (we show it in the message list as well as directly above the container already)
> * [ ]  We can show the sender name for the same reason
> * [ ]  I didnt even include the first line of text which we also have client-side already – but as far as I understand we don’t have line-breaks or anything in there so maybe best to not use it?
> * [ ]  The skeleton lines should be for the mail content. And for that it looks nicer to have several lines instead of 1 big block
>   Mail skeleton screen

ok, this was supposed to be the message skeleton so that is how the loading state of message should look like. And thats helpful for the second pr.

Current PR here is suppose to show the loading state of the thread before the message is loaded.  So this pr's skeleton is shown when i click the envelope and the thread start loading -   then once the message start loading, your mock up shows up.

Let me know if you want only one skeleton(your mockup), but during our call yesterday you mentioned 2 phases, thats why i splited  it

GretaD:
> > Nice! :) Here is a mockup of how it should look like (animation is missing of course as it is a screenshot):
> > 
> > * [ ]   We can show the container around it because we know it will be there
> > * [ ]   We can show the avatar because we already have it client-side (we show it in the message list as well as directly above the container already)
> > * [ ]   We can show the sender name for the same reason
> > * [ ]   I didnt even include the first line of text which we also have client-side already – but as far as I understand we don’t have line-breaks or anything in there so maybe best to not use it?
> > * [ ]   The skeleton lines should be for the mail content. And for that it looks nicer to have several lines instead of 1 big block
> 
> ok, this is supposed to be the message skeleton so that is how the loading state of message should look like. And thats helpful for the second pr.
> 
> Current PR here is suppose to show the loading state of the thread. So this pr's skeleton is shown when i click the envelope and the thread start loading - then once the message start loading, your mock up shows up.
> 
> Let me know if you want only one skeleton(your mockup), but during our call yesterday you mentioned 2 phases, thats why i split it

we had a call and agreed on the points on the description of this ticket

jancborchardt (approved):
Very nice! :)

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+	<div>
+		<div class="item-list__entry">
+			<div class="item__details">
+				<div class="line-one">
+&nbsp;
```
ChristophWurst, line 6: indentation

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+	<div>
+		<div class="item-list__entry">
```
ChristophWurst, line 3:
```suggestion
		<div class="message-loading-skeleton">
```

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,100 @@
...
+			<div class="item__details">
```
ChristophWurst, line 4:
```suggestion
			<div class="message-loading-sekeleton__body">
```

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,100 @@
...
+				<div class="line-one">
+&nbsp;
+				</div>
+				<div class="line-two">
```
ChristophWurst, line 8:
```suggestion
				<div class="message-loading-sekeleton__body__line-two">
```

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+	<div>
```
ChristophWurst, line 2: outermost div is not necessary, is it?

src/components/LoadingSkeletonThread.vue
```diff
@@ -0,0 +1,100 @@
...
+				</div>
+			</div>
+		</div>
+	</div>
+</template>
+
+<script>
+
+export default {
+	name: 'LoadingSkeletonThread',
```
ChristophWurst, line 31:
```suggestion
	name: 'MessageLoadingSkeleton',
```

ChristophWurst (changes requested):
Naming & BEM

src/components/MessageLoadingSkeleton.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+	<div class="message-loading-skeleton">
+		<div class="message-loading-skeleton__body">
+			<div class="message-loading-sekeleton__body__line-one">
+&nbsp;
```
ChristophWurst, line 5: indentation

src/components/MessageLoadingSkeleton.vue
```diff
@@ -0,0 +1,98 @@
...
+<style lang="scss" scoped>
+
+/* skeleton */
+.message-loading-skeleton {
+	display: flex;
+	flex-grow: 1;
+	align-items: flex-start;
+	padding: 8px;
+	width: 100%;
+	.message-loading-skeleton__body {
```
ChristophWurst, line 42:
```suggestion
	&__body {
```

src/components/MessageLoadingSkeleton.vue
```diff
@@ -0,0 +1,98 @@
...
+		max-height: 500px;
+		max-width: 1200px;
+		flex: 1 1 100%;
+		display: flex;
+		flex-direction: column;
+		margin-left: 50px;
+		.message-loading-sekeleton__body__line-one,
```
ChristophWurst, line 49:
```suggestion
		&__line-one,
```

ChristophWurst (changes requested):
The power of BEM

ChristophWurst (commented):
Looks and works great otherwise!

ChristophWurst (approved):

PR #7430, opened 2022-10-14
Add Google OAuth support

ChristophWurst:
Clean version of https://github.com/nextcloud/mail/pull/6830. Reviewable commit by commit.

Fixes https://github.com/nextcloud/mail/issues/6454

## To do

- [x] Google Oauth admin settings
- [x] Account setup flow adaptation
  - [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/3369
- [x] Token refresh logic
- [x] Account re-auth for expired/revoke refresh tokens
  - [x] https://github.com/nextcloud/mail/pull/7476
- [x] (App) password to oauth migration
- [x] Edge case: email mismatch between Mail form and Google consent flow. ~~Fetch email from Gmail so the token definitely works.~~ Consent windows allows us to "hint" the account email address
- [x] Documentation

### Scenario one: fresh ~~avocado~~ account
1) Follow the admin docs to obtain and save a client ID and secret
2) Open the app
3) Enter your gmail address
  * See the form submit button change to *Sign in with Google*
4) Follow the instructions on the screen

### Scenario two: account migration
0) Unlink gmail oauth if you set a client ID/secret already
1) Enable 2FA for your gmail account, create an app password
2) Open the app
3) Add the account using the app password - the form submit button should say *Connect*
4) Follow the admin docs to obtain and save a client ID and secret
5) Open the app - account should still sync and open mail
6) Open account settings, scroll to server settings - the form submit button should say 'Reconnect Google account*
7) Reconnect the account - it now uses oauth instead of the app password

## Screenshots

| Screen | Shot |
|---|---|
| Admin settings | Bildschirmfoto vom 2022-10-14 13-42-14 |
| Before setup | Bildschirmfoto vom 2022-10-14 17-03-37 |
| During setup | Bildschirmfoto vom 2022-10-14 17-10-37 |

PR #7476, opened 2022-10-21
Fix handling of account update API response

ChristophWurst:
Axios gives us the response contents inside `data`, but our response also has a key `data`. So we need `.data.data`.

Found in #7430


miaulalala (approved):
image

PR #7485, opened 2022-10-24
Fix OOM during cache clean-up

ChristophWurst:
array_map is too memory hungry for large accounts

## How to test

1) Add an account with 600k messages
2) Run `occ mail:clean-up`

Master: :boom: 
Here: :sunglasses:


kesselb:
Good finding :+1:

kesselb (approved):

PR #7486, opened 2022-10-24
Fix transfer encoding issue

### LINKED ISSUE #7408: Decoding mangled for some first message lines in preview
### Steps to reproduce

1.  Go to Mail app
2.  List messages (either on inbox or navigating to folder)
3.  See mangled preview

### Expected behavior

Correct preview decoding

### Actual behavior

Mangled preview / unintelligible on message list 

### Mail app version

1.14.0

### Mailserver or service

Dovecot

### Operating system

Debian 11.0

### PHP engine version

PHP 7.4

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

It is a particular issue as this happens only for some specific messages from various senders (there are senders where preview is shown correctly), also, this was not happening some Mail versions ago.

Attached is a message source summary with the first 3 _actual_ message lines (all sensitive information has been redacted sans subject)

Also, a short capture on how the message preview on the message list is seen.

sshot_nexcloud_mail
[message_summary.txt](https://github.com/nextcloud/mail/files/9756094/message_summary.txt)

miaulalala:
Fixes #7408

### DIFF
--- a/lib/Migration/Version1140Date20221027171138.php
+++ b/lib/Migration/Version1140Date20221027171138.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2022 Anna Larch <anna.larch@gmx.net>
+ *
+ * @author Anna Larch <anna.larch@gmx.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Mail\Migration;
+
+use Closure;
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):
If we reset `structure_analyzed` it will also fix existing data with wrong encoding

miaulalala:
> If we reset `structure_analyzed` it will also fix existing data with wrong encoding

Repair step?

miaulalala:
@ChristophWurst how would we handle this for backports? Is this an issue in Mail 2 only?

ChristophWurst:
The feature exists since 1.14. You can take inspiration at https://github.com/nextcloud/mail/blob/main/lib/Migration/Version1140Date20220808203258.php#L44-L55 for resetting the flag.

Target branch updated because 2.1 has already been branched-off. Main will be 2.2.0

lib/Migration/Version1140Date20221027171138.php
```diff
@@ -0,0 +1,59 @@
...
+	public function __construct(MessageMapper $messageMapper,
+		LoggerInterface $logger) {
+		$this->logger = $logger;
+		$this->messageMapper = $messageMapper;
+	}
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+g	 * @param array $options
```
ChristophWurst, line 50: g

lib/Migration/Version1140Date20221027171138.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2022 Your name <your@email.com>
```
ChristophWurst, line 6: :eyes:

lib/Migration/Version1140Date20221027171138.php
miaulalala: gg

ChristophWurst (approved):

Issue #6887, opened 2022-07-15
Memory exhausted during thread building with large mailbox


ChristophWurst:
### Steps to reproduce

1. Have a mailbox with 600k messages
2. Synchronize the account from the CLI

### Expected behavior

Sync and threading pass

### Actual behavior

Sync passes, threading :boom: 

```
[debug] account is up to date, skipping mailbox sync
[debug] Syncing 2490
[debug] Locking mailbox 2490 for new messages sync
[debug] Locking mailbox 2490 for changed messages sync
[debug] Locking mailbox 2490 for vanished messages sync
[debug] Running partial sync for 2490
[debug] partial sync 3151:INBOX - get all known UIDs took 1s. 44/317MB memory used
[debug] partial sync 3151:INBOX - get new messages via Horde took 3s. 44/317MB memory used
[debug] partial sync 3151:INBOX - persist new messages took 0s. 44/317MB memory used
[debug] partial sync 3151:INBOX - get changed messages via Horde took 0s. 44/317MB memory used
[debug] partial sync 3151:INBOX - persist changed messages took 0s. 44/317MB memory used
[debug] partial sync 3151:INBOX - get vanished messages via Horde took 0s. 45/317MB memory used
[debug] partial sync 3151:INBOX - delete vanished messages took 0s. 45/317MB memory used
[debug] partial sync 3151:INBOX took 4s
[debug] Unlocking mailbox 2490 from vanished messages sync
[debug] Unlocking mailbox 2490 from changed messages sync
[debug] Unlocking mailbox 2490 from new messages sync
[debug] Account 3151 has 619683 messages for threading
[debug] Threading 619683 messages - build ID table took 3s. 427/445MB memory used
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20971520 bytes) in apps/mail/lib/IMAP/Threading/Container.php on line 102
``` 

### Mail app version

1.13, 1.14 and main

PR #7494, opened 2022-10-25
Fix high memory usage during thread building

ChristophWurst:
## Problem

Threading is done across all messages of one account. With the number of messages the amount of input data for the threading algorithm increases. As some point, the available memory size of a PHP process is reached.

A testing account with 600k dummy emails could be sync'ed into the database case, but at threading the process always ran out of memory.

## Solution

If only messages with threading info are loaded we can significantly reduce the amount of memory consumed to build threads. This is because many messages are not part of a thread. And if they are, the references and in-reply-to headers are enough information to know that there was a previous message. So we build the threads out of existence of the thread roots. As long as the threading result is the same, this doesn't matter.

### Example 1: we have message X and Y. Y points to X as parent message.

Previously we loaded them both. Then found out the are related and save X's ID as *thread root id*.
Now we only load Y, because X doesn't point to any other message. We still know that Y is part of a thread with X's id. So we save X's id as Y's *thread root id*. X is not changed, but keeps its default self-reference for the thread root. Both messages therefore belong to X's id.

### Example 2: we have message Y that points to unknown message X

Previously and now we only load message Y. But previously we discarded the empty tree node for X by replacing it with Y. Therefore Y referenced itself as thread root. If we ever found X, the algorithm would stop replacing the now not empty node and Y's thread root updates to X. The thread is built explicitly.
Now we already save Y's id as X's thread root. We don't care if the parent exists or not. If Y appears, it doesn't have any influence on the threading algorithm. The thread is therefore built implicitly because X and Y have the same thread root ids.

### Example 3: we have a message that points to no other

Previously this message was loaded for threading. If no other message pointed to it it was a useless node in the tree that did not get any database update.
Now we don't even load the message. It doesn't have any effect on the algorithm output.

## Todo

- [x] Make it work
- [x] Make it work *exactly* like the previous algorithm. ~~Right now there are very few messages/threads where the new algorithm causes different results than the old.~~ Different thread root IDs because we keep references of (yet) unknown messages. But the resulting threads are identical.
- [ ] What about subject threading?!

Fixes https://github.com/nextcloud/mail/issues/6887

---

> There are two types of people in the world: Those who can extrapolate from incomplete data.

Unknown.

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -204,7 +204,13 @@ class MessageMapper extends QBMapper {
 		$messagesQuery->select('id', 'subject', 'message_id', 'in_reply_to', 'references', 'thread_root_id')
 			->from($this->getTableName())
 			->where($messagesQuery->expr()->in('mailbox_id', $messagesQuery->createFunction($mailboxesQuery->getSQL()), IQueryBuilder::PARAM_INT_ARRAY))
-			->andWhere($messagesQuery->expr()->isNotNull('message_id'));
+			->andWhere(
+				$messagesQuery->expr()->isNotNull('message_id'),
+				$messagesQuery->expr()->orX(
+					$messagesQuery->expr()->isNotNull('in_reply_to'),
+					$messagesQuery->expr()->neq('references', $messagesQuery->createNamedParameter('[]'))
+				),
+			);
 
 		$result = $messagesQuery->execute();
 		$messages = [];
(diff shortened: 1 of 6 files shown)


ChristophWurst:
Could someone review/test this please? :)

ChristophWurst:
I'll look into disabling subject threading in a separate PR.

miaulalala:
Tested it - if the "sent" mailbox is loaded, the threading super speedy (almost instant).

I tested by sending a few messages back and forth between two accounts, and then removing and adding the mail account in the Mail app.

What surprised me was the following scenario:

When I artificially produce an initial sync by removing the account and adding it again, I have a relatively "new" mail account from a DB perspective. In this case, I haven't yet synced all mailboxes (for example, by clicking on them in the UI). In this case, the messages are not threaded at all in the UI. Is this expected?

miaulalala (approved):
image

PR #7512, opened 2022-11-02
Migrate Vuex actions from promises to async-await

ChristophWurst:
This will make it easier to add consistent try-catch blocks for https://github.com/nextcloud/mail/issues/5714.


GretaD (approved):
:star2:

Issue #4768, opened 2021-03-15
Rework draft handling (on IMAP)


ChristophWurst:
## Current state

Drafts do not work so well right now. Due to IMAP restrictions we recreate the messages for every byte that changes in the editor. Sometimes old version remain, other times you send the final message and there is a large backlog of drafts to be stored before the message is going out. As recently discussed we have to improve this.

We need to do some research about how other web mail clients handle this. Do they also store to IMAP directly? Could we save to the db first and only when the user *exits* the current message to to IMAP? …

## Implementation

This section aims to describe the high level mechanics of the improved drafts handling with working copies.

### Principles

1) Outside the boundaries of a transaction no more than one version of a draft exists. The draft is either on IMAP or in the local database.
2) The drafts handling is compatible with other IMAP clients.

### Use Case 1 - composing a new message that is sent right away

1) **The user opens the Mail app and clicks *New message***
2) **As the user types, pauses, continues, etc, the app periodically saves *working copies***-> T1+T3
3) **The user clicks *Send***: the frontend saves the last editor state as another draft revision. Then the frontend sends a request to the backend to convert the local draft to a message in the local outbox. -> T5
4) **The message is sent by a trigger from the frontend or a background job in the backend like it was done before**

### Use Case 2 - composing a new message without sending it
1) **The user opens the Mail app and clicks *New message***
2) **As the user types, pauses, continues, etc, the app periodically saves *working copies*** -> T1+T3
3) **The user closes the composer modal**: the frontend saves the last editor state as another draft revision. Then the frontend sends a request to the backend to move the local draft to a draft message on IMAP. The move operation deletes the local draft from the database. -> T6

### Use Case 3 - composing a new message but closing the browser
1) **The user opens the Mail app and clicks *New message***
2) **As the user types, pauses, continues, etc, the app periodically saves *working copies*** -> T1+T3
3) **The user closes the browser**
4) **A background job in the backend finds the stale local draft**: the backend moves the local draft to a draft message on IMAP. The move operation deletes the local draft from the database. A threshold between current time and the draft last revision timestamp is the indicator for the staleness of a local draft. -> T6

### Use Case 4 - composing a new message but discarding it
1) **The user opens the Mail app and clicks *New message***
2) **As the user types, pauses, continues, etc, the app periodically saves *working copies*** -> T1+T3
3) **The user clicks the draft delete icon**: the frontend sends a request to the backend to delete the local draft. -> T4

### Use Case 5 - open a draft but make no edits
[...]

PR #7517, opened 2022-11-03
Rework local draft handling (Backend)

### LINKED ISSUE #7076: Rework Draft Handling - Backend Part

miaulalala:
Fixes #7076

Thoroughly documented here: https://github.com/nextcloud/mail/issues/4768
To do:
### Transaction 1 - create local draft for new message

Frontend sends draft data to backend
- [x] Backend stores draft data in database
- [x] Backend returns draft data including its database id

### Transaction 2 - create local draft from IMAP draft

Frontend sends draft data and/or draft message ID to backend. The id is the id of the database cache, not the IMAP UID
- [x] Backend stores draft data in database
- [x] Backend deletes draft message from IMAP
- [x] Backend returns draft data including its database id

### Transaction 3 - create revision of local draft

Frontend sends draft id and data to the backend
- [x] Backend replaces data of existing record in database.
- [x] Backend returns the updated draft including its unchanged database id

### Transaction 4 - delete local draft

Frontend sends draft id to the backend
- [x] Backend deletes draft from database

### Transaction 5 - send local draft

Frontend sends draft id to the backend
- [x] Backend converts local draft to local outbox message

### Transaction 6 - create IMAP draft from local draft

Frontend sends draft id to the backend
- [x] Backend creates an IMAP message from the draft data in the user's drafts mailbox
- [x] Backend deletes local draft from database

### Tests

- [x] DraftsController Unit Tests
- [x] DraftsService Unit Tests
- [x] DraftsService Integration Tests
- [x] ~~LocalMessageMapper Integration Test for new Type~~

### Background job

- [x] sync to IMAP when draft has been stale for a certain amount of time (TBD)

### DIFF
--- a/lib/BackgroundJob/IMipMessageJob.php
+++ b/lib/BackgroundJob/IMipMessageJob.php
@@ -33,12 +33,12 @@ class IMipMessageJob extends TimedJob {
 	private IMipService $iMipService;
 
 	public function __construct(ITimeFactory $time,
-								IMipService $iMipService) {
+								IMipService $draftsService) {
 		parent::__construct($time);
 
 		// Run once per hour
 		$this->setInterval(60 * 60);
-		$this->iMipService = $iMipService;
+		$this->iMipService = $draftsService;
 	}
 
 	protected function run($argument): void {
--- a/lib/Controller/DraftsController.php
+++ b/lib/Controller/DraftsController.php
@@ -1,0 +1,198 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Mail App
+ *
+ * @copyright 2022 Anna Larch <anna.larch@gmx.net>
+ *
+ * @author Anna Larch <anna.larch@gmx.net>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Mail\Controller;
+
+use OCA\Mail\Db\LocalMessage;
+use OCA\Mail\Http\JsonResponse;
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\DraftsService;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\IRequest;
+
+class DraftsController extends Controller {
+	private DraftsService $service;
+	private string $userId;
+	private AccountService $accountService;
+	private ITimeFactory $timeFactory;
+
+	public function __construct(string $appName,
+								$UserId,
+								IRequest $request,
+								DraftsService $service,
+								AccountService $accountService,
+								ITimeFactory $timeFactory) {
+		parent::__construct($appName, $request);
+		$this->userId = $UserId;
+		$this->service = $service;
+		$this->accountService = $accountService;
+		$this->timeFactory = $timeFactory;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @TrapError
+	 *
+	 * @param int $accountId
+	 * @param string $subject
+	 * @param string $body
+	 * @param string $editorBody
+	 * @param bool $isHtml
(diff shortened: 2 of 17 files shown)


lib/Controller/DraftsController.php
```diff
@@ -0,0 +1,154 @@
...
+		string  $editorBody,
+		bool    $isHtml,
+		bool    $failed = false,
+		array   $to = [],
+		array   $cc = [],
+		array   $bcc = [],
+		array   $attachments = [],
+		?int    $aliasId = null,
+		?string $inReplyToMessageId = null,
+		?int $sendAt = null): JsonResponse {
```
ChristophWurst, line 136: I don't think this should ever not be not for a draft message, should it? :thinking:

ChristophWurst: Unless the update method also handles conversion to an outbox message message

miaulalala: I think it would be good to have that in the update. Use case creating and continuosly updating a draft and then sending it comes to mind.

ChristophWurst: So it is the conversion aspect

miaulalala: Current logic: if there's a sendAt timestamp, the message will be converted to an outbox message.

lib/BackgroundJob/IMipMessageJob.php
```diff
@@ -33,12 +33,12 @@ class IMipMessageJob extends TimedJob {
 	private IMipService $iMipService;
 
 	public function __construct(ITimeFactory $time,
-								IMipService $iMipService) {
+								IMipService $draftsService) {
```
ChristophWurst, line 36: :monocle_face:

lib/Migration/Version2200Date20221103140538.php
```diff
@@ -0,0 +1,57 @@
...
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$localMessagesTable = $schema->getTable('mail_local_messages');
+		$localMessagesTable->addColumn('updated_at', Types::INTEGER, [
```
ChristophWurst, line 47: check for column existence so the migration can be run more than once in the case of an error

ChristophWurst (commented):
One more thing: bump the version to alpha.1 so a migration check is triggered

lib/Controller/DraftsController.php
miaulalala: Am I ok to resolve?

ChristophWurst (approved):
Looks good

lib/Migration/Version2020Date20221103140538.php
```diff
@@ -0,0 +1,58 @@
...
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version2020Date20221103140538 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
```
ChristophWurst, line 38: try with psalm-param

the parent class has the broader type Closure https://github.com/nextcloud-deps/ocp/blob/d1f996e8d03a10184d48a0120282886b3f7806f7/OCP/Migration/IMigrationStep.php#L63

miaulalala: so `@psalm-param Closure()` ?

ChristophWurst: Try it

miaulalala: Works

lib/BackgroundJob/IMipMessageJob.php
```diff
@@ -33,12 +33,12 @@ class IMipMessageJob extends TimedJob {
...
 		parent::__construct($time);
 
 		// Run once per hour
 		$this->setInterval(60 * 60);
-		$this->iMipService = $iMipService;
+		$this->iMipService = $draftsService;
```
kesselb, line 41: :confused:

ChristophWurst:
> Transaction 6 - create IMAP draft from local draft
>    Frotnend sends draft id to the backend
>    Backend creates an IMAP message from the draft data in the user's drafts mailbox
>    Backend deletes local draft from database

Where can I find that?

lib/Service/DraftsService.php
```diff
@@ -0,0 +1,240 @@
...
+					// Ignore message of non-existent account
+					continue;
+				}
+				$this->sendMessage(
+					$message,
+					$account,
+				);
+				$this->logger->debug('Draft {id} moved to IMAP', [
+					'id' => $message->getId(),
+				]);
```
ChristophWurst, line 229: this doesn't seem right. the message is moved as part of the sending process. but it shouldn't be sent

ChristophWurst: It actually works because we pass a special flag to *send message* which prevents it from sending the message.

PR #7519, opened 2022-11-03
Add icon for load more on PI

### LINKED ISSUE #6142: Improve look of priority inbox

GretaD:
before
Screenshot from 2022-11-03 17-25-23
after

Screenshot from 2022-11-03 17-23-52

ref and contributes to #6142

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -196,6 +196,7 @@
 				:key="'list-collapse-' + searchQuery"
 				class="load-more"
 				@click="$emit('load-more')">
+				<AddIcon :size="20" />
 				{{ t('mail', 'Load more') }}
 			</div>
 			<div id="load-more-mail-messages" key="loadingMore" :class="{'icon-loading-small': loadingMore}" />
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

sazanof (approved):
Tested and works!

PR #7523, opened 2022-11-04
Prevent trim'ing and strpos'ing null in integration tests

ChristophWurst:
Fixes a flock of 
```
PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in /home/christoph/workspace/nextcloud/apps/mail/vendor/bytestream/horde-mail/lib/Horde/Mail/Rfc822.php on line 224
```
and
```
PHP Deprecated:  strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /home/christoph/workspace/nextcloud/apps/mail/vendor/bytestream/horde-mime/lib/Horde/Mime.php on line 217
``` 

that pollute the text output.

### DIFF
--- a/tests/Integration/Framework/MessageBuilder.php
+++ b/tests/Integration/Framework/MessageBuilder.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  *
@@ -28,13 +30,13 @@ class MessageBuilder {
 	/** @var string */
 	private $to;
 
-	/** @var string */
+	/** @var string|null */
 	private $cc;
 
-	/** @var string */
+	/** @var string|null */
 	private $bcc;
 
-	/** @var string */
+	/** @var string|null */
 	private $date;
 
 	/** @var string */
@@ -54,7 +56,7 @@ class MessageBuilder {
 	 * @param string $from
 	 * @return MessageBuilder
 	 */
-	public function from($from) {
+	public function from(string $from) {
 		$this->from = $from;
 		return $this;
 	}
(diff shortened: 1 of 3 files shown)


tests/Integration/Framework/MessageBuilder.php
```diff
@@ -72,7 +74,7 @@ public function to($to) {
 	 * @param string $cc
 	 * @return MessageBuilder
 	 */
-	public function cc($cc) {
+	public function cc(string $cc) {
```
miaulalala, line 77:
```suggestion
	public function cc(?string $cc) {
```

tests/Integration/Framework/MessageBuilder.php
```diff
@@ -81,7 +83,7 @@ public function cc($cc) {
 	 * @param string $bcc
 	 * @return MessageBuilder
 	 */
-	public function bcc($bcc) {
+	public function bcc(string $bcc) {
```
miaulalala, line 86:
```suggestion
	public function bcc(?string $bcc) {
```

miaulalala (changes requested):

tests/Integration/Framework/MessageBuilder.php
ChristophWurst: This doesn't make much sense IMO. If there is no CC, this method shouldn't be called

tests/Integration/Framework/MessageBuilder.php
ChristophWurst: Same here

tests/Integration/Framework/MessageBuilder.php
miaulalala: Maxence

miaulalala (approved):

PR #7539, opened 2022-11-07
Fix open composer actions after changing encryption mode

ChristophWurst:
If the email encryption is turned on or off, the menu should close. NcActionCheckbox does not come with a closeAfterClick like ActionButton, we need to keep track of the menu state ourselves and close manually.

## How to test

1) Set up https://mailvelope.com/en/ for your test email account
2) Configure the account in Mail
3) Open the message composer
4) Turn on encryption

Main: the menu stays open.
This banch: the menu closes.

Bildschirmfoto vom 2022-11-07 10-27-57

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -232,7 +232,8 @@
 				</ButtonVue>
 			</div>
 			<div class="composer-actions--secondary-actions">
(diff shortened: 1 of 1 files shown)


sazanof (approved):
Tested and works.

PR #7540, opened 2022-11-07
Fix envelope and message previews showing PGP ciphertext

ChristophWurst:
| Before | After |
|---|---|
| Bildschirmfoto vom 2022-11-07 10-49-44 | Bildschirmfoto vom 2022-11-07 13-08-26 |
| Bildschirmfoto vom 2022-11-07 11-07-18 | Bildschirmfoto vom 2022-11-07 13-08-33 |

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -65,7 +65,7 @@
 			</div>
 			<div v-if="data.previewText"
 				class="envelope__preview-text">
-				{{ data.previewText }}
+				{{ isEncrypted ? t('mail', 'Encrypted message') : data.previewText }}
 			</div>
 		</template>
 		<template #indicator>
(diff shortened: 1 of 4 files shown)


GretaD (approved):
code looks good

jancborchardt (changes requested):
Looks good, but as per design guidelines we never use italic. Otherwise nice. :)

ChristophWurst:
> as per design guidelines

https://docs.nextcloud.com/server/latest/developer_manual/search.html?q=italic&check_keywords=yes&area=default

ChristophWurst:
Italic styling removed.

jancborchardt:
@ChristophWurst yep, sorry for missing the link :) direct: https://docs.nextcloud.com/server/latest/developer_manual/design/foundations.html#typography
> No italics or UPPER CASE as these text styles are less legible

jancborchardt (approved):

miaulalala (approved):

PR #7556, opened 2022-11-09
Add empty content for loading thread

GretaD:
before ref: #6045 

after
Screenshot from 2022-11-09 13-01-23


GretaD:
another option to show would be
Screenshot from 2022-11-09 13-05-05

src/components/Thread.vue
```diff
@@ -423,4 +423,9 @@ export default {
 .user-bubble__title {
 	cursor: pointer;
 }
+.emptycontent {
+	display: flex;
+	position: relative;
+	margin-top: 50vh;
+}
```
ChristophWurst, line 430: isn't there an off-the-shelf component for this?

sazanof (approved):

src/components/Thread.vue
GretaD: yes, the loading component, the styling is slightly off here. If i dont add any styling it will be aligned on the left.

ChristophWurst: Can we use that component? I doubt that the margin-top:50vh is responsive across window sizes

GretaD: we are using it, see on the template

jancborchardt:
Hmm, it just shows an empty screen with a spinner and a line of text now? We do already have the info of sender, subject and so on, so we can already show that and the box structure.

The main point of the issue was that it changes sizes several times, which should not happen.

ChristophWurst:
> The main point of the issue was that it changes sizes several times, which should not happen.

We are aware: https://github.com/nextcloud/mail/issues/6045#issuecomment-1308609308.

ChristophWurst (approved):
Better than the previous loader :+1:

Issue #7344, opened 2022-09-29
Nextcloud Mail App Not Working With Mailvelope


zohaib09:
### Steps to reproduce

1. Install mailvelop in chrome
2. Add NC domain to authorized domain
3. Create new email with mailvelop Email body and try to encrypt
4. When Enable API in authorized domain for our NC domain then Mailvelope option disappeared during compose of new emails

### Expected behavior

Emails Should be encrypted as other webmail services.

### Actual behavior

Email does not encryptes

### Mail app version

1.13.8

### Mailserver or service

Zimbra

### Operating system

Debian 11

zohaib09:
@ChristophWurst  plesse advice

miaulalala:
Do you have any error messages in your logs or in the browser console?

ChristophWurst:
Tested and the email *is* encrypted. Please provide screenshots.

zohaib09:
Hi

Mailvelope installed

![image](https://user-images.githubusercontent.com/47688270/193426665-6a5bf37c-15df-4b3a-b8e7-c1306bcd9626.png)

Domain is authorized without API enabled (Default settings) then went to NC mail app, click New Message

![image](https://user-images.githubusercontent.com/47688270/193426783-06f86023-7269-4e94-9d44-fef11bc64c66.png)

Then clicked on mailvelooped icon, composed email in Mailvelope-Editor 

![image](https://user-images.githubusercontent.com/47688270/193426911-8e608e06-0d73-4692-afb1-23fd92408271.png)

Then click on encrypt

Then prompted password to sign email (entered the password, clicked ok)

![image](https://user-images.githubusercontent.com/47688270/193427031-cf669b0b-f9f2-4b76-8295-ad3ec7e99b14.png)

Then this empty mail app editor appears

![image](https://user-images.githubusercontent.com/47688270/193427125-d5a7c980-3803-43d4-a82c-34d946d88e68.png)


This is perfectly working for webmail client of zimbra, gmail etc. but not for mail app.

zohaib09:
@ChristophWurst  Hi, please suggest a solution for this. What may be the cause.

ChristophWurst:
You forgot to enable *API mode* for the domain.

zohaib09:
Another thing, in mail app of nextcloud , decryption is perfectly working. Only encryption have issue

[...]

PR #7558, opened 2022-11-09
Fix sending PGP messages as HTML

ChristophWurst:
They are always plain text messages. Otherwise the PGP ciphertext is not recognized on the receiving end.

Fixes https://github.com/nextcloud/mail/issues/7344

## How to test

0. Set up Mailvelope
1. Open the app
2. Start a new message
3. Turn on formatting if it's not alreay on because of the account default
4. Turn on encryption
5. Send the message

Here: messages is sent as text/plain
Main: message is sent as text/html

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -887,7 +887,7 @@ export default {
 				body: this.encrypt ? plain(this.bodyVal) : html(this.bodyVal),
 				attachments: this.attachments,
 				inReplyToMessageId: this.inReplyToMessageId ?? (this.replyTo ? this.replyTo.messageId : undefined),
-				isHtml: !this.editorPlainText,
+				isHtml: !this.encrypt && !this.editorPlainText,
 				requestMdn: this.requestMdn,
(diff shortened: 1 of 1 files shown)


sazanof:
Despite the fact that the message view opens correctly after sending, in the message source contains

`Content-Type: text/plain; charset=utf-8`

Checked before testing - `rich text` is enabled in the account.

The Mailvelope message window does not support html, even if html is intentionally inserted there

ChristophWurst:
Yes, we inline the GPG cipher into the text part of a message.

> The Mailvelope message window does not support html, even if html is intentionally inserted there

Correct. It can only do plain text.

The reproduction steps include *Turn on formatting* because clicking that is necessary to trigger the bug. It doesn't mean that Mailvelope should show formatting because it can't. Formatting only works without encryption.

sazanof (approved):
Thanks for the clarification. Yes, it works.Tested it.

miaulalala (approved):

PR #7562, opened 2022-11-09
Improve loading experience for message on thread

GretaD:
ref #6045 

 

- [x] Fix the message loading screen

Add minimum height to element so the spinner looks less broken. This should be less than an action message
After the loading finished, animate that the element grows to the actual message size (especially text messages)

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -233,6 +233,7 @@ export default {
 	data() {
 		return {
 			loading: false,
+			hidden: true,
 			error: undefined,
 			message: undefined,
 			importantSvg,
@@ -490,6 +491,7 @@ export default {
 		margin-right: 10px;
 		background-color: var(--color-main-background);
 		padding-bottom: 28px;
+		min-height: 150px;
 
 		& + .envelope {
 			margin-top: -28px;
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -145,6 +145,7 @@
 		</div>
 		<Loading v-if="loading" />
 		<Message v-else-if="message"
+			class="message-animation"
 			:envelope="envelope"
 			:message="message"
 			:full-height="fullHeight" />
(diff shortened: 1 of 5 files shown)


GretaD:
https://user-images.githubusercontent.com/12728974/201327212-76c95f9b-1989-428d-8b7c-94119c7086f0.mp4

src/components/ThreadEnvelope.vue
```diff
@@ -490,6 +491,7 @@ export default {
 		margin-right: 10px;
 		background-color: var(--color-main-background);
 		padding-bottom: 28px;
+		min-height: 150px;
```
ChristophWurst, line 494: this doesn't seem to work. the message to iframe loader still causes a jump

GretaD:
the loading starts at sec 16
i cannot see a jump with any thread im testing
[Screencast from 14.11.2022 17:19:42.webm](https://user-images.githubusercontent.com/12728974/201711485-cd00060c-376f-49d1-b4a7-92d5b87799f3.webm)

GretaD:
Bildschirmfoto vom 2022-11-14 17-33-34

ahh, but thats the second icon(before the iframe loads) that i havent removed yet. You proposed to do it 3 prs and this is the second one. On this pr's description i added that this pr does.

ChristophWurst (approved):
:+1: :elephant:

jancborchardt:
Next to the jumping loading icon, 2 other things (possibly for other PRs, your call):
- The "Loading thread" message in the beginning is not needed, just show the loading thread directly
- The height of the container is still very small. Only mails which are just 1 sentence or word long will fit in there. So the default message container height (when the loading spinner is shown) can be increased. E.g. let's say that a mail with 5-7 lines of text could fit.

jancborchardt (changes requested):
Next to the jumping loading icon, 2 other things (possibly for other PRs, your call):
- [ ] The "Loading thread" message in the beginning is not needed, just show the loading thread directly
- [ ] The height of the container is still very small. Only mails which are just 1 sentence or word long will fit in there. So the default message container height (when the loading spinner is shown) can be increased. E.g. let's say that a mail with 5-7 lines of text could fit.

GretaD:
Superseded by #7615

Issue #7497, opened 2022-10-27
Encoding problem with outgoing messages


Thomas-AIRS-Informatique:
### Steps to reproduce

1.Hello, I just realized that I have a formatting problem with sending emails via the Mail application. I am on version 2.0.3. Formatting is enabled in the app. On Nextcloud version 25.0.0
2.
3.

### Expected behavior

Correct text formatting

### Actual behavior

For exemple : 
`Je reste Ã  votre disposition pour tout complÃ©ment d'information. `

### Mail app version

2.0.3

### Mailserver or service

IMAP

### Operating system

Ubuntu 20.04

### PHP engine version

PHP 7.4

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

Hello, I just realized that I have a formatting problem with sending emails via the Mail application. I am on version 2.0.3. Formatting is enabled in the app.
For exemple : `Un exemple : Je reste Ã  votre disposition pour tout complÃ©ment d'information. `

Thomas-AIRS-Informatique:
I performed a test on version 1.6.3 on Nextcloud v24.0.6 and there is no problem.

Extarys:
Happens to me also on Nextcloud 25.0.0 with Mail v2.0.3.
It didn't happen on Nextcloud 24.x

All accents are broken. Let me know if you need someone to test potential fixes/configuration changes.

jejusek:
Does anyone know a way to fix this fault? Let me mention that the national characters in the subject line are correct. Only in the text of the message, the recipient gets strange letters.
I will be grateful. I like to use this app to reply to email.

bastien-sansas:
Hello, it's a critical issue for many people. The application is broken for a lot of non-English communities. We can just send plain text. 

There is also a problem with the content preview of incoming mail (in the mails list). For instance, accent characters are replaced by '?'.

In fact, the problem seems to be appearing when the mail is in rich text mode.

Thanks for your contribution to this project and your work.

mvalberto:
Same problem in spanish, nextcloud 25.0.1, and mail app 2.1.0,  it's a critical issue.

ChristophWurst:
Outgoing messages are temporarily stored in a database table `oc_mail_loca_messages`. Could this be an issue with the utf8 support of your db?

Check https://docs.nextcloud.com/server/stable/admin_manual/configuration_database/mysql_4byte_support.html if you use mysql/mariadb.

mvalberto:
[...]

PR #7586, opened 2022-11-13
Don't decode content for preview twice

kesselb:
Follow-up for #7486 and #7497

Hi :wave: 

I'm unsure about this one. 

We got some reports (e.g., https://github.com/nextcloud/mail/issues/7497#issuecomment-1311073366) that the preview is still wrong. 

I can reproduce the behavior locally with base64 encoded text/plain messages. 
The message data is already decoded when our code to decode the data runs. 

How to test:

- Reset data `update oc_mail_messages set structure_analyzed = null and preview_text = null where 1;` 
- Reload mailbox and wait a sec


ChristophWurst (approved):
:exploding_head:

PR #7592, opened 2022-11-14
Fix saving provisioning and anti-spam settings

olvidalo:
Saving provisioning settings has been broken since e2472148b1f74bd256f2d6041e5782786ce89e80. This fixes the save button by making `native-type` a string instead of a method binding.

It also fixes the button for saving anti-spam settings by adding `native-type="submit"`.

### DIFF
--- a/src/components/settings/AntiSpamSettings.vue
+++ b/src/components/settings/AntiSpamSettings.vue
@@ -52,8 +52,9 @@
 							type="email">
 						<br>
 						<Button
-							type="secondary submit"
+							type="secondary"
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):
Code makes sense!

olvidalo:
Ah you're right, saving would work but there'd still be a warning `Invalid prop: type check failed for prop "nativeType". Expected String, got Function` which is gone after this change.
Sorry for missing this and thanks for approving!

miaulalala (approved):
LGTM!

PR #7605, opened 2022-11-16
Catch auth errors and stop syncing in the frontend

ChristophWurst:


ChristophWurst:
We should rate limit authentication when it fails and only try authentication for x times until the user changes the password or a timer expires (e.g. when a server temporarily doesn't authenticate an account).

Closing

PR #7607, opened 2022-11-16
Show thread subject while loading thread messages

ChristophWurst:
Contributes to https://github.com/nextcloud/mail/issues/6045#issuecomment-1308609308

This removes the empty content *Loading thread* for the case that we know at least one message of the thread. Then we already show the subject on the top of the page and an inner loader (to be replaced by a skeleton view).

Bildschirmfoto vom 2022-11-16 10-42-49

In the case of a thread with no known message, e.g. at a page load or of a thread with an unknown ID, we will still need a loading screen for the whole thread. That is why the earlier empty content has to stay as fallback.

Bildschirmfoto vom 2022-11-16 10-43-01

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -1,13 +1,14 @@
 <template>
 	<AppContentDetails id="mail-message">
-		<Loading v-if="loading" :hint="t('mail', 'Loading thread')" />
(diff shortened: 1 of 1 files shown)


GretaD (approved):
tested and works

PR #7612, opened 2022-11-16
Make HTML body loading a single step experience

ChristophWurst:
Contributes to https://github.com/nextcloud/mail/issues/6045

We have to fetch the message and then wait for the HTML body iframe to load. Those are two steps. We can't change that.

So we fake the experience by only showing one spinner that vanishes when the message and its body have loaded.


GretaD (approved):
works

PR #7615, opened 2022-11-16
Add animation for loading mail content

GretaD:
contributes #6045 

Please test it, i cannot capture the exact animation with a screen recorder


GretaD:
we animate the whole container know see https://github.com/nextcloud/mail/pull/7472

PR #7617, opened 2022-11-16
Handle error when thread id is unknown

### LINKED ISSUE #7608: Inexistent thread loads forever due to lacking error handling
### Steps to reproduce

1. Set up an account
2. Open `/apps/mail/box/priority/thread/12345678987654321`

### Expected behavior

The loading ends once the app sees that the thread does not exist (HTTP 403) and shows a nice info that the thread doesn't exist or was deleted

### Actual behavior

The thread loads forever

### Mail app version

2.2

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

GretaD:
fixes #7608 

- [x] needs #7643

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -2,6 +2,10 @@
 	<AppContentDetails id="mail-message">
 		<!-- Show outer loading screen only if we have no data about the thread -->
 		<Loading v-if="loading && thread.length === 0" :hint="t('mail', 'Loading thread')" />
+		<Error
+			v-else-if="error"
+			:error="error && error.message ? error.message : t('mail', 'Not found')"
+			:message="errorMessage" />
 		<template v-else>
 			<div id="mail-thread-header">
 				<div id="mail-thread-header-fields">
@@ -56,6 +60,7 @@ import debounce from 'lodash/fp/debounce'
 import { getRandomMessageErrorMessage } from '../util/ErrorMessageFactory'
 import Loading from './Loading'
 import logger from '../logger'
+import Error from './Error'
 import RecipientBubble from './RecipientBubble'
 import ThreadEnvelope from './ThreadEnvelope'
 
@@ -64,6 +69,7 @@ export default {
 	components: {
 		RecipientBubble,
 		AppContentDetails,
+		Error,
 		Loading,
 		ThreadEnvelope,
 		Popover,
@@ -215,8 +221,11 @@ export default {
 		},
 		async resetThread() {
 			this.expandedThreads = [this.threadId]
+			this.errorMessage = ''
+			this.error = undefined
 			await this.fetchThread()
 			this.updateParticipantsToDisplay()
+
 		},
 		async fetchThread() {
 			this.loading = true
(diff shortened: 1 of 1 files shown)


src/components/Thread.vue
```diff
@@ -247,8 +248,8 @@ export default {
 				this.loading = false
 			} catch (error) {
 				logger.error('could not load envelope thread', { threadId, error })
-				if (error.isError) {
-					this.errorMessage = t('mail', 'Could not load your message thread')
+				if (error?.response?.status === 403) {
+					showError(t('mail', 'Could not load your message thread, it doesnt exist or is deleted'))
```
ChristophWurst, line 252:
```suggestion
					showError(t('mail', 'Could not load your message thread because it doesn't exist or has been deleted'))
```

src/components/Thread.vue
```diff
@@ -247,8 +248,8 @@ export default {
...
 					this.error = error
 					this.loading = false
 				}
```
ChristophWurst, line 255: add an `else` for any other error and show *Could not load your message thread*

that could happen if the network is down, etc.

ChristophWurst (changes requested):

ChristophWurst (commented):
Why did you go for `showError` instead of `this.errorMessage`?

Your toast will be temporary and the loading is still infinite, isn't it?

GretaD:
> Why did you go for `showError` instead of `this.errorMessage`?
> 
> Your toast will be temporary ~and the loading is still infinite, isn't it?~ loading stops.

Just so i could have a feedback on the UI which is the toast.

ChristophWurst:
You are right. `errorMessage` is no longer used in the template. Let me see why.

ChristophWurst:
> Let me see why.

It was the usual suspects: https://github.com/nextcloud/mail/pull/3540

ChristophWurst:
Can we revive the old error empty content to show instead of the loading view when the thread can't be loaded? This makes the error show in place and permanent. Toasts vanish.

GretaD:
yes, done. Theres only one problem. I cannot click an envelope now, i can scroll and click mailboxes but not envelopes.

Screenshot from 2022-11-17 11-58-22

ChristophWurst:
I think Vue recycles the `Thread` component. The component has no input props but relies on the *route*. See the `$route` watcher. It calls `resetThread` for every route change. Reset `this.error` and `this.errorMessage` in that method.

GretaD:
after resetting `this.error ` and  `this.errorMessage` the envelope list is not responsive anymore

ChristophWurst:
It is repsonsive but the custom .emtpycontent div of the error overlaps the list. Use your browser inspection on top of the evenlope list. 
Bildschirmfoto vom 2022-11-17 18-21-03

GretaD:
> after resetting `this.error ` and `this.errorMessage` the envelope list is not responsive anymore

now it is :)

src/components/Thread.vue
```diff
@@ -247,10 +256,12 @@ export default {
 				this.loading = false
 			} catch (error) {
 				logger.error('could not load envelope thread', { threadId, error })
-				if (error.isError) {
-					this.errorMessage = t('mail', 'Could not load your message thread')
+				if (error?.response?.status === 403) {
+					this.errorMessage = t('mail', 'Could not load your message thread, because it doesn\'t exist or has been deleted')
 					this.error = error
```
ChristophWurst, line 261:
```suggestion
					this.error = t('mail', 'Could not load your message thread')
					this.errorMessage = t('mail', 'The thread doesn\'t exist or has been deleted')
```

ChristophWurst (commented):
Less technical, more human

src/components/Thread.vue
GretaD: :robot:

ChristophWurst (approved):
An icon would make it perfect

PR #7618, opened 2022-11-16
Improve the animation of enevelope list skeleton

GretaD:
its looks way better with gradient rather than the blink

[Screencast from 16.11.2022 18:55:14.webm](https://user-images.githubusercontent.com/12728974/202256845-d378f885-4df8-45c4-889d-c5e885a71ed7.webm)


ChristophWurst:
It's the error at the end of the video that sells it

ChristophWurst (approved):

PR #7643, opened 2022-11-21
Replace emptycontent div with EmptyContent vue component

GretaD:
the empty content in mail is not styled properly because we dont use the EmptyContent vue component

### DIFF
--- a/src/components/NoMessageSelected.vue
+++ b/src/components/NoMessageSelected.vue
@@ -21,30 +21,25 @@
 
 <template>
 	<AppContentDetails>
-		<div id="emptycontent">
-			<IconMail :size="65" />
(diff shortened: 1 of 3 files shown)


src/components/NoMessageSelected.vue
```diff
@@ -21,22 +21,25 @@
 
 <template>
 	<AppContentDetails>
-		<div id="emptycontent">
-			<IconMail :size="65" />
-			<h2>{{ t('mail', 'No message selected') }}</h2>
```
ChristophWurst, line 26: there is some styling for the h2. that is now dead. could you remove it from the bottom of the file?

ChristophWurst (changes requested):

PR #7645, opened 2022-11-21
Fix XOAUTH2 auth via SMTP

ChristophWurst:
Forgot SMTP in https://github.com/nextcloud/mail/pull/6819.

Discovered in https://github.com/nextcloud/mail/pull/7430


miaulalala (approved):
LGTM

PR #7652, opened 2022-11-22
Fix XOAUTH2 auth via SMTP

ChristophWurst:
Corrected version of https://github.com/nextcloud/mail/pull/7645.

Like IMAP, SMTP needs a special format for the OAuth token. Horde can create that for us.


ChristophWurst:
Tested & works with Gmail

GretaD (approved):
i didnt test it

PR #7661, opened 2022-11-23
Do not emit TextEditor input event if nothing changed

ChristophWurst:
Noticed working on use case 4 of https://github.com/nextcloud/mail/issues/4768 where we don't want to save any changes until the user makes any.

Without this patch the freshly initialized text editor emits an input event. But the value is identical to the prop value.

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -189,8 +189,10 @@ export default {
 			this.$emit('ready', editor)
 		},
 		onEditorInput(text) {
-			logger.debug(`TextEditor input changed to <${text}>`)
-			this.$emit('input', text)
+			if (text !== this.value) {
+				logger.debug(`TextEditor input changed to <${text}>`)
+				this.$emit('input', text)
+			}
 		},
 		appendToBodyAtCursor(toAppend) {
 			// https://ckeditor.com/docs/ckeditor5/latest/builds/guides/faq.html#where-are-the-editorinserthtml-and-editorinserttext-methods-how-to-insert-some-content


ChristophWurst:
> Be careful: When you load a message body with HTML images, sanitized by DOMPurify, the content is different because the sanitized value is a bit different.

What should I look out for? Shouldn't the logic catch that? If the editor changes the value it will be different and still emit, no?

kesselb:
> What should I look out for? Shouldn't the logic catch that? If the editor changes the value it will be different and still emit, no?

> where we don't want to save any changes until the user makes any.

The input event may be emitted without "user interaction" when CKEditor "sanitized" the provided content.

kesselb:
| Scenario | Result |
| --- | --- |
| New Message: Change alias => Replace signature (incl. signature above quote) | :heavy_check_mark:  |
| New Message: Enable/disable formatting | :heavy_check_mark: |
| Reply: Change alias => Replace signature (incl. signature above quote) | :heavy_check_mark:  |
| Reply: Enable/disable formatting | :heavy_check_mark: |

st3iny (commented):
Disabling formatting while removing all formatting (red button) does not trigger an event, while it does on main. Is this alright?

cc @kesselb

kesselb:
> Disabling formatting while removing all formatting (red button) does not trigger an event, while it does on main. Is this alright?

Well spotted :+1: 

Disable formatting = Reload the editor with a different set of plugins. 
The content stays the same (it's always html even for plain text).

ChristophWurst:
> The content stays the same (it's always html even for plain text).

so the lack of an input event is fine, right?

kesselb:
> so the lack of an input event is fine, right?

Good question. 

1. Compose a message with rich text
2. Disable formatting
3. Close composer
4. Open message from drafts again

=> Rich text is still enabled. We did not save a draft after changing the editor mode. 

If we need this change for https://github.com/nextcloud/mail/issues/4768 to work, please merge ;)

We can call callSaveDraft from setEditorModeHtml and setEditorModeText to trigger a draft update after an editor mode change.

ChristophWurst:
> If we need this change for #4768 to work, please merge ;)
> 
> We can call callSaveDraft from setEditorModeHtml and setEditorModeText to trigger a draft update after an editor mode change.

:+1:

PR #7677, opened 2022-11-24
Show only existing mailboxes, not all subscribed mailboxes

miaulalala:
Discovered while testing #7430

Mailboxes with the flags / attributes "\NoSelect" or "\NonExistent" should not be shown even if they might be subscribed mailboxes. See [rfc5258](https://www.rfc-editor.org/rfc/rfc5258.html) for the deets.

Not sure if the folder mapper stats needs an update too to additionally filter for `\\noselect` and/or `\\nonexistent`. Let's see after this is merged and #7430 is rebased.

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -55,7 +55,7 @@ class FolderMapper {
 	 */
 	public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
 							   string $pattern = '*'): array {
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):
During testing the change might not show right away because we cache mailbox info for 2h

ChristophWurst:
Should be safe to backport as well

ChristophWurst:
> Should be safe to backport as well

Turntables

PR #7678, opened 2022-11-25
Fix dark theme loading envelopes and thread colour

GretaD:
before
Screenshot from 2022-11-25 10-59-49

after
Screenshot from 2022-11-25 11-07-20


ChristophWurst (approved):
:sunglasses:

ChristophWurst:
backports?

GretaD:
the backport was not needed

PR #7688, opened 2022-11-28
Fix auto config routes for non-admins

ChristophWurst:
Only ISPDB discovery was accessible for non-admins. This means that email addresses of larger providers worked for non-admins during the auto discovery. More custom setup failed hard because the MX and port check routes were blocked.

## How to test

1) Log in as non-admin user
2) Open the Mail app
3) Open the account setup
4) Add an email account from a custom domain

---

Fixes https://github.com/nextcloud/mail/issues/7713


ChristophWurst:
Wrong button :roll_eyes:

PR #7648, opened 2022-11-21
Rework drafts handling on the frontend

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/7077
Fixes https://github.com/nextcloud/mail/issues/4768

- [x] Requires https://github.com/nextcloud/mail/pull/7661
- [x] Requires https://github.com/nextcloud/mail/pull/7691

## Todo
- [x] UC1
- [ ] UC2
- [x] UC3
- [x] UC4
- [x] UC5 https://github.com/nextcloud/mail/pull/7661
- [x] UC6
- [x] T1
- [x] T2
- [x] T3
- [x] T4
- [x] T5
- [ ] T6

PR #7691, opened 2022-11-28
Make the composer component stateless

ChristophWurst:
This moves state (drafts, final message) out of the composer component and up to the message modal. This makes it possible to save a draft when the modal closes in https://github.com/nextcloud/mail/pull/7648. Right we'd need an indirection hack to invoke that from the parent. If we move the logic up a level everything is at the level where we need it.

What I've tested so far
1) Auto-saving a draft
2) Auto-saving another draft revision
3) Draft error handling + retry
4) Sending a message
5) Attach local files

I'll see if we can unit-test a few more critical paths in those two components.

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1,5 +1,5 @@
 <template>
-	<div v-if="state === STATES.EDITING" class="message-composer">
+	<div class="message-composer">
 		<div class="composer-fields composer-fields__from mail-account">
 			<label class="from-label" for="from">
 				{{ t('mail', 'From') }}
@@ -41,7 +41,7 @@
 					:loading="loadingIndicatorTo"
 					:auto-limit="autoLimit"
 					:options-limit="30"
-					@input="callSaveDraft(true, getMessageData)"
+					@input="saveDraftDebounced"
 					@tag="onNewToAddr"
 					@search-change="onAutocomplete($event, 'to')">
 					<template #tag="{ option }">
@@ -90,7 +90,7 @@
 					:auto-limit="autoLimit"
 					:hide-selected="true"
 					:options-limit="30"
-					@input="callSaveDraft(true, getMessageData)"
+					@input="saveDraftDebounced"
 					@tag="onNewCcAddr"
 					@search-change="onAutocomplete($event, 'cc')">
 					<template #tag="{ option }">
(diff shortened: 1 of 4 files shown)


ChristophWurst:
@GretaD could you try and see if you can break anything?

GretaD:
> @GretaD could you try and see if you can break anything?

i broke my dev instance today, so i will break this tomorrow

GretaD:
- As soon as i open composer from new message button - the draft saved is shown.
- After testing several times to write a message and then close it with x or clicking outside, i noticed that on draft folder only 2 messages are saved but no subject or body is in them. My tests had body and a subject.
- I start writing an email then i click discard and close, this error is shown:  `ConsoleLogger.js:64 [ERROR] mail: Could not discard draft 
{level: 0, app: 'mail', uid: 'greta', error: AxiosError}
app
: 
"mail"
error
: 
AxiosError
code
: 
"ERR_BAD_REQUEST"
config
: 
{transitional: {…}, transformRequest: Array(1), transformResponse: Array(1), timeout: 0, adapter: ƒ, …}
message
: 
"Request failed with status code 403"
name
: 
"AxiosError"`

- Messages are sent correctly
- None of my reply drafts were saved.

ChristophWurst:
> * After testing several times to write a message and then close it with x or clicking outside, i noticed that on draft mailbox only 2 messages are saved but no subject or body is in them. My tests had body and a subject.

I can not reproduce this. Subject and body are saved. When I resume the draft those two are still there.

ChristophWurst:
> Request failed with status code 403

Returned if message, mailbox or account doesn't exist.

1) Could you preproduce another time and inspect the DELETE request a bit closer?
2) What `id` is sent in the URL? 
3) Is there a prior DELETE request to delete the same message? 
4) Does the `id` match the last POST to .../draft and the *returned* id?

ChristophWurst:
> None of my reply drafts were saved.

They are saved for me. Even when I just click reply and let the prefilled editor untouched for a few seconds. And also when I make modifications.

GretaD:
After pulling your latest changes, i cannot reproduce any of my errors above

ChristophWurst:
@kesselb could you have a look at this change?

st3iny (approved):
Tested and works.

I experienced 2 bugs that aren't specific to this branch and can be reproduced on master:
1. Old drafts are not deleted after saving a new draft and sending an email.
2. Images/Attachments are not saved in drafts.

Issue #7699, opened 2022-11-30
2.1.3: Inbox is gone and Priority inbox is empty


jakobroehrl:
After the update there is no inbox any more and the Priority inbox is shown as empty (but I my Inbox is full in fact)

![grafik](https://user-images.githubusercontent.com/24530680/204739315-d5a62541-2d41-47ef-8f02-3770e3a7369e.png)

LostinSpacetime:
Same here. Please fix!

ChristophWurst:
> Please fix!

Thanks for the hint

ChristophWurst:
https://github.com/nextcloud/mail/pull/7677 could be the reason for this breakage according to @GretaD. I'm investigating.

LostinSpacetime:
> > Please fix!
> 
> Thanks for the hint

Sorry, got a bit stressed

ChristophWurst:
I'm not fully able to reproduce the issue myself yet but I notice that our app no longer syncs any mailboxes that are not subscribed. That means the *Show only subscribed mailboxes* has no effect.

ChristophWurst:
If you can, let me allow to propose an experiment.

1) Run ``occ mail:account:export <userid>`` to find the id of the affected account
2) Revert the changes https://github.com/nextcloud/mail/pull/7677 did in ``lib/IMAP/FolderMapper.php`` so that it says ``Horde_Imap_Client::MBOX_ALL_SUBSCRIBED`` again.
3) Run the DB statement ``UPDATE oc_mail_accounts SET last_mailbox_sync=0 WHERE id=<accountid>``
4) Reload the page

The app should sync mailboxes again. This time it will include mailboxes that are not subscribed.

ChristophWurst:
@jakobroehrl please fill in the issue form with your next report.

LostinSpacetime:
> If you can, let me allow to propose an experiment.
> 
>     1. Run `occ mail:account:export <userid>` to find the id of the affected account
> 
>     2. Revert the changes [Show only existing mailboxes, not all subscribed mailboxes #7677](https://github.com/nextcloud/mail/pull/7677) did in `lib/IMAP/FolderMapper.php` so that it says `Horde_Imap_Client::MBOX_ALL_SUBSCRIBED` again.
> 
>     3. Run the DB statement `UPDATE oc_mail_accounts SET last_mailbox_sync=0 WHERE id=<accountid>`
> 
>     4. Reload the page
> 
> 
[...]

PR #7700, opened 2022-11-30
Revert "Show only existing mailboxes, not all subscribed mailboxes"

ChristophWurst:
Reverts nextcloud/mail#7677

Fixes https://github.com/nextcloud/mail/issues/7699

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -55,7 +55,7 @@ class FolderMapper {
 	 */
 	public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
 							   string $pattern = '*'): array {
-		$mailboxes = $client->listMailboxes($pattern, Horde_Imap_Client::MBOX_SUBSCRIBED_EXISTS, [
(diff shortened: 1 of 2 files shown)


GretaD (approved):

lib/IMAP/FolderMapper.php
```diff
@@ -55,7 +55,7 @@ class FolderMapper {
 	 */
 	public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
 							   string $pattern = '*'): array {
-		$mailboxes = $client->listMailboxes($pattern, Horde_Imap_Client::MBOX_SUBSCRIBED_EXISTS, [
+		$mailboxes = $client->listMailboxes($pattern, Horde_Imap_Client::MBOX_ALL_SUBSCRIBED, [
```
ChristophWurst, line 58: @miaulalala I got myself confused as well but *all subscribed* doesn't mean we only get the subscribed mailboxes here. It actually returns *all*.

> Return all mailboxes regardless of subscription status, and ensure
>     *     the '\subscribed' attribute is set if mailbox is subscribed
>     *     (implies 'attributes' option is true).

for all subscribed

> Return subscribed mailboxes that exist on the server.

for subscribed exists

PR #1350, opened 2018-11-30
Vue

ChristophWurst:
- [x] Requires https://github.com/nextcloud/mail/projects/4

PR #7706, opened 2022-11-30
Do not show nonexistent or noselect mailboxes

miaulalala:
This fix will filter out any mailboxes with attributes
`\nonexistent` or `\noselect`  as these should never be shown to the
user irrespective of subscription status.

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -67,6 +67,10 @@ class FolderMapper {
 				return null;
 			}
 
+			if (!empty(array_intersect($mailbox['attributes'], ['\noselect', '\nonexistent']))) {
+				return null;
+			}
+
 			try {
 				$client->status($mailbox["mailbox"]);
 			} catch (Horde_Imap_Client_Exception $e) {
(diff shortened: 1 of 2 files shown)


ChristophWurst:
*What if* there is a mailbox that is \\noselect but has children that are selectable?

miaulalala:
> _What if_ there is a mailbox that is \noselect but has children that are selectable?

well good question. but doesn't the listMailboxes function get all mailboxes anyway? If it is a \nonexistent / \noselect mailbox it shouldn't be show, but the child maibox should be in the list. But true, hiding it on the frontend is also an option. I don't know if mailboxes like this would contain data - if they do, we would cache data we shouldn't show and add more syncs than neccesary. tho.

ChristophWurst:
> but the child maibox should be in the list.

Yes, they should. But I think we don't handle children without a parent. It would require code that creates a new virtual parent so that the tree can be rendered in the navigation.

> But true, hiding it on the frontend is also an option. I don't know if mailboxes like this would contain data - if they do, we would cache data we shouldn't show and add more syncs than neccesary. tho.

We don't need to necessarily *hide* them but we have to make sure they can't be opened, if that makes sense.

The old PR was https://github.com/nextcloud/mail/pull/402. The logic might have been lost with https://github.com/nextcloud/mail/pull/1350.

miaulalala:
Okay cool, a frontend solution it is :grin:

PR #7707, opened 2022-11-30
Show outbox only when there is content

### LINKED ISSUE #7694: Outbox shown too prominently

The "Outbox" shown permanently is quite confusing and technical, especially also because it is not common from other mail apps.
I catch myself that when I send an email I always wait for the (1) counter on the Outbox to go away until I do anything else (like deleting the email I responded to). And that kind of defeats the purpose of the inline sending/undo.

What was suggested by @small1 is to only use the Drafts and Sent folder for the flow and not have Outbox – and I agree with that.

It simplifies the mental model a lot as then a mail you are writing/sending can only be in 2 places (Drafts or Sent) instead of 3 (Outbox additionally).

If there’s a mail in scheduled to be sent or an issue with a mail supposed to be sent, the Drafts folder could have the counter.

@GretaD as discussed.

_No response_

### Additional context

Here is how Outlook and Gmail do it.
Gmail has a "Scheduled" entry but most of the time it’s empty anyway.

Outlook | Gmail
-|-
|image

GretaD:
fix #7694
Screenshot from 2022-12-01 17-49-53
Screenshot from 2022-12-01 17-49-30

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -30,7 +30,6 @@
 				:key="'mailbox-' + mailbox.databaseId"
 				:account="unifiedAccount"
 				:mailbox="mailbox" />
-			<NavigationOutbox />
 			<AppNavigationSpacer />
 
 			<!-- All other mailboxes grouped by their account -->
@@ -70,6 +69,9 @@
 			</template>
 		</template>
 		<template #footer>
+			<div v-if="outboxMessages.length !== 0" class="outbox__border">
+				<NavigationOutbox class="outbox" />
+			</div>
 			<AppNavigationSettings :title="t('mail', 'Mail settings')">
 				<template #icon>
 					<IconSetting :size="20" />
@@ -143,6 +145,9 @@ export default {
 		 */
 		passwordIsUnavailable() {
 			return this.$store.getters.getPreference('password-is-unavailable', false)
+		},
+		outboxMessages() {
+			return this.$store.getters['outbox/getAllMessages']
 		},
 	},
 	methods: {
@@ -223,5 +228,19 @@ to {
 :deep(.settings-button) {
 	opacity: .7 !important;
 	font-weight: bold !important;
+	z-index: 1;
+}
+.outbox {
+	margin-left: 6px;
+	width: auto;
+	&__border {
+		border-top: 1px solid var(--color-background-darker);
+	}
+}
+
+:deep(.app-navigation-entry) {
+	&.active {
+		background-color: transparent !important;
+	}
 }
 </style>
--- a/src/components/NavigationOutbox.vue
+++ b/src/components/NavigationOutbox.vue
@@ -65,11 +65,10 @@ export default {
 
 <style lang="scss" scoped>
 :deep(.counter-bubble__counter) {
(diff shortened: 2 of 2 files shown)


ChristophWurst:
You are too fast. We will rethink https://github.com/nextcloud/mail/issues/7694 because the conditional render will make the navigation entries jump

src/components/Navigation.vue
```diff
@@ -223,5 +228,23 @@ to {
...
+	z-index: 1;
+}
+.outbox-border {
+	border-top: 1px solid var(--color-background-darker);
+}
+.outbox {
+	margin-left: 8px;
+	height: 54px;
+}
+:deep(.app-navigation-entry:active),
```
ChristophWurst, line 240: I have the feeling we are using deep styles a bit too often. Is there no other way? Should we add a slot to the component that allows it to render without a background?

src/components/Navigation.vue
```diff
@@ -223,5 +228,23 @@ to {
...
 	opacity: .7 !important;
 	font-weight: bold !important;
+	z-index: 1;
+}
+.outbox-border {
+	border-top: 1px solid var(--color-background-darker);
+}
+.outbox {
+	margin-left: 8px;
+	height: 54px;
```
ChristophWurst, line 238: why does this need a fixed height? can't it size automatically?

GretaD: you are right, most probably not, i needed the height because i wanted to style it with the fade in effect, but it doest work for open state of settings so i move on with border for now.

src/components/Navigation.vue
GretaD: i fixed the code a bit, i only left the .active state to be transparent as it aesthetically nicer.

src/components/Navigation.vue
```diff
@@ -144,6 +146,9 @@ export default {
 		passwordIsUnavailable() {
 			return this.$store.getters.getPreference('password-is-unavailable', false)
 		},
+		messages() {
```
ChristophWurst, line 149: nit: broad name. make it `outboxMessages`

ChristophWurst (approved):

src/components/Navigation.vue
```diff
@@ -70,6 +69,9 @@
 			</template>
 		</template>
 		<template #footer>
+			<div v-if="messages.length !== 0" class="outbox-border">
+				<NavigationOutbox v-if="messages.length !== 0" class="outbox" />
```
ChristophWurst, line 73: the inner v-if is probably not necessary, is it?

ChristophWurst (commented):
Code looks good

src/components/Navigation.vue
```diff
@@ -223,5 +228,18 @@ to {
...
 	font-weight: bold !important;
+	z-index: 1;
+}
+.outbox-border {
+	border-top: 1px solid var(--color-background-darker);
+}
+.outbox {
+	margin-left: 6px;
+	width: auto;
+}
```
ChristophWurst, line 239: BEM

ChristophWurst (approved):

Issue #7687, opened 2022-11-27
Email preview displayed incorrectly when for Content-Transfer-Encoding: quoted-printable messages


ilicz:
### Steps to reproduce

1. Display message encoded with Content-Transfer-Encoding: quoted-printable

### Expected behavior

Correctly displaying encoding of the previewed messages

### Actual behavior

Preview of the message text in the message list as well as in the  "threaded display" in the message window is not decoded correctly and display non-ASCII characters in their encoded form.  
Czech sentence normally displayed as `Přátelé střeleckého sportu a LOSíku`
is then displayed as `P=C5=99=C3=A1tel=C3=A9 st=C5=99eleck=C3=A9ho sportu a LOS=C3=ADku`

This happens only for messages comming with `Content-Transfer-Encoding: quoted-printable` encoding (no matter whether they have Content-type text/html or text/plain)

Preview in the message window:
![image](https://user-images.githubusercontent.com/98597308/204144799-848d04fe-e1ee-46c5-a127-f8caa0d6d3c6.png)
Preview in the message list in the box listing:
![image](https://user-images.githubusercontent.com/98597308/204144723-ce398b68-faae-4daa-b725-bc3e4b0c3624.png)

### Mail app version

2.1.2

### Mailserver or service

IceWarp

### Operating system

DEbian

### PHP engine version

PHP 8.0

### Web server

Apache (supported)

### Database

MariaDB

kesselb:
Hi, can you provide us the message source.

miaulalala:
Thanks I will take a look this week!

miaulalala:
If you are able to, you could give the linked PR a shot @ilicz

kesselb:
Hi,

I cannot reproduce your issue with the provided message source.
It seems to work fine on my dovecot based mail server. Imported the message via thunderbird.

![image](https://user-images.githubusercontent.com/3902676/205335748-13c730eb-00ef-403d-ad81-69e49ea67a6e.png)

![image](https://user-images.githubusercontent.com/3902676/205336337-8e428b8b-4d63-448d-b178-94d37d56f8da.png)

kesselb:
@ilicz could you provide us some more data? 

1. Enable logging for mail: https://github.com/nextcloud/mail/blob/main/doc/admin.md#logging
2. Reset preview text for the broken message: `update oc_mail_messages set structure_analyzed = null and preview_text = null where id = INSERT THE ID OF THE MESSAGE HERE`
[...]

PR #7709, opened 2022-11-30
Check for quoted-printable transfer encoding and

miaulalala:
decode messages before parsing html

Fixes https://github.com/nextcloud/mail/issues/7687


ilicz:
Tested the patch on my playground server and everything seems to be fine now.

ChristophWurst:
Could you check with the original approach in https://github.com/nextcloud/mail/commit/90a90dd24d21da7130cee7a514ca348f0de1fb24? Why did the decoding not work when we use Horde? I would prefer Horde decoding in the hopes that we address more encodings, if any, and not just quoted-printable.

ChristophWurst:
We also use the setTransferenCoding+getContents trick for attachments. That is why I wonder why it works only in some places.

miaulalala:
The issue comes from the `'decode' => true` on the fetch query. Removed that and added back the setContentTransferEncoding and setContent.
@ilicz I would apprechiate a second test if you will? :smile:

miaulalala:
uh looks like the setTransferEncoding is not possibe this way. I'll see how I can work this out.

lib/IMAP/MessageMapper.php
```diff
@@ -726,11 +724,19 @@ public function getBodyStructureData(Horde_Imap_Client_Socket $client,
...
+				}
 				$html = new Html2Text($htmlBody);
 				return new MessageStructureData($hasAttachments, trim($html->getText()), $isImipMessage);
 			}
 			$textBody = $part->getBodyPart($textBodyId);
 			if (!empty($textBody)) {
+				$mimeHeaders = $part->getMimeHeader($textBodyId, Horde_Imap_Client_Data_Fetch::HEADER_PARSE);
+				if ($enc = $mimeHeaders->getValue('content-transfer-encoding')) {
+					$structure->setTransferEncoding($enc);
+				}
```
ChristophWurst, line 739: I believe the ``$structure->setContents(...)`` and ``getContents()`` parts are missing

miaulalala:
Added the content check on the structure. it was indeed missing :sweat_smile:

ilicz:
I applied new patch and it seems the UTF-8 messages are fine now.
Sadly, I only now noticed meeting request sent from MS Outlook server - in iso8859-2. Such messages display correctly in reading pane, but the message preview in the message listing is not displayed correctly:

```
Content-Type: text/plain; charset="iso-8859-2"
Content-Transfer-Encoding: quoted-printable
```

image

kesselb:
I'm unsure if we are on the right track :thinking: 

The provided message works fine on my email server: https://github.com/nextcloud/mail/issues/7687#issuecomment-1335489245

ChristophWurst:
> The provided message works fine on my email server: [#7687 (comment)](https://github.com/nextcloud/mail/issues/7687#issuecomment-1335489245)

@miaulalala found out that if we request decoding when the message is fetched, the BINARY IMAP extension is required. And so depending on the capabilities of your server the result differs for the same message.

lib/IMAP/MessageMapper.php
```diff
@@ -703,7 +703,6 @@ public function getBodyStructureData(Horde_Imap_Client_Socket $client,
 			$partsQuery = new Horde_Imap_Client_Fetch_Query();
 			if ($htmlBodyId !== null) {
 				$partsQuery->bodyPart($htmlBodyId, [
-					'decode' => true,
```
ChristophWurst, line 706: @miaulalala could you document your findings for later reference?

kesselb:
- Added migration to reset preview text
- Rebased the pull request
- Documented our findings https://github.com/nextcloud/mail/issues/7746

ChristophWurst (approved):

lib/IMAP/MessageMapper.php
miaulalala: Leaving a comment here so I don't forget. Not every IMIP server allows for the decdoing to happen there as it needs capability BINARY

PR #7721, opened 2022-12-02
Move OAuth access token into a dedicated column

ChristophWurst:
Depending on the vendor access tokens can be very long and the encrypted value exceeds 4000 chars allowed for our inbound_password column.

Required for https://github.com/nextcloud/mail/issues/6591.

## Todo

- [x] Adjust automated tests
- [x] Manual testing


miaulalala (approved):
Looks gut

PR #7722, opened 2022-12-02
Add Microsoft xoauth2 support

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6591

## Todo
- [x] Requires https://github.com/nextcloud/mail/pull/7721
- [x] Oauth workflow
- [x] User UI
- [x] Admin UI

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -369,6 +369,21 @@ return [
 			'url' => '/integration/google-auth',
 			'verb' => 'GET',
 		],
+		[
+			'name' => 'microsoftIntegration#configure',
+			'url' => '/api/integration/microsoft',
+			'verb' => 'POST',
+		],
+		[
+			'name' => 'microsoftIntegration#unlink',
+			'url' => '/api/integration/microsoft',
+			'verb' => 'DELETE',
+		],
+		[
+			'name' => 'microsoftIntegration#oauthRedirect',
+			'url' => '/integration/microsoft-auth',
+			'verb' => 'GET',
+		],
 	],
 	'resources' => [
 		'accounts' => ['url' => '/api/accounts'],
--- a/src/components/settings/MicrosoftAdminOauthSettings.vue
+++ b/src/components/settings/MicrosoftAdminOauthSettings.vue
@@ -1,0 +1,116 @@
+<!--
+  - @copyright 2022 Christoph Wurst <christoph@winzerhof-wurst.at>
+  -
+  - @author 2022 Christoph Wurst <christoph@winzerhof-wurst.at>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<form @submit.prevent="onSubmit">
+		<label for="mail-microsoft-oauth-tenant-id"> {{ t('mail', 'Tenant ID (optional)') }} </label>
+		<input
+			id="mail-microsoft-oauth-tenant-id"
+			v-model="tenantIdVal"
+			:disabled="loading"
+			type="text">
+		<label for="mail-microsoft-oauth-client-id"> {{ t('mail', 'Client ID') }} </label>
+		<input
+			id="mail-microsoft-oauth-client-id"
+			v-model="clientIdVal"
+			:disabled="loading"
+			type="text"
+			required>
+		<label for="mail-microsoft-oauth-client-secret"> {{ t('mail', 'Client secret') }} </label>
+		<input
+			id="mail-microsoft-oauth-client-secret"
+			v-model="clientSecret"
+			:disabled="loading"
+			type="password"
+			required>
+		<button type="submit" :disabled="!clientIdVal || !clientSecret || clientSecret === PASSWORD_PLACEHOLDER || loading" class="primary">
+			{{ t('mail', 'Save') }}
+		</button>
+		<button :disabled="loading" @click.prevent="onUnlink">
+			{{ t('mail', 'Unlink') }}
+		</button>
+	</form>
+</template>
+
+<script>
+import { showError, showSuccess } from '@nextcloud/dialogs'
+import { translate as t } from '@nextcloud/l10n'
+
+import { configure, unlink } from '../../service/MicrosoftIntegrationService'
+import logger from '../../logger'
+
+const PASSWORD_PLACEHOLDER = '*****'
+
+export default {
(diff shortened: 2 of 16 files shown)


ChristophWurst:
> User UI

I'm facing a little challenge here.

Generally speaking, IMAP accounts with password auth require us to ask the user for a password. XOAUTH2 accounts do not need a password but use a popup screen for the vendor.

With the Gmail XOAUTH2 integration I chose to hide the password input as soon as someone enters an email address that ends with `@gmail.com`. This works. But it only works for Gmail accounts with the default Gmail domain. Any custom domain hosted by Google falls short. This was acceptable for a first version.

Now, with Microsoft, it seems even more likely that someone uses a domain that doesn't easily allow us to detect it's hosted by Microsoft. Some accounts seem to end with `@onmicrosoft.com` but there are many more.

A more sophisticated approach for detecting Gmail and Microsoft accounts is to use the IMAP and SMTP hosts found by auto config detection. But auto config detection only happens *after* the user clicks on *Connect*. At that point we should already have hidden the password, or not.

So there are at least two approaches we can take here.

1) Always show the password input, but no longer make it required for auto mode. Manual mode stays as is. 
    If we detect that the account uses password auth then we abort the setup right after auto detection
    If we detect that it's a Gmail or Microsoft account we just discard the password.
2) Change the setup into a multi step dialogue
    1. Ask for the name and email address
    2. Perform auto detection in the background
    3. If XOAUTH2, open the vendor popup
    3. If config for a password account could be detected, ask for the password
    3. If no config could be detected, show the manual setup

For simplicity I'll probably go with the first one.

ChristophWurst:
Tada

Bildschirmfoto vom 2023-01-19 11-39-35

ChristophWurst:
Until it is done

ChristophWurst:
Please be patient and refrain from commenting on PRs unless the comment adds value to the code changes or discussion. We are struggling too many notifications already. The comments just slow us down. Thank you.

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<form @submit.prevent="onSubmit">
+		<label for="mail-gmail-oauth-tenant-id"> {{ t('mail', 'Tenant ID') }} </label>
```
st3iny, line 24:
```suggestion
		<label for="mail-microsoft-oauth-tenant-id"> {{ t('mail', 'Tenant ID') }} </label>
```

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+		<input
+			id="mail-gmail-oauth-tenant-id"
```
st3iny, line 26:
```suggestion
			id="mail-microsoft-oauth-tenant-id"
```

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+			v-model="tenantIdVal"
+			:disabled="loading"
+			type="text"
+			required>
+		<label for="mail-gmail-oauth-client-id"> {{ t('mail', 'Client ID') }} </label>
```
st3iny, line 31:
```suggestion
		<label for="mail-microsoft-oauth-client-id"> {{ t('mail', 'Client ID') }} </label>
```

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+		<input
+			id="mail-gmail-oauth-client-id"
```
st3iny, line 33:
```suggestion
			id="mail-microsoft-oauth-client-id"
```

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+			v-model="clientIdVal"
+			:disabled="loading"
+			type="text"
+			required>
+		<label for="mail-gmail-oauth-client-secret"> {{ t('mail', 'Client secret') }} </label>
```
st3iny, line 38:
```suggestion
		<label for="mail-microsoft-oauth-client-secret"> {{ t('mail', 'Client secret') }} </label>
```

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+		<input
+			id="mail-gmail-oauth-client-secret"
```
st3iny, line 40:
```suggestion
			id="mail-microsoft-oauth-client-secret"
```

src/components/settings/MicrosoftAdminOauthSettings.vue
st3iny: Styling might be adjusted. Or is it just used for the label? (Applies to the other occurrences as well.)

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+			v-model="clientSecret"
+			:disabled="loading"
+			type="password"
+			required>
+		<button type="submit" :disabled="!tenantId || !clientIdVal || !clientSecret || clientSecret === PASSWORD_PLACEHOLDER || loading" class="primary">
```
st3iny, line 45:
```suggestion
		<button type="submit" :disabled="!tenantIdVal || !clientIdVal || !clientSecret || clientSecret === PASSWORD_PLACEHOLDER || loading" class="primary">
```

Fixes save button never being enabled.

src/components/settings/MicrosoftAdminOauthSettings.vue
```diff
@@ -0,0 +1,117 @@
...
+			clientIdVal: this.clientId,
+			clientSecret: this.clientId ? PASSWORD_PLACEHOLDER : '',
+			PASSWORD_PLACEHOLDER,
+		}
+	},
+	methods: {
+		async onSubmit() {
+			this.loading = true
+			try {
+				await configure(this.tenantId, this.clientIdVal, this.clientSecret)
```
st3iny, line 88:
```suggestion
				await configure(this.tenantIdVal, this.clientIdVal, this.clientSecret)
```

Fixes tenant id never being sent to the backend.

src/components/settings/MicrosoftAdminOauthSettings.vue
st3iny: Done

appinfo/routes.php
```diff
@@ -369,6 +369,21 @@
...
 			'verb' => 'GET',
 		],
+		[
+			'name' => 'microsoftIntegration#configure',
+			'url' => '/api/integration/microsoft',
+			'verb' => 'POST',
+		],
+		[
+			'name' => 'microsoftIntegration#unlink',
+			'url' => '/api/integration/google',
```
kesselb, line 379:
```suggestion
			'url' => '/api/integration/microsoft',
```

;)

kesselb (approved):

st3iny (approved):
Tested and works with my personal account.

flowlee:
This does not currently work for addresses with the @bildung.gv.at domain. These are the official education addresses used by teachers in Austria and they are hosted by Microsoft and normally accessed via outlook.com.

PR #7756, opened 2022-12-09
Show subject on a thread when the subject is changed

GretaD:
While working on showing the preview of a thread, the clean subject was removed: https://github.com/nextcloud/mail/pull/7021

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -60,6 +60,9 @@
 				<div class="sender">
 					{{ envelope.from && envelope.from[0] ? envelope.from[0].label : '' }}
 				</div>
+				<div v-if="hasChangedSubject" class="subline">
+					{{ cleanSubject }}
+				</div>
 				<div v-if="showSubline" class="subline">
 					<span class="preview">
 						{{ isEncrypted ? t('mail', 'Encrypted message') : envelope.previewText }}
@@ -236,6 +239,10 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+		threadSubject: {
+			required: true,
+			type: String,
+		},
 	},
 	data() {
 		return {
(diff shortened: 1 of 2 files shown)


GretaD:
there is only this scenario that i dont understand if its shown correctly 
Screenshot from 2022-12-14 13-18-15

ChristophWurst:
> there is only this scenario that i dont understand if its shown correctly

We can't tell if you don't give us infos about the messages in the thread :wink: 

Did the normalized subject (without re/fwd/etc) change?

GretaD:
> > there is only this scenario that i dont understand if its shown correctly
> 
> We can't tell if you don't give us infos about the messages in the thread wink
> 
> Did the normalized subject (without re/fwd/etc) change?

i am not sure what to check, this thread has only this message, the email that was forwarded to me has `RE: [EXTERNAL] ..`

ChristophWurst:
If there is only one message then there is no changing subject. The thread subject is identical to the message subject. The thread should *not* show the subject therefore.

src/components/ThreadEnvelope.vue
```diff
@@ -310,6 +318,15 @@ export default {
...
 			)
 		},
+		hasChangedSubject() {
+			return this.cleanSubject !== this.cleanThreadSubject
+		},
+		cleanSubject() {
+			return this.envelope.subject.replace(/((?:[\t ]*(?:R|RE|F|FW|FWD):[\t ]*)*)/i, '')
+		},
+		cleanThreadSubject() {
+			return this.threadSubject.replace(/((?:[\t ]*(?:R|RE|F|FW|FWD):[\t ]*)*)/i, '')
```
st3iny, line 328: Please move the cleaning logic to a reusable method. You could name it `filterSubject()`. Otherwise, the code is duplicated.

st3iny (commented):
Looks good.

src/components/ThreadEnvelope.vue
```diff
@@ -33,6 +33,7 @@
 			<div
 				v-if="isImportant"
 				class="app-content-list-item-star icon-important"
+				:class="{ 'important-position-not-expanded' : importantPositionWithSubjectNotExpanded }"
```
ChristophWurst, line 36: bem?

ChristophWurst (approved):

ChristophWurst (approved):

ChristophWurst:
Breaks js unit tests

PR #7793, opened 2022-12-20
ci: Do not check for fixups in draft PRs

ChristophWurst:
It's discouraging when fixups are actually preferred.


ChristophWurst:
Sent upstream https://github.com/nextcloud/.github/pull/102

PR #7796, opened 2022-12-21
Validate remote hosts

ChristophWurst:
This borrows some code from https://github.com/nextcloud/server/pull/34852 so we can use the feature earlier.


ChristophWurst:
> Class, interface or enum named IPLib\Factory does not exist

Shipped by server

kesselb (approved):

PR #7805, opened 2022-12-21
feat(occ): Add ML prediction command

ChristophWurst:
An easy way to test a trained model.

1) Train with `occ mail:account:train`
2) Test with `occ mail:predict-importance`


ChristophWurst:
CI failure addressed in https://github.com/nextcloud/mail/pull/7806

st3iny (approved):

PR #7806, opened 2022-12-22
chore(static analysis): Ignore Symfony Command classes for Psalm

ChristophWurst:
Static analysis fails in https://github.com/nextcloud/mail/pull/7805 due to the use of the Symfony Command class.

As long as there is no Nextcloud abstraction we are out of alternatives. Instead of updating the baseline for every new command we can also tell Psalm to not report this missing class.

:crossed_fingers:


st3iny (approved):
:elephant:

kesselb (approved):

PR #7807, opened 2022-12-22
ci: Use setup-nextcloud action

ChristophWurst:
Ref https://github.com/marketplace/actions/set-up-nextcloud


st3iny (approved):
Nice one! Apply it to Calendar as well?

ChristophWurst:
I'm experimenting a bit more but yes I hope to get rid of some duplication in our workflows

PR #7813, opened 2022-12-23
chore(deps): Use conventional commits for bumps

ChristophWurst:
feat for major/minor bumps of production deps
fix for patch bumps of production deps
chore for the rest

The configuration is a mess. I know.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,9 @@ updates:
   reviewers:
   - "ChristophWurst"
   - "kesselb"
+  commit-message:
+    prefix: chore
+    include: scope
 - package-ecosystem: npm
   directory: "/"
   schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -26,6 +29,55 @@ updates:
...
+      update-types: ["version-update:semver-major", "version-update:semver-minor"]
+  allow:
+    - dependency-type: production
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: weekly
+    day: saturday
+    time: "03:00"
+    timezone: Europe/Paris
```
kesselb, line 65:
```suggestion
    timezone: Europe/Vienna
```

kesselb (approved):
:sunglasses:

.github/dependabot.yml
ChristophWurst: An important detail

PR #7846, opened 2022-12-27
fix: use unmodified email body for forward/reply

### LINKED ISSUE #5233: Replying to / forwarding rich text emails replaces original links with redirect links
### Expected behavior
When an email account is set to use rich text and an email containing links is forwarded to another email address and viewed from a different client than Mail, the original links should in the email and not the redirect links added when viewing the email in Mail.

If the forwarded email is view in Mail, the link should only redirect once.

### Actual behavior
When viewing the forwarded email from a different client than Mail, the redirect links inserted by Mail have replaced the original links. So if the email is forwarded to someone who doesn't have a user on the Nextcloud instance the email was sent from, they get redirected to the login screen of that Nextcloud instance when clicking on the link.

If the forwarded email is viewed from Mail another redirect is added to the links, so the user is redirected twice before getting sent to the original destination.

### Mail app

**Mail app version:** 1.9.5

**Mailserver or service:** Tested on Runbox and Simply.com

### Server configuration

**Operating system**: Ubuntu 20.04

**Web server:** Apache

**Database:** MySQL 8.0.24

**PHP version:** 7.4.18

**Nextcloud Version:** 21.0.1 (snap)

### Client configuration
**Browser:** Firefox 89

**Operating system:** KDE Neon (Ubuntu 20.04)

kesselb:
Fix #5233

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -506,28 +506,30 @@ class MessagesController extends Controller {
 
 			$client = $this->clientFactory->getClient($account);
 			try {
-				$html = $this->mailManager->getImapMessage(
+				$imapMessage = $this->mailManager->getImapMessage(
 					$client,
 					$account,
 					$mailbox,
 					$message->getUid(),
 					true
-				)->getHtmlBody(
-					$id
 				);
 			} finally {
 				$client->logout();
 			}
 
-			$htmlResponse = $plain ?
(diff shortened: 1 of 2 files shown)


hamza221:
Is this still needed? I couldn't reproduce on main

ChristophWurst:
I can't reproduce either.

Issue #7095, opened 2022-08-30
Navigation expand icon moved to the right


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Configure an account
3. Create a hierarchy of mailboxes
4. Hover the parent mailbox

### Expected behavior

Expand icon shows on the left, instead of the mailbox icon.

### Actual behavior

Expand icon shows on the right.

![Bildschirmfoto vom 2022-08-23 19-40-13](https://user-images.githubusercontent.com/1374172/186227576-9a5b7738-e335-4272-9533-d1738fc11b5a.png)

### Mail app version

v1.4 46c445b3fa227376734303c92248aaae605a2bf5

ChristophWurst:
I think this is a regression of https://github.com/nextcloud/nextcloud-vue/pull/2704.

GretaD:
fixed on nc/vue 7.2.0

ChristophWurst:
link?

GretaD:
there is no specific fix for this problem, but on nc/vue 7.2.0 the problem doesnt exist anymore. I can have a look to see what exactly fixed this one too.

GretaD:
> fixed on nc/vue 7.2.0

well, looks like on 7.3.0 it broken again

ChristophWurst:
It looks fixed with v3 :sunglasses:

PR #7856, opened 2023-01-02
Force the icon-collapse to align left

GretaD:
fixes: #7095

This is an old bug. Most probably this should be fixed on nc/vue but i dont see which change caused this error there. This error can be caused also from some server css, but so far i failed to find the right change, so pushing this pr till we figured it out.
before
Screenshot from 2023-01-02 18-09-13
after
Screenshot from 2023-01-02 18-08-52

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -695,4 +695,8 @@ export default {
 .counter-bubble__counter {
 	max-width: initial;
 }
+// TODO: upsteream collapse icon position fix
+  :deep(.icon-collapse) {
+	  position: absolute !important;
+  }
 </style>


GretaD:
@juliushaertl i added you as a reviewer since this error is also happening on deck: https://github.com/nextcloud/deck/issues/4352

juliushaertl:
I think I found a potential upstream fix. Looks like this only works in the vue component docs because the button style is loaded after the NcAppNavigationIconCollapsible one. Will push something shortly.

juliushaertl:
https://github.com/nextcloud/nextcloud-vue/pull/3608 works with deck, so likely also fixes the issue for mail

GretaD:
> I think I found a potential upstream fix. Looks like this only works in the vue component docs because the button style is loaded after the NcAppNavigationIconCollapsible one. Will push something shortly.

testing in a bit, thank you so much <3

GretaD:
tested and it works with Julius pr :partying_face:

PR #7858, opened 2023-01-02
[stable1.15] fix: check that IpUtils is available

kesselb:
IpUtils is required to check if a given ip address is withing a range.

To backport the functionality the class was copied from the library to lib/private therefore the different namespace.

Signed-off-by: Daniel Kesselberg <mail@danielkesselberg.de>

### DIFF
--- a/lib/Net/IpAddressClassifier.php
+++ b/lib/Net/IpAddressClassifier.php
@@ -28,7 +28,6 @@ namespace OCA\Mail\Net;
 use IPLib\Address\IPv6;
 use IPLib\Factory;
 use IPLib\ParseStringFlag;
-use Symfony\Component\HttpFoundation\IpUtils;
 use function filter_var;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Net/IpAddressClassifier.php
```diff
@@ -69,9 +68,26 @@ public function isLocalAddress(string $ip): bool {
...
+			if (\OC\Http\IpUtils::checkIp($ip, self::LOCAL_ADDRESS_RANGES)) {
+				/* Within local range */
+				return true;
+			}
+		} else {
+			// < Nextcloud 24.0.4
+			// < Nextcloud 23.0.8
+			// < Nextcloud 22.2.11
+			// IpUtils is not available.
 		}
```
miaulalala, line 91: can't you remove that else block since it will return false below anyway?

miaulalala (approved):

miaulalala (approved):

lib/Net/IpAddressClassifier.php
kesselb: done done

PR #7885, opened 2023-01-10
chore(oauth): Remove branding from generic OAUTH settings

ChristophWurst:
Extracted from https://github.com/nextcloud/mail/pull/7722.

Removes Google branding from all file names and component names that are not specific to Google.


st3iny (approved):
Creating and using a google mail account still works. Code looks good too.

PR #7886, opened 2023-01-11
fix(XOAUTH2): Defer OAUTH account detection

ChristophWurst:
Defer the detection of OAUTH until auto config is done. Only if the IMAP and SMTP servers are from Google and we have OAUTH credentials then we try OAUTH. This makes it possible to use Gmail OAUTh with domains other than `@gmail.com`.

As a side effect we will no longer require a password unless Google integration is not set up.

Ref https://github.com/nextcloud/mail/pull/7722#issuecomment-1378412116

## How to test

0) Set up Gmail in the admin settings
1) Open the account setup
2) Enter a dummy email `test@hotmail.com`
3) Leave the password empty

-> Auto detection succeeds but you get the feedback *Password required*

4) Fill in a password

-> Auto detection succeeds but you get the feedback that username or password is wrong.

5) Enter a valid gmail account (can end with `@gmail.com` but no longer has to)

-> You get the popup window to configure it

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -26,16 +26,13 @@
 				<p v-if="!isValidEmail(emailAddress)" class="account-form--error">
 					{{ t('mail', 'Please enter an email of the format name@example.com') }}
 				</p>
-				<label v-if="!useGoogleSso"
-					for="auto-password"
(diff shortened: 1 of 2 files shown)


st3iny (approved):
Tested and works. I really like the UX of the whole auth flow!

ChristophWurst:
I'm working on the jest failures

PR #7921, opened 2023-01-20
perf(PHP): Use static closures where possible

ChristophWurst:
They allow GC because they do not capture `$this`


miaulalala (approved):
image

st3iny (approved):
:broom:

PR #7939, opened 2023-01-25
refactor(php): Use attribute instead of annotation for TrapError

ChristophWurst:
New year, new PHP goodies.


miaulalala (approved):
image

st3iny (approved):
Truly epic!

PR #7987, opened 2023-02-06
Add Provision info to occ account:export command

hamza221:
This info is helpful when debugging production instances where users have multiple accounts and it isn't clear which ones they created themselves and which are provisioned.


ChristophWurst (approved):
Please squash your commits into one

PR #7991, opened 2023-02-07
Use keyboard shortcut only if ACL allows it

### LINKED ISSUE #7910: A user should be able to use keyboard shortcuts only if ACLs allow it

As a recipient of a shared mailbox I want the app to only show me options allowed by ACL

Change the front end for actions on keyboard shortcuts:
- Compose new message
- Toggle start
- Toggle unread
- Delete
- Send

_No response_

### Additional context

This is part of https://github.com/nextcloud/mail/issues/7057

GretaD:
fixes #7910

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -306,6 +306,24 @@ export default {
 				this.loadingMore = false
 			}
 		},
+		hasDeleteAcl() {
+			if (!this.mailbox.myAcls) {
+				return true
+			}
+			return this.mailbox.myAcls.indexOf('d') !== -1
+		},
+		hasSeenAcl() {
+			if (!this.mailbox.myAcls) {
+				return true
+			}
+			return this.mailbox.myAcls.indexOf('s') !== -1
+		},
+		hasArchiveAcl() {
+			if (!this.mailbox.myAcls) {
+				return true
+			}
+			return this.mailbox.myAcls.indexOf('t') !== -1 && this.mailbox.myAcls.indexOf('e') !== -1
+		},
 		async handleShortcut(e) {
 			const envelopes = this.envelopes
 			const currentId = parseInt(this.$route.params.threadId, 10)
@@ -350,6 +368,9 @@ export default {
 				})
 				break
 			case 'del':
+				if (!this.hasDeleteAcl()) {
+					return
+				}
 				logger.debug('deleting', { env })
 				this.onDelete(env.databaseId)
 				try {
@@ -374,6 +395,9 @@ export default {
 
 				break
 			case 'arch':
+				if (this.hasArchiveAcl()) {
+					return
+				}
 				logger.debug('archiving', { env })
 				this.onArchive(env.databaseId)
 				try {
@@ -389,7 +413,6 @@ export default {
 
 					showError(t('mail', 'Could not archive message'))
 				}
-
 				break
 			case 'flag':
 				logger.debug('flagging envelope via shortkey', { env })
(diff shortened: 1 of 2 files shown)


ChristophWurst:
We use https://www.npmjs.com/package/vue-shortkey for keyboard shortcuts. The directive is attached to `AppContentList`. `MailboxThread::onShortcut` handles the shortcut and emits an event. In ``Mailbox::handleShortcut`` you find the shortcut handler method. Adjust that method to ACLs.

src/components/Mailbox.vue
```diff
@@ -351,7 +369,9 @@ export default {
 				break
 			case 'del':
 				logger.debug('deleting', { env })
-				this.onDelete(env.databaseId)
+				if (e.srcKey === 'del' && this.hasDeleteAcl()) {
```
ChristophWurst, line 372: this code is inside a switch case where e.shortKey matches `del`. I don't see a need for a second check. The first part of the if condition will always be `true`

src/components/Mailbox.vue
```diff
@@ -351,7 +369,9 @@ export default {
...
+					this.onDelete(env.databaseId)
+				}
 				try {
 					await this.$store.dispatch('deleteThread', {
```
ChristophWurst, line 376: this still executes, even with lacking ACL rights, doesn't it?

ChristophWurst (changes requested):

src/components/Mailbox.vue
GretaD: Here i noticed a bug, the message is gone from the list giving you the impression that is deleted but after reload message it is still there.

src/components/Mailbox.vue
```diff
@@ -407,12 +429,14 @@ export default {
 				break
 			case 'unseen':
 				logger.debug('marking message as seen/unseen via shortkey', { env })
```
ChristophWurst, line 431: nit: move into the if

nit2: use an early return like above

src/views/KeyboardShortcuts.vue
```diff
@@ -29,6 +29,10 @@
 				<dt><kbd>U</kbd></dt>
 				<dd>{{ t('mail', 'Toggle unread') }}</dd>
 			</div>
+			<div>
+				<dt><kbd>A</kbd></dt>
+				<dd>{{ t('mail', 'Toggle archiving') }}</dd>
```
ChristophWurst, line 34:
```suggestion
				<dd>{{ t('mail', 'Archive') }}</dd>
```

src/components/Mailbox.vue
```diff
@@ -350,7 +368,11 @@ export default {
 				})
 				break
 			case 'del':
-				logger.debug('deleting', { env })
+				if (!this.hasDeleteAcl()) {
+					logger.debug('deleting', { env })
```
ChristophWurst, line 372: it logs when it doesn't delete, and doesn't log when it does delete

src/components/Mailbox.vue
```diff
@@ -374,7 +396,10 @@ export default {
 
 				break
 			case 'arch':
-				logger.debug('archiving', { env })
+				if (this.hasArchiveAcl()) {
+					logger.debug('archiving', { env })
```
ChristophWurst, line 400: same here

src/components/Mailbox.vue
```diff
@@ -406,13 +430,18 @@ export default {
 
 				break
 			case 'unseen':
-				logger.debug('marking message as seen/unseen via shortkey', { env })
+				if (this.hasSeenAcl()) {
+					logger.debug('marking message as seen/unseen via shortkey', { env })
```
ChristophWurst, line 434: same

ChristophWurst (commented):
Move the log *after* the if

PR #7996, opened 2023-02-08
fix: allow sending of messages with empty body content

miaulalala:

### DIFF
--- a/lib/Service/MimeMessage.php
+++ b/lib/Service/MimeMessage.php
@@ -46,47 +46,51 @@ class MimeMessage {
 	 */
 	public function build(bool $isHtml, string $content, array $attachments): Horde_Mime_Part {
 		if ($isHtml) {
-			$doc = new DOMDocument();
-			$doc->loadHTML(mb_convert_encoding($content, 'HTML-ENTITIES', 'UTF-8'), LIBXML_HTML_NODEFDTD | LIBXML_HTML_NOIMPLIED);
-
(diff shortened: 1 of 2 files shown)


lib/Service/MimeMessage.php
```diff
@@ -127,7 +127,7 @@ public function build(bool $isHtml, string $content, array $attachments): Horde_
 			$bodyPart = new Horde_Mime_Part();
 			$bodyPart->setType('text/plain');
 			$bodyPart->setCharset('UTF-8');
-			$bodyPart->setContents($content);
+			$bodyPart->setContents($content ?? '');
```
miaulalala, line 130: this can go actually

GretaD (approved):
tested and works <3

ChristophWurst:
Could you extend https://github.com/nextcloud/mail/blob/main/tests/Unit/Service/MimeMessageTest.php to cover the previous error case?

ChristophWurst (approved):

ChristophWurst:
Does this fix the error `Could not send outbox message 2696: DOMDocument::loadHTML(): Argument #1 ($source) must not be empty` @miaulalala?

Also, what versions are affected and how far do we need to backport?

kesselb:
> Does this fix the error `Could not send outbox message 2696: DOMDocument::loadHTML(): Argument #1 ($source) must not be empty`

Yes :partying_face:

PR #7997, opened 2023-02-08
Show special use icons based on account preference instead of IMAP attributes

### LINKED ISSUE #3917: Show special use icons based on account preference, not (only?) the IMAP attribute
### Feature Request

As of https://github.com/nextcloud/mail/pull/3820 the user can pick their special mailboxes, which means they can potentially select a drafts/sent/trash mailbox that isn't attributed as such on an IMAP level. RIght now the special icons for those mailboxes are based on the attribute. But we should consider making the account preference our "source of truth", or at least have the special icons for mailboxes with the IMAP attribute *or* being selected by the user.

#### Summary

As per above.

---

@jancborchardt do you have a preference? Should a drafts/sent/trash mailbox still have a special icon even if it's not used for what the attribute otherwise claims?

hamza221:
closes #3917 
Signed-off-by: hamza221 <hamzamahjoubi221@gmail.com>

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -48,13 +48,13 @@
 					:size="20" />
 				<IconFavorite v-else-if="filter === 'starred'"
 					:size="20" />
-				<IconDraft v-else-if="mailbox.specialRole === 'drafts'"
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Clean and simple :shipit:

PR #7998, opened 2023-02-09
ci(code-coverage): Use codecov token to improve CI stability

ChristophWurst:
Ref https://github.com/codecov/codecov-action/issues/903


kesselb (approved):

GretaD:
test.yml on stable1.15 is completely different that's why it fails

PR #4968, opened 2021-04-19
Add multi account provisioning

miaulalala:
Rework this feature to use a table instead of the JSON that's saved in the APPVALUE parameter.

Fixes #4459

PR #8004, opened 2023-02-10
fix(provisioning): Return database ID of new configs

ChristophWurst:
https://github.com/nextcloud/mail/pull/4968 oversight.

## How to test

1) Open ``/settings/admin/groupware``
2) Click *+ Add new config*
3) Enter `aaa` as provisioning doman, `test` as email address template
4) Click *Save config*
5) Click *Unprovision & Delete config*

Main: config disappears visually, but request sends a `{id}` for the config ID.
Here: real database ID is sent for deletion. Config is removed from the DB as well.

Beware: even if the config is deleted from the DB you might still see it after a page reload: https://github.com/nextcloud/mail/pull/7992

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -64,14 +64,14 @@ class SettingsController extends Controller {
 
 	public function createProvisioning(array $data): JSONResponse {
 		try {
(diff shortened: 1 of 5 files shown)


miaulalala (approved):
its-me-im-the-problem

PR #8020, opened 2023-02-14
chore(autoloading): Only use authoritative classmaps for production

ChristophWurst:
## How to test

1) Run `composer i`
2) Check `vendor/composer/autoload_real.php`

Main: you find a `$loader->setClassMapAuthoritative(true);` and switching between branches could potentially trigger missing classes if you don't run `composer i` or `composer dump-autoload` to update the class map
Here: class map is still optimized on dev but tries to look up missing classes. On production we use the authoritative classmaps nevertheless.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,6 @@
 			"bamarni/composer-bin-plugin": true
 		},
 		"optimize-autoloader": true,
-		"classmap-authoritative": true,
 		"autoloader-suffix": "Mail"
 	},
 	"require": {
--- a/krankerl.toml
+++ b/krankerl.toml
@@ -1,6 +1,6 @@
 [package]
 before_cmds = [
-	"composer install --no-dev -o",
+	"composer install --no-dev -a",
 	"npm install --deps",
 	"npm run build",
 ]


miaulalala (changes requested):
I tried testing this with the following steps:

1) Check out main
2) Run composer i
3) check out this branch
4) create new empty migration via occ and bump info.xml
5) run the upgrade from the web ui

Unfortunately the lookup didn't work with the typical error that happens when the composer autoloader wasn't dumped:

image

st3iny (approved):
Tested and works.

---

@miaulalala The existing autoloader has the authoritative state ingrained and thus it has to be generated once more after this PR. The setting from `composer.json` is only read on dumping the autoloader (or installing dependencies).

`vendor/composer/autoload_real.php`
```
  41   │         }
  42   │
  43   │         $loader->setClassMapAuthoritative(true);
  44   │         $loader->register(true);
  45   │
  46   │         if ($useStaticLoader) {
```

ChristophWurst:
Excellent catch. That explains the missing class despite my changes.

miaulalala (approved):
Tested again:

1) Checked out the PR
2) composer i
3) created migration
4) bumped info.xml
5) started upgrade via webUI
6) Migration runs through
7) Profit!

PR #8021, opened 2023-02-14
fix: use api to read input value

kesselb:
> The internal dom structure was changed by https://github.com/nextcloud/nextcloud-vue/pull/3517

stable2.2 uses nextcloud/vue 7.1 => no backport

Test Case 1:

- Open menu for account in navigation
- Click "Add mailbox"
- Add a mailbox

Test Case 2:

- Open menu for mailbox in navigation
- Click "Add submailbox" 
- Add a sub mailbox

It does not work on main because the internal dom strucure was changed.


GretaD (approved):
:+1:

PR #8036, opened 2023-02-16
Check special use via dedicated method

miaulalala:
instead of using a string comparison for the special use property.

### DIFF
--- a/lib/Listener/FlagRepliedMessageListener.php
+++ b/lib/Listener/FlagRepliedMessageListener.php
@@ -88,7 +88,7 @@ class FlagRepliedMessageListener implements IEventListener {
 						continue;
 					}
 					// ignore drafts and sent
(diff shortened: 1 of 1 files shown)


lib/Listener/FlagRepliedMessageListener.php
```diff
@@ -84,7 +84,7 @@ public function handle(Event $event): void {
 				try {
 					$mailbox = $this->mailboxMapper->findById($message->getMailboxId());
 					// ignore drafts and sent
-					if ($mailbox->getSpecialUse() === '["sent"]' || $mailbox->getSpecialUse() === '["drafts"]') {
+					if ($mailbox->isSpecialUse( 'sent') || $mailbox->isSpecialUse('drafts')) {
```
ChristophWurst, line 87:
```suggestion
					if ($mailbox->isSpecialUse('sent') || $mailbox->isSpecialUse('drafts')) {
```

ChristophWurst (approved):
Makes the release tarball 1B smaller

PR #8049, opened 2023-02-20
feat(acl): check delete acl for move operation

kesselb:
Rename mailbox needs x (delete mailbox) for the source mailbox. 
Hide the button if we cannot.

Renamed the label from Move to Move mailbox like the other operations (e.g. Clear mailbox, Delete mailbox).


hamza221 (approved):
Tested ,LGTM

PR #8053, opened 2023-02-21
chore(deps): Enable renovate bumps on main

ChristophWurst:
Ref https://github.com/nextcloud/groupware/issues/63

* Production dep bumps on Wednesday
* Dev dep bumps monthly
* Security bumps for stable branches (if that works at all)

I did not declare the ckeditor family as a dependency group because it looks like renovate detects monorepos automatically.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,10 +7,11 @@
 	],
 	"timezone": "Europe/Vienna",
 	"schedule": [
-		"before 5am every weekday"
+		"before 5am on wednesday"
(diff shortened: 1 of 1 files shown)


kesselb (approved):

ChristophWurst:
> I did not declare the ckeditor family as a dependency group because it looks like renovate detects monorepos automatically.

It does not: https://github.com/nextcloud/mail/pull/8067

PR #8094, opened 2023-02-23
replace hasAcls() with mailboxHasRights()

hamza221:
Part of [#7057](https://github.com/nextcloud/mail/issues/7057)

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -227,7 +227,7 @@ import ImportantIcon from './icons/ImportantIcon'
 import IconSend from 'vue-material-design-icons/Send'
 import MoveMailboxModal from './MoveMailboxModal'
 import { PRIORITY_INBOX_ID, UNIFIED_INBOX_ID } from '../store/constants'
-
+import { mailboxHasRights } from '../util/acl'
 import { clearCache } from '../service/MessageService'
 import { getMailboxStatus } from '../service/MailboxService'
 import logger from '../logger'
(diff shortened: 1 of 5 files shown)


src/tests/unit/components/Envelope.vue.spec.js
```diff
@@ -80,7 +81,7 @@ describe('Envelope', () => {
 			localVue,
 		})
 
-		expect(view.vm.hasSeenAcl).toBe(true)
+		expect(mailboxHasRights(view.vm.mailbox,'s')).toBe(true)
```
ChristophWurst, line 84: if the component no longer has a `hasSeenAcl` drop this test

if there is a `hasSeenAcl` that calls `mailboxHasRights` keep the test and test `hasSeenAcl` like before

src/components/NavigationMailbox.vue
```diff
@@ -379,7 +379,7 @@ export default {
 			return this.$route.params.mailboxId === this.mailbox.databaseId
 		},
 		isValidDropTarget() {
-			if (this.isActive || !this.hasInsertAcl) {
+			if (this.isActive || !mailboxHasRights(this.mailbox, 'i')) {
```
kesselb, line 382: I think hasInsertAcl would improve the readability here.

hamza221: on it

ChristophWurst (approved):
Clever code reuse and still readable. I like it :shipit:

PR #8112, opened 2023-02-24
chore(deps): Configure renovate commit titles

ChristophWurst:
* `Bump` instead of `update dependency`
* Include current version

Ref https://github.com/nextcloud/groupware/issues/63#issuecomment-1443055107

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,6 +14,9 @@
 		"dependencies",
 		"3. to review"
 	],
+	"commitMessageAction": "Bump",
+	"commitMessageTopic": "{{depName}}",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -14,6 +14,9 @@
 		"dependencies",
 		"3. to review"
 	],
+	"commitMessageAction": "Bump",
+	"commitMessageTopic": "{{depName}}",
+	"commitMessageExtra": "from {{currentVersion}} to {{#if isPinDigest}}{{{newDigestShort}}}{{else}}{{#if isMajor}}{{prettyNewMajor}}{{else}}{{#if isSingleVersion}}{{prettyNewVersion}}{{else}}{{#if newValue}}{{{newValue}}}{{else}}{{{newDigestShort}}}{{/if}}{{/if}}{{/if}}{{/if}}",
```
st3iny, line 19: I could imagine `currentVersion` to be undefined in some cases (pinned digests). We'll see how it behaves in practice. The official documentation does not elaborate further on the value of this variable.

st3iny (approved):
LGTM!

renovate.json
ChristophWurst: I took inspiration from https://github.com/DefectDojo/django-DefectDojo/blob/master/.github/renovate.json

ChristophWurst: Bildschirmfoto vom 2023-02-24 10-38-53

seems to work ok

PR #8170, opened 2023-03-01
Use NcRichText component from `@nextcloud/vue`

julien-nc:
Circular dependency is definitely going down with the next `@nextcloud/vue` release which will include `NcRichText` and everything needed for the link picker and the reference widgets.

refs https://github.com/nextcloud/nextcloud-vue/pull/3841 https://github.com/nextcloud/nextcloud-vue/issues/3812

I'll update `@nextcloud/vue` and set this to "ready for review" once the new release is out.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,6 @@
         "@nextcloud/router": "^2.0.1",
         "@nextcloud/vue": "~7.8.0",
         "@nextcloud/vue-dashboard": "^2",
-        "@nextcloud/vue-richtext": "^2.1.0-beta.5",
         "@riophae/vue-treeselect": "^0.4.0",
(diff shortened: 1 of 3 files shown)


julien-nc:
Ready IMO.

GretaD:
The Richtext change looks good, but the upgrade to nc/vue 7.8 have this problem

Screenshot from 2023-03-02 17-27-10

julien-nc:
@GretaD I can't reproduce. Tested on Firefox and Chromium. Any specific window size or element of context?
CTRL+F5? :grin: 

Rebased on main

GretaD (approved):
I cannot reproduce the error anymore

Issue #8157, opened 2023-02-28
Show sharing icon for mailboxes shared with me


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a recipient of mailbox shares I want to be able to visually distinguish between my own and shared mailboxes.

### Describe the solution you'd like

Show a sharing icon like we do in Files
 
![Bildschirmfoto vom 2023-02-28 10-37-31](https://user-images.githubusercontent.com/1374172/221813038-1242ba65-ce09-4b6a-a205-f674aa133557.png)

### Describe alternatives you've considered

N/a

### Additional context

https://github.com/nextcloud/mail/issues/7057

ChristophWurst:
If `myAcls` is not sufficient to distinguish between owned and shared mailbox, we need to fetch more ACL data: https://datatracker.ietf.org/doc/html/rfc4314.html#section-3

GretaD:
While brainstorming with Christoph we realized that:

1) not possible to distinguish based on my ACL rights
2) not possible to distinguish reliably based on mailbox path

PR #8171, opened 2023-03-01
feat(imap): Persist if mailbox is shared or not

ChristophWurst:
Groundwork for https://github.com/nextcloud/mail/issues/8157

### How to test

1) Add an account that has ACL support and receives mailboxes
2) Inspect oc_mail_mailboxes for that account
3) All received mailboxes have shared=1

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -12,7 +12,7 @@
 - **🙈 We’re not reinventing the wheel!** Based on the great [Horde](https://horde.org) libraries.
 - **📬 Want to host your own mail server?** We do not have to reimplement this as you could set up [Mail-in-a-Box](https://mailinabox.email)!
 	]]></description>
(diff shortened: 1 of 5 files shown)


GretaD:
i tested it couple of time and i still see 0 for the shared ones. Am i testing it wrong

did the steps written in this ticket,
also, created a new mailbox and shared it then opened the account that i shared this mailbox, i see the mailbox but shared column has 0

ChristophWurst:
Did you add the account anew or is it an account that existed before you checked out the branch? Mailboxes are cached for 2h

GretaD:
> Did you add the account anew or is it an account that existed before you checked out the branch? Mailboxes are cached for 2h

Yes, and i can see 1 now.

GretaD (approved):
tested and it works

PR #8180, opened 2023-03-01
Fix error when to mail is null

### LINKED ISSUE #8179: Reply all does not work when any to email is null
### Steps to reproduce

1. Click reply all on mail which has a to entry with null value as email

### Expected behavior

Reply mail modal opens and you can write a reply

### Actual behavior

Error in browser console and empty modal opens

### Mail app version

2.2.4

### Mailserver or service

Strato

### Operating system

Linux

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MySQL

### Additional info

TypeError: Cannot read properties of null (reading 'startsWith')
    at Composer.vue:594:1
    at Array.some (<anonymous>)
    at i.data (Composer.vue:594:1)
    at vue.runtime.esm.js:5419:21
    at vue.runtime.esm.js:5383:42
    at Rr (vue.runtime.esm.js:5324:9)
    at e._init (vue.runtime.esm.js:5646:9)
    at new i (vue.runtime.esm.js:5781:18)
    at vue.runtime.esm.js:4512:12
    at init (vue.runtime.esm.js:4374:52)

Used browser: Google Chrome 110.0.5481.177
Nextcloud server version: 25.0.4

IchbinkeinReh:
Fixes #8179

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -590,7 +590,7 @@ export default {
 			subjectVal: this.subject,
 			bodyVal: this.editorBody,
 			attachments: this.attachmentsData,
-			noReply: this.to.some((to) => to.email.startsWith('noreply@') || to.email.startsWith('no-reply@')),
+			noReply: this.to.some((to) => to.email?.startsWith('noreply@') || to.email?.startsWith('no-reply@')),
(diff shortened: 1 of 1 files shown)


st3iny (approved):
Can't reproduce but code looks good to me.

PR #8198, opened 2023-03-03
[stable3.0] chore(CI): Adjust testing matrix for Nextcloud 26

GretaD:


.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         php-versions: ['8.0', '8.1', '8.2']
-        nextcloud-versions: ['master']
+        nextcloud-versions: ['master', 'stable26']
```
ChristophWurst, line 13:
```suggestion
        nextcloud-versions: ['stable26']
```

ChristophWurst (changes requested):
master is 27 and doesn't work. Mail stable3.0 can only test against server stable26

ChristophWurst (approved):

.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         php-versions: ['8.0', '8.1', '8.2']
-        nextcloud-versions: ['master']
+        nextcloud-versions: ['stable26']
```
ChristophWurst, line 13: there are two more usages of `master` in this workflow file

ChristophWurst (changes requested):

ChristophWurst:
psalm worlkflow should also switch to dev-stable26

ChristophWurst (approved):

st3iny (approved):

PR #8206, opened 2023-03-06
Fix important and favorite position

### LINKED ISSUE #8192: Favorite/Important icon is not position properly on some threads
### Steps to reproduce

1. Open mail app (any size screen)
2. Open/Create a thread with different messages
3. Marks some of it as important and favorite

### Expected behavior

Favorite and important icon is positioned correctly on avatar

### Actual behavior

Favorite and important icon moves depending if the thread has a subject preview or a tag

### Mail app version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

Screenshot from 2023-03-03 11-05-04

GretaD:
fixes #8192

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -41,14 +41,14 @@
 				v-if="envelope.flags.flagged"
 				fill-color="#f9cf3d"
 				:size="18"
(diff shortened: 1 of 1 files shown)


GretaD:
@hamza221 I fixed all the scenarios but one, and that is when a thread has a tag. I cannot find a way how to fix that without breaking something else. Can you please have a look if you can see something that i dont. We have too many actions on the header so its not easy to add/remove/move stuff there.

PR #8254, opened 2023-03-21
chore(deps): Fine-tune renovate security bumps

ChristophWurst:
Check daily and open PR immediately


GretaD (approved):
cool cool cool cool

PR #8256, opened 2023-03-21
chore(deps): Use "fix" type for semantic vuln bumps

ChristophWurst:


st3iny (approved):
Can't find documentation on the `semanticCommitType` param but LGTM.

PR #8294, opened 2023-03-23
Show message timestamp in action menu

### LINKED ISSUE #8267: Show message timestamp in action menu

If I look at my inbox I see a relative time indicator for when a message was sent, but not always in a precise format. sometimes it is the time, sometimes just the day. As a user I want to see at which time a message was sent, regardless if it's recent old older.

Hovering the time doesn't give more info because then the time is replaced with the menu handle

Bildschirmfoto vom 2023-03-22 17-21-30

Include the timestamp in the actions menu like Talk does

Bildschirmfoto vom 2023-03-22 17-18-32

ref https://github.com/nextcloud/spreed/pull/9003

_No response_

### Additional context

@jancborchardt @nimishavijay yay or nay?

hamza221:
closes #8267
image

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -122,6 +122,16 @@
 				</ActionButton>
 			</EnvelopePrimaryActions>
 			<template v-if="!moreActionsOpen">
+				<ActionText>
+					<template #icon>
+						<ClockOutlineIcon
+							:size="20" />
+					</template>
+					{{
+						messageLongDate
+					}}
+				</ActionText>
+				<ActionSeparator />
 				<ActionButton v-if="hasWriteAcl"
 					:close-after-click="true"
 					@click.prevent="onToggleJunk">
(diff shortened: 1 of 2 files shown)


src/components/Envelope.vue
```diff
@@ -520,6 +524,9 @@ export default {
 		formatted() {
 			return shortRelativeDatetime(new Date(this.data.dateInt * 1000))
 		},
+		messageLongDate() {
```
ChristophWurst, line 527: make this a computed prop

ChristophWurst (commented):
Awesome! Let's wait for the approval of the designers

nimishavijay (changes requested):
Looks good! One suggestion: add a [clock icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Aschedule%3A) on the left of the item so that all the text is nicely aligned :)

src/util/shortRelativeDatetime.js
```diff
@@ -42,4 +42,8 @@ export const shortDatetime = curry((ref, date) => {
 	return momentDate.format('MMM D, YYYY')
 })
 
+export const messageDateTime = (date) => {
+	return moment(date * 1000).format(('MMMM D, YYYY H:mm'))
```
ChristophWurst, line 46: this is a tricky topic. I think the format should depend on the user's locale. Let's go with the format `lll`. See https://momentjs.com/ for examples.

nimishavijay (approved):
Nice! 🚀

PR #8297, opened 2023-03-24
Fix clickable area partially covering timestamp

goulvench:
## Before

Timestamp is partially covered by clickable area
Nextcloud Mail app - sender clickable area covering timestamp

## After

Timestamp can be hovered over
Nextcloud Mail app - sender clickable area fixed


GretaD:
I think that was intentional, the clickable area should have been the whole(not-actionable) header. And that includes the timestamp, but that means that you cannot hover the timestamp, so this fixup is good

PR #8311, opened 2023-03-28
fix no sent mailbox configured

### LINKED ISSUE #4153: "No sent mailbox configured" Error when sending a message
### Expected behavior
Tell us what should happen
After filling in the complete form to create a new mail account and saving without errors and syncing the inbox from the new account succussfully, I expect the app to send E-Mails without an error, assuming the remote IMAP server is available. The sent mail should occur in the sent folder.

### Actual behavior
Tell us what happens instead
Sending a message results in a 400 error response with the information: {"status":"fail","data":{"message":"No sent mailbox configured","type":"OCA\\Mail\\Exception\\SentMailboxNotSetException"}} (to be seen in the developer tools of chromium in the network section). The mail is not sent and not stored in the configured folder (INBOX.sent)

Sidenote: while writing a 400 error occurrs when the app tries to save the draft (No drafts mailbox configured...).

### Mail app

1.7.0

Remote hosted webserver (by onyxhosting) with postfix

### Server configuration

Ubuntu 20.04 LTS

Nginx 1.7

PostgreSQL 12.5

**PHP version:** (e.g. 7.0)
7.4

20.0.2

#### Client configuration
Chromium 87.0 

Arch Linux

### LINKED ISSUE #10546: Can't send mail using OVH SMTP
### Steps to reproduce

1. create a NC/Mail account hosted at OVH with automatic IMAP/SMTP settings
2. create a new mail using this mail account
3. send this new mail

### Expected behavior

The mail is sent to the recipient(s)

### Actual behavior

1st a popup  on the right says the mail is sent, then after a few seconds another pop up says it could not be sent.
The mail remains in the Outbox.

### Mail app version

3.7.9

### Nextcloud version

28.0.7 Enterprise

### Mailserver or service

OVH

### Operating system

Linux 5.15.0-122-generic x86_64

### PHP engine version

PHP 8.2

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

The NC instance is hosted by IONOS as a "Managed Nextcloud". I can't upgrade NC nor any app.

When I create another mail account hosted for instance at hostinger, everything works fine.

I receive mails delivered to this OVH mail account in NC/Mail.

This is not a matter of OVH mail password.

When I configure the exact same OVH email account in thunderbird, everything works fine.

I'm not sure the following error is related to this bug:
```
Exception
OCA\Mail\Controller\DraftsController::update(): Argument #2 ($accountId) must be of type int, null given, called in /var/www/html/lib/private/AppFramework/Http/Dispatcher.php on line 230 in file '/var/www/html/apps/mail/lib/Controller/DraftsController.php' line 157

```

hamza221:
Fix #4153 
Fixes https://github.com/nextcloud/mail/issues/10546

Repro: 

1. remove the default folder for sent ("you can do it in vuex by setting sentMailboxId to null")
2. Try to send a message 

Fix: 
Ps: Other ideas are appreciated 
1. Check if a mailbox named Sent or equivalent in the local language

- if it exists set it as Sent default folder and resend message, if not:

2. create a new mailbox 
3. set it as send default folder
4. resend the message 
5. Catch exceptions and inform the user with toasts accordingly


src/components/Composer.vue
```diff
@@ -1121,6 +1121,13 @@ export default {
 			this.saveDraftDebounced()
 		},
 		async onSend(_, force = false) {
+			const selectedAccount = this.$store.getters.getAccount(this.selectedAlias.id)
```
ChristophWurst, line 1124: does this work with aliases too? I think `getAccount` will only return an object if the id references an account (the main address), not an alias

hamza221: https://github.com/nextcloud/mail/blob/87c3ac8db07a365b358f71e8a644b584a68289d4/src/components/Composer.vue#L654-L658
selectedAlias.id here is always the account id

src/components/NewMessageModal.vue
```diff
@@ -128,6 +135,69 @@ export default {
 	methods: {
 		toHtml,
 		plain,
+		async onNewSentMailbox(data) {
+			const account = this.$store.getters.getAccount(data.accountId)
+			showWarning(t('mail', 'Setting Sent default folder'))
+			this.creatingSentMailbox = true
+
+			// check if sent mailbox already exists
+			let sentMailboxId = await getMailboxExists(data.accountId, account.personalNamespace + 'Sent').catch((error) => {
```
ChristophWurst, line 144: would it not be sufficient to search for the mailbox in the local data?

JohannesGGE:
Where exactly do you set sentMailboxId to null? (I can't get the error reproduced)

hamza221:
>Where exactly do you set sentMailboxId to null? (I can't get the error reproduced)
I do it in the vue.js devtools extension

src/components/NewMessageModal.vue
```diff
@@ -128,6 +134,60 @@ export default {
...
+			const mailboxes = this.$store.getters.getMailboxes(data.accountId)
+			const sentMailboxId = mailboxes.find((mailbox) => mailbox.name === account.personalNamespace + 'Sent' || mailbox.name === account.personalNamespace + t('mail', 'Sent'))?.databaseId
+			if (sentMailboxId) {
+				await this.setSentMailboxAndResend(account, sentMailboxId, data)
+				return
+			}
+			logger.info(`creating ${t('mail', 'Sent')} mailbox`)
+			await this.$store
+				.dispatch('createMailbox', { account, name: account.personalNamespace + t('mail', 'Sent') })
+				.then((e) => {
```
ChristophWurst, line 150: the method is async. you don't need .then. use await

ChristophWurst (commented):
The changes feel a bit too complex and messy. I think we should try to find a cleaner, less complicated spot to place the logic. I'll think about an alternative.

ChristophWurst:
> The changes feel a bit too complex and messy. I think we should try to find a cleaner, less complicated spot to place the logic. I'll think about an alternative.

How about we just move the logic into the head of NewMessageModal::onSend? We have access to the accountId and can fetch the account, check for the mailbox existence and act, then proceed with the usual send logic.

The good news is that this isn't too far off the current state. The code just has to be pushed around a bit.

I also think we don't necessarily need a special loading text for this state. For the user it's sufficient if we show *Sending …* while creating/assigning the mailbox

src/components/NewMessageModal.vue
```diff
@@ -294,6 +294,51 @@ export default {
 		},
 		toHtml,
 		plain,
+		async onNewSentMailbox(data, account) {
+			showWarning(t('mail', 'Setting Sent default folder'))
+			let newSentMailboxId = null
+			const mailboxes = this.mainStore.getMailboxes(data.accountId)
+			const sentMailboxId = mailboxes.find((mailbox) => mailbox.name === account.personalNamespace + 'Sent' || mailbox.name === account.personalNamespace + t('mail', 'Sent'))?.databaseId
```
ChristophWurst, line 301: nit. first look for the localized, then search for `Sent`. this ensures that the localized takes priority.

src/components/NewMessageModal.vue
```diff
@@ -294,6 +294,51 @@ export default {
...
+			if (sentMailboxId) {
+				await this.setSentMailboxAndResend(account, sentMailboxId, data)
```
ChristophWurst, line 303: I'm missing a return or anything else that leaves this method after setting the existing mailbox

src/components/NewMessageModal.vue
```diff
@@ -294,6 +294,51 @@ export default {
...
+			}
+			logger.info(`creating ${t('mail', 'Sent')} mailbox`)
+			try {
+				const newSentMailbox = await this.mainStore.createMailbox({ account, name: account.personalNamespace + t('mail', 'Sent') })
```
ChristophWurst, line 307: if the IMAP server supports it (likely) we should also set the specialuse attribute to `\sent`

ChristophWurst (commented):
The logical flow looks good otherwise

lib/Controller/MailboxesController.php
```diff
@@ -246,10 +246,10 @@ public function update() {
 	 * @throws ClientException
 	 */
 	#[TrapError]
-	public function create(int $accountId, string $name): JSONResponse {
+	public function create(int $accountId, string $name, bool $isSentMailbox = false): JSONResponse {
```
ChristophWurst, line 249: This makes the new parameter only usable for specialuse=sent. Change it to allow a generic specialuse attribute. Then we will be able to also create other special mailboxes.

appinfo/routes.php
```diff
@@ -155,6 +155,11 @@
 			'url' => '/api/mailboxes/{id}',
 			'verb' => 'PATCH'
 		],
+		[
+			'name' => 'mailboxes#setImapSpecialUseAtrribute',
```
ChristophWurst, line 159: dead route

ChristophWurst (changes requested):
No test coverage

hamza221:
Creating a new mailbox is not tested, not sure how to without an account

lib/IMAP/FolderMapper.php
```diff
@@ -76,8 +86,17 @@ public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
...
+		array $specialUseAttributes = []): Folder {
+		$opts = [];
+	
+		foreach ($specialUseAttributes as $attribute) {
+			if (!in_array($attribute, self::SUPPORTED_SPECIAL_USE_ATTRIBUTES, true)) {
+				throw new ServiceException('Unsupported special use attribute: ' . $attribute);
+			}
+			$opts[] = $attribute;
+		}
+		$client->createMailbox($name, $opts);
```
kesselb, line 99: createMailbox expects an array "opts" with a key "special_use" and there the list of attributes.

lib/IMAP/FolderMapper.php
```diff
@@ -76,8 +86,17 @@ public function getFolders(Account $account, Horde_Imap_Client_Socket $client,
...
 		Account $account,
-		string $name): Folder {
-		$client->createMailbox($name);
+		string $name,
+		array $specialUseAttributes = []): Folder {
+		$opts = [];
+	
+		foreach ($specialUseAttributes as $attribute) {
+			if (!in_array($attribute, self::SUPPORTED_SPECIAL_USE_ATTRIBUTES, true)) {
+				throw new ServiceException('Unsupported special use attribute: ' . $attribute);
```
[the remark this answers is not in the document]
ChristophWurst: It's sensible validation. You could pull this up into the HTTP layer and do the check in the controller. Return a BAD REQUEST or UNPROCESSIBLE ENTITY.

lib/Controller/MailboxesController.php
```diff
@@ -248,8 +248,12 @@ public function update() {
 	#[TrapError]
 	public function create(int $accountId, string $name, array $specialUseAttributes = []): JSONResponse {
 		$account = $this->accountService->find($this->currentUserId, $accountId);
-
-		return new JSONResponse($this->mailManager->createMailbox($account, $name, $specialUseAttributes));
+		try {
+			$mailbox = $this->mailManager->createMailbox($account, $name, $specialUseAttributes);
+		} catch (ServiceException $e) {
+			return new JSONResponse([], Http::STATUS_BAD_REQUEST);
```
ChristophWurst, line 254: This will now catch *any* service exception and return a nondescriptive  HTTP response. This would be painful to debug.

Do a simple in_array check comparing the sent special use to the allowed ones. Then carry on with the string and add type hints for the allowed values for the deeper service methods.

ChristophWurst:
Works, but not as expected.

Bildschirmfoto vom 2025-05-16 13-28-28

I see this chain of toasts, so the logic kicks in.

Bildschirmfoto vom 2025-05-16 13-29-15

But my sent mailbox is now set to *nullSent*. Which is not just a UI glitch, it created this mailbox:

image

The creation shouldn't even happen, because I do have a mailbox named **Sent** and it does have the **\\sent** attribute:

image

src/components/NewMessageModal.vue
```diff
@@ -286,6 +286,51 @@ export default {
...
+			showWarning(t('mail', 'Setting Sent default folder'))
+			let newSentMailboxId = null
+			const mailboxes = this.mainStore.getMailboxes(data.accountId)
+			const sentMailboxId = mailboxes.find((mailbox) => mailbox.name === account.personalNamespace + 'Sent' || mailbox.name === account.personalNamespace + t('mail', 'Sent'))?.databaseId
+			if (sentMailboxId) {
+				await this.setSentMailboxAndResend(account, sentMailboxId, data)
+			}
+			logger.info(`creating ${t('mail', 'Sent')} mailbox`)
+			try {
+				const newSentMailbox = await this.mainStore.createMailbox({ account, name: account.personalNamespace + t('mail', 'Sent'), specialUseAttributes: ['\\Sent'] })
```
ChristophWurst, line 299: I guess the ``nullSent`` happens when there is no personalNamespace. In that case, don't prepend anything

src/components/NewMessageModal.vue
ChristophWurst: > hamza221 marked this conversation as resolved.

Isn't this precisely the reason why I'm running into errors despite being able to send the message?

To have constructive and comprehensive collaboration please be more elaborative. If you close my comment as resolved silently I assume it's because you have addressed it.

hamza221:
> It fails to detect my existing mailbox _Sent_ and runs into a HTTP400. The toasts are honestly overwhelming and confusing.
> 
> First I get this Bildschirmfoto vom 2025-05-27 09-52-22
> 
> Bildschirmfoto vom 2025-05-27 09-52-25
> 
> image
> 
> The message _is sent and stored in my mailbox called "Sent"_, nevertheless?!


Toasts were all over the place, 
It should make sense now, I added early returns and removed extra Toasts

lib/Contracts/IMailManager.php
```diff
@@ -45,12 +45,13 @@ public function getMailboxes(Account $account): array;
 	/**
 	 * @param Account $account
 	 * @param string $name
+	 * @param array<Horde_Imap_Client::SPECIAL_USE_*> $specialUseAttributes
```
ChristophWurst, line 48: ^ see Psalm. Missing an import

src/components/NewMessageModal.vue
```diff
@@ -286,6 +286,57 @@ export default {
...
+				} catch (error) {
+					logger.error('could not set sent mailbox', { error })
+					showError(t('mail', 'Couldn\'t set sent default folder, please try manually before sending a new message'))
+				}
+				return
+
+			}
+			logger.info(`creating ${t('mail', 'Sent')} mailbox`)
+			try {
+				const newSentMailbox = await this.mainStore.createMailbox({ account, name: (account.personalNamespace ?? '') + t('mail', 'Sent'), specialUseAttributes: ['\\Sent'] })
```
ChristophWurst, line 309: nit: extract new mailbox name into a variable for code readability and to have the actual mailbox name logged, not just the translation

src/components/NewMessageModal.vue
```diff
@@ -286,6 +286,57 @@ export default {
...
+		async onNewSentMailbox(data, account) {
+			showWarning(t('mail', 'Setting Sent default folder...'))
+			let newSentMailboxId = null
+			const mailboxes = this.mainStore.getMailboxes(data.accountId)
+			const sentMailboxId = mailboxes.find((mailbox) => (mailbox.name === (account.personalNamespace ?? '') + 'Sent') || (mailbox.name === (account.personalNamespace ?? '') + t('mail', 'Sent')))?.databaseId
+			if (sentMailboxId) {
+				try {
+					await this.setSentMailboxAndResend(account, sentMailboxId, data)
+					showSuccess(t('mail', 'Default sent folder set'))
+					this.onSend(data)
```
ChristophWurst, line 298: Missing `await`

src/components/NewMessageModal.vue
```diff
@@ -286,6 +286,57 @@ export default {
...
+			} catch (error) {
+				showError(t('mail', 'Could not create new mailbox, please try setting a sent mailbox manually'))
+				logger.error('could not create mailbox', { error })
+				this.$emit('close')
+				return
+			}
+
+			try {
+				await this.setSentMailboxAndResend(account, newSentMailboxId, data)
+				this.onSend(data)
```
ChristophWurst, line 322: missing `await`

src/components/NewMessageModal.vue
```diff
@@ -385,6 +436,11 @@ export default {
 				.catch((error) => logger.error('could not upload attachments', { error }))
 		},
 		async onSend(data, force = false) {
+			const account = this.mainStore.getAccount(data.accountId)
+			if (!account?.sentMailboxId) {
+				this.onNewSentMailbox(data, account)
```
ChristophWurst, line 441: missing `await`

src/components/NewMessageModal.vue
```diff
@@ -286,6 +286,57 @@ export default {
...
 			this.additionalTrapElements.push(element)
 		},
+		async onNewSentMailbox(data, account) {
+			showWarning(t('mail', 'Setting Sent default folder...'))
+			let newSentMailboxId = null
+			const mailboxes = this.mainStore.getMailboxes(data.accountId)
+			const sentMailboxId = mailboxes.find((mailbox) => (mailbox.name === (account.personalNamespace ?? '') + 'Sent') || (mailbox.name === (account.personalNamespace ?? '') + t('mail', 'Sent')))?.databaseId
+			if (sentMailboxId) {
+				try {
+					await this.setSentMailboxAndResend(account, sentMailboxId, data)
```
ChristophWurst, line 296: the method takes to params. `data` is superfluous

src/components/NewMessageModal.vue
ChristophWurst, line 309: `createMailbox` prepends the namespace too. This can lead to double namespacing.

ChristophWurst:
Superseded by https://github.com/nextcloud/mail/pull/12555

PR #8327, opened 2023-04-03
mb_convert_encoding does not support windows-1250. #1037

Astinus-Eberhard:


lib/Model/IMAPMessage.php
```diff
@@ -658,7 +658,11 @@ private function loadBodyData(Horde_Mime_Part $p, $partNo): string {
 		// data is utf-8 by default.
 		$charset = $p->getContentTypeParameter('charset');
 		if ($charset !== null && strtoupper($charset) !== 'UTF-8') {
-			$data = mb_convert_encoding($data, 'UTF-8', $charset);
+			try {
+				$data = mb_convert_encoding($data, 'UTF-8', $charset);
+			} catch (\Throwable $ex) {
+				$data = iconv($charset, 'UTF-8//TRANSLIT', $data);
```
ChristophWurst, line 664: is translit generally available or does it depend on the compilation options of iconv? we had UTF8//IGNORE in the past and it was troublesome

Astinus-Eberhard: No idea, but used a lot in so many projects & environments and I haven't spotted a single issue. I found both `//TRANSLIT` and `//IGNORE` options everywhere I read anything about Iconv. Honestly I've never even think this can failed (at least with clear from and to encodings and without autodetections).

There was some problems in some Iconv implementations, for example on some docker alphine images, but IMHO nothing so unusual & there was some fixes for it (mainly updates binaries from edges).

This PR is as a suggestion, if someone has a better experience with character encoding then please review and replace.

ChristophWurst: https://github.com/nextcloud/mail/issues/3908 was the older thread about //IGNORE issues.

ririko5834:
When it willl be merged

lib/Model/IMAPMessage.php
Astinus-Eberhard: Honestly I see no risk here, cause this change does even matter only if there was an exception thrown, which is currently not handled properly, preventing some users from reading emails with corrupted encodings, which IMHO is more annoying than potential risk of misleading characters replacement. If under any environment this will fail, then again, now it is failing anyway too, so nothing change really.

ChristophWurst: The risk is that there are possibly environments that currently work and will break with this change.

Could you check if Alpine's iconv handles //TRANSLIT without any issues?

https://github.com/nextcloud/mail/issues/3908#issuecomment-762377816 for the specific issues we had earlier.

If translit works with all iconv implementations I'm okay with this change :+1: 

Thanks

ririko5834:
Does anyone think that this PR can fix my issue https://github.com/anonaddy/anonaddy/issues/485

ChristophWurst:
Closing due to lack of clarity regarding the support of //TRANSLIT across platforms https://github.com/nextcloud/mail/pull/8327#discussion_r1289715428

PR #8336, opened 2023-04-05
fix: colors button dark-theme #3231

### LINKED ISSUE #3231: Further dark theme fixes
These are left over – the primary button icons should always be the same as the text color, which depends on theming color, not light/dark theme. 
- [x] Reply button icon
  
- [x] Send button icon
  
- [x] "Add mail account" button in settings
   

~~Also possibly something to look into is this menu, which might have an integrated dark theme?~~ https://github.com/nextcloud/mail/issues/6378

JohannesGGE:
Fix #3231 

Change color of text and icon to match with other buttons.

| Before | After |
| --- | --- |
| image |image|

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -270,7 +270,7 @@ p.app-settings {
 	margin-top: 6px;
 }
 .app-settings-button.button.primary.new-button {
-	color: var(--color-main-background);
+	color: var(--color-primary-text);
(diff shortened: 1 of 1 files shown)


ChristophWurst:
To ease reviews of visual changes we post before/after screenshots. You can see an example in https://github.com/nextcloud/mail/pull/8069.

ChristophWurst (approved):

Issue #8328, opened 2023-04-04
Sort tags alphabetically


maximelehericy:
### Is your feature request related to a problem? Please describe.

I easily have 70-100 tags today.
Unfortunately, in Nextcloud Mail, tags aren't sorted alphabetically, which makes quite hard to find the right one in the list.

### Describe the solution you'd like

Display tags sorted alphabetically ?

### Describe alternatives you've considered

Use find in page, but it isn't really user friendly.

### Additional context

I am trying to switch from TB to Nextcloud Mail thanks to the great work of you guys :)

I basically create one tag per prospect/customer. Then, whoever is the sender and recipient (colleague, partner, supplier, customer), I can easily gather all emails related to one prospect/customer.

It is the best approach I've found to get complete history about one prospect/customer without spending time in the search trying to remember what keywords or what names were used in the email thread.

PR #8350, opened 2023-04-11
enhancement: sort tags alphabetically

JohannesGGE:
Fixes: #8328

Sort tags alphabetically (after default and used tags)

### DIFF
--- a/src/components/SearchMessages.vue
+++ b/src/components/SearchMessages.vue
@@ -290,6 +290,7 @@ import Magnify from 'vue-material-design-icons/Magnify'
 import debouncePromise from 'debounce-promise'
 import { findRecipient } from '../service/AutocompleteService'
 import uniqBy from 'lodash/fp/uniqBy'
+import { hiddenTags } from './tags'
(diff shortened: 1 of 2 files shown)


hamza221 (commented):
Sorting tags in message search can also be useful here, if I got the issue right 
https://github.com/nextcloud/mail/blob/c03ee464bd7c7c9fd14fb28c5d5e2a3e43a54cdd/src/components/SearchMessages.vue#L337-L339

ChristophWurst:
Makes sense to sort them consistently everywhere

JohannesGGE:
:heavy_check_mark:   Implemented it for the search as well

hamza221:
Test and lgtm
One small thing tho, The default tags are on top only in the Tag Modal. Is that the expected behavior?
**Tag Modal:**
image
**Message search:**
image

ChristophWurst:
@hamza221 could you check another time?

hamza221 (approved):
Tested and works

Issue #8354, opened 2023-04-12
Button "Save sieve script" looks disabled


ChristophWurst:
### Steps to reproduce

1. Add an account
2. Open the account settings
3. Enable filtering (sieve)
4. Enter a filter script
5. Look at the save button

### Expected behavior

It looks active

### Actual behavior

It looks disabled

| Default | Hover |
|--------|--------|
| ![Bildschirmfoto vom 2023-04-12 09-15-18](https://user-images.githubusercontent.com/1374172/231381034-f7a3ba30-4091-4007-b685-bdf27ce8d7b0.png) | ![Bildschirmfoto vom 2023-04-12 09-15-21](https://user-images.githubusercontent.com/1374172/231381026-2a41c460-5c32-400f-9254-65517312075a.png) |

### Mail app version

3.1

### PHP engine version

None

### Web server

None

### Database

None

PR #8355, opened 2023-04-12
fix: style of "Save sieve script" button

JohannesGGE:
Fixes: #8354

|before|after|
|-|-|
| image |image |


ChristophWurst (approved):
Nice

Issue #6038, opened 2022-02-14
Sync failed: IMAP error reported by server / IMAP command too long


solracsf:
### Steps to reproduce

1. Setup a mailbox
2. Run `occ mail:account:sync`

### Expected behavior

Mailbox is synced

### Actual behavior

```
[debug] account is up to date, skipping mailbox sync
[debug] Skipping mailbox sync for 3
[debug] Skipping mailbox sync for 12
[debug] Skipping mailbox sync for 15
[debug] Skipping mailbox sync for 18
[debug] Skipping mailbox sync for 21
[debug] Skipping mailbox sync for 24
[debug] Skipping mailbox sync for 27
[debug] Skipping mailbox sync for 30
[debug] Skipping mailbox sync for 33
[debug] Skipping mailbox sync for 36
[debug] Skipping mailbox sync for 54
[debug] Skipping mailbox sync for 57
[debug] Syncing 6
[debug] Locking mailbox 6 for new messages sync
[debug] Locking mailbox 6 for changed messages sync
[debug] Locking mailbox 6 for vanished messages sync
[debug] Running partial sync for 6
[debug] partial sync 3:INBOX - get all known UIDs took 0s
[debug] partial sync 3:INBOX - get new messages via Horde took 1s
[debug] partial sync 3:INBOX - classified a chunk of new messages took 1s
[debug] partial sync 3:INBOX - persist new messages took 0s
[debug] partial sync 3:INBOX - get changed messages via Horde took 188s
[debug] partial sync 3:INBOX - persist changed messages took 98s
[debug] Unlocking mailbox 6 from vanished messages sync
[debug] Unlocking mailbox 6 from changed messages sync
[debug] Unlocking mailbox 6 from new messages sync

In ImapToDbSynchronizer.php line 249:

  [OCA\Mail\Exception\ServiceException]
  Sync failed for 3:INBOX: IMAP error reported by server.

Exception trace:
  at /apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php:249
 OCA\Mail\Service\Sync\ImapToDbSynchronizer->sync() at /apps/mail/lib/Service/Sync/ImapToDbSynchronizer.php:135
 OCA\Mail\Service\Sync\ImapToDbSynchronizer->syncAccount() at /apps/mail/lib/Command/SyncAccount.php:110
 OCA\Mail\Command\SyncAccount->sync() at /apps/mail/lib/Command/SyncAccount.php:94
[...]

PR #8367, opened 2023-04-12
fix(imap): Chunk UIDs by string length

ChristophWurst:
IMAP sync requires an IMAP FETCH command that fetches specific messages. If the full mailbox has to be sync'ed, that can be a long list of UIDs.

The synchronizer had already broken UIDs into chunks by a maximum chunk size. Yet the chunk size does not correlate with the resulting IMAP FETCH command length.

An advanced chunk function calculates the UID range string and decreases the chunk size until the UIDs produce a string that fits the given byte size.

Fixes https://github.com/nextcloud/mail/issues/6038


miaulalala (approved):
Looks gut, better than doing the highest and lowest only. Tested, works, although I'm not sure I could reproduce with my mailbox.

ChristophWurst:
Bildschirmfoto vom 2023-04-12 19-42-31

there is also no significant performance regression

Issue #6127, opened 2022-03-14
Convert Email to Task


thstyl2000:
Detailed feature request taken from https://github.com/nextcloud/mail/issues/6678.

## Is your feature request related to a problem? Please describe.

I love that we already have the ability to create an event from an e-mail, but what would be even more useful to me would be to be able to create tasks with an optional due date from an e-mail. I often create myself a todo list to manage incoming e-mails, so this feature could speed up that work.

## Describe the solution you'd like

The flow could be almost identical to creating events:

1. Click on the 3 dots menu on a specified e-mail and go into the "More actions" sub-menu. There you would find a new button below "Create event" such as "Create task"
![image](https://user-images.githubusercontent.com/36501744/172541295-5869f520-49a4-44ca-b60d-48f3e3d8f8f1.png)

2.  Click on "Create task". This would open a modal similar to the "Create event" one, with some or all of the following options:
    - Name (defaults to e-mail title)
    - Start date (defaults to now)
    - Due date (defaults to now + 1 hour for example)
    - "All day" checkbox (similar to events, turns the two previous datetime selections into date only selections)
    - Calendar selection (to which calendar should the task be attached)
    - Optional: Priority ?
    - Optional: Status ?
    - Optional: Tags ?
    - Optional: Notes (defaults to e-mail body)
Here is the "Create event" modal as a reference:
![image](https://user-images.githubusercontent.com/36501744/172542515-fe38c87f-c90c-4c05-b7da-3ab5d685dc27.png)

## Acceptance criteria

* Optional: put *Create task* into existing *Create event* action and modal
  * Use https://github.com/nextcloud/nextcloud-vue/pull/3945 to switch between event and task

## Work packages

- [ ] Main implementation that fulfill acceptance criteria
- [ ] https://github.com/nextcloud/documentation/issues/10016

miaulalala:
Nice idea!

ufgum:
Another benefit could be to "create a deck card" subject as card name, body as attachment or as description. Also to add to an existing card. With **task or deck**, it would be possible to  handle projects with the mail APP.

olgru:
I'm looking for a way to document requests somehow. Most are reaching me by mail, most got respond by mail too. But I don't want my mail client to be the documentation system. Would be much nicer to have this here as other related team activities are also within NC. 
Si I support this request... ;-)

cydnax:
Hi. Any update on this yet?

PR #8375, opened 2023-04-13
Create a task from a message

hamza221:
closes [#6127](https://github.com/nextcloud/mail/issues/6127)

- [x] Move task.js and calendar.js to  [calendar-js](https://github.com/nextcloud/calendar-js) ?
- [x] ~~Wait for https://github.com/nextcloud/nextcloud-vue/pull/3945 to get merged and add the feature in Create event modal~~ follow-up
- [ ] Fix warns
image

### DIFF
--- a/src/components/TaskModal.vue
+++ b/src/components/TaskModal.vue
@@ -1,0 +1,243 @@
+<template>
+	<Modal @close="onClose">
+		<div class="modal-content">
+			<h2>{{ t('mail', 'Create task') }}</h2>
+			<div class="taskTitle">
+				<input v-model="taskTitle" type="text">
+			</div>
+			<div class="all-day">
+				<DatetimePicker v-model="startDate"
+					:format="dateFormat"
+					:clearable="false"
+					:minute-step="5"
+					:show-second="false"
+					:type="datePickerType"
+					:show-timezone-select="true"
+					:timezone-id="startTimezoneId" />
+				<DatetimePicker v-model="endDate"
+					:format="dateFormat"
+					:clearable="false"
+					:minute-step="5"
+					:show-second="false"
+					:type="datePickerType"
+					:show-timezone-select="true"
+					:timezone-id="endTimezoneId" />
+			</div>
+			<label for="note">Description:</label>
+			<textarea id="note" v-model="note" rows="7" />
+			<div class="all-day">
+				<input
+					id="allDay"
+					v-model="isAllDay"
+					type="checkbox"
+					class="checkbox">
+				<label
+					for="allDay">
+					{{ t('mail', 'All day') }}
+				</label>
+			</div>
+			<Multiselect
+				v-model="selectedCalendar"
+				label="displayname"
+				track-by="url"
+				:placeholder="t('mail', 'Select calendar')"
+				:allow-empty="false"
+				:options="calendars">
+				<template #option="{option}">
+					<CalendarPickerOption
+						v-bind="option" />
+				</template>
+				<template #singleLabel="{option}">
+					<CalendarPickerOption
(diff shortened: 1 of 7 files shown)


hamza221:
@ChristophWurst  lmk if you think task.js and calendar.js should be moved [calendar-js](https://github.com/nextcloud/calendar-js) when you find time to review it

src/store/calendar.js
```diff
@@ -0,0 +1,94 @@
+/**
+ * Maps a dav collection to our calendar object model
```
ChristophWurst, line 2: *our* model is the one used in Calendar, right? unsure if we want to push that into calendar-js. @st3iny what do you think?

ChristophWurst: we will leave it as-is

ChristophWurst:
conflicts

ChristophWurst (approved):
Code looks good. Didn't test

ChristophWurst:
eslint needs a fix

src/task.js
```diff
@@ -0,0 +1,719 @@
...
+
+		this.syncStatus = null
+
+		// Time in seconds before the task is going to be deleted
+		this.deleteCountdown = null
+
+		// Queue for update requests with concurrency 1,
+		// because we only want to allow one request at a time
+		// (otherwise we will run into problems with changed ETags).
+		this.updateQueue = new PQueue({ concurrency: 1 })
```
ChristophWurst, line 72: never used -> drop pqueue

JohannesGGE:
Screenshot from 2023-05-03 12-07-47_
I was a bit confused at first because I used the action-menu on the right and couldn't find the option. Maybe it should be added there too?

JohannesGGE:
image
I noticed some things:
- The headline is "Create event", is that intended?
- I cannot select the calendar to which the task should be added (list is empty)
- Unlike the create event, it is not clear that the datetimes are "Start date" and "Due date" (at least for me) (most of the time I personally set only a due date)
- "Tell us your story:" is the description field? If yes, I would just write "Description" to prevent confusion.

ChristophWurst:
*Create* should be disabled if you have no calendars with VTODO support, I guess

hamza221:
> _Create_ should be disabled if you have no calendars with VTODO support, I guess

on it

ChristophWurst:
@JohannesGGE please re-review :)

jancborchardt (commented):
- [ ] The height of the description field could be increased so that it’s nicer to type in there, and that the "Create" button doesn’t float in the air but is exactly at the bottom.

src/components/TaskModal.vue
```diff
@@ -0,0 +1,242 @@
...
+				</template>
+				<template #singleLabel="{option}">
+					<CalendarPickerOption
+						:display-icon="option.displayIcon"
+						v-bind="option" />
+				</template>
+			</Multiselect>
+
+			<br>
+			<button class="primary" :disabled="saving" @click="onSave">
```
JohannesGGE, line 57:
```suggestion
			<button class="primary" :disabled="disabled" @click="onSave">
```

JohannesGGE (commented):
I would format the date+time (all day disabled) like the "create event" does it. Right now there is just the time shown.
Screenshot from 2023-05-04 12-02-17
image

ChristophWurst:
Nit: *Select option* -> *Select calendar*
Nit: *List is empty* -> *No calendars with task list support*

hamza221:
image

ChristophWurst (approved):
Works

JohannesGGE (approved):

ChristophWurst:
Please squash

Issue #7884, opened 2023-01-10
Show senders e-mail adress


rollanders:
### Is your feature request related to a problem? Please describe.

I receive a mail. I only see the name of the sender. I have not found anyway to see his e-mail-adress. 

### Describe the solution you'd like

You should be able to show the adresss.

rollanders:
I can find it by klicking on the three dots and then choose Show source.

ChristophWurst:
Yes, but it should also show when the sender name is clicked

rollanders:
But it does not. I updated the mail-app. Same situation.

ChristophWurst:
That is what I meant. You currently can't see the email. But it would be desirable if we could.

ufgum:
Don't forget the recipient address.
The recipient is also important if you use alias addresses. You see only you're name, not the address this email was sent to.
I would go for "mouseover title tag" the sender AND recipient names.

JakobGode:
In my situation i click on the sendes name and the sender adress is shown, but completly covered by an bar.
milliseconds after clicking the senders adress is shown and then the bar pops up.
![sender adress coverd](https://user-images.githubusercontent.com/61253957/213698459-7976d3a3-2d21-47e1-9042-af0abecad258.PNG)

ufgum:
> In my situation i click on the senders name and the sender adress is shown, but completly covered by an bar. Milliseconds after clicking the senders address is shown and then the bar pops up.

And how to see the email in Firefox and Chromium:
1. click the name, wait the bar covers the email-address
2. click on the bar (between the links)
3. click again the name and the email-address is shown

It works, but a bit complicated. Therefore I would prefer the `title` tag, it pop's up without clicking.

JakobGode:
Works, but uncomfortable. And if there are more adresses in the conversation, the mailaderss i missing

koying:
> And how to see the email in Firefox and Chromium

Oh wow. Thanks for this, but extremely cumbersome...

KimmoJ:
Obfuscating any sender info or making it something you have to click through to find is just wrong. Who it is from is a crucial bit of information before you know how to react to an email. In fact, remove the display name if need be and leave only the emal address, even that would be more correct. 

[...]

PR #8376, opened 2023-04-14
Display email address 

hamza221:
Before:
image
After:
image

Fixes https://github.com/nextcloud/mail/issues/8379

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -52,6 +52,7 @@
 				class="app-content-list-item-star junk-icon-style"
 				:data-starred="envelope.flags.$junk ? 'true' : 'false'"
 				@click.prevent="hasWriteAcl ? onToggleJunk() : false" />
+
 			<router-link
 				:to="route"
 				event=""
@@ -60,6 +61,9 @@
 				@click.native.prevent="$emit('toggle-expand', $event)">
 				<div class="sender">
 					{{ envelope.from && envelope.from[0] ? envelope.from[0].label : '' }}
+					<p class="sender__email">
+						{{ envelope.from && envelope.from[0] ? envelope.from[0].email : '' }}
+					</p>
 				</div>
 				<div v-if="hasChangedSubject" class="subline">
 					{{ cleanSubject }}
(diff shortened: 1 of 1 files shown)


mat-m:
I would prefer to keep the labels and have the mail address on hovering.

ChristophWurst:
@hamza221 please create a ticket for this change so that we can discuss the visual changes with designers.

This change originates from https://github.com/nextcloud/mail/issues/7884.

src/components/ThreadEnvelope.vue
```diff
@@ -60,6 +61,9 @@
 				@click.native.prevent="$emit('toggle-expand', $event)">
 				<div class="sender">
 					{{ envelope.from && envelope.from[0] ? envelope.from[0].label : '' }}
+					<p class="sender-email">
```
ChristophWurst, line 64:
```suggestion
					<p class="envelope__header__sender-email">
```

for the BEM convention

src/components/ThreadEnvelope.vue
```diff
@@ -652,6 +660,7 @@ export default {
 			padding: 10px;
 			border-radius: var(--border-radius);
 			min-height: 68px; /* prevents jumping between open/collapsed */
+
```
ChristophWurst, line 663: unrelated :)

ChristophWurst (commented):
Code looks good

Please update the screenshots in the PR description

hamza221:
image
Does this look fine? there's no real visual distinction between the address and subject except the mail format

ChristophWurst:
> Does this look fine? there's no real visual distinction between the address and subject except the mail format
> Also is the order fine ? Label->email->subject?

Looks good to me :)

src/components/ThreadEnvelope.vue
```diff
@@ -60,6 +61,9 @@
 				@click.native.prevent="$emit('toggle-expand', $event)">
 				<div class="sender">
 					{{ envelope.from && envelope.from[0] ? envelope.from[0].label : '' }}
+					<p class="sender__email">
+						{{ envelope.from[0].email }}
```
ChristophWurst, line 65: like on L63 we should save-guard a possibly missing sender. messages might not have a `from` address

ChristophWurst (approved):
:shipit:

PR #8404, opened 2023-04-24
ci: Only run integration tests if unit tests pass

ChristophWurst:
Unit tests are fast. Integration tests are heavy. We can align the two to stop PRs that won't get merged anyway.

Caveat: serializes test execution.

Ref https://docs.github.com/en/actions/using-workflows/about-workflows#creating-dependent-jobs


ChristophWurst:
Unsure if this is a good idea or not. Closing for now.

PR #8405, opened 2023-04-25
chore: format js code with eslint

ChristophWurst:
This is the result of running `npm run lint:fix` on `main`.


st3iny (approved):
:broom:

PR #8418, opened 2023-04-27
perf(frontend): Load NewMessageModal async

ChristophWurst:
| Before | After |
|--------|--------|
| Bildschirmfoto vom 2023-04-27 12-00-12 | Bildschirmfoto vom 2023-04-27 12-07-22 | 

It's not a lot but it's significant enough for a page load. The component is fetched when *New message* is clicked.


st3iny (approved):
Tested briefly and works. Epic!

I didn't know that async handling is so convenient to implement. I expected it to be way more complicated.

Issue #5387, opened 2021-07-31
One Click Suggested Unsubscribe


ChildLearningClub:
## Feature Request

Easily unsubscribe to promotion emails by grouping them into a list and allowing for simple one click unsubscribe.

[Edison email](https://mail.edison.tech/) as well as apparently CleanFox have a feature to quickly unsubscribe to newsletters and promotional emails.

## Specification

* https://datatracker.ietf.org/doc/html/rfc2369
* https://datatracker.ietf.org/doc/html/rfc8058

## Work packages

- [x] https://datatracker.ietf.org/doc/html/rfc2369 https://github.com/nextcloud/mail/pull/8395
- [x] https://datatracker.ietf.org/doc/html/rfc2369 https://github.com/nextcloud/mail/pull/8434
- [ ] https://datatracker.ietf.org/doc/html/rfc8058 https://github.com/nextcloud/mail/pull/8438
- [x] https://github.com/nextcloud/documentation/issues/10015

ChristophWurst:
I assume they get that info from some headers? Would you have an example email where that quick unsubscribe is shown so we can inspect the header info?

philippstappert:
there is a `list-unsubscribe` header in most newsletters/mailing lists with a link to unsubscribe.

See RFC2369: https://datatracker.ietf.org/doc/html/rfc2369

nimishavijay:
This is there in Gmail, Apple mail and Outlook so it sounds like a great idea!
As for how it looks on the frontend, something like this could work:
![image](https://user-images.githubusercontent.com/52440189/156173003-dfcb1685-ef2a-4149-8da8-887fa6489131.png)

Screenshots for other email clients are at [Penpot](https://design.penpot.app/#/workspace/464ebb90-b243-11eb-95bf-bcf80c3b5feb/c337d4c0-92fd-11ec-bd38-efdb6fa63305?page-id=e8226c50-9956-11ec-824f-49723f33a3ff&layout=assets)
cc @jancborchardt also

ChristophWurst:
> there is a `list-unsubscribe` header in most newsletters/mailing lists with a link to unsubscribe.
> 
> See RFC2369: https://datatracker.ietf.org/doc/html/rfc2369

[...]

Issue #8191, opened 2023-03-03
Primary action menu overlaps with recipient on small screen


GretaD:
### Steps to reproduce

1. Open mail on small screen
2. Look at one thread header
3.

### Expected behavior

Nice design

### Actual behavior

I cannot read the recipient

### Mail app version

3.0.0-beta

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

![Screenshot from 2023-03-03 09-57-49](https://user-images.githubusercontent.com/12728974/222690523-41ce5d4b-67a2-4737-992b-bb01c0605526.png)

GretaD:
@jancborchardt @nimishavijay @ChristophWurst this is a space issue. How should we solve it design wise?

nimishavijay:
For sure a space issue. As the screen width decreases the items of the primary actually menu could go into the 3 dot menu, starting from the rightmost action (so as the screen size decreases, the Delete action would be the first to go inside the 3 dot menu). 

Additionally, since Reply is the most important action here we could show that outside the 3 dot menu always. What do you think?

ChristophWurst:
More examples

![Bildschirmfoto vom 2023-03-21 09-27-49](https://user-images.githubusercontent.com/1374172/226557417-778226f3-901a-4d7b-b56b-391a8074d7e5.png)
![Bildschirmfoto vom 2023-03-21 09-28-56](https://user-images.githubusercontent.com/1374172/226557430-66457619-5459-4d66-ac3c-6f191953fb32.png)

nimishavijay:
Since there is a lot of useful information shown in that header area, we could also make the avatar smaller to 32px on mobile since there are no actions associated with it. One use of it would be to detect fraudulent mails (for example the fake Frank doesn't have a photo), and that is easy to identify even if the avatar is smaller.

PR #8434, opened 2023-05-02
feat: Easy unsubscribe from lists with mailto unsubscribe header

ChristophWurst:
For https://github.com/nextcloud/mail/issues/5387

Sibling of https://github.com/nextcloud/mail/pull/8395. Now lists that only send a `mailto` link are supported too. If both HTTP and mailto are sent we take the HTTP.

Ref https://github.com/nextcloud/mail/issues/5387

| Default | Confirmation | While unsubscribing |
|--------|--------|--------|
| Bildschirmfoto vom 2023-05-02 11-45-06 | Bildschirmfoto vom 2023-05-02 11-45-17 | Bildschirmfoto vom 2023-05-02 11-48-45 | 

## How to test

1) Open a message with a mailto unsubscribe header (e.g. Linkedin message)
2) See and click the *Unsubscribe* button
3) Click *Send unsubscribe email*
4) Wait a moment
5) See message in *Sent* mailbox

### DIFF
--- a/lib/IMAP/ImapMessageFetcher.php
+++ b/lib/IMAP/ImapMessageFetcher.php
@@ -70,6 +70,7 @@ class ImapMessageFetcher {
 	private string $rawReferences = '';
 	private string $dispositionNotificationTo = '';
 	private ?string $unsubscribeUrl = null;
+	private ?string $unsubscribeMailto = null;
 
 	public function __construct(int $uid,
 		string $mailbox,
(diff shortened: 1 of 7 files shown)


lib/IMAP/ImapMessageFetcher.php
```diff
@@ -520,6 +522,10 @@ private function parseHeaders(Horde_Imap_Client_Data_Fetch $fetch): void {
 					$this->unsubscribeUrl = $header->url;
 					break;
```
ChristophWurst, line 523: uncomment these two lines to test mailing listst that have a http and mailto list

GretaD:
it works, but it overlaps with the primary actions and in some not so small screens you cannot use the primary actions
Screenshot from 2023-05-02 12-29-32

ChristophWurst:
> it works, but it overlaps with the primary actions and in some not so small screens you cannot use the primary actions

https://github.com/nextcloud/mail/issues/8191

PR #8438, opened 2023-05-02
feat: One-click unsubscribe

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/5387
Close https://github.com/nextcloud/mail/issues/5387

Refinement of https://github.com/nextcloud/mail/pull/8395 for services that support one-click unsubscribe.

https://datatracker.ietf.org/doc/html/rfc8058

## How to test

1) Set up an account in Mail
2) Open a message from a mailing list (it has to have `List-Unsubscribe: <http…>` and `List-Unsubscribe-Post=List-Unsubscribe=One-Click` headers)
3) See the *Unsubscribe* button
4) Click the *Unsubscribe* button and see a confirmation


src/components/ThreadEnvelope.vue
```diff
@@ -612,6 +620,19 @@ export default {
 				return t('mail', 'Could not archive message')
 			}
 		},
+		async unsubscribeViaOneClick() {
+			try {
+				this.unsubscribing = true
+
+				await unsubscribe(this.envelope.databaseId)
+			} catch (error) {
+				logger.error('Could one-click unsubscribe', { error })
```
JohannesGGE, line 629: Could `not` one-click unsubscribe ?

JohannesGGE (approved):
Tested and works as expected (I think).
The unsubscribe landing page is not shown (4. wrong in the test desc?) and the list is unsubscribed.
Personally, I would appreciate a little feedback, the unsubscribe was send.

src/components/ThreadEnvelope.vue
ChristophWurst: Yeah :see_no_evil:

ChristophWurst:
> The unsubscribe landing page is not shown

My bad. Copied over the test steps from https://github.com/nextcloud/mail/pull/8395.

Will add a confirmation toast. Good idea!

JohannesGGE (approved):

miaulalala:
Tested, worked fine, cool stuff!

Two nitpicks:

1. Location of the unsubscribe button.  The current location crowds the lefthand side and doesn't really fit with the styling of the current header. I would prefer something like this:
image
cc @jancborchardt 

Alternatively add a few pixels of margin to the button:
| Before | After |
|--------|--------|
|image | image | 

2. The toast for the unsubscribe stays visible for a long time compared to other toasts.

ChristophWurst:
> Location of the unsubscribe button.  The current location crowds the lefthand side and doesn't really fit with the styling of the current header. I would prefer something like this:

See https://github.com/nextcloud/mail/issues/5387 and https://github.com/nextcloud/mail/issues/5387#issuecomment-1517635640 (bottom). The button was placed there intentionally.

> The toast for the unsubscribe stays visible for a long time compared to other toasts.

It uses the default for *success* toasts.

> the button is visible on mobile but messes up the header

Tracked in https://github.com/nextcloud/mail/issues/8191

miaulalala (approved):

kesselb:
Cool :+1: 

> A mail receiver can do a one-click unsubscription by performing an
   HTTPS POST to the HTTPS URI in the List-Unsubscribe header.  It sends
   the key/value pair in the List-Unsubscribe-Post header as the request
   body.

> The POST content SHOULD be sent as 'multipart/form-data' [[RFC7578](https://datatracker.ietf.org/doc/html/rfc7578)] or
   MAY be sent as 'application/x-www-form-urlencoded'.  These encodings
   are the ones used by web browsers when sending forms.  The target of
   the POST action is the same as the one in the GET action for a manual
   unsubscription, so this is intended to allow the same server code to
   handle both


To comply with the RFC, we need to send the List-Unsubscribe-Post key/value pair to the unsubscribe url.

ChristophWurst:
Thanks @kesselb. Didn't see any header values in the messages I tested with.

kesselb:
To add more fun: 

> As described below, the
   message MUST have a valid DomainKeys Identified Mail (DKIM) signature
   that covers at least the List-Unsubscribe and List-Unsubscribe-Post
   headers.

> The message needs at least one valid authentication identifier.  In
   this version of the specification, the only supported identifier type
   is DKIM [[RFC6376](https://datatracker.ietf.org/doc/html/rfc6376)].  Hence, senders MUST apply at least one valid DKIM
   signature to the message.

>   The List-Unsubscribe and List-Unsubscribe-Post headers MUST be
   covered by the signature and included in the "h=" tag of a valid
   DKIM-Signature header field.

 >  If the message does not have the required DKIM signature, the mail
   receiver SHOULD NOT offer a one-click unsubscribe for that message.

Possible library to validate DKIM: https://github.com/pimlie/php-dkim (horde already has a couple of methods to fetch and normalize headers, maybe we don't need an additional lib).

Plugin for Roundcube: https://github.com/pimlie/authres_status (very helpful readme with lots of information).

ChristophWurst:
@kesselb does your rebase mean this is done and ready to review? :)

lib/IMAP/ImapMessageFetcher.php
```diff
@@ -512,19 +515,29 @@ private function parseHeaders(Horde_Imap_Client_Data_Fetch $fetch): void {
...
-			$headers = $listHeaders->parse($listUnsubscribeHeader->name, $listUnsubscribeHeader->value_single);
-			foreach ($headers as $header) {
-				if (str_starts_with($header->url, 'http')) {
-					$this->unsubscribeUrl = $header->url;
-					break;
-				}
-				if (str_starts_with($header->url, 'mailto')) {
-					$this->unsubscribeMailto = $header->url;
-					break;
+		$dkimSignatureHeader = $parsedHeaders->getHeader('dkim-signature');
```
miaulalala, line 518: **nit:** early return would lead to better readability here and save about 3 brackets :wink:

miaulalala (approved):

kesselb:
@ChristophWurst ready to review.

lib/IMAP/ImapMessageFetcher.php
kesselb: Thanks. I will take that into account for the follow-up to add the DKIM signature validation.

ChristophWurst (commented):
:+1: latest additions look good!

PR #8453, opened 2023-05-04
perf(imap): FETCH only flags for partial message updates

### LINKED ISSUE #8451: Partial sync only has to FETCH flags, nothing else

Partial sync can be expensive without QRESYC/CONDSTORE capabilities. Horde will flag all given UIDs as changed. We then fetch the messages another time but only really use the flags and tags data to update the cache. Headers, subject, imapdate, etc are fetched without a purpose.

Adjust `\OCA\Mail\IMAP\Sync\Synchronizer::sync` to call a minimal version of `\OCA\Mail\IMAP\MessageMapper::findByIds` for changed message flags.

_No response_

### Additional context

_No response_

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/8451

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -257,6 +257,7 @@ class MessageMapper {
 								   string $userId,
 								   bool $loadBody = false): array {
 		$query = new Horde_Imap_Client_Fetch_Query();
+		$query->uid();
 		$query->flags();
 
 		return $this->findByQuery($ids, $query, $mailbox, $client, $loadBody, $userId);
@@ -328,11 +329,11 @@ class MessageMapper {
 		}
 
 		$fetchResults = array_values(array_filter($fetchResults, static function (Horde_Imap_Client_Data_Fetch $fetchResult) {
-			return $fetchResult->exists(Horde_Imap_Client::FETCH_ENVELOPE);
+			return $fetchResult->exists(Horde_Imap_Client::FETCH_UID);
 		}));
 
 		if ($fetchResults === []) {
-			$this->logger->debug("findByIds in $mailbox got " . count($ids) . " UIDs but found none");
+			$this->logger->debug("findByQuery in $mailbox got " . count($ids) . " UIDs but found none");
 		} else {
 			$minFetched = $fetchResults[0]->getUid();
 			$maxFetched = $fetchResults[count($fetchResults) - 1]->getUid();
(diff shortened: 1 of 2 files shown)


JohannesGGE:
before:
```
C: 7 UID FETCH 2726 (FLAGS BODY.PEEK[HEADER])
S: * 1 FETCH (UID 2726 FLAGS (\Seen NonJunk) BODY[HEADER] {8549}
Return-Path: <***@***>
Authentication-Results: ***
Received: ***
DKIM-Signature: ***
X-UI-Sender-Class: ***
Received: from ***
Date: ***
To: ***
From: ***
Reply-To: ***
Subject: ***
Message-ID: ***
X-MA-Provider: ***
X-Data: ***
MIME-Version: ***
Content-Type: ***
X-Provags-ID: ***
Envelope-To: ***
x-tdresult: ***
x-tdcapabilities: ***
X-Spam-Flag: ***
UI-InboundReport: ***

S: )
S: 7 OK UID FETCH completed
>> Command 7 took 0.047 seconds.
```

after:
```
C: 7 UID FETCH 2726 (FLAGS)
S: * 1 FETCH (UID 2726 FLAGS (\Seen NonJunk))
S: 7 OK UID FETCH completed
>> Command 7 took 0.0289 seconds.
```

ChristophWurst:
something is odd. 

> 341x	55.8 ms	↴ select * from oc_mail_tags where (imap_label = ?) and (user_id = ?)
> 341x	47.6 ms	↴ insert into oc_mail_message_tags(...) values (?)

from blackfire

JohannesGGE:
Found the issue. We use the messageID to check and update the tags. Without the header, we don't fetch the messageID anymore.

It could be possible to adjust the update of the tags with an additional join of mail_message_tags and mail_messages to get the uid that is fetched. Could be worth a try.

JohannesGGE:
The current state fails because the entries in `mail_message_tags` are stored with the message_id. The message_id is not part of the header and because of that not fetched in the new partial sync. `\OCA\Mail\Db\MessageMapper::updateTags` tries to diff on message_id and creates a generated message_id because none is found. Because of the mismatch of the message_id in `mail_message_tags` and `mail_messages` the partial sync always adds all tags to `mail_message_tags` (always with auto generated message_id).

It could be an option to join additionally on `mail_messages` in `\OCA\Mail\Db\TagMapper::getAllTagsForMessages`. This requires a bit more complex change because `\OCA\Mail\Db\MessageMapper::updateTags` needs the uid to diff and the matching message_id to create the db entry in `mail_message_tags` then.

For some reason the initial sync is also broken in this branch and the message_id is generated. I don't understand why, in my understanding only the partial sync is affected by the changes.

ChristophWurst:
Stale and conflicting

PR #8459, opened 2023-05-08
fix: Check strict cookies for image proxy

ChristophWurst:
https://github.com/nextcloud/mail/pull/2153's little sibling. It makes sense to check for the cookie existence here too.

## How to test

1) Set up an account
2) Open a HTML message with external images
3) Click *Show images* and *Show images temporarily*

main: images show
here: images show


GretaD (approved):
didnt test

Issue #8402, opened 2023-04-24
Lack of user feedback for Sieve script error


ChristophWurst:
### Steps to reproduce

1. Configure an account that has Sieve support
2. Enable Sieve filters
3. Enter a script but leave out the imports

### Expected behavior

Error message about validation, syntax error, or similar.

### Actual behavior

> Oh Snap! Request failed with status code 500

### Mail app version

3.1

### PHP engine version

None

### Web server

None

### Database

None

PR #8461, opened 2023-05-09
fix(sieve): show feedback on syntax error

JohannesGGE:
Fix: #8402 

`debug = false`

|before|after|
|--|--|
|Screenshot from 2023-05-11 14-44-21|Screenshot from 2023-05-11 14-46-36|

### DIFF
--- a/lib/Controller/SieveController.php
+++ b/lib/Controller/SieveController.php
@@ -39,6 +39,7 @@ use OCP\AppFramework\Http\JSONResponse;
 use OCP\IRequest;
 use OCP\Security\ICrypto;
 use OCP\Security\IRemoteHostValidator;
+use Psr\Log\LoggerInterface;
 
 class SieveController extends Controller {
 	private AccountService $accountService;
@@ -47,6 +48,7 @@ class SieveController extends Controller {
 	private string $currentUserId;
 	private ICrypto $crypto;
 	private IRemoteHostValidator $hostValidator;
+	private LoggerInterface $logger;
 
 	public function __construct(IRequest $request,
 								string $UserId,
@@ -54,7 +56,8 @@ class SieveController extends Controller {
 								MailAccountMapper $mailAccountMapper,
 								SieveClientFactory $sieveClientFactory,
 								ICrypto $crypto,
-								IRemoteHostValidator $hostValidator
+								IRemoteHostValidator $hostValidator,
+								LoggerInterface $logger
 	) {
 		parent::__construct(Application::APP_ID, $request);
 		$this->currentUserId = $UserId;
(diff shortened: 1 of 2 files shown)


lib/Controller/SieveController.php
```diff
@@ -109,7 +109,12 @@ public function updateActiveScript(int $id, string $script): JSONResponse {
 		$sieve = $this->getClient($id);
 
 		$scriptName = $sieve->getActive() ?? 'nextcloud';
-		$sieve->installScript($scriptName, $script, true);
+
+		try {
+			$sieve->installScript($scriptName, $script, true);
+		} catch (ManagesieveException $e) {
+			return new JSONResponse(statusCode: 422);
```
ChristophWurst, line 116: inject the LoggerInterface and log the error as warning or error including the exception object

a day will come where we have to debug a script saving issue and then we'll be thankful to find information in the logs :)

ChristophWurst: use the HTTP::STATUS_* constant for the status code

kesselb:
Hi, thanks for taking care :+1: 

The original code did not catch the ManageSieve exception => The [ErrorMiddleware](https://github.com/nextcloud/mail/blob/main/lib/Http/Middleware/ErrorMiddleware.php) will handle it. 

What I overlooked back then: The ErrorMiddleware behaves different for debug = true/false. 

debug = true you see the message like in the before screenshot. Oh snap! followed by some technical stuff. That's not very user friendly but at least a bit helpful to figer out the problem. 

debug = false you get Oh snap! without any explanation what's wrong :see_no_evil: 

The syntax seems to be incorrect is better. Even better if we include some data from the managesieve exception to give the user a idea what's wrong.

lib/Controller/SieveController.php
```diff
@@ -109,7 +113,13 @@ public function updateActiveScript(int $id, string $script): JSONResponse {
 		$sieve = $this->getClient($id);
 
 		$scriptName = $sieve->getActive() ?? 'nextcloud';
-		$sieve->installScript($scriptName, $script, true);
+
+		try {
+			$sieve->installScript($scriptName, $script, true);
+		} catch (ManagesieveException $e) {
+			$this->logger->error('Installing sieve script failed: ' . $e->getMessage(), ['app' => 'mail']);
```
ChristophWurst, line 120:
```suggestion
			$this->logger->error('Installing sieve script failed: ' . $e->getMessage(), ['app' => 'mail', 'exception' => $e]);
```

to log the trace

https://www.php-fig.org/psr/psr-3/#13-context

JohannesGGE:
@kesselb What do you suggest to show? The technical stuff is the only helpful message I can get from the exception, so I can show a part of that? e.g.:`error: unknown command 'fileinto' (only reported once at first occurrence).`

kesselb:
@JohannesGGE okay. I guess we have to forward the error message then.

kesselb (approved):
LGTM :+1:

Issue #8464, opened 2023-05-10
Sieve script is not loading in settings


JohannesGGE:
### Steps to reproduce

1. Use an account with enabled and set sieve script
2. Open Account Settings
3. Go to Sieve filter rules

### Expected behavior

Sieve script is loaded and shown.

### Actual behavior

Sieve script is loading... (screenshoot)

Can be fixed by clicking button "Save sieve settings" in Sieve filter server settings again.

![image](https://github.com/nextcloud/mail/assets/74607597/5b0f60e9-5ea4-476a-9a14-a918f6fd0c05)

### Mail app version

main

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
does this only happen right after adding the account or also if you reload the page between adding the account and opening the sieve settings?

do you see a js error or a failing XHR in the browser console?

JohannesGGE:
As far I can tell it happen always. There is no failing XHR. Seems the call isn't fired at all before "Save sieve settings" click.

PR #8466, opened 2023-05-10
fix(sieve): load sieve scipt on open settings

JohannesGGE:
Fix: #8464 

Sieve script loads when open the settings.


ChristophWurst (approved):
So it was a regression of https://github.com/nextcloud/mail/pull/8420

Nice!

Issue #7077, opened 2022-08-24
Rework Draft Handling - Frontend Part


miaulalala:
Front-end part of https://github.com/nextcloud/mail/issues/4768

- [x] Requires https://github.com/nextcloud/mail/issues/7076

ChristophWurst:
That was a misunderstanding. Local drafts are not visible to the user.

ChristophWurst:
The todo above is only draft, we haven't discussed it yet. I'll spec out the feature when the backend is done.

neeraj-nandan:
By when can we expect this issue to be resolved?

Please provide a date. It would be highly regarded.

PR #8469, opened 2023-05-11
feat: Rework draft handling front-end

hamza221:
Fixes https://github.com/nextcloud/mail/issues/7077
Closes https://github.com/nextcloud/mail/issues/4768

- [x] UC1
- [x] UC2
- [ ] UC3 => will be solved in a separate PR
- [x] UC4
- [x] UC5
- [x] UC6

### DIFF
--- a/src/store/outbox/actions.js
+++ b/src/store/outbox/actions.js
@@ -63,7 +63,6 @@ export default {
 	},
 
 	async enqueueMessage({ commit }, { message }) {
-		message = await OutboxService.enqueueMessage(message)
 		commit('addMessage', { message })
 
 		// Future drafts/sends after an error should go through outbox logic
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -46,6 +46,7 @@ Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud
 		<job>OCA\Mail\BackgroundJob\CleanupJob</job>
 		<job>OCA\Mail\BackgroundJob\OutboxWorkerJob</job>
 		<job>OCA\Mail\BackgroundJob\IMipMessageJob</job>
+		<job>OCA\Mail\BackgroundJob\DraftsJob</job>
 	</background-jobs>
 	<repair-steps>
 		<post-migration>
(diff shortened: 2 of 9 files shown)


JohannesGGE:
There is something wrong with the background job. 

### UC1
- local_messages still contains entry

### UC2
- works fine

### UC3
- local_messages entry is not converted and remains in db

### UC4
- works fine

### UC5
- works fine

### UC6
- UC1 -> works fine. local_messages entry seems to be deleted this time ('send_at' value is set)
- UC2 -> works fine
- UC3 -> local_messages up to date but not deleted. Draft in messages disappeared and no new updated version appears

hamza221:
UC3 still missing

ChristophWurst:
UC1: the drafts are stored locally. adding more text updates that draft. sending puts the message into the outbox. but the message is not sent.

ChristophWurst:
UC2+UC6 work :heavy_check_mark:

ChristophWurst:
> UC1: the drafts are stored locally. adding more text updates that draft. sending puts the message into the outbox. but the message is not sent.

I think the only thing missing is starting a timer with the undo timeout and triggering a send of the new outbox message.

On systems with reliably cron the message will be sent after ~5mins. but I think the frontend should trigger sending right after the undo timeout.

hamza221:
@ChristophWurst what. do you think should be done for UC3? should we handle it in a separate issue ?

hamza221:
The timer is already there but the message's sendAt was not updated correctly, Also I noticed that messages get enqueue when the draft is saved, I changed that too

src/store/outbox/actions.js
```diff
@@ -170,7 +169,7 @@ export default {
 					logger.error('Could not delay-send message ' + message.id, { message })
 					reject(error)
 				}
-			}, UNDO_DELAY)
+			}, UNDO_DELAY + 10)
```
ChristophWurst, line 172: for testing?

hamza221: Yes :/ , I removed it :))

JohannesGGE (approved):
With https://github.com/nextcloud/mail/pull/8576 U3 works now.

I only noticed one small thing:
In UC5: When there is no change the composer doesn't close completly. The minimized composer stays. You have to click on close again on the minimized composer.

JohannesGGE (approved):

ChristophWurst:
Green CI. Time for a squash.

PR #8473, opened 2023-05-12
Use the color-primary-element* variables

szaimen:
Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.

### DIFF
--- a/css/mail.scss
+++ b/css/mail.scss
@@ -349,7 +349,7 @@
 				&,
 				> a {
 					opacity: 1;
-					box-shadow: inset 2px 0 var(--color-primary);
+					box-shadow: inset 2px 0 var(--color-primary-element);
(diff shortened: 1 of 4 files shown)


JohannesGGE (approved):

szaimen:
Merge?

GretaD (approved):
didnt test but makes sense

ChristophWurst:
screeshots? backports?

GretaD:
> Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.

@szaimen when was this problem introduced?

szaimen:
Not sure honestly

PR #8489, opened 2023-05-22
Add key property to dashboard messages and fix misc errors for emails with empty from array

### LINKED ISSUE #4580: wrong avatars on the widget "Wichtige E-Mail"

IchbinkeinReh:
fixes #4580
this is because DashboardWidget uses field id as key for the v-for and the messages do not have an id field (see https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcDashboardWidget/NcDashboardWidget.vue#L190 )

### DIFF
--- a/src/components/Dashboard.vue
+++ b/src/components/Dashboard.vue
@@ -33,7 +33,7 @@
 				:main-text="itemMainText(item)"
 				:sub-text="itemSubText(item)">
 				<template #avatar>
-					<Avatar v-if="item.from"
+					<Avatar v-if="item.from && item.from.length"
 						:email="item.from[0].email"
 						:display-name="item.from[0].label"
 						:disable-tooltip="true"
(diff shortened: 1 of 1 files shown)


IchbinkeinReh:
@ChristophWurst sry for asking. Do i have to do anything to get this PR reviewed or did you have no time yet to review the PR?

src/components/Dashboard.vue
```diff
@@ -128,6 +128,7 @@ export default {
 
 		await Promise.all(inboxes.map(async (mailbox) => {
 			const messages = await fetchEnvelopes(mailbox.accountId, mailbox.databaseId, this.query, undefined, 10)
+			messages.forEach(message => { message.id = message.databaseId })
```
ChristophWurst, line 131: I don't comprehend why this is necessary. Could you elaborate?

PR #8494, opened 2023-05-24
perf(imap): Reduce number of STATUS commands for background mailbox sync

ChristophWurst:
Before this patch, the mailbox STATUS is fetched for *all* mailboxes. This is fine for accounts with a handful of mailboxes. Yet there are users who organize their email into hundreds of mailboxes and at some point this leads to excessive STATUS commands for every mailbox list sync.

This patch reduces the number of STATUS commands by only running them for the INBOX, mailboxes marked for background sync and 5 additional mailboxes.

If a user has hundreds of mailboxes and marks them all for background sync, we'll still have a performance problem. But otherwise the sync should be faster now, and mailboxes not marked for sync still get an occasional update about the unread/total messages stats.

This is a follow-up to https://github.com/nextcloud/mail/pull/8449

## How to test

0) Check out latest `main`
1) Add an account with at least 6 mailboxes
2) Comment out https://github.com/nextcloud/mail/blob/1115beaf2bf8b674fd0a6b50ee499605b7455dc6/lib/Command/SyncAccount.php#L104
3) Empty data/horde_imap.log
4) Run ``NC_debug=true php occ mail:account:sync <account id> --force``
5) Run ``cat data/horde_imap.log | wc -l`` and write down the number of lines reported
6) Check out ``perf/imap/reduce-mailbox-status-background-sync``
7) Empty data/horde_imap.log
8) Run ``NC_debug=true php occ mail:account:sync <account id> --force``
9) Run ``cat data/horde_imap.log | wc -l`` and write down the number of lines reported

For my test account I get 116 lines of IMAP communication on main and 67 lines with the patch.

### DIFF
--- a/lib/IMAP/MailboxSync.php
+++ b/lib/IMAP/MailboxSync.php
@@ -43,8 +43,11 @@ use OCP\IDBConnection;
 use Psr\Log\LoggerInterface;
 use function array_combine;
 use function array_map;
+use function array_reduce;
+use function array_slice;
 use function in_array;
 use function json_encode;
+use function shuffle;
 use function sprintf;
 use function str_starts_with;
 
@@ -103,14 +106,16 @@ class MailboxSync {
 				$namespaces = $client->getNamespaces([], [
 					'ob_return' => true,
 				]);
+				$personalNamespace = $this->getPersonalNamespace($namespaces);
 				$account->getMailAccount()->setPersonalNamespace(
-					$this->getPersonalNamespace($namespaces)
+					$personalNamespace
 				);
 			} catch (Horde_Imap_Client_Exception $e) {
 				$logger->debug('Getting namespaces for account ' . $account->getId() . ' failed: ' . $e->getMessage(), [
 					'exception' => $e,
 				]);
 				$namespaces = null;
+				$personalNamespace = null;
 			}
 
 			try {
@@ -125,7 +130,7 @@ class MailboxSync {
 			}
 			$this->folderMapper->detectFolderSpecialUse($folders);
 
-			$this->atomic(function () use ($account, $folders, $namespaces) {
+			$mailboxes = $this->atomic(function () use ($account, $folders, $namespaces) {
 				$old = $this->mailboxMapper->findAll($account);
 				$indexedOld = array_combine(
 					array_map(static function (Mailbox $mb) {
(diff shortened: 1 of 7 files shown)


ChristophWurst:
Branch updated to resolve conflicts

JohannesGGE:
I tested a bit and got weired results:

On main I get:
|sync in background|subscribed|lines in log|
|--|--|--|
|true|true|39 (no mailbox)|
|true|false|39 (no mailbox)|
|false|false|39 (no mailbox)|

On perf/imap/reduce-mailbox-status-background-sync I get:
|sync in background|subscribed|lines in log|
|--|--|--|
|true|true|95 (all mailboxes)|
|true|false|95 (all mailboxes)|
|false|false|63 (not all, random selection)|

lib/IMAP/MailboxSync.php
```diff
@@ -250,4 +268,36 @@ private function isMailboxShared(?Horde_Imap_Client_Namespace_List $namespaces,
...
+				[...$doNotSync, $mailbox],
+			];
+		}, [[], []]);
+
+		// Synchronize the mailboxes selected for sync and keep the rest updated occasionally
+		shuffle($doNotSync);
+		/** @var Mailbox[] $syncStatus */
+		$syncStatus = [...$sync, ...array_slice($doNotSync, 0, 5)];
+		foreach ($syncStatus as $mailbox) {
+			$stats = $this->folderMapper->getFoldersStatusAsObject($client, $mailbox->getName());
```
ChristophWurst, line 293: this is the only place we directly fetch the STATUS. @JohannesGGE please attach a debugger and see if this loop is responsible for the many STATUS commands, or if STATUS is requested somewhere else implicitly when we carry out another operation

ChristophWurst:
Bildschirmfoto vom 2023-06-05 12-36-25

I still only see STATUS commands for the mailboxes synced in background + five random other mailboxes

ChristophWurst:
The issue was that accounts with the ACL capabilities had MYRIGHTS sent more often than necessary.

st3iny:
Tested again and there is a single command more now when applying this PR (15 on main vs 16 here).

ChristophWurst:
how many mailboxes do you have and how many are marked for background sync?

ChristophWurst:
The production instance I wrote this patch for could reduce its cron load significantly after this patch. Let's get this in :)

JohannesGGE (approved):

ChristophWurst:
- [ ] BUG: An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1364 Field 'messages' doesn't have a default value

ChristophWurst:
After 300 attempts CI is green

ChristophWurst:
:heavy_check_mark: initial and partial sync of my test account still works

ChristophWurst:
Tried a local cherry-pick to v3.2. It's not trivial :s

PR #8496, opened 2023-05-25
build: update dependencies on nextcloud/ocp install

kesselb:
Follow-up for https://github.com/nextcloud/mail/pull/8490

I'm unsure if that's a sane solution.


kesselb:
Resolved via https://github.com/nextcloud/.github/pull/153 and https://github.com/nextcloud/mail/pull/8515

Issue #8007, opened 2023-02-10
mb_convert_encoding(): Handling HTML entities via mbstring is deprecated


ChristophWurst:
### Steps to reproduce

1. Run the app or tests

### Expected behavior

No deprecation warnings

### Actual behavior

> PHP Deprecated:  mb_convert_encoding(): Handling HTML entities via mbstring is deprecated; use htmlspecialchars, htmlentities, or mb_encode_numericentity/mb_decode_numericentity instead in /nextcloud/apps/mail/lib/Service/MimeMessage.php on line 50

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
Introduced with https://github.com/nextcloud/mail/commit/6f4fe62507a59754c6ed15a9546582a5a1a93201 :eyes: @kesselb

PR #8499, opened 2023-05-28
fix: html-entitites via mb_convert_encoding was deprecated with php 8.2

kesselb:
Fix https://github.com/nextcloud/mail/issues/8007

The email messages, built via CKEditor, is without <html> structure.
Without explicit definition (e.g. `<meta http-equiv="content-type" content="text/html; charset=UTF-8">`) DOMDocument uses ISO-8859-1.

By wrapping the email message in a <html> structure and adding a charset definition,
it's no longer necessary to convert the html-entities to work around the ISO-8891-1 encoding because DOMDocument uses the correct encoding.


Chartman123:
@kesselb Just tried it on my instance with PHP 8.2 and HTML mails seem to render just fine 👍🏻

kesselb:
Thanks for testing @Chartman123 :+1: 

Please note: This pull request is about sending ;) 
The original problem was that some characters were rendered wrong.

Chartman123:
Tested sending as well right now. :)

ChristophWurst (approved):

PR #8504, opened 2023-05-31
feat: set app name for mail

kesselb:
| Before | After |
| --- | --- |
| Screenshot from 2023-05-31 14-55-49 | Screenshot from 2023-05-31 15-23-12 |


ChristophWurst (approved):
Should we do the same for the other `src/main-*.js` files?

ChristophWurst (approved):

JohannesGGE (approved):

PR #8536, opened 2023-06-09
Fix calendar import popup

### LINKED ISSUE #7444: Mail app: importing *.ics files no longer works
### Steps to reproduce

1. get E-Mail sent with an *.ics File attached
2. click on "Import into calendar"
3. nothing happens, no further menue opens

### Expected behavior

after clicking on "Import into calendar" a select option should appear in wich calender the event shoud be added.

### Actual behavior

The select option for the wanted caldendar never appears

### Mail app version

1.14.1

### Mailserver or service

_No response_

### Operating system

Debian

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MySQL

### Additional info

I tested this behaiviour on my test cloud where version 1.13.9 was still installed of the mail app. It did work there, after updating to 1.14.1 the problem also appeared on my test cloud

GretaD:
Screenshot from 2023-06-09 11-57-08

to do

- [x] add a toas that says smth like "event imported"

fixes #7444

### DIFF
--- a/src/components/MessageAttachment.vue
+++ b/src/components/MessageAttachment.vue
@@ -34,45 +34,51 @@
 			<span class="attachment-size">{{ humanReadable(size) }}</span>
 		</div>
 		<Actions :boundaries-element="boundariesElement">
-			<ActionButton
-				v-if="isCalendarEvent"
-				class="attachment-import calendar"
-				:disabled="loadingCalendars"
-				@click.stop="loadCalendars">
-				<template #icon>
-					<IconAdd v-if="!loadingCalendars" :size="20" />
-					<IconLoading v-else-if="loadingCalendars" :size="20" />
-				</template>
-				{{ t('mail', 'Import into calendar') }}
-			</ActionButton>
-			<ActionButton
-				class="attachment-download"
-				@click="download">
-				<template #icon>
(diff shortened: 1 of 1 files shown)


src/components/MessageAttachment.vue
```diff
@@ -34,45 +34,51 @@
...
-				class="attachment-save-to-cloud"
-				:disabled="savingToCloud"
-				@click.stop="saveToCloud">
-				<template #icon>
-					<IconSave v-if="!savingToCloud" :size="20" />
-					<IconLoading v-else-if="savingToCloud" :size="20" />
-				</template>
-				{{ t('mail', 'Save to Files') }}
-			</ActionButton>
-			<div
```
ChristophWurst, line 66: would it not be sufficient to move the div outside of the actions element?

GretaD: no, because popovermenu is deprecated as a component and the second calendar popover is shown behind the "body" of the event.

JohannesGGE (approved):

ChristophWurst:
backport?

nanaksingh13:
calendar import still not working in Mail 3.2.1

Issue #1215, opened 2018-10-20
Searching for (LDAP)-Nextcloud-Accounts doesn't work


scroom:
Hello,

I configured the mail app in config.php so that every user gets his email account already integrated (LDAP server for mail and cloud). Every Nextcloud user is also registered with his email address via LDAP at Nextcloud.

I would expect that when I start typing in the address field, the mail app will suggest different Nextcloud users who match the letter combination. But unfortunately nothing happens.  Only locally created groups are suggested to me.

Nextcloud: 14.0.3
Mail-App: 0.11.0
Server: Ubuntu 18.04
Nginx

I've got this in the log:

> ini_restore() has been disabled for security reasons at /var/www/nextcloud/apps/mail/vendor/pear-pear.horde.org/Horde_Util/Horde/String.php#153

ChristophWurst:
IIRC we don't enumerate users from the instance, but just contacts from you address book and the ones you've sent messages before.

ChristophWurst:
-> https://github.com/nextcloud/mail/issues/848

scroom:
@ChristophWurst From the linked bug reports I conclude that some users don't want addresses to be used by the whole instance in autocompletion, but others want exactly that. 

Would be great to be able to use this optionally (as opt-in) -> checkbox "Read contacts from user database".

ChristophWurst:
> @ChristophWurst From the linked bug reports I conclude that some users don't want addresses to be used by the whole instance in autocompletion, but others want exactly that.
> 
> Would be great to be able to use this optionally (as opt-in) -> checkbox "Read contacts from user database".

IMO there should be a global option in the Nextcloud admin settings about this as it effects not just this app but is a global privacy setting.

Samonitari:
Pardon me the question?

Is there any progress regarding this?
Current behaviour (Nextcloud 23.0.0, Mail 1.11.6):
- Mail doesn't autocomplete LDAP users
- Mail **does however** LDAP autocomplete groups
- Global user search (next to you profile icon in the right corner)
  - **does**  find LDAP users
  - **can** view selected LDAP user's profile with a click on the "person icon"
  - **cannot** jump to mail to, only displays the address - "mail icon" is mailto link (not specific to LDAP users)

I think we can conclude that the Nextcloud-wise user search handling behaviour makes it:
- security-wise not a bit more secure than if the Mail would also autocomplete the users
[...]

PR #8551, opened 2023-06-15
Fix: Autocompletetion for system address book (not just full matches)

JohannesGGE:
Fix: #1215 

|before|after|
|--|--|
|Screenshot from 2023-06-15 15-07-30|Screenshot from 2023-06-15 15-04-33|

### How to test:
1. Checkout `mail/fix/8551/contact-manager-search` https://github.com/nextcloud/server/pull/38907 in the server repo
2. Enable SAB access in the admin settings (Allow username autocompletion in share dialog and allow access to the system address book)
3. Write a Mail and type first letters of a SAB contact in the TO field

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -72,7 +72,7 @@ class ContactsIntegration {
 		$shareeEnumerationFullMatchUserId = $shareeEnumerationFullMatch && $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_userid', 'yes') === 'yes';
 		$shareeEnumerationFullMatchEmail = $shareeEnumerationFullMatch && $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_full_match_email', 'yes') === 'yes';
 
-		$result = $this->contactsManager->search($term, ['UID', 'FN', 'EMAIL'], ['enumeration' => $shareeEnumeration]);
+		$result = $this->contactsManager->search($term, ['UID', 'FN', 'EMAIL'], ['enumeration' => $shareeEnumeration, 'fullmatch' => $shareeEnumerationFullMatch]);
 		if (empty($result)) {
 			return [];
 		}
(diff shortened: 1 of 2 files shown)


miaulalala (approved):
That if condition is one hecking crazy thing yo

lib/Service/ContactsIntegration.php
```diff
@@ -118,10 +118,14 @@ public function getMatchingRecipient(string $userId, string $term): array {
 					continue;
 				}
 				$lowerTerm = strtolower($term);
-				if (!$isSystemUser || $isInSameGroup || ($lowerTerm !== '' && (
-					($shareeEnumerationFullMatch && !empty($fn) && $lowerTerm === strtolower($fn)) ||
-					($shareeEnumerationFullMatchUserId && $lowerTerm === strtolower($id)) ||
-					($shareeEnumerationFullMatchEmail && $lowerTerm === strtolower($e))))) {
+				if (!$isSystemUser || // Not a user in SAB
```
ChristophWurst, line 121: similar to the first if in the loop we could flip the conditions and change to a series of small checks with continue

the code is indeed very hard to read. it was also hard to read before :) 

Add a test to https://github.com/nextcloud/mail/blob/main/tests/Unit/Service/ContactsIntegrationTest.php too for this case, please

ChristophWurst:
Ready to review/test? :)

ChristophWurst:
I don't know what I'm doing wrong but I do see other (local database) users even with main+master. Does only LDAP need this fix?

JohannesGGE:
@ChristophWurst Tested it again with a new docker setup (without LDAP user) and can't autocomplete user (only fullmatch) on main+master :confused: 
The search result is chached in some way, so if you already found the user before, the user will be shown after. Maybe thats the reason?

ChristophWurst:
> The search result is cached in some way

At which layer? I am testing with disabled HTTP caching

Bildschirmfoto vom 2023-06-29 08-55-51

is what ``\OC\ContactsManager::search`` passes into ``\OCP\IAddressBook::search`` for the SAB

JohannesGGE:
Yeah, the caching happens in the frontend.

I get the same results in the mentioned parts. My matching users are filtered out in the crazy if (https://github.com/nextcloud/mail/blob/f7d0678e8f20d23a1c19035141ec5ab73db7f86e/lib/Service/ContactsIntegration.php#L121-L124).

Your users pass somehow here?

PR #8553, opened 2023-06-15
Fix: OAuth refreshed access token was saved in the wrong field

tmrlvi:
Google OAuth was saving the token in wrong field in `refresh` (previous commit fixed `finishConnect` only)


ChristophWurst (approved):
Excellent find. I forgot this code when I moved the data from another column in https://github.com/nextcloud/mail/pull/7721

ChristophWurst:
The other OAuth implementation https://github.com/nextcloud/mail/blob/main/lib/Integration/MicrosoftIntegration.php does not have the problem :relieved:

PR #8562, opened 2023-06-16
Support for selecting messages with crtl key

### LINKED ISSUE #8048: Support Multiselect of Emails using Shift key

Right now you need to start multiselect of emails via the ... menu which is cumbersome. I would expect that pressing shift would allow me to select multiple messages but instead it opens (in chrome) a new window.

Implement multiselect using shift + click on the keyboard for many messages and ctrl/cmd + click for selected few messages.

_No response_

### Additional context

_No response_

rcsilva83:
Partially fixes #8048 because using shift key only works after some message is selected.

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -15,7 +15,8 @@
 		:data-envelope-id="data.databaseId"
 		:title="addresses"
 		:details="formatted()"
-		@click="onClick">
+		@click="onClick"
+		@click.ctrl="toggleSelected">
 		<template #icon>
 			<Star
 				v-if="data.flags.flagged"


rcsilva83:
My PR doesn't have any PHP code modifications. I think these test failures have nothing to do with my code.

ChristophWurst:
It's only the coverage report upload that fails :+1:

Chartman123:
@GretaD @ChristophWurst I just updated to Mail 3.3.0 on my instance and selecting mails with ctrl key opens every mail in a new tab... Should I open a separate issue?

rcsilva83:
> @GretaD @ChristophWurst I just updated to Mail 3.3.0 on my instance and selecting mails with ctrl key opens every mail in a new tab... Should I open a separate issue?

Just made a pull request to fix this: https://github.com/nextcloud/mail/pull/8798

Issue #8534, opened 2023-06-08
Unable to create mailbox folders on a new clean NextCloud installation


Hainer1:
### Steps to reproduce

1. click on the three menu points on the mail account bar icon
2. go to add mailbox
3. add name of the mailbox e.g. send , receive , deleted ,... 

### Expected behavior

The mailbox folder should be created

### Actual behavior

no mailbox folders are created 

### Mail app version

3.2.0

### Mailserver or service

flatbooster

### Operating system

Ubuntu 22.04

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

New, clean installation of ubuntu, and nextcloud.
I cannot send or receive email because of missing mailbox folders.
I get an exclamation mark on the left side of the mail account bar icon.

snetat:
Have you configured shared folders on the mail server ?

Hainer1:
No shared folders.
I have just managed to install the old 3.1.1 on my current VM. It is definitely the case that 3.1.1 works for creating mailboxes and including them in the setup. There must be a bug in 3.2.0.

ChristophWurst:
> I get an exclamation mark on the left side of the mail account bar icon.

The IMAP connect must be failing. Check your nextcloud.log entries written when you (re)load the Mail page.

Hainer1:
Yes, if I remember correctly, IMAP failed because there were no folders and I cannot create these folders in 3.2.0. If I create them, which is possible in 3.1.1, IMAP works. It may be possible to switch from 3.1.1 to 3.2.0 after setting up all the mails and folders, but I am not sure at the moment if I have configured all the users mails.

ChristophWurst:
Check your nextcloud.log entries written when you (re)load the Mail page and post them here.

Hainer1:
Here is the log after the failed manual IMAP configuration and reloading the mail page:
[...]

PR #8563, opened 2023-06-16
fix(imap): Ignore non existent mailboxes / no select for MYRIGHTS

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/8534

Regression of https://github.com/nextcloud/mail/pull/8449. The code https://github.com/nextcloud/mail/blob/5055ba884ed12ae820a9816d12175ae9a9bb9bc7/lib/IMAP/FolderMapper.php#L70-L80 was dropped, which previously caught access to invalid mailboxes and silently ignored the mailbox.

Mailbox synchronization fails if a nonexistent mailbox is access during the `MYRIGHTS` command, if the server supports ACLs.

```
C: 4 LIST () "" (*) RETURN (SUBSCRIBED SPECIAL-USE)
S: * LIST (\Subscribed \Archive) "/" Archives
S: * LIST (\Subscribed \Drafts) "/" Drafts
S: * LIST (\Subscribed \Sent) "/" Sent
S: * LIST (\Subscribed \Trash) "/" Trash
S: * LIST (\NonExistent) "/" shared
S: * LIST (\Subscribed) "/" shared/otheruser@domain.tld
S: * LIST (\Subscribed) "/" INBOX
```

but then

```
C: 11 MYRIGHTS shared
>> ERROR: read/timeout error.
>> ERROR: Server closed the connection.
```

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -63,6 +63,11 @@ class FolderMapper {
 		]);
 
 		return array_filter(array_map(static function (array $mailbox) use ($account) {
+			$attributes = array_flip(array_map('strtolower', $mailbox['attributes']));
+			if (isset($attributes['\\nonexistent'])) {
+				// Ignore mailbox that does not exist, similar to \Horde_Imap_Client::MBOX_SUBSCRIBED_EXISTS mode
+				return null;
+			}
 			if (in_array($mailbox['mailbox']->utf8, self::DOVECOT_SIEVE_FOLDERS, true)) {
 				// This is a special folder that must not be shown
 				return null;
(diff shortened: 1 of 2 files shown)


ChristophWurst:
This patch can be tested with the bitgrid test accounts

image

kesselb:
Do we need the same logic for \NoSelect?
Found a bug report for Thunderbird https://bugzilla.mozilla.org/show_bug.cgi?id=222394

ChristophWurst:
That probably makes sense. Let me add that too.

hamza221 (approved):
Tested creating new mailbox doesn't work on main and works with these fixes

st3iny (approved):
Wasn't able to reproduce the initial problem but the code looks good.

Issue #7988, opened 2023-02-06
Stared/Favourites folder empty


St4n:
### Steps to reproduce

1. Click on favourites folder on the left pane
2. Nothing is shown, the app says impossible to display messages

### Expected behavior

The fabourite folder should contain all stared messages

### Actual behavior

The starred messages are not shown, the app says impossible to display messages

### Mail app version

2.2.2

### Mailserver or service

Docker version of nextcloud, launched with docker-compose

### Operating system

Ubuntu 22.04 as docker host

### PHP engine version

PHP 8.1

### Web server

Nginx

### Database

MySQL

### Additional info

The Web server is the one included in the official Nextcloud docker container. I don't know which one it is.

Mer0me:
Same behaviour here, Nextcloud 26 / Mail 3.0.2, tested with Exchange and Dovecot IMAP server : starring a message add the star, make the message appears in Priority inbox, but the Favorite folder is always empty.
Is it supposed to work without IMAP server config ?

PR #8572, opened 2023-06-22
fix(message-filter): Show starred messages in Favourites again

JohannesGGE:
Fix: #7988 

Show starred messages in 'Favourites' again.


JohannesGGE:
Currently only starred messages in the INBOX are shown. So maybe we should filter all mailboxes?

ChristophWurst:
> Currently only starred messages in the INBOX are shown. So maybe we should filter all mailboxes?

It's intentionally just INBOX

ChristophWurst (approved):

GretaD (approved):

st3iny (approved):

PR #8576, opened 2023-06-26
fix(bgJob): Fix DB-Query for open drafts

JohannesGGE:
Fix: #7076 

local messages that don't have a send_at value should be converted to drafts, not sended messages.


ChristophWurst (approved):
Good find 🔎

Issue #4101, opened 2020-11-18
Mark as spam auto move message to spamfolder


luvis:
# Feature Request

Basically I would love to have the option to automatically move an email to the spamfolder when marked as spam. Once I've marked a message as spam I don't see the point in keeping it in my inbox. Maybe other people like to keep them, in which case might be able to add this as an option in settings? Its just annoying to have to mark as spam and then move -> to folder.. It takes to long, then I end up just deleting the email and spamassassin doesn't learn anything from it..

## Summary

When marking a message as spam, move message to spam folder

## Work packages

- [x] Implementation
- [x] https://github.com/nextcloud/documentation/issues/10958

ChristophWurst:
Yeah, though we have to be careful as some mail servers do this automatically when you set the flag and others don't.

StCyr:
What do you mean? The message would be moved on the IMAP server and the move would only be visible at next sync?

luvis:
> Yeah, though we have to be careful as some mail servers do this automatically when you set the flag and others don't.

Sorry, I missed your response. Even if this is the case, does it matter? If i move a message and the server plans on doing it automatically anyway, won't it just ignore it since it's already been moved? Or "move it again"? Either which, if it's an option in settings, it could be disabled by default. Those who's server does it automatically, won't need to enable it but those (like me) who's server doesn't automatically move the message, can enable it. That would be great 👍

ChristophWurst:
I guess it depends on when the server does it. If it moves right after the client adds the flag, the message to move won't exist anymore. So maybe we have to move first and then put the flag.

StCyr:
> So maybe we have to move first and then put the flag.

Seems legit. That way we do all the work and we don't rely on the server possibly moving the message

alexeysmovzh:
Hello Guys,

Does it any advance on this issue?

This is desired feature. 
Because users already used to this behavior with other popular web-mail systems.

ChristophWurst:
No but we had some more findings in a duplicate ticket https://github.com/nextcloud/mail/issues/5326#issuecomment-880038311

yitzhaq:
(From dup. #5326):

> Yes, we need a setting for the preferred junk mailbox like we have for draft, sent and trash.

[...]

PR #8581, opened 2023-06-27
feat: move messages to junk folder

kesselb:
For https://github.com/nextcloud/mail/issues/4101

**To Do**

- [x] Decide: Merge JunkSettings into AccountDefaultSettings
- [x] ~Decide: Rename "Mark as Spam" to "Move to Spam" with checkbox "Move message to Junk" enabled~
- [x] Inline documentation about the action/getter behavior
- [x] Testing

**Test cases**

| Message List | Message View |
| --- | --- |
|Screenshot from 2023-07-07 18-45-40 |Screenshot from 2023-07-07 18-46-01|

1. Add a new account. See a different icon for an existing folder/mailbox with the special role "junk".
2. Open account settings. A section "Junk settings" exist. The checkbox "Move message to Junk" is off by default.
3. Open action menu from the message list. Click "Mark as Spam". The message is flagged as spam, but not moved. 
4. Open action menu from the message view. Click "Mark as Spam". The message is flagged as spam, but not moved. 
5. Open account settings. Toggle checkbox "Move message to Junk".
6. Open action menu from the message list. Click "Mark as Spam". The message is flagged as spam and moved.
7. Open action menu from the message view. Click "Mark as Spam". The message is flagged as spam and moved. 
8. Open junk mailbox and repeat 6 and 7.

### DIFF
--- a/lib/Listener/MoveJunkListener.php
+++ b/lib/Listener/MoveJunkListener.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * @author Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Mail\Listener;
+
+use OCA\Mail\Contracts\IMailManager;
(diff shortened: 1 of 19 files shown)


lib/Listener/MoveJunkListener.php
```diff
@@ -0,0 +1,102 @@
...
+				]);
+			}
+		} elseif (!$event->isSet() && 'INBOX' !== $mailbox->getName()) {
+			try {
+				$this->mailManager->moveMessage(
+					$account,
+					$mailbox->getName(),
+					$event->getUid(),
+					$account,
+					'INBOX',
```
ChristophWurst, line 92: can we assume the name to be always exactly like that? what about capitalization? personal namspace?

jancborchardt (commented):
> Decide: Rename "Mark as Spam" to "Move to Spam" with checkbox "Move message to Junk" enabled

It should stay a simple "Mark as spam" action. It is expected from (the majority of) other clients that "Mark as spam" will also move it to spam/junk.

kesselb:
image

I will add the junk mailbox to default folders.

ChristophWurst (approved):
Works

ChristophWurst:
```
OCA\Mail\Tests\Integration\Db\MailAccountTest::testMailAccountConstruct
--- Expected
+++ Actual
@@ @@
     'quotaPercentage' => null
     'trashRetentionDays' => 60
     'name' => 'Peter Parker'
+    'junkMailboxId' => null
+    'moveJunk' => false
 )
```

jancborchardt:
Follow-up as talked with @ChristophWurst and mentioned above:
>  It is expected from (the majority of) other clients that "Mark as spam" will also move it to spam/junk.

Meaning the toggle to need to enable this functionality can be removed as it should be enabled by default.

PR #8594, opened 2023-07-03
Add slash command and emoji picker to mail composer

### LINKED ISSUE #8514: Bring CKEditor up to standards with NcRichContenteditable

As a user of Nextcloud I want to find uniform text input behavior. Most apps use [NcRichContenteditable](https://nextcloud-vue-components.netlify.app/#/Components/NcRichContenteditable) now where I can
* use <kbd>/</kbd> to open the smart picker
* use <kbd>:</kbd> to suggest emojis

In the Mail app none is available.

Find a way to bring CKEditor up to standards with NcRichContenteditable features or adapt NcRichContenteditable so it can be used as rich text editor in Mail.

_No response_

### Additional context

_No response_

hamza221:
Closes #8514 
image
image

### DIFF
--- a/src/ckeditor/smartpicker/InsertItemCommand.js
+++ b/src/ckeditor/smartpicker/InsertItemCommand.js
@@ -1,0 +1,81 @@
+/**
+ * @author Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+import Command from '@ckeditor/ckeditor5-core/src/command'
+export default class InsertItemCommand extends Command {
+
+	/**
+	 * @param {module:core/editor/editor~Editor} editor instance
+	 * @param {module:engine/model/writer~Writer} writer instance
+	 * @param {string} item smart picker or emoji picker
+	 * @param {string} trigger the character to replace
+	 */
+	insertItem(editor, writer, item, trigger) {
+		const currentPosition = editor.model.document.selection.getLastPosition()
+		if (currentPosition === null) {
+			// null as current position is probably not possible
+			// @TODO Add error to handle such a situation in the callback
+			return
+		}
+
+		const range = editor.model.createRange(
+			currentPosition.getShiftedBy(-5),
+			currentPosition
+		)
+
+		// Iterate over all items in this range:
+		const walker = range.getWalker({ shallow: false, direction: 'backward' })
+
+		for (const value of walker) {
+			if (value.type === 'text' && value.item.data.includes(trigger)) {
+				writer.remove(value.item)
+
+				const text = value.item.data
+				const lastSlash = text.lastIndexOf(trigger)
+
+				const textElement = writer.createElement('paragraph')
+				writer.insertText(text.substring(0, lastSlash), textElement)
+				editor.model.insertContent(textElement)
+
+				const itemElement = writer.createElement('paragraph')
+				writer.insertText(item, itemElement)
+				editor.model.insertContent(itemElement)
+
+				return
+			}
+		}
+
+		// @TODO If we end up here, we did not find the slash. We should throw an error maybe.
+	}
+
+	/**
+	 * @param {string}  item link from smart picker or emoji from emoji picker
+	 * @param {string} trigger the character to replace
+	 */
+	execute(item, trigger) {
+		this.editor.model.change(writer => {
+			this.insertItem(this.editor, writer, item, trigger)
+		})
+	}
+
+	refresh() {
+		this.isEnabled = true
+	}
+
+}
--- a/src/ckeditor/smartpicker/PickerPlugin.js
+++ b/src/ckeditor/smartpicker/PickerPlugin.js
@@ -1,0 +1,31 @@
+/**
+ * @author Daniel Kesselberg <mail@danielkesselberg.de>
+ *
+ * Mail
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin'
+import InsertItemCommand from './InsertItemCommand'
+export default class PickerPlugin extends Plugin {
+
+	init() {
+		this.editor.commands.add(
+			'insertItem',
+			new InsertItemCommand(this.editor)
+		)
+	}
+
+}
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -53,7 +53,9 @@ import ImagePlugin from '@ckeditor/ckeditor5-image/src/image'
 import ImageResizePlugin from '@ckeditor/ckeditor5-image/src/imageresize'
 import ImageUploadPlugin from '@ckeditor/ckeditor5-image/src/imageupload'
 import MailPlugin from '../ckeditor/mail/MailPlugin'
-
+import Tribute from 'tributejs/dist/tribute.esm.js'
+import { searchProvider, getLinkWithPicker } from '@nextcloud/vue/dist/Components/NcRichText'
+import { emojiSearch, emojiAddRecent } from '@nextcloud/vue/dist/Functions/emoji'
 import { getLanguage } from '@nextcloud/l10n'
 
 import logger from '../logger'
@@ -131,6 +133,9 @@ export default {
 		}
 
 		return {
+			linkTribute: null,
+			emojiTribute: null,
+			textSmiles: [],
 			ready: false,
 			editor: Editor,
 			config: {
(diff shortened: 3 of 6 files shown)


ChristophWurst:
> Inserting 'any link' always puts the cursor at the start of the message

is it because the focus is lost when the picker shows? Could we get the last cursor position on blur?

kesselb:
@hamza221 mind to resolve the conflicts and rebase on main?

kesselb:
> cursor position after inserting link is nondeterministic

I can reproduce. CKEditor uses an internal structure to represent the content, and therefore it's a bit tricky to insert data or modify the content externally. 

```
this.editorInstance.execute('delete')
this.appendToBodyAtCursor(link)
```

The approach to delete the last character only works for `/` and `:`. 
It does not work for `/fil` or `:wav`. Especially for emojis, it is quite common to filter the list. 

What we actually want is to take the current position, go back to the last slash and replace with the link.
Here is a small poc for the smart picker: https://github.com/nextcloud/mail/pull/8898

I found your approach connecting Tribute.js and CKEditor very creative, yet I have mixed feelings about it. CKEditor has a mentions plugin https://ckeditor.com/docs/ckeditor5/latest/features/mentions.htm and there is also a poc for a slash commands based on it https://github.com/ckeditor/ckeditor5/commit/865c940598a27d65caab7de506931be2abc20761. 

Likely, an implementation based on the mentions plugin will take some time. 
On the other hand, Tribute.js is unmaintained as per https://github.com/zurb/tribute/issues/765.

ChristophWurst:
CKEditor slash commands need a license https://github.com/nextcloud/mail/issues/8514#issuecomment-1599391483

kesselb:
Slash commands is a premium feature and requires a commercial license. 
Mentions are not a premium feature.

My hope was/is that we could implement our slash commands based on the mentions plugin. The CKEditor slash commands is also using the mentions plugin internally. https://github.com/ckeditor/ckeditor5/commit/865c940598a27d65caab7de506931be2abc20761 is a poc done by a CKEditor developer. 

I would prefer an integration with CKEditor based on the mentions plugin, but cannot give you a estimate. 
Fine by me to continue with the Tribute.js approach and our small plugin to handle the insertion.

kesselb:
> Cursor position resets after inserting a Link everything else works fine

Can you confirm that it only happens for the first link and the second, third, … work as expected? 
That makes it even more weird ;)

kesselb:
- `writer.setSelection(itemElement, 'after')` in insertItem does not help because the cursor position is correct over there. 
- I cannot reproduce the reset cursor problem with `const link = 'https://www.nextcloud.com'` instead of the smart picker call. As you suggested, something done by the smart picker let ckeditor reset the cursor position but only once in my testings.

src/components/TextEditor.vue
```diff
@@ -178,6 +259,25 @@ export default {
...
+				const eventData = data[0]
+				const item = eventData.mention
+				if (eventData.marker === ':') {
+					emojiAddRecent(item)
+					this.editorInstance.execute('InsertItem', item.native, ':')
+				}
+				if (eventData.marker === '/') {
+					getLinkWithPicker(item.id)
+						.then((link) => {
+							this.editorInstance.execute('InsertItem', link, '/')
```
kesselb, line 274:
```suggestion
							this.editorInstance.execute('InsertItem', link, '/')
							this.editorInstance.editing.view.focus()
```

Seems to "workaround" the reset cursor issue. 

If the reference picker is open window._nc_focus_trap shows three focus traps.
When closing the picker the focus goes back to ckeditor.
I'm still lost what is causing ckeditor to reset the cursor then, especially because we had a selection before.

src/ckeditor/smartpicker/InsertItem.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+import Command from '@ckeditor/ckeditor5-core/src/command'
+export default class InsertPickerItemCommand extends Command {
+
+	/**
+	 * @param {module:core/editor/editor~Editor} editor instance
+	 * @param {module:engine/model/writer~Writer} writer instance
+	 * @param {string} item smart picker or emoji picker
+	 * @param trigger
```
kesselb, line 27:
```suggestion
	 * @param {string} trigger the character to replace
```

src/ckeditor/smartpicker/InsertItem.js
```diff
@@ -0,0 +1,81 @@
...
+				return
+			}
+		}
+
+		// @TODO If we end up here, we did not find the slash. We should throw an error maybe.
+	}
+
+	/**
+	 * @param {string}  item link from smart picker or emoji from emoji picker
+	 * @param trigger
```
kesselb, line 69:
```suggestion
	 * @param {string} trigger the character to replace
```

src/ckeditor/smartpicker/InsertItem.js
```diff
@@ -0,0 +1,81 @@
...
+import Command from '@ckeditor/ckeditor5-core/src/command'
+export default class InsertPickerItemCommand extends Command {
+
+	/**
+	 * @param {module:core/editor/editor~Editor} editor instance
+	 * @param {module:engine/model/writer~Writer} writer instance
+	 * @param {string} item smart picker or emoji picker
+	 * @param trigger
+	 */
+	InsertItem(editor, writer, item, trigger) {
```
kesselb, line 29:
```suggestion
	insertItem(editor, writer, item, trigger) {
```

src/ckeditor/smartpicker/InsertItem.js
```diff
@@ -0,0 +1,81 @@
...
+		// @TODO If we end up here, we did not find the slash. We should throw an error maybe.
+	}
+
+	/**
+	 * @param {string}  item link from smart picker or emoji from emoji picker
+	 * @param trigger
+	 */
+	execute(item, trigger) {
+		this.editor.model.change(writer => {
+			this.InsertItem(this.editor, writer, item, trigger)
```
kesselb, line 73:
```suggestion
			this.insertItem(this.editor, writer, item, trigger)
```

src/ckeditor/smartpicker/InsertItem.js
```diff
@@ -0,0 +1,81 @@
+/**
```
kesselb, line 1: Please rename the file to InsertItemCommand.js

(to keep the naming somewhat in sync with the signature plugin)

src/ckeditor/smartpicker/InsertItem.js
```diff
@@ -0,0 +1,81 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+import Command from '@ckeditor/ckeditor5-core/src/command'
+export default class InsertPickerItemCommand extends Command {
```
kesselb, line 21:
```suggestion
export default class InsertItemCommand extends Command {
```

src/ckeditor/smartpicker/PickerPlugin.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ */
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin'
+import InsertItem from './InsertItem'
+export default class PickerPlugin extends Plugin {
+
+	init() {
+		this.editor.commands.add(
+			'InsertItem',
```
kesselb, line 26:
```suggestion
			'insertItem',
```

src/ckeditor/smartpicker/PickerPlugin.js
```diff
@@ -0,0 +1,31 @@
...
+			new InsertItem(this.editor)
```
kesselb, line 27:
```suggestion
			new InsertItemCommand(this.editor)
```

src/components/TextEditor.vue
kesselb, line 274:
```suggestion
							this.editorInstance.execute('insertItem', link, '/')
```

src/tests/unit/components/PickerPlugin.vue.spec.js
```diff
@@ -0,0 +1,51 @@
...
+			const editor = await VirtualTestEditor.create({
+				initialData: text,
+				plugins: [ParagraphPlugin, PickerPlugin],
+			})
+			
+			editor.model.change(writer => {
+				writer.setSelection(editor.model.document.getRoot(), 'end')
+			})
+
+			editor.execute('InsertItem',
```
kesselb, line 42:
```suggestion
			editor.execute('insertItem',
```

src/components/TextEditor.vue
```diff
@@ -178,6 +259,25 @@ export default {
...
 		onEditorReady(editor) {
 			logger.debug('TextEditor is ready', { editor })
+
+			editor.commands.get('mention')?.on('execute', (event, data) => {
+				event.stop()
+				const eventData = data[0]
+				const item = eventData.mention
+				if (eventData.marker === ':') {
+					emojiAddRecent(item)
+					this.editorInstance.execute('InsertItem', item.native, ':')
```
kesselb, line 269:
```suggestion
					this.editorInstance.execute('insertItem', item.native, ':')
```

src/components/TextEditor.vue
hamza221: The workaround works 👍

src/ckeditor/smartpicker/InsertItemCommand.js
```diff
@@ -0,0 +1,81 @@
...
+				return
+			}
+		}
+
+		// @TODO If we end up here, we did not find the slash. We should throw an error maybe.
+	}
+
+	/**
+	 * @param {string}  item link from smart picker or emoji from emoji picker
+	 * @param trigger
```
kesselb, line 69:
```suggestion
	 * @param {string} trigger the character to replace
```

ChristophWurst:
- [ ] BUG: Type `hello / this is a test` and then <kbd>Enter</kbd>

I expect the smart picker only to show if I press enter right after the slash

hamza221:
> * [x]  BUG: Type `hello / this is a test` and then Enter
> 
> 
> I expect the smart picker only to show if I press enter right after the slash

It's because searchProvider returns 'Any Link'  if there's no match which is not useful here. 
I added a workaround -> Now the options only show if there's a match so you can still search the options through typing

ChristophWurst (approved):
I can't comprehend the CKEditor magic but the feature works flawless now

:rocket:

src/ckeditor/smartpicker/InsertItem.js
kesselb, line 1: Please don't forget to delete this file ;)

hamza221: Must've did something wrong while squashing. Thank you for the heads up

kesselb (approved):

kesselb:
Possible follow up:

- Compose new message
- Insert a file
- After selecting a file you see a small box for a few seconds. 
- If you click on the box or somewhere else the insert is canceled.  I guess that's the default modal click outside logic. It would be nice to not cancel the process on outside click if possible.

hamza221:
@kesselb  I'm not sure I understand , Do you mean what's happening in the screen recording?
[Screencast from 2023-10-27 17-34-45.webm](https://github.com/nextcloud/mail/assets/40746210/4ffacf97-6e69-4ca7-bbf3-29894a9a8565)

PR #8603, opened 2023-07-05
chore: Ignore HTTP response class templates

ChristophWurst:
They are new in 28 and missing for 27 and older.

Ref https://github.com/nextcloud/server/pull/38802

### DIFF
--- a/lib/Http/ProxyDownloadResponse.php
+++ b/lib/Http/ProxyDownloadResponse.php
@@ -27,6 +27,10 @@ namespace OCA\Mail\Http;
 use DateTime;
 use OCP\AppFramework\Http\DownloadResponse;
 
+/**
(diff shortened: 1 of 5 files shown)


lib/Http/ProxyDownloadResponse.php
```diff
@@ -27,6 +27,11 @@
 use DateTime;
 use OCP\AppFramework\Http\DownloadResponse;
 
+/**
+ * @see https://github.com/omniti-labs/jsend
```
provokateurin, line 31: Copy paste mistake from JsonResponse

provokateurin (approved):

lib/Http/ProxyDownloadResponse.php
ChristophWurst: yes indeedy

PR #8605, opened 2023-07-05
chore: Remove unused phpdoc params

ChristophWurst:
In preparation of Psalm 6.

Removed phpdoc params have no real usage in the code.


kesselb (approved):
:broom:

PR #8606, opened 2023-07-05
fix(provisioning): Clear cache before returning

ChristophWurst:
I'm proud of how stupid my mistake is.

Discovered by Psalm.

Ref https://github.com/nextcloud/mail/pull/7992


st3iny (approved):
🙈

PR #8611, opened 2023-07-06
fix: add preview enhancement job to new accounts joblist

miaulalala:
Must have slipped under the radar along the way.

### DIFF
--- a/lib/Service/AccountService.php
+++ b/lib/Service/AccountService.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
 namespace OCA\Mail\Service;
 
 use OCA\Mail\Account;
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

ChristophWurst:
Job class added in https://github.com/nextcloud/mail/pull/6812. Does that mean IMIP processing didn't run or only for the messages shown in the UI?

kesselb (approved):

miaulalala:
> Job class added in #6812. Does that mean IMIP processing didn't run or only for the messages shown in the UI?

I think the jobslist would have only contained jobs for existing accounts.

PR #8616, opened 2023-07-11
feat: parse mailto addresses

kesselb:
To improve compatibility with contacts send email feature.

Send email will compose a mailto link with all email addresses from the contact group. A mailto link with multiple email addresses needs some preprocessing in mail.

**How to test:**

- Mail: Enable "Register as application for mail links" 
- Contacts: Create a contact group. Add some contacts. Make sure the contacts have an email address.
- Contacts: Click "Send email" for the contact group.
- Mail: See all email addresses in the To field ;)

**Side note:**

The mailto link generated by contacts (https://github.com/nextcloud/contacts/blob/a72b5849337354bac60db24fd56c7e062cf3135d/src/components/AppNavigation/GroupNavigationItem.vue#L177) does not comply with https://datatracker.ietf.org/doc/html/rfc6068. 

However, most clients seem to throw the mailto link into a parser and are not too strict about the format.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,7 @@
         "@nextcloud/vue-dashboard": "^2.0.1",
         "@riophae/vue-treeselect": "^0.4.0",
         "@vue/babel-preset-app": "^5.0.8",
+        "address-rfc2822": "^2.1.0",
         "buffer": "^6.0.3",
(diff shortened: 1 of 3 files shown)


st3iny (approved):
Tested and works.

Issue #3562, opened 2020-09-07
mail 'search function doesn't search in senders' address and name


StCyr:
### Expected behavior
Searching for the name of the sender of an email should match the corresponding email

### Actual behavior
Searching for the name of the sender of an email doesn't return any result

### Mail app

**Mail app version:** master

**Mailserver or service:** (e.g. Outlook, Yahoo, Gmail, Exchange,...)

### Server configuration

**Operating system**: (e.g. Debian 8)

**Web server:** (e.g. Apache, Nginx,...)

**Database:** (e.g. MariaDB, SQLite or PostgreSQL)

**PHP version:** (e.g. 7.0)

**Nextcloud Version:** NC20beta2

#### Client configuration
**Browser:** (e.g. Firefox 48)

**Operating system:** (e.g. Arch Linux)

ChristophWurst:
right … only if you use `from:user@domain.com`

StCyr:
would be cool if it'd it automagicaly: The `from:...` syntax only works with email adresses not names, and doesn't support partial match (ie: "cyrille" for matching "cyrille@bollu.be")

StCyr:
Ohoh... The search function doesn't seem to search in mails' body neither...

I guess you don't want to accept the performance penalty of such extended search functions.

(FYI, I believe the search happens here: https://github.com/nextcloud/mail/blob/7edf71937ce86ab4642ef617649e45305ef7a1ba/lib/Db/MessageMapper.php#L564)

Possible solutions/workarounds that I imagine:

1. (easy) Document mail 'search function in the Settings pane (same as for the keyboard shortcuts documentation);
2. (hard) Add a FULLTEXT index for the messages' body and add support for fulltext search;
3. (bad idea) Add an administrator option to allow searching in mails' body and use `LIKE '%foobar%'` queries to actually search in messages' body. (I was thinking about this idea for small NC instances, but, eventually, I've noticed my small instance, with only me as mail user, as already about 11000 messages in its DB)

Same possible workaround/solutions probably apply for searching in sender addresses and names.

ChristophWurst:
This is known and I put down some possible solutions at https://github.com/nextcloud/mail/issues/3481.

StCyr:
> 
> 
> Ohoh... The search function doesn't seem to search in mails' body neither...
> 
> I guess you don't want to accept the performance penalty of such extended search functions.
> 
> (FYI, I believe the search happens here:
> 
[...]

PR #8645, opened 2023-07-21
feat(search): Use case-independent wildcard matches and match labels

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/3562

Search only found exact matches before (full string and correct case). Now it searches with wildcard, case-independent and also looks into the label of a recipient.

How to test
1) Search by partial email address, e.g. messages from *@gmail.com*
2) Search by recipient label, e.g. *jane doe*

main branch: no results
here: results


GretaD:
I tested it but i get no result if i type _i.e @yahoo.com_. or name surname. I checked the branch, run a watcher and start searching. Am i missing something

ChristophWurst:
It only works if you open the advanced search and type the domain into the *From* field. The main search field only searches in the subject.

GretaD (approved):
it works

kesselb (approved):
Nice :+1:

PR #8524, opened 2023-06-06
Fix:re-open account settings

hamza221:
Fix #8523

PR #8650, opened 2023-07-25
feat: load active sieve script on demand

kesselb:
**How to reproduce**

- Open mail
- See xhr request for every configured mail account for the active sieve script

https://github.com/nextcloud/mail/pull/8466 added `dispatch(fetchActiveSieveScript)` to load the active script when the settings component is mounted.  

https://github.com/nextcloud/mail/pull/8524 changed the component from mounted on demand to always mounted and therefore triggers  `dispatch(fetchActiveSieveScript)` on page load. 

I'm using account.sieveEnabled as default for fetchActiveSieveScript to load only when sieve is enabled for the account.

### DIFF
--- a/src/components/AccountSettings.vue
+++ b/src/components/AccountSettings.vue
@@ -128,6 +128,7 @@ import SieveFilterForm from './SieveFilterForm'
 import OutOfOfficeForm from './OutOfOfficeForm'
 import CertificateSettings from './CertificateSettings'
 import TrashRetentionSettings from './TrashRetentionSettings'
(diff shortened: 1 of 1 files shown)


src/components/AccountSettings.vue
```diff
@@ -150,6 +151,7 @@ export default {
 	data() {
 		return {
 			trapElements: [],
+			fetchActiveSieveScript: this.account.sieveEnabled,
```
kesselb, line 154: It's safe to access the prop here, right?

PR #8653, opened 2023-07-26
add thread summary ui

### LINKED ISSUE #8508: Summarize email thread

As a user I receive long email threads and I want to efficiently get the gist of them without reading the details.

Show a text box below (long) threads with a few sentences of summary of the messages above.

A reference concept can be seen at https://www.youtube.com/watch?v=6DaJVZBXETE&t=25s.

## Implementation

- [x] https://github.com/nextcloud/server/issues/38578 (to be done by the Nextcloud integration team)
- [x] Requires mockup and interaction specs from designers
  * Where to render the text
  * Starting at how many messages in a thread do we summarize? 1? 2?
- [x] MVP
  * Make it work with integration_openai mainly because that is fast
  * Process summaries on demand when the user opens the thread (in a web request, no background processing)
  * Fetch thread message bodies in a service inside the Mail app and send the data for processing to https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/text_processing.html
- [x] Cache processed threads
- [x] https://github.com/nextcloud/documentation/issues/10917

An API will be provided to which we can send text and retrieve a summary. Therefore, the Mail app has to be changed to collect a thread's message texts, have the text summarized and show the results in the UI.

### Sync vs async processing

If the summary should show instantly, when loading the thread, the processing has to happen in the background and the result persisted with the thread data.
If the summary is generated async, the data can be processed on demand. Depending on the performance of the AI there can be a significant delay.

### Opt-out

There will be people who don't want this feature. We should allow them to turn off the feature.

_No response_

## Additional context

_No response_

hamza221:
closes #8508


ChristophWurst:
Set to draft because we'll do the full MVP on this branch

lib/Controller/ThreadController.php
```diff
@@ -111,4 +115,34 @@ public function delete(int $id): JSONResponse {
...
+		}
+
+		try {
+			$summary = $this->aiIntergrationsService->summarizeThread(
+				$id,
+				$messages,
+				$this->currentUserId,
+			);
+		} catch (ServiceException $e) {
+			return new JSONResponse([], Http::STATUS_NO_CONTENT);
```
ChristophWurst, line 142: log the exception

lib/Service/AIIntergrationsService.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace OCA\Mail\Service;
+
+use OCA\Mail\Exception\ServiceException;
+use OCP\TextProcessing\IManager;
+use OCP\TextProcessing\SummaryTaskType;
+use OCP\TextProcessing\Task;
+use function array_map;
+
+class AIIntergrationsService {
```
ChristophWurst, line 32: you found yourself an interesting naming problem: https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

lib/Service/AIIntergrationsService.php
```diff
@@ -0,0 +1,61 @@
...
+
+	public function __construct(IManager $manager) {
+		$this->manager = $manager;
+	}
+	/**
+	 *
+	 * @throws ServiceException
+	 */
+	public function summarizeThread(string $threadId, array $messages, $currentUserId): string {
+		if(in_array(SummaryTaskType::class, $this->manager->getAvailableTaskTypes())) {
```
ChristophWurst, line 45:
```suggestion
		if(in_array(SummaryTaskType::class, $this->manager->getAvailableTaskTypes(), true)) {
```

won't change a lot but let's use strict in_array consistently

src/components/Thread.vue
```diff
@@ -165,6 +173,15 @@ export default {
 		window.removeEventListener('resize', this.resizeDebounced)
 	},
 	methods: {
+		async updateSummary() {
+			if (this.thread.length > 2 && this.enabledThreadSummary) {
+				this.summaryLoading = true
+				 await summarizeThread(this.thread[0].threadRootId, this.accountId).then(summary => {
+					this.summaryText = summary
+					this.summaryLoading = false
+				})
```
ChristophWurst, line 182: mix of async-await and .then. use async syntax only

lib/Service/AIIntergrationsService.php
hamza221: so what's the verdict AIIntergrationsService or AiIntergrationsService?

ChristophWurst: I'd go with AiIntergrationsService like we did with SmimeService

src/components/Thread.vue
```diff
@@ -57,34 +58,48 @@ import { NcAppContentDetails as AppContentDetails, NcPopover as Popover } from '
...
+			summaryLoading: false,
 			loading: true,
 			message: undefined,
 			errorMessage: '',
 			error: undefined,
 			expandedThreads: [],
 			participantsToDisplay: 999,
 			resizeDebounced: debounce(500, this.updateParticipantsToDisplay),
+			enabledThreadSummary: loadState('mail', 'enabled_thread_summary', false),
+			// TODO: remove this, only a placeholder
```
ChristophWurst, line 101: to do when? ;)

src/components/Thread.vue
```diff
@@ -158,6 +173,15 @@ export default {
 		window.removeEventListener('resize', this.resizeDebounced)
 	},
 	methods: {
+		updateSummary() {
+			if (this.thread.length > 2 && this.enabledThreadSummary) {
+				this.summaryLoading = true
+				  summarizeThread(this.thread[0].threadRootId, this.accountId).then(summary => {
+					this.summaryText = summary
+					this.summaryLoading = false
+				})
```
ChristophWurst, line 182:
```suggestion
				this.summaryLoading = true
				const summary = await summarizeThread(this.thread[0].threadRootId, this.accountId
				this.summaryText = summary
				this.summaryLoading = false

```

src/components/Thread.vue
hamza221: oh I forgot it, It's already changed

ChristophWurst:
Text processing APIs only exist with 27.1, 28 and later. This app supports older releases. So we need to find a way to hide this feature when the API is not there.

The AiIntegrationsService could inject ContainerInterface instead and then do something like

```php
try {
   $manager = $this->container->get(IManager::class);
   $manager->magic();
} catch (ContainerError $e) {
   // hide feature
}
```

ChristophWurst:
Not blocking but very nice to have: unit tests for all new classes

hamza221:
> Not blocking but very nice to have: unit tests for all new classes

I can create an issue and add them in a separate PR? so this gets released asap

hamza221:
Any advice on how to make psalm happy for older versions?

ChristophWurst:
> Any advice on how to make psalm happy for older versions?

Extend https://github.com/nextcloud/mail/blob/main/tests/psalm-baseline.xml

appinfo/routes.php
```diff
@@ -320,6 +320,11 @@
 			'url' => '/api/settings/allownewaccounts',
 			'verb' => 'POST'
 		],
+		[
+			'name' => 'settings#setEnabledThreadSummary',
+			'url' => '/api/settings/threadsummary',
+			'verb' => 'POST'
```
ChristophWurst, line 326: can be a PUT because the operation is idempotent. It would be ok for the frontend to send the same requests more than once with the same result.

appinfo/routes.php
```diff
@@ -360,6 +365,11 @@
 			'url' => '/api/thread/{id}',
 			'verb' => 'POST'
 		],
+		[
+			'name' => 'thread#summarize',
+			'url' => '/api/thread/{id}/summary/',
```
ChristophWurst, line 370:
```suggestion
			'url' => '/api/thread/{id}/summary',
```

lib/Service/AiIntegrationsService.php
```diff
@@ -0,0 +1,73 @@
...
+					return $message->getPreviewText();
+				}, $messages);
+
+				$taskPrompt = implode("\n", $messagesBodies);
+				$summaryTask = new Task(SummaryTaskType::class, $taskPrompt, "mail", $currentUserId, $threadId);
+				$manager->runTask($summaryTask);
+
+				return $summaryTask->getOutput();
+			} else {
+				throw new ServiceException('No language model available for summary');
```
ChristophWurst, line 66: will always be caught two lines below

lib/Service/AiIntegrationsService.php
```diff
@@ -0,0 +1,73 @@
...
+			}
+		} catch (\Throwable $e) {
+			throw new ServiceException('Text processing is not available in your current Nextcloud version');
```
ChristophWurst, line 69: note: this will go but if you throw an exception in a catch block always add the previous exception to the new one to have a holistic stack trace in the logs

src/components/LoadingSkeleton.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<div v-for="i in numberOfLines" :key="i" class="item-list__entry">
-			<div
+			<div v-if="withAvatar"
```
ChristophWurst, line 4: this smells unrelated :D

src/components/Thread.vue
```diff
@@ -38,6 +38,7 @@
 					</div>
 				</div>
 			</div>
+			<ThreadSummary v-if="thread.length>2 && enabledThreadSummary" :loading="summaryLoading" :summary="summaryText" />
```
ChristophWurst, line 41:
```suggestion
			<ThreadSummary v-if="thread.length > 2 && enabledThreadSummary" :loading="summaryLoading" :summary="summaryText" />
```

coding style

lib/Service/AiIntegrationsService.php
```diff
@@ -0,0 +1,73 @@
...
+	public function summarizeThread(string $threadId, array $messages, string $currentUserId): null|string {
+		try {
+			$manager = $this->container->get(IManager::class);
+			if(in_array(SummaryTaskType::class, $manager->getAvailableTaskTypes(), true)) {
+
+				$messagesBodies = array_map(function ($message) {
+					return $message->getPreviewText();
+				}, $messages);
+
+				$taskPrompt = implode("\n", $messagesBodies);
```
marcelklehr, line 60: Sorry for butting in; I saw the PR being referenced and was curious. The summaries might be better if we structure them a little. Something like this for example:

```
Mail from XXX with title "Bla bla bla":

Hello ...


Regards,
XXX

Mail from YYY with title "sdsfg":

...
```

lib/Controller/ThreadController.php
```diff
@@ -111,4 +121,37 @@ public function delete(int $id): JSONResponse {
...
+		} catch (ClientException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		try {
+			$summary = $this->aiIntergrationsService->summarizeThread(
+				$id,
+				$messages,
+				$this->currentUserId,
+			);
```
marcelklehr, line 146: Calling this in a route is risky as it will likely exceed the PHP max execution time. Also, you might want to cache the output. :)

lib/Service/AiIntegrationsService.php
ChristophWurst: That is on our radar :)

We want to make it work first, then fine-tune the model integration.

src/components/LoadingSkeleton.vue
hamza221: It is but it was missing from LoadingSkeleton, I'll create a separate Pr for it

hamza221: #8674

lib/Controller/ThreadController.php
```diff
@@ -111,4 +121,40 @@ public function delete(int $id): JSONResponse {
...
+			$summary = $this->aiIntergrationsService->summarizeThread(
+				$message->getThreadRootId(),
+				$thread,
+				$this->currentUserId,
+			);
+		} catch (\Throwable $e) {
+			$this->logger->error('Summarizing thread failed: ' . $e->getMessage(), [
+				'exception' => $e,
+			]);
+			return new JSONResponse([], Http::STATUS_NO_CONTENT);
```
st3iny, line 154: Hide the whole summary box in case of error. Currently, it will be rendered with no text which looks weird.

src/components/Thread.vue
```diff
@@ -38,6 +38,7 @@
 					</div>
 				</div>
 			</div>
+			<ThreadSummary v-if="thread.length > 2 && enabledThreadSummary" :loading="summaryLoading" :summary="summaryText" />
```
st3iny, line 41: Hide on error here. (See my other comment above for reference.)

src/components/Thread.vue
```diff
@@ -158,6 +172,14 @@ export default {
 		window.removeEventListener('resize', this.resizeDebounced)
 	},
 	methods: {
+		async updateSummary() {
+			if (this.thread.length < 2 && !this.enabledThreadSummary) return
```
st3iny, line 176:
```suggestion
			if (this.thread.length < 2 || !this.enabledThreadSummary) return
```

Should be an or condition, right? Thread is small or summary is disabled globally.

src/components/Thread.vue
```diff
@@ -158,6 +172,14 @@ export default {
...
+
+			this.summaryLoading = true
+			const summary = await summarizeThread(this.thread[0].databaseId)
+			this.summaryText = summary
+			this.summaryLoading = false
```
st3iny, line 181: This is triggered when navigating to another thread where the summary would be disabled. We should probably check the route to see if the user navigates away (to another thread).

Have a look at the following screen cast:

https://github.com/nextcloud/mail/assets/1479486/1fa7ff59-4604-4688-b0a9-5017e67b3627

src/service/SettingsService.js
```diff
@@ -76,3 +76,11 @@ export const updateAllowNewMailAccounts = (allowed) => {
 	}
 	return axios.post(url, data).then((resp) => resp.data)
 }
+
+export const updateEnabledThreadSummary = async (enabled) => {
+	const url = generateUrl('/apps/mail/api/settings/threadsummary')
+	const data = {
+		enabled,
+	}
+	return await axios.post(url, data).then((resp) => resp.data)
```
st3iny, line 85:
```suggestion
	const resp = await axios.post(url, data)
	return resp.data
```

Mixing async/await and callbacks is confusing.

lib/Controller/SettingsController.php
```diff
@@ -119,4 +119,7 @@ public function deleteAntiSpamEmail(): JSONResponse {
 	public function setAllowNewMailAccounts(bool $allowed) {
 		$this->config->setAppValue('mail', 'allow_new_mail_accounts', $allowed ? 'yes' : 'no');
 	}
+	public function setEnabledThreadSummary(bool $enabled) {
```
st3iny, line 122:
```suggestion

	public function setEnabledThreadSummary(bool $enabled) {
```

lib/Controller/ThreadController.php
```diff
@@ -28,27 +28,37 @@
...
-		AccountService $accountService,
-		IMailManager $mailManager) {
+
+	public function __construct(string                $appName,
+		IRequest              $request,
+		string                $UserId,
+		AccountService        $accountService,
+		IMailManager          $mailManager,
+		AiIntegrationsService $aiIntergrationsService,
+		LoggerInterface       $logger,
```
st3iny, line 53: Indentation of parameters is weird. This is probably caused by PhpStorm.

st3iny (changes requested):
Tested and works. However, there are some issues in the code.

src/components/Thread.vue
st3iny: PS: This could also be related to my other comment above (regarding the `||` condition).

lib/Controller/ThreadController.php
ChristophWurst: Solid error handling will be important when someone uses the llm app as text processing backend and it runs into a timeout :+1:

src/components/ThreadSummary.vue
```diff
@@ -0,0 +1,99 @@
+<template>
+	<div class="summary">
+		<div class="summary__header">
+			<div class="summary__header__title">
+				<span class="summary__header__title__brand">
+					<CreationIcon class="summary__header__title__brand__icon" />
+					<p>Nextcloud Assistant</p>
```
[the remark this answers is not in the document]
hamza221: @jancborchardt @nimishavijay what do you think about this? should it come from theming, and/or should it be translated ?

ChristophWurst (commented):
Tested without any llm backend

Ideally we hide the whole feature when there is no text summarization support. That is hide the admin settings and the summary box above threads.

src/components/ThreadSummary.vue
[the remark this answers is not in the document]
nimishavijay: I would say yes, without theming it would look confusing to users who have never interacted with the Nextcloud brand on their instance. And yes it should be translated. But double checking with @jancborchardt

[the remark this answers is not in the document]
jancborchardt: Yep, what @nimishavijay said! :)

ChristophWurst:
I found myself in a funny situation. I have thread summaries enabled since yesterday. But there is no UI to disable them.

Nextcloud users could run into this as well if they previously had an llm integration enabled but turned it off.

src/service/SettingsService.js
```diff
@@ -76,3 +76,19 @@ export const updateAllowNewMailAccounts = (allowed) => {
...
+	const url = generateUrl('/apps/mail/api/settings/threadsummary')
+	const data = {
+		enabled,
+	}
+	const resp = await axios.put(url, data)
+	return resp.data
+}
+
+export const isLlmConfigured = async () => {
+	const url = generateUrl('/apps/mail/api/settings/llmconfigured')
```
ChristophWurst, line 90: nit: we can save one request at page load if this retrieval is moved into initial state

lib/Controller/PageController.php
```diff
@@ -244,6 +244,11 @@ public function index(): TemplateResponse {
 			$this->config->getAppValue('mail', 'allow_new_mail_accounts', 'yes') === 'yes'
 		);
 
+		$this->initialStateService->provideInitialState(
```
ChristophWurst, line 247: always set summaries to `no` if llm apis are not available

ChristophWurst (commented):
Requesting small changes one last time

Solid PR :+1: :sunglasses:

ChristophWurst (approved):

jakobroehrl:
> image image
> 
> To Test the setting should be enabled first in admin settings
> 
> closes #8508

@hamza221 I updated the Mail App to 3.4.0-beta.2 but I can't find the "enable thread setting" in my groupware settings. 
grafik

Could you help?

st3iny:
The necessary parts in the server are not yet released. You have to be a bit more patient (or install a beta at your own risk).

ChristophWurst:
https://github.com/nextcloud/documentation/pull/11025

PR #8659, opened 2023-07-31
Update changelog on main

GretaD:
A lot of leftovers from previews releases. Sorry :(
I copy paste from nc-release repo, so no need to prof read.


st3iny (approved):
:broom:

PR #8674, opened 2023-08-05
Add loading skeletons without avatars

hamza221:
With avatar
image
Without avatar 
image

### DIFF
--- a/src/components/LoadingSkeleton.vue
+++ b/src/components/LoadingSkeleton.vue
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<div v-for="i in numberOfLines" :key="i" class="item-list__entry">
-			<div
+			<div v-if="withAvatar"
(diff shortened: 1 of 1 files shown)


GretaD (approved):

ChristophWurst:
What does this change and is it a fix or a feature? Do we need backports?

hamza221:
> What does this change and is it a fix or a feature? Do we need backports?

The condition was missing, to show loading skeletons without avatars. but the prop was already there. 
I'll add screenshots for with/without skeletons. 
We need backports I think, i'll check when this was introduced and add them

hamza221:
@ChristophWurst we never needed this use case before thread summary, even tho the condition is missing. I don't think the backport is really needed. what do you think?

ChristophWurst:
skipping backports sounds fine

PR #8686, opened 2023-08-08
fix(proxy): Add image proxy rate limit

ChristophWurst:
Avoid users sending too many requests through the server.

Ref https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#rate-limiting

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -19,6 +19,11 @@
 		<directory name="lib/Vendor" />
     </extraFiles>
 	<issueHandlers>
+		<UndefinedAttributeClass>
+			<errorLevel type="info">
+				<referencedClass name="OCP\AppFramework\Http\Attribute\UserRateLimit" />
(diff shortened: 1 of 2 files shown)


kesselb (approved):
Make sense :+1:

ChristophWurst:
Psalm needs to be silenced about the missing attribute for older versions

psalm.xml
```diff
@@ -19,6 +19,11 @@
 		<directory name="lib/Vendor" />
     </extraFiles>
 	<issueHandlers>
+		<UndefinedAttributeClass>
+			<errorLevel type="info">
```
ChristophWurst, line 23: I'm testing the consequences of using this over *suppress*

ChristophWurst: Psalm passes :thinking: :tada: :thinking:

Issue #2048, opened 2019-09-29
Snooze mails


jakobroehrl:
## Feature Request
To keep the inbox clean and clear it would be nice to snooze mails for x hours or y days.
Often the inbox is a kind of todo list. So if you can't do the todo right now you can snooze it for later.
Of I you send an importand mail you will keep track of it. So you can snooze the send mail and it will apear in the future, so you can ask about the status or send the demand again. 

## Summary
snozze a mail and move it automaticly back to the inbox after the choosen snozze time has expired

- [x] Requires https://github.com/nextcloud/mail/issues/3664

## Work packages

* [x] ~~Shallow prototype (in case MVP is not doable for feature freeze https://github.com/nextcloud/groupware/issues/70)~~
  * [x] ~~Add an action for *Snooze* to messages but only temporarily remove the message in the frontend~~
* [x] MVP
  * Add action *Snooze* with a few simple time ranges that moves the selected message to a dedicated "Snoozed" mailbox. System saves the timestamp until when the messages is snoozed.
  * Background job watches messages in Snoozed/Later and moves them back to INBOX when the time expires (source mailbox before snooze is not saved)
  * Message moved back to INBOX is sorted as usual (not yet at the top)
  * ~~Nextcloud notification for expired snoozes~~
* [x] Snoozed mailbox created at first use. Mailbox name is localized in UI. https://github.com/nextcloud/mail/issues/8742
* [x] ~~More detailed time ranges via a custom date picker~~ part of the mvp
* [x] https://github.com/nextcloud/mail/issues/8745
* [x] ~~Sort message back to the top~~ will not be done
* [x] https://github.com/nextcloud/documentation/issues/10968
* [x] https://github.com/nextcloud/mail/issues/8758
* [x] https://github.com/nextcloud/mail/issues/8763

## Follow-ups
* Move messages back to their original mailbox (not the INBOX) https://github.com/nextcloud/mail/issues/8779
* https://github.com/nextcloud/mail/issues/8744
* https://github.com/nextcloud/mail/issues/8746
* Interoperability
  * Look into options to make the Snooze/Later mailbox discoverable for us and other clients with a dedicated IMAP attribute similar to `\sent`, `\draft`, etc.
  * Look into options to push the snooze time to IMAP for other clients to read

ChristophWurst:
This requires us to store more of the IMAP state locally.

szaimen:
Maybe you could do it  the way Microsoft is doing it in Outlook on Android: 
[...]

PR #8694, opened 2023-08-09
feat: add snooze mvp

JohannesGGE:
fix: #2048

### DIFF
--- a/lib/Controller/ThreadController.php
+++ b/lib/Controller/ThreadController.php
@@ -29,6 +29,7 @@ use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\Http\TrapError;
 use OCA\Mail\Service\AccountService;
 use OCA\Mail\Service\AiIntegrationsService;
+use OCA\Mail\Service\SnoozeService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
@@ -40,6 +41,7 @@ class ThreadController extends Controller {
 	private string $currentUserId;
 	private AccountService $accountService;
 	private IMailManager $mailManager;
+	private SnoozeService $snoozeService;
 	private AiIntegrationsService $aiIntergrationsService;
 	private LoggerInterface $logger;
 
@@ -49,12 +51,14 @@ class ThreadController extends Controller {
 		string $UserId,
 		AccountService $accountService,
 		IMailManager $mailManager,
+		SnoozeService $snoozeService,
 		AiIntegrationsService $aiIntergrationsService,
 		LoggerInterface $logger) {
 		parent::__construct($appName, $request);
 		$this->currentUserId = $UserId;
 		$this->accountService = $accountService;
 		$this->mailManager = $mailManager;
+		$this->snoozeService = $snoozeService;
 		$this->aiIntergrationsService = $aiIntergrationsService;
 		$this->logger = $logger;
 	}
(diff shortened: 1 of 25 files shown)


lib/BackgroundJob/UnSnoozeJob.php
```diff
@@ -0,0 +1,127 @@
...
+		try {
+			$snoozeMailbox = $this->mailboxMapper->findById($snoozeMailboxId);
+		} catch (DoesNotExistException $e) {
+			return;
+		}
+
+		$now = $this->time->getTime();
+		$messages = $this->messageMapper->findMessagesToUnSnooze(
+			$snoozeMailboxId,
+			$now,
```
ChristophWurst, line 101: the job runs once an hour. if you snooze a message until *tomorrow morning, 9am* and the job always runs at :45, the message will be woken at 9:45 only.

idea: consider 30min look-ahead for the background job. then messages are woken +-30min of the target date

ChristophWurst: the job runs every 60 *seconds*. there is no delay.

ChristophWurst (commented):
Early architectural nitpicking :heavy_check_mark:

lib/BackgroundJob/UnSnoozeJob.php
```diff
@@ -0,0 +1,127 @@
...
+				$this->unsnoozeMails($account);
+			} catch (ServiceException|ClientException $e) {
+				$this->logger->error('Could not unSnooze Mails', [
+					'exception' => $e,
+					'userId' => $account->getUserId(),
+					'accountId' => $account->getId(),
+					'snoozeMailboxId' => $account->getMailAccount()->getSnoozeMailboxId(),
+				]);
+			}
+		}
```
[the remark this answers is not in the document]
JohannesGGE: I added a SnoozeService and put the logic there, let me know if it's done correctly.
I was also not sure if we put an Interface here (ISnoozeService). Some services have one, others don't.

lib/Controller/ThreadController.php
```diff
@@ -84,6 +88,27 @@ public function move(int $id, int $destMailboxId): JSONResponse {
...
+	 */
+	#[TrapError]
+	public function snooze(int $id, int $unixTimestamp): JSONResponse {
+		try {
+			$selectedMessage = $this->mailManager->getMessage($this->currentUserId, $id);
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		$this->snoozeService->snoozeThread($selectedMessage, $unixTimestamp);
```
ChristophWurst, line 107: pretty cool that a whole thread can be snoozed as well :)

lib/Db/ThreadMapper.php
```diff
@@ -70,4 +70,26 @@ public function findMessageUidsAndMailboxNamesByAccountAndThreadRoot(MailAccount
...
+			->where(
+				$qb->expr()->eq('messages.thread_root_id', $qb->createNamedParameter($threadRootId, IQueryBuilder::PARAM_STR)),
+			);
+
+		$result = $qb->executeQuery();
+		$rows = array_map(static function (array $row) {
+			return [
+				'messageId' => (string)$row[0]
+			];
+		}, $result->fetchAll(\PDO::FETCH_NUM));
```
ChristophWurst, line 90: https://www.php.net/manual/en/function.array-column.php

lib/Migration/Version3300Date20230807300513.php
```diff
@@ -0,0 +1,77 @@
...
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version3300Date20230807300513 extends SimpleMigrationStep {
```
ChristophWurst, line 35:
```suggestion
class Version3400Date20230807300513 extends SimpleMigrationStep {
```

because the next minor will be 3.4. 3.3 has already been branched off

ChristophWurst (approved):
Code looks super clean! Will also give this a test shortly

ChristophWurst (commented):
bump the version in info.xml so that the backrground job is inserted and the migration run automatically

ChristophWurst (approved):
Works!

Message can be snoozed (moved) :heavy_check_mark: 
When timestamp expires (updated the database) the message move back :heavy_check_mark:

ChristophWurst:
:grimacing: needs rebase onto latest main to get rid of the conflicts

lib/Migration/Version3400Date20230807300513.php
```diff
@@ -0,0 +1,77 @@
...
+			$messagesRetentionTable->addColumn('message_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 1024,
+			]);
+			$messagesRetentionTable->addColumn('snoozed_until', Types::INTEGER, [
+				'notnull' => true,
+				'length' => 4,
+			]);
+			$messagesRetentionTable->setPrimaryKey(['id'], 'mail_msg_snoozed_id_idx');
+			$messagesRetentionTable->addIndex(['message_id'], 'mail_msg_snoozed_msgid_idx');
```
ChristophWurst, line 72: should we make it unique? two entries don't make any sense

ChristophWurst (approved):
:shipit: Otherwise

PR #8697, opened 2023-08-09
Fix alignment for settings buttons and it's icons

### LINKED ISSUE #8681: Button to register mailto handler needs line wrap
### Steps to reproduce

1. Open the app
2. Open the settings menu bottom left
3. Look

### Expected behavior

Good design

### Actual behavior

1) No line wrap
2) Buttons could have consistent, full width

image

### Mail app version

v3.3.0-alpha.3

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

auroraanna:
Fixes #8681

Before:
the buttons and icons within are not aligned nicely
After: 
the buttons and icons within are nicely aligned now and the butttons all have the same width

The buttons are all the same width now and the previous overflowing text is not cut off by and ellipsis anyone but broken into a new line. Also that lock icon doesn't have a margin-right anymore which made it not align with the other icons.

@GretaD said the buttons should maybe be the same width as the other ones in the sidebar outside of the mail settings but I don't agree. I think the buttons being smaller makes it clear that they're contained in the mail settings.

cc: @jancborchardt @nimishavijay What is your opinion on the design?

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -275,15 +275,24 @@ p.app-settings {
 	background-position: 10px center;
 	text-align: left;
 	margin-top: 6px;
+	width: 100%;
 }
 .app-settings-button.button.primary.new-button {
 	color: var(--color-primary-element-text);
 	//this style will be removed after we migrate also the  'add mail account' to material design
 	padding-left: 34px;
 	gap: 4px;
+	width: fit-content;
 }
 .app-settings-link {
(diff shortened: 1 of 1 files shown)


jancborchardt (approved):
Very nice fix, looks good @auroraanna! And welcome! :)

GretaD (approved):
if the design team is fine with smaller buttons, so am i :)

PR #8709, opened 2023-08-10
fix(setup): Rate limit auto config attempts

ChristophWurst:
If a user sends more attempts they either have incorrect options anyway or just try to overload the server.

## How to test

1) Open the app
2) Click *+ New account* in the navigation
3) Enter some invalid email address
4) Hammer the *Connect* button

`main`: You can try endlessly
here: You can try a bunch of times, then you see a nice error message:

image


GretaD (approved):
works

PR #8716, opened 2023-08-10
Add unit tests for thread summary feature

hamza221:
ref https://github.com/nextcloud/mail/issues/8508

### DIFF
--- a/tests/Unit/Controller/ThreadControllerTest.php
+++ b/tests/Unit/Controller/ThreadControllerTest.php
@@ -242,4 +242,50 @@ class ThreadControllerTest extends TestCase {
 
 		$this->assertEquals(Http::STATUS_OK, $response->getStatus());
 	}
+
+	public function testSummarizeThread(): void {
+		$mailAccount = new MailAccount();
+		$mailAccount->setId(1);
+		$this->accountService
+				->expects(self::once())
+				->method('find')
+				->with('john', 1)
+				->willReturn(new Account($mailAccount));
+		$mailbox = new Mailbox();
+		$mailbox->setId(20);
+		$mailbox->setAccountId($mailAccount->getId());
+		$this->mailManager
+			->method('getMailbox')
+			->willReturn($mailbox);
+		$message1 = new Message();
+		$message1->setId(300);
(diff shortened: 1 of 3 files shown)


tests/Unit/Controller/ThreadControllerTest.php
```diff
@@ -237,4 +237,48 @@ public function testDeleteTrash(): void {
 
 		$this->assertEquals(Http::STATUS_OK, $response->getStatus());
 	}
+
+	public function testSummarizeThread(): void {
+		$mailAccount = new MailAccount();
+		$mailAccount->setId(1);
+		$this->accountService
+			->method('find')
```
ChristophWurst, line 245: nit: if you add `->with(...)` you can assert it's looking for the expected id

tests/Unit/Service/AiIntegrationsServiceTest.php
```diff
@@ -0,0 +1,86 @@
...
+		parent::setUp();
+		$this->container = $this->createMock(ContainerInterface::class);
+		$this->manager = $this->createMock(IManager::class);
+
+		$this->aiIntegrationsService = new AiIntegrationsService(
+			$this->container
+		);
+	}
+
+	public function testSummarizeThreadNoBackend() {
```
ChristophWurst, line 59: add a test where `get` throws

tests/Unit/Controller/ThreadControllerTest.php
hamza221: you mean ->with() on the accountService?

ChristophWurst: yes

ChristophWurst: ``$this->accountService->expect(self::once())->method('find')->with(1)->willReturn(new Account($mailAccount));``

then you know
1) it's been called exactly once
2) it's been called with the expected inputs

ChristophWurst:
26 fails with `Object of class OCA\Mail\Exception\ServiceException could not be converted to int`

kesselb (approved):

PR #8717, opened 2023-08-10
Add Caching to Thread summaries

hamza221:
Ref  #8508

### DIFF
--- a/lib/Service/AiIntegrations/Cache.php
+++ b/lib/Service/AiIntegrations/Cache.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2023 Hamza Mahjoubi<hamzamahjoubi22@proton.met>
+ *
+ * @author 2023 Hamza Mahjoubi <hamzamahjoubi22@proton.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


lib/Service/AiIntegrations/Cache.php
```diff
@@ -0,0 +1,79 @@
...
+class Cache {
+	// Cache for one week
+	public const CACHE_TTL = 7 * 24 * 60 * 60;
+
+	/** @var ICache */
+	private $cache;
+
+
+	public function __construct(ICacheFactory $cacheFactory) {
+		$this->cache = $cacheFactory->createDistributed('mail.ai');
```
ChristophWurst, line 41: distributed caches are always more expensive than local ones because of the net connection they have to establish. local caches can use sockets.

let's use a local cache instance for this. in the worst case one thread is summarized a second time. so be it.

lib/Service/AiIntegrations/Cache.php
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	/**
+	 * @param array $ids
+	 * @return string
+	 */
+	private function buildUrlKey(array $ids): string {
+		return base64_encode(json_encode($ids));
```
ChristophWurst, line 49: :+1: cache key changes when there is a new message in the thread

lib/Service/AiIntegrations/Cache.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2017 Hamza Mahjoubi<hamzamahjoubi22@proton.met>
```
ChristophWurst, line 6: time traveler?

ChristophWurst (approved):
Bonus points for unit tests that verify that thread summaries are not fetched from the text APIs if cached

hamza221:
I'll add the test to https://github.com/nextcloud/mail/pull/8716 cause it makes sense to add it the service's Tests.

ChristophWurst (approved):

Issue #8724, opened 2023-08-11
Can't snooze mails on Gmail


ChristophWurst:
### Steps to reproduce

1. Set up a Gmail account
2. Create and assign a *Snoozed* mailbox
3. Snooze a message

### Expected behavior

:sleeping: 

### Actual behavior

:open_mouth: 

server errors. I lack details because I don't have access to the backend or logs of the instance.

### Mail app version

3.4.0-beta.2

### Mailserver or service

Gmail

### PHP engine version

None

### Web server

None

### Database

None

PR #8728, opened 2023-08-11
fix(Snooze): Allow snoozing gmail messages

JohannesGGE:
fix: #8724 

Due to the message appearing twice in the account (a folder which holds all mails) the snooze thread tries to insert the message twice in the db. Column is unique, so the server respons with error.

Also interesting for further implementation/bugfixing: When the message is snoozed and moved back by the normal move, the message can't be snoozed again because the db entry won't be deleted (probably because the move fails in the wake job).

### DIFF
--- a/lib/Db/ThreadMapper.php
+++ b/lib/Db/ThreadMapper.php
@@ -80,7 +80,7 @@ class ThreadMapper extends QBMapper {
 			->from($this->tableName, 'messages')
 			->where(
 				$qb->expr()->eq('messages.thread_root_id', $qb->createNamedParameter($threadRootId, IQueryBuilder::PARAM_STR)),
-			);
+			)->groupBy('messages.message_id');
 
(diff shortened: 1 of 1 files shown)


lib/Service/SnoozeService.php
```diff
@@ -108,6 +108,8 @@ public function snoozeThread(Message $selectedMessage, int $unixTimestamp): void
 			$selectedMessage->getThreadRootId()
 		);
 
+		$messages = array_unique($messages);
```
ChristophWurst, line 111: could it make sense to adjust `\OCA\Mail\Db\ThreadMapper::findMessageIdsByThreadRoot` with a `GROUP BY` and let the database deduplicate?

it's a nano optimization because there will probably rarely be more than one duplicate

JohannesGGE: :white_check_mark:

Issue #8747, opened 2023-08-14
Snooze MVP follow up


JohannesGGE:
- [x] Update snooze options aligned to https://github.com/nextcloud/spreed/pull/10177
- [x] Set alarm icon for selected dedicated mailbox (not just name match "Snoozed")
- [x] Just one "snooze" request to the backend (not snooze and move)
- [x] Enable snoozing message if DB entry for the message already exists
- [x] Remove DB entry for the message if moving fails
- [x] Move snooze logic to SnoozeService

PR #8754, opened 2023-08-15
Snooze mvp follow up

JohannesGGE:
Fix: #8747 

This follow up is fixing some minor things:

- Update snooze options aligned to https://github.com/nextcloud/spreed/pull/10177
- Set alarm icon for selected dedicated mailbox (not just name match "Snoozed")
- Only one "snooze" request to the backend (not snooze and move)
- Enable snoozing message if DB entry for the message already exists
- Remove DB entry for the message if moving fails
- Move snooze logic to SnoozeService

### DIFF
--- a/lib/Service/SnoozeService.php
+++ b/lib/Service/SnoozeService.php
@@ -29,6 +29,7 @@ namespace OCA\Mail\Service;
 use OCA\Mail\Account;
 use OCA\Mail\Contracts\IMailManager;
 use OCA\Mail\Db\MailAccountMapper;
+use OCA\Mail\Db\Mailbox;
 use OCA\Mail\Db\MailboxMapper;
 use OCA\Mail\Db\Message;
 use OCA\Mail\Db\MessageMapper;
@@ -40,7 +41,9 @@ use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\IMAP\IMAPClientFactory;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\DB\Exception;
 use Psr\Log\LoggerInterface;
+use Throwable;
 
 class SnoozeService {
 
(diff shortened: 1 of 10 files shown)


lib/Service/SnoozeService.php
```diff
@@ -81,6 +83,79 @@ public function wakeMessages(): void {
...
+			$this->mailManager->moveMessage(
+				$srcAccount,
+				$srcMailbox->getName(),
+				$message->getUid(),
+				$dstAccount,
+				$dstMailbox->getName()
+			);
+		} catch (\Throwable $e) {
+			$this->unSnoozeMessageDB($message);
+			throw new ServiceException($e->getMessage());
```
ChristophWurst, line 118: loses all of the previous error trace. either just rethrow the original throwable or pass the original as *previous* of the new service exception

lib/Service/SnoozeService.php
```diff
@@ -81,6 +83,79 @@ public function wakeMessages(): void {
...
+			$this->mailManager->moveThread(
+				$srcAccount,
+				$srcMailbox,
+				$dstAccount,
+				$dstMailbox,
+				$selectedMessage->getThreadRootId()
+			);
+		} catch (\Throwable $e) {
+			$this->unSnoozeThreadDB($selectedMessage);
+			throw new ServiceException($e->getMessage());
```
ChristophWurst, line 154: same as above

lib/Service/SnoozeService.php
```diff
@@ -89,11 +164,26 @@ public function wakeMessages(): void {
...
 	 */
-	public function snoozeMessage(Message $message, int $unixTimestamp): void {
+	public function snoozeMessageDB(Message $message, int $unixTimestamp): void {
 		$snooze = new MessageSnooze();
 		$snooze->setMessageId($message->getMessageId());
 		$snooze->setSnoozedUntil($unixTimestamp);
-		$this->messageSnoozeMapper->insert($snooze);
+		try {
+			$this->messageSnoozeMapper->insert($snooze);
+		} catch(Exception $e) {
```
ChristophWurst, line 173: can you check the reason of the error and only handle duplicates and rethrow the rest?

lib/Service/SnoozeService.php
```diff
@@ -112,7 +202,28 @@ public function snoozeThread(Message $selectedMessage, int $unixTimestamp): void
...
+	 * @return void
+	 */
+	public function unSnoozeThreadDB(Message $selectedMessage): void {
+		$messages = $this->threadMapper->findMessageIdsByThreadRoot(
+			$selectedMessage->getThreadRootId()
+		);
+
+		foreach ($messages as $message) {
+			$this->messageSnoozeMapper->deleteByMessageId($message['messageId']);
 		}
```
ChristophWurst, line 227: perf: would be faster and atomic if this is done in one query by the db

lib/Service/SnoozeService.php
JohannesGGE: :white_check_mark:

lib/Service/SnoozeService.php
JohannesGGE: :white_check_mark:

lib/Service/SnoozeService.php
JohannesGGE: Added a check `$e->getPrevious() instanceof \Doctrine\DBAL\Exception\UniqueConstraintViolationException`. Do you think this is okay?

lib/Service/SnoozeService.php
JohannesGGE: Changed `deleteByMessageId()` to take array of ids and query db by where->in

lib/Service/SnoozeService.php
JohannesGGE: Apparently, tests fail because the UniqueConstraintViolationException is part of 3rdparty. Is there a way to use it without errors or is this not an option?

ChristophWurst: ``\OCP\DB\Exception::getReason`` and comparison with ``\OCP\DB\Exception::REASON_UNIQUE_CONSTRAINT_VIOLATION``

lib/Service/SnoozeService.php
```diff
@@ -144,6 +285,7 @@ private function wakeMessagesByAccount(Account $account): void {
 
 		$client = $this->clientFactory->getClient($account);
 		try {
+			$messageIdsToDelete = array();
```
ChristophWurst, line 288:
```suggestion
			$messageIdsToDelete = [];
```

ChristophWurst (approved):
Looks good!

Issue #8745, opened 2023-08-14
Deactivate Snooze on Ajax background trigger


JohannesGGE:
### Is your feature request related to a problem? Please describe.

Contributes to: https://github.com/nextcloud/mail/issues/2048

### Describe the solution you'd like

If Ajax is selected as a background-job trigger, the snooze feature should be deactivated due to unreliable wake job execution.

PR #8759, opened 2023-08-16
feat(snooze): Disable snooze on AJAX cron

JohannesGGE:
Fix: #8745

Don't allow to snooze mails if AJAX is selected as background cron trigger.


ChristophWurst (approved):

ChristophWurst:
nitpick on the sematic commit message: this is a `fix` :)

st3iny (approved):

PR #8761, opened 2023-08-16
fix: remove moveJunk from integration test as well

kesselb:
Screenshot from 2023-08-16 16-06-14


kesselb:
Replaced by https://github.com/nextcloud/mail/pull/8759

PR #8762, opened 2023-08-16
Added "Mark as spam", "Mark as not spam", "Edit tags" buttons to multiselect menu

### LINKED ISSUE #7006: Multiselect menu missing "Mark as spam" and "Tags"
Whereas the menu on single mails has "Mark as spam" and "Tags", the multiselect menu is missing those:

Single menu | Multiselect menu
-|-
Single select menu
Multiselect menu

(Latest stable Mail version, on cloud.nc)

auroraanna:
Fixes #7006

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -96,6 +96,46 @@
 						}}
 					</ActionButton>
 					<ActionButton
+						v-if="isAtLeastOneSelectedNotJunk"
+						:close-after-click="true"
+						@click.prevent="markSelectionJunk">
+						<template #icon>
+							<AlertOctagonIcon
+								:size="20" />
+						</template>
+						{{
+							n(
+								'mail',
+								'Mark {number} as spam',
+								'Mark {number} as spam',
+								selection.length,
+								{
+									number: selection.length,
+								}
+							)
+						}}
+					</ActionButton>
+					<ActionButton
+						v-if="isAtLeastOneSelectedJunk"
+						:close-after-click="true"
+						@click.prevent="markSelectionNotJunk">
+						<template #icon>
+							<AlertOctagonIcon
+								:size="20" />
+						</template>
+						{{
+							n(
+								'mail',
+								'Mark {number} as not spam',
+								'Mark {number} as not spam',
+								selection.length,
+								{
+									number: selection.length,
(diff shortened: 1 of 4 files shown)


auroraanna:
Showing the "Mark … as spam" and "Mark … as not spam" buttons conditionally is working, but the don't do anything yet. I'm confused on how to write the methods for them.

There seem to only exist toggle actions for junk. For these button, just toggling doesn't work, they all need to be set to one state.

I haven't created the "Edit tags of …" button yet.

auroraanna:
The spam button work now. The "Edit tags" button doesn't because the TagModal component only works for one envelope.

auroraanna:
"Edit tags" button should be working too now.

src/components/EnvelopeList.vue
```diff
@@ -367,6 +448,28 @@ export default {
 			})
 			this.unselectAll()
 		},
+		markSelectionJunk() {
+			this.selectedEnvelopes.forEach((envelope) => {
+				if (!envelope.flags.$junk) {
+					this.$store.dispatch('toggleEnvelopeJunk', {
+						envelope,
+						removeEnvelope: true,
```
kesselb, line 456:
```suggestion
						removeEnvelope: await this.$store.dispatch('moveEnvelopeToJunk', envelope),
```

Unfortunately, async/await does not work inside forEach: https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of work well with async/await.

kesselb:
Very well, thank you :pray: 

Having a junk folder is optional. If you have one and mark a message as junk, the message is moved, and otherwise it stays in the current folder. 

<table>
<tr>
 <td>removeEnvelope
 <td>outcome
<tr>
 <td>true
 <td>envelope will be moved to the junk folder
<tr>
 <td>false
 <td>envelope will stay in the current folder
</table>

I will send a pull request next week to add some inline documentation for toggleEnvelopeJunk action. 

The scenario "user does not have a junk folder" does not work properly. 
In this case, removeEnvelope should be false to not remove the envelope from the store.

GretaD (approved):
very nice :100:

Issue #8763, opened 2023-08-16
Snooze: Clean up leftover DB entries


JohannesGGE:
### Steps to reproduce

1. Snooze a message
2. Move the snoozed message somewhere else by the "Move" action

### Expected behavior

The DB entry for the message in the snoozed table should be deleted (by clean up job after one month begining from wake date)

### Actual behavior

The DB entry stays forever.

### PHP engine version

None

### Web server

None

### Database

None

PR #8765, opened 2023-08-16
fix(snooze): Add cleanup orphan db entries

JohannesGGE:
Fix: #8763 

Adds a cleanup call to delete all db entries in the snooze table that should have been unsnoozed 30d ago. We assume that these messages no longer exist in the snoozed mailbox.

### DIFF
--- a/lib/Db/MessageSnoozeMapper.php
+++ b/lib/Db/MessageSnoozeMapper.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace OCA\Mail\Db;
 
 use OCP\AppFramework\Db\QBMapper;
+use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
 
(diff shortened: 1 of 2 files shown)


lib/Db/MessageSnoozeMapper.php
```diff
@@ -57,4 +58,19 @@ public function deleteByMessageId(string $messageId): void {
...
+	/**
+	 * Delete all orphaned entries that should have been unsnoozed a month ago.
+	 * We assume that these messages no longer exist in the snoozed mailbox.
+	 */
+	public function deleteOrphans(): void {
+		$deleteQb = $this->db->getQueryBuilder();
+		$deleteQb->delete($this->getTableName())
+			->where($deleteQb->expr()->lt(
+				'snoozed_until',
+				$deleteQb->createNamedParameter(($this->time->getTime() - 2592000), IQueryBuilder::PARAM_INT),
```
ChristophWurst, line 71:
```suggestion
				$deleteQb->createNamedParameter(($this->time->getTime() - (30*7*24*3600)), IQueryBuilder::PARAM_INT),
```

for readability and the de-optimization won't hurt

ChristophWurst (approved):

lib/Db/MessageSnoozeMapper.php
JohannesGGE: :white_check_mark: Removed the 7*, otherwise we would wait 7 months ;)

Issue #8742, opened 2023-08-14
Create Snooze Mailbox on first Snooze


JohannesGGE:
### Is your feature request related to a problem? Please describe.

Contributes to: https://github.com/nextcloud/mail/issues/2048

### Describe the solution you'd like

A dedicated mailbox "Snoozed" is created automatically on the first snooze operation

### Describe alternatives you've considered

Currently the mailbox has to be created and selected as snooze mailbox manually

PR #8771, opened 2023-08-17
fix(snooze): create snooze mailbox on first snooze

JohannesGGE:
Fix: #8742

Creates a new "Snoozed" and select it as dedicated snooze mailbox. If a mailbox "Snoozed" already exists, this mailbox ist selected.


ChristophWurst (commented):
Replace .then with awaits please

src/components/Envelope.vue
```diff
@@ -790,6 +790,37 @@ export default {
 			// Remove from selection first
 			this.setSelected(false)
 
+			if (!this.account.snoozeMailboxId) {
+				const name = 'Snoozed'
+				let snoozeMailboxId
+
+				const createMailboxResponse = await this.$store
+					.dispatch('createMailbox', { account: this.account, name })
+					.catch((error) => {
```
ChristophWurst, line 799: .catch -> try-catch :laughing:

ChristophWurst (commented):
Move the logic into a Vuex action to deduplicate the two code fragments

:+1: then

ChristophWurst (approved):
Looks good

Issue #8746, opened 2023-08-14
Show "Snoozed" Mailbox as primary mailbox


JohannesGGE:
### Is your feature request related to a problem? Please describe.

Contributes to: https://github.com/nextcloud/mail/issues/2048

### Describe the solution you'd like

The "Snoozed" mailbox should be shown as primary mailbox like the Inbox and Favourites

PR #8774, opened 2023-08-18
fix(Snooze): Sort snooze mailbox as specialUse

JohannesGGE:
Fix: #8746 

If a dedicated snooze mailbox is selected in the account settings, it is shown as a special use mailbox before none special mailboxes

|before|after|
|---|---|
|image|image|


hamza221 (approved):

st3iny (approved):
Didn't test but looks good.

st3iny (changes requested):
Frontend tests are failing :eyes:

st3iny (approved):

Issue #8779, opened 2023-08-18
Snooze: Move message back to src mailbox


JohannesGGE:
### Is your feature request related to a problem? Please describe.

Currently the message is moved to the INBOX on wake

### Describe the solution you'd like

The message is moved back to the original mailbox.
If the mailbox has been deleted, move back to inbox.

### Additional context

Contributes to: https://github.com/nextcloud/mail/issues/2048

PR #8780, opened 2023-08-18
fix(Snooze): Move the message back to src mailbox on wake

JohannesGGE:
Fix: #8779 

Moves the message back to the original mailbox instead of INBOX on wake.
(If the original mailbox is gone, back to INBOX)

### DIFF
--- a/lib/Db/MessageSnoozeMapper.php
+++ b/lib/Db/MessageSnoozeMapper.php
@@ -40,6 +40,39 @@ class MessageSnoozeMapper extends QBMapper {
 	}
 
 	/**
+	 * Returns srcMailboxId (before snooze) for message
+	 * Return null if no entry for message or srcMailboxId is null
+	 *
+	 * @param string $messageId
+	 *
+	 * @return int|null
+	 */
+	public function getSrcMailboxId(string $messageId): ?int {
+		$qb = $this->db->getQueryBuilder();
+
+		$select = $qb->select('src_mailbox_id')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->eq(
+					'message_id',
+					$qb->createNamedParameter($messageId, IQueryBuilder::PARAM_STR),
+					IQueryBuilder::PARAM_STR
+				),
+				$qb->expr()->isNotNull(
+					'src_mailbox_id'
+				),
+			);
+
+		$result = $select->executeQuery();
+		$row = $result->fetch();
+		$result->closeCursor();
+		if ($row === false) {
+			return null;
+		}
+		return (int)$row['src_mailbox_id'];
+	}
+
+	/**
 	 * Deletes DB Entry for a waked message
 	 *
 	 * @param string[] $messageIds
--- a/lib/Migration/Version3400Date20230818160236.php
+++ b/lib/Migration/Version3400Date20230818160236.php
@@ -24,32 +24,34 @@ declare(strict_types=1);
  *
  */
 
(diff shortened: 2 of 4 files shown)


lib/Service/SnoozeService.php
```diff
@@ -287,12 +289,25 @@ private function wakeMessagesByAccount(Account $account): void {
 		try {
 			$messageIdsToDelete = [];
 			foreach ($messages as $message) {
+				$srcMailboxId = $this->messageSnoozeMapper->getSrcMailboxId($message->getMessageId());
```
kesselb, line 292: The additional roundtrip for srcMailboxId is a bit sad.

Idea 1: Introduce SnoozedMessage and SnoozedMessageMapper. SnoozedMessage should extend Message. SnoozedMessageMapper should extend QBMapper. That's nice, but also a bit weird because SnoozedMessageMapper inherits a couple of methods from QBMapper which we need to adjust. 

Idea 2: Don't use a DTO at all. For the unsnooze operation, we need the srcMailboxId and message uid. I think it's okay to change findMessagesToUnSnooze to return an array. I think that's much simpler than idea 1. 

If you do not have time, I'm happy to continue with the current iteration.

lib/Migration/Version3400Date20230818160236.php
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$snoozeTable = $schema->getTable('mail_messages_snoozed');
+		if (!$snoozeTable->hasColumn('src_mailbox_id')) {
+			$snoozeTable->addColumn('src_mailbox_id', Types::INTEGER, [
+				'notnull' => false,
+				'length' => 4,
```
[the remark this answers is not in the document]
JohannesGGE: :white_check_mark:

lib/Db/MessageSnoozeMapper.php
```diff
@@ -39,6 +39,34 @@ public function __construct(IDBConnection $db, private ITimeFactory $time) {
...
+		$select = $qb->select('src_mailbox_id')
+			->from($this->getTableName())
+			->where($qb->expr()->eq(
+				'message_id',
+				$qb->createNamedParameter($messageId, IQueryBuilder::PARAM_STR),
+				IQueryBuilder::PARAM_STR,
+			));
+
+		$result = $select->executeQuery();
+		$rows = $result->fetchAll();
```
kesselb, line 62:
```suggestion
		$rows = $result->fetch();
```
You can use fetch if you expect only one row.

lib/Db/MessageSnoozeMapper.php
```diff
@@ -39,6 +39,34 @@ public function __construct(IDBConnection $db, private ITimeFactory $time) {
...
+		$result->closeCursor();
+		if (empty($rows) || $rows[0]['src_mailbox_id'] === null) {
```
kesselb, line 64: Null check for src_mailbox_id should go into the where clause. Find message_snooze where id = x and src_mailbox_id is not null. Then check here whether we have a row or not.

kesselb (approved):

lib/Service/SnoozeService.php
kesselb: :+1:

PR #8786, opened 2023-08-21
Added missing check to see if envelope is eligible to be removed

auroraanna:
Followup to #8762 which was missing this PRs changes:
> @auroraanna Aloha, could you send a pull request to also adjust markSelectionNotJunk() to use for of + await like for markSelectionJunk? Sorry I overlooked it in my first review.
>
> -- @kesselb

The `moveEnvelopeToJunk` action covers both cases according to @kesselb so `markSelectionNotJunk()` should have it too.

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -451,7 +451,7 @@ export default {
 		async markSelectionJunk() {
 			for (const envelope of this.selectedEnvelopes) {
 				if (!envelope.flags.$junk) {
-					this.$store.dispatch('toggleEnvelopeJunk', {
+					await this.$store.dispatch('toggleEnvelopeJunk', {
 						envelope,
 						removeEnvelope: await this.$store.dispatch('moveEnvelopeToJunk', envelope),
 					})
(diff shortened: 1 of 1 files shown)


src/components/EnvelopeList.vue
```diff
@@ -459,15 +459,15 @@ export default {
 			}
 			this.unselectAll()
 		},
-		markSelectionNotJunk() {
-			this.selectedEnvelopes.forEach((envelope) => {
+		async markSelectionNotJunk() {
+			for (const envelope of this.selectedEnvelopes) {
 				if (envelope.flags.$junk) {
 					this.$store.dispatch('toggleEnvelopeJunk', {
```
kesselb, line 465:
```suggestion
					await this.$store.dispatch('toggleEnvelopeJunk', {
```

kesselb: Please also adjust for markSelectionJunk. 
await for dispatch has the result, that the requests are sequential. 
It may take a bit longer, but won't overload the backend with too many requests at once.

kesselb (approved):

kesselb:
Please sign off the commits to make the DCO check happy.

PR #8800, opened 2023-08-23
fix: drop unique index on mail_messages_snoozed.message_id

JohannesGGE:
Fix: #8748

Followup to #8694

I had to rewrite the snooze db insert. Pls check if the throwing of the ServiceException is correct, I'm not quite sure.

### DIFF
--- a/lib/Service/SnoozeService.php
+++ b/lib/Service/SnoozeService.php
@@ -163,30 +163,26 @@ class SnoozeService {
 	 *
 	 * @param Message $message
 	 * @param int $unixTimestamp
+	 * @param Mailbox $srcMailbox
 	 *
 	 * @return void
 	 *
 	 * @throws Exception
+	 * @throws ServiceException
 	 */
 	public function snoozeMessageDB(Message $message, int $unixTimestamp, Mailbox $srcMailbox): void {
 		$snooze = new MessageSnooze();
 		$snooze->setMessageId($message->getMessageId());
 		$snooze->setSnoozedUntil($unixTimestamp);
 		$snooze->setSrcMailboxId($srcMailbox->getId());
-		try {
-			$this->messageSnoozeMapper->insert($snooze);
-		} catch(Exception $e) {
-			if ($e->getReason() === Exception::REASON_UNIQUE_CONSTRAINT_VIOLATION) {
-				$messageId = $message->getMessageId();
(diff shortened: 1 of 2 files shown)


lib/Service/SnoozeService.php
```diff
@@ -222,20 +218,14 @@ public function snoozeThreadDB(Message $selectedMessage, int $unixTimestamp, Mai
...
-					if($messageId === null) {
-						throw $e;
-					}
-					$this->messageSnoozeMapper->deleteByMessageIds(array($messageId));
-					$this->messageSnoozeMapper->insert($snooze);
-				} else {
-					throw $e;
-				}
+
+			$messageId = $message['messageId'];
```
kesselb, line 222: Why would findMessageIdsByThreadRoot return null for messageId?

lib/Service/SnoozeService.php
```diff
@@ -222,20 +218,14 @@ public function snoozeThreadDB(Message $selectedMessage, int $unixTimestamp, Mai
...
+			if($messageId === null) {
+				throw new ServiceException('Cannot extract messageId from message to snooze.');
 			}
+
+			$this->messageSnoozeMapper->deleteByMessageIds(array($messageId));
```
kesselb, line 227:
```suggestion
			$this->messageSnoozeMapper->deleteByMessageIds([$messageId]);
```

lib/Service/SnoozeService.php
```diff
@@ -167,26 +167,21 @@ public function snoozeThread(
...
-			} else {
-				throw $e;
-			}
+
+		$messageId = $message->getMessageId();
+		if($messageId === null) {
+			throw new ServiceException('Cannot extract messageId from message to snooze.');
 		}
+
+		$this->messageSnoozeMapper->deleteByMessageIds(array($messageId));
```
kesselb, line 183:
```suggestion
		$this->messageSnoozeMapper->deleteByMessageIds([$messageId]);
```

lib/Service/SnoozeService.php
JohannesGGE: You're right, messageId shouldn't be null. I just adapted the pattern from the $message->getMessageId(); where the null check was needed.

kesselb (approved):

Issue #8744, opened 2023-08-14
Add unsnooze/wake action


JohannesGGE:
### Is your feature request related to a problem? Please describe.

Contributes to: https://github.com/nextcloud/mail/issues/2048

### Describe the solution you'd like

The action "Snooze" has to be replaced by a action "Unsnooze"/"Wake" if the message/thread is already snoozed

PR #8808, opened 2023-08-24
fix(Snooze): Add unsnooze action

JohannesGGE:
Fix: #8744

Replaces the Snooze button with a Unsnooze button if the selected mailbox is the Snoozed-Mailbox and add unsnooze functionality.

+Mark message as unread after wake

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -261,6 +261,11 @@ return [
 			'verb' => 'POST'
 		],
 		[
+			'name' => 'messages#unSnooze',
+			'url' => '/api/messages/{id}/unsnooze',
+			'verb' => 'POST'
+		],
+		[
 			'name' => 'messages#mdn',
 			'url' => '/api/messages/{id}/mdn',
 			'verb' => 'POST'
@@ -373,6 +378,11 @@ return [
 		[
 			'name' => 'thread#snooze',
 			'url' => '/api/thread/{id}/snooze',
+			'verb' => 'POST'
+		],
+		[
+			'name' => 'thread#unSnooze',
+			'url' => '/api/thread/{id}/unsnooze',
 			'verb' => 'POST'
 		],
 		[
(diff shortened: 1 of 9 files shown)


appinfo/routes.php
```diff
@@ -375,6 +380,11 @@
 			'url' => '/api/thread/{id}/snooze',
 			'verb' => 'POST'
 		],
+		[
+			'name' => 'thread#unSnooze',
+			'url' => '/api/thread/{id}/unsnooze',
+			'verb' => 'GET'
```
ChristophWurst, line 386: POST

lib/Service/SnoozeService.php
```diff
@@ -151,12 +191,72 @@ public function snoozeThread(
...
+	 */
+	public function unSnoozeThread(
+		Message $selectedMessage,
+		?string $currentUserId,
+	):void {
+		$snoozedMailbox = $this->mailManager->getMailbox($currentUserId, $selectedMessage->getMailboxId());
+		$srcAccount = $this->accountService->find($currentUserId, $snoozedMailbox->getAccountId());
+
+		$mailAccount = $srcAccount->getMailAccount();
+		$messageInTrash = $snoozedMailbox->getId() === $mailAccount->getTrashMailboxId(); //Nicht unbedingt nötig?
```
ChristophWurst, line 220: :de:

lib/Service/SnoozeService.php
```diff
@@ -267,8 +339,6 @@ private function wakeMessagesByAccount(Account $account): void {
 					$srcMailboxName
 				);
 
-				//TODO mark message as unread?
```
ChristophWurst, line 270: is that done?

st3iny (approved):
Tested and works.

Restored messages is marked as read though. Should that be fixed?

ChristophWurst (approved):

PR #8815, opened 2023-08-28
Fix misplaced action menu

### LINKED ISSUE #8474: Misplaced actions menues
### Steps to reproduce

1. Go to Mail and login
2. Ad an account and see that there's no configure option
3. Ad another account and see that there is a configure option
4.

### Expected behavior

Configure your first email account.

### Actual behavior

You are unable to configure your first email account.

### Mail app version

3.1.1

### Mailserver or service

IMAP/SMTP

### Operating system

Linux (OpenSuSe Leap 15.4

### PHP engine version

PHP 8.0

### Web server

Apache (supported)

### Database

MySQL

### Additional info

Nextcloud 26.0.1

GretaD:
fixes #8474

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -35,7 +35,8 @@
 			<IconBullet v-else-if="bulletColor" :size="16" :fill-color="bulletColor" />
 		</template>
 		<!-- Actions -->
-		<template #actions>
+		<NcActions
+			:container="container">
 			<template v-if="isDisabled">
 				<ActionText :title="t('mail', 'Provisioned account is disabled')">
 					<template #icon>
@@ -113,7 +114,7 @@
 					{{ t('mail', 'Remove account') }}
 				</ActionButton>
 			</template>
-		</template>
+		</NcActions>
 		<template #extra>
 			<AccountSettings :open="showSettings" :account="account" @update:open="toggleAccountSettings" />
 		</template>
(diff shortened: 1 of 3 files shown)


GretaD:
@kesselb Joas proposed on the frontend channel to add a container. After adding a container, the issue looks like it is fixed, but the three dot menu is positioned wrong. Can you see if im doing something wrong
Screenshot from 2023-08-28 18-00-25

kesselb:
I'm a bit lost as well.

Left: Broken state after adding a new account
Right: Right after page reload

Screenshot from 2023-08-30 17-36-43
Noteable different: The computed property actionsBoundariesElement

Screenshot from 2023-08-30 17-38-22

The automated calculation for the placement seems wrong. 

When I changed the preventOverflow for VPopper component to false for the left (broken) window, the placement works.

GretaD:
@raimund-schluessler you fixed something similar on deck(action menu scrolling) do you have any idea how to fix this properly? Thank you in advance,

raimund-schluessler:
The boundaries element of the actions is computed here:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/7e2dcd2b60670f1844751d831997479fd290dc00/src/components/NcAppNavigationItem/NcAppNavigationItem.vue#L633-L635

Seems like it results in `#content-vue` when the account is created and `undefined` when the app is reloaded. I don't know why one fails and the other works. But if setting `container` or `boundaries-element` fixes the issue, one could create a prop for for `NcAppNavigationItem` and forward it here:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/7e2dcd2b60670f1844751d831997479fd290dc00/src/components/NcAppNavigationItem/NcAppNavigationItem.vue#L318-L319

GretaD:
closing as this is not a solution anyway

PR #8839, opened 2023-09-07
fix(db): Identify retention and snooze mailboxes uniquely

ChristophWurst:
The IMAP message id is not unique across mailboxes or accounts. Multiple accounts that receive the same message will cause a conflict.

### DIFF
--- a/lib/Db/MessageSnooze.php
+++ b/lib/Db/MessageSnooze.php
@@ -29,8 +29,10 @@ namespace OCA\Mail\Db;
 use OCP\AppFramework\Db\Entity;
 
 /**
- * @method void setMessageId(string $messageId)
- * @method string getMessageId()
+ * @method void setMailboxId(int $mailboxId)
+ * @method int getMailboxId()
+ * @method void setUid(int $uid)
+ * @method int getUid()
  * @method void setSnoozedUntil(int $snoozeUntil)
  * @method int getSnoozedUntil()
  * @method void setSrcMailboxId(int $srcMailboxId)
@@ -38,8 +40,11 @@ use OCP\AppFramework\Db\Entity;
  */
 class MessageSnooze extends Entity {
 
-	/** @var string */
-	protected $messageId;
+	/** @var int */
+	protected $mailboxId;
+
+	/** @var int */
+	protected $uid;
 
 	/** @var int */
 	protected $snoozedUntil;
@@ -48,7 +53,8 @@ class MessageSnooze extends Entity {
 	protected $srcMailboxId;
 
 	public function __construct() {
-		$this->addType('messageId', 'string');
+		$this->addType('mailboxId', 'integer');
+		$this->addType('uid', 'integer');
 		$this->addType('snoozedUntil', 'integer');
 		$this->addType('srcMailboxId', 'integer');
 	}
--- a/lib/Db/MessageSnoozeMapper.php
+++ b/lib/Db/MessageSnoozeMapper.php
@@ -47,16 +47,21 @@ class MessageSnoozeMapper extends QBMapper {
 	 *
 	 * @return int|null
 	 */
-	public function getSrcMailboxId(string $messageId): ?int {
+	public function getSrcMailboxId(int $mailboxId, int $uid): ?int {
 		$qb = $this->db->getQueryBuilder();
 
 		$select = $qb->select('src_mailbox_id')
 			->from($this->getTableName())
 			->where(
 				$qb->expr()->eq(
-					'message_id',
-					$qb->createNamedParameter($messageId, IQueryBuilder::PARAM_STR),
-					IQueryBuilder::PARAM_STR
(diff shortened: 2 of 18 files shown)


ChristophWurst:
I *think* snoozing works again. @JohannesGGE please have a critical look.

ChristophWurst:
@st3iny same for trash retention. I think it works (again) but please check carefully.

lib/IMAP/MessageMapper.php
```diff
@@ -317,17 +317,20 @@ public function findByIds(Horde_Imap_Client_Base $client,
...
 		string $sourceFolderId,
 		int $messageId,
-		string $destFolderId): void {
+		string $destFolderId): int {
 		try {
-			$client->copy($sourceFolderId, $destFolderId,
+			$mapping = $client->copy($sourceFolderId, $destFolderId,
 				[
 					'ids' => new Horde_Imap_Client_Ids($messageId),
 					'move' => true,
```
kesselb, line 330:
```suggestion
					'move' => true,
					'force_map' => true,
```

Nice :+1: 


If we get the UID mapping depends on the UIDPLUS extension. 
Copy returns either an array (with UIDPLUS) or true (without UIDPLUS).

The force_map option tells horde to always return an array. Without the UIDPLUS extension, it tries to look up the UID by the message ID. Not perfect but our best option I guess ;)

lib/Db/MessageSnooze.php
```diff
@@ -48,8 +53,8 @@ class MessageSnooze extends Entity {
 	protected $srcMailboxId;
 
 	public function __construct() {
-		$this->addType('messageId', 'string');
-		$this->addType('snoozedUntil', 'integer');
+		$this->addType('mailboxId', 'integer');
+		$this->addType('uid', 'integer');
 		$this->addType('srcMailboxId', 'integer');
```
JohannesGGE, line 58:
```suggestion
		$this->addType('snoozedUntil', 'integer');
```

(No deletion. Just the snoozedUntil column is missing)

lib/Db/MessageSnoozeMapper.php
```diff
@@ -73,17 +78,24 @@ public function getSrcMailboxId(string $messageId): ?int {
...
+				$qb->expr()->in(
+					'mailbox_id',
+					$qb->createNamedParameter($mailboxId, IQueryBuilder::PARAM_INT),
+					IQueryBuilder::PARAM_INT,
+				),
+				$qb->expr()->in(
+					'uid',
+					$qb->createNamedParameter($uid, IQueryBuilder::PARAM_INT),
+					IQueryBuilder::PARAM_INT,
+				),
```
JohannesGGE, line 97: Is IN still okay, after we do not have an array anymore?

lib/Service/SnoozeService.php
```diff
@@ -209,41 +187,34 @@ public function unSnoozeMessageDB(Message $message): void {
...
-			$snooze->setMessageId($message['messageId']);
+			$snooze->setMailboxId($dstMailbox->getId());
+			$snooze->setUid($uid);
 			$snooze->setSnoozedUntil($unixTimestamp);
 			$snooze->setSrcMailboxId($srcMailbox->getId());
 
-			$messageId = $message['messageId'];
-
-			$this->messageSnoozeMapper->deleteByMessageIds([$messageId]);
+			// TODO: why delete, then insert?!
```
JohannesGGE, line 198: Because we do not have the unique index anymore. And that was the quickest way to make sure that there was just one entry for a message :sweat_smile: .
But if we put an index again on the combi of uid and mailboxId we can catch the unique exception again, I think.
What do you think?

JohannesGGE (commented):
Otherwise :+1: 
Thank you for the clean up in the SnoozeService ;)

lib/Migration/Version3400Date20230907103114.php
```diff
@@ -0,0 +1,101 @@
...
+			]);
+		}
+		if (!$snoozedTable->hasColumn('uid')) {
+			$snoozedTable->addColumn('uid', Types::INTEGER, [
+				'notnull' => true,
+			]);
+		}
+		if (!$retentionTable->hasIndex('mail_msg_snoozed_mbuid_idx')) {
+			$retentionTable->addUniqueIndex(['mailbox_id', 'uid'], 'mail_msg_snoozed_mbuid_idx');
+		}
```
JohannesGGE, line 94:
```suggestion
		if (!$snoozedTable->hasIndex('mail_msg_snoozed_mbuid_idx')) {
			$snoozedTable->addUniqueIndex(['mailbox_id', 'uid'], 'mail_msg_snoozed_mbuid_idx');
		}
```

JohannesGGE (approved):

ChristophWurst:
Did another self-review and can't find anything missed. This should be good

miaulalala (approved):

Issue #8840, opened 2023-09-07
Quota notification shows placeholder instead of account email


miaulalala:
### Steps to reproduce

![image](https://github.com/nextcloud/mail/assets/7427347/26978f22-d735-4796-bc87-fd773e67f476)

### Expected behavior

Account email should be shown

### Actual behavior

Placeholder is shown

### Mail app version

3.4.0-beta2

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
![Bildschirmfoto vom 2023-09-07 14-53-22](https://github.com/nextcloud/mail/assets/1374172/7916e4d9-037e-4322-b9ba-2fdce8f4e8a1)

The notification in nextcloud is correct. The desktop notification is not.

PR #8841, opened 2023-09-07
fix(quota): Rename placeholder and fix desktop notification

miaulalala:
Fixes https://github.com/nextcloud/mail/issues/8840


ChristophWurst:
Works now but also works on main when tested with a fake quota

miaulalala:
> Works now but also works on main when tested with a fake quota

That's my experience too.

miaulalala:
Works:

image

image

ChristophWurst (approved):
:+1: pr/commit titles are outdated

PR #8863, opened 2023-09-13
Smart replies for mail

### LINKED ISSUE #8505: Suggested replies for email

It could be useful to have some suggested replies for a message to quickly send an email, especially simple replies like "Thank you", "That works for me", or something similar

There could be a few suggested replies at the bottom of the last message in a thread, with some relevant suggestions based on the contents of the message. Clicking on it could send the message immediately (with the option to undo of course). This would save 2 clicks and some typing.

Rough designs and some variations in placement and button style, cc @jancborchardt and @szaimen for feedback
Note that in the mockups the action is reply, we could also do reply all, what do you think?

## Implementation

* Synchronous mode, like the thread summary. Deployment needs fast LLM. Slow LLM not supported for the first version.
* Use free prompt API if possible, or ask Nextcloud integration team for help with a new backend/API.

## Work packages
* MVP
* Admin docs
* User docs

_No response_

## Additional context

_No response_

hamza221:
Fixes  #8505

| Composer | Message |
|--------|--------|
| image | image | 

- [x] Add caching
- [x] Fix composer update 
- [x] unit tests 
- [x] Better prompt 
- [x] Work on email body, not preview text

For now at least one prompt is executed for every opened message
- [x]  Find a way to filter out Promotional / generated emails without the use of LLM

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -49,6 +49,7 @@ use OCA\Mail\Http\TrapError;
 use OCA\Mail\IMAP\IMAPClientFactory;
 use OCA\Mail\Model\SmimeData;
 use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\AiIntegrations\AiIntegrationsService;
 use OCA\Mail\Service\ItineraryService;
 use OCA\Mail\Service\SmimeService;
 use OCA\Mail\Service\SnoozeService;
@@ -90,6 +91,7 @@ class MessagesController extends Controller {
 	private IDkimService $dkimService;
 	private IUserPreferences $preferences;
 	private SnoozeService $snoozeService;
+	private AiIntegrationsService $aiIntegrationService;
 
 	public function __construct(string $appName,
 		IRequest $request,
@@ -110,7 +112,8 @@ class MessagesController extends Controller {
 		IMAPClientFactory $clientFactory,
 		IDkimService $dkimService,
 		IUserPreferences $preferences,
-		SnoozeService $snoozeService) {
+		SnoozeService $snoozeService,
+		AiIntegrationsService $aiIntegrationService) {
 		parent::__construct($appName, $request);
 		$this->accountService = $accountService;
 		$this->mailManager = $mailManager;
@@ -130,6 +133,7 @@ class MessagesController extends Controller {
 		$this->dkimService = $dkimService;
 		$this->preferences = $preferences;
 		$this->snoozeService = $snoozeService;
+		$this->aiIntegrationService = $aiIntegrationService;
 	}
 
 	/**
@@ -885,6 +889,34 @@ class MessagesController extends Controller {
 			$message->getUid()
 		);
 		return new JSONResponse();
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @param int $messageId
+	 *
+	 * @return JSONResponse
+	 */
+	#[TrapError]
+	public function smartReply(int $messageId):JSONResponse {
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $messageId);
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+		try {
+			$replies = $this->aiIntegrationService->getSmartReply($account, $mailbox, $message, $this->currentUserId);
+		} catch (ServiceException $e) {
+			$this->logger->error('Smart reply failed: ' . $e->getMessage(), [
+				'exception' => $e,
+			]);
+			return new JSONResponse([], Http::STATUS_NO_CONTENT);
+		}
+		return new JSONResponse($replies);
+
 	}
 
 	/**
--- a/lib/Service/AiIntegrations/AiIntegrationsService.php
+++ b/lib/Service/AiIntegrations/AiIntegrationsService.php
@@ -26,8 +26,11 @@ namespace OCA\Mail\Service\AiIntegrations;
 use OCA\Mail\Account;
 use OCA\Mail\Contracts\IMailManager;
 use OCA\Mail\Db\Mailbox;
+use OCA\Mail\Db\Message;
 use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\IMAP\IMAPClientFactory;
+use OCA\Mail\Model\IMAPMessage;
+use OCP\TextProcessing\FreePromptTaskType;
 use OCP\TextProcessing\IManager;
 use OCP\TextProcessing\SummaryTaskType;
 use OCP\TextProcessing\Task;
@@ -64,7 +67,7 @@ class AiIntegrationsService {
 	 *
 	 * @throws ServiceException
 	 */
-	public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, array $messages, string $currentUserId): null|string {
+	public function summarizeThread(Account $account, Mailbox $mailbox, string $threadId, array $messages, string $currentUserId): null|string {
 		try {
 			$manager = $this->container->get(IManager::class);
 		} catch (\Throwable $e) {
@@ -74,7 +77,7 @@ class AiIntegrationsService {
 			$messageIds = array_map(function ($message) {
 				return $message->getMessageId();
 			}, $messages);
-			$cachedSummary = $this->cache->getSummary($messageIds);
+			$cachedSummary = $this->cache->getValue($this->cache->buildUrlKey($messageIds));
 			if($cachedSummary) {
 				return $cachedSummary;
 			}
@@ -99,7 +102,7 @@ class AiIntegrationsService {
 			$manager->runTask($summaryTask);
 			$summary = $summaryTask->getOutput();
 
-			$this->cache->addSummary($messageIds, $summary);
+			$this->cache->addValue($this->cache->buildUrlKey($messageIds), $summary);
 
 			return $summary;
 		} else {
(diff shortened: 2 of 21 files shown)


ChristophWurst:
> POC of #8505

What's missing for a MVC?

hamza221:
> What's missing for a MVC?

I'm still not confident about the method of detecting Personal vs non personal emails, aside from that, nothing ig

JohannesGGE:
Is there anything to consider while testing? I dont't get any reply suggestions. Tried simple mails like in your screenshot. `http://nextcloud.local/index.php/apps/mail/api/messages/3/smartreply` answers with 200, but empty response.

hamza221:
> Is there anything to consider while testing? I dont't get any reply suggestions. Tried simple mails like in your screenshot. `http://nextcloud.local/index.php/apps/mail/api/messages/3/smartreply` answers with 200, but empty response.

If you're getting 200 that means they're getting flagged as generated mails, for some reason. Can you please add a breakpoint and check if its because of `isPersonalEmail()` or LLM output ?

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,14 +102,64 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+			$client = $this->clientFactory->getClient($account);
+			try {
+				$imapMessage = $this->mailManager->getImapMessage(
+					$client,
+					$account,
+					$mailbox,
+					$message->getUid(), true
+				);
+				if($this->isPersonalEmail($imapMessage)){
+					return [];
```
JohannesGGE, line 136: It exits here. Don't we want suggestions for personal emails?

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -115,4 +168,33 @@ public function isLlmAvailable(): bool {
...
+			'noreply', 'no-reply', 'info', 'notification', 'donotreply', 'donot-reply',
+			'admin', 'support', 'contact', 'webmaster', 'helpdesk', 'sales', 'marketing',
+			'service', 'billing', 'newsletter', 'noreply-', 'do-not-reply', 'automated',
+			'donotreply-', 'contact-us', 'noreply.', 'info.', 'marketing.', 'billing',
+			'customer-support', 'service.', 'noreply_', 'do_not_reply', 'unsubscribe',
+			'notification', 'no_reply', 'no-reply', 'marketing-', 'service.',
+			'automated-', 'do_not_reply', 'helpdesk', 'support', 'billing_',
+			'newsletter_', 'news', 'alerts', 'noreply+', 'noreply1', 'donotreply1'
+		];
+		$senderAddress = $imapMessage->getFrom()->first()->getEmail();
```
ChristophWurst, line 183: missing null checks

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -115,4 +168,33 @@ public function isLlmAvailable(): bool {
 		}
 		return in_array(SummaryTaskType::class, $manager->getAvailableTaskTypes(), true);
 	}
+
+	private function isPersonalEmail(IMAPMessage $imapMessage): bool {
+		$commonPatterns = [
+			'noreply', 'no-reply', 'info', 'notification', 'donotreply', 'donot-reply',
+			'admin', 'support', 'contact', 'webmaster', 'helpdesk', 'sales', 'marketing',
```
ChristophWurst, line 175: this might give false negatives. if you are contacted by support@domain.tld it might still be an email personalized to you and awaiting a response.

IMO when in doubt I'd rather show the suggested replies than hide them for legitimate messages.

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,14 +102,64 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
 			return $summary;
 		} else {
 			throw new ServiceException('No language model available for summary');
 		}
 	}
 
+	/**
+	 * @return array
+	 */
+	public function SmartReply(Account $account, Mailbox $mailbox, Message $message, string $currentUserId): array {
```
ChristophWurst, line 116: C# method naming. make it something like `getSmartReply`

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,14 +102,64 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+			$this->cache->addValue($this->cache->uildUrlKey($messageIds), $summary);
 
 			return $summary;
 		} else {
 			throw new ServiceException('No language model available for summary');
 		}
 	}
 
+	/**
+	 * @return array
```
ChristophWurst, line 114:
```suggestion
	 * @return string[]
```

lib/Service/AiIntegrations/AiIntegrationsService.php
hamza221: The I would suggest removing the first llm check. Cause it's usually a false negative

ChristophWurst: Makes sense

hamza221:
image

I started getting this bug, I would say a way around it would be to prompt the LLM twice, once for each question, what do you think @ChristophWurst ?

lib/Service/AiIntegrations/AiIntegrationsService.php
hamza221: You're right, corrected

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,14 +102,57 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+	 * @return string[]
+	 */
+	public function getSmartReply(Account $account, Mailbox $mailbox, Message $message, string $currentUserId): array {
+		try {
+			$manager = $this->container->get(IManager::class);
+		} catch (\Throwable $e) {
+			throw new ServiceException('Text processing is not available in your current Nextcloud version', 0, $e);
+		}
+		if(in_array(FreePromptTaskType::class, $manager->getAvailableTaskTypes(), true)) {
+			$cachedReplies = $this->cache->getValue('smartReplies_'.$message->getUid());
```
ChristophWurst, line 123: the UID is not unique across mailboxes. Use the database ID instead

ChristophWurst (changes requested):
Like with thread summaries we should offer a way to turn this feature off by admins

lib/Service/AiIntegrations/AiIntegrationsService.php
ChristophWurst: support@ and friends are still checked. If you receive an email from support asking if you have tried turning it off and on again the app should still suggest replies.

Let's drop the pattern matching

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -50,6 +50,8 @@ class AiIntegrationsService {
 
 	/** @var IMailManager  */
 	private IMailManager $mailManager;
+	public const SummaryTask = SummaryTaskType::class;
```
ChristophWurst, line 53: what is the idea behind these constants?

conventional php consts are all upper case

hamza221: In order to use `isLlmAvailable()` for different task types.

ChristophWurst: But you could also call isLlmAvailable with SummaryTaskType::class directly

hamza221: I just didn't want to add extra imports in other files, but I can change it

ChristophWurst: I believe the code is easier to read without the const/alias

lib/Service/AiIntegrations/AiIntegrationsService.php
hamza221: Can we leave the basic ones?
`'noreply', 'no-reply', 'notification', 'donotreply', 'donot-reply','noreply-', 'do-not-reply', 'automated',
'donotreply-', 'noreply.', 'noreply_', 'do_not_reply', 'no_reply', 'no-reply', 
'automated-', 'do_not_reply', 'noreply+',`

ChristophWurst: Sure

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,20 +104,93 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+	/**
+	 * @return string[]
+	 */
+	public function getSmartReply(Account $account, Mailbox $mailbox, Message $message, string $currentUserId): array {
 		try {
 			$manager = $this->container->get(IManager::class);
 		} catch (\Throwable $e) {
+			throw new ServiceException('Text processing is not available in your current Nextcloud version', 0, $e);
+		}
+		if(in_array(FreePromptTaskType::class, $manager->getAvailableTaskTypes(), true)) {
```
ChristophWurst, line 124:
```suggestion
		if (in_array(FreePromptTaskType::class, $manager->getAvailableTaskTypes(), true)) {
```

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,20 +104,93 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+			$cachedReplies = $this->cache->getValue('smartReplies_'.$message->getId());
+			if($cachedReplies) {
```
ChristophWurst, line 126:
```suggestion
			if ($cachedReplies) {
```

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,20 +104,93 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+			}
+			$client = $this->clientFactory->getClient($account);
+			try {
+				$imapMessage = $this->mailManager->getImapMessage(
+					$client,
+					$account,
+					$mailbox,
+					$message->getUid(), true
+				);
+				if(!$this->isPersonalEmail($imapMessage)) {
```
ChristophWurst, line 137: you get the idea ;)

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -99,20 +104,93 @@ public function summarizeThread(Account $account, Mailbox $mailbox, $threadId, a
...
+					return [];
+				}
+				$messageBody = $imapMessage->getPlainBody();
+
+			} finally {
+				$client->logout();
+			}
+				$prompt = "Suggest 2 replies to the following email. Each reply should be 25 characters max. Separate the replies with  \"| \", like for example  \"Yes! | No, I'm not available \". Do not print anything else. The email contents are : ".$messageBody."";
```
ChristophWurst, line 145: indent

ChristophWurst (approved):
Tested against chatgpt3.5 & works

The suggestions are OK, some are a bit absurd. But that's fine, we can improve the prompt :+1:

ChristophWurst (approved):

ChristophWurst:
> OCA\Mail\Tests\Unit\Service\AiIntegrationsServiceTest::testCached

> This test did not perform any assertions

lib/Controller/MessagesController.php
```diff
@@ -887,6 +891,35 @@ public function destroy(int $id): JSONResponse {
...
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $messageId);
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+		try {
+			$replies = $this->aiIntegrationService->getSmartReply($account, $mailbox, $message, $this->currentUserId);
+
```
ChristophWurst, line 912: nit: this line can go

PR #8885, opened 2023-09-26
Fix: use message bodies for thread summary

hamza221:


lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -52,24 +64,36 @@ public function __construct(ContainerInterface $container, Cache $cache) {
...
+			$client = $this->clientFactory->getClient($account);
+			try {
+				$messagesBodies = array_map(function ($message) use ($client, $account, $mailbox) {
+					$imapMessage = $this->mailManager->getImapMessage(
+						$client,
+						$account,
+						$mailbox,
+						$message->getUid(), true
+					);
+					$a = $imapMessage->getPlainBody();
```
ChristophWurst, line 90: inline the expression

ChristophWurst (approved):

lib/Service/AiIntegrations/AiIntegrationsService.php
hamza221: Forgot it after debugging 🥲

hamza221:
> @hamza221 is there any noticeable performance impact for the new roundtrip or the larger text fed into the LLM?

The thread I tested with for both is not that long, so I can't tell. something I'm afraid of though is openAi's prompt length limitation

ChristophWurst (approved):

Issue #8050, opened 2023-02-20
🔎📄👤 Search message body


frederikb96:
### Is your feature request related to a problem? Please describe.

If I enter a search phrase in the search box of a folder, only the message subjects are searched.

### Describe the solution you'd like

Option 1: Automatically also search within the message body and the sender emails adresses.
Option 2: Under filter, provide a checkbox for search message body and senders.

### Describe alternatives you've considered

Solution can be combined with #8030

### Additional context

It would be very nice to have this feature, as it is standard in other mail clients.

Thank you very much :)

Djiock:
Glad to see there is already an issue opened, it is indeed pretty standard. The app improved greatly lately though, thanks for that ! Hoping to see this implemented soon.

KimmoJ:
Agreed, came to point this out also. Right now, search is pretty bad since you need to know the subject to have a shot at finding anything, really. This applies both to the global search (at least the email section in the global search) and the search inside the mail app itself. Arguably, if you just enter a search term it should search _all text anywhere probably even including headers_ in my opinion. Then you could optionally limit the choices in the targeted search UI. My workaround for now is firing up another webmail solution entirely (SOGO) whenever I need to search. But yeah, just entering a search term should find that in any text in any mail imo, unless you specify you want to search just a subsection like topic or sender.

ChristophWurst:
I've studied the code a bit. The app does have capabilities to contact IMAP for body search. Right now the search bar always sends off a search for a subject match only. E.g. if you enter "test" in the search input, a request with "subject:test" is sent. This is also documented at https://github.com/nextcloud/mail/pull/6546#issuecomment-1195253032.

TheDBP:
Just adding my interest in seeing this feature added. I love all the Nextcloud integration, the design is clean, in general I really love this webmail client but unfortunately with the search in the state it's in, my organization can't rely on it. Would much prefer to use this client than any other, but we really need a more robust search feature for that to be feasible.

ChristophWurst:
We don't use bounties

ChristophWurst:
Sender/recipient is searched since https://github.com/nextcloud/mail/pull/8645

PR #8897, opened 2023-09-27
Feat: Search mail bodies

hamza221:
Fixes https://github.com/nextcloud/mail/issues/8050

- [x] Add a field in advanced search for body only search

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -836,9 +836,17 @@ class MessageMapper extends QBMapper {
 
 		// createParameter
 		if ($uids !== null) {
-			$select->andWhere(
-				$qb->expr()->in('m.uid', $qb->createParameter('uids'))
-			);
+			// In the case of body+subject search we need a combination of both results,
+			// thus the orWhere in every other case andWhere should do the job.
+			if(!empty($query->getSubjects())) {
+				$select->orWhere(
+					$qb->expr()->in('m.uid', $qb->createParameter('uids'))
+				);
+			} else {
+				$select->andWhere(
+					$qb->expr()->in('m.uid', $qb->createParameter('uids'))
+				);
+			}
 		}
 		foreach ($query->getFlags() as $flag) {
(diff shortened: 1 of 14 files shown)


lib/Db/MessageMapper.php
```diff
@@ -832,7 +832,7 @@ public function findIdsByQuery(Mailbox $mailbox, SearchQuery $query, ?int $limit
 
 		// createParameter
 		if ($uids !== null) {
-			$select->andWhere(
+			$select->orWhere(
```
kesselb, line 835: Debugging leftover?

ChristophWurst:
Code looks good!

Right now there is a new dedicated input in the advanced search. We also discussed an opt-in feature to search in bodies in the search box on top of envelopes, right? I'm fine with doing another PR.

sergioeix:
cant blive this is not option yet, very usefull. Cant wait to test it :)

ChristophWurst:
If I search for a body term only, the search doesn't really have an affect. I also get messages without the search term. Is it the AND vs OR combination of the IMAP and DB results?

hamza221:
> If I search for a body term only, the search doesn't really have an affect. I also get messages without the search term. Is it the AND vs OR combination of the IMAP and DB results?

It is. So if I'm correct,  the only case where we need an orWhere is when we're searching both body and subject, all other combinations should be andWhere.

lib/Db/MessageMapper.php
```diff
@@ -832,9 +832,15 @@ public function findIdsByQuery(Mailbox $mailbox, SearchQuery $query, ?int $limit
 
 		// createParameter
 		if ($uids !== null) {
-			$select->andWhere(
-				$qb->expr()->in('m.uid', $qb->createParameter('uids'))
-			);
+			if(!empty($query->getSubjects())) {
```
ChristophWurst, line 835: please leave a comment for this logic. one of us will wonder about this `if` in the future :)

ChristophWurst (approved):

ChristophWurst:
Works!

HLFH:
@ChristophWurst Hi. Could we have this released in Nextcloud Mail 3.4.5? Thank you!

PR #8898, opened 2023-09-27
Add slash command to mail composer

kesselb:
image
image


hamza221:
@kesselb I tried generalizing the plugin by adding  a 'trigger' argument to insertLink() so it can also work with emoji picker and it's working flawlessly locally. 
Do you think it's a good way to go. If so I'll push a commit here then merge it with the other branch.

hamza221:
I pushed it anyways we can reset it, if there's a problem

kesselb:
Sounds good to me :+1: 

If possible, please add a test.

As example, the test for the signature plugin: https://github.com/nextcloud/mail/blob/main/src/tests/unit/components/SignaturePlugin.spec.js (it's also a command called from TextEditor component).

hamza221:
The mention plugin works, I did a poc in my last commit for the emoji picker,  It still needs some work ui wise. With the next commit I'll add slash commands, fix Ui and add tests then hopefully the feature would be ready

hamza221:
@kesselb  are you okay with megring and doing the review on the main Pr?

Issue #849, opened 2018-03-26
Feature Request: allow specifying password for default account


amaccuish:
My users login using SAML, so nextcloud doesn't have access to their password from LDAP. It would be sweet if I there could also be a "imapPass" and "smtpPass" in the config file, I have a master user setup on my mail server which can impersonate users so the user's actual password isn't needed, just the master one.

Cheers!

rasos:
We are facing a similar challenge: we plan to move from LDAP to SAML authentication for nextcloud to offer a true Single-Sign-On user experience. 

When looking at the `oc_mail_accounts` table, I see `inbound_password `and `outbound_password` and I learned, that this table stores only any _additionally_ account configured by the user in the settings of the Mail app. The default mail account uses the parameters from the config file and the same password given at login, which we don't have anymore available with SAML authentication. The `Auto Mail Accounts` App would create also a default account in the `oc_mail_accounts`table, but we still would not get the password.

A master user setup at the mail server is not an option to us, as our users can define their mail account name themselves, which would allow anybody to impersonate. 

It seems that this is our only choice: Remove the default mail config option and write the mail parameters directly into the  `oc_mail_accounts`table. Any password change in LDAP would need to be synced to `inbound_password `and `outbound_password`, though. This way we would be able to write the encrypted and salted password from LDAP, and not use a symmetric one, so we also gain security. Any concerns so far?

tjareson:
Have you ever been successful with writing login data for an external mail directly to oc_mail_accounts? I would need the same to provision the access to a mail server without user interaction, so it is all ready to use. Configuration by the user is a hassle, specifically if autodetection is not working and users have to deal with IMAP server, ports and protocols. 

I was just wondering: If an account is added in oc_mail_accounts is there a need to create other data in e.g. oc_mail_mailboxes as well or is it done automatically when the user first opens mail?
How is the password encrypted? Is it a hash function?

oliv3r:
I'm curious about this as well, or rather, lets say I would use imap as authentication to nextcloud, can we make it so we can skip the authentication of the mail app? This would make it seemless for users.

ChristophWurst:
That works already

oliv3r:
@ChristophWurst Ah, that's great, can you share some more info? All I found was that you need 'external_user' plugin to do imap auth (which I haven't even got working yet :p)

ChristophWurst:
See the admin groupware settings. For support please use https://help.nextcloud.com/c/support/7

[...]

PR #8917, opened 2023-09-30
Allow SSO users to send and recieve mail via master password

mickenordin:
SSO users do not have a password set, and the auto provisioning of mail does not work for SSO user. It is also inconvenient to synchronize the password database between Nextcloud and the mail server used.

So to allow SSO user to use the mail app we can instead configure a shared password for all users, this will work well with for example Dovecot that has a concept of a "master password"[0] that can be used to authenticate users. To use this feature we must convince the mail app that the user has a password available, which we can set with occ like so:
```
  ./occ config:app:set mail master_password --value 'very-secret-master-password'
```
We can then configure dovecot to allow this password from the Nextcloud server, in this example 89.46.21.198:
```
passdb {
  args = password=very-secret-master-password allow_nets=89.46.21.198/32
  driver = static
}
```
If we configure postfix to use SASL auth against dovecot, we can then both send and recieve mail from Nextcloud mail app, for SSO users.

0. https://doc.dovecot.org/configuration_manual/authentication/master_users/


ChristophWurst:
> SO users do not have a password set, and the auto provisioning of mail does not work for SSO user.

I think it would make most sense if the static password can be an option of provisioning. With the current approach it will be impossible to allow installations to have provisioned accounts and accounts on other imap servers.

mickenordin:
> I think it would make most sense if the static password can be an option of provisioning. With the current approach it will be impossible to allow installations to have provisioned accounts and accounts on other imap servers.

Do you mean that I should add an option for the admin settings like a checkbox "Use master password" and also a textbox for entering a password?

ChristophWurst:
Yes, that is what I mean.

mickenordin:
> Yes, that is what I mean.

Cool! I will start working on that :)

ChristophWurst:
Code looks clean, kudos!

I see that the master password is a global option. How about we move it into a provisioning config? The benefit is that one would be able to add two or more provisioned accounts with master passwords.

mickenordin:
> Code looks clean, kudos!
Thanks! It is not fully working yet though, the master password is not being prefilled if it is set, like I would expect (so that you can intuitivly remove it). Do you have any idea what I did wrong there?

> I see that the master password is a global option. How about we move it into a provisioning config? The benefit is that one would be able to add two or more provisioned accounts with master passwords.

Absolutly, I could do that. How can we detect if an account can be associated with a master password in IMAP and SMTP factories, in that case?

mickenordin:
Also I just noticed #849 . I noticed that they wanted imap and smtp default passord to be separated. Should we do that? And is "default password" a better nomenclature than "master password"?

mickenordin:
> Could setting the password for provisioned accounts with a master password moved to `\OCA\Mail\Http\Middleware\ProvisioningMiddleware::beforeController`?

So setting the master password  as password for the user inside the if clause here? https://github.com/nextcloud/mail/blob/d570006f25ef35a581261a60add4cb9b6ad898b7/lib/Http/Middleware/ProvisioningMiddleware.php#L78

or will the exceptions be triggered anyway you think?

mickenordin:
I moved the settings in to the provisioning/config but I fail to grasp where the values are saved to `oc_mail_provisionings`. Any pointers here would be welcome, I tried to mimic how the sieve settings were handled, but so far no luck as far as getting any settings in to the database.

Also I don't understand how I can get the master password out of the config in \OCA\Mail\Http\Middleware\ProvisioningMiddleware::beforeController (assuming I can get it in there :sweat_smile: ), can I loop through the $configs and get the values out somehow?

mickenordin:
> The provisioning configs will need a database schema change to hold the new password (or two if we split imap and smtp) and changes in the entity that represents the database entries. Then you will be able read the password through a getter of the config.

Does that mean we need a migration?

And after that do I only need to add the values to lib/Db/Provisioning.php or something else?

ChristophWurst:
Yes, a migration and the changes on Provisioning.php, correct :)

lib/Migration/Version3500Date20231005091430.php
```diff
@@ -0,0 +1,118 @@
...
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('mail_provisionings')) {
```
ChristophWurst, line 44: assume the table to exist but check if the columns do

this allows the migration to be executed more than once, e.g. for a rerun on error.

mickenordin: Yes, that makes sense :)

mickenordin:
This is working fully now. I had to add a getter for the mail account in https://github.com/nextcloud/mail/pull/8917/commits/1d116ad2aff41fd9b25533615d3735746ce263bf# 

If there is a better way to get around this, please point me in the right direction. But this is what I came up with :).

Anyway, it works for me know, and this is what it looks like:
Screenshot from 2023-10-05 13-25-36
Screenshot from 2023-10-05 13-25-12

MohammedNoureldin (approved):
I came here because I was searching if there is a way to use mail when logging in using SSO, and I found this great PR. I did a quick review and looks good to me.

mickenordin:
This works really well in our testing, and I was able to configure both postfix and dovecot to used the nextcloud app password  database as a user backend. This means that nextcloud app passwords works for authentication for imap and smtp as well, so this can achieve a really tight integration of self hosted mail with Nextcloud as a web frontend.

MohammedNoureldin:
Hey, @mickenordin!

Thanks for this great commit. Could you provide more details about configuring postfix and dovecot to be able to use this feature?

Do you have any idea when this will be merged to main?

mickenordin:
> Thanks for this great commit. Could you provide more details about configuring postfix and dovecot to be able to use this feature? I mean any minimal simplified config that will get this working.

It is not completly straight forward and involves sasl auth for postfix a lua passdb in dovecot and sql userdb in dovecot and mysql user accounts in postfix. But I will do a writeup if this gets merged, because I think it will be immensly useful to any organization using nextcloud mail. 

> Do you have any idea when this will be merged to main?

This is up to @ChristophWurst :)

MohammedNoureldin:
That would be great. Thank you! I already use SASL for postfix to authenticate against LDAP. So it should be not to much changes for my case at least, I think.

lib/Http/Middleware/ProvisioningMiddleware.php
```diff
@@ -70,6 +70,17 @@ public function beforeController($controller, $methodName) {
 		try {
 			$this->provisioningManager->provisionSingleUser($configs, $user);
 			$password = $this->credentialStore->getLoginCredentials()->getPassword();
+
+			$mailAccount = $this->provisioningManager->getMailAccount($user);
+			$provisioningId = $mailAccount->getProvisioningId();
+			$provisioning = $this->provisioningManager->getConfigById($provisioningId);
```
ChristophWurst, line 76: loading account and provisionings here is expensive. `$configs` holds the provisionings. could you move the logic into `$this->provisioningManager->updatePassword`? Pass the configs into the method. If the account(s) have a config with master password, set the password to the master password. else use the existing logic.

this has the same result as your current implementation, but with fewer database roundtrips

ChristophWurst (commented):
Looks good!

ChristophWurst:
Tested and works :+1: 

Please bump the info.xml version from alpha.2 to alpha3. A version bump causes Nextcloud to check for new migrations and run the one you've added.

lib/Http/Middleware/ProvisioningMiddleware.php
mickenordin: Should be fixed in https://github.com/nextcloud/mail/pull/8917/commits/8fd4982930b1a5fc4741cdca85f89ee0c36b4e8c

mickenordin:
> Tested and works 👍
> 
> Please bump the info.xml version from alpha.2 to alpha3. A version bump causes Nextcloud to check for new migrations and run the one you've added.

Done!

ChristophWurst:
I lack permission to push to your branch

ChristophWurst:
You can find my amendments at https://github.com/nextcloud/mail/pull/9008. It adds a safety check and fixes typing issues.

mickenordin:
@MohammedNoureldin seems like this will be merged soon, and here is my prommised write up: https://wiki.sunet.se/pages/viewpage.action?pageId=164888692

MohammedNoureldin:
I am interesting in knowing if Nextcloud support logging in to Email if the server supports OIDC login. Or is our only chance to auto-login to emails is using this Master password approach?

mickenordin:
Superceded by #9008

Issue #8568, opened 2023-06-21
Missing error handling for message header parsing


sergiupapuc:
### Steps to reproduce

1.error in log
2.foreach() argument must be of type array|object, bool given at /var/www/nextcloud/apps/mail/lib/IMAP/ImapMessageFetcher.php#527
3.

### Expected behavior

foreach() argument must be of type array|object, bool given at /var/www/nextcloud/apps/mail/lib/IMAP/ImapMessageFetcher.php#527

### Actual behavior

foreach() argument must be of type array|object, bool given at /var/www/nextcloud/apps/mail/lib/IMAP/ImapMessageFetcher.php#527

### Mail app version

3.2.2

### Mailserver or service

imap

### Operating system

ubuntu

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

foreach() argument must be of type array|object, bool given at /var/www/nextcloud/apps/mail/lib/IMAP/ImapMessageFetcher.php#527

GretaD:
Hello, can you please explain what where you doing on mail app that this error was shown. Please complete the template with the importation needed

sergiupapuc:
...in the interface of Mail App, no error...but in the logs is the line with error !...nothing special is done, just read mail!

ChristophWurst:
It is probably the loop at https://github.com/nextcloud/mail/blob/3a8cf2dc0abc24b5cae6febdb05f06482302034a/lib/IMAP/ImapMessageFetcher.php#LL528C5-L528C5. ``\Horde_ListHeaders::parse`` returns `false` when the headers can't be parsed.

ChristophWurst:
> error in log

@sergiupapuc please look up the full line in nextcloud.log. We need the complete json entry with its error trace.

sergiupapuc:
foreach() argument must be of type array|object, bool given at /var/www/nextcloud/apps/mail/lib/IMAP/ImapMessageFetcher.php#527

...this is the full line !

ChristophWurst:
> ...this is the full line !

It certainly isn't. nextcloud.log contains json.

kesselb:
```
Index: lib/IMAP/ImapMessageFetcher.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/IMAP/ImapMessageFetcher.php b/lib/IMAP/ImapMessageFetcher.php
[...]

PR #8930, opened 2023-10-05
fix(sync): return if headers couldn't be parsed

miaulalala:
The ```Horde_ListHeaders::parse()``` function can return `false` if headers are a `list-...` header.

Return early to avoid the foreach exploding.


kesselb:
Example from https://github.com/nextcloud/mail/issues/8568#issuecomment-1711668075

{"reqId":"4nmsa8iC8OG4RJV74DJO","level":3,"time":"2023-09-08T12:40:15+00:00","remoteAddr":"","user":"--","app":"mail","method":"","url":"--","message":"failedto parse list-unsubscribeheader","userAgent":"--","version":"27.0.2.1","data":{"app":"mail","name":"List-Unsubscribe","value_single":""}}

Weird that the List-Unsubscribe header is there with an empty value.

kesselb (approved):

lib/IMAP/ImapMessageFetcher.php
```diff
@@ -525,6 +525,10 @@ private function parseHeaders(Horde_Imap_Client_Data_Fetch $fetch): void {
 			$listHeaders = new Horde_ListHeaders();
 			/** @var Horde_ListHeaders_Base[] $headers */
```
ChristophWurst, line 526: this should probably go. it my erase the false variant of the union return type of `parse`

ChristophWurst (approved):

lib/IMAP/ImapMessageFetcher.php
```diff
@@ -523,8 +523,11 @@ private function parseHeaders(Horde_Imap_Client_Data_Fetch $fetch): void {
 		$listUnsubscribeHeader = $parsedHeaders->getHeader('list-unsubscribe');
 		if ($listUnsubscribeHeader !== null) {
 			$listHeaders = new Horde_ListHeaders();
-			/** @var Horde_ListHeaders_Base[] $headers */
 			$headers = $listHeaders->parse($listUnsubscribeHeader->name, $listUnsubscribeHeader->value_single);
+			if(!$headers) {
```
solracsf, line 527:
```suggestion
			if (!$headers) {
```

PR #8944, opened 2023-10-10
Avoid calling Mailbox.vue::loadEnvelopes twice when switching between two accounts

GretaD:
Continuation of #6301 
fixes https://github.com/nextcloud/mail/issues/3746
before
Screenshot from 2023-10-10 19-05-41

after
Screenshot from 2023-10-10 19-06-04

The account watcher is no longer needed. Fetching messages on mailbox change is all we need

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -138,9 +138,6 @@ export default {
 		},
 	},
 	watch: {
-		account() {
-			this.loadEnvelopes()
-		},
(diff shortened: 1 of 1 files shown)


GretaD:
cc @IchbinkeinReh

ChristophWurst (approved):
:+1: works

switching between INBOXes of two accounts causes two requests, this branch only causes one. The mailbox is still synced after fetching.

PR #8951, opened 2023-10-11
perf: Use local caches for avatars

ChristophWurst:
Local caches are faster. Distributed caches are only needed if cache is used for consistency. Populating avatar/provisioning caches for each application server should be cheaper than the remote cache access for a lot of requests reading avatars and provisionings.


kesselb (approved):

miaulalala (approved):

ChristophWurst:
I'm realizing that for provisionings it's tricky. We don't use cache as a cache but as storage with explicit clearing to propagate changes. There will be strange artifacts on distributed systems.

ChristophWurst:
Provisioning cache changes undone

Issue #8952, opened 2023-10-11
Mail-App Advanced Search Error


AlexanderZuehr:
### Steps to reproduce

Hey if I try to search in the mail-app with the advanced mail-app search function I am getting an error if I use one of the following fields.

To:
Cc:
Bcc:

I'll attach pictures and the log from nextcloud.

<img width="600" alt="Bildschirmfoto 2023-10-11 um 11 43 32" src="https://github.com/nextcloud/mail/assets/76497172/f98b81e1-9936-4659-96e1-0464fc314997">
<img width="605" alt="Bildschirmfoto 2023-10-11 um 11 43 14" src="https://github.com/nextcloud/mail/assets/76497172/59ee05a5-afaf-4117-9cfb-b6395f98f2f8">
<img width="602" alt="Bildschirmfoto 2023-10-11 um 11 42 55" src="https://github.com/nextcloud/mail/assets/76497172/b9e2cbef-4e7e-4743-9cd7-77c6d145b7b6">
<img width="300" alt="Bildschirmfoto 2023-10-11 um 11 42 48" src="https://github.com/nextcloud/mail/assets/76497172/2e8ebb89-4f6f-4304-8dfb-955aed4358ed">

### Mail app version

3.4.0

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

[mail] Fehler: OC\DB\Exceptions\DbalException: An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'r0.email' in 'where clause' at <<closure>>

 0. /var/www/html/lib/private/DB/QueryBuilder/QueryBuilder.php line 295
    OC\DB\Exceptions\DbalException::wrap()
 1. /var/www/html/lib/public/AppFramework/Db/QBMapper.php line 335
    OC\DB\QueryBuilder\QueryBuilder->executeQuery()
 2. /var/www/html/apps/mail/lib/Db/MessageMapper.php line 869
    OCP\AppFramework\Db\QBMapper->findEntities()
 3. /var/www/html/apps/mail/lib/Service/Search/MailSearch.php line 169
    OCA\Mail\Db\MessageMapper->findIdsByQuery()
 4. /var/www/html/apps/mail/lib/Service/Search/MailSearch.php line 133
    OCA\Mail\Service\Search\MailSearch->getIdsLocally()
 5. /var/www/html/apps/mail/lib/Controller/MessagesController.php line 159
[...]

PR #8953, opened 2023-10-11
fix(search): Use corresponding table alias for recipient search

ChristophWurst:
Each recipient type has its own join and join table alias. The WHERE clauses have to use the corresponding alias.

Fixes https://github.com/nextcloud/mail/issues/8952
Fixes https://github.com/nextcloud/mail/issues/8954

Regression of https://github.com/nextcloud/mail/pull/8645.

## How to test

1) Open the app
2) Configure an account
3) Search for *To* **and omit From**


GretaD (approved):
works

kesselb:
Good finding :+1:

PR #8962, opened 2023-10-11
Chore: drop use of \OCA\Mail\Account::getImapConnection

hamza221:


kesselb (approved):
Thank you, good finding :+1:

Issue #8753, opened 2023-08-15
Modal closes automatically


kesselb:
### Steps to reproduce

1. Open Mail
2. Pick a message
3. Click move message (e.g. in the menu envelope)
4. Wait a few seconds
5. 😯

### Expected behavior

Let me choose the mailbox without having to hurry up ;)

### Actual behavior

[Screencast from 2023-08-15 12-41-11.webm](https://github.com/nextcloud/mail/assets/3902676/e57ea6aa-38a0-4a9b-a23b-40df776d5a0a)

### Mail app version

https://github.com/nextcloud/mail/commit/49b34a6c78efe446979ab41c899a4b3c3877e11d

### PHP engine version

None

### Web server

None

### Database

None

JohannesGGE:
I reverted the snooze changes and could reproduce the bug, so this shouldn't be a regression from snooze.
I tried to understand why the modal behave like this, can't relate. The component states `showMoveModal` still seems to be set to true after the the modal disappears.

ChristophWurst:
@GretaD could you look into this please?

PR #8232, opened 2023-03-15
Fix the overlapping of primary menu with recipient

GretaD:
fixes #8191

Im trying to adapt text logic for mail: https://github.com/nextcloud/text/blob/8a17046aa440df841fe9182205d80ce937068c1a/src/components/Menu/MenuBar.vue

and also using inline prop that `NcActions` uses

[Screencast from 08.05.2023 19:06:28.webm](https://user-images.githubusercontent.com/12728974/236886363-2937e7fa-2f8a-4c17-b9d2-3050b7c77c23.webm)

PR #8985, opened 2023-10-20
fix(envelope menu): prevent modals from closing automatically

JohannesGGE:
Fix: #8753

Seems to be a structure issue caused by this pr: https://github.com/nextcloud/mail/pull/8232

I moved the modals from `MenuEnvelope.vue` to `ThreadEnvelope.vue` so that they are no longer part of the NCAction Tag.

I also removed the leftover `vue-frag` package and renamed a forgotten ButtonVue to NCActionButton.

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -84,13 +84,13 @@
 					</div>
 				</div>
 				<div class="envelope__header__left__unsubscribe">
-					<ButtonVue
+					<NcActionButton
 						v-if="message && message.dkimValid && (message.unsubscribeUrl || message.unsubscribeMailto)"
 						type="tertiary"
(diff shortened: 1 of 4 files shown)


src/components/ThreadEnvelope.vue
```diff
@@ -729,6 +775,46 @@ export default {
...
+			this.showTaskModal = false
+		},
+		onOpenTagModal() {
+			this.showTagModal = true
+		},
+		onCloseTagModal() {
+			this.showTagModal = false
+		},
+		async onShowSourceModal() {
+			const resp = await axios.get(
```
hamza221, line 806: suggestion: Only query it the first time the modal is opened.

hamza221 (approved):
Tested

PR #9001, opened 2023-10-25
fix(classification): Delete historic data

ChristophWurst:
It's interesting to analyze for improvements but otherwise just bloats app data and the database.

## How to test

1) Set up an account
2) Run `mail:account:train` 10x 
3) Wait two months and a day (or change oc_mail_classifiers.created_at timestamps in the database)
4) Run `SELECT COUNT(*) FROM oc_mail_classifiers` and remember the number
5) Run `occ mail:clean-up`
6) Run `SELECT COUNT(*) FROM oc_mail_classifiers` and compare it to the number before

### DIFF
--- a/lib/Db/ClassifierMapper.php
+++ b/lib/Db/ClassifierMapper.php
@@ -58,4 +58,16 @@ class ClassifierMapper extends QBMapper {
 
 		return $this->findEntity($select);
 	}
+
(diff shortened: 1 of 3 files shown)


miaulalala (approved):
Looks gud

JohannesGGE (approved):

Issue #8864, opened 2023-09-13
🔎 Allow message date and recipient search in advanced search


ChristophWurst:
### Is your feature request related to a problem? Please describe.

Extend advanced search in server with Mail specifics, like searching sender and recipients.

### Describe the solution you'd like

Integrate with the new seach in Hub 7 (details follow)

### Additional context

https://github.com/nextcloud/server/issues/39162

ChristophWurst:
API draft is at https://github.com/nextcloud/server/pull/40618.

JohannesGGE:
Current state and more information in this draft-pr: https://github.com/nextcloud/mail/pull/9002

PR #9002, opened 2023-10-25
feat(search): allow date and recipient search

JohannesGGE:
Fix: https://github.com/nextcloud/mail/issues/8864
Server PR: https://github.com/nextcloud/server/pull/40618

---

Currently it's either possible to provide `email:` (will be mapped from `person`) or `from:/to:/cc:/bcc:`. `email` will search in all fields (or), `from/to/cc/bcc` in the explicit field (and). If both filter types are provided  `email` will be used.

## How to test

`curl -u admin:admin https://localhost/ocs/v2.php/search/providers/mail/search?term=spam&person=nc-user_admin`

## Open Questions:
- [x] What does the `person` filter provide exactly? An  IUser

## ToDos:
- [x] Get email addresses for `person` (depends on what person provide

## Follow-ups
- Is it possible to refactor `\OCA\Mail\Db\MessageMapper::findIdsByQuery` and `\OCA\Mail\Db\MessageMapper::findIdsGloballyByQuery`?
- There is a bug introduced in https://github.com/nextcloud/mail/pull/8897/files#diff-39657503b9cdd0f0fa8292196e6432d8a918b4d16ab8212f2a5c9b6f67857e86 . If just term is provided the result is wrong (All emails because no filter. But should be parsed as `subject:term`) -> Issue: https://github.com/nextcloud/mail/issues/9043

### DIFF
--- a/lib/Search/FilteringProvider.php
+++ b/lib/Search/FilteringProvider.php
@@ -1,0 +1,94 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright 2023 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2023 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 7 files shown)


ChristophWurst:
I think the API is implemented correctly. Can't test it right now because searching for `since:1699474523` doens't end up in Mail's search provider. Neither does `in:mail since:1699474523` :shrug:

lib/Search/FilteringProvider.php
```diff
@@ -0,0 +1,94 @@
...
+		}
+		if ($until = $query->getFilter('until')?->get()) {
+			if ($until instanceof DateTimeImmutable) {
+				$ts = $until->getTimestamp();
+				$filters[] = "end:$ts";
+			}
+		}
+		if ($userFilter = $query->getFilter('person')?->get()) {
+			if ($userFilter instanceof IUser) {
+				$email = $userFilter->getEMailAddress();
```
hamza221, line 58:
```suggestion
				$email = $userFilter->getEmailAddress();
```

ChristophWurst: I'm pretty sure `\OCP\IUser::getEMailAddress` is the correct method capitalization

Issue #8325, opened 2023-04-03
"Contacts with this address:" but nothing shown


ChristophWurst:
### Steps to reproduce

1. Save yourself to the addressbook
2. Send yourself an email
3. Click the recipient

### Expected behavior

The app finds the contact and shows it

### Actual behavior

The app finds the contact but nothing shows

![image](https://user-images.githubusercontent.com/1374172/229497619-1fa391ec-6a88-42dd-80e7-0386d4d2e5f4.png)

### Mail app version

3.0

### PHP engine version

None

### Web server

None

### Database

None

auroraanna:
![image](https://github.com/nextcloud/mail/assets/81317317/fe82a3e6-c7a5-4b57-9220-8d2c036e1af4)

I followed the instructions but it does show the email address. It doesn't display "Contacts with this address" in front of it anymore though. Seems like it's been fixed?

PR #9004, opened 2023-10-26
fix(sender-details): show contact names again

JohannesGGE:
Fix: #8325

Regression of: #6995

|before|after|
|--|--|
|image|image|


GretaD (approved):
good catch

PR #9008, opened 2023-10-27
feat(provisioning): Master password support

ChristophWurst:
Amended and squashed version of https://github.com/nextcloud/mail/pull/8917.

## How to test

1) Build https://github.com/ChristophWurst/docker-imap-devel/pull/13
2) Set up provisioning for the static password from the config file

## How to test – Alternative

1) Set up provisioning with your testing account's email address and the "static" IMAP password :wink: 
2) Verify the account works

### DIFF
--- a/lib/Db/Provisioning.php
+++ b/lib/Db/Provisioning.php
@@ -51,6 +51,10 @@ use ReturnTypeWillChange;
  * @method void setSmtpPort(int $smtpPort)
  * @method string getSmtpSslMode()
  * @method void setSmtpSslMode(string $smtpSslMode)
+ * @method bool|null getMasterPasswordEnabled()
+ * @method void setMasterPasswordEnabled(bool $masterPasswordEnabled)
+ * @method string|null getMasterPassword()
+ * @method void setMasterPassword(string $masterPassword)
  * @method bool|null getSieveEnabled()
  * @method void setSieveEnabled(bool $sieveEnabled)
  * @method string|null getSieveHost()
@@ -70,6 +74,7 @@ use ReturnTypeWillChange;
  */
 class Provisioning extends Entity implements JsonSerializable {
 	public const WILDCARD = '*';
+	public const MASTER_PASSWORD_PLACEHOLDER = '********';
 
 	protected $provisioningDomain;
 	protected $emailTemplate;
@@ -81,6 +86,8 @@ class Provisioning extends Entity implements JsonSerializable {
 	protected $smtpHost;
 	protected $smtpPort;
 	protected $smtpSslMode;
+	protected $masterPasswordEnabled;
+	protected $masterPassword;
 	protected $sieveEnabled;
 	protected $sieveUser;
 	protected $sieveHost;
@@ -93,6 +100,8 @@ class Provisioning extends Entity implements JsonSerializable {
 	public function __construct() {
 		$this->addType('imapPort', 'integer');
 		$this->addType('smtpPort', 'integer');
+		$this->addType('masterPasswordEnabled', 'boolean');
+		$this->addType('masterPassword', 'string');
 		$this->addType('sieveEnabled', 'boolean');
 		$this->addType('sievePort', 'integer');
 		$this->addType('ldapAliasesProvisioning', 'boolean');
@@ -112,6 +121,8 @@ class Provisioning extends Entity implements JsonSerializable {
 			'smtpHost' => $this->getSmtpHost(),
 			'smtpPort' => $this->getSmtpPort(),
 			'smtpSslMode' => $this->getSmtpSslMode(),
+			'masterPasswordEnabled' => $this->getMasterPasswordEnabled(),
+			'masterPassword' => !empty($this->getMasterPassword()) ? self::MASTER_PASSWORD_PLACEHOLDER : null,
 			'sieveEnabled' => $this->getSieveEnabled(),
 			'sieveUser' => $this->getSieveUser(),
 			'sieveHost' => $this->getSieveHost(),
--- a/lib/Service/Provisioning/Manager.php
+++ b/lib/Service/Provisioning/Manager.php
@@ -316,9 +316,23 @@ class Manager {
 		}
 	}
 
-	public function updatePassword(IUser $user, string $password): void {
+	/**
+	 * @param Provisioning[] $provisionings
+	 */
+	public function updatePassword(IUser $user, string $password, array $provisionings): void {
 		try {
 			$account = $this->mailAccountMapper->findProvisionedAccount($user);
+
+			$provisioning = $this->findMatchingConfig($provisionings, $user);
+			if ($provisioning === null) {
+				return;
+			}
+			$masterPassword = $provisioning->getMasterPassword();
+			$masterPasswordEnabled = $provisioning->getMasterPasswordEnabled();
+			if ($masterPasswordEnabled && $masterPassword !== null) {
+				$password = $masterPassword;
+				$this->logger->debug('Password set to master password for ' . $user->getUID());
+			}
 
 			if (!empty($account->getInboundPassword())
 				&& $this->crypto->decrypt($account->getInboundPassword()) === $password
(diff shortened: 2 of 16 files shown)


ChristophWurst:
Code coverage decreased

lib/Db/Provisioning.php
```diff
@@ -112,6 +120,8 @@ public function jsonSerialize() {
 			'smtpHost' => $this->getSmtpHost(),
 			'smtpPort' => $this->getSmtpPort(),
 			'smtpSslMode' => $this->getSmtpSslMode(),
+			'masterPasswordEnabled' => $this->getMasterPasswordEnabled(),
+			'masterPassword' => $this->getMasterPassword(),
```
ChristophWurst, line 124: We should not expose this

src/components/settings/ProvisioningSettings.vue
```diff
@@ -197,6 +197,37 @@
...
+								type="checkbox"
+								class="checkbox">
+							<label :for="'mail-master-password-enabled' + setting.id">
+								{{ t('mail', 'Use master password') }}
+							</label>
+						</div>
+						<div>
+							<input
+								:id="'mail-password-id' + setting.id"
+								v-model="passwordId"
```
ChristophWurst, line 218: > [Vue warn]: Property or method "passwordId" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://v2.vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

mickenordin:
Will this be merged in time for 3.5.0? I saw that you guys were tagging beta releases?

ChristophWurst:
Yes, no rush. 3.5 will only be branched off in a week. This PR should make it.

ChristophWurst:
@mickenordin may I ask you to also give the latest state a final test? :pray:

mickenordin:
> @mickenordin may I ask you to also give the latest state a final test? 🙏

Definetly! Will get right on it!

mickenordin:
Unfortunately this does no longer work for me. The app now fails to detect that the user should see the mail box, and instead claims that it is disabled due to being a SSO account.

The actual code works though it seems. If I set the master password and give a local user the same email it works, but not for a user signed in with single signon.

I am pretty sure that this used to work, so I am not sure what happened here. How can it have stopped detecting that the user account should be able to see the mail box?

ChristophWurst:
I'm also unable to test the changes at the moment

mickenordin:
@ChristophWurst I see what is missing now, it is the check in 

src/components/Navigation.vue in https://github.com/nextcloud/mail/pull/8917/files#diff-578fcad0646f9b6ab3e4f8bbe2f8f946e54d39670c0da653db620aaead7d9663

mickenordin:
From: Micke Nordin ***@***.***>

---
 src/components/Navigation.vue | 3 ++-
 src/store/getters.js          | 1 +
 src/store/index.js            | 1 +
 src/store/mutations.js        | 3 +++
 4 files changed, 7 insertions(+), 1 deletion(-)

diff --git a/src/components/Navigation.vue b/src/components/Navigation.vue
index 5feaf97b0..adcaa26d2 100644
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -182,7 +182,8 @@ export default {
 		 * @return {boolean} True if the account should be disabled
 		 */
 		isDisabled(account) {
-			return this.passwordIsUnavailable && !!account.provisioningId
+
+			return (this.passwordIsUnavailable && !!account.provisioningId) && !!this.$store.getters.masterPasswordEnabled
 		},
 	},
 }
diff --git a/src/store/getters.js b/src/store/getters.js
index da6424222..e2d7f1bf5 100644
--- a/src/store/getters.js
+++ b/src/store/getters.js
@@ -117,6 +117,7 @@ export const getters = {
 	isScheduledSendingDisabled: (state) => state.isScheduledSendingDisabled,
 	isSnoozeDisabled: (state) => state.isSnoozeDisabled,
 	googleOauthUrl: (state) => state.googleOauthUrl,
+	masterPasswordEnabled: (state) => state.masterPasswordEnabled,
 	microsoftOauthUrl: (state) => state.microsoftOauthUrl,
 	getActiveSieveScript: (state) => (accountId) => state.sieveScript[accountId],
 	getCurrentUserPrincipal: (state) => state.currentUserPrincipal,
diff --git a/src/store/index.js b/src/store/index.js
index fbd245183..6bdbda6fe 100644
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -103,6 +103,7 @@ export default new Store({
 				isSnoozeDisabled: false,
 				currentUserPrincipal: undefined,
 				googleOauthUrl: null,
+				masterPasswordEnabled: false,
 				sieveScript: {},
 				calendars: [],
 				smimeCertificates: [],
diff --git a/src/store/mutations.js b/src/store/mutations.js
index ad7c1beb8..3719dc699 100644
--- a/src/store/mutations.js
+++ b/src/store/mutations.js
@@ -475,6 +475,9 @@ export default {
 	setGoogleOauthUrl(state, url) {
 		state.googleOauthUrl = url
 	},
+	setMasterPasswordEnabled(state, value) {
+		state.masterPasswordEnabled = value
+	},
 	setMicrosoftOauthUrl(state, url) {
 		state.microsoftOauthUrl = url
 	},
-- 
2.42.0

mickenordin:
The patch above fixes the issue.

ChristophWurst:
The feature works again :rocket:

ChristophWurst (commented):
:+1: on @mickenordin's changes :)

Thanks a lot

lib/Db/Provisioning.php
```diff
@@ -51,6 +51,10 @@
  * @method void setSmtpPort(int $smtpPort)
  * @method string getSmtpSslMode()
  * @method void setSmtpSslMode(string $smtpSslMode)
+ * @method bool|null getMasterPasswordEnabled()
+ * @method void setMasterPasswordEnabled(bool $masterPasswordEnabled)
+ * @method string|null getMasterPassword()
+ * @method void setMasterPassword(string $masterPassword)
```
miaulalala, line 57:
```suggestion
 * @method void setMasterPassword(?string $masterPassword)
```

miaulalala: To unset the master password?

lib/Service/Provisioning/Manager.php
```diff
@@ -316,10 +316,24 @@ public function deprovision(Provisioning $provisioning): void {
...
 
+			$provisioning = $this->findMatchingConfig($provisionings, $user);
+			if ($provisioning === null) {
+				return;
+			}
+			$masterPassword = $provisioning->getMasterPassword();
+			$masterPasswordEnabled = $provisioning->getMasterPasswordEnabled();
+			if ($masterPasswordEnabled && $masterPassword !== null) {
+				$password = $masterPassword;
+				$this->logger->debug('Password set to master password for ' . $user->getUID());
```
miaulalala, line 334: early return here? Having both a master pw and a regular pw is unneccessary, right?

miaulalala (approved):

sevmonster:
This PR doesn't work if the user never had a password to begin with. This is the case in my environment as none of my users were created with a password, they were all created via `user_saml`.

https://github.com/nextcloud/mail/blob/3fc016ffa4c95c24c32bfe6d8a21a2a37fa0af78/lib/Http/Middleware/ProvisioningMiddleware.php#L60

I am not sure why the password logic is split between two places like this when it seems the only place it is used is [in the tests](https://github.com/search?q=repo%3Anextcloud/mail%20updatePassword&type=code). Could all of this logic be moved into the `updatePassword` function for simplicity and also to help fix this regression? I am not a NC dev so I don't know if this split is a requirement for some reason.

mickenordin:
My users also don't have a password set and were created by the global site selector. This is what this PR aims to fix by allowing the administrator to set a master password.  Can you explain a bit more clearly what is not working, and what the regression is?

ChristophWurst:
Please put this into a new ticket.

sevmonster:
> My users also don't have a password set and were created by the global site selector.
Your users probably have *something* set as their password in the database; they would not be able to use provisioned accounts at all if they didn't. Please check your actual database for values in the `password` row.

> Can you explain a bit more clearly what is not working, and what the regression is?

@mickenordin @ChristophWurst #9175

PR #9030, opened 2023-10-30
ci: Turn off builds for pushes and merges to main/stable*

ChristophWurst:
We don't check the results so CI minutes are used without purpose.

### DIFF
--- a/.github/workflows/lint-info-xml.yml
+++ b/.github/workflows/lint-info-xml.yml
@@ -5,13 +5,7 @@
 
 name: Lint info.xml
 
-on:
-  pull_request:
-  push:
-    branches:
-      - main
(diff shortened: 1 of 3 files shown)


.github/workflows/lint-info-xml.yml
```diff
@@ -7,7 +7,6 @@ name: Lint info.xml
 
 on:
   pull_request:
-  push:
     branches:
```
nickvergessen, line 10: This was part of push and is now part of pr, intended?
Doesn't allow sending follow up prs, but fine either way for normal work

nickvergessen (approved):

nickvergessen (approved):

Issue #9050, opened 2023-11-03
With "Enable mail body search" message bodies are still not searched


ChristophWurst:
### Steps to reproduce

1. Configure an account
2. Open the account settings
3. Tick *Enable mail body search*
4. Close the settings
5. Search messages with the input on top of the message list (**not** the advanced search dialogue)

### Expected behavior

Search term is used for subject and body search

### Actual behavior

Search term is used for subject search

### Mail app version

3.5.0 beta 1

### PHP engine version

None

### Web server

None

### Database

None

hamza221:
I can't reproduce 
![image](https://github.com/nextcloud/mail/assets/40746210/7769dfe3-2d69-4c36-aa90-bf7227415be7)

ChristophWurst:
Possibly unrelated but shows while testing the feature with various accounts:

```
[ERROR] mail: Could not fetch envelopes of mailbox 80 (subject:Test body:Test)
TypeError: i is undefined
    addEnvelope mutations.js:286
    y vuex.esm.js:844
    commit vuex.esm.js:466
    commit vuex.esm.js:465
    _withCommit vuex.esm.js:624
    commit vuex.esm.js:464
    commit vuex.esm.js:409
    fetchEnvelopes actions.js:585
    U _map.js:7
    Z map.js:62
    a _dispatchable.js:44
    t _curry2.js:20
    t _curry1.js:16
    le tap.js:30
    a _dispatchable.js:44
    t _curry2.js:20
    t _curry1.js:16
    promise callback*65505/oe< andThen.js:39
    t _curry2.js:20
    t _curry1.js:16
    g _pipe.js:3
    r _arity.js:31
    fetchEnvelopes actions.js:593
    Ne sessionExpiryHandler.js:26
    fetchEnvelopes actions.js:545
    y vuex.esm.js:851
    dispatch vuex.esm.js:516
    dispatch vuex.esm.js:406
    loadEnvelopes Mailbox.vue:193
    searchQuery Mailbox.vue:149
    VueJS 12
    onUpdateSearchQuery MailboxThread.vue:297
    VueJS 4
    sendQueryEvent SearchMessages.vue:444
    s index.js:54
    setTimeout handler*69496/e.exports/< index.js:38
    query SearchMessages.vue:420
    VueJS 12
    input SearchMessages.vue:1
    VueJS 3
```

ChristophWurst:
> Possibly unrelated

Narrowed it down a bit.

[...]

PR #9053, opened 2023-11-06
Add possiblitity to search message bodies in priority inbox

hamza221:
ref: #9050

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -5,6 +5,19 @@
 			{{ t('mail', 'Add mail account') }}
 		</router-link>
 
+		<p v-if="loadingPrioritySettings" class="app-settings">
+			{{ prioritySettingsText }}
+		</p>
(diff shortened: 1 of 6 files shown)


src/components/AppSettingsMenu.vue
```diff
@@ -198,6 +216,19 @@ export default {
...
+
+			this.$store
+				.dispatch('savePreference', {
+					key: 'search-priority-body',
+					value: e.target.checked ? 'true' : 'false',
+				})
+				.catch((error) => Logger.error('could not save preferences', { error }))
+				.then(() => {
+					this.loadingPrioritySettings = false
+				})
```
ChristophWurst, line 230: my typical requested review: please make the method async and use the await syntax with a try-catch :)

hamza221: I should start remembering that

ChristophWurst (approved):

PR #9056, opened 2023-11-08
Fix multiple tags with same label

### LINKED ISSUE #9021: Multiple tags with same label
### Steps to reproduce

1.Press edit tags on any envelope
2.write tag name in the input
3.click on the button fast multiple times
image

### Expected behavior

Button is disabled and tag is only created once

### Actual behavior

The tag is created multiple times.
PS: after a page refresh only one copy would stay as the api call will fail for the duplicates

### Mail app version

latest main

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

ScratchX98:
Fixes #9021

I figured it would be reasonable to keep using the `showSaving` ref to check that the async request finishes

### DIFF
--- a/src/components/TagModal.vue
+++ b/src/components/TagModal.vue
@@ -48,7 +48,7 @@
 					</template>
 					{{ t('mail', 'Add tag') }}
 				</NcButton>
-				<ActionInput v-if="editing" @submit="createTag" :disabled="showSaving">
+				<ActionInput v-if="editing" :disabled="showSaving" @submit="createTag">
 					<template #icon>
 						<IconTag :size="20" />
 					</template>
@@ -174,7 +174,9 @@ export default {
 				showError(this.t('mail', 'Tag already exists'))
 				return
 			}
-			if(displayName !== null && displayName !== '') this.showSaving = true
+			if (displayName !== null && displayName !== '') {
+				this.showSaving = true
+			}
 			try {
 				await this.$store.dispatch('createTag', {
 					displayName,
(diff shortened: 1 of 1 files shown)


src/components/TagModal.vue
```diff
@@ -163,6 +163,8 @@ export default {
 		},
 		async createTag(event) {
 			this.editing = true
+			if (this.showSaving) return
```
ChristophWurst, line 166: It would be more elegant if the button is disabled while the previous async request is still happening. Then `createTag` can never be called while `this.showSaving` is true ;)

ScratchX98: Right, I added a commit disabling the input when needed, but leaving the `showSaving` condition in `createTag` for safety, or do you reckon that should be removed to simplify?

ChristophWurst: Fair enough. We can do both!

ChristophWurst, line 166:
```suggestion
			if (this.showSaving) {
				return
			}
```

ChristophWurst (changes requested):
Coding style

ChristophWurst (approved):
Looks good!

src/components/TagModal.vue
```diff
@@ -172,6 +176,9 @@ export default {
 				showError(this.t('mail', 'Tag already exists'))
 				return
 			}
+			if (displayName !== null && displayName !== '') {
```
GretaD, line 179: we dont need both checks, right? We can just write: `if (displayName !== '')` and the result is the same

GretaD (approved):
nice, it works :) thank you @ScratchX98

GretaD:
wait, i tested adding a tag not editing one. retesting

GretaD (approved):

PR #9057, opened 2023-11-09
Fix Mail-App Deutsch (Persönlich:Du) de.js

AlexanderZuehr:
Added missing translations for one of the German languages you can choose. Deutsch (Förmlich: Sie) is already correct, but Deutsch (Persönlich:Du) is missing some strings, so I've added these strings and changed them to fit the Persönlich:Du Style.


ChristophWurst (changes requested):
Hi @AlexanderZuehr! Translations are maintained on https://www.transifex.com/nextcloud and synced to git by automation. Please make your changes on Transifex :)

PR #9062, opened 2023-11-14
Fix: delete tags unset before confirmation

hamza221:
Tags get unset when clicking on delete tag before confirming

### DIFF
--- a/src/components/DeleteTagModal.vue
+++ b/src/components/DeleteTagModal.vue
@@ -39,6 +39,11 @@ export default {
 			type: Object,
 			required: true,
 		},
+		envelopes: {
+			// The envelopes on which this menu will act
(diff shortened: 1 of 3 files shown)


ChristophWurst:
I'm not sure I understand the problem. Could you describe before/after behavior and how this can be tested? THanks :)

hamza221:
> I'm not sure I understand the problem. Could you describe before/after behavior and how this can be tested? THanks :) 

Before:
- set a test tag on some messages,
- click on delete tags 
- don't confirm
- envelopes get untagged anyways

ChristophWurst:
Doesn't work for me. Untagging still happens before confirm.

hamza221:
Should work now

PR #9074, opened 2023-11-17
Exclude empty tags from saving

### LINKED ISSUE #9020: Empty Tag
### Steps to reproduce

1.Press edit tags on any envelope
2. add an empty tag
image

### Expected behavior

Error message or disabled button, maybe also add a check in the controller

### Actual behavior

Tag is created

### Mail app version

main

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes #9020

### DIFF
--- a/src/components/TagModal.vue
+++ b/src/components/TagModal.vue
@@ -177,8 +177,9 @@ export default {
 				showError(this.t('mail', 'Tag already exists'))
 				return
 			}
-			if (displayName !== null && displayName !== '') {
-				this.showSaving = true
+			if (displayName.trim() === '') {
+				showError(this.t('mail', 'Tag name cannot be empty'))
+				return
 			}
 			try {
 				await this.$store.dispatch('createTag', {


hamza221:
I was thinking of just disabling the button if there's no text

src/components/TagModal.vue
```diff
@@ -177,7 +177,11 @@ export default {
 				showError(this.t('mail', 'Tag already exists'))
 				return
 			}
-			if (displayName !== null && displayName !== '') {
+			if (displayName.trim() === '') {
+				showError(this.t('mail', 'Tag name cannot be empty'))
+				return
+			}
+			if (displayName !== '') {
```
ChristophWurst, line 184: isn't this check superfluous if you're testing the trimmed name for emptyness above?

GretaD: true, i didnt add this one, it was there, but i should have removed it instead of simplifying it.

ChristophWurst (approved):

PR #9091, opened 2023-11-24
fix: add name_hash as nullable

### LINKED ISSUE #9089: Postgres error upgrading to v3.5.0 RC1: column "name_hash" contains null values
### Steps to reproduce

1. Update to Mail v3.5.0 RC1

### Expected behavior

Smooth upgrade

### Actual behavior

```
Updating <mail> ...
Exception: Database error when running migration 3500Date20231115182612 for app mail
An exception occurred while executing a query: SQLSTATE[23502]: Not null violation: 7 ERROR:  column "name_hash" contains null values
Update failed
```

### Mail app version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

Caused by https://github.com/nextcloud/mail/pull/9071

kesselb:
Fix #9089 

Otherwise the migration on pgsql don't work.

### DIFF
--- a/lib/Migration/Version3500Date20231115184458.php
+++ b/lib/Migration/Version3500Date20231115184458.php
@@ -56,6 +56,11 @@ class Version3500Date20231115184458 extends SimpleMigrationStep {
 			$mailboxesTable->addUniqueIndex(['account_id', 'name_hash'], $indexNew);
 		}
 
+		$nameHashColumn = $mailboxesTable->getColumn('name_hash');
+		if (!$nameHashColumn->getNotnull()) {
+			$nameHashColumn->setNotnull(true);
+		}
+
 		return $schema;
 	}
 }
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -29,7 +29,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>3.6.0-alpha.2</version>
+	<version>3.6.0-alpha.3</version>
 	<licence>agpl</licence>
 	<author>Christoph Wurst</author>
 	<author homepage="https://github.com/nextcloud/groupware">Nextcloud Groupware Team</author>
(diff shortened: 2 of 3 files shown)


lib/Migration/Version3500Date20231115184458.php
```diff
@@ -56,6 +56,11 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$mailboxesTable->addUniqueIndex(['account_id', 'name_hash'], $indexNew);
 		}
 
+		$nameHashColumn = $mailboxesTable->getColumn('name_hash');
+		if (!$nameHashColumn->getNotnull()) {
+			$nameHashColumn->setNotnull(true);
```
kesselb, line 61: Unsure about this one. 

The default is still NULL and the value is omitted on insert a not null violation is thrown. 
Do we need a new default as well?

```
-- auto-generated definition
create table oc_mail_mailboxes
(
    id                  serial
        primary key,
    name                varchar(255)                                 not null,
    account_id          integer                                      not null,
    attributes          varchar(255) default '[]'::character varying not null,
    delimiter           varchar(1)   default NULL::character varying,
    messages            integer                                      not null,
    unseen              integer                                      not null,
    selectable          boolean      default false,
    special_use         varchar(255) default '[]'::character varying not null,
    sync_new_lock       integer,
    sync_changed_lock   integer,
    sync_vanished_lock  integer,
    sync_new_token      varchar(255) default NULL::character varying,
    sync_changed_token  varchar(255) default NULL::character varying,
    sync_vanished_token varchar(255) default NULL::character varying,
    sync_in_background  boolean      default false,
    my_acls             varchar(32)  default NULL::character varying,
    shared              boolean      default false,
    name_hash           varchar(255) default NULL::character varying not null
);
```

ChristophWurst: Don't we want the opposite, and make the column nullable?

*not null* and violation when inserting with out a value is expected

ChristophWurst: When we change not from not null to null here, all rows should have their name_hash filled. And for inserts we have to make sure to always write with the hash. If we insert without, there will be an error. I think that is fine and nothing we can improve.

ChristophWurst (approved):
Tested with a pg setup. Could trigger the bug with an existing account before going to the original migration. The migration passes with the changes of this PR.

Thank you :pray:

PR #9115, opened 2023-11-30
style: Import external library files with their extension

ChristophWurst:
In preparation of the eslint 8 upgrade.

This was automated with some regexes. I might have missed some imports. I'll fix them in the es8 upgrade PR.


st3iny (approved):
:elephant:

Issue #5149, opened 2021-06-09
Not all tags are readable due to background contrast


ChristophWurst:
### Expected behavior
Tags should be readable, despite the picked color

### Actual behavior
![Bildschirmfoto von 2021-06-09 17-41-28](https://user-images.githubusercontent.com/1374172/121386146-0abf3c80-c94a-11eb-96ce-6ea8d5448daa.png)

### Mail app

**Mail app version:** master

meichthys:
Also relevant in Dark Mode:
![image](https://user-images.githubusercontent.com/10717998/186681535-3166bc43-d4dd-4e58-a76e-7a6da46e92db.png)

meichthys:
For me, the issue only seems to be apparent in the "Edit Tags" dialog, not on the email itself.

ChristophWurst:
![Bildschirmfoto vom 2023-03-22 17-39-59](https://user-images.githubusercontent.com/1374172/226975900-7778d628-aa01-49c4-b2c5-3532f00a5e6c.png)

still happens and it make the app look unpolished.

GretaD:
@miaulalala we had a brainstorm together related to this some time ago: https://github.com/nextcloud/mail/pull/5217

is there anything on backend that can be done related to this? Or should i look it only on frontend part?

ChristophWurst:
The default color is unfortunate but you should still expect ever color to work if the user can pick one.

GretaD:
> The default color is unfortunate but you should still expect ever color to work if the user can pick one.

the colors that are unreadable are tags that are auto-generated, like junk, redirected, spam, pishing ect. I doubt these tags should be shown there at the first place.

PR #9116, opened 2023-11-30
fix(tags): Make unreadable tags readable

ChristophWurst:
Make bright on white and dark on black tag labels readable.

| Before | After |
|--------|--------|
| Bildschirmfoto vom 2023-12-01 18-14-07 | Bildschirmfoto vom 2023-12-01 18-13-44 | 

## To do

- [x] Envelope list tags
- [x] Thread header tags
- [x] Tag modal tags

Fixes https://github.com/nextcloud/mail/issues/5149


ChristophWurst:
cc @meichthys if you have a setup to test changes

ChristophWurst:
Envelope list needs a fix, too

ChristophWurst:
| Before | After |
|--------|--------|
| Bildschirmfoto vom 2023-11-30 20-39-14 | Bildschirmfoto vom 2023-11-30 20-42-49 | 

^ @nimishavijay colored background and colored font is touch to get right. Would this be ok instead? I'm using the default font color + 15% opacity color background. Don't throw specific contrast requirements at me but I think this becomes a lot more readable than before.

ChristophWurst:
@nimishavijay pleaes see https://github.com/nextcloud/mail/pull/9116#issue-2018503114. It's now always the same tag component and therefore has consistent styling. The pill radius is taken from the css variable for consistency with the rest of Nextcloud. Tags are more compact now to improve the information density.

nimishavijay:
Pill radius and height adjustment is great 🚀 I'm only not sure about the text color, the style with the colored text looked really nice! Perhaps we could use a darker version of the selected color for the text and a lighter version for the bg, what do you think?

ChristophWurst:
> Perhaps we could use a darker version of the selected color for the text and a lighter version for the bg, what do you think?

Do you know a way to make this work with all possible RGB colors?

nimishavijay:
How about if you select a color, we use it darkened by 60% for text and lightened by 80% for bg? I used [this tool](https://noeldelgado.github.io/shadowlord/#8b37d0) to play around with it, contrast seems to be fine 

Default tag colors and white and black look like this: 

image

ChristophWurst:
| Before | After |
|--------|--------|
| Bildschirmfoto vom 2023-12-01 18-14-07 | Bildschirmfoto vom 2023-12-01 18-13-44 | 

like this?

GretaD:
I tested it with the junk tag that is autogenerated. the result is the one below. The dark theme is still shown white for me.
Screenshot from 2023-12-27 14-30-57
Screenshot from 2023-12-27 14-30-27

ChristophWurst:
Has to be redone

PR #9121, opened 2023-11-30
Feat/show email on hovering a `RecipientListItem` in the new email modal 

### LINKED ISSUE #5074: Show email addresses when hovering over recipient list name
### Feature Request

After selecting recipients on the list, it might be helpful to show their email address, when one hovers with the mouse above the name.

#### Summary

The "problem" is that sometimes people have the same name with different addresses, and, after eg replying to a message, the name is added, one cannot see the email address anymore.

GVodyanov:
Fixes #5074


ChristophWurst (approved):
Tested & works

I can see the full email address even for long display names where the email address is otherwise cut

Issue #9132, opened 2023-12-04
Never ending desktop notifications for new emails


ChristophWurst:
### Steps to reproduce

1. Set up an account
2. Leave the tab alone
3. Receive a new email

### Expected behavior

I get a desktop notification *once*

### Actual behavior

I get a desktop notification about the very same message *every 30 seconds*.

### Mail app version

3.5.0 RC3

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
This could be caused by https://github.com/nextcloud/mail/pull/8231.

@hamza221 could you try to reproduce?

ChristophWurst:
![image](https://github.com/nextcloud/mail/assets/1374172/c4e5d35f-56f0-49bf-8a1e-11da8bf58eeb)


the message is constantly reported as new

ChristophWurst:
The thread has two latest messages because they were sent at the exact same unix timestamp. This seems to make the sorting toggle between one or the other message constantly, therefore creating desktop notifications.

We can lower the priority of this fix

ChristophWurst:
Happens on my prod instance too now using Mail v3.5.0-rc.3

ChristophWurst:
| Request | Response |
|--------|--------|
| ![Bildschirmfoto vom 2023-12-11 08-45-31](https://github.com/nextcloud/mail/assets/1374172/5d3b89ae-8bd0-487f-9ef2-7b3e5e1aef7e) | ![Bildschirmfoto vom 2023-12-11 08-45-45](https://github.com/nextcloud/mail/assets/1374172/204f942d-f373-4015-9166-269e1bbaa333) | 

It is the request that syncs my INBOX in the background, **not** the sync of the priority inbox that is shown.

On the console I see some related errors:

![image](https://github.com/nextcloud/mail/assets/1374172/d7df9031-c65a-45a7-9612-cee7ae42b376)

Which tells us that the messages are **removed on thefrontend**. With the next background sync, they are brought back. Then deleted, synced again, deleted …

ChristophWurst:
Setting priority back to high because this is quite annoying to get the notification every 30s. It makes me close the Mail tab.

ChristophWurst:
It only happens with more than one account

hamza221:
Only one account set
https://github.com/nextcloud/mail/assets/40746210/ed33e2e7-68be-42d6-a637-889020d8932e

ChristophWurst:
[...]

PR #9157, opened 2023-12-11
fix: Alien envelopes that vanish and reappear continuously

ChristophWurst:
When adding envelopes to the store we must only append them to the mailbox they belong to, or the unified inbox. If they are added to any other mailbox, the sync removes them, adds them back, removes them, adds them, removes them, ...

## How to test

1. Add account 1
2. Add account 2
3. Reload the page
4. Watch

main: some messages continuously vanish and return
here: view stays unchanged even with background sync

Fixes https://github.com/nextcloud/mail/issues/9132

### DIFF
--- a/src/store/mutations.js
+++ b/src/store/mutations.js
@@ -284,13 +284,13 @@ export default {
 			return
 		}
 
-		const mailbox = state.mailboxes[envelopes[0].mailboxId]
 		const idToDateInt = (id) => state.envelopes[id].dateInt
 
 		const listId = normalizedEnvelopeListId(query)
 		const orderByDateInt = orderBy(idToDateInt, state.preferences['sort-order'] === 'newest' ? 'desc' : 'asc')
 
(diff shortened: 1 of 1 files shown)


src/store/mutations.js
```diff
@@ -284,13 +284,13 @@ export default {
 			return
 		}
 
-		const mailbox = state.mailboxes[envelopes[0].mailboxId]
```
ChristophWurst, line 287: assumes that all envelopes passed have the same mailboxId. That is not true for unified mailboxes because their envelopes still point to the mailboxId of their original mailbox.

hamza221:
I can still reproduce the 'same timestamp' bug tho

ChristophWurst:
yes, me too

ChristophWurst:
there must be two infinite sync bugs

hamza221 (approved):

PR #9163, opened 2023-12-12
Add an option to maximize composer modal

### LINKED ISSUE #9100: Add a toggle to the existing modal for different sizes.

Right now we have only one size composer modal. For some people that use big screens, that is too small. Lets add a toggle that changes the composer modal size, depending on the user preference.

Add a toggle that changes the composer modal size, depending on the user preference. Large size should not be full screen.

_No response_

### Additional context

Ref https://github.com/nextcloud/mail/issues/6589
ref https://github.com/nextcloud/mail/pull/6717

GretaD:
Screenshot from 2023-12-14 16-22-53
Screenshot from 2023-12-14 16-22-43

fixes #9100

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -1,6 +1,6 @@
 <template>
 	<Modal v-if="showMessageComposer"
-		size="normal"
+		:size="largerModal ? 'large' : 'normal'"
 		:title="modalTitle"
 		:additional-trap-elements="toolbarElements"
 		@close="$event.type === 'click' ? onClose() : onMinimize()">
@@ -10,16 +10,16 @@
 			role="alert">
 			<p>{{ error }}</p>
 			<template #action>
-				<ButtonVue type="tertiary"
+				<NcButton type="tertiary"
 					:aria-label="t('mail', 'Go back')"
 					@click="error = undefined">
 					{{ t('mail', 'Go back') }}
-				</ButtonVue>
-				<ButtonVue type="tertiary"
+				</NcButton>
+				<NcButton type="tertiary"
 					:aria-label="t('mail', 'Retry')"
 					@click="onSend">
 					{{ t('mail', 'Retry') }}
-				</ButtonVue>
+				</NcButton>
 			</template>
 		</EmptyContent>
 		<Loading v-else-if="uploadingAttachments" :hint="t('mail', 'Uploading attachments …')" role="alert" />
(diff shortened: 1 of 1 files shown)


nimishavijay:
This is actually a great start, some follow up points:

- [ ] I would suggest using the [open-in-full MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aopen_in_full%3A) instead as it is clear that it is expanding the size, and make sure that all buttons are 44px to avoid misclicks. 
- [ ] This button should also not be shown for every screen size, as for very small screens (eg. mobile) the composer is already full screen. 
- [ ] The expanded composer can be upto 80vw and 90vh: that results in something like this on my 1920x1080 laptop:
image
And for larger screens (Gmail has it even bigger, [here is a screenshot](https://github.com/nextcloud/mail/assets/52440189/06358fe8-781d-45b4-840d-5a23ffd94c23)):
image
- [ ] The expanded composer should also have a minimize button instead of expand and it can use this [close fullscreen icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aclose_fullscreen%3A)
- [ ] The setting is remembered per user 

ccing the rest of @nextcloud/designers as there was some discussion around this :)

GretaD:
Nice, thank you nimisha, i will wait for more feedback then, because we have time :)

jancborchardt (commented):
Agree with @nimishavijay’s feedback :) no further feedback at this stage

GretaD:
> The setting is remembered per user

As i can see from other mail clients, like gmail and outlook, this is not remembered, their default size(the wider one) is always open, no matter what size you had opened last. Should we still remember the last size used by the user? cc @jancborchardt @nimishavijay

jancborchardt:
@GretaD yes, remembering the size is a really nice quality of life feature. Btw when testing with Gmail it does not open the bigger one by default but rather the smaller one on the bottom right always.

GretaD:
> @GretaD yes, remembering the size is a really nice quality of life feature. Btw when testing with Gmail it does not open the bigger one by default but rather the smaller one on the bottom right always.
yes, that one, sorry, but the point was that they dont save the last size. 

sure, if we want to save ours, then lets do that.

hamza221:
> The expanded composer should also have a minimize button instead of expand and it can use this [close fullscreen icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aclose_fullscreen%3A)

I think the Icons are reversed 
| Expanded | Minimized |
|--------|--------|
| image | image |

src/components/NewMessageModal.vue
```diff
@@ -34,76 +34,84 @@
...
-				@upload-attachment="onAttachmentUploading"
-				@send="onSend"
-				@show-toolbar="handleShow" />
+			<NcButton class="maximize-button"
+				type="tertiary-no-background"
+				:aria-label="t('mail', 'Maximize composer')"
+				@click="onMaximize">
+				<template #icon>
+					<MaximizeIcon v-if="!largerModal" :size="20" />
+					<DefaultComposerIcon v-if="largerModal" :size="20" />
```
hamza221, line 56:
```suggestion
					<DefaultComposerIcon v-else :size="20" />
```

src/components/NewMessageModal.vue
```diff
@@ -195,6 +206,10 @@ export default {
 		this.updateCookedComposerData()
 	},
 	methods: {
+		onMaximize() {
+			this.largerModal = !this.largerModal
+			localStorage.setItem('modalSize', this.largerModal ? 'large' : 'normal')
```
hamza221, line 211: The problem with local storage is that it's gonna be only remembered on the same browser. I think we better user user Preferences here

hamza221 (changes requested):
The button still exits in dom in mobile view, it's just not visible because of lack of space. 
I think it should have display:none for small viewports

GretaD:
> > The expanded composer should also have a minimize button instead of expand and it can use this [close fullscreen icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aclose_fullscreen%3A)
> 
> * [ ]  I think the Icons are reversed
> 
> Expanded	Minimized
> image	image
> * [ ]  It's not remembered

do you have the latest changes, my lart fix up fixed exactly these 2 points

src/components/NewMessageModal.vue
GretaD: ok, good point

hamza221:
> do you have the latest changes, my lart fix up fixed exactly these 2 points

My bad, turns out I was not on latest commit, The Icons are correct indeed

src/components/NewMessageModal.vue
```diff
@@ -34,76 +34,84 @@
...
+			<NcButton class="minimize-button"
+				type="tertiary-no-background"
+				:aria-label="t('mail', 'Minimize composer')"
+				@click="onMinimize">
+				<template #icon>
+					<MinimizeIcon :size="20" />
+				</template>
+			</NcButton>
 		</template>
+		<Composer ref="composer"
```
ChristophWurst, line 68: I don't think the composer should move out of the v-else template. It means the composer would also render when the emptycontent view shows

GretaD: I don't remember moving the template closure on purpose, so I'm assuming it's done by ESLint while i was moving code around. Will push a fixup soon, if you dont have any other comment.

ChristophWurst (approved):
Tested & works

Issue #9098, opened 2023-11-27
Left-trim message preview text from whitespace


ChristophWurst:
### Steps to reproduce

![Bildschirmfoto vom 2023-11-27 18-45-35](https://github.com/nextcloud/mail/assets/1374172/f5206e8d-df4f-4c78-8759-840e5b1b22ec)

1. Send a message to yourself, make it start with whitespace, like "          hello"
2. Wait for the message to show in the inbox
3. Look at the preview text

### Expected behavior

No whitespace. Preview text is left-aligned with sender and subject text

### Actual behavior

![image](https://github.com/nextcloud/mail/assets/1374172/82518c73-1157-4c6d-947a-870abbad2f20)

### Mail app version

3.5.0 rc2

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
We can fix this in two ways:

1) Trim the text in the front-end (also works for historic, cached data)
2) Trim the text in the back-end (works only for new messages)

If we do both we are on the save side.

ChristophWurst:
The backend persists the text in ``\OCA\Mail\Db\MessageMapper::updatePreviewDataBulk``

PR #9174, opened 2023-12-14
Enh/trim message preview

GVodyanov:
Close https://github.com/nextcloud/mail/issues/9098

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -507,15 +507,10 @@ class MessageMapper extends QBMapper {
 				$query->setParameter('flag_attachments', $message->getFlagAttachments(), $message->getFlagAttachments() === null ? IQueryBuilder::PARAM_NULL : IQueryBuilder::PARAM_BOOL);
 				$previewText = null;
 				if ($message->getPreviewText() !== null) {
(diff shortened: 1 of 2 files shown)


lib/Db/MessageMapper.php
```diff
@@ -508,6 +508,9 @@ public function updatePreviewDataBulk(Message ...$messages): array {
 				$previewText = null;
 				if ($message->getPreviewText() !== null) {
 					$previewText = mb_strcut(mb_convert_encoding($message->getPreviewText(), 'UTF-8', 'UTF-8'), 0, 255);
+
+					//converting the spaces is necessary for ltrim to work
```
hamza221, line 512:
```suggestion
						$convertedText = mb_convert_encoding($message->getPreviewText(), 'UTF-8', 'UTF-8');
					//converting the spaces is necessary for ltrim to work
					$previewText = mb_strcut(ltrim(preg_replace('/\s/u', ' ', $convertedText)), 0, 255);

					// Make sure modifications are visible when these objects are used right away
					$message->setPreviewText($previewText);
```
I'm still not sure why we're converting from UTF-8 to UTF-8,
but I think this way we can keep the 255 characters limit in case of an edge-case where a message starts with a huge amount of spaces

hamza221 (approved):
Squash commits then we can merge 🚀

Issue #9109, opened 2023-11-30
Remove margin-right of mailbox counter if it has sub mailboxes


ChristophWurst:
### Steps to reproduce

1. Configure an account
2. Have or create a two level mailbox hierarchy
3. Have or create unread messages in the first level mailbox
4. Look at the sidebar

### Expected behavior

Efficient use of space

### Actual behavior

| :framed_picture:  | :mag:  | :mag_right:  |
|--------|--------|--------|
| ![Bildschirmfoto vom 2023-11-30 08-11-10](https://github.com/nextcloud/mail/assets/1374172/9f24c90e-5fc8-4cfe-83cc-52afefd8c179) | ![Bildschirmfoto vom 2023-11-30 08-11-58](https://github.com/nextcloud/mail/assets/1374172/8c12bb24-4800-4960-9726-e93a4ba27b9f) | ![Bildschirmfoto vom 2023-11-30 08-12-11](https://github.com/nextcloud/mail/assets/1374172/ffe94311-0a9b-43f3-898e-ffcd92e39504) | 

### Mail app version

3.5

### PHP engine version

None

### Web server

None

### Database

None

PR #9179, opened 2023-12-17
Enh/align-mailbox-counter

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/9109

I understand that not having scoped CSS in this case might not be the best, suggestions to improve appreciated.

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -34,6 +34,7 @@
 		:title="title"
 		:to="to"
 		:open.sync="showSubMailboxes"
+		:class="{ subMailBoxCounter: showSubMailboxes }"
 		@update:menuOpen="onMenuToggle">
 		<template #icon>
 			<div>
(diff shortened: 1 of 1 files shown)


GretaD:
please add a before and after screenshot for design tickets :)

src/components/NavigationMailbox.vue
```diff
@@ -679,3 +680,9 @@ export default {
 	max-width: initial;
 }
 </style>
+
+<style lang="scss">
+.subMailBoxCounter .app-navigation-entry__counter-wrapper {
```
ChristophWurst, line 685: `.app-navigation-entry__counter-wrapper` is a selector for an internal element of AppNavigationItem. This works, but it's fragile. Internal restructuring of the component may break this CSS rule.

Could we fix this upstream in `@nextcloud/vue`?

Positive side effect: the issue will also be fixed for any other apps that use AppNavigationItem with a counter.

PR #9184, opened 2023-12-18
Fix/make smime certificate modal responsive

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/8082


ChristophWurst (approved):
Is not response on main but here :+1:

Issue #9160, opened 2023-12-12
Slow tag orphan cleanup due to missing index


ChristophWurst:
### Steps to reproduce

1. Run `SELECT mt.id FROM oc_mail_messages m
    RIGHT JOIN oc_mail_message_tags mt ON mt.imap_message_id = m.message_id
    WHERE m.message_id IS NULL;`

### Expected behavior

Fast result

### Actual behavior

Slow result

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

Suggested index is
```sql
create index oc_mail_messages_msgid_idx
    on oc_mail_messages (`message_id(128)`);
```

miaulalala:
Applied on a large instance, worked well.

nickvergessen:
Correct syntax seems to be:
```sql
ALTER TABLE `oc_mail_messages` ADD INDEX `oc_mail_messages_msgid_idx` (`message_id`(128)); 
```

ChristophWurst:
Adding the index can take a few minutes. Let's use the optional indices API and not add the index in a migration. But let's still add the index to the existing migrations so it's added for new installations.

nickvergessen:
> Adding the index can take a few minutes. Let's use the optional indices API and not add the index in a migration. 

4.22 seconds with 1.28m mails here. Sure you can try the new API, but we did worse things in the past

ChristophWurst:
It took 5mins for an affected customer

PR #9206, opened 2023-12-28
perf(db): Add message_id index for mail_messages table

ChristophWurst:
* Index created by default for new installations
* Optional index for existing installations

Fixes https://github.com/nextcloud/mail/issues/9160

## How to test

1) Set up a new installation

OR

1) Have an existing installation
2) Run `occ db:add-missing-indices`

main: no index
this branch: index

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -66,6 +66,7 @@ use OCA\Mail\Listener\MessageKnownSinceListener;
 use OCA\Mail\Listener\MoveJunkListener;
 use OCA\Mail\Listener\NewMessageClassificationListener;
 use OCA\Mail\Listener\OauthTokenRefreshListener;
+use OCA\Mail\Listener\OptionalIndicesListener;
 use OCA\Mail\Listener\OutOfOfficeListener;
 use OCA\Mail\Listener\SaveSentMessageListener;
 use OCA\Mail\Listener\SpamReportListener;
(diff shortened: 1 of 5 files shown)


lib/Migration/Version1040Date20200422142920.php
```diff
@@ -111,6 +111,13 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			'mail_msg_mb_uid_idx'
 		);
 		$messagesTable->addIndex(['sent_at'], 'mail_msg_sent_idx');
+		// mail_messages_msgid_idx was added later and may not exist until optional indices are created
+		$messagesTable->addIndex(
+			['message_id'],
+			'mail_messages_msgid_idx',
+			[],
+			['lengths' => [128]],
+		);
```
JohannesGGE, line 120: This won't end up as an index in the table at new instance setup because we remove all indices in the next migration.

https://github.com/nextcloud/mail/blob/f50605daaa5e21fde1e46abd645b22589b628276/lib/Migration/Version1130Date20220412111833.php#L160-L165

I'm not sure why we do this, but it will probably lead to further issues when editing indices in former migrations. Already happened with `mail_msg_sent_idx` on `sent_at` which wasn't add again and the unique index `mail_msg_mb_uid_idx` on `['uid','mailbox_id']` which was removed but not added again right away. It was [added again](https://github.com/nextcloud/mail/pull/6509) after some time as a non-unique index `mail_messages_mb_id_uid` when perf issues occurred.

JohannesGGE (commented):
Otherwise the index and the `occ db:add-missing-indices` work.

JohannesGGE (approved):

miaulalala (approved):

PR #9208, opened 2023-12-29
Fix: select alias

### LINKED ISSUE #9196: Unable to select alias to send from
### Steps to reproduce

1. Add alias to existing account
2. Compose new message
3. Attempt to select alias in "From" dropdown

### Expected behavior

Selecting the alias in the dropdown should work

### Actual behavior

Clicking on the alias in the dropdown has no effect. Unable to send as alias.

### Mail app version

3.5.0

### Mailserver or service

ISP

### Operating system

Linux

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

No errors or warnings produced in backend or frontend, even with debug logging enabled.

hamza221:
Fix #9196

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -10,7 +10,7 @@
 					:value="selectedAlias"
 					:options="aliases"
 					label="name"
-					track-by="selectId"
+					:get-option-key="(option)=>option.selectId"
 					:searchable="false"
 					:placeholder="t('mail', 'Select account')"
 					:clear-on-select="false"


src/components/Composer.vue
```diff
@@ -10,7 +10,7 @@
 					:value="selectedAlias"
 					:options="aliases"
 					label="name"
-					track-by="selectId"
+					:get-option-key="(option)=>option.selectId"
```
ChristophWurst, line 13: we use track-by in other places too. is that a problem everywhere?

ChristophWurst: apparently this is only needed if the objects haven an `id` and we don't want that value to be used for identification https://vue-select.org/api/props.html#getoptionkey

hamza221: `track-by` doesn't exist in [NcSelect](https://vue-select.org/api/props.html#getoptionkey). So If we have options we same Key, we would probably run into the same error

hamza221: By default key = option.id

hamza221: > apparently this is only needed if the objects haven an `id` and we don't want that value to be used for identification https://vue-select.org/api/props.html#getoptionkey

Yes in this case the alias has the same Id as the account

GretaD (approved):

Issue #9170, opened 2023-12-13
Undefined array key when runing Auto Config


solracsf:
### Steps to reproduce

1. Run auto-config

### Expected behavior

No Undefined array key

### Actual behavior

Undefined array key

### Mail app version

3.5.0

### Mailserver or service

rzone.de

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

```
{
  "reqId": "dGacsqzhQnrOpfijmA7K",
  "level": 3,
  "time": "2023-12-11T19:11:38+01:00",
  "remoteAddr": "REDACTED",
  "user": "REDACTED@REDACTED.com",
  "app": "PHP",
  "method": "GET",
  "url": "/apps/mail/api/autoconfig/ispdb/REDACTED%40REDACTED.com",
  "message": "Undefined array key 1 at /var/www/nextcloud/lib/private/AppFramework/Utility/ControllerMethodReflector.php#65",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36",
  "version": "27.1.5.0",
  "data": {
    "app": "PHP"
  },
  "id": "6579affe4ebce"
}
```

PR #9216, opened 2024-01-05
fix(setup): Fix rate limit annotation syntax

brad2014:
Fixes nextcloud/mail#9170.

Automatic method reflector code expects "=" not ":" in Doc parameter lists.  Ref: https://github.com/nextcloud/server/blob/7502c19ddd43853c3b4fad1e2df91aed19e6b626/lib/private/AppFramework/Utility/ControllerMethodReflector.php#L66

In addition to the above, I had to do the following (I don't know if it should be allowed by mail app by default):

```
% php occ config:system:set --value true --type boolean allow_local_remote_servers
```

### DIFF
--- a/lib/Controller/AutoConfigController.php
+++ b/lib/Controller/AutoConfigController.php
@@ -61,7 +61,7 @@ class AutoConfigController extends Controller {
 	 * @param string $email
 	 *
 	 * @NoAdminRequired
-	 * @UserRateThrottle(limit: 5, period: 60)
+	 * @UserRateThrottle(limit=5, period=60)
 	 *
 	 * @return JsonResponse
 	 */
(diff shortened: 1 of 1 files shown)


ChristophWurst:
@brad2014 thanks a lot for the fixes!

Could you please submit them separately? That makes automatic backports to older versions alot easier.

brad2014:
> @brad2014 thanks a lot for the fixes!
> 
> Could you please submit them separately? That makes automatic backports to older versions a lot easier.

I backed out the second change related to fixing the other bug, which I filed separately as #9221, and will generate a separate PR.

ChristophWurst (approved):

ChristophWurst:
@brad2014 mind giving those manual backports a try? Git might have detected some small conflicts. That makes the bot fail the fully automated port.

ChristophWurst:
https://docs.nextcloud.com/server/latest/developer_manual/getting_started/development_process.html#manual-backport

PR #9227, opened 2024-01-08
[stable2.2] fix(setup): Fix rate limit annotation syntax

brad2014:
This PR fixes nextcloud/mail#9170.

- Automatic method reflector code expects "=" not ":" in Doc parameter lists. Ref: https://github.com/nextcloud/server/blob/7502c19ddd43853c3b4fad1e2df91aed19e6b626/lib/private/AppFramework/Utility/ControllerMethodReflector.php#L66


ChristophWurst (approved):
Looks good. TrapError was an annotation for 2.2 and is an attribute for 3.x

Issue #9256, opened 2024-01-18
Classifier clean-up is slower than new classifier build, up


ChristophWurst:
### Steps to reproduce

1. Have an instance with ~100 Mail accounts

### Expected behavior

Database table mail_classifiers stays reasonably small

### Actual behavior

Database table mail_classifiers grows because 
* `\OCA\Mail\BackgroundJob\SyncJob` runs 24*100 times a day
* `\OCA\Mail\BackgroundJob\CleanupJob` runce once a day, clearing 100 classifiers

### Mail app version

3.5

### PHP engine version

None

### Web server

None

### Database

None

PR #9267, opened 2024-01-22
fix(clean-up): remove all outdated classifier not only 100 per day

JohannesGGE:
Fix: #9256 

Instead of removing a fixed number of 100 classifiers which are older than two months, now all old classifiers are removed on cleanup. So only one classifier per type and account remains.

## How to test

1) Set up an account
2) Run `mail:account:train` 10x 
3) Run `SELECT COUNT(*) FROM oc_mail_classifiers` and remember the number
4) Run `occ mail:clean-up`
5) Run `SELECT COUNT(*) FROM oc_mail_classifiers` and compare it to the number before


ChristophWurst:
This is tricky. I am not sure if we should wipe everything at a time. The reasoning is that we are not only wiping database rows but clear files from app data storage.

The instance we monitored has 500k classifiers that can go. My gut feeling says deleting 500k rows + 500k files is not going to be done quickly.

ChristophWurst (approved):

Issue #9061, opened 2023-11-14
QuotaJob: Division by zero


kesselb:
### Steps to reproduce

1. Run the quota job
2. See a division by zero error
3. :disappointed: 

### Expected behavior

No division by zero ;)

### Actual behavior

From a customer log:

```json
{
  "reqId": "",
  "level": 3,
  "time": "October 31, 2023 04:02:52",
  "remoteAddr": "",
  "user": "",
  "app": "core",
  "method": "",
  "url": "--",
  "message": "Error while running background job (class: OCA\\Mail\\BackgroundJob\\QuotaJob, arguments: Array\n(\n    [accountId] => someAccount\n)\n)",
  "userAgent": "--",
  "version": "27.1.2.2",
  "exception": {
    "Exception": "DivisionByZeroError",
    "Message": "Division by zero",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/apps/mail/lib/BackgroundJob/QuotaJob.php",
        "line": 98,
        "function": "calculateAndSetQuotaPercentage",
        "class": "OCA\\Mail\\Account",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/public/BackgroundJob/Job.php",
        "line": 81,
        "function": "run",
        "class": "OCA\\Mail\\BackgroundJob\\QuotaJob",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/public/BackgroundJob/TimedJob.php",
        "line": 103,
        "function": "start",
        "class": "OCP\\BackgroundJob\\Job",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/public/BackgroundJob/TimedJob.php",
        "line": 93,
        "function": "start",
        "class": "OCP\\BackgroundJob\\TimedJob",
        "type": "->"
      },
      {
        "file": "/var/www/html/cron.php",
        "line": 152,
        "function": "execute",
        "class": "OCP\\BackgroundJob\\TimedJob",
        "type": "->"
      }
    ],
    "File": "/var/www/html/apps/mail/lib/Account.php",
    "Line": 240,
    "message": "Error while running background job (class: OCA\\Mail\\BackgroundJob\\QuotaJob, arguments: Array\n(\n    [accountId] => someAccount\n)\n)",
    "exception": {},
[...]

PR #9269, opened 2024-01-23
fix(quota): don't divide by limit zero

miaulalala:
Fixes https://github.com/nextcloud/mail/issues/9061

If the limit is 0, don't do the maths as the maths is bad.

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -218,6 +218,10 @@ class Account implements JsonSerializable {
 	 * @return void
 	 */
 	public function calculateAndSetQuotaPercentage(Quota $quota): void {
+		if ($quota->getLimit() === 0) {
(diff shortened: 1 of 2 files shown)


lib/Account.php
```diff
@@ -218,6 +218,11 @@ public function newMessage() {
 	 * @return void
 	 */
 	public function calculateAndSetQuotaPercentage(Quota $quota): void {
+		// No division by 0 pls
+		if($quota->getLimit() === 0) {
```
ChristophWurst, line 222:
```suggestion
		if ($quota->getLimit() === 0) {
```

ChristophWurst (approved):

solracsf:
Fixes the issue on v27 🆗

solracsf (approved):

PR #9283, opened 2024-01-30
Add aria-label-combobox on NcSelect

GretaD:
Its necessary for accessibility, see here: https://github.com/nextcloud-libraries/nextcloud-vue/commit/ba409b3340c199d1d5396936a56fc832315a86a8#diff-e220135f30cfa99bbb7f8f8c79e4d517b44dfacf958bd2a80ecfeb4a8a471a63R589

### DIFF
--- a/src/components/TaskModal.vue
+++ b/src/components/TaskModal.vue
@@ -38,6 +38,7 @@
 				label="displayname"
 				track-by="url"
 				:placeholder="t('mail', 'Select calendar')"
+				:aria-label-combobox="t('mail', 'Select calendar')"
 				:allow-empty="false"
 				:options="calendars">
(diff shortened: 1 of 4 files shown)


src/components/TaskModal.vue
```diff
@@ -38,6 +38,7 @@
 				label="displayname"
 				track-by="url"
 				:placeholder="t('mail', 'Select calendar')"
+				:aria-label-combobox="t('mail','Select calendar')"
```
ChristophWurst, line 41:
```suggestion
				:aria-label-combobox="t('mail', 'Select calendar')"
```

GretaD: where is lint autofix when i need it?

Issue #8322, opened 2023-04-01
kitinerary-extractor: Permission denied


mercury1337:
### Steps to reproduce

1. Installed Mail App
2. Saw the following messages in the log

sh: 1: /var/www/nextcloud/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: Permission denied
sh: 1: flatpak: not found
sh: 1: kitinerary-extractor: not found

### Expected behavior

No error message

### Actual behavior

sh: 1: /var/www/nextcloud/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: Permission denied
sh: 1: flatpak: not found
sh: 1: kitinerary-extractor: not found

### Mail app version

3.0.2

### Operating system

Ubuntu 22.04.2 LTS

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MySQL

### Additional info

I found https://github.com/nextcloud/mail/issues/2489 and it asked to raise a new issue if the problem occurs again => done

judicandus:
Brand new installation of nextcloud using traefik + nextcloud-aio, installed Mail, encountered the same errors in the logs:

christinloehner:
It is still present in Nextcloud Hub 8 (29.0.8) .....  :-(

yankees5963:
Still having this issue with NC 34.0 using k8s rootless

Issue #7899, opened 2023-01-14
flatpak and kitinerary-extractor not found - PHP Warning


timreeves:
### Steps to reproduce

Simply use the mail app

### Expected behavior

No PHP Warning

### Actual behavior

[14-Jan-2023 12:20:21] WARNING: [pool owncloud.reeves.one] child 19025 said into stderr: "/var/www/vhosts/reeves.one/own
cloud.reeves.one/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: /lib/x86_64-linux-gnu/l
ibm.so.6: version `GLIBC_2.29' not found (required by /var/www/vhosts/reeves.one/owncloud.reeves.one/apps/mail/vendor/ch
ristophwurst/kitinerary-bin/src/../bin/kitinerary-extractor)"
[14-Jan-2023 12:20:21] WARNING: [pool owncloud.reeves.one] child 19025 said into stderr: "/var/www/vhosts/reeves.one/own
cloud.reeves.one/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: /lib/x86_64-linux-gnu/l
ibm.so.6: version `GLIBC_2.35' not found (required by /var/www/vhosts/reeves.one/owncloud.reeves.one/apps/mail/vendor/ch
ristophwurst/kitinerary-bin/src/../bin/kitinerary-extractor)"
[14-Jan-2023 12:20:21] WARNING: [pool owncloud.reeves.one] child 19025 said into stderr: "/var/www/vhosts/reeves.one/own
cloud.reeves.one/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: /lib/x86_64-linux-gnu/l
ibc.so.6: version `GLIBC_2.34' not found (required by /var/www/vhosts/reeves.one/owncloud.reeves.one/apps/mail/vendor/ch
ristophwurst/kitinerary-bin/src/../bin/kitinerary-extractor)"
[14-Jan-2023 12:20:21] WARNING: [pool owncloud.reeves.one] child 19025 said into stderr: "/var/www/vhosts/reeves.one/own
cloud.reeves.one/apps/mail/vendor/christophwurst/kitinerary-bin/src/../bin/kitinerary-extractor: /lib/x86_64-linux-gnu/l
[...]

PR #9291, opened 2024-02-03
SML Markup rendering

jaudriga:
Support mails containing JSON-LD with Schema.org data.

This adds initial support for the upcoming SML standard (see [SML IETF WG](https://datatracker.ietf.org/wg/sml) and https://structured.email ) by implementing:

* schema extraction from HTML body of a mail via https://github.com/audriga/html2jsonld-php
* generic schema rendering via https://github.com/audriga/jsonld2html-cards

**Note on schema extraction library:** audriga/html2jsonld-php is NOT supposed to replace KItinerary. KItinerary is clearly the more powerful solution. However, KItinerary cannot be used in some Nextcloud installations (e.g. #8322 or #7899 ) . audriga/html2jsonld-php is supposed to be a light-weight alternative for some usage scenarios. The idea is to use html2jsonld-php only if configured.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
 		"php": ">=8.0 <=8.3",
 		"ext-openssl": "*",
 		"arthurhoaro/favicon": "^2.0.0",
-		"audriga/html2jsonld-php": "dev-master as 1.x-dev",
+		"audriga/html2jsonld-php": ">=1.0.0",
 		"bamarni/composer-bin-plugin": "^1.8.2",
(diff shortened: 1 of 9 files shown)


jaudriga:
Closing this for now, because we have opened this a bit too prematurely. Sorry for that.

In its current form this PR will actually make JSON-LD rendering worse. For example, FlightReservations have yet to get proper support in LD2H. We would like to push few more improvements in the next days - probably wait with further consideration. We may also open two distinct PRs.

Issue #9162, opened 2023-12-12
v3.5+ DB performance analysis


ChristophWurst:
### Steps to reproduce

1. Use the app

### Expected behavior

Fast DB operations

### Actual behavior

?

### Mail app version

v3.5.0 RC3 / RC4

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

Queries of my personal instance for roughly a day of use with two mail accounts:

```
 133752 UPDATE `oc_mail_messages` SET `flag_answered` = :flag_answered, `flag_deleted` = :flag_deleted, `flag_draft` = :flag_draft, `flag_flagged` = :flag_flagged, `flag_seen` = :flag_seen, `flag_forwarded` = :flag_forwarded, `flag_junk` = :flag_junk, `flag_notjunk` = :flag_notjunk, `flag_mdnsent` = :flag_mdnsent, `flag_important` = :flag_important, `updated_at` = :dcValue1 WHERE (`uid` = :uid) AND (`mailbox_id` = :mailbox_id)
   3774 UPDATE `oc_mail_mailboxes` SET `sync_new_lock` = :dcValue1, `sync_changed_lock` = :dcValue2, `sync_vanished_lock` = :dcValue3 WHERE `id` = :dcValue4
   1811 SELECT DISTINCT `t`.*, `mt`.`imap_message_id` FROM `oc_mail_tags` `t` INNER JOIN `oc_mail_message_tags` `mt` ON `t`.`id` = `mt`.`tag_id` WHERE (`mt`.`imap_message_id` IN (:ids)) AND (`t`.`user_id` = :dcValue1)
   1126 SELECT `uid` FROM `oc_mail_messages` WHERE (`mailbox_id` = :dcValue1) AND (`id` IN (:ids))
   1018 SELECT `label`, `email`, `type`, `message_id` FROM `oc_mail_recipients` WHERE `message_id` IN (:ids)
    909 SELECT * FROM `oc_mail_accounts` WHERE (`user_id` = :dcValue1) AND (`id` = :dcValue2)
    883 SELECT `mb`.* FROM `oc_mail_mailboxes` `mb` INNER JOIN `oc_mail_accounts` `a` ON `mb`.`account_id` = `a`.`id` WHERE (`a`.`user_id` = :dcValue1) AND (`mb`.`id` = :dcValue2)
[...]

PR #9295, opened 2024-02-05
perf(db): optimise indices of mail tables

JohannesGGE:
Fix: #9162 

Needs https://github.com/nextcloud/server/pull/43209 to also replace indices.

What should happen:

## oc_mail_messages

- add index on `structure_analyzed` -> `mail_messages_strucanalyz_idx`
- drop index `mail_messages_mailbox_id` because `mail_messages_mb_id_uid` already contains `mailbox_id` on first position
- make `mail_messages_mb_id_uid` unique -> `mail_messages_mb_id_uid_uidx`

## oc_mail_classifiers

- add index on `created_at` -> `mail_class_creat_idx`

## oc_mail_accounts

- add index on `provisioning_id` -> `mail_acc_prov_idx`

## oc_mail_smime_certificates

- extend `mail_smime_certs_uid_idx` to cover `(user_id,email_address)` -> `mail_smime_certs_uid_email_idx`
- drop `mail_smime_certs_id_uid_idx`

## oc_mail_tags

- drop `mail_msg_tags_usr_id_index` because `mail_msg_tags_usr_lbl_idx` already contains `user_id` on first position

## oc_mail_aliases

- add index on `account_id` -> `mail_alias_accid_idx`

## oc_mail_trusted_senders

- add index on `(user_id,email,type)` and drop `mail_trusted_senders_type` -> `mail_trusted_senders_idx`

## oc_mail_coll_addresses

- add index on `(user_id, email, display_name)` and drop `mail_coll_addr_userid_index` and `mail_coll_addr_email_index` -> `mail_coll_idx`

### DIFF
--- a/lib/Listener/OptionalIndicesListener.php
+++ b/lib/Listener/OptionalIndicesListener.php
@@ -57,6 +57,64 @@ class OptionalIndicesListener implements IEventListener {
 				],
 			);
 		}
+
+		$event->addMissingIndex(
+			'mail_messages',
+			'mail_messages_strucanalyz_idx',
+			['structure_analyzed']
+		);
+
+		$event->addMissingIndex(
+			'mail_classifiers',
+			'mail_class_creat_idx',
+			['created_at']
+		);
+
+		$event->addMissingIndex(
+			'mail_accounts',
+			'mail_acc_prov_idx',
+			['provisioning_id']
+		);
+
+		$event->addMissingIndex(
+			'mail_aliases',
+			'mail_alias_accid_idx',
+			['account_id']
+		);
+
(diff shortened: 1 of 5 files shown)


miaulalala (approved):

JohannesGGE:
I commented out the index creation of the indices to drop.

It would be possible to also comment out the creation of the indices that are replaced and add the new ones in the migration, to have new instances up to date without "occ missingindices". The Migration will have a lot of checks in this case, which can be confusing. What do you think?

st3iny (approved):
Tested and works. I'd suggest to keep the new indices out of old migrations to keep it clean.

ChristophWurst (approved):

PR #9323, opened 2024-02-12
enh: change reply button placement

hamza221:
Default reply button is now on the right 
image


src/components/Message.vue
```diff
@@ -73,6 +66,13 @@
 				</template>
 				{{ reply }}
 			</NcButton>
+			<NcButton type="primary"
+				@click="onReply">
+				<template #icon>
+					<ReplyIcon />
+				</template>
+				Reply
```
ChristophWurst, line 74: not exactly related but doesn't this button need translation?

ChristophWurst (approved):

PR #9324, opened 2024-02-13
fix(setup): Sort MX records by weight

ChristophWurst:


st3iny (approved):
Code looks good.

https://3v4l.org/8EMTs

lib/Service/AutoConfig/MxRecord.php
```diff
@@ -38,10 +40,15 @@ public function __construct(LoggerInterface $logger) {
...
+		if (getmxrr($host, $mxRecords, $mxWeights) === false) {
 			$this->logger->debug("no MX records for host <$host> found");
 			return [];
 		}
+
+		// Sort MX records by weight
+		$sortedRecords = array_combine($mxRecords, $mxWeights);
+		asort($sortedRecords, SORT_NUMERIC);
+
 		$mxRecords = array_filter($mxRecords, static function ($record) {
```
ChristophWurst, line 52: forgot to use the sorted array here :laughing:

PR #9326, opened 2024-02-13
Fix calendar picker for create tasks

GretaD:


src/components/TaskModal.vue
```diff
@@ -130,6 +132,13 @@ export default {
 		}
 	},
 	methods: {
+		/**
+		 * @param {string} id The calendar id
+		 * @return {object|undefined} The calendar object (if it exists)
+		 */
+		getCalendarById(id) {
+			return this.calendars.find((cal) => cal.id === id)
+		},
```
GretaD, line 141: @st3iny the initial error is gone, but now _Invalid prop: type check failed for prop "displayname". Expected String, got Undefined_  same for color, is shown, i think its because of this. Am i doing this right?

src/components/TaskModal.vue
```diff
@@ -115,7 +117,7 @@ export default {
 			return this.$store.getters.getAllTags
 		},
 		calendars() {
-			return this.$store.getters.getTaskCalendarsForCurrentUser
+			return this.$store.getters.getTaskCalendarsForCurrentUser.map(calendar => ({ id: calendar.id }))
```
ChristophWurst, line 120: inspect with the Vue dev tools. you will get an array of objects that only have an `id` property, nothing else. you might need to map color and displayname too

GretaD:
[Screencast from 14.02.2024 14:45:13.webm](https://github.com/nextcloud/mail/assets/12728974/36f777a1-c790-4862-9425-e4abeab673d9)

ChristophWurst (approved):

PR #9327, opened 2024-02-13
fix(setup): Use MX record TLD for ISPDB lookup

ChristophWurst:
1. MX points to server for outgoing email
2. TLD of MX points to the hoster
3. ISPDB of the hoster gives the IMAP/SMTP config
4. If ISPDB fails, try MX as IMAP/SMTP server

~~This is best enjoyed with https://github.com/nextcloud/mail/pull/9328~~ rebased


miaulalala (approved):
image

Issue #8435, opened 2023-05-02
Mail both "Sent" and stuck in Outbox


jancborchardt:
### Steps to reproduce

1. Send a mail
2. Wait for it to leave outbox

### Expected behavior

The mail is sent.

### Actual behavior

Recently I often have the issue that mails are both sent _and_ still stuck in the outbox. This creates a situation where I'm not sure if the mail is sent, or I send it twice to make sure.

E.g. right now I have a situation where I know the mail was sent as one of the recipients is a forwarder I am in as well, but still in the Outbox it says "Message could not be sent", even after a refresh.
For debugging on our mail server, mail was sent at 12:38

### Mail app version

Current on cloud.nextcloud.com

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
Nothing in the logs, nothing on Sentry.

juliushaertl:
@jancborchardt Any reason to add this and others to the office team board? 😕

gutiersa:
I am having this problem, but my messages are not sent at all. It occurred two updates ago. It was sending fine before that.

ChristophWurst:
That's different @gutiersa. Your messages can probably not be sent. Capture the Nextcloud log and look for errors when you click *Send now* on one of the messages. File that as a new ticket please.

gutiersa:
Ok, thank you.

jancborchardt:
@ChristophWurst ran into this again just now (literally this past minute), here are the console logs:
```
jQuery is deprecated: The global jQuery is deprecated. It will be removed in a later versions without another warning. Please ship your own. 296 [sentry.js:2:62947](https://cloud.nextcloud.com/apps/sentry/js/sentry.js?v=0c566cac-1762)
[ERROR] mail: Failed to send message 4173 from outbox 
Object { app: "mail", uid: "jan", level: 2, error: {…} }
[sentry.js:2:62947](https://cloud.nextcloud.com/apps/sentry/js/sentry.js?v=0c566cac-1762)
[ERROR] mail: Could not delay-send message 4173 
Object { app: "mail", uid: "jan", level: 2, message: {…} }
[sentry.js:2:62947](https://cloud.nextcloud.com/apps/sentry/js/sentry.js?v=0c566cac-1762)
Uncaught (in promise) 
[...]

PR #9364, opened 2024-02-21
fix(outbox): catch failed SENT copy operation

miaulalala:
Fixes https://github.com/nextcloud/mail/issues/8435

Sending a mail could fail at any point in the send operation.

This PR aims to introduce clear status codes that allow for the send code to redo the operations that fail.

Everything except the actual IMAP send and copying the sent message to the SENT mailbox are idempotent.


lib/Db/LocalMessage.php
```diff
@@ -60,11 +60,29 @@
...
+ * @method bool isMdnRequested()
+ * @method setMdnRequested(bool $mdnRequested)
+ * @method string|null getRaw()
+ * @method setRaw(string|null $raw)
  */
 class LocalMessage extends Entity implements JsonSerializable {
 	public const TYPE_OUTGOING = 0;
 	public const TYPE_DRAFT = 1;
 
+	public const STATUS_UNPROCESSED = 0;
```
ChristophWurst, line 74:
```suggestion
	public const STATUS_RAW = 0;
```

/s

ChristophWurst:
Unassigning from the milestone because this is not a must have. We can merge at any time and backport.

lib/Db/LocalMessage.php
```diff
@@ -116,6 +133,15 @@ class LocalMessage extends Entity implements JsonSerializable {
 	/** @var bool|null */
 	protected $smimeEncrypt;
 
+	/**
+	 * @var int
+	 * @psalm-var self::STATUS_*
```
ChristophWurst, line 138: https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#int-mask1-2-4

lib/Db/LocalMessage.php
```diff
@@ -60,11 +60,28 @@
  * @method setSmimeCertificateId(?int $smimeCertificateId)
  * @method bool|null getSmimeEncrypt()
  * @method setSmimeEncrypt (bool $smimeEncryt)
+ * @method int getStatus();
```
ChristophWurst, line 63:
```suggestion
 * @method int|null getStatus();
```

the db column is nullable

lib/Service/AntiAbuseService.php
```diff
@@ -85,11 +86,11 @@ private function checkNumberOfRecipients(IUser $user,
...
 		}
 
-		$actualNumberOfRecipients = count($messageData->getTo())
-			+ count($messageData->getCc())
-			+ count($messageData->getBcc());
+		$actualNumberOfRecipients = count($message->getRecipients());
 
 		if ($actualNumberOfRecipients >= $numberOfRecipientsThreshold) {
+			$message->setStatus(LocalMessage::STATUS_TOO_MANY_RECIPIENTS);
+			$this->eventDispatcher->dispatchTyped(new OutboxMessageStatusChangeEvent($message));
```
ChristophWurst, line 93: that is a lot of indirection. why is it necessary to dispatch an event here for a database update?

lib/Controller/OutboxController.php
```diff
@@ -209,6 +209,9 @@ public function update(
 		?int $sendAt = null
 	): JsonResponse {
 		$message = $this->service->getMessage($id, $this->userId);
+		if ($message->getStatus() === LocalMessage::STATUS_IMAP_SENT_MAILBOX_FAIL || $message->getStatus() === LocalMessage::STATUS_PROCESSED) {
+			return JsonResponse::error('Cannot modify already sent message', Http::STATUS_FORBIDDEN, [$message]);
```
ChristophWurst, line 213:
```suggestion
			return JsonResponse::error('Cannot modify already sent message', Http::STATUS_FORBIDDEN, $message);
```

lib/Controller/OutboxController.php
```diff
@@ -217,11 +220,13 @@ public function update(
...
 		$message->setEditorBody($editorBody);
 		$message->setHtml($isHtml);
-		$message->setFailed($failed);
 		$message->setInReplyToMessageId($inReplyToMessageId);
 		$message->setSendAt($sendAt);
 		$message->setSmimeSign($smimeSign);
 		$message->setSmimeEncrypt($smimeEncrypt);
+		// Explicitly overwrite the status, so we can try sending again
+		// in case the user has updated the failing component
+		$message->setStatus(LocalMessage::STATUS_RAW);
```
ChristophWurst, line 229: this is business logic. move it into the service.

lib/Listener/OutboxStatusChangeListener.php
```diff
@@ -0,0 +1,47 @@
...
+	public function handle(Event $event): void {
+		if (!($event instanceof OutboxMessageStatusChangeEvent)) {
+			return;
+		}
+
+		$message = $event->getMessage();
+		$this->messageMapper->update($message);
+
+	}
+}
```
ChristophWurst, line 47: this does not feel like a use case for events. can you inline the mapper call and make persistence explicit?

lib/Send/AntiAbuseHandler.php
```diff
@@ -0,0 +1,63 @@
...
+			return $this->processNext($account, $localMessage);
+		}
+
+		$user = $this->userManager->get($account->getUserId());
+		if ($user === null) {
+			$this->logger->error('User {user} for mail account {id} does not exist', [
+				'user' => $account->getUserId(),
+				'id' => $account->getId(),
+			]);
+			// What to do here?
```
ChristophWurst, line 50: it could only happen if a message of a deleted user is sent -> abort the chain

lib/Send/Chain.php
```diff
@@ -0,0 +1,53 @@
...
+	public function process(Account $account, LocalMessage $localMessage): void {
+		$chain = $this->sentMailboxHandler;
+		$chain->setNext($this->antiAbuseHandler)
+			->setNext($this->sendHandler)
+			->setNext($this->copySentMessageHandler)
+			->setNext($this->flagRepliedMessageHandler)
+			->setNext($this->removeMessageHandler);
+
+		$result = $chain->process($account, $localMessage);
+		if($result->getStatus() !== LocalMessage::STATUS_PROCESSED) {
```
ChristophWurst, line 49:
```suggestion
		if ($result->getStatus() !== LocalMessage::STATUS_PROCESSED) {
```

not sure why cs fixer does not complain but there should be a space

lib/Send/RemoveMessageHandler.php
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	public function process(Account $account, LocalMessage $localMessage): LocalMessage {
+		if($localMessage->getStatus() === LocalMessage::STATUS_PROCESSED) {
+			$this->attachmentService->deleteLocalMessageAttachments($account->getUserId(), $localMessage->getId());
+			$this->messageMapper->deleteWithRecipients($localMessage);
+		}
+		return $localMessage;
+	}
+}
```
ChristophWurst, line 42: I would handle update/delete of the message outside of the chain. if the chain exits early save the message. if the chain exits cleanly delete the message.

ChristophWurst (commented):
Looks sane but haven't tested yet

Tests missing for all the new handlers

lib/Listener/OutboxStatusChangeListener.php
miaulalala: yeah that event can go, it's not needed with the handlers any more.

lib/Send/Chain.php
miaulalala: yeah I run the fixer pretty regularely but it never fixes those :cry:

lib/Send/RemoveMessageHandler.php
miaulalala: good point, that makes sense.

lib/Controller/OutboxController.php
miaulalala: I don't think that's right?

https://github.com/nextcloud/mail/blob/3285422eef37b9716cb31aeb66b76dcacd5ef3b2/lib/Controller/OutboxController.php#L150

miaulalala: `error::$data` ist typed to `[]` unlike `success` which has no enforced type

lib/Send/CopySentMessageHandler.php
```diff
@@ -0,0 +1,94 @@
...
+class CopySentMessageHandler extends AHandler {
+
+
+	public function __construct(private IMAPClientFactory $imapClientFactory,
+		private MailboxMapper $mailboxMapper,
+		private LoggerInterface $logger,
+		private MessageMapper $messageMapper
+	) {
+		parent::__construct();
+	}
```
ChristophWurst, line 43:
```suggestion

	public function __construct(private IMAPClientFactory $imapClientFactory,
		private MailboxMapper $mailboxMapper,
		private LoggerInterface $logger,
		private MessageMapper $messageMapper
	)  {
		parent::__construct();
	}

```

lib/Send/AHandler.php
```diff
@@ -0,0 +1,47 @@
...
+
+use OCA\Mail\Account;
+use OCA\Mail\Db\LocalMessage;
+
+abstract class AHandler {
+
+	protected AHandler|null $next = null;
+
+	public function __construct() {
+	}
```
ChristophWurst, line 32: unused -> drop. an empty constructor only makes sense for public API and where we have to avoid the fragile baseclass problem. we can always adjust all sub classes if a constructor is needed here in the future.

lib/Send/AntiAbuseHandler.php
```diff
@@ -0,0 +1,64 @@
...
+		parent::__construct();
+	}
+	public function process(Account $account, LocalMessage $localMessage): LocalMessage {
+		if($localMessage->getStatus() === LocalMessage::STATUS_IMAP_SENT_MAILBOX_FAIL
+			|| $localMessage->getStatus() === LocalMessage::STATUS_PROCESSED) {
+			return $this->processNext($account, $localMessage);
+		}
+
+		$user = $this->userManager->get($account->getUserId());
+		if ($user === null) {
```
ChristophWurst, line 45:
```suggestion
		if ($localMessage->getStatus() === LocalMessage::STATUS_IMAP_SENT_MAILBOX_FAIL
			|| $localMessage->getStatus() === LocalMessage::STATUS_PROCESSED) {
			return $this->processNext($account, $localMessage);
		}

		$user = $this->userManager->get($account->getUserId());
		if ($user === null) {
```

ChristophWurst (changes requested):
Testing the case where CopySentMessageHandler fails:`` An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'raw' in 'field list'`` resulting in a HTTP500.

`\OCP\AppFramework\Db\QBMapper::update` is called with a `LocalMessage`

ChristophWurst:
``\OCA\Mail\Controller\OutboxController::send`` must not return a HTTP200 when the chain fails. Right now the outbox message disappears from the outbox until a page reload, then the message is back.

miaulalala:
> Testing the case where CopySentMessageHandler fails:` An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'raw' in 'field list'` resulting in a HTTP500.
> 
> `\OCP\AppFramework\Db\QBMapper::update` is called with a `LocalMessage`

Did you run the migration again? The "raw" field was introduced later.

ChristophWurst:
Raw is a column for local messages, not messages

ChristophWurst (changes requested):
Successful case works :heavy_check_mark: 
An error in `\OCA\Mail\Send\CopySentMessageHandler::process` shows an error toast in the front-end :heavy_check_mark: 
After the error the message vanishes from the outbox. From a user perspective it's simply *gone* :x: 

Messages that failed to copy on IMAP but were sent should show in a transparent matter. As in, they stay in the outbox, visible, but read-only.

ChristophWurst:
- [ ] BUG: Message that failed to copy to IMAP shows in inbox with no further information. The message is clickable, but clicking results in a silent error. Console:

image

- [ ] BUG: *Send now* does not work. I would expect to be able to manually retry/resume the sending process if it failed.

ChristophWurst (commented):
Failed messages appear in the outbox :heavy_check_mark: 
Retried messages are processed by the worker job and finally disappear from the outbox :heavy_check_mark: 

Frontend handling can be improved ^

ChristophWurst:
Clicking messages stuck in the outbox no longer causes an error :heavy_check_mark: 

But the UX is strange. The message appears *clickable* but nothing happens on click. The cursor should be the default one, not *pointer*.

With the latest UI changes the ability to force resend/retry was lost. It's only possible to delete stuck outbox messages or wait for the cron job to retry/resend.

miaulalala:
> Clicking messages stuck in the outbox no longer causes an error ✔️
> 
> But the UX is strange. The message appears _clickable_ but nothing happens on click. The cursor should be the default one, not _pointer_.
> 
> With the latest UI changes the ability to force resend/retry was lost. It's only possible to delete stuck outbox messages or wait for the cron job to retry/resend.

We can use an NcListItemIcon instead of the NcListItem that does the rendering for the pointer,. Since NcListItem it also does the rendering for the status message, it's kinda weird though. 

There's no prop that I can see that will allow the pointer to be set for the NcListItem component. The styles don't really work well together for the NcListItem and NcListItemIcon (first one is NcListItemIcon, second and third are NcListItem):

image

miaulalala:
On further thinking I'd actually like to see the right panel being used. We could treat outbox messages like "Sent" messages and display them like a regular email, with an "Edit" button for anything that's not been sent yet. Would get rid of the awkward unused message content space and allow for a preview of any sent messages. cc @nimishavijay 

For now, I'd leave the NcListItem even though it's a tiny bit awkward with the non- action on click unless a frontender wants to take a real shot at refactoring the whole thing to render better than my awkward try with the NcListItemIcon.

ChristophWurst:
Then we leave the cursor pointer or use :v-deep to overwrite it

Once force send is back, conflicts are resolved and CI is green we can merge this :shipit:

miaulalala:
> Then we leave the cursor pointer or use :v-deep to overwrite it
> 
> Once force send is back, conflicts are resolved and CI is green we can merge this :shipit:

force send is a different PR, no?

ChristophWurst:
> With the latest UI changes the ability to force resend/retry was lost. It's only possible to delete stuck outbox messages or wait for the cron job to retry/resend.

This is what I mean by "force" resend. When storing on IMAP fails temporarily I want to be able to retry.

ChristophWurst:
- [ ] Unhandled js errors when simulating the error case of https://github.com/nextcloud/mail/issues/8435

image

ChristophWurst:
When I retry and it still fails

image

ChristophWurst:
Related tests fail

ChristophWurst (commented):
Tested & works

src/components/NewMessageModal.vue
```diff
@@ -392,7 +392,9 @@ export default {
 
 				if (!data.sendAt || data.sendAt < Math.floor((now + UNDO_DELAY) / 1000)) {
 					// Awaiting here would keep the modal open for a long time and thus block the user
-					this.$store.dispatch('outbox/sendMessageWithUndo', { id: dataForServer.id })
+					this.$store.dispatch('outbox/sendMessageWithUndo', { id: dataForServer.id }).catch((error) => {
+						logger.debug('Could not send message', { error })
+					})
```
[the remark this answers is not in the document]
miaulalala: Because there is nothing to do about the error - sendMessageWithUndo handles the rejection of the promise if it fails. The error handling is to avoid the console errors for an uncaught in Promise exception from the rejection.

src/components/OutboxMessageListItem.vue
```diff
@@ -139,9 +171,21 @@ export default {
...
 			await this.$store.dispatch('outbox/updateMessage', { message, id: message.id })
-			await this.$store.dispatch('outbox/sendMessageWithUndo', { id: message.id })
+			try {
+				if (this.message.status !== 11) {
+					await this.$store.dispatch('outbox/sendMessageWithUndo', { id: message.id })
+				} else {
+					await this.$store.dispatch('outbox/copyMessageToSentMailbox', { id: message.id })
+				}
+			} catch (error) {
+				logger.error('Could not send or copy message', { error })
```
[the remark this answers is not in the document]
miaulalala: We could check if the error has data, and not update the message otherwise?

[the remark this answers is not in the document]
miaulalala: Already added

src/components/NewMessageModal.vue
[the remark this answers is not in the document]
miaulalala: Ah, and we can't try- catch because we can't await the store.dispatch. looked it through with richard and if we did, we would block the user from interactions until the 30 seconds of sending are done. So we have to do it the "ugly" way with a .catch on the promise

PR #9415, opened 2024-02-28
chore(deps): Drop vue-slide-up-down

ChristophWurst:
Added in https://github.com/nextcloud/mail/pull/1155 but was never used. Should have asked for reviews.


GretaD (commented):
sorry, i approve this anyway :)

PR #9423, opened 2024-02-29
fix(db): Avoid dirty reads while deleting messages by uid

ChristophWurst:
1. Instead of SELECT all data into memory and run a second DELETE query, use a sub query.
2. Instead of looking up IDs for UIDs and deleting by ID, delete by UID directly.

This method was prone to dirty reads before. Now there are only write queries.

Ref https://github.com/nextcloud/groupware/issues/24

## How to test

1. Set up an account in Mail and a second client
2. Delete messages in the other client
3. Sync Mail

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -705,39 +705,32 @@ class MessageMapper extends QBMapper {
 	}
 
 	public function deleteByUid(Mailbox $mailbox, int ...$uids): void {
-		$messageIdQuery = $this->db->getQueryBuilder();
+		$messageIdSubQuery = $this->db->getQueryBuilder();
 		$deleteRecipientsQuery = $this->db->getQueryBuilder();
 		$deleteMessagesQuery = $this->db->getQueryBuilder();
 
-		// Get all message ids query
-		$messageIdQuery->select('id')
+		$messageIdSubQuery->select('id')
 			->from($this->getTableName())
 			->where(
-				$messageIdQuery->expr()->eq('mailbox_id', $messageIdQuery->createNamedParameter($mailbox->getId())),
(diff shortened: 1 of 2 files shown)


lib/Db/MessageMapper.php
```diff
@@ -705,39 +705,32 @@ public function deleteAll(Mailbox $mailbox): void {
...
 		$deleteMessagesQuery->delete($this->getTableName())
-			->where($deleteMessagesQuery->expr()->in('id', $deleteMessagesQuery->createParameter('messageIds')));
+			->where(
+				$deleteMessagesQuery->expr()->eq('mailbox_id', $deleteMessagesQuery->createNamedParameter($mailbox->getId())),
+				$deleteMessagesQuery->expr()->in('uid', $deleteMessagesQuery->createParameter('uids')),
+			);
 
 		foreach (array_chunk($uids, 1000) as $chunk) {
-			$messageIdQuery->setParameter('uids', $chunk, IQueryBuilder::PARAM_INT_ARRAY);
-			$cursor = $messageIdQuery->executeQuery();
```
ChristophWurst, line 728: The original implementation of deleteByUid did use a sub query too: https://github.com/nextcloud/mail/pull/4663. Then https://github.com/nextcloud/mail/pull/4947 added an intermediate fetch. @miaulalala do you remember why this was done?

miaulalala: Most likely because I ran into the usual subquery issues (where to put the brackets) and didn't know how to fix it back then.

ChristophWurst: Good. Then this change is safe :)

st3iny (approved):
Tested and works. Logs were clean.

PR #9434, opened 2024-03-05
Fix: Thread summaries fail for threads across multiple mailboxes

### LINKED ISSUE #9367: Thread summaries fail for threads across multiple mailboxes
### Steps to reproduce

0. Install integration_openai
1. Receive an email
2. Respond to that email
3. Open the thread

### Expected behavior

Thread summaries can be generated

### Actual behavior

I get an error with "Message not found". The reason for that is that `\OCA\Mail\Service\AiIntegrations\AiIntegrationsService::summarizeThread` uses `\OCA\Mail\Contracts\IMailManager::getImapMessage` with a specific mailbox. Only one of the two messages is in there.

### Mail app version

3.5

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

cc @hamza221

hamza221:
Fix #9367

### DIFF
--- a/lib/Controller/ThreadController.php
+++ b/lib/Controller/ThreadController.php
@@ -168,7 +168,6 @@ class ThreadController extends Controller {
 		try {
 			$summary = $this->aiIntergrationsService->summarizeThread(
 				$account,
-				$mailbox,
 				$message->getThreadRootId(),
 				$thread,
 				$this->currentUserId,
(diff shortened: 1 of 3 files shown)


ChristophWurst (approved):
We'll need the same fix for the event data generation because I copy pasted your code ;)

PR #9435, opened 2024-03-05
Fix: Add smart reply to composer

hamza221:


ChristophWurst (approved):
That must have been an unfortunate timing with your previous PR and https://github.com/nextcloud/mail/pull/9211

PR #9439, opened 2024-03-06
Enh: Improve smart reply UI

hamza221:
| B | A |
|--------|--------|
| image | image |

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -56,22 +56,22 @@
 			:message="message" />
 		<MessageAttachments :attachments="message.attachments" :envelope="envelope" />
 		<div id="reply-composer" />
-		<div v-if="smartReplies.length>0" class="reply-buttons">
-			<NcButton v-for="(reply,index) in smartReplies"
-				:key="index"
-				type="secondary"
-				@click="onReply(reply)">
-				<template #icon>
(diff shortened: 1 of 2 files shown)


lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -178,7 +178,7 @@ public function getSmartReply(Account $account, Mailbox $mailbox, Message $messa
 			throw new ServiceException('Text processing is not available in your current Nextcloud version', 0, $e);
 		}
 		if (in_array(FreePromptTaskType::class, $manager->getAvailableTaskTypes(), true)) {
-			$cachedReplies = $this->cache->getValue('smartReplies_'.$message->getId());
+			$cachedReplies = $this->cache->getValue('smartReplies_'.$message->getUid());
```
ChristophWurst, line 181: how is this related to the UI enhancements? UIDs are not globally unique. Wouldn't you have a conflict for two identical UIDs in two mailboxes?

hamza221: Git add . 🥲
Was not supposed to be in the pr but I noticed that caching was broken and forgot to discard it.

Issue #5181, opened 2021-06-15
iCloud messages indexed but not displayed


Minnietheandroid:
Hi.
Trying out the app (it looks fantastic btw) and came across this error trying to connect an iCloud mailbox.

This or a similar issue came up a few days back and was closed for various reasons. #3401
Wasn't sure if I should continue it or not, but seeing as I'm on NC Snap, with Apache thought I'd do a fresh report.
Apologies if that was the wrong decision. Nextcloud error attached and any advice would be most appreciated.

[[PHP] Error: Error: A non-numeric value encountered.txt](https://github.com/nextcloud/mail/files/6652471/PHP.Error.Error.A.non-numeric.value.encountered.txt)

### Expected behavior
Mailbox displays messages in right hand pane

### Actual behavior
Mailbox reports correct number of messages/unread items in mailbox but doesn't display the messages or content

### Mail app

**Mail app version:** 1.9.5

**Mailserver or service:** iCloud

### Server configuration

**Operating system**: Ubuntu 20.04 - Nextcloud Snap

**Web server:** Apache 2.4 
 
**Database:** MySQL 8

**PHP version:** 7.4

**Nextcloud Version:** Snap 21.01

#### Client configuration
**Browser:** Firefox 89.0, Brave 1.25.68

**Operating system:** Manjaro 21

ChristophWurst:
> A non-numeric value encountered

This one still puzzles me. Could you try with detailed logging as documented at https://github.com/nextcloud/mail/blob/master/doc/admin.md#logging? I'd like to see the IMAP communication where an UID isn't a number.

Minnietheandroid:
Thank you for the reply Christoph.

I was able to retrieve the nextcloud.log and the horde_imap.log but couldn't find the smtp one, 
perhaps because I haven't sent an email out yet? I don't know.

Personal info replaced with USERNAME and LOCALIP.

Thank you for looking.

[nextcloud.log](https://github.com/nextcloud/mail/files/6654635/nextcloud.log)
[horde_imap.log](https://github.com/nextcloud/mail/files/6654636/horde_imap.log)

Minnietheandroid:
Sorry, wasn't thinking clearly.
I sent an email and got the smtp log you need.

[horde_smtp.log](https://github.com/nextcloud/mail/files/6658458/horde_smtp.log)

Minnietheandroid:
Hi and thanks very much @miaulalala for having a look.
[...]

PR #9454, opened 2024-03-08
fix: display iCloud messages

miaulalala:
Resolves: https://github.com/nextcloud/mail/issues/5181

Since iCloud doesn't support ESEARCH ([RFC 4731](https://www.rfc-editor.org/rfc/rfc4731)) properly, we need to add an ALL fetch query to get the MIN and MAX uids.

It's definitely gonna be a bit slower, but that's for the initial sync only.

Thanks to https://github.com/ktemkin who also debugged this and put me on the right track.


Luncheon3462:
Can’t wait for this to get fixed.

Luncheon3462:
#9454

miaulalala:
Needs https://github.com/bytestream/Imap_Client/pull/15

lib/IMAP/MessageMapper.php
```diff
@@ -146,6 +146,17 @@ public function findAll(Horde_Imap_Client_Socket $client,
 			];
 		}
 
+		// This can happen for iCloud
+		if($metaResults['max'] === null) {
```
ChristophWurst, line 150:
```suggestion
		if ($metaResults['max'] === null) {
```

ChristophWurst (approved):
Tested and works with Dovecot

romcheg:
@miaulalala Thank you for this one!

strich3:
@miaulalala Thank you!

Luncheon3462:
@miaulalala does this mean that the fix for iCloud emails will be in the 3.6.0 version of the mail app?

Luncheon3462:
@miaulalala ???

Luncheon3462:
je regret.  merci tres beaucoup.

Issue #8309, opened 2023-03-28
Allow stacking message list and reading pane vertically


goulvench:
### Is your feature request related to a problem? Please describe.

On desktop, the default view is composed of 3 columns: mailbox list, message list, and reading pane.
On tablet/mobile, the view switches between message list and reading pane.

On narrow desktop screens (13"), especially when the accounts pane is open, the message list and reading pane are very narrow and hard to read/use.

### Describe the solution you'd like

Moving to a 2-column design, with the reading pane under the message list, would make better use of vertical available space, and match user expectations (most mail clients offer to have the reading pane either to the side or under the message list). 

The message list preview would have much more breathing room, making it easier to triage messages without opening them for instance.

### Describe alternatives you've considered

- Animation showing how to switch between right pane, stacked panes, and no pane on Gmail https://www.googleworkspaceguides.com/wp-content/uploads/2022/02/Animation.gif
- Reading pane toggler for Outlook https://support.content.office.net/en-us/media/7be78ac1-d931-4ee9-865e-3f5d150290a2.png

### Additional context

This solution would be a solution to #3294, and help users keep track of context (when comparing two different conversations for instance).

GretaD:
@nextcloud/designers I would like to start working on this asap, please add it for review on your backlog and please send me the mockups/insights when you have them :)

jancborchardt:
@GretaD @nimishavijay the way Gmail does it seems best here:
- A layout toggle between: "No split", "Side by side" and "Top and bottom" (or different wording if you can think of some :) Default being "Side by side"
- The side by side view has the resizable message list, so it is necessary for top and bottom layout as well

@GretaD do you need an additional mockup besides that and https://www.googleworkspaceguides.com/wp-content/uploads/2022/02/Animation.gif ?

GretaD:
> @GretaD do you need an additional mockup besides that and https://www.googleworkspaceguides.com/wp-content/uploads/2022/02/Animation.gif ?

So having only one toggle that switch between full view and 3 column design(the one we have now).
I am not sure where to add the toggle, so just that would be my question for now.

marcoambrosini:
I think we should not implement this view. Both the messages list and the email content are vertical scroll containers, so it doesn't make much sense to stack them vertically IMO.

ChristophWurst:
It won't be the default. It does make sense to some people, like me.

jancborchardt:
> So having only one toggle that switch between full view and 3 column design(the one we have now).
[...]

PR #9462, opened 2024-03-13
Add horizontal split for mailing list

GretaD:
- [x] Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5401
- [ ] Documentation at user_manual

Tasks

- [x] Add horizontal split

Follow up
- - [ ] Increase the number of messages showed by default on vertical and list view, and less on horizontal view

How to reproduce:
1. Go to app settings
2. Click horizontal view 

For https://github.com/nextcloud/mail/issues/8309

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -4,14 +4,14 @@
 			:name="t('mail', 'Mail settings')"
 			:show-navigation="true"
 			:open.sync="showSettings">
-			<NcAppSettingsSection id="mail-list-view" :name="t('mail', 'Layout')">
+			<NcAppSettingsSection v-if="!isMobile" id="mail-list-view" :name="t('mail', 'Layout')">
 				<NcCheckboxRadioSwitch value="no-split"
 					:button-variant="true"
 					name="mail-layout"
 					type="radio"
 					:checked="layoutMode"
 					button-variant-grouped="vertical"
-					@update:checked="setLayout">
+					@update:checked="setLayout('no-split')">
 					<template #icon>
 						<CompactMode :size="20" />
 					</template>
@@ -23,11 +23,23 @@
 					type="radio"
 					:checked="layoutMode"
 					button-variant-grouped="vertical"
-					@update:checked="setLayout">
+					@update:checked="setLayout('vertical-split')">
 					<template #icon>
 						<VerticalSplit :size="20" />
 					</template>
 					{{ t('mail', 'Vertical split') }}
+				</NcCheckboxRadioSwitch>
+				<NcCheckboxRadioSwitch value="horizontal-split"
+					:button-variant="true"
+					name="mail-layout"
+					type="radio"
+					:checked="layoutMode"
+					button-variant-grouped="vertical"
+					@update:checked="setLayout('horizontal-split')">
+					<template #icon>
+						<HorizontalSplit :size="20" />
+					</template>
+					{{ t('mail', 'Horizontal split') }}
 				</NcCheckboxRadioSwitch>
 			</NcAppSettingsSection>
 			<NcAppSettingsSection id="account-settings" :name="t('mail', 'Account creation')">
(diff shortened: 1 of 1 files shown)


ChristophWurst (commented):
Works great on desktop :+1: 

My only request: allow the thread list to be higher than ~40%. If the screen is big, I want to use 2/3 for the list and 1/3 for the message.

---

Does not work on mobile

image

GretaD:
> Does not work on mobile
> 

thats because of `npm link`, if you copy `/dist` from nextcloud-vue to node_modules/@nextcloud/vue in Mail it will work fine. I had the same issue with `no-split`, and Grigorii mention that its because of `npm link`

package.json
```diff
@@ -48,7 +48,7 @@
     "@nextcloud/moment": "^1.3.1",
     "@nextcloud/paths": "^2.1.0",
     "@nextcloud/router": "^3.0.0",
-    "@nextcloud/vue": "^8.10.0",
+    "@nextcloud/vue": "^8.11.0",
```
GretaD, line 51: i will upgrade the vue in a separate pr, but leaving this here as it is easy to test without linking with nc/vue. It will be removed after reviewing

src/components/AppSettingsMenu.vue
```diff
@@ -4,14 +4,14 @@
...
-			<NcAppSettingsSection id="mail-list-view" :name="t('mail', 'Layout')">
+			<NcAppSettingsSection v-if="!isMobile" id="mail-list-view" :name="t('mail', 'Layout')">
 				<NcCheckboxRadioSwitch value="no-split"
 					:button-variant="true"
 					name="mail-layout"
 					type="radio"
 					:checked="layoutMode"
 					button-variant-grouped="vertical"
-					@update:checked="setLayout">
+					@update:checked="() => setLayout('no-split')">
```
ChristophWurst, line 14:
```suggestion
					@update:checked="setLayout('no-split')">
```

this works the same, no?

src/components/Mailbox.vue
```diff
@@ -94,7 +94,7 @@ export default {
 		},
 		initialPageSize: {
 			type: Number,
-			default: 20,
+			default: 30,
```
ChristophWurst, line 97: yes but why do we need this for the horizontal split? horizontal split is the view with the fewest visible items in the list

ChristophWurst (commented):
Looks good

src/components/Mailbox.vue
GretaD: i can change the initialPageSize to be dynamic

ChristophWurst: but why?

GretaD: so i can show different messages for different layouts.

GretaD:
> After playing around a bit with it a request I would have is some way to be able to close the split, because say I want my full page email list after reading an email (useful for searching). Currently, I don't see any way to do this.

Im not sure if i understand what you mean, if you want to go back to full list of vertical split, you go back to app settings and switch to a different layout.

GVodyanov:
> > After playing around a bit with it a request I would have is some way to be able to close the split, because say I want my full page email list after reading an email (useful for searching). Currently, I don't see any way to do this.
> 
> Im not sure if i understand what you mean, if you want to go back to full list of vertical split, you go back to app settings and switch to a different layout.

Screenshot from 2024-03-18 16-46-43

GVodyanov (commented):
horizontal mail split

Test on Firefox (v121), seems to work well enough, stops showing on mobile (with nextcloud-vue v8.11), but I assume that's intentional.

ChristophWurst (approved):

ratermir:
Hello, I really appreciate the horizontal split of screen, great work. The whole mail app seems to be perfectly usable  in the version 3.6. Only I'd prefer bigger part of mail list (now it is approx. 40%, which is too few). By me it could be from 20% to 80% of screen space - let it on the user whot he/she want to see.

PR #9474, opened 2024-03-19
feat(utility): make json methods work the same for `$data`

miaulalala:
The code used to type `$data` to an array for `error` and `errorFromThrowable`.

Now all methods take a type of `array|JsonSerializable|bool|string $data`

### DIFF
--- a/lib/Http/JsonResponse.php
+++ b/lib/Http/JsonResponse.php
@@ -93,9 +93,16 @@ class JsonResponse extends Base {
 		);
 	}
 
+	/**
+	 * @param string $message
+	 * @param int $status
+	 * @param array|JsonSerializable|bool|string $data
+	 *
+	 * @return static
+	 */
 	public static function error(string $message,
 		int $status = Http::STATUS_INTERNAL_SERVER_ERROR,
-		array $data = [],
+		$data = [],
 		int $code = 0): self {
 		return new self(
 			[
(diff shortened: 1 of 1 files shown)


lib/Http/JsonResponse.php
```diff
@@ -109,11 +116,18 @@ public static function error(string $message,
...
+	 * @param int $status
+	 * @param array|JsonSerializable|bool|string $data
+	 *
+	 * @return static
 	 */
 	public static function errorFromThrowable(Throwable $error,
 		int $status = Http::STATUS_INTERNAL_SERVER_ERROR,
-		array $data = []): self {
+		$data = []): self {
+		if(!is_array($data)) {
```
ChristophWurst, line 128:
```suggestion
		if (!is_array($data)) {
```

ChristophWurst (commented):
s/enh/feat/

lib/Http/JsonResponse.php
```diff
@@ -109,11 +116,18 @@ public static function error(string $message,
...
+	 * @param int $status
+	 * @param array|JsonSerializable|bool|string $data
+	 *
+	 * @return static
 	 */
 	public static function errorFromThrowable(Throwable $error,
 		int $status = Http::STATUS_INTERNAL_SERVER_ERROR,
-		array $data = []): self {
+		$data = []): self {
+		if (!is_array($data)) {
```
ChristophWurst, line 128: can we not keep the unwrapped value?

miaulalala: Hm, not really since we merge the data and serialized exception which is `array|null`

ChristophWurst: You are right

ChristophWurst (approved):
:shipit:

PR #9490, opened 2024-03-21
Fix color for preview when in active state

### LINKED ISSUE #9466: message body in message list for current message does not have enough contrast
### Steps to reproduce

1. Open Mail app in vertical split view
2. click on any message

### Expected behavior

Third row containing body text should be white or another suitable color depending on primary color (as far as I understand `--color-primary-element-text` should work)

image

### Actual behavior

third row is always maxcontrast leading to insufficient contrast between background and text

image

### Mail app version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes #9466 

please test it

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -885,6 +885,12 @@ export default {
 	}
 }
 
+.list-item__wrapper--active {
+	.envelope__preview-text {
+		color: var(--color-primary-element-text);
+	}
+}
+
 .icon-important {
 	:deep(path) {
 	fill: #ffcc00;


hamza221 (dismissed):
Doesn't work because the text itself is not in an active state

`.list-item__wrapper--active{
	.envelope__preview-text {
		color: var(--color-primary-element-text);
	}
}`

Works

hamza221 (approved):
Tested ✅
image

src/components/Envelope.vue
```diff
@@ -885,6 +885,12 @@ export default {
 	}
 }
 
+.list-item__wrapper--active {
+	.envelope__preview-text {
+	color: var(--color-primary-element-text);
+	}
```
ChristophWurst, line 891:
```suggestion
	.envelope__preview-text {
		color: var(--color-primary-element-text);
	}
```

to my eyes

ChristophWurst (approved):

PR #9491, opened 2024-03-21
Fix: misplaced border ncselect

### LINKED ISSUE #9445: Partial border for recipient select
### Steps to reproduce

1. Set up an account
2. Click *New message*
3. Focus the *To* field

### Expected behavior

Good design

### Actual behavior

image

### Mail app version

3.6.0 beta1

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

@GretaD @hamza221 how can we fix this?

hamza221:
Fix #9445 

| b | a |
|--------|--------|
| image |image |

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -48,6 +48,7 @@
 					:reducible="true"
 					:clearable="true"
 					:no-wrap="false"
+					:append-to-body="false"
 					:create-option="createRecipientOption"
 					@input="saveDraftDebounced"
 					@option:selecting="onNewToAddr"
@@ -97,7 +98,7 @@
 					:filter-by="(option, label, search)=>filterOption(option, label, search,'cc')"
 					:taggable="true"
 					:close-on-select="true"
-
+					:append-to-body="false"
 					:multiple="true"
 					:placeholder="t('mail', 'Contact or email address …')"
 					:aria-label-combobox="t('mail', 'Contact or email address …')"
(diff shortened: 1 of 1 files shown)


GretaD:
i thought the issue was that the dropdown not having a top border.

hamza221:
> i thought the issue was that the dropdown not having a top border.

Yes, I realised that later after checking pre-vue8 , that's why I turned it into draft

Pre vue8 there was no border and no slot for no results

hamza221:
image

hamza221:
I left the No-options slot in 
image

hamza221:
PHP-CS unrelate to this PR

ChristophWurst:
https://github.com/nextcloud/mail/pull/9552

src/components/Composer.vue
```diff
@@ -96,7 +97,7 @@
 					:filter-by="(option, label, search)=>filterOption(option, label, search,'cc')"
 					:taggable="true"
 					:close-on-select="true"
-
+					::append-to-body="false"
```
GVodyanov, line 100: Typo

GVodyanov (approved):
Works!

GretaD (commented):
when i setup an account for the first time and i have no search results, it is shown like this. 
If i have at least one search result, it is shown correctly.

Screenshot from 2024-04-19 11-21-10

ChristophWurst:
The borderless design does not look very appealing. Could we fully switch to the default ncselect styling with borders? We have bordered and stacked inputs in other places too: 
image

So I guess it would be fine to also have the borders in the Mail composer. @jancborchardt objections? if so, please provide mockups for a borderless select component

hamza221:
@nextcloud/designers any suggestions?

marcoambrosini (commented):
I think it makes sense to use our components as they natively look as @ChristophWurst suggests. Border-with will decrease to 1 pixel with the next release so they won't look as prominent. 

If with time we realise that it's still too much, I would prefer finding a solution upstream, since these dense forms are ubiquitous in Nextcloud.

ChristophWurst (approved):

PR #9492, opened 2024-03-22
Fix imap smtp button on firefox

### LINKED ISSUE #9467: IMAP/SMTP security button styling
### Steps to reproduce

1. Open the account setup
2. Change to manual mode

### Expected behavior

Good design

### Actual behavior

image

### Mail app version

3.6.0

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes #9467 

after
Screenshot from 2024-03-22 11-55-00

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -741,7 +741,7 @@ input[type='radio'][disabled] + label {
 }
 #account-form {
 	z-index: 1001;
-	width: 250px;
+	width: 300px;
 	top: 15%;
 	padding-bottom: 50px;
 	margin: 0 auto;


src/components/AccountForm.vue
```diff
@@ -694,7 +694,15 @@ h4 {
 .flex-row {
 	display: flex;
 }
-
+@-moz-document url-prefix() {
```
ChristophWurst, line 697: could you find out why firefox styles it differently?

GretaD: we have dedicated styles for firefox in other parts of our code. I dont know though, why specifically these buttons are broken only for FF.

ChristophWurst (changes requested):
With the current solution the buttons are wider than the rest of the form. This looks out of place.

Firefox must make the buttons higher because there is not enough space otherwise.

What about making `#account-form` 280px wide?

Bildschirmfoto vom 2024-03-25 17-35-09
Bildschirmfoto vom 2024-03-25 17-34-59

ChristophWurst:
This is the reason why it wasn't broken on Chrome:

image

… which means it was broken too, just differently. Firefox switched to a line break and preserved the maximum width. Chrome kept the single line and escaped the maximum width.

ChristophWurst (approved):

PR #9494, opened 2024-03-22
fix: remove deprecated prototype.substr() method

miaulalala:
Apparently substr() is dead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

### DIFF
--- a/src/util/eventAttendee.js
+++ b/src/util/eventAttendee.js
@@ -27,8 +27,8 @@
  * @return {string} URI without a mailto prefix
  */
 export function removeMailtoPrefix(uri) {
-	if (uri.startsWith('mailto:')) {
-		return uri.substr(7)
+	if (uri.startsWith('mailto:') || uri.startsWith('MAILTO:')) {
+		return uri.substring(7)
 	}
 
 	return uri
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

src/util/eventAttendee.js
```diff
@@ -27,8 +27,8 @@
  * @return {string} URI without a mailto prefix
  */
 export function removeMailtoPrefix(uri) {
-	if (uri.startsWith('mailto:')) {
-		return uri.substr(7)
+	if (uri.startsWith('mailto:') || uri.startsWith('MAILTO:')) {
```
ChristophWurst, line 30: I missed this. is this change intentional?

miaulalala: yeah? The ics I got from Patrick contains an uppercase `MAILTO:`. I figured `substring()` is case sensitive. Is that not the case?

ChristophWurst: > Is that not the case?

Good question. Did you test that?

miaulalala: uh, `startsWith` is definitely case sensitive

PR #9509, opened 2024-04-02
Filter messages sent in the last 7 days and from me

### LINKED ISSUE #9182: Add filter shortcuts (like Has attachment, Last 7 days and From me)

To make mail app more efficient we should add some more filter on the advance search modal.
3 proposals would be:

1. Has attachment, 
2. Last 7 days  
3. From me

_No response_

_No response_

### Additional context

_No response_

GretaD:
fixes #9182 

with pressed state

Screenshot from 2024-04-18 13-23-12
Screenshot from 2024-04-18 13-22-38

### DIFF
--- a/src/components/SearchMessages.vue
+++ b/src/components/SearchMessages.vue
@@ -1,276 +1,306 @@
 <template>
-	<div class="search-messages">
-		<input v-model="query"
-			type="text"
-			class="search-messages--input"
-			:placeholder="t('mail', 'Search in mailbox')">
-		<NcButton v-if="filterChanged"
-			:aria-label="t('mail', 'Close')"
-			class="search-messages--close"
-			@click="resetFilter()">
-			<template #icon>
-				<Close :size="24" />
-			</template>
-		</NcButton>
-
-		<span v-if="filterChanged"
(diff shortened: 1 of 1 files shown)


GretaD:
when i click "from me" the from field is field with my email, but the option is not clicked
Screenshot from 2024-04-04 17-36-54

ChristophWurst:
I think the filter shortcuts should be regular buttons, not radio buttons

GretaD:
@st3iny had a comment that he would prefer that the "has attachment" to be a checkbox instead of a button. I left it as a button, as it was requested that the shortcuts to be buttons. But i can change it again, if more agree with Richard.

st3iny (changes requested):
The search is not reset after disabling a button again. I guess that should work?

E.g. I click `Last 7 days` and then click it again and expect that all mails are now shown again as the filter is toggled. However, the filter stays on last 7 days even if the button is not primary any longer.


https://github.com/nextcloud/mail/assets/1479486/327ad8ad-4d5c-42b3-8381-1825ee14645a

st3iny (approved):
Tested and works.

PR #9522, opened 2024-04-03
fix(deps): Replace @nextcloud/vue-dashboard with @nextcloud/vue

ChristophWurst:
`@nextcloud/vue-dashboard` is dead.

## How to test

1. Set up an account
2. Open the dashboard
3. Enable unread/import widgets

Main: it works
Here: it works


st3iny (approved):
Dashboard still works.

Issue #6627, opened 2022-06-02
App does not prevent accidental leave that causes data loss


ChristophWurst:
### Steps to reproduce

1. Compose a new message
2. Before draft saving kicks in, close the tab

### Expected behavior

The app asks me for confirmation and only drops my data if I consent.

### Actual behavior

Data is silently dropped.

### Additional info

This can be solved by JavaScript's `onbeforeunload`.

thstyl2000:
It's actually even more troublesome in the following case:

1. Start writing an email.
2. Draft is saved for the first time. "Draft saved" notification is shown.
3. Continue writing for some time. "Draft saved" notification is continuously shown.
4. Remember sth you need to check from another email.
5. You close the editor directly.
6. You lose your last couple of sentences because the email is saved periodically.

PR #9525, opened 2024-04-04
fix(composer): Prevent leaving the tab with unsaved changes

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/6627

## How to test
1. Open the app
2. Start writing a new message
3. (optional: Minimize the composer)
4. Close the tab

Main: data is lost.
Here: browser ask the user if they really want to leave. If the composer was minimized it will be maximized.

Ref https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -205,11 +205,15 @@ export default {
 		if (id) {
 			this.draftsPromise = Promise.resolve(id)
 		}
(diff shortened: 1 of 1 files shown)


GVodyanov (approved):
Works for Firefox 124.0.1 and Chromium Version 123.0.6312.105 (Official Build), Manjaro Linux (64-bit)

hamza221 (approved):

PR #9527, opened 2024-04-05
ci: Silence PR feedback acount about Nextcloud GmbH employees

ChristophWurst:
There have been dozens of false reports in the past few days. This should silence them.


marcelklehr (approved):
If these people are full employees, the ideal way to deal with this would be to add a link to their github to http://nextcloud.com/team, but this will do in the meantime.

ChristophWurst:
Grigory is on the team page

marcelklehr:
ah, there's a final / at the end of his github handle that freaks out the regex, I can fix it next week

PR #9553, opened 2024-04-11
fix: Sanitize forward slashes from name before generating url

SebastianKrupinski:
Sanitize any forward slashes from guest user name before generating a url

Fixes https://github.com/nextcloud/mail/pull/9553


ChristophWurst (approved):

ChristophWurst:
@SebastianKrupinski please see https://github.com/nextcloud/mail/pull/9553/checks?check_run_id=23709692723. You need to sign off your commits

Issue #9452, opened 2024-03-08
Apply recipient search input on blur


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud user I might forget to *submit* the input in the *to*/*cc*/*bcc* fields and just click/tab to the message body. Right now that doesn't register my input as the value.

#### Steps to reproduce

1. Open the app
2. Click *+ New message*
3. Enter *email@domain.tld*
4. Press <kbd>Tab</kbd>

### Describe the solution you'd like

On NcSelect blur (lost focus), take current search input and add it as recipient.

In Jan's words (https://github.com/nextcloud/mail/issues/7149)
> Tabbing out or unfocusing the field should also accept what was put in as recipient.

This makes perfect sense for email addresses in the NcSelect search field. Applying "Jane Doe" as recipient probably makes no sense.

### Describe alternatives you've considered

N/a

### Additional context

https://github.com/nextcloud/mail/issues/7149

PR #9566, opened 2024-04-14
Enh/add recipient on blur

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/9452

Adds mail recipients automatically when the NcSelect is deselected


src/components/Composer.vue
```diff
@@ -1209,16 +1216,27 @@ export default {
...
+			if (
+				(option === null || option === undefined)
+				&& this.recipientSearchTerms[type] !== undefined
+				&& this.recipientSearchTerms[type] !== ''
+			) {
+				option = {}
+				option.email = this.recipientSearchTerms[type]
+				option.label = this.recipientSearchTerms[type]
+				this.recipientSearchTerms[type] = ''
+			}
```
ChristophWurst, line 1237: I'm not sure I understand this trick. Is this for when you leave the input when a suggestion is shown in the dropdown?

GVodyanov: Took me a while to remember what was going on here too 😅

---

So ```(option === null || option === undefined)``` is needed because that function is called in two cases:
1. You selected one of the suggested options from the dropdown
2. You click outside the dropdown, so the recipient is added anyway

If you were to click on the dropdown without this check it would fire twice, once for the selection of the dropdown (`@option:selecting`), and once for the click closing the dropdown (`@search:blur`).

---

```
&& this.recipientSearchTerms[type] !== undefined
&& this.recipientSearchTerms[type] !== ''
``` 
Just normal checks

---
```
option = {}
option.email = this.recipientSearchTerms[type]
option.label = this.recipientSearchTerms[type]
```
Overriding the `option` parameter with not the dropdown content but the string content of the search of the `NcSelect`

```
this.recipientSearchTerms[type] = ''
```
Resetting the string content of the search of the `NcSelect` seeing as that string has already been added as a recipient

Hope this clears things up @ChristophWurst :)

src/components/Composer.vue
```diff
@@ -1178,6 +1183,8 @@ export default {
 			this.loadingIndicatorTo = loadingIndicator === 'to'
 			this.loadingIndicatorCc = loadingIndicator === 'cc'
 			this.loadingIndicatorBcc = loadingIndicator === 'bcc'
+			this.recipientSearchTerms[loadingIndicator] = term
+
```
hamza221, line 1187: loadingIndicator should maybe be renamed , for code comprehension since it's used as loading Indicator and an input indicator now . 
Maybe something more general like `addressType` or just `type` would make more sense here

hamza221 (approved):
Tested ✅ maybe as a followup we check if the value is in email format, before adding the address

GVodyanov:
> Tested ✅ maybe as a followup we check if the value is in email format, before adding the address

Right, do we have a specific function for that? I wasn't able to find anything but it feels like we should have a standard for that, currently I just inserted a check on the `@` character.

Also an issue I'm encountering is the search term being cleared even if the term wasn't added to the email list, I will investigate it further later but it's a little weird.

hamza221:
I wrote a little hacky work-arround, But idk if it's really necessary. what do you think @ChristophWurst? 
Feel free to discard it, If you find a better way

GVodyanov:
> I wrote a little hacky work-arround, But idk if it's really necessary. what do you think @ChristophWurst? Feel free to discard it, If you find a better way

Peek 2024-05-28 18-07
Tested it, works well, and in my opinion it would do @hamza221

ChristophWurst (approved):

Issue #9099, opened 2023-11-27
Recipient search avatar flickers


ChristophWurst:
### Steps to reproduce

1. Set up an account
2. Compose a new message
3. Enter a recipient name or email

### Expected behavior

After each key stroke, the avatar changes color

https://github.com/nextcloud/mail/assets/1374172/f12d01dd-a252-4765-8e3e-eca95c125553

### Actual behavior

No flicker or simply no avatar until I select a contact or confirm my input

### Mail app version

3.5.0

### PHP engine version

None

### Web server

None

### Database

None

PR #9576, opened 2024-04-18
Add static icon for unknown user

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/mail/issues/9099

Add static icon for unknown user to prevent new avatar generation for each input.

**Before:**

https://github.com/nextcloud/mail/assets/1374172/f12d01dd-a252-4765-8e3e-eca95c125553

**After:** 

Screenshot from 2024-04-18 11-00-00

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -69,6 +69,7 @@
 							<ListItemIcon :no-margin="true"
 								:name="option.label"
 								:subtitle="option.email"
+								:icon-class="!option.id ? 'icon-user' : null"
 								:url="option.photo"
 								:avatar-size="24" />
 						</div>
@@ -126,6 +127,7 @@
 								:name="option.label"
 								:subtitle="option.email"
 								:url="option.photo"
+								:icon-class="!option.id ? 'icon-user' : null"
 								:avatar-size="24" />
 						</div>
 					</template>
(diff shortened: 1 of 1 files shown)


hamza221 (approved):

src/components/Composer.vue
```diff
@@ -69,6 +69,7 @@
 							<ListItemIcon :no-margin="true"
 								:name="option.label"
 								:subtitle="option.email"
+								:iconClass="!option.id ? 'icon-user' : null"
```
ChristophWurst, line 72: I am not sure if icon classes are still supported. I remember that we wanted to get rid of them an replace with inline (material) icons.

icon-class is not documented at https://nextcloud-vue-components.netlify.app/#/Components/NcListItems?id=nclistitemicon either

JuliaKirschenheuter: > icon-class is not documented at 

`iconClass` is a valid prop of `NcAvatar` https://nextcloud-vue-components.netlify.app/#/Components/NcAvatar and will be thrown via attrs from `NcListItems`.

In current solution of `NcListItems` it is not possible to use material icons instead. Vue component need to be adjusted + NcVue release is needed.

ChristophWurst (approved):

src/components/Composer.vue
ChristophWurst, line 72: Attribute ':iconClass' must be hyphenated

st3iny (approved):
Tested and works.

JuliaKirschenheuter:
CI is not related

PR #9582, opened 2024-04-19
fix: remove unused package

GretaD:


ChristophWurst (changes requested):
1. This is a fix, not a chore. You *remove* the package, you don't bump it
2. package-lock is out of sync

ChristophWurst (approved):
Makes sense

PR #9584, opened 2024-04-19
fix(outbox): handle indeterminate smtp errors

miaulalala:
Sometimes, if the SMTP server has connectivity issues, SMTP errors are thrown even if the mail has already been sent (although it's an indeterminate state :cat2: :atom: :skull: ). In those cases, we have no way to check if the email has actually been sent to all intended recipients.

This PR introduces a new state for these types of errors, and aims to give the user more information in the outbox list items on what happened. The outbox now allows (re)sending the message or copying the message to the sent outbox.

As yet to be addressed is the case where there are existing outbox messages that have failed previously. 
We should set them to the new error status. But, as their raw message text is not available, this will cause errors when trying to copy them to the sent mailbox. We could also treat them entirely different by checking for the raw body text and only allowing a resend and delete operation.

Some styling issues for the outbox are also blocking this issue.

To Do:

- [x] How to handle existing messages without `raw` body - rebuild the message body as a mime message from the stored data.
- [x] Styling the ListItem

### DIFF
--- a/lib/Send/Chain.php
+++ b/lib/Send/Chain.php
@@ -25,7 +25,9 @@ namespace OCA\Mail\Send;
 use OCA\Mail\Account;
 use OCA\Mail\Db\LocalMessage;
 use OCA\Mail\Db\LocalMessageMapper;
+use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\Service\Attachment\AttachmentService;
+use OCP\DB\Exception;
 
 class Chain {
 	public function __construct(private SentMailboxHandler $sentMailboxHandler,
@@ -38,12 +40,26 @@ class Chain {
 	) {
 	}
 
+	/**
+	 * @throws \Throwable
+	 * @throws Exception
+	 * @throws ServiceException
+	 */
 	public function process(Account $account, LocalMessage $localMessage): void {
 		$handlers = $this->sentMailboxHandler;
 		$handlers->setNext($this->antiAbuseHandler)
 			->setNext($this->sendHandler)
 			->setNext($this->copySentMessageHandler)
 			->setNext($this->flagRepliedMessageHandler);
+
+		/**
+		 * Skip all messages that errored out indeterminedly in the SMTP send.
+		 * @see \Horde_Smtp_Exception  for the error codes that are inderminate
+		 * They might or might not have been sent already.
+		 */
+		if ($localMessage->getStatus() === LocalMessage::STATUS_ERROR) {
+			throw new ServiceException('Could not send message because a previous send operation produced an unclear sent state.');
+		}
 
 		$result = $handlers->process($account, $localMessage);
 		if ($result->getStatus() === LocalMessage::STATUS_PROCESSED) {
--- a/lib/Send/CopySentMessageHandler.php
+++ b/lib/Send/CopySentMessageHandler.php
@@ -35,7 +35,7 @@ class CopySentMessageHandler extends AHandler {
 	public function __construct(private IMAPClientFactory $imapClientFactory,
 		private MailboxMapper $mailboxMapper,
 		private LoggerInterface $logger,
-		private MessageMapper $messageMapper
+		private MessageMapper $messageMapper,
 	) {
 	}
 	public function process(Account $account, LocalMessage $localMessage): LocalMessage {
--- a/lib/Service/MailTransmission.php
+++ b/lib/Service/MailTransmission.php
@@ -38,6 +38,7 @@ use Horde_Mime_Headers_MessageId;
 use Horde_Mime_Headers_Subject;
 use Horde_Mime_Mail;
 use Horde_Mime_Mdn;
+use Horde_Smtp_Exception;
 use OCA\Mail\Account;
 use OCA\Mail\Address;
 use OCA\Mail\AddressList;
@@ -63,6 +64,12 @@ use OCP\EventDispatcher\IEventDispatcher;
 use Psr\Log\LoggerInterface;
 
 class MailTransmission implements IMailTransmission {
+	private const RETRIABLE_CODES = [
+		Horde_Smtp_Exception::INSUFFICIENT_STORAGE,
+		Horde_Smtp_Exception::OVERQUOTA,
+		Horde_Smtp_Exception::LOGIN_REQUIREAUTHENTICATION,
+	];
+
 	public function __construct(
 		private IMAPClientFactory $imapClientFactory,
 		private SmtpClientFactory $smtpClientFactory,
@@ -142,8 +149,12 @@ class MailTransmission implements IMailTransmission {
 			$mail->send($transport, false, false);
 			$localMessage->setRaw($mail->getRaw(false));
 		} catch (Horde_Mime_Exception $e) {
-			$localMessage->setStatus(LocalMessage::STATUS_SMPT_SEND_FAIL);
 			$this->logger->error($e->getMessage(), ['exception' => $e]);
+			if (in_array($e->getCode(), self::RETRIABLE_CODES, true)) {
+				$localMessage->setStatus(LocalMessage::STATUS_SMPT_SEND_FAIL);
+				return;
+			}
+			$localMessage->setStatus(LocalMessage::STATUS_ERROR);
 			return;
 		}
 
--- a/src/components/OutboxMessageListItem.vue
+++ b/src/components/OutboxMessageListItem.vue
@@ -21,8 +21,7 @@
   -->
 
 <template>
-	<ListItem v-if="message.status !== statusImapSentMailboxFail()"
-		class="outbox-message"
+	<ListItem class="outbox-message"
 		:class="{ selected }"
 		:name="title"
 		:details="details"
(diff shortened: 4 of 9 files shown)


lib/Migration/Version3600Date20240220134813.php
```diff
@@ -61,4 +68,17 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		return $schema;
 	}
 
+	public function postSchemaChange(IOutput $output, \Closure $schemaClosure, array $options) {
+		// Let's buffer this a bit
+		$aMinuteAgo = time() - 60;
+		/** @var QueryBuilder $query */
+		$query = $this->connection->getQueryBuilder();
+		$query->update('mail_local_messages')
+			->set('status', $query->createNamedParameter(LocalMessage::STATUS_IMAP_SENT_MAILBOX_FAIL, IQueryBuilder::PARAM_INT))
```
ChristophWurst, line 77: this will explode for we updates if the LocalMessage class has been loaded before the code is replaced and the update is run. Inline the const.

tests/Unit/Send/ChainTest.php
```diff
@@ -139,4 +138,37 @@ public function testProcessNotProcessed() {
...
+		$expected->setId(100);
+
+		$this->sentMailboxHandler->expects(self::any())
+			->method('setNext')
+			->withConsecutive(
+				[$this->antiAbuseHandler],
+				[$this->sentMailboxHandler],
+				[$this->copySentMessageHandler],
+				[$this->flagRepliedMessageHandler],
+			);
```
ChristophWurst, line 161: sentMailboxHandler::setNext is only called once, not consecutively. try to change self::any() to self::exactly(4) and you should see failing assertions

tests/Unit/Send/ChainTest.php
```diff
@@ -139,4 +138,37 @@ public function testProcessNotProcessed() {
...
+		$this->sentMailboxHandler->expects(self::never())
+			->method('process');
+		$this->attachmentService->expects(self::never())
+			->method('deleteLocalMessageAttachments');
+		$this->localMessageMapper->expects(self::never())
+			->method('deleteWithRecipients');
+		$this->localMessageMapper->expects(self::never())
+			->method('update');
```
ChristophWurst, line 169: you are testing the behavior of sentMailboxHandler here. SentMailboxHandler is a stub and calling process will never call any other methods. If you want to test this path, do it in the unit test for the sent mailbox handler.

lib/Send/CopySentMessageHandler.php
```diff
@@ -68,6 +80,11 @@ public function process(Account $account, LocalMessage $localMessage): LocalMess
 			return $localMessage;
 		}
 
+		if(empty($localMessage->getRaw())) {
+			$mail = $this->buildRaw($account, $localMessage);
+			$localMessage->setRaw($mail->getRaw(false));
+		}
```
ChristophWurst, line 86: we can probably not fix it but messages where the mime is generated here are not identical to the sent messages. e.g. they won't have a message-id set, so they will not show in threads.

lib/Send/CopySentMessageHandler.php
```diff
@@ -88,4 +105,64 @@ public function process(Account $account, LocalMessage $localMessage): LocalMess
 
 		return $this->processNext($account, $localMessage);
 	}
+
+	private function buildRaw(Account $account, LocalMessage $localMessage): \Horde_Mime_Mail {
```
ChristophWurst, line 109: would it be possible to deduplicate code from ``\OCA\Mail\Service\MailTransmission::sendMessage``?

lib/Migration/Version3600Date20240220134813.php
```diff
@@ -61,4 +68,17 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		return $schema;
 	}
 
+	public function postSchemaChange(IOutput $output, \Closure $schemaClosure, array $options) {
+		// Let's buffer this a bit
+		$aMinuteAgo = time() - 60;
+		/** @var QueryBuilder $query */
```
ChristophWurst, line 74: program to to the interface. drop this type comment.

lib/Service/MailTransmission.php
```diff
@@ -142,8 +149,12 @@ public function sendMessage(Account $account, LocalMessage $localMessage): void
 			$mail->send($transport, false, false);
 			$localMessage->setRaw($mail->getRaw(false));
 		} catch (Horde_Mime_Exception $e) {
-			$localMessage->setStatus(LocalMessage::STATUS_SMPT_SEND_FAIL);
 			$this->logger->error($e->getMessage(), ['exception' => $e]);
+			if (in_array($e->getCode(), $this->retriableCodes)) {
```
ChristophWurst, line 153:
```suggestion
			if (in_array($e->getCode(), $this->retriableCodes, true)) {
```

Install https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-

lib/Send/CopySentMessageHandler.php
miaulalala: We could move it to the TransmissionService. Shall I do the refactoring in this PR?

ChristophWurst: Will it make the changes smaller if we have less duplicated code?

lib/Service/MailTransmission.php
```diff
@@ -142,8 +149,12 @@ public function sendMessage(Account $account, LocalMessage $localMessage): void
...
+				$localMessage->setStatus(LocalMessage::STATUS_SMPT_SEND_FAIL);
+				return;
+			}
+			$localMessage->setStatus(LocalMessage::STATUS_ERROR);
 			return;
```
ChristophWurst, line 158: this seems like the essential *fix* that we can also backport. It makes recoverable errors retryable automatically. Could we extract/reduce this as the backportable part and have the copy to sent mailbox *feature* only for main/3.7?

lib/Send/CopySentMessageHandler.php
miaulalala: leaving this in for now - I've tried unsuccessfully to keep the whole code in tune with the $localMessage having to be returned to the chain when there's an SMIME error. I'm probably just too f*king stupid to make it work, so sorry in advance.

lib/Migration/Version3600Date20240220134813.php
```diff
@@ -61,4 +66,16 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
 
+	public function postSchemaChange(IOutput $output, \Closure $schemaClosure, array $options) {
+		// Let's buffer this a bit
+		$aMinuteAgo = time() - 60;
+		$query = $this->connection->getQueryBuilder();
+		$query->update('mail_local_messages')
+			->set('status', $query->createNamedParameter(11, IQueryBuilder::PARAM_INT))
+			->where(
+				$query->expr()->lt('send_at', $query->createNamedParameter($aMinuteAgo, IQueryBuilder::PARAM_INT))
+			);
```
ChristophWurst, line 77: should we only set this for type=outbox?

lib/Service/MailTransmission.php
```diff
@@ -63,6 +64,12 @@
 use Psr\Log\LoggerInterface;
 
 class MailTransmission implements IMailTransmission {
+	private array $retriableCodes = [
```
ChristophWurst, line 67: should be a const

src/components/OutboxMessageListItem.vue
```diff
@@ -176,22 +171,37 @@ export default {
...
-					await this.$store.dispatch('outbox/copyMessageToSentMailbox', { id: message.id })
+				await this.$store.dispatch('outbox/sendMessageWithUndo', { id: message.id })
+			} catch (error) {
+				logger.error('Could not send message', { error })
+				if (error.data !== undefined) {
+					await this.$store.dispatch('outbox/updateMessage', { message: error.data[0], id: message.id })
 				}
+			}
+		},
+		async copyToSentMailbox() {
```
ChristophWurst, line 185: is this needed for the fix?

miaulalala: I can revert but it simply splits the previous `if` into two separate methods

lib/Send/Chain.php
```diff
@@ -38,13 +40,27 @@ public function __construct(private SentMailboxHandler $sentMailboxHandler,
...
 			->setNext($this->sendHandler)
 			->setNext($this->copySentMessageHandler)
 			->setNext($this->flagRepliedMessageHandler);
 
+		/**
+		 * Skip all messages that errored out indeterminedly in the SMTP send.
+		 * @see \Horde_Smtp_Exception  for the error codes that are inderminate
+		 * They might or might not have been sent already.
+		 */
+		if($localMessage->getStatus() === LocalMessage::STATUS_ERROR) {
```
ChristophWurst, line 60:
```suggestion
		if ($localMessage->getStatus() === LocalMessage::STATUS_ERROR) {
```

ChristophWurst (approved):
:shipit:

PR #9601, opened 2024-04-25
Important mailS (not mail)

Jerome-Herbinet:
Because, most of the time, there are more than "one" important mail

### DIFF
--- a/lib/Dashboard/ImportantMailWidget.php
+++ b/lib/Dashboard/ImportantMailWidget.php
@@ -38,7 +38,7 @@ class ImportantMailWidget extends MailWidget {
 	 * @inheritDoc
 	 */
 	public function getTitle(): string {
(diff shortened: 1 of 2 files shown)


ChristophWurst:
https://www.nytimes.com/2010/05/09/magazine/09FOB-onlanguage-t.html

Jerome-Herbinet:
> https://www.nytimes.com/2010/05/09/magazine/09FOB-onlanguage-t.html

Not everyone seems to agree, and there are same quirks that divide people in French too 🙂. 

In the meantime, to play devil's advocate, or confront my opinion with yours, here's a screenshot taken at Thunderbird :

image

What do you think ?

st3iny:
Thunderbird is using Emails where the plural makes sense IMO. But for the word mail it makes sense to keep the current form.

Issue #9453, opened 2024-03-08
Phishing detection


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud Mail user, I want the app to help me detect phishing emails. Some of these phishing emails are not easy to spot.

### Describe the solution you'd like

Run some sanity checks on the email (to be brainstormed) and
1. Show a warning
2. Disable external links
3. Offer to move the message to spam

### Work packages
- [x] Research, brainstorming and goal definition
  - [x] Compare sender with sender label
  - [x] Trusted domains
  - [x] DKIM
  - [x] BIMI https://github.com/nextcloud/mail/pull/5784
- [x] Implementation
   - [x]  anchor tag text vs link checking #9994 
   - [x] Highlighting messages with a different reply-to #9610
   - [x] check full name address against contacts #9610
- [x] #9891
- [ ] Unit test for link checking https://github.com/nextcloud/mail/pull/10075
- [x] Admin docs if applicable (Not applicable) 
- [x] https://github.com/nextcloud/documentation/pull/12152

### Follow-up 

- [ ] Add a setting to disable HTML emails by default  
- [ ] Add quick actions (example: add email address to contact)
- [ ] Check emails for selectively applied CSS on forwards
context : https://www.schneier.com/blog/archives/2024/04/security-vulnerability-of-html-emails.html 

### Suggestion 
- [ ] Add a hint if a mail comes from an address that is neither in "contacts" nor has send any mails before. @kaffeeringe 

### Out of scope 
- resolving domain names
- Build a process to integrate against configurable Lists
- SPF 
- Preconfigured trust custom header tags 

### Describe alternatives you've considered

N/a

### Additional context

https://github.com/nextcloud/mail/issues/7864

kaffeeringe:
Really important topic. I personally would like to get rid off all the fake DHL-notifications. But this is an essential feature for organisations.

kaffeeringe:
Maybe add a hint if a mail comes from an address that is neither in "contacts" nor has send any mails before.

PR #9610, opened 2024-04-29
Feat: phishing detection

hamza221:
Ref https://github.com/nextcloud/mail/issues/9453


lib/Service/PhishingDetectionService.php
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	private function replyToCheckPass(string $fromEmail, ?string $replyToEmail): array {
+		if(!(isset($replyToEmail))) {
+			return ["check" => true];
+		}
+		if($replyToEmail !== $fromEmail) {
+			$this->warn = true;
+		}
+		return ["check" => false , "message" => "Reply-To email: ${$replyToEmail} is different from the sender email: ${$fromEmail}"];
```
ChristophWurst, line 66: l10n

lib/Service/PhishingDetectionService.php
```diff
@@ -0,0 +1,190 @@
...
+ */
+
+namespace OCA\Mail\Service;
+
+use DateTime;
+use Horde_Mime_Headers;
+use OCA\Mail\AddressList;
+use OCA\Mail\Contracts\ITrustedSenderService;
+
+class PhishingDetectionService {
```
ChristophWurst, line 33: this new class needs tests

lib/Service/PhishingDetectionService.php
```diff
@@ -0,0 +1,190 @@
...
+
+		if(!$trusted) {
+			$this->warn = true;
+			return ["check" => false, "message" => "Sender email: ${$email} is not trusted"];
+		}
+		return ["check" => true];
+	}
+
+	private function isLink(string $text): bool {
+		$pattern = '/^(https?:\/\/|www\.|[a-zA-Z0-9-]+\.[a-zA-Z]{2,})/i';
```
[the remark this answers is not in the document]
pabzm: +1 for not using a custom Regexp.

But if I understand correctly this method isn't supposed to match only fully qualified URLs, but all text that looks roughly like a URL (as part of a link text).

If I'm correct then `FILTER_VALIDATE_URL` would be too strict because it doesn't accept e.g. `cloud.nextcloud.com/p/something`. Additionally it doesn't accept URLs containing non-ASCII characters.

I'd suggest to use a library to find URLs in text. [The Horde lib "Text_Filter" contains a class](https://dev.horde.org/api/master/lib/Text_Filter/classes/Horde_Text_Filter_Linkurls.html) that links all found URLs, maybe some methods of that can be used? 

Also maybe the method could be renamed to something more telling (e.g. "textLooksLikeALink()"), and get a comment about its purpose?

lib/Service/PhishingDetectionService.php
```diff
@@ -0,0 +1,190 @@
...
+
+	private bool $warn = false;
+
+
+	public function __construct(ContactsIntegration $contactIntegration, ITrustedSenderService $trustedSenderService) {
+		$this->contactIntegration = $contactIntegration;
+		$this->trustedSenderService = $trustedSenderService;
+	}
+
+	private function checkDatePass(string $date): array {
```
pabzm, line 49: I would suggest to name all checking methods in a similar way to make this more comprehensible. Some have "check" in the beginning, some at the end. And what does the "Pass"-suffix on some methods mean?

lib/Service/PhishingDetectionService.php
```diff
@@ -0,0 +1,190 @@
...
+	private function parseAnchorTags(string $htmlMessage): array {
+
+		$results = [];
+		$zippedArray = [];
+
+		$dom = new \DOMDocument();
+		libxml_use_internal_errors(true);
+		$dom->loadHTML($htmlMessage);
+		libxml_use_internal_errors();
+		$anchors = $dom->getElementsByTagName('a');
```
pabzm, line 138: `area`-elements can contain links and could be checked, too. See https://developer.mozilla.org/en-US/docs/Web/API/Document/links

hamza221: So we would check href against alt in <area> ?

pabzm: Yes, or match the URL against a block-list.

hamza221: block-list is out of scope. 
I'll add a check against alt

lib/Service/PhishingDetectionService.php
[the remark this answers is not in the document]
pabzm: You could use the original source, yes. But a library might bring the benefit that possible changes (there's a list of TLDs in there, which may change) are tended for by multiple people/projects – I'm unsure about the Horde libs in this regard, though.

kesselb:
Great presentation today :+1: 

Sorry, I didn't have the opportunity to review the pr. 

1) I'm uncertain about the reply-to warning. The purpose of reply-to is to send the response to a different email. I wonder if we want to use a less scary warning for such a case. For example, our calendar invitations are also using reply-to. 

2) The PhishingDetectionService reminds me a bit about our CheckSetupController in server. I think we could move every check to an own class. Testing is also easier than. Wdyt @ChristophWurst?

lib/PhishingDetectionList.php
```diff
@@ -0,0 +1,69 @@
...
+	private bool $warning = false;
+
+	public function __construct(array $checks = []) {
+		$this->checks = $checks;
+	}
+
+	public function addCheck(PhishingDetectionResult $check) {
+		$this->checks[] = $check;
+	}
+	private function isWarning() {
```
st3iny, line 49:
```suggestion
	}
	
	private function isWarning() {
```

Nit

lib/PhishingDetectionResult.php
```diff
@@ -0,0 +1,74 @@
...
+		$this->message = $message;
+		$this->isPhishing = $isPhishing;
+		$this->additionalData = $additionalData;
+
+	}
+
+	public function getType(): string {
+		return $this->type;
+	}
+	public function isPhishing(): bool {
```
st3iny, line 60:
```suggestion
	}
	
	public function isPhishing(): bool {
```

Nit

lib/Service/PhishingDetection/DateCheck.php
```diff
@@ -0,0 +1,49 @@
...
+use OCP\IL10N;
+
+class DateCheck {
+	protected IL10N $l10n;
+
+	public function __construct(IL10N $l10n) {
+		$this->l10n = $l10n;
+	}
+
+	public function run(string $date):PhishingDetectionResult {
```
st3iny, line 39:
```suggestion
	public function run(string $date): PhishingDetectionResult {
```

Nit

lib/Address.php
```diff
@@ -68,9 +68,27 @@ public function getLabel(): ?string {
 			// Fallback
 			return $this->getEmail();
 		}
+		$personal = trim(explode('<', $personal)[0]); // Remove the email part if present
```
st3iny, line 71: I don't quite understand this line. Please elaborate: Why does the personal part contain `<`? Is this a bug in Horde?

st3iny (changes requested):
Did not test but left some comments on the code.

lib/Service/PhishingDetection/PhishingDetectionService.php
```diff
@@ -0,0 +1,61 @@
...
+
+	public function __construct(private ContactCheck $contactCheck, private CustomEmailCheck $customEmailCheck, private DateCheck $dateCheck, private ReplyToCheck $replyToCheck) {
+		$this->contactCheck = $contactCheck;
+		$this->customEmailCheck = $customEmailCheck;
+		$this->dateCheck = $dateCheck;
+		$this->replyToCheck = $replyToCheck;
+		$this->list = new PhishingDetectionList();
+	}
+
+
```
st3iny, line 46:
```suggestion


```

Nit

lib/Service/PhishingDetection/PhishingDetectionService.php
```diff
@@ -0,0 +1,61 @@
...
+use Horde_Mime_Headers;
+use OCA\Mail\AddressList;
+use OCA\Mail\PhishingDetectionList;
+
+class PhishingDetectionService {
+
+
+	/** @var PhishingDetectionList */
+	private $list ;
```
st3iny, line 36:
```suggestion
	private PhishingDetectionList $list;
```

Let's use modern PHP

st3iny (dismissed):

lib/Address.php
hamza221: Some clients allow customising sender address such as ThunderBird, In that case Personal would be 
`custom label <custom.address@example.com>`  and from `$email` would still contain the original email.

st3iny: Ahh, now I get it. Thanks.

Code makes sense: https://3v4l.org/rAY8u

lib/Service/PhishingDetection/PhishingDetectionService.php
```diff
@@ -0,0 +1,58 @@
...
+class PhishingDetectionService {
+
+	private PhishingDetectionList $list ;
+
+	public function __construct(private ContactCheck $contactCheck, private CustomEmailCheck $customEmailCheck, private DateCheck $dateCheck, private ReplyToCheck $replyToCheck) {
+		$this->contactCheck = $contactCheck;
+		$this->customEmailCheck = $customEmailCheck;
+		$this->dateCheck = $dateCheck;
+		$this->replyToCheck = $replyToCheck;
+		$this->list = new PhishingDetectionList();
```
ChristophWurst, line 41: This looks like a temporary variable. Move it into `checkHeadersForPhishing` or anywhere else it's used.

The current way introduces side effects and if you call `checkHeadersForPhishing` twice you get outdated results too from the previous invocation.

lib/Service/PhishingDetection/ContactCheck.php
```diff
@@ -0,0 +1,62 @@
...
+			foreach ($contact['email'] as $contactEmail) {
+				$emailInContacts = true;
+				$emails .= $contactEmail.",";
+				if ($contactEmail === $email) {
+					return new PhishingDetectionResult(PhishingDetectionResult::CONTACTS_CHECK, false);
+				}
+			}
+		}
+		if ($emailInContacts) {
+			return new PhishingDetectionResult(PhishingDetectionResult::CONTACTS_CHECK, true, $this->l10n->t('Sender email: %1$s is not in the contacts list, but the sender name: %2$s is in the contacts list with the following emails: %3$s', [$email, $fn, $emails]));
```
ChristophWurst, line 56: *contacts list* -> address book

lib/Service/PhishingDetection/DateCheck.php
```diff
@@ -0,0 +1,48 @@
...
+
+class DateCheck {
+	protected IL10N $l10n;
+
+	public function __construct(IL10N $l10n) {
+		$this->l10n = $l10n;
+	}
+
+	public function run(string $date): PhishingDetectionResult {
+		$now = new DateTime();
```
ChristophWurst, line 40: inject the ITimeFactory and use getTime or similar to make this class fully testable

tests/Integration/Service/Phishing/PhishingDetectionServiceIntegrationTest.php
```diff
@@ -0,0 +1,91 @@
...
+		$this->contactCheck = new ContactCheck($this->contactsIntegration, $this->l10n);
+		$this->customEmailCheck = new CustomEmailCheck($this->l10n);
+		$this->dateCheck = new DateCheck($this->l10n);
+		$this->replyToCheck = new ReplyToCheck($this->l10n);
+		$this->service = new PhishingDetectionService($this->contactCheck, $this->customEmailCheck, $this->dateCheck, $this->replyToCheck);
+	}
+
+	
+
+	public function testContactCheck() {
```
ChristophWurst, line 61: nit: tests should have a void return type

lib/Service/PhishingDetection/DateCheck.php
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Mail\Service\PhishingDetection;
+
+use DateTime;
+use OCA\Mail\PhishingDetectionResult;
+use OCP\IL10N;
+
+class DateCheck {
```
ChristophWurst, line 32: this class has no tests ;)

ChristophWurst (commented):
Looks good

ChristophWurst:
> codecov/project — 43.34% (-6.87%) compared to c6e28fd 

could you check test coverage of new classes?

hamza221:
> > codecov/project — 43.34% (-6.87%) compared to [c6e28fd](https://github.com/nextcloud/mail/commit/c6e28fd59188ded88c3688a2e3d4988fd09df508)
> 
> could you check test coverage of new classes?

image

lib/Service/PhishingDetection/DateCheck.php
hamza221: it does 
https://github.com/nextcloud/mail/blob/feat/phishing-detection/tests/Integration/Service/Phishing/PhishingDetectionServiceIntegrationTest.php#L76

ChristophWurst:
Can we go to 100? :)

lib/PhishingDetectionList.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
kesselb, line 25:
```
/**
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 */
```

hamza221:
image

lib/IMAP/ImapMessageFetcher.php
```diff
@@ -495,7 +514,7 @@ private function decodeSubject(Horde_Imap_Client_Data_Envelope $envelope): strin
 
 	private function parseHeaders(Horde_Imap_Client_Data_Fetch $fetch): void {
 		/** @var resource $headersStream */
-		$headersStream = $fetch->getHeaderText('0', Horde_Imap_Client_Data_Fetch::HEADER_STREAM);
+		$headersStream = $fetch->getHeaderText('0', 10);
```
ChristophWurst, line 517: magic number?

ChristophWurst (approved):
:clap:

lib/IMAP/ImapMessageFetcher.php
hamza221: mb :'(, I can't remember why probably testing something

ChristophWurst (approved):
:shipit:

PR #9615, opened 2024-04-30
Fix:Search mailbox by address

hamza221:
Reproduce: 

1. open advanced search 
2. Search in from / To / CC/ BCC 
3.test both contacts and explicit mail addresses


GretaD (approved):
works

hamza221:
@ChristophWurst  any argument against multiple addresses per slot ?

PR #9658, opened 2024-05-20
chore(tests): Clean-up phpunit.unit.xml

joshtrichards:
- Fix spacing (it was bugging me)
- Switch up ordering
- Validate against correct schema version
- Remove options that are using default values (convert{Errors, Notices, Warnings}ToExceptions=true, cacheResult=true)
- Remove small/medium/large timeouts which aren't being used (and are all the same anyhow)
  - note: we're effectively using `defaultTimeLimit` globally currently (which defaults to `0`)


ChristophWurst (approved):
:broom:

Issue #8866, opened 2023-09-15
Print out email


lp-SRO:
### Is your feature request related to a problem? Please describe.

Is it not possible to print out the contents of an email? Even in today's times, this is still necessary from time to time.

### Describe the solution you'd like

You should add a printer symbol at the top of an email. For example, in front of the Symbohl wastebasket. You can then use this symbol to print an email.

### Additional context

Thank you for your work :-)

ChristophWurst:
Have you tried printing the page? There are special print stylesheets in place to make the contents of an email printable.

lp-SRO:
Where can I find these?

ChristophWurst:
Ctrl+P

lp-SRO:
Yes, but then only what you can see will be printed. If the content is longer and it becomes two pages, it won't print it.
That doesn't make any sense at all. Basically it just prints out a screenshot.

BiglifeMatt:
Print preview would certainly be desired by my end users.

handkerchief333:
There is no need to discuss whether this is a useful feature. In fact, it is an extremely useful and necessary feature. So much so that you almost have to use a different e-mail program because it's missing. So +1
See #9301 for my 2 cents.

PR #9669, opened 2024-05-25
Enh/Improve email printing

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/8866

TODO:
- [x] Add printing button
- [x] Implement printing iframe
- [x] Add additional information and styling


GVodyanov:
Current Issue:
image

Longer emails are cut and do not display multiple pages

hamza221:
I think a Print preview would be the easiest way around either a pop up or a new page.
It would also solve the problem of big threads, Theoretically with this approach you would have to programatically  expand all of them one by one and then print. 
Gmail solves this problem by opening a new page with adequate design and triggering the print event instantly

GVodyanov:
> I think a Print preview would be the easiest way around either a pop up or a new page. It would also solve the problem of big threads, Theoretically with this approach you would have to programatically expand all of them one by one and then print. Gmail solves this problem by opening a new page with adequate design and triggering the print event instantly

@hamza221 I found a solution to be overlaying an `iframe` and inserting everything needed into it, tell me if you think it looks sane. I know it's hacky, but I believe anything at this point will be, and doing it like this doesn't require a new view.

GVodyanov:
image

Here is what it currently looks like, and yes, there are multiple pages, they don't get cut off

src/components/MenuEnvelope.vue
```diff
@@ -567,8 +568,45 @@ export default {
...
+			frame.style.position = 'absolute'
+			frame.style.top = '-1000000px'
+			document.body.appendChild(frame)
+			const frameDoc = frame.contentWindow ? frame.contentWindow : frame.contentDocument.document ? frame.contentDocument.document : frame.contentDocument
+			frameDoc.document.open()
+			frameDoc.document.write('<html><head><style>')
+			frameDoc.document.write(printStyles)
+			frameDoc.document.write('</style></head><body>')
+			frameDoc.document.write(document.querySelector('#mail-thread-header h2').outerHTML)
+			frameDoc.document.write(`<h4>${this.envelope.from[0].label}, ${this.envelope.from[0].email}</h4>`)
```
ChristophWurst: you have to sanitize user input before putting it into DOM. the label could contain HTML markup

GVodyanov: I added a method to do that, but tell me if we have a more standard way somewhere

ChristophWurst:
I am not confident with the way this works. Building HTML by hand is scary. Why do we need this separate page? Can't we use the existing page and adjust the layout for print? E.g. do not limit the height of messages in the thread when printing so that everything is included, hide unwanted elements, etc.

hamza221:
> I am not confident with the way this works. Building HTML by hand is scary. Why do we need this separate page? Can't we use the existing page and adjust the layout for print? E.g. do not limit the height of messages in the thread when printing so that everything is included, hide unwanted elements, etc.

We can also use the already existing IFrame but we can't include the sender's address...

PR #9697, opened 2024-06-03
Fix: query share attributes for filepicker

hamza221:


st3iny (approved):
Tested and works (together with https://github.com/nextcloud/mail/pull/9592)

Issue #9573, opened 2024-04-17
Increase the size of the top split of the horizontal split


GretaD:
### Is your feature request related to a problem? Please describe.

The part of mail list is too small and you cannot see more than a couple of messages, now it is approx. 40%, which is too few.

### Describe the solution you'd like

Make the mail list part wider

### Additional context

Ref #9462

PR #9700, opened 2024-06-03
feat: increaze the min and max width for horizontal view

GretaD:
fixes https://github.com/nextcloud/mail/issues/9573

Screenshot from 2024-06-03 19-14-35

i added random numbers @jancborchardt please test it and let me know how much the % should be

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -2,6 +2,8 @@
 	<AppContent pane-config-key="mail"
 		:layout="layoutMode"
 		:show-details="isThreadShown"
+		:list-min-width="horizontalListMinWidth"
+		:list-max-width="horizontalListMaxWidth"
 		@update:showDetails="hideMessage">
 		<template #list>
 			<div :class="{ list__wrapper: !showThread || !isMobile }">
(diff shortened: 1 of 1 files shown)


GretaD:
i added random numbers, @jancborchardt please test it and let me know how much the % should be

src/components/MailboxThread.vue
```diff
@@ -150,6 +152,18 @@ export default {
 		layoutMode() {
 			return this.$store.getters.getPreference('layout-mode', 'vertical-split')
 		},
+		horizontalListMinWidth() {
+			return this.layoutMode === 'horizontal-split' ? 60 : this.defaultListMinWidth
```
ShGKme, line 156: `defaultListMinWidth` is actually a `vertical-split` value. IMO, it can be a constant number instead of "default".

ShGKme:
One minor issue, `NcAppContent` doesn't take into account the `min` and `max` values until user tries to resize it manually. For example, if `min-max` are 40..60 but restored (or from the previous layout) is `20`, then it is `20`. And only when the user touches it, it becomes 40

ShGKme (approved):
All fine from implementation.

PR #9719, opened 2024-06-07
fix: change appnavigation to appnavigationcaption for the email account

### LINKED ISSUE #9756: Use NcNavigationCaption for account navigation element. Remove color circle

GretaD:
before
Screenshot from 2024-06-07 12-37-39
after
Screenshot from 2024-06-07 12-38-03
fixes #9756

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -21,103 +21,93 @@
   -->
 
 <template>
-	<AppNavigationItem v-if="visible"
-		:id="id"
-		:key="id"
-		:menu-open.sync="menuOpen"
-		:name="account.emailAddress"
-		:to="firstMailboxRoute"
-		:exact="true"
-		@update:menuOpen="onMenuToggle">
-		<template #icon>
-			<IconError v-if="account.error || isDisabled" :size="20" />
-			<IconBullet v-else-if="bulletColor" :size="16" :fill-color="bulletColor" />
-		</template>
-		<!-- Actions -->
-		<template #actions>
-			<template v-if="isDisabled">
-				<ActionText :name="t('mail', 'Provisioned account is disabled')">
-					<template #icon>
-						<IconInfo :size="20" />
-					</template>
-					{{ t('mail', 'Please login using a password to enable this account. The current session is using passwordless authentication, e.g. SSO or WebAuthn.') }}
-				</ActionText>
+	<Fragment>
+		<NcAppNavigationCaption v-if="visible"
+			:id="id"
+			:key="id"
+			:name="account.emailAddress"
+			@update:menuOpen="onMenuToggle">
+			<!-- Actions -->
+			<template #actions>
+				<template v-if="isDisabled">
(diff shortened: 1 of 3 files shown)


GretaD:
cc @marcoambrosini thats what you meant during the call, right?

src/components/NavigationAccount.vue
```diff
@@ -21,12 +21,11 @@
   -->
 
 <template>
-	<AppNavigationItem v-if="visible"
+	<NcAppNavigationCaption v-if="visible"
 		:id="id"
 		:key="id"
 		:menu-open.sync="menuOpen"
```
ChristophWurst, line 27: Does this prop exist? https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppNavigation?id=ncappnavigationcaption

src/components/NavigationAccount.vue
```diff
@@ -21,12 +21,11 @@
...
 		:name="account.emailAddress"
-		:to="firstMailboxRoute"
 		:exact="true"
```
ChristophWurst, line 29: Same. This is not a router element

ChristophWurst (commented):
:broom:

src/components/NavigationAccount.vue
```diff
@@ -21,12 +21,11 @@
...
 		@update:menuOpen="onMenuToggle">
 		<template #icon>
```
marcoambrosini, line 31: Also the bulletcolor component goes away

marcoambrosini (commented):
more 🧹

src/components/NavigationAccount.vue
marcoambrosini: Sorry wrong line, I mean in `template #icon`

ChristophWurst (approved):

src/components/NavigationAccount.vue
```diff
@@ -106,16 +99,15 @@
 		<template #extra>
```
GretaD, line 99: NcAppNavigationCaption doesnt have an extra slot, need to find a way how to make account settings work

GretaD: @marcoambrosini should i add an extra slot at appnavigataioncaption, or should we use a different component?

marcoambrosini: Maybe use a <Fragment> component @GretaD https://www.npmjs.com/package/vue-fragment
cc @ShGKme @susnux for opinions

ShGKme: Yes, fragment will work here for Vue 2. But we are usually using `vue-frag`: https://www.npmjs.com/package/vue-frag 

Mail already has it installed.

ShGKme (approved):
Tested, all fine.

Now it is not possible to open the inbox by clicking on the account name. I guess it is intended, because we cannot collapse it anyway and the inbox  link is always visible.

ShGKme:
(auto-merged after restarting failed CI job)

PR #9749, opened 2024-06-17
feat(ocs): notify of new messages and provide API endpoint to retrieve its contents

miaulalala:
To Do

- [x] GET Email as JSON structure
- [x] GET Email as RFC5322 raw email
- [x] GET Attachments via URL (single attachment only)
- [x] Event for new incoming messages
- [x] Tests

### DIFF
--- a/lib/Controller/MessageApiController.php
+++ b/lib/Controller/MessageApiController.php
@@ -1,0 +1,240 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\Mail\Controller;
+
+use OCA\Mail\Contracts\IDkimService;
+use OCA\Mail\Exception\ClientException;
+use OCA\Mail\Exception\ServiceException;
+use OCA\Mail\Exception\SmimeDecryptException;
+use OCA\Mail\Http\TrapError;
+use OCA\Mail\IMAP\IMAPClientFactory;
+use OCA\Mail\Model\SmimeData;
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\AliasesService;
+use OCA\Mail\Service\Attachment\AttachmentService;
+use OCA\Mail\Service\ItineraryService;
+use OCA\Mail\Service\MailManager;
+use OCA\Mail\Service\OutboxService;
+use OCA\Mail\Service\Search\MailSearch;
+use OCA\Mail\Service\TrustedSenderService;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\BruteForceProtection;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCSController;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Files\IMimeTypeDetector;
+use OCP\IRequest;
+use OCP\IURLGenerator;
+use Psr\Log\LoggerInterface;
+
+class MessageApiController extends OCSController {
+
+	private ?string $userId;
+
+	public function __construct(
+		string $appName,
+		?string $userId,
+		IRequest $request,
+		private AccountService $accountService,
+		private AliasesService $aliasesService,
+		private AttachmentService $attachmentService,
+		private OutboxService $outboxService,
+		private MailSearch $mailSearch,
+		private MailManager $mailManager,
+		private IMAPClientFactory $clientFactory,
+		private LoggerInterface $logger,
+		private ITimeFactory $time,
+		private IURLGenerator $urlGenerator,
+		private IMimeTypeDetector $mimeTypeDetector,
+		private IDkimService $dkimService,
+		private ItineraryService $itineraryService,
+		private TrustedSenderService $trustedSenderService,
+	) {
+		parent::__construct($appName, $request);
+		$this->userId = $userId;
+	}
+
+	/**
+	 * @param int $id
+	 * @return DataResponse
+	 */
+	#[BruteForceProtection('mailGetMessage')]
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function get(int $id): DataResponse {
+		if ($this->userId === null) {
+			return new DataResponse('Account not found.', Http::STATUS_NOT_FOUND);
+		}
+
+		try {
+			$message = $this->mailManager->getMessage($this->userId, $id);
+			$mailbox = $this->mailManager->getMailbox($this->userId, $message->getMailboxId());
+			$account = $this->accountService->find($this->userId, $mailbox->getAccountId());
+		} catch (ClientException | DoesNotExistException $e) {
+			$this->logger->error('Message, Account or Mailbox not found', ['exception' => $e->getMessage()]);
+			return new DataResponse('Account not found.', Http::STATUS_NOT_FOUND);
+		}
+
+		$loadBody = true;
+		$client = $this->clientFactory->getClient($account);
+		try {
+			$imapMessage = $this->mailManager->getImapMessage(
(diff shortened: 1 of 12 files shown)


miaulalala:
Sent via api:

image

lib/Controller/MessageApiController.php
```diff
@@ -0,0 +1,93 @@
...
+			return new DataResponse('Not found', Http::STATUS_NOT_FOUND);
+		} finally {
+			$client->logout();
+		}
+
+		$messageContent = new AttachmentDownloadResponse(
+			$source,
+			$message->getSubject() . '.eml',
+			'message/rfc822',
+		);
```
SebastianKrupinski, line 89: I would recommend returning the message as a Json object instead of a raw email string, this way the consuming service does not have to re-phrase the message in again to use/read its contents, like we do in the UI.

Something like this... (pulled from a UI request)

```
{
    "uid": 254,
    "messageId": "\u003C3f488a7111b4a63cf073ec0ec4f86ff9@ksamail.ca\u003E",
    "from": [
        {
            "label": "User One via Nextcloud",
            "email": "ncdev@ksamail.ca"
        }
    ],
    "to": [
        {
            "label": "User Two",
            "email": "user2@ksamail.ca"
        }
    ],
    "replyTo": [
        {
            "label": "User One",
            "email": "user1@ksamail.ca"
        }
    ],
    "cc": [],
    "bcc": [],
    "subject": "Invitation: Recurrance Test",
    "dateInt": 1718673419,
    "flags": {
        "seen": true,
        "flagged": false,
        "answered": false,
        "deleted": false,
        "draft": false,
        "forwarded": false,
        "hasAttachments": false,
        "mdnsent": false,
        "important": false
    },
    "hasHtmlBody": true,
    "dispositionNotificationTo": "",
    "hasDkimSignature": false,
    "unsubscribeUrl": null,
    "isOneClickUnsubscribe": false,
    "unsubscribeMailto": null,
    "body": "==== Email Body =====",
    "attachments": [
        {
            "id": "2",
            "messageId": 254,
            "fileName": "event.ics",
            "mime": "text\/calendar",
            "size": 1566,
            "cid": null,
            "disposition": "inline",
            "downloadUrl": "https:\/\/localhost\/apps\/mail\/api\/messages\/876\/attachment\/2",
            "mimeUrl": "\/apps\/theming\/img\/core\/filetypes\/text-calendar.svg?v=704c4934",
            "isImage": false,
            "isCalendarEvent": true
        }
    ],
    "accountId": 2,
    "mailboxId": 7,
    "databaseId": 876,
    "isSenderTrusted": false,
    "smime": {
        "isSigned": false,
        "signatureIsValid": null,
        "isEncrypted": false
    }
}
```

SebastianKrupinski (commented):
Other than the one recommendation, code works.

lib/Controller/MessageApiController.php
miaulalala: Thanks for the suggestion - I think a combo of both parsed data and the raw email might be beneficial so that the workflow engine can proceed with whatever data they think is best. Good tip!

SebastianKrupinski: You're welcome.

You could add a raw flag to the get request, that way the consuming server can choose...

```
public function get(int $id, bool $raw = false) {

}
```

Or  something like this....

```
public function get(int $id, string $format = 'json') {

}
```

miaulalala: I think it might be better to split the get into two separate endpoints even

miaulalala:
@ChristophWurst shall I also expose an endpoint to download the attachments? Or is there a way to add them to the JSON without doing the base64_enocde dance? We have a download in the MessageController already that I could copy.

SebastianKrupinski (approved):
Tested works good.

The Json message object just didn't have the body parameter, but all other information was there.

SebastianKrupinski:
> @ChristophWurst shall I also expose an endpoint to download the attachments? Or is there a way to add them to the JSON without doing the base64_enocde dance? We have a download in the MessageController already that I could copy.

@miaulalala i wouldn't include attachments in the get method output... That can cause a lot of traffic for something that might not be used... on that note the attachments are already in the raw message, so I don't know if the raw message should be included unless requested that way...

This was the encoded attachment in the message I pulled.
```
inline; name=event.ics; filename=event.ics\r\n\r\nQkVHSU46VkNBTEVOREFSDQpWRVJTSU9OOjIuMA0KUFJPRElEOi0vL1NhYnJlLy9TYWJyZSBWT2Jq\r\nZWN0IDQuNS40Ly9FTg0KQ0FMU0NBTEU6R1JFR09SSUFODQpNRVRIT0Q6Q0FOQ0VMDQpCRUdJTjpW\r\nRVZFTlQNClVJRDpkZWM2NzQzNS1iN2QwLTQ0YTMtODQxMS0zOTFjYTZmNGJiY2ENCkRUU1RBTVA6\r\nMjAyNDA2MjVUMjE1ODA2Wg0KU0VRVUVOQ0U6Mw0KU1VNTUFSWTpUZXN0aW5nIFNpbmdsZXRvbiBB\r\nbGwgRGF5IEV2ZW50DQpEVFNUQVJUO1ZBTFVFPURBVEU6MjAyNDA3MDMNCkRURU5EO1ZBTFVFPURB\r\nVEU6MjAyNDA3MDQNCk9SR0FOSVpFUjtDTj1Vc2VyIE9uZTptYWlsdG86dXNlcjFAa3NhbWFpbC5j\r\nYQ0KQVRURU5ERUU7Q049VXNlciBUd286bWFpbHRvOnVzZXIyQGtzYW1haWwuY2ENCkVORDpWRVZF\r\nTlQNCkVORDpWQ0FMRU5EQVINCg==\r\n--SM8uFlts--\r\n\r\n"
```

miaulalala:
JSON response:


```json
{
    "ocs": {
        "meta": {
            "status": "ok",
            "statuscode": 200,
            "message": "OK"
        },
        "data": {
            "uid": 5396,
            "messageId": "<kjsdflkdsjflksdkjf@geopod-ismtpd-39>",
            "from": [
                {
                    "label": "FromLabel",
                    "email": "from@teefury.com"
                }
            ],
            "to": [
                {
                    "label": "fhfhf@gmx.net",
                    "email": "fhfhf@gmx.net"
                }
            ],
            "replyTo": [
                {
                    "label": "ReplyToLabel",
                    "email": "email@replyto.com"
                }
            ],
            "cc": [],
            "bcc": [],
            "subject": "Subject",
            "dateInt": 1719374564,
            "flags": {
                "seen": true,
                "flagged": false,
                "answered": false,
                "deleted": false,
                "draft": false,
                "forwarded": false,
                "hasAttachments": false,
                "mdnsent": false,
                "important": true
            },
            "hasHtmlBody": true,
            "dispositionNotificationTo": "",
            "hasDkimSignature": true,
            "unsubscribeUrl": "sjfifjsdlksdjflksdjf",
            "isOneClickUnsubscribe": true,
            "unsubscribeMailto": null,
            "scheduling": [],
            "body": "Body data here",
            "attachments": [],
            "id": 186737,
            "isSenderTrusted": false,
            "smime": {
                "isSigned": false,
                "signatureIsValid": null,
                "isEncrypted": false
            }
        }
    }
}
```

st3iny:
When an S/MIME encrypted message is fetched via `MessageApiController::get()` and the decryption fails (due to a missing key), a 404 response is returned with an empty message.

This could be improved by catching, retrying without fetching the body and adding a field `bodyRawUrl` field to the data response pointing to `getRaw()`. The HTTP status code should be `206 Partial Content`.

We cannot get the body but the meta information of a message (to, cc, subject etc.) could still be interesting to an API consumer.

https://github.com/nextcloud/mail/blob/d7989048871d8d06e207178914cc216b0d6d5b69/lib/Controller/MessageApiController.php#L81-L92

Fetching with `$loadBody = false` in line 87 will return the IMAP message without the body but will not fail on missing keys.

---

The thrown exception in `SmimeService::decryptDataFetch()` has to adjusted accordingly. Currently, a `ServiceException` is thrown which cannot be distinguished easily from the the decryption error.

https://github.com/nextcloud/mail/blob/fefd2a77b8f7d098fe565b17d8ee58e345f4cb4d/lib/Service/SmimeService.php#L524-L526

st3iny (approved):
S/MIME is handled correctly now.

hamza221 (approved):

PR #9778, opened 2024-06-26
fix: declare all properties in unit tests

miaulalala:


tests/Unit/Service/AntiSpamServiceTest.php
```diff
@@ -34,6 +34,8 @@ class AntiSpamServiceTest extends TestCase {
 	private SmtpClientFactory|MockObject $smtpClientFactory;
 	private MockObject|ImapMessageMapper $imapMessageMapper;
 	private LoggerInterface|MockObject $logger;
+	private MockObject|IMailTransmission $transmission;
```
ChristophWurst, line 37:
```suggestion
	private IMailTransmission|MockObject $transmission;
```

for consistency and readability

ChristophWurst (approved):

tests/Unit/Service/AntiSpamServiceTest.php
ChristophWurst: not a blocker. we can merge regardless

PR #9798, opened 2024-07-04
Remove instance name from share link add button

Jerome-Herbinet:
Just check the commit. IMO, the variable (which seems to incorporate the instance name) is unnecessary and makes the working unnecessarily long.


jancborchardt (approved):
Sounds good, will also most likely prevent the line break, and we use the same wording right above "Add attachment from Files".

PR #9806, opened 2024-07-05
fix: close the ncselect dropdown when clicked somewhere else

GretaD:
noticed while testing #9801 
when i select a 'to' and i want to add a 'cc' the dropdown doesnt close

Screenshot from 2024-07-05 14-34-31


GVodyanov (approved):
Cool

PR #9824, opened 2024-07-10
feat: add recipient info on the right side of the composer

### LINKED ISSUE #9622: Recipient info next to mail composer

When you write a mail to someone or reply, you only have the mail itself as context. You don’t see anything about the person itself except maybe if they have a signature.

We could show their contact info next to the composer. Maybe you have some notes in their contact which are relevant for that, or their company name and position. I remember some mail app did this but can not recall which.

Here is a quick mockup for the contact info showing on the right side of the composer.
- If there is only one person, just show their info right away
- For multiple people, we could collapse it to just the avatar, main line with name and subline with title, and make it expandable
- Data can come from Profile and/or Contacts (see below)

Mail composer contacts info

In Talk for example we will likely have to use the Profile info instead of the Contacts info. So to keep things consistent, maybe it’s better to use the same for both.
Although for Mail we need to have Contacts as a fallback since otherwise this feature would not work for external people to your organization, nor for single-user instances.

@nickvergessen said we should have a unified endpoint as an OCS API for this, so it would probably best for you to coordinate there @ChristophWurst.

### Additional context

Very similar issue in Talk as well for showing additional info of a person when starting a chat: https://github.com/nextcloud/spreed/issues/6293

GretaD:
fixes #9622
requires: https://github.com/nextcloud/contacts/pull/4194
to do

- [x]  change the classes to BEM styling.

PR description:
When a user is added to the "To" field in the mail composer, a right pane displays the contact details from their profile. Adding a second recipient collapses the list of contact details, with an option to toggle the visibility of the pane to view the data. Clicking the "maximize" button removes the right pane entirely to focus on the composer. Reducing the composer size restores the right pane, making the contact details visible again.

Single recipient
Screenshot from 2024-11-26 12-33-46
2 recipients collapsed view
Screenshot from 2024-11-26 12-33-56
2 reicpients, 1 expanded
Screenshot from 2024-11-26 12-34-10

### DIFF
--- a/src/components/RecipientInfo.vue
+++ b/src/components/RecipientInfo.vue
@@ -1,0 +1,151 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+<template>
+	<div class="recipient-info">
+		<!-- For a single recipient -->
+		<div v-if="recipients.length === 1" class="recipient-info__single">
+			<div class="recipient-info__header">
+				<div class="recipient-info__avatar">
+					<Avatar :display-name="recipients[0].label"
+						:email="recipients[0].email"
+						:size="55"
+						:disable-tooltip="true"
+						:disable-menu="true" />
+				</div>
+				<div class="recipient-info__details">
+					<DisplayContactDetails :email="recipients[0].email" />
+				</div>
+			</div>
+		</div>
+
+		<!-- For multiple recipients -->
+		<div v-else class="recipient-info__multiple">
+			<div v-for="(recipient, index) in recipients" :key="recipient.email" class="recipient-info__item">
+				<div class="recipient-info__header">
+					<div class="recipient-info__avatar">
+						<Avatar :display-name="recipient.label"
+							:email="recipient.email"
+							:size="55"
+							:disable-tooltip="true"
+							:disable-menu="true" />
+					</div>
+					<div v-if="!expandedRecipients[index]" class="recipient-info__name">
+						<h6>{{ recipient.email }}</h6>
+					</div>
+					<div class="recipient-info__expand-toggle" @click="toggleExpand(index)">
+						<template v-if="isExpanded(index)">
+							<div class="recipient-info__show-less">
+								<IconArrowUp :size="16" />
+								<span>{{ t('mail', 'Show less') }}</span>
+							</div>
+						</template>
+						<template v-else>
+							<IconArrowDown :size="16" />
+							<span>{{ t('mail', 'Show more') }}</span>
+						</template>
+					</div>
+				</div>
+				<div v-show="expandedRecipients[index]" class="recipient-info__details">
+					<DisplayContactDetails :email="recipient.email" />
+				</div>
+			</div>
+		</div>
+	</div>
+</template>
+
+<script>
+import { mapGetters } from 'vuex'
+import IconArrowDown from 'vue-material-design-icons/ArrowDown.vue'
+import IconArrowUp from 'vue-material-design-icons/ArrowUp.vue'
+import Avatar from './Avatar.vue'
+import DisplayContactDetails from './DisplayContactDetails.vue'
+
+export default {
+	components: {
+		Avatar,
+		IconArrowDown,
+		IconArrowUp,
+		DisplayContactDetails,
+	},
+	data() {
+		return {
+			expandedRecipients: [],
+		}
+	},
+	computed: {
+		...mapGetters(['composerMessage']),
+		recipients() {
+			return this.composerMessage.data.to
+		},
+	},
+	watch: {
+		recipients: {
+			immediate: true,
+			handler() {
+				this.expandedRecipients = this.recipients.map(() => false)
+			},
+		},
+	},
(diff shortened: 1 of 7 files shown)


GretaD:
Hi @ShGKme we need your vue expertise to decide how to move on here.
Screenshot from 2024-07-22 12-42-31

we would like to find a way how to properly show info on the right pane, that should look more or less like the contact one (see mockup on the initial ticket). Duplicating code is not the right way to go for obvious reasons. Do you have any idea what would be best to do here? This feature has to go in on Thursday.

ShGKme:
Hi @GretaD 

> This feature has to go in on Thursday.

Having this, I'd start with duplicating the code. Then you can do it faster. And also see what will be different, what customization or interface will be required.

Then, when it works, we should move it to a library. Either `@nextcloud/vue` or a new one.

Also, there could be two ways:
1. Reuse only UI components, but not the login with VCard queries
2. Reuse everything, like `NcTeam*` components do in `@nextcloud/vue`

I'd go with the second one. Not sure this component could be needed as a pure UI component.

ChristophWurst:
This change needs a counterpart in Contacts. Where can we find that change?

GretaD:
> This change needs a counterpart in Contacts. Where can we find that change?

the description was not updated recently, sorry about that, i added the info of the current state.

GVodyanov (changes requested):
This looks really cool ✨! I found a small issue though:

Screenshot from 2024-11-24 19-49-46
Screenshot from 2024-11-24 19-47-55

I only see the email address in the right pane when having two recipients, no error in console

GretaD:
> I only see the email address in the right pane when having two recipients, no error in console

thanks so much for reviewing. i will wait for a second reviewer and deliver all in one fixup, <3

src/components/RecipientInfo.vue
```diff
@@ -0,0 +1,156 @@
...
+			<div v-for="(recipient, index) in recipients" :key="index" class="recipient-item">
+				<div class="recipient-header">
+					<div class="recipient-avatar">
+						<Avatar :display-name="recipient.label"
+							:email="recipient.email"
+							:size="55"
+							:disable-tooltip="true"
+							:disable-menu="true" />
+					</div>
+					<div v-if="!expandedRecipients[index]" class="recipient-name">
```
GVodyanov, line 31: Why is this section not present for a single recipient?

src/components/RecipientInfo.vue
```diff
@@ -0,0 +1,156 @@
...
+						<h6>{{ recipient.email }}</h6>
+					</div>
+					<div class="expand-toggle" @click="toggleExpand(index)">
+						<template v-if="isExpanded(index)">
+							<div class="show-less">
+							<IconArrowUp :size="16" />
```
GVodyanov, line 37: Indentation

kesselb:
- I'm not a fan of showing the sidebar when entering an email address / selecting a contact because it shifts the composer modal. What do you think about always showing the sidebar?
- It's possible to select a contact group as recipient. This will also trigger the xhr request to find the contact details. Could we add a check if recipient starts with "nextcloud:" then don't search the contact details? A possible enhancement for a follow-up could be to list the members of the group.

kesselb:
- [ ] Something weird happens when deleting a recipient with the contact picture:

[Screencast From 2024-12-03 15-02-58.webm](https://github.com/user-attachments/assets/4590d1fa-6db5-4a5d-a256-d6a942bab542)

src/store/getters.js
```diff
@@ -159,5 +159,4 @@ export const getters = {
 	isFollowUpFeatureAvailable: (state) => state.followUpFeatureAvailable,
 	getInternalAddresses: (state) => state.internalAddress?.filter(internalAddress => internalAddress !== undefined),
 	hasCurrentUserPrincipalAndCollections: (state) => state.hasCurrentUserPrincipalAndCollections,
-	showSettingsForAccount: (state) => (accountId) => state.showAccountSettings === accountId,
```
kesselb, line 162: Please restore the line because it's still needed ;)

GretaD: thats a mistake of the rebase not on purpose of course :(

src/components/RecipientInfo.vue
GretaD: you mean the v-if? That is because when its one recipient only, it shouldnt be collapsed. It collapses when a second one is added.

SebastianKrupinski (approved):
Tested. Works.

The only thing I would add is the ability to minimize the side pane. It takes up a lot of room and makes the email message area very small.

image

GretaD:
> Tested. Works.
> 
> The only thing I would add is the ability to minimize the side pane. It takes up a lot of room and makes the email message area very small.
> 

its there, if you click the minimaze/maximize button on top right :) you can remove it and continue with your composer only view

SebastianKrupinski:
> its there, if you click the minimaze/maximize button on top right :) you can remove it and continue with your composer only view

Ah. Okay, That might be a bit confusing. I thought that was to minimize the entire modal. Would be better to put a > icon in the top left corner of the pane. But that is up to design.

st3iny (approved):
Tested and works. The empty content in case there is no data available is not perfectly centered, but I don't see this as a blocker.

image

Issue #9814, opened 2024-07-08
Don't log AttachmentMissingError as an error 


kesselb:
### Is your feature request related to a problem? Please describe.

The missing attachment warning is logged as an error and, if configured, forwarded to logging services like sentry.

How to reproduce: 
- Open browser console
- Write an email and use the word attachment
- Send
- See the error 

### Describe the solution you'd like

Log the message with a lower log level because it's not an actual error.

PR #9825, opened 2024-07-10
feat: Log AttachmentMissingError as a warning

PeterJGDe:
fixes: #9814

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -414,26 +414,28 @@ export default {
 				await this.$store.dispatch('stopComposerSession')
 				this.$emit('close')
 			} catch (error) {
-				logger.error('could not send message', { error })
 				this.error = await matchError(error, {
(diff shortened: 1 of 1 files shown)


kesselb (commented):
Thank you :+1: 

Works as expected.

It should be fine to lower the loglevel from warning to info and adjust the wording a bit. That warning is nothing an admin would care about. 

Please add `logger.error('could not send message', { error })` to

https://github.com/nextcloud/mail/blob/c1d14c4fa32f8d66543db9dc2402d5914127d7d8/src/components/NewMessageModal.vue#L426-L428

Please add `logger.warn('could not send message', { error })` to

https://github.com/nextcloud/mail/blob/c1d14c4fa32f8d66543db9dc2402d5914127d7d8/src/components/NewMessageModal.vue#L435-L437

Without `logger.error` and `logger.warn` in the default case errors from the backend, without a mapping in match error, are just ignored.

kesselb (approved):

Issue #9823, opened 2024-07-10
In the composer, change the icon for toggling the formatting bar to format color text icon


GretaD:
In the composer, change the icon for toggling the formatting bar to [format color text icon](https://fonts.google.com/icons?icon.style=Filled&icon.set=Material+Icons&icon.query=format&selected=Material+Icons:format_color_text:&icon.size=24&icon.color=%23222222)

PR #9826, opened 2024-07-10
fix: change format button icon

PeterJGDe:
fixes: #9823

Before: 
image

After:
image


GretaD:
we used the Tt icon, because underlined A, doesnt exist on the material design.

Issue #9725, opened 2024-06-10
Add JSON format to `mail:account:export`


WilliamDEdwards:
### Is your feature request related to a problem? Please describe.

`php occ mail:account:delete` takes a `user-id`. This user ID can only be retrieved with `php occ mail:account:export`, but it doesn't support JSON output, which makes automatically deleting mail accounts nearly impossible.

PR #9828, opened 2024-07-10
feat: add JSON to occ mail export

PeterJGDe:
fixes: #9725

### DIFF
--- a/lib/Command/ExportAccount.php
+++ b/lib/Command/ExportAccount.php
@@ -12,10 +12,12 @@ use OCP\Security\ICrypto;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class ExportAccount extends Command {
 	public const ARGUMENT_USER_ID = 'user-id';
+	public const ARGUMENT_OUTPUT_FORMAT = 'output';
 
 	private AccountService $accountService;
 	private ICrypto $crypto;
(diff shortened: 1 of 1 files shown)


GretaD:
we tested it, and it works as expected, a php dev should have a look just in case :)

kesselb:
Works ;)

We usually use something like `occ mail:account:export --output=json` to tell a command to provide a json output.

Commands in the server extending from `\OC\Command\Base` provide the output argument by default: https://github.com/nextcloud/server/blob/ead87dfa01282698acafa9600e5e4a52bf306e33/core/Command/Base.php#L29-L37.

I'm unsure if we want to extend from \OC\Command\Base in mail because it's strictly speaking a private component in server. 

Fine by me to merge the pr in the current state but would recommend

- To merge the JSON command into the other one.
- Make the JSON output all lowercase and avoid whitespaces (e.g. "Account ID" => "account_id") to make it easier to consume. I'm not aware that we have a strict rule here, just best practice.

lib/Command/ExportAccount.php
```diff
@@ -34,22 +36,58 @@ protected function configure() {
...
+						'host' => $account->getMailAccount()->getOutboundHost(),
+						'port' => $account->getMailAccount()->getOutboundPort(),
+						'security' => $account->getMailAccount()->getOutboundSslMode()
+					]
+				];
+			}
+		
+			return $accountsData;
+		}
+		if ($input->getOption(self::ARGUMENT_OUTPUT_FORMAT) == 'json') {
```
ChristophWurst, line 76:
```suggestion
		if ($input->getOption(self::ARGUMENT_OUTPUT_FORMAT) === 'json') {
```

same for the other comparisons. we always use strict comparison in nextcloud. it will also make sure that we are comparing two strings :)

PeterJGDe:
@kesselb I removed `\OC\Command\Base` and merged i with `--output` into `mail:account:export`. What do you think about it?

kesselb:
Yep, not using the server component is probably better :+1:

lib/Command/ExportAccount.php
```diff
@@ -34,22 +36,58 @@ protected function configure() {
...
+					]
+				];
+			}
+		
+			return $accountsData;
+		}
+		if ($input->getOption(self::ARGUMENT_OUTPUT_FORMAT) === 'json') {
+			$output->writeln(json_encode(getAccountsData($accounts)));
+		} elseif ($input->getOption(self::ARGUMENT_OUTPUT_FORMAT) === 'json_pretty') {
+			$output->writeln(json_encode(getAccountsData($accounts, 'json_pretty')));
```
kesselb, line 79:
```suggestion
			$output->writeln(json_encode(getAccountsData($accounts), JSON_PRETTY_PRINT));
```

lib/Command/ExportAccount.php
```diff
@@ -34,22 +36,58 @@ protected function configure() {
...
-		foreach ($accounts as $account) {
-			$output->writeln("<info>Account " . $account->getId() . ":</info>");
-			$output->writeln("- E-Mail: " . $account->getEmail());
-			$output->writeln("- Name: " . $account->getName());
-			$output->writeln("- Provision: " . ($account->getMailAccount()->getProvisioningId() ? "set" : "none"). " ID: " . ($account->getMailAccount()->getProvisioningId() ? $account->getMailAccount()->getProvisioningId():"N/A"));
-			$output->writeln("- IMAP user: " . $account->getMailAccount()->getInboundUser());
-			$output->writeln("- IMAP host: " . $account->getMailAccount()->getInboundHost() . ":" . $account->getMailAccount()->getInboundPort() . ", security: " . $account->getMailAccount()->getInboundSslMode());
-			$output->writeln("- SMTP user: " . $account->getMailAccount()->getOutboundUser());
-			$output->writeln("- SMTP host: " . $account->getMailAccount()->getOutboundHost() . ":" . $account->getMailAccount()->getOutboundPort() . ", security: " . $account->getMailAccount()->getOutboundSslMode());
+		function getAccountsData($accounts) {
```
kesselb, line 47: Please make getAccountsData a class method. 

A class method would be on the same level as `protected function execute`.

For example, `private function getAccountsData($accounts) { ... }`

Watch out, you also need to change the calling code:

 `getAccountsData($accounts)`  => `$this->getAccountsData($accounts)`

kesselb (approved):

PR #9672, opened 2024-05-27
feat(ocs): send a message via api

miaulalala:
[Mühselig ernährt sich das Eichhörnchen](https://english.stackexchange.com/questions/347961/english-idiom-for-the-squirrel-feeds-itself-arduously)
https://github.com/nextcloud/mail/issues/9450

Sent via api:

image

PR #9832, opened 2024-07-10
enh(ocs): add OCS extractor and workflow

miaulalala:
Needed for https://github.com/nextcloud/mail/pull/9749 and https://github.com/nextcloud/mail/pull/9672


.github/workflows/openapi.yml
```diff
@@ -0,0 +1,94 @@
...
+          CYPRESS_INSTALL_BINARY: 0
+          PUPPETEER_SKIP_DOWNLOAD: true
+        run: |
+          npm ci
+
+      - name: Set up dependencies
+        run: composer i
+
+      - name: Regenerate OpenAPI
+        run: composer run openapi
```
provokateurin, line 83: Need to add this script to composer.json to make it work (`generate-spec`)

composer.json
```diff
@@ -74,7 +74,8 @@
 		"test:integration": "phpunit -c tests/phpunit.integration.xml --fail-on-warning",
 		"test:integration:dev": "phpunit -c tests/phpunit.integration.xml --no-coverage --order-by=defects --stop-on-defect --fail-on-warning --stop-on-error --stop-on-failure",
 		"test:unit": "phpunit -c tests/phpunit.unit.xml --fail-on-warning",
-		"test:unit:dev": "phpunit -c tests/phpunit.unit.xml --no-coverage --order-by=defects --stop-on-defect --fail-on-warning --stop-on-error --stop-on-failure"
+		"test:unit:dev": "phpunit -c tests/phpunit.unit.xml --no-coverage --order-by=defects --stop-on-defect --fail-on-warning --stop-on-error --stop-on-failure",
+		"openapi": "php vendor/bin/generate-spec"
```
provokateurin, line 78:
```suggestion
		"openapi": "generate-spec"
```
This is enough, it will be resolved automatically

provokateurin (approved):

composer.json
miaulalala: nope, it doesn't for me locally at least

miaulalala: I think it's because it's installed via vendor-bin

provokateurin: Talk also has it in vendor-bin and there it works https://github.com/nextcloud/spreed/blob/22e1ff018fe3f96a40e60e292f5bf4b09967e754/composer.json#L22 :thinking:

miaulalala: does it matter? It works.

ChristophWurst: Tested and it works as @provokateurin suggested. The bin plugin puts "binaries" of vendor-bin/*/bin to the global vendor/bin/ so that they can be called as if they were installed globally.

ChristophWurst (changes requested):
1. Commit check in the workflow is not necessary. We don't want to vendor the openapi output in git.
2. Adjust https://github.com/nextcloud/mail/blob/b9bdff0b88baf3b8f211059a07a129b4fd3ca8cc/krankerl.toml#L5-L7 to generate the opnapi files before packaging.
3. The new vendor bin should be monitored by renovate bot. Add it to https://github.com/nextcloud/mail/blob/b9bdff0b88baf3b8f211059a07a129b4fd3ca8cc/renovate.json.

provokateurin:
The check is necessary to detect errors, even if you don't want to commit the result. Otherwise you'll only learn of the problems once you try to do a release...

ChristophWurst:
@provokateurin I mean "Check openapi*.json and typescript changes" to go. "Regenerate OpenAPI" has to stay of course.

ChristophWurst (commented):
The new vendor bin should be monitored by renovate bot. Add it to https://github.com/nextcloud/mail/blob/b9bdff0b88baf3b8f211059a07a129b4fd3ca8cc/renovate.json.

ChristophWurst (approved):

ChristophWurst:
Squash harder

miaulalala:
> Squash harder

image

ChristophWurst (approved):

ChristophWurst (changes requested):
We should only need `nextcloud/openapi-extractor` in the bin, not the global composer.* files

composer.json
```diff
@@ -88,5 +89,11 @@
 				"gravatarphp/gravatar"
 			]
 		}
-	}
+	},
+	"repositories": [
```
ChristophWurst, line 93: :fire:

PR #9848, opened 2024-07-12
fix: remove the global styling from the composer list

### LINKED ISSUE #9608: ordered & unordered list formatting does not work
### Steps to reproduce

since about a week I noticed that ordered & unordered list formatting does not work on cloud.nextcloud.com mail webinterface. Other formatting like bold, quotation works fine.

324926716-2cf37fd7-a19d-4544-bdde-4c1f17a4602e

### Expected behavior

list formatting works

### Actual behavior

list formatting doesn't work

### Mail app version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes #9608

`padding: 0` from global styling, kills the ckeditor `padding-inline-start: 40px;` on our app

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -380,6 +380,10 @@ https://github.com/ckeditor/ckeditor5/issues/1142
  .ck .ck-reset {
 	background: var(--color-main-background) !important;
  }
+/* Default ckeditor value of padding-inline-start, to overwrite the global styling from server */
+.ck-content ul, .ck-content ol {
+	padding-inline-start: 40px;
+}
 .ck-list__item {
 	.ck-off {
 		background:var(--color-main-background) !important;


src/components/TextEditor.vue
```diff
@@ -380,6 +380,9 @@ https://github.com/ckeditor/ckeditor5/issues/1142
  .ck .ck-reset {
 	background: var(--color-main-background) !important;
  }
+.ck-content ul, .ck-content ol {
+	padding-inline-start: 40px;
```
ChristophWurst, line 384: Where does the magic number come from? Is that the CKEditor default?

GretaD: > Where does the magic number come from? Is that the CKEditor default?

yes

ChristophWurst: then leave this as code comment please

ChristophWurst (approved):
Except the missing code comment

PR #9852, opened 2024-07-15
QoL improvements to email printing

GVodyanov:
Fixes https://github.com/nextcloud/mail/issues/8866

Improve upon https://github.com/nextcloud/mail/pull/9818

TODO:
- [x] Add a print button in actions menu
- [x] Remove reply button
- [x] Remove envelope border
- [x] Remove unnecessary action items

image

### DIFF
--- a/src/components/MenuEnvelope.vue
+++ b/src/components/MenuEnvelope.vue
@@ -164,6 +164,13 @@
 				</template>
 				{{ t('mail', 'View source') }}
 			</ActionButton>
+			<ActionButton :close-after-click="true"
+				@click="onPrint">
+				<template #icon>
+					<PrinterIcon :size="20" />
+				</template>
+				{{ t('mail', 'Print message') }}
+			</ActionButton>
 			<ActionLink :close-after-click="true"
 				:href="exportMessageLink">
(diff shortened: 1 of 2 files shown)


src/components/MenuEnvelope.vue
```diff
@@ -562,6 +571,12 @@ export default {
 		setCustomSnooze() {
 			this.onSnooze(this.customSnoozeDateTime.valueOf())
 		},
+		onPrint() {
+			// needed for the actions menu to actually close and not be shown in the print preview
+			setTimeout(() => {
+				window.print()
+			}, 10)
```
ChristophWurst, line 578: Could you use Vue's $nextTick callback?

GVodyanov: Doing
```js
this.$nextTick(() => {
	window.print()
})
```
Does not seem to work, is there a better way to do it?

ChristophWurst: The only thing I could imagine is adding a dedicated close-after-click event in the upstream component and listening to that instead of the click event

ChristophWurst (approved):

PR #9877, opened 2024-07-17
chore: update psalm/phar to 5.9.0

miaulalala:
For https://github.com/nextcloud/mail/pull/9832

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
 	},
 	"require-dev": {
 		"nextcloud/ocp": "dev-stable26",
-		"psalm/phar": "^5.25.0",
+		"psalm/phar": "^5.9.0",
 		"roave/security-advisories": "dev-master"
 	},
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -43,7 +43,7 @@
 	},
 	"require-dev": {
 		"nextcloud/ocp": "dev-stable26",
-		"psalm/phar": "^5.25.0",
+		"psalm/phar": "^5.9.0",
```
provokateurin, line 46: This is wrong, you already have good version of psalm. No need to downgrade it, anything >=5.9.0 is fine so 5.25.0 is fine too.

provokateurin (changes requested):

composer.json
miaulalala: ugh I'm so confused, of course 5.9 is lower than 5.25 :sweat:

PR #9889, opened 2024-07-19
fix: set link icon size explicitly

hamza221:
| b | a |
|--------|--------|
| image | image |

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -397,6 +397,10 @@ https://github.com/ckeditor/ckeditor5/issues/1142
  }
  .link-title{
 	color: var(--color-main-text) !important;
+	margin-left: var(--default-grid-baseline) !important;
+ }
+ .link-icon {
+	width: 16px !important;
  }
  .custom-item {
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
I think I saw a css var for icon size the other day. We could use that to be even more flexible

hamza221:
> I think I saw a css var for icon size the other day. We could use that to be even more flexible

Couldn't find it, only found ck editor's vars

src/components/TextEditor.vue
```diff
@@ -398,6 +398,9 @@ https://github.com/ckeditor/ckeditor5/issues/1142
  .link-title{
 	color: var(--color-main-text) !important;
  }
```
GVodyanov, line 400:
```suggestion
 .link-title{
	color: var(--color-main-text) !important;
	margin-left: var(--default-grid-baseline);
 }
```

GVodyanov: This looks a little nicer

GVodyanov (approved):

PR #9891, opened 2024-07-19
feat: add internal addresses

hamza221:


src/components/RecipientListItem.vue
```diff
@@ -29,6 +30,16 @@ export default {
...
 		},
 	},
+	data() {
+		return {
+			isInternal: true,
+		}
+	},
+	async onMounted() {
+		if (this.$store.getters.getPreference('internal-addresses', 'false') === 'true') {
+			this.isInternal = await isInternal(this.option.email)
```
ChristophWurst, line 40: I think we should push this to the server, e.g. to `\OCA\Mail\IMAP\PreviewEnhancer::process`. Otherwise there will be one request per recipient. That does not scale

src/components/ThreadEnvelope.vue
```diff
@@ -898,6 +903,10 @@ export default {
 		}
 	}
 
+	.external{
```
ChristophWurst, line 906:
```suggestion
	.external {
```

src/components/RecipientListItem.vue
ChristophWurst: Alternative: process on the front-end

The list of internal addresses is likely very short with 1-5 domains. Load the list with initial state and do the checks in the front-end.

hamza221: I'm utilising the store instead now

src/components/InternalAddress.vue
```diff
@@ -0,0 +1,160 @@
...
+
+<template>
+	<div>
+		<div v-for="domain in sortedDomains"
+			:key="domain.address">
+			{{ domain.address }}
+			{{ senderType(domain.type) }}
+			<ButtonVue type="tertiary"
+				class="button"
+				:aria-label="t('mail','Remove')"
```
ChristophWurst, line 14:
```suggestion
				:aria-label="t('mail', 'Remove')"
```

the usual styling nitpicks ;-)

src/components/InternalAddress.vue
```diff
@@ -0,0 +1,160 @@
...
+				@click="removeInternalAddress(domain)">
+				{{ t('mail','Remove') }}
+			</ButtonVue>
+		</div>
+		<div v-for="email in sortedEmails"
+			:key="email.address">
+			{{ email.address }}
+			{{ t('mail', 'domain') }}
```
ChristophWurst, line 22: I'm not a designer but the `user@domain.tld domain [Remove]` line looks a bit wonky.

We could put the `domain` part in parenthesis and make it text max contrast.

Just an idea

src/components/InternalAddress.vue
```diff
@@ -0,0 +1,160 @@
...
+			<template #icon>
+				<IconAdd :size="20" />
+			</template>
+			{{ t('mail', 'Add internal address') }}
+		</ButtonVue>
+		<NcDialog :open.sync="openDialog"
+			:buttons="buttons"
+			:name="t('mail', 'Add internal address')"
+			@close="openDialog = false">
+			<NcTextField class="input" :label="t('mail', 'Add internal email or domain')" :value.sync="newAddress" />
```
ChristophWurst, line 41: a tiny bit of text would also help this dialog. you could show examples, like how to actually set a domain as internal. is it `domain.tld` or `@domain.tld` or `*@domain.tld` a user has to enter? :)

src/components/ThreadEnvelope.vue
```diff
@@ -898,6 +902,10 @@ export default {
 		}
 	}
 
+	.external {
```
ChristophWurst, line 905: could this be BEM modifier?

templates/index.php
```diff
@@ -16,4 +16,5 @@
 <input type="hidden" id="search-priority-body" value="<?php p($_['search-priority-body']); ?>">
 <input type="hidden" id="layout-mode" value="<?php p($_['layout-mode']); ?>">
 <input type="hidden" id="follow-up-reminders" value="<?php p($_['follow-up-reminders']); ?>">
+<input type="hidden" id="internal-addresses" value="<?php p($_['internal-addresses']); ?>">
```
ChristophWurst, line 19: use initial state pls

src/components/InternalAddress.vue
hamza221: all of them work
If you add name@domain.tld it will be saved as an address 
if you add @domain.tld it will be stripped into domain.tld 
if you add domain.tld it will be saved as a domain an addresses ares gonna be checked against it

lib/Controller/InternalAddressController.php
```diff
@@ -0,0 +1,82 @@
...
+			$address,
+			$type,
+			false
+		);
+
+		return JsonResponse::success();
+	}
+
+	/**
+	 * @NoAdminRequired
```
ChristophWurst, line 70: Use ``\OCP\AppFramework\Http\Attribute\NoAdminRequired``

lib/Controller/InternalAddressController.php
```diff
@@ -0,0 +1,82 @@
...
+	 *
+	 * @return JsonResponse
+	 */
+	#[TrapError]
+	public function list(): JsonResponse {
+		$list = $this->internalAddressService->getInternalAddresses(
+			$this->uid
```
ChristophWurst, line 77: Handle the null case and return a 401

lib/Controller/InternalAddressController.php
```diff
@@ -0,0 +1,82 @@
...
+	 * @NoAdminRequired
+	 *
+	 * @param string $address
+	 * @param string $type
+	 * @return JsonResponse
+	 */
+	#[TrapError]
+	public function removeAddress(string $address, string $type): JsonResponse {
+		$this->internalAddressService->add(
+			$this->uid,
```
ChristophWurst, line 60: 401 handling

lib/Controller/InternalAddressController.php
```diff
@@ -0,0 +1,82 @@
...
+use OCA\Mail\Service\InternalAddressService;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
+use OCP\IRequest;
+
+class InternalAddressController extends Controller {
+	private ?string $uid;
+
+	public function __construct(IRequest        $request,
+		?string                                 $UserId,
```
ChristophWurst, line 24:
```suggestion
		?string                                 $userId,
```

https://docs.nextcloud.com/server/latest/developer_manual/basics/dependency_injection.html#predefined-core-services

lib/Controller/PageController.php
```diff
@@ -112,6 +114,7 @@ public function __construct(string $appName,
 		$this->aiIntegrationsService = $aiIntegrationsService;
 		$this->userManager = $userManager;
 		$this->classificationSettingsService = $classificationSettingsService;
+		$this->internalAddressService = $internalAddressService;
```
ChristophWurst, line 117: property is not declared

lib/Db/InternalAddressMapper.php
```diff
@@ -0,0 +1,98 @@
...
+		$qb = $this->db->getQueryBuilder();
+
+		$insert = $qb->insert($this->getTableName())
+			->values([
+				'user_id' => $qb->createNamedParameter($uid),
+				'address' => $qb->createNamedParameter($address),
+				'type' => $qb->createNamedParameter($type),
+			]);
+
+		$insert->executeStatement();
```
ChristophWurst, line 60: the base class has an insert method. You might just be able to use it. the advantage is that it will also handle returning the ID of the inserted row

lib/Db/InternalAddressMapper.php
```diff
@@ -0,0 +1,98 @@
...
+
+	public function find(string $uid, string $address): ?InternalAddress {
+		$qb = $this->db->getQueryBuilder();
+		$select = $qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->eq('user_id', $qb->createNamedParameter($uid)),
+				$qb->expr()->eq('address', $qb->createNamedParameter($address))
+			);
+		return $this->findEntity($select);
```
ChristophWurst, line 96: findEntity throws when no row is found. this is not handled. you have to catch the exception and "translate" it to null

ChristophWurst (commented):
Looks good so far!

ChristophWurst (approved):

PR #9910, opened 2024-07-24
Fix attachment downloadUrl in OCS getMail endpoint

julien-nc:
It was always returning `/ocs/v2.php` because the `mail.messageApi.downloadAttachment` does not exist.


miaulalala (approved):
Thanks for the fix Julien!

Issue #9838, opened 2024-07-11
Unread style and early ellipsis for envelopes


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Look at unread messages

### Expected behavior

* Sender names show in bold

### Actual behavior

*Sender and subject show ellipsis too early

### Mail app version

4

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
This is probably related to https://github.com/nextcloud/mail/issues/9334. It's best tackled together.

GretaD:
The eclipsed issue comes from an unnecessary display flex for the subject class. I can push a fixup for this.

While the issue with the space when it has the unread indicator is because the unread indicator is inside the details/extra slot. See pic below. That can be probably fixed on nc/vue.
![Screenshot from 2024-07-24 10-46-41](https://github.com/user-attachments/assets/2d12a3e7-414c-4612-8998-1f740f0844ba)

ChristophWurst:
> That can be probably fixed on nc/vue.

Let's do this

GretaD:
> > That can be probably fixed on nc/vue.
> 
> Let's do this

sure, but the 2 issues are not connected. The display flex is added by us on the mail app that breaks the ellipsis.

ChristophWurst:
Fair enough

GVodyanov:
I'm stealing this

PR #9911, opened 2024-07-24
fix: ellips the subject and position of the important icon

GretaD:
this fixes the impotant icon position
before
Screenshot from 2024-07-24 12-12-26

after
Screenshot from 2024-07-24 12-33-32

and
the ellipsed subject reported here #9838 
Screenshot from 2024-07-24 12-34-14

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -903,8 +903,6 @@ export default {
 			text-overflow: ellipsis;
 			white-space: nowrap;
 			line-height: 130%;
-			display: flex;
-			align-items: center;
 		}
 	}
 	&__preview-text {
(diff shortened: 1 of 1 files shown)


GVodyanov (approved):
Cool!

src/components/Envelope.vue
```diff
@@ -903,7 +903,6 @@ export default {
 			text-overflow: ellipsis;
 			white-space: nowrap;
 			line-height: 130%;
-			display: flex;
 			align-items: center;
```
ChristophWurst, line 906:
```suggestion
```

https://developer.mozilla.org/en-US/docs/Web/CSS/align-items this is a flexbox prop. Make sure elements are still aligned the same way.

GretaD: yes, the elements are aligned in the center in all layouts

ChristophWurst: good. then let's just drop the dead prop

ChristophWurst (approved):

Issue #9915, opened 2024-07-25
Spacing issues in accounts setup/edit form


ChristophWurst:
### Steps to reproduce

1. Open the app on a fresh installation

### Expected behavior

A nicely designed account settings form

### Actual behavior

1. The (None SSL/TLS STARTTLS) switcher has rendering errors and the text is a bit too cramped
2.  The overall spacing between the options is not good. Not enough spacing between selection.

<img width="733" alt="b5" src="https://github.com/user-attachments/assets/0fda7c75-bd8f-4bbd-b421-d838b27a934c">

### Mail app version

4.0.0

### PHP engine version

None

### Web server

None

### Database

None

PR #9922, opened 2024-07-26
AccountForm: improve spacing and remove button borders

GVodyanov:
Fix  https://github.com/nextcloud/mail/issues/9915

| After | Before |
| - | - |
| Screenshot from 2024-07-26 10-04-32 | image |

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -8,29 +8,24 @@
 			cache-lifetime="0"
 			@changed="onModeChanged">
 			<Tab id="auto" key="auto" :name="t('mail', 'Auto')">
-				<label for="auto-name">{{ t('mail', 'Name') }}</label>
 				<NcInputField id="auto-name"
 					:value.sync="accountName"
+					:label="t('mail', 'Name')"
 					type="text"
 					:placeholder="t('mail', 'Name')"
 					:disabled="loading"
-					:label-outside="true"
 					autofocus />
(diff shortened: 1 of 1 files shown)


src/components/AccountForm.vue
```diff
@@ -703,6 +703,19 @@ export default {
 	font-weight: bold;
 }
 
+:deep(.checkbox-radio-switch) {
+	border: 0 !important;
```
ChristophWurst, line 707: this border is indended: https://nextcloud-vue-components.netlify.app/#/Components/NcCheckboxRadioSwitch

we have to fix that it's intermitted

src/components/AccountForm.vue
```diff
@@ -703,6 +703,19 @@ export default {
...
+}
+
+:deep(.input-field) {
+	margin-bottom: calc(var(--default-grid-baseline) * 3);
+	margin-top: var(--default-grid-baseline);
+}
+
+label {
+	margin-top: calc(var(--default-grid-baseline) * 2);
```
ChristophWurst, line 716: as an alternative we could convert every HTML label+input into NcInput. It is even more compact

src/components/AccountForm.vue
GVodyanov: Yep

src/components/AccountForm.vue
GVodyanov: Done

GVodyanov:
Here's how it's looking now 
image

Issue #9917, opened 2024-07-25
Composer alignment and text issues


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Add an account
3. Click *+ New message*

### Expected behavior

Good design

### Actual behavior

1. The From selector and the To and CC boxes are not aligned properly
2. *From* has a scrollbar with accounts that have very long names or email addresses (or both)

![image](https://github.com/user-attachments/assets/2a353d84-433d-47c8-b415-6b232fd51f77)

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
I'm still discussing how the alignment can actually be improved

1. *From* is shorter to make room for the modal minimize/maximize/close actions
2. *To* is shorter to make room for the cc/bcc toggle
3. *CC*/*BCC* take full width

Some options
1. Move everything down to avoid the collision with the modal actions and make all NcSelects as wide as the *To* NcSelect
2. Make all NcSelects full width and put the cc/bcc toggle elsewhere (actions menu?)

cc @jancborchardt @nimishavijay

ChristophWurst:
New idea:

> * All the boxes (From, To, CC) should have the same width. 
> * The From field should always be a one line box. Cut the email if it doens't fit. Everyone will get the different sender addresses based on the visual part

ChristophWurst:
![Bildschirmfoto vom 2024-07-26 13-34-51](https://github.com/user-attachments/assets/42a4b129-9117-4636-8d02-488ba1d508ab)

we could do it like this. If we always show the cc/bcc inputs then we eliminate the toggle button.

@jancborchardt @nimishavijay please let us know if this is ok for you

nimishavijay:
> Move everything down to avoid the collision with the modal actions 

This sounds good! In the future, if the width of the from/to/cc/bcc labels can be reduced, they can be left aligned and on the top row with the modal actions there could be a "New message" heading for the modal too.

> Make all NcSelects full width and put the cc/bcc toggle elsewhere

This sounds good to me too. Not sure about an action menu elsewhere, but would it be possible to have a button inside the "To" field which toggles cc/bcc fields? Something like the emoji picker button in Talk.

> If we always show the cc/bcc inputs then we eliminate the toggle button.

This I would generally advise against, as then the body text field will be very small, and this is where people spend most time while composing. 

> The From field should always be a one line box. Cut the email if it doens't fit. Everyone will get the different sender addresses based on the visual part

[...]

PR #9923, opened 2024-07-26
Composer: improve alignment and height of NcSelects

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/9917

| After | Before |
| - | - | 
| Screenshot from 2024-07-26 10-29-59|image|

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1684,8 +1684,12 @@ export default {
 	max-height: var(--default-clickable-area);
 }
 
+:deep(.vs__selected-options) {
+	overflow-y: hidden;
+}
+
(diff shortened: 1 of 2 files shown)


src/components/Composer.vue
```diff
@@ -1684,6 +1686,11 @@ export default {
 	width: 100%;
 }
 
+:deep(.v-select.select .vs__selected) {
+	height: unset;
+	min-height: calc(var(--default-clickable-area) - var(--default-grid-baseline));
+}
+
```
ChristophWurst, line 1693: This feels hacky. Can we set a max width? The NcSelect seems to elipsize its selection when the width of the container is restrained. I'd prefer that over line wraps.

GVodyanov:
Here's how it looks now 
image

image

PR #9924, opened 2024-07-26
Navigation: make new message button not scroll

GVodyanov:
Workaround for https://github.com/nextcloud/mail/issues/9916

| After | Before |
| - | - | 
| image | image |

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -5,7 +5,9 @@
 
 <template>
 	<AppNavigation>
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
OK as a workaround but not a fix. The scrolling is a regression of https://github.com/nextcloud-libraries/nextcloud-vue/pull/5831. I suggest we leave our ticket and migrate back to the body slot soon.

Issue #9918, opened 2024-07-25
"Notjunk" label should be hidden


ChristophWurst:
### Steps to reproduce

1. Set up an account hat has messages tagged with "Nonjunk"

### Expected behavior

I don't see the tags

### Actual behavior

<img width="464" alt="b8" src="https://github.com/user-attachments/assets/05e1a626-75e3-497f-b32a-82f3de43acda">

### PHP engine version

None

### Web server

None

### Database

None

PR #9925, opened 2024-07-26
fix(tags): Hide Notjunk tag

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/9918

I could not test this. Both email servers I tested with automatically unset Notjunk once it's set.

### DIFF
--- a/src/components/tags.js
+++ b/src/components/tags.js
@@ -9,6 +9,7 @@ export const hiddenTags = {
 	has_cal: '',
 	'has cal': '',
 	hasnoattachment: '',
+	notjunk: '',
 	loadremoteimages: '',
 	'unsubscribe newsletter': '',
 }


src/components/tags.js
```diff
@@ -9,6 +9,7 @@ export const hiddenTags = {
 	has_cal: '',
 	'has cal': '',
 	hasnoattachment: '',
+	Notjunk: '',
```
GVodyanov, line 12:
```suggestion
	notjunk: '',
```

Because of this https://github.com/nextcloud/mail/blob/4501d9b59a78494412634227cd7c8b635e0ce018/src/components/TagModal.vue#L104C104-L104C115

GVodyanov (dismissed):

hamza221 (approved):
Couldn't test

Issue #9919, opened 2024-07-25
Quick search bar alignment


ChristophWurst:
### Steps to reproduce

1. Set up an account

### Expected behavior

Good design

### Actual behavior

1. The layout of the search bar going completely to the right
2. The scrollbar starts below which looks somehow weird

![image](https://github.com/user-attachments/assets/aad33aca-77b1-4807-a804-a0c9144d4adf)

### Mail app version

v4.0.0-alpha1

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
@jancborchardt @nimishavijay we need your input.

1. The bar is moved to the right to make room for the nav toggle button.
2. The scrollbar starts below because the search bar is sticky.

ChristophWurst:
The advanced search icon was moved into the input intentionally: https://github.com/nextcloud/mail/issues/9755

nimishavijay:
There seems to be extra space to the right of the nav toggle, where did that come from? 

Should be: (margin on search-messages is `-1px 0 0 48px;`)
![image](https://github.com/user-attachments/assets/6ffc46e3-1390-48c2-a4fe-1be22925da0a)

But on the current daily: (margin on search-messages is `-1px 0 0 52px;`)
![image](https://github.com/user-attachments/assets/8eb2c764-c228-4726-b72e-1a99a3c5dd3e)


Tbh scrollbar looks great to me, not going all the way up is expected behaviour at this point (ref. Gmail (scrollbar focused for clarity), Confluence, Notion)

![image](https://github.com/user-attachments/assets/3b3b2d10-af4f-450e-ab77-48c51db3f6b1)
![image](https://github.com/user-attachments/assets/487d0900-6d3e-47bd-98f8-fcfa98c20366)
![image](https://github.com/user-attachments/assets/d937c1b9-0a23-4602-bab4-5d0b0aaac7ce)

ChristophWurst:
> Tbh scrollbar looks great to me, not going all the way up is expected behaviour at this point (ref. Gmail (scrollbar focused for clarity), Confluence, Notion)

If we compare to Talk with the sticky conversation header, the header has a subtle border at the bottom. We could also have one for the sticky search bar in Mail. That may make the scrollbar a bit less "weird".

ChristophWurst:
| Before | After |
|--------|--------|
[...]

PR #9938, opened 2024-07-29
fix: Separate quick search and threads list with a line

ChristophWurst:
| Before | After |
|--------|--------|
| image  | image | 

The element height stays at 52px so it aligns with the other elements
image

There is some whitespace diff. View the changes without whitespace changes.

Fixes https://github.com/nextcloud/mail/issues/9919


st3iny (approved):
Tested and works. Looks cleaner now.

PR #9941, opened 2024-07-30
fix(search): Show placeholder for the mailbox search input

ChristophWurst:
One day this will become an NcSelect. Today is not that day, so we add a HTML placeholder.

| Before | After |
|--------|--------|
| Bildschirmfoto vom 2024-07-30 09-52-49 | Bildschirmfoto vom 2024-07-30 09-54-35 |


ChristophWurst:
The placeholder was lost in https://github.com/nextcloud/mail/pull/9901.

st3iny (approved):

PR #9943, opened 2024-07-30
fix: Fine-tune thread summary box design

ChristophWurst:
* Increase information density
* Fix the heading case
* Make the assistant bubble a NcChip

| Before | After |
|--------|--------|
|Bildschirmfoto vom 2024-07-30 11-18-23 | Bildschirmfoto vom 2024-07-30 11-17-47 | 

10px padding make the elements align with the buttons below

image

### DIFF
--- a/src/components/ThreadSummary.vue
+++ b/src/components/ThreadSummary.vue
@@ -6,11 +6,9 @@
 	<div class="summary">
 		<div class="summary__header">
 			<div class="summary__header__actions">
-				<span class="summary__header__actions__brand">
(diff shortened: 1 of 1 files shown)


jancborchardt (approved):
Looks much nicer already! Moving to variables is a separate step I assume?

Similar for the whitespace above "Thread summary", which could be reduced to be the same as below the heading.

ChristophWurst:
> Moving to variables is a separate step I assume?

Yes: https://github.com/nextcloud/mail/issues/9944

PR #9956, opened 2024-08-01
Thread: make recipient bubbles prettier

GVodyanov:
For https://github.com/nextcloud/groupware/issues/80#issuecomment-2252284378

| After | Before |
| - | - |
| image | image |

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -480,19 +480,26 @@ export default {
 }
 
 .avatar-header {
-	max-height: 24px;
+	height: var(--default-clickable-area);
 	overflow: hidden;
+	display: flex;
+	align-items: stretch;
 }
 .avatar-more {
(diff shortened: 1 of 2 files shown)


src/components/Thread.vue
```diff
@@ -501,8 +501,21 @@ export default {
...
+	border-radius: var(--border-radius-large);
+	padding: calc(var(--default-grid-baseline) / 2);
+}
+
+.v-popper__popper--shown .user-bubble__wrapper {
+	margin-right: 0 !important;
 }
+
+.user-bubble__content {
+	height: unset !important;
```
ChristophWurst, line 515: According to https://nextcloud-vue-components.netlify.app/#/Components/NcUserBubble there is also a size prop for the user bubble

ChristophWurst (approved):

PR #9991, opened 2024-08-13
fix: case insensitive comparison for contact emails

hamza221:
- Make email comparison case insensitive 
- refactor to use singular form and remove trailing comma in case of one mail address in the address book


ChristophWurst (approved):
Bonus points for a unit test that covers this case

PR #9992, opened 2024-08-13
fix: send imip when importing an event in mail

kesselb:
## Summary

An iMIP message for attendees is only generated when updating an existing event (cf. https://github.com/nextcloud/3rdparty/blob/ea2fabbd358c9e0f9dae43bcb242b0cf8ee0d178/sabre/vobject/lib/ITip/Broker.php#L245-L254).

To also trigger an iMIP message from mail, we first import then event and update the attendance status afterward.

## How to test

- Send an imip message from an external party (e.g. google calendar) to your email
- Click "accept"
- The event is added to your calendar, but the organizer does not get a response.

### DIFF
--- a/src/components/Imip.vue
+++ b/src/components/Imip.vue
@@ -69,30 +69,32 @@
 		<template v-if="isRequest && userIsAttendee">
 			<div v-if="!wasProcessed && eventIsInFuture && existingEventFetched"
 				class="imip__actions imip__actions--buttons">
-				<ButtonVue type="secondary"
-					:loading="loading"
+				<NcButton type="secondary"
+					:disabled="loading"
 					:aria-label="t('mail', 'Accept')"
 					@click="accept">
 					{{ t('mail', 'Accept') }}
-				</ButtonVue>
-				<ButtonVue type="tertiary"
-					:loading="loading"
+				</NcButton>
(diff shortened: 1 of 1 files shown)


src/components/Imip.vue
```diff
@@ -414,44 +414,60 @@ export default {
...
-				logger.error('Failed to save event to calendar', {
-					error,
-					attendee,
-					calendar,
-					vEvent,
-					vCalendar,
-					existingEvent: this.existingEvent,
-				})
-			} finally {
-				this.loading = false
```
ChristophWurst, line 445: I liked unsetting loading in finally because this was it is ensured that the reset happens even on error

miaulalala (approved):

src/components/Imip.vue
kesselb: I still reset, just a few lines later. We have two try-catch blocks now, and both are wrapped in an if statement. First the loading state is enabled, then one or both try-catch statements run, and last the loading state is disabled. 

The current loading state does not work because ButtonVue (an alias for NcButton) has no property called "loading". I've changed the code to disable the button and added the loading icon to the right.

ChristophWurst (approved):

PR #10000, opened 2024-08-14
feat(deps): Add Nextcloud 31 support

nickvergessen:

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -34,7 +34,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>4.0.0-beta.2</version>
+	<version>4.1.0-dev.0</version>
 	<licence>agpl</licence>
 	<author homepage="https://github.com/ChristophWurst">Christoph Wurst</author>
 	<author homepage="https://github.com/GretaD">GretaD</author>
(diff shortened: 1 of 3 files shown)


kesselb (approved):

kesselb:
Lowered min-version to 30

appinfo/info.xml
```diff
@@ -34,7 +34,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>4.0.0-alpha.3</version>
+	<version>5.0.0-dev.0</version>
```
kesselb, line 37: Check with Christoph if we want to go with 5.0.0 or 4.1.0

kesselb (changes requested):

appinfo/info.xml
ChristophWurst: 4.1 as long as we keep 30 as min

kesselb (approved):

appinfo/info.xml
```diff
@@ -34,7 +34,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>4.0.0-beta.2</version>
+	<version>4.1.0-dev.0</version>
```
ChristophWurst, line 37: can we use alpha? I think the versions are sorted alphabetically. If we start with dev we can only go to RC and not to alpha or beta

PR #10007, opened 2024-08-19
fix:flagged icons position

GretaD:
before list layout:
Screenshot from 2024-08-19 10-01-39

after
Screenshot from 2024-08-19 09-49-25

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -937,7 +937,7 @@ export default {
 	// In message list, but not the one in the action menu
 	&.app-content-list-item-star {
 		background-image: none;
-		left: 4px;
+		left: 5px;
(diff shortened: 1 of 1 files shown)


src/components/Envelope.vue
```diff
@@ -989,6 +990,9 @@ export default {
 		opacity: .1;
 	}
 }
+.one-lie.junk-icon-style {
```
ChristophWurst, line 993: Liar

ChristophWurst (changes requested):

src/components/Envelope.vue
GretaD: :woman_facepalming:

ChristophWurst (approved):

PR #10009, opened 2024-08-19
Composer, RecipientListItem: make long recipients clip and subject unconstrained

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/9917#issuecomment-2296234952

| After | Before |
| - | - |
| Screenshot from 2024-08-19 15-44-16 | image |


ChristophWurst (approved):
You confused me with your after/before once again

:+1: good looks

PR #10022, opened 2024-08-20
fix: polish list item

GretaD:
Request:

950-714-max

after. Please notice that the avatar is already 40px
Screenshot from 2024-08-20 11-27-06
Screenshot from 2024-08-20 11-21-34
Screenshot from 2024-08-20 11-21-30

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -937,7 +937,7 @@ export default {
 	// In message list, but not the one in the action menu
 	&.app-content-list-item-star {
 		background-image: none;
-		left: 5px;
+		left: 1px;
 		top: 8px;
 		opacity: 1;
 
@@ -977,7 +977,7 @@ export default {
 .junk-icon-style {
 	opacity: .2;
 	display: flex;
-	top: 42px;
+	top: 32px;
 	left: 32px;
 	background-size: 16px;
 	height: 20px;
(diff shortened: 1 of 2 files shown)


ChristophWurst (changes requested):
* Avatar has size=40 but is 45x45px tall
* Tags looks wonky
* Tags resize on hover and make the element appear jittery
* Actions menu handle overflows the element image

src/components/Envelope.vue
```diff
@@ -1072,13 +1072,19 @@ export default {
 }
 :deep(.list-item) {
 	flex-wrap: wrap;
+	padding: var(--default-grid-baseline);
```
ChristophWurst, line 1075: Could you move this into the skeleton component? list item is our own component now. we don't need the deep selectors to patch styles

src/components/Envelope.vue
```diff
@@ -1072,13 +1072,19 @@ export default {
...
 }
 :deep(.line-two__subtitle) {
 	display: flex;
 	flex-basis: 100%;
 	padding-left: 40px;
 	width: 450px;
 }
+:deep(.list-item__actions) {
```
ChristophWurst, line 1083: same here

ChristophWurst (changes requested):

ChristophWurst (approved):
Wygląda dobrze!

image

nit: tags should probably wrap line when there are many or have a long label. the ellipsis looks strange

GretaD:
> Wygląda dobrze!
> 
> image
> 
> nit: tags should probably wrap line when there are many or have a long label. the ellipsis looks strange

the ellipsis happen only when there are too many tags, up to 3 its shown correctly, unless these 3 have longer translations or are personalized ones with long string. But yes, i agree with you.

PR #10023, opened 2024-08-20
fix: use correct type for button, type for prop

kesselb:
Fix two warnings from the phishing component 

- [Vue warn]: Invalid prop: type check failed for prop "phishingData". Expected Object, got Array
- [Vue warn]: Invalid prop: custom validator check failed for prop "type".


ChristophWurst (approved):
Good soup

miaulalala (approved):

st3iny (approved):
Here, have a :cookie:

PR #10024, opened 2024-08-20
fix: wrong attendance status for imip events

kesselb:
- [x] Needs/Based on https://github.com/nextcloud/mail/pull/9992

### Problem

 The attendence status for an imip event might be wrong. 

### Before

[Screencast from 2024-08-20 13-53-29.webm](https://github.com/user-attachments/assets/513c24df-4849-441c-b571-1d3e9364c1d7)

### After

[Screencast from 2024-08-20 13-52-35.webm](https://github.com/user-attachments/assets/b65c22ef-11ad-47e2-8a35-f97ec43eaaa2)

Commit a1ecb2e15a4ea848e171919bd50b42888eb593da

One may have imported an event, but the attendance status is not properly fetched when rendering the email.
The reason is a timing / state problem.

- Loading the user principal and collections is initialized in https://github.com/nextcloud/mail/blob/6fc45eb0630b9065f9ccb4c1da5cc9557f7df834/src/App.vue#L49-L50
- If the backend request for the message body is faster, than loading the principal and collections, then Imip.fetchExistingEvent runs without having calendars and changes existingEventFetched to true that prevents the method from running again.
- Solution: Render the imip component when principal and collections are fetched.

Commit d677ce7af6dae7deb33bf0b471fa8c5a3aa96edc

Accepting a calendar invitation should always go to a writable calendar, and therefore we can skip the check if the event exists in a read-only calendar.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -48,6 +48,7 @@ export default {
 		this.sync()
 		await this.$store.dispatch('fetchCurrentUserPrincipal')
 		await this.$store.dispatch('loadCollections')
+		this.$store.commit('hasCurrentUserPrincipalAndCollections', true)
 	},
 	methods: {
 		reload() {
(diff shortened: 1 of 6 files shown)


st3iny (approved):
Looks good, as discussed in our call.

Issue #10043, opened 2024-08-23
List layout is not shown in one line


GretaD:
### Steps to reproduce

1. Open mail app
2. Change the layout from mail settings to list layout
3. See the style

### Expected behavior

Messages are shown in one line

### Actual behavior

Messages are shown in 2 lines
![Screenshot from 2024-08-23 12-18-57](https://github.com/user-attachments/assets/d2c3b270-5907-45b8-a3e6-fe5577cdd889)

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
This is important to be fixed before the final release

GretaD:
@GVodyanov do you have some time to have a look at this one? You probably missed some styling when you created our own component, you can compare with my original pr: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5209
it should be something like flex-direction: row and some width, that is missing. Otherwise i can have a look after my customer call.

GVodyanov:
My bad, for some reason I remembered it having two lines and not just one

PR #10044, opened 2024-08-23
EnvelopeSkeleton: make oneline envelopes actually one line and not two

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/10043

| Before | After |
| - | - |
| image | Screenshot from 2024-08-23 14-46-35 |

### DIFF
--- a/src/components/EnvelopeSkeleton.vue
+++ b/src/components/EnvelopeSkeleton.vue
@@ -552,24 +552,47 @@ export default {
 		--list-item-border-radius: var(--border-radius-element, calc(var(--default-clickable-area) / 2));
 		padding-block: calc(var(--list-item-padding) * 2);
 		--list-item-padding: 2px;
+		height: unset;
 
-		.list-item-content__inner__main {
-			display: flex;
-			justify-content: start;
-			gap: 12px;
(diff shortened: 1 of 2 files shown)


src/components/EnvelopeSkeleton.vue
```diff
@@ -552,24 +552,51 @@ export default {
...
 		}
 
 		a {
 			margin: 0;
+			align-items: center;
+			height: unset;
+		}
+
+		:deep(.avatardiv__initials) {
+			line-height: 40px !important;
```
ChristophWurst, line 594: why the deep overwrite?

GVodyanov: Double-checked and it is indeed not needed. The idea was to have the avatar be `40px` in oneline and `45px` (as per Marco's suggestion some time ago), but now I noticed that Greta made it `40px` everywhere and not just in oneline, I assume per a new design decision.

GretaD: yes, the avatar should be 40px everywhere on mail :)

GretaD (approved):
tested and looks good

Issue #9942, opened 2024-07-30
Confusing search icon (needs to be replaced by a more suggestive one)


Jerome-Herbinet:
### Is your feature request related to a problem? Please describe.

In my opinion, the button below is confusing. The icon used suggests more a "settings" notion of the current application or account, or the current folder, but has nothing to do with the notion of search (or filter). Instinctively, when I look for a search or filter button in an interface, it looks like a magnifying glass or a funnel, but certainly not like the icon currently in use.
![Capture d’écran du 2024-07-30 09-40-37](https://github.com/user-attachments/assets/b1960166-112c-473b-baa6-022f1b071c9a)

### Describe the solution you'd like

We need an icon that really suggests the notion of search or filtering. Here's a test I've done with my browser's code inspector. I've replaced the tune icon with a magnifying glass icon directly suggesting the notion of search. It could also have been a funnel icon suggesting a notion of filtering. 
I also modified the tooltip, thinking it would be clearer that way, but that's a detail compared to the icon problem. Depending on what the design team says, I may open a PR.
![Capture d’écran du 2024-07-30 09-45-07](https://github.com/user-attachments/assets/e70118a5-150b-41a9-b7f5-3ac2dc8b39eb)

@nextcloud/designers

ChristophWurst:
The icon was spec'ed at https://github.com/nextcloud/mail/pull/6546#issuecomment-1193940095.

@nimishavijay shall we replace the icon?

marcoambrosini:
I think a filter icon would be the most proper here 
https://fonts.google.com/icons?selected=Material+Symbols+Outlined:filter_alt:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=filter&icon.size=24&icon.color=%235f6368

ChristophWurst:
Let me know which icon to use and I'll change it. The filter icon was there before the settings icon: https://github.com/nextcloud/mail/pull/6546#issuecomment-1193887676.

Jerome-Herbinet:
The share and/or lock icon should be displayed as an overlay on top-right of the profile or initial image (with a higher z-index).

nimishavijay:
Filter icon makes sense here. Originally the settings icon (MDI "tune" icon) was used for parity with Gmail, but you're right, Gmail's search flow is different (entirely new screen shows up with search results , so that button was used to "tune" the search). Over here since the current view is just being filtered it makes sense to use the filter icon 👍

[...]

PR #10056, opened 2024-08-27
Replace tune icon with filter icon

Jerome-Herbinet:
Fixes: #9942 

Check commit.

**I don't have a test environment. Test from reviewers need.**

### DIFF
--- a/src/components/SearchMessages.vue
+++ b/src/components/SearchMessages.vue
@@ -13,7 +13,7 @@
 				@click="toggleButtons">
 			<NcButton type="tertiary" class="search-messages--filter" @click="moreSearchActions = true">
 				<template #icon>
-					<Tune :size="20" />
+					<FilterVariantIcon :size="20" />
 				</template>
 			</NcButton>
 			<NcButton v-if="filterChanged"
@@ -296,7 +296,7 @@ import NcDateTimePicker from '@nextcloud/vue/dist/Components/NcDateTimePicker.js
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import NcCheckboxRadioSwitch
 	from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
-import Tune from 'vue-material-design-icons/Tune.vue'
+import FilterVariantIcon from 'vue-material-design-icons/FilterVariant.vue'
 import Close from 'vue-material-design-icons/Close.vue'
 import IconClose from '@mdi/svg/svg/close.svg'
 import IconMagnify from '@mdi/svg/svg/magnify.svg'
@@ -317,7 +317,7 @@ export default {
 		NcDateTimePicker,
 		NcButton,
 		NcCheckboxRadioSwitch,
-		Tune,
+		FilterVariantIcon,
 		Close,
 	},
 	props: {


ChristophWurst (changes requested):
The icon component is not registered on the Vue component

Jerome-Herbinet:
> The icon component is not registered on the Vue component

@ChristophWurst this is new for me ; how to do that ? Can you help me ?

ChristophWurst:
Delete https://github.com/nextcloud/mail/blob/295a1dcf4c98da0ca46a03847c5bb9a320295f03/src/components/SearchMessages.vue#L299 and replace Tune with FilterIcon at https://github.com/nextcloud/mail/blob/295a1dcf4c98da0ca46a03847c5bb9a320295f03/src/components/SearchMessages.vue#L321

jancborchardt (changes requested):
@ChristophWurst @Jerome-Herbinet I think it will have to be FilterVariantIcon so it is going to be the list/filter icon (not the funnel) as per https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%235f6368&selected=Material+Symbols+Outlined:filter_list:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=filter ?

Jerome-Herbinet:
> Delete
> 
> https://github.com/nextcloud/mail/blob/295a1dcf4c98da0ca46a03847c5bb9a320295f03/src/components/SearchMessages.vue#L299
> and replace Tune with FilterIcon at
> 
> https://github.com/nextcloud/mail/blob/295a1dcf4c98da0ca46a03847c5bb9a320295f03/src/components/SearchMessages.vue#L321

@ChristophWurst OK

Jerome-Herbinet:
@jancborchardt could you test it ?

jancborchardt (commented):
@Jerome-Herbinet as per my comment before, the icon probably needs to be "FilterVariant".

It should _not_ look like this [funnel icon](https://pictogrammers.com/library/mdi/icon/filter/), but rather like this [list filter icon](https://pictogrammers.com/library/mdi/icon/filter-variant/).

Jerome-Herbinet:
> @Jerome-Herbinet as per my comment before, the icon probably needs to be "FilterVariant".
> 
> It should _not_ look like this [funnel icon](https://pictogrammers.com/library/mdi/icon/filter/), but rather like this [list filter icon](https://pictogrammers.com/library/mdi/icon/filter-variant/).

OK @jancborchardt ; to make finding icons easier, can you tell me how (and/ore where) you easily browse your entire Nextcloud icon collection to select the right one (AFAIK I can't open files viewed as an image...).

jancborchardt:
@Jerome-Herbinet as far as I know, that is https://pictogrammers.com/library/mdi/ – but a frontender would have to help here.

st3iny (commented):
Works so far. Just needs the correct icon, as Jan mentioned.

grafik

ChristophWurst:
What was force pushed? The icon is still to be replaced, right?

Jerome-Herbinet:
> What was force pushed? The icon is still to be replaced, right?

Yes, forget it, I still have to make discussed changes.

jancborchardt (approved):
Seems good – don’t know about the icon naming conventions but it seems the variables usually don’t have "Icon" appended, so you could call it simply "FilterVariant" instead of "FilterVariantIcon"?

st3iny (approved):
Tested and works. Thank you very much :)

Issue #10045, opened 2024-08-23
Alter Outbox logic to handle automated messages better


SebastianKrupinski:
### Is your feature request related to a problem? Please describe.

Yes, at the moment the outbox logic requires the editorBody property to be set. This works correctly for messages generated with the front end UI but causes errors/issues for automated messages generated from the server or other apps.

### Describe the solution you'd like

Perform check to see if the editorBody is set then disable editing if it is missing and maybe only show a preview of the message in the preview pane.

Also handle "Send now" better so that it does not require the editorBody to be set.

### Describe alternatives you've considered

None

SebastianKrupinski:
Follow up to #9651

PR #10059, opened 2024-08-27
Alter Outbox logic to handle automated messages better

SebastianKrupinski:
Resolves: #10045

- Disabled viewing of messages that have NO editor body
- Fixed outbox controller update routine to accept null editor body

### DIFF
--- a/src/components/OutboxMessageListItem.vue
+++ b/src/components/OutboxMessageListItem.vue
@@ -163,10 +163,14 @@ export default {
 				if (error.data !== undefined) {
 					await this.outboxStore.updateMessage({ message: error.data[0], id: message.id })
 				}
+
 			}
 		},
 		async openModal() {
 			if (this.message.status === STATUS_IMAP_SENT_MAILBOX_FAIL) {
+				return
+			}
+			if (this.message.editorBody === null) {
 				return
 			}
 			await this.$store.dispatch('startComposerSession', {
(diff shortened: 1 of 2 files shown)


src/components/OutboxMessageListItem.vue
```diff
@@ -151,7 +151,7 @@ export default {
 				failed: false,
 				sendAt: (new Date().getTime() + UNDO_DELAY) / 1000,
 			}
-			await this.outboxStore.updateMessage({ message, id: message.id })
+			await this.outboxStore.sendMessage({ id: message.id, force: true })
```
ChristophWurst, line 154: > Fixed send now to use proper send method instead of updating the message

I think this was intentional. We set the new sendAt value with the undo delay. Then the actual sending happens.

src/components/OutboxMessageListItem.vue
```diff
@@ -151,7 +151,7 @@ export default {
...
 			try {
 				if (this.message.status !== STATUS_IMAP_SENT_MAILBOX_FAIL) {
 					await this.outboxStore.sendMessageWithUndo({ id: message.id })
```
ChristophWurst, line 157: If this line is reached, the message would be resent (if the outbox still has it) or cause an error

ChristophWurst: The *else* is even more problematic. Messages with status=STATUS_IMAP_SENT_MAILBOX_FAIL must not be sent again but just copied to the \Sent mailbox

ChristophWurst (approved):

PR #10069, opened 2024-08-28
fix: create mailbox doesnt reset after saving

GretaD:
When creating a mailbox, there is no indication that is saving or what is happening. Also, if you want to add a second mailbox, the last mailbox's name is still on the input field.
before

[Screencast from 28.08.2024 16:51:14.webm](https://github.com/user-attachments/assets/144cedcf-a3dc-4805-b21e-bd83ed31ba13)

after

[Screencast from 28.08.2024 16:49:54.webm](https://github.com/user-attachments/assets/aef66882-5bfe-4687-97d0-8f828e5eeaae)

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -40,13 +40,13 @@
 						@update:checked="changeShowSubscribedOnly">
 						{{ t('mail', 'Show only subscribed mailboxes') }}
 					</ActionCheckbox>
-					<ActionButton v-if="!editing" @click="openCreateMailbox">
+					<ActionButton v-if="!editing && nameLabel" @click="openCreateMailbox">
 						<template #icon>
 							<IconFolderAdd :size="16" />
 						</template>
 						{{ t('mail', 'Add mailbox') }}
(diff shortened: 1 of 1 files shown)


src/components/NavigationAccount.vue
```diff
@@ -176,23 +179,30 @@ export default {
...
+			} catch (error) {
+				logger.error('could not create mailbox', { error })
+				throw error
+			} finally {
+				this.showSaving = false
+				this.nameInput = false
+				this.editing = false
+				this.createMailboxName = ''
+			}
+			logger.info(`mailbox ${name} created`)
```
GVodyanov, line 201: I don't think these are supposed to be backticks instead of apostrophes

GVodyanov (approved):
Works and I really like it!

src/components/NavigationAccount.vue
ChristophWurst: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

PR #10073, opened 2024-08-30
fix: link checking missing scheme

hamza221:


lib/Service/PhishingDetection/LinkCheck.php
```diff
@@ -40,6 +40,13 @@ private function getInnerText(\DOMElement $node) : string {
 		return $innerText;
 	}
 
+	private function parse(string $url): string {	
+		if(strpos($url, "://") === false) {
```
ChristophWurst, line 44: https://www.php.net/manual/en/function.str-contains.php

lib/Service/PhishingDetection/LinkCheck.php
```diff
@@ -40,6 +40,13 @@ private function getInnerText(\DOMElement $node) : string {
...
+			$url = "http://".$url;
```
ChristophWurst, line 45: nit: just return the concatenated string

ChristophWurst (commented):
This smells like a case that can be unit-tested :smiling_imp:

lib/Service/PhishingDetection/LinkCheck.php
```diff
@@ -40,6 +40,13 @@ private function getInnerText(\DOMElement $node) : string {
 		return $innerText;
 	}
 
+	private function parse(string $url): string {
+		if(strpos($url, "://") === false) {
```
kesselb, line 44:
```suggestion
		if(strpos($url, '://') === false) {
```

lib/Service/PhishingDetection/LinkCheck.php
```diff
@@ -40,6 +40,13 @@ private function getInnerText(\DOMElement $node) : string {
...
+			$url = "http://".$url;
```
kesselb, line 45:
```suggestion
			$url = 'http://' . $url;
```

lib/Service/PhishingDetection/LinkCheck.php
SebastianKrupinski, line 44: The regex pattern seem to only check for "https://" and NOT "http://" so adding "http://" to a URL that is missing the scheme would produce a non matching result, stripping of the scheme for all links might be better

SebastianKrupinski:
Does the code check for dangerous URL schemes like "file://" or "ftp://"?

lib/Service/PhishingDetection/LinkCheck.php
hamza221: The regex matches 
> URLs -- http, https, and naked domains like "example.com".  
 but also this method is called later than the regex it adds http:// ( the protocol here doesn't really matter) it's only added so [parse_url](https://www.php.net/manual/en/function.parse-url.php) can work properly

hamza221:
> Does the code check for dangerous URL schemes like "file://" or "ftp://"?

How are these schemes dangerous for phishing? sorry my knowledge is limited about the cases. 
Let's discuss either in a private message and I'll open an issue for it, or you can also open one and assign me.

kesselb:
This pull request is about fixing a problem we discussed in the public chat. 
As shown below the warning is triggered even if the urls are using the same host.

image

The reason is that parse_url does not work if the given string does not contain something that looks like a protocol: 
https://3v4l.org/cap3C (it would also work to use "dummy://" instead of "http://".

An additional check, does the mail contain links with unusual protocols, sounds like a good enhancement to me. 
I'd say that should go into a new issue.

SebastianKrupinski:
> > Does the code check for dangerous URL schemes like "file://" or "ftp://"?
> 
> How are these schemes dangerous for phishing? sorry my knowledge is limited about the cases. Let's discuss either in a private message and I'll open an issue for it, or you can also open one and assign me.

I don't know if they are dangerous for phishing specifically but they are dangerous in general. For instance "ftp://domain/folder/file.doc" will trigger the download and opening of a external file. "file://folder/file.exe" will trigger the opening of a local executable. This in not a blocker for me, just thought the warning could be included also.

kesselb (approved):

PR #10084, opened 2024-09-02
fix: thread disappearing after refresh

### LINKED ISSUE #9759: On list layout, the thread disappears after refresh
### Steps to reproduce

1. open the mail app
2. change to list layout
3. open a message
4. reload the page

### Expected behavior

i still see the thread

### Actual behavior

the thread disappears

### Mail app version

3.7

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes #9759

to test
1. open mail app on list view
2. open a message
3. refresh

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -286,6 +286,8 @@ export default {
 			this.handleMailto()
 			if (to.name === 'mailbox' && to.params.mailboxId === PRIORITY_INBOX_ID) {
 				await this.onPriorityMailboxOpened()
+			} else if (this.isThreadShown) {
+				await this.fetchEnvelopes()
 			}
 		},
 		async hasFollowUpEnvelopes(value) {
@@ -298,6 +300,7 @@ export default {
 		mailbox() {
 			clearTimeout(this.startMailboxTimer)
 			setTimeout(this.saveStartMailbox, START_MAILBOX_DEBOUNCE)
+			this.fetchEnvelopes()
 		},
 	},
 	created() {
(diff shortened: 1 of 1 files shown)


ChristophWurst (commented):
Will this cause double fetching when you don't refresh the page but navigate from one mailbox to another?

GretaD:
> Will this cause double fetching when you don't refresh the page but navigate from one mailbox to another?

very good point/question. I will add a check to see if the envelopes are already being loaded before making the fetch request.

GVodyanov (commented):
I am testing this out, but am a little confused. On firefox the issue was not reproducible on main. On chromium, reloading the page causes the thread screen to be blank, both on main and with the fix.

GretaD:
@kesselb did you change anything when you rebased? :D

kesselb:
I'm not aware :confused: 

The commits https://github.com/nextcloud/mail/commit/ebb7a3d7bc0c31b20c85d53ddcaf83b61c509c64 and https://github.com/nextcloud/mail/commit/b7af7837673acdeac13f7c4e518f0b7793203aa6 look pretty much like https://github.com/nextcloud/mail/commit/ebb7a3d7bc0c31b20c85d53ddcaf83b61c509c64.

kesselb:
> On firefox the issue was not reproducible on main. 

Same for me. 

> On chromium, reloading the page causes the thread screen to be blank, both on main and with the fix.

Works over here ;) It's also blank on main, but on this branch the selected thread is shown.

It's a bit odd that Thread depends on data loaded by Mailbox, but Mailbox might not be there. I hope we can restructure that in the future to have a clearer flow.

kesselb (approved):

src/components/MailboxThread.vue
```diff
@@ -298,18 +300,31 @@ export default {
...
+			this.fetchEnvelopes()
 		},
 	},
 	created() {
 		this.handleMailto()
 	},
 	async mounted() {
 		setTimeout(this.saveStartMailbox, START_MAILBOX_DEBOUNCE)
+		if (this.isThreadShown) {
+			this.fetchEnvelopes()
```
kesselb, line 312: await?

src/components/MailboxThread.vue
```diff
@@ -286,6 +286,8 @@ export default {
 			this.handleMailto()
 			if (to.name === 'mailbox' && to.params.mailboxId === PRIORITY_INBOX_ID) {
 				await this.onPriorityMailboxOpened()
+			} else if (this.isThreadShown) {
+				this.fetchEnvelopes()
```
kesselb, line 290: await?

kesselb:
Shall we backport to stable4.0 and stable3.7?

PR #10100, opened 2024-09-04
chore: rename sieve security to sieve filter

kesselb:
- Rename "Enable sieve security" to "Enable sieve filter"
- Use the translation function
- Add a small introduction for the feature

image

### DIFF
--- a/src/components/SieveAccountForm.vue
+++ b/src/components/SieveAccountForm.vue
@@ -5,8 +5,11 @@
 <template>
 	<form id="sieve-form">
 		<p>
+			{{ t('mail', 'Sieve is a powerful language for writing filters for your mailbox. You can manage the sieve scripts in Mail if your email service supports it.') }}
+		</p>
+		<p>
 			<NcCheckboxRadioSwitch :checked.sync="sieveConfig.sieveEnabled">
-				{{ ('mail', 'Enable sieve security') }}
+				{{ t('mail', 'Enable sieve filter') }}
 			</NcCheckboxRadioSwitch>
 		</p>
 		<div v-if="sieveConfig.sieveEnabled">


src/components/SieveAccountForm.vue
```diff
@@ -4,9 +4,12 @@
 -->
 <template>
 	<form id="sieve-form">
+		<p>
+			{{ t('mail', 'Sieve is a powerful language for writing filters for your mailbox. You can control the sieve scripts from Nextcloud Mail if your email service supports it.')}}
```
ChristophWurst, line 8:
```suggestion
			{{ t('mail', 'Sieve is a powerful language for writing filters for your mailbox. You can control the sieve scripts from {product} Mail if your email service supports it.', { product: OCA.theme.????? })}}
```

we use it in other places too. forgot the exact variable ;-)

ChristophWurst: https://github.com/nextcloud/mail/blob/629a9fa693648d9414ecc8cb0fc3f315305dbba8/src/components/NoMessageSelected.vue#L33

kesselb: I like the idea :+1: 

https://github.com/nextcloud/mail/pull/10100/commits/f17aba480a09e31d8b58f06a09861b4a00618842 did not work, even if `window?.OC?.theme?.name` evaluates to "Testcloud". I assume that OC.theme is not populated when the component is rendered, but didn't look at it properly. 

Pushed another fixup to use a wording without Nextcloud. Let me know if you want me to take a look later to include the instance name or product name, I will create an issue then.

ChristophWurst: Then let's avoid the name :+1:

PR #10104, opened 2024-09-05
fix: phishing warning layout

hamza221:
| b | a |
|--------|--------|
| image | image |

### DIFF
--- a/src/components/PhishingWarning.vue
+++ b/src/components/PhishingWarning.vue
@@ -3,20 +3,22 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<div class="warning">
-		<div class="warning__title">
+	<div class="phishing-warning">
+		<div class="phishing-warning__title">
 			<IconAlertOutline :size="16" :title="t('mail', 'Phishing email')" />
 			This email might be a phishing attempt
 		</div>
-		<ul v-for="(warning,index) in warnings" :key="index" class="warning__list">
-			<li>{{ warning.message }}</li>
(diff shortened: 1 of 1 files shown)


kesselb (approved):

marcoambrosini (commented):
I think a is more appropriate but to be consistent with how we show warnings and info boxes, I would also add a `calc(var(--default-grid-baseline) * 2)` margin

src/components/PhishingWarning.vue
```diff
@@ -64,7 +64,7 @@ export default {
     width: 100%;
     text-align: left;
     padding: 8px;
-    margin-bottom: 0 5px;
+    margin-bottom: 0 calc(var(--default-grid-baseline) * 2);
```
marcoambrosini, line 67:
```suggestion
    margin: calc(var(--default-grid-baseline) * 2);
```

src/components/PhishingWarning.vue
```diff
@@ -64,7 +64,7 @@ export default {
     width: 100%;
```
marcoambrosini, line 64:
```suggestion
```

src/components/PhishingWarning.vue
hamza221: `-bottom` is a leftover mb, But I do believe that the margin should only be top - bottom. don't you thing so? 
so more like `margin: calc(var(--default-grid-baseline) * 2) 0` like this ->
image
otherwise it wouldn't it pop out unless that's what you're going for 
image

marcoambrosini: I think we're going for something more like in your second screenshot. Pop factor looks very similar to me

PR #10106, opened 2024-09-05
chore: replace unconventional commits workflow

kesselb:
- Enforce the check also for main
- Get updates from the org template


ChristophWurst (approved):
Enforcing is fine by me :+1:

Issue #10039, opened 2024-08-21
Sync and Send NOT reusing connections for operations


SebastianKrupinski:
### Steps to reproduce

1. Send or Sync messages
2. Check server or mail client logs

### Expected behavior

Use a single connection for all operations then close connection once completed

### Actual behavior

Sync operation opens new IMAP connection for every mailbox that needs to sync.

Send operation opens one new SMTP connection and two new IMAP connections for every email it sends.

### Mail app version

Lastest Dev

### Mailserver or service

IMAP/SMTP

### PHP engine version

PHP 8.3

### Web server

Nginx

### Database

MySQL

SebastianKrupinski:
Test Results of sending 3 messages at the same time from the same account.

```
Thu, 08 Aug 2024 12:37:22 - INFO - Network connection ended (network.connection-end) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 60796, elapsed = 29ms
Thu, 08 Aug 2024 12:37:22 - INFO - SMTP EHLO command (smtp.ehlo) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 60796, domain = "localhost"
Thu, 08 Aug 2024 12:37:22 - INFO - Network connection started (network.connection-start) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 60796

Thu, 08 Aug 2024 12:37:20 - INFO - Network connection ended (network.connection-end) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 60792, elapsed = 28ms
Thu, 08 Aug 2024 12:37:20 - INFO - SMTP EHLO command (smtp.ehlo) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 60792, domain = "localhost"
Thu, 08 Aug 2024 12:37:20 - INFO - Network connection started (network.connection-start) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 60792

Thu, 08 Aug 2024 12:37:17 - INFO - Network connection ended (network.connection-end) listenerId = "submission", protocol = smtp, remoteIp = 127.0.0.1, remotePort = 58302, elapsed = 55ms
[...]

Issue #10098, opened 2024-09-04
Reuse IMAP client for ``\OCA\Mail\Service\Sync\ImapToDbSynchronizer::syncAccount``


ChristophWurst:
### Is your feature request related to a problem? Please describe.

The method ``\OCA\Mail\Service\Sync\ImapToDbSynchronizer::syncAccount`` iterates over an account's mailboxes and synchronizes them. For each mailbox we create a new client, which causes a new connection, and slows down the process.

### Describe the solution you'd like

* Create the client instance in ``\OCA\Mail\Service\Sync\ImapToDbSynchronizer::syncAccount``
* Adjust ``\OCA\Mail\Service\Sync\ImapToDbSynchronizer::sync`` to take the client as parameter instead of creating it via the factory

### Describe alternatives you've considered

Cache connections in the IMAP client factory: https://github.com/nextcloud/mail/commit/d68749d60e28a6ec42bce87dd2729b5b799c4cea. Reverted via https://github.com/nextcloud/mail/commit/a23e8238dffd46115c04fd43d835f6db32510023 because it caused too many open connections.

PR #10108, opened 2024-09-06
feat: imap connection caching and reuse

SebastianKrupinski:
#10039

This is a proof of concept for caching and reusing IMAP connection.


ChristophWurst (changes requested):
Please follow the instructions in https://github.com/nextcloud/mail/issues/10098

We should be able to solve the problem without reintroducing statefulness

SebastianKrupinski:
@ChristophWurst this is test code, that was written before you created/assigned #10098 it was very meant to be a solution, just posted for the purpose of saving it.

ChristophWurst:
This is obsolete. We have improved the connection usage. Closing :)

PR #10122, opened 2024-09-09
feat: add mention to mail

### LINKED ISSUE #10110: Add recipients to a message using the @ symbol

- [x] Search for contacts in address book using @ symbol in composer
- [x] only show contacts with at least one email address
- [x] Add the contact's first email address the "To" Field
- [x] Render the mention correctly (FN using https://nextcloud-vue-components.netlify.app/#/Components/NcUserBubble ) [UPDATE: will use anchor tags instead so we don't alter how the message is meant to be displayed} 

_No response_

_No response_

### Additional context

_No response_

hamza221:
fix #10110

https://github.com/user-attachments/assets/9d8f5a99-61c9-4cc5-9adb-0ca5ec2cb77e

### DIFF
--- a/lib/Controller/ContactIntegrationController.php
+++ b/lib/Controller/ContactIntegrationController.php
@@ -15,18 +15,27 @@ use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\JSONResponse;
+use OCP\ICache;
+use OCP\ICacheFactory;
 use OCP\IRequest;
 
 #[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
 class ContactIntegrationController extends Controller {
 	private ContactIntegrationService $service;
+	private ICache $cache;
+	private string $uid;
+
 
 	public function __construct(string $appName,
 		IRequest $request,
-		ContactIntegrationService $service) {
+		ContactIntegrationService $service,
+		ICacheFactory $cacheFactory,
+		string $UserId) {
 		parent::__construct($appName, $request);
 
 		$this->service = $service;
+		$this->cache = $cacheFactory->createLocal('mail.contacts');
+		$this->uid = $UserId;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


ChristophWurst (approved):
Superb feature!

Works well! I only noticed a few minor issues
* There are XHR requests to search for empty names
* There is an XHR request to search once the mention suggestion is selected with tab/enter

Nice to have
* Avoid searching the same term again. Cache the responses.

hamza221:
@ChristophWurst 
>There is an XHR request to search once the mention suggestion is selected with tab/enter 

Unfortunately that can't / will be hard to fix since that's how the mention plugin work in CKEditor (a feature not a bug) 
since the trigger '@'  here is kept unlike for (links and emojis) 
anytime you have your cursor near a trigger it will call the search function to update the state

hamza221:
> Fair, but we can just not do a request for an empty string ;-)

yes that's fixed

ChristophWurst (approved):

lib/Controller/ContactIntegrationController.php
```diff
@@ -15,18 +15,23 @@
...
 class ContactIntegrationController extends Controller {
 	private ContactIntegrationService $service;
+	private ICache $cache;
+
 
 	public function __construct(string $appName,
 		IRequest $request,
-		ContactIntegrationService $service) {
+		ContactIntegrationService $service,
+		ICache $cache) {
```
ChristophWurst, line 30: I've never seen ICache injected. Will this give you a local or distributed cache?

lib/Controller/ContactIntegrationController.php
```diff
@@ -75,8 +80,13 @@ public function newContact(?string $contactName = null, ?string $mail = null): J
...
-	public function autoComplete(string $term): JSONResponse {
-		$res = $this->service->autoComplete($term);
-		return (new JSONResponse($res))->cacheFor(60 * 60, false, true);
+	public function autoComplete(string $term, bool $forceSAB = false): JSONResponse {
+		$cached = $this->cache->get($term);
+		if($cached !== null) {
+			return new JSONResponse(json_decode($cached));
+		}
+		$res = $this->service->autoComplete($term, $forceSAB);
+		$this->cache->set($term, json_encode($res), 7 * 24 * 3600);
```
ChristophWurst, line 89: 1 week is a bit long. Do an hour, that will catch frequently mentioned people and still let changes propagate

lib/Controller/ContactIntegrationController.php
```diff
@@ -75,8 +80,13 @@ public function newContact(?string $contactName = null, ?string $mail = null): J
 	 * @return JSONResponse
 	 */
 	#[TrapError]
-	public function autoComplete(string $term): JSONResponse {
-		$res = $this->service->autoComplete($term);
-		return (new JSONResponse($res))->cacheFor(60 * 60, false, true);
+	public function autoComplete(string $term, bool $forceSAB = false): JSONResponse {
+		$cached = $this->cache->get($term);
```
ChristophWurst, line 84: the cache key is not scoped to a user. you will cache-spill results from one user to another

PR #10128, opened 2024-09-11
refactor: Apply PHP5.6-7.3 rector sets

ChristophWurst:
Commits can be reviewed one by one.

PHP7.4+ comes with lots of changes. We can do them later.


kesselb (approved):
Nice

kesselb (approved):

PR #10129, opened 2024-09-11
refactor: Use Nextcloud sets for rector

ChristophWurst:


lib/Controller/AccountsController.php
```diff
@@ -54,7 +55,7 @@ class AccountsController extends Controller {
 	public function __construct(string $appName,
 		IRequest $request,
 		AccountService $accountService,
-		$UserId,
+		$userId,
```
ChristophWurst, line 58: @kesselb only the argument is renamed, not its usaged :speak_no_evil:

PR #10162, opened 2024-09-17
fix: messages containing an invitation doesnt horizontally scroll

GretaD:
ref #10245

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -213,4 +213,12 @@ export default {
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


src/components/Message.vue
```diff
@@ -213,4 +213,12 @@ export default {
 		}
 	}
 }
+@media screen and (max-width: 500px)
+:deep(.tbody) {
```
GretaD, line 217: :deep(tbody)*

GretaD:
fix here #10252

PR #10176, opened 2024-09-19
fix(threading): Handle threads with duplicate send times

ChristophWurst:
A thread with two messages with identical sent_at result in a problematic situation. The backend will return *both* messages when listing the inbox. The front-end will de-duplicate and only show one of the two. At every browser-server sync, the picked duplicate will be replaced by the other message. At the next sync they are replaced again. Etc. This causes endless notifications.

I ran into this with a thread for an outgoing message that was replied to by the mail server. This seems to be the rare case where two messages in a thread have an identical send time because they were processed in the same second.

## How to test

1. Create a thread with two messages
2. Go into the database, find the thread and change send_at to be identical for both messages
3. Reload the app
4. Wait for the sync

main: the two messages toggle.
here: only one of the messages is loaded and never replaced.

---

Note: I can't come up with a good way to find out which of the messages with identical send time are "newer". I just sort them by message id. It's a bit "random" semantically but keeps the sorting stable.


hamza221 (approved):

st3iny (approved):
Makes sense.

Issue #10089, opened 2024-09-03
*Reply to sender only* replies all


ChristophWurst:
### Steps to reproduce

1. Receive an email with multiple recipients, or an email to a mailing list
2. Click *Reply to sender only*

### Expected behavior

The sender is selected as recipient of the reply.

### Actual behavior

All previous thread participants are selected, or the mailing list in case the original message went to a list.

### Mail app version

4.0.0

### PHP engine version

None

### Web server

None

### Database

None

GretaD:
this is regression of https://github.com/nextcloud/mail/pull/9921

PR #10190, opened 2024-09-25
fix: reply to sender only

GretaD:
fixes https://github.com/nextcloud/mail/issues/10089


GretaD:
nevermind, Christoph already sent a fixup: #10177

PR #10235, opened 2024-10-08
chore: remove psr/log as dependency

kesselb:
The psr/log package is provided by nextcloud server.


kesselb:
Closing in favor of https://github.com/nextcloud/mail/pull/10237

PR #10249, opened 2024-10-11
[stable4.0] build: run tests against stable30

kesselb:


ChristophWurst (commented):
Please change ``app:enable -f mail`` to ``app:enable mail``. AFAIR the force option was only there to make the app installable with "wrong" server after the branch-off

ChristophWurst (approved):

PR #10260, opened 2024-10-15
fix: app version

hamza221:
see comment https://github.com/nextcloud/mail/pull/10243#discussion_r1801017202


ChristophWurst (approved):
This is a *chore* commit because it should not show up in a changelog

PR #10271, opened 2024-10-17
fix: increase default SMTP timeout

miaulalala:
These are the [RFC](https://www.rfc-editor.org/rfc/rfc5321.html#page-65) recommendations:

```
Recommended Timeout Values:
    Initial Connection: Set the timeout for the initial 220 message to at least 5 minutes.
    Mail Commands: Allocate a minimum of 5 minutes for both the MAIL and RCPT commands.
    Data Initiation: Allow at least 2 minutes for initiating the DATA command.
    Data Block: Set a timeout of at least 3 minutes for sending a data block.
    Data Termination: Provide a generous 10-minute timeout for terminating the DATA command.
    Server Timeout: Configure the server to have a timeout of at least 5 minutes.
```

We have to be cognizant of the PHP timeout, so 5 minutes are out.

20 seconds to start, otherwise we need to see what else we can do and if we need to recommend higher PHP timeouts as well as a higher SMTP timeout for the Mail app.

### DIFF
--- a/doc/admin.md
+++ b/doc/admin.md
@@ -40,7 +40,7 @@ Depending on your mail host, it may be necessary to increase your IMAP and/or SM
 ```
 #### SMTP timeout
 ```php
-'app.mail.smtp.timeout' => 2
+'app.mail.smtp.timeout' => 20
 ```
 #### Sieve timeout
 ```php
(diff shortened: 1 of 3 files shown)


ChristophWurst (commented):
:+1: does it influence autodiscovery?

ChristophWurst:
I meant https://github.com/nextcloud/mail/blob/main/lib/Service/AutoConfig/ConnectivityTester.php but the timeout is static

ChristophWurst (approved):

PR #10280, opened 2024-10-22
fix: perform bulk message actions

SebastianKrupinski:
Bulk mail message actions

### DIFF
--- a/lib/Service/MessageOperationService.php
+++ b/lib/Service/MessageOperationService.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Service;
+
+use Horde_Imap_Client_Exception;
+use OCA\Mail\Account;
+use OCA\Mail\Db\MailAccountMapper;
+use OCA\Mail\Db\MailboxMapper;
+use OCA\Mail\Db\MessageMapper;
+use OCA\Mail\IMAP\IMAPClientFactory;
+use OCA\Mail\IMAP\MessageMapper as ImapMessageMapper;
+use OCA\Mail\Service\MailManager;
+
+class MessageOperationService {
+
+    public function __construct(
+        protected IMAPClientFactory $clientFactory,
+        protected MailAccountMapper $accountMapper,
+        protected MailboxMapper $mailboxMapper,
+        protected MessageMapper $messageMapper,
+        protected MailManager $mailManager,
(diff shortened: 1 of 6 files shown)


lib/Service/MessageOperationService.php
```diff
@@ -0,0 +1,153 @@
...
+		// retrieve all mailboxes and group by account
+		$mailboxes = $this->groupByAccount($this->mailboxMapper->findByIds(array_keys($messages)));
+		// retrieve all accounts
+		$accounts = $this->accountMapper->findByIds(array_keys($mailboxes));
+		// process every account
+		$results = [];
+		foreach ($accounts as $account) {
+			$account = new Account($account);
+			// determine if account belongs to the user and skip if not
+			if ($account->getUserId() != $userId) {
```
ChristophWurst, line 108: Good to check, but please move this down into the db when looking up accounts. Make it a WHERE id IN (xxx) AND user_id = 'sebastian'. That avoids loaded any foreign accounts.

ChristophWurst (commented):
Looks good so far

SebastianKrupinski:
Pslam complaints: will be dealt with, once the reviewed

SebastianKrupinski:
Merge: broke front end, reverting to draft

ChristophWurst:
@SebastianKrupinski the PR state says *ready for review* and the label say *2. developing*. What is the status here?

SebastianKrupinski:
> @SebastianKrupinski the PR state says _ready for review_ and the label say _2. developing_. What is the status here?

Yes, its ready to review. It was late at night after along day, when I uploaded this an forgot to change the labels

ChristophWurst (approved):
Looks good

vite.config.js
```diff
@@ -0,0 +1,31 @@
+/**
```
kesselb, line 1: Please remove

lib/Service/MessageOperationService.php
```diff
@@ -0,0 +1,147 @@
...
+	 * @param string $userId system user id
+	 * @param array<int> $identifiers message ids
+	 * @param array<string,bool> $flags message flags
+	 *
+	 * @return array<int,bool> operation results
+	 */
+	public function changeFlags(string $userId, array $identifiers, array $flags): array {
+
+		// retrieve message meta data [uid, mailbox_id] for all messages and group by mailbox id
+		$messages = $this->groupByMailbox($this->messageMapper->findMailboxAndUid($identifiers));
```
kesselb, line 98: Is the `MessageFlaggedEvent` event not needed anymore? 

https://github.com/nextcloud/mail/blob/37e631ec406072797a07091ab5d6c9303e023b39/lib/Service/MailManager.php#L481-L491

kesselb (changes requested):

PR #10293, opened 2024-10-24
chore: bring back needs info workflow

kesselb:
> debug-only: true

The action will not post or add labels for now, only print a list of issues in the action output. 
I will check and change the repository variable then.


.github/workflows/stale.yml
```diff
@@ -0,0 +1,34 @@
...
+        repo-token: ${{ secrets.COMMAND_BOT_PAT }}
+        stale-issue-message: >
+          Hi there! It looks like this issue hasn’t seen much activity for a while and might need a bit more information to move forward.
+          If we don’t hear back soon, we may close it to keep things organized.
+          Thanks again for your contributions – we really appreciate it!
+        stale-issue-label: 'stale'
+        only-labels: 'needs info'
+        labels-to-remove-when-unstale: 'needs info,stale'
+        exempt-issue-labels: '1. to develop,2. developing,3. to review,4. to release,discussion'
+        days-before-stale: 30
```
ChristophWurst, line 31: Let's go higher or the action might tag too many tickets as stale in the next few days. Our issue tracker is still not fully triaged.

How about 90?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,34 @@
...
+    - uses: actions/stale@v9
+      with:
+        repo-token: ${{ secrets.COMMAND_BOT_PAT }}
+        stale-issue-message: >
+          Hi there! It looks like this issue hasn’t seen much activity for a while and might need a bit more information to move forward.
+          If we don’t hear back soon, we may close it to keep things organized.
+          Thanks again for your contributions – we really appreciate it!
+        stale-issue-label: 'stale'
+        only-labels: 'needs info'
+        labels-to-remove-when-unstale: 'needs info,stale'
```
ChristophWurst, line 29: Not sure if the *needs info* label should be rmoved automatically or by the maintainer who asked for information? When something else is commented, or another person comments, the information is not always provided and the label would have to be added back manually.

ChristophWurst (commented):
Thanks

Just some minor comments

.github/workflows/stale.yml
kesselb: > Let's go higher or the action might tag too many tickets as stale in the next few days. Our issue tracker is still not fully triaged.

I think that's fine because the stale action will only comment on issues with the needs info label. Everything else is ignored.

.github/workflows/stale.yml
kesselb: Good catch, thank you!

PR #10301, opened 2024-10-28
feat: per account imap and smtp debugging

SebastianKrupinski:
At the moment there is no ability to debug a single mail account.

This adds the ability to enable debugging from the command line using a occ command on a single account and protocol. This also creates a separate file for each account that is enable instead of logging all accounts to the same log file. ( {data directory}/mail-{user id}-{account id}-imap.log )

// command arguments and options
mail:account:debug [--on] [--off] <account-id>

// enable debugging for imap
mail:account:debug 14 --on

// disable debugging
mail:account:debug 14 --off
mail:account:debug 14

// enable for single run with environmental variable
NC_MAIL_DEBUG=true php occ mail:account:sync 14

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -167,7 +167,7 @@ jobs:
             run: php -f nextcloud/occ app:enable mail
           - name: Configure Nextcloud for testing
             run: |
-                php -f nextcloud/occ config:system:set debug --type bool --value true
+                php -f nextcloud/occ config:system:set app.mail.debug --type bool --value true
                 php -f nextcloud/occ config:system:set app.mail.verify-tls-peer --type bool --value false
           - name: Enable slow mysql query logs
             if: ${{ matrix.db == 'mysql' }}
@@ -192,7 +192,7 @@ jobs:
             run: echo "SELECT * FROM mysql.slow_log WHERE sql_text LIKE '%oc_mail%' AND sql_text NOT LIKE '%information_schema%'" | mysql -h 127.0.0.1 -u root -pmy-secret-pw
           - name: Print debug logs
             if: ${{ always() }}
-            run: cat nextcloud/data/horde_*.log
+            run: cat nextcloud/data/mail-*-*-imap.log
           - name: Report coverage
             uses: codecov/codecov-action@b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 # v4.6.0
             if: ${{ always() && matrix.db == 'mysql' }}
--- a/lib/Command/DebugAccount.php
+++ b/lib/Command/DebugAccount.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Command;
+
+use OCA\Mail\Service\AccountService;
+use OCP\AppFramework\Db\DoesNotExistException;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DebugAccount extends Command {
+	protected const ARGUMENT_ACCOUNT_ID = 'account-id';
+	protected const OPTION_DEBUG_ON = 'on';
+	protected const OPTION_DEBUG_OFF = 'off';
+
+	public function __construct(
+		private AccountService $accountService,
+		private LoggerInterface $logger,
+	) {
+		parent::__construct();
+	}
+
+	/**
+	 * @return void
+	 */
+	protected function configure(): void {
+		$this->setName('mail:account:debug');
+		$this->setDescription('Enable or Disable IMAP/SMTP debugging for an account');
+		$this->addArgument(self::ARGUMENT_ACCOUNT_ID, InputArgument::REQUIRED);
+		$this->addOption(self::OPTION_DEBUG_ON, null, InputOption::VALUE_NONE);
+		$this->addOption(self::OPTION_DEBUG_OFF, null, InputOption::VALUE_NONE);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$accountId = (int)$input->getArgument(self::ARGUMENT_ACCOUNT_ID);
+		$debugOn = $input->getOption(self::OPTION_DEBUG_ON);
+		$debugOff = $input->getOption(self::OPTION_DEBUG_OFF);
+		$debug = false;
+
+		try {
+			$account = $this->accountService->findById($accountId)->getMailAccount();
+		} catch (DoesNotExistException $e) {
+			$output->writeln("<error>Account $accountId does not exist</error>");
+			return 1;
+		}
+
+		if ($debugOn && $debugOff) {
+			$output->writeln('<error>Cannot use both --on and --off at the same time</error>');
+			return 1;
+		}
+
+		if ($debugOn) {
+			$debug = true;
+		}
+
+		$account->setDebug($debug);
+		$this->accountService->save($account);
+
+		return 0;
+	}
+}
--- a/lib/IMAP/IMAPClientFactory.php
+++ b/lib/IMAP/IMAPClientFactory.php
@@ -117,8 +117,9 @@ class IMAPClientFactory {
 				'backend' => $this->hordeCacheFactory->newCache($account),
 			];
 		}
-		if ($this->config->getSystemValue('debug', false)) {
-			$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde_imap.log';
+		if ($account->getDebug() || $this->config->getSystemValueBool('app.mail.debug')) {
+			$fn = 'mail-' . $account->getUserId() . '-' . $account->getId() . '-imap.log';
+			$params['debug'] = $this->config->getSystemValue('datadirectory') . '/' . $fn;
 		}
 
 		$client = new HordeImapClient($params);
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -34,7 +34,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>5.0.0-dev.1</version>
+	<version>5.0.0-dev.2</version>
 	<licence>agpl</licence>
 	<author homepage="https://github.com/ChristophWurst">Christoph Wurst</author>
 	<author homepage="https://github.com/GretaD">GretaD</author>
@@ -80,6 +80,7 @@ Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud
 		<command>OCA\Mail\Command\CleanUp</command>
 		<command>OCA\Mail\Command\CreateAccount</command>
 		<command>OCA\Mail\Command\CreateTagMigrationJobEntry</command>
+		<command>OCA\Mail\Command\DebugAccount</command>
 		<command>OCA\Mail\Command\DeleteAccount</command>
 		<command>OCA\Mail\Command\DiagnoseAccount</command>
 		<command>OCA\Mail\Command\ExportAccount</command>
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -62,6 +62,13 @@ class Account implements JsonSerializable {
 	}
 
 	/**
+	 * @return bool
+	 */
+	public function getDebug(): bool {
+		return $this->account->getDebug();
+	}
+
+	/**
 	 * Set the quota percentage
 	 * @param Quota $quota
 	 * @return void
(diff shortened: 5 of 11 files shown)


SebastianKrupinski:
TODO: Fix integration test

SebastianKrupinski:
> TODO: Fix integration test

Done

lib/IMAP/IMAPClientFactory.php
```diff
@@ -40,6 +40,9 @@ class IMAPClientFactory {
 	private ITimeFactory $timeFactory;
 	private HordeCacheFactory $hordeCacheFactory;
 
+	private int $debugDefault = 1 << 0;   // 1 (0000 0001)
+	private int $debugFull = 1 << 1;     // 2 (0000 0010)
```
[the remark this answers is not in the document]
SebastianKrupinski: Sure. I've removed the bit shifting, we can still just use a regular integer and do a bit comparison.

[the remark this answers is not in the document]
st3iny: That's what I meant. I don't like having a compressed int as a field. That make it hard to understand when encountered outside of the code, e.g. in a database dump.

As this is not performance critical code I would prefer another way, e.g. a serialized JSON string or concatenated string ("imap|smtp") or similar.

st3iny (commented):
I posted a reply in the thread above.

> I meant both. I don't like having a compressed int as a field. That makes it hard to understand when encountered outside of the code, e.g. in a database dump.
>
> As this is not performance critical code I would prefer another way, e.g. a serialized JSON string or concatenated string ("imap|smtp") or similar.

kesselb:
Please change "4.1.0-alpha.2" to "4.1.0-alpha.3" in info.xml to trigger the migration.

kesselb:
> Please change "4.1.0-alpha.2" to "4.1.0-alpha.3" in info.xml to trigger the migration.

You may want to wait until https://github.com/nextcloud/mail/pull/10326 is merged and bump after a rebase.

lib/IMAP/IMAPClientFactory.php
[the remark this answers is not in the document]
SebastianKrupinski: Done, I've changed it to a concatenated string, although I think int would have cleaner.

SebastianKrupinski:
> > Please change "4.1.0-alpha.2" to "4.1.0-alpha.3" in info.xml to trigger the migration.
> 
> You may want to wait until #10326 is merged and bump after a rebase.

Done. I've changed the version to '4.2.0-alpha.1' that way it will migrate properly after the '4.2.0-alpha.0' release.

ChristophWurst (commented):
What I'm missing now is a way to enable Mail debug mode for one occ run.

Before I could do ```php NC_debug=1 occ mail:account:sync ...```. This allowed running one sync in debug mode and leave the rest (of production) untouched. Now I have to globally set the account for debug and it **will** influence production, e.g. through background jobs or when the user has the app open in the browser.

Any idea how we can bring the old trick back?

SebastianKrupinski:
> What I'm missing now is a way to enable Mail debug mode for one occ run.
> 
> Before I could do `php NC_debug=1 occ mail:account:sync ...`. This allowed running one sync in debug mode and leave the rest (of production) untouched. Now I have to globally set the account for debug and it **will** influence production, e.g. through background jobs or when the user has the app open in the browser.
> 
> Any idea how we can bring the old trick back?

So you want to use a single use environmental variable to enable debugging, no problem.

```
NC_MAIL_DEBUG=imap php occ mail:account:sync 14
```

kesselb:
I’d like to suggest removing the option to enable debugging separately for each protocol (like IMAP or SMTP). This would make the setup simpler, with just one debug option instead. I think enabling debugging for individual protocols adds extra complexity without much benefit.

.github/workflows/test.yml
```diff
@@ -184,7 +184,7 @@ jobs:
             run: echo "SELECT * FROM mysql.slow_log WHERE sql_text LIKE '%oc_mail%' AND sql_text NOT LIKE '%information_schema%'" | mysql -h 127.0.0.1 -u root -pmy-secret-pw
           - name: Print debug logs
             if: ${{ always() }}
-            run: cat nextcloud/data/horde_*.log
+            run: cat nextcloud/data/mail-*-*-imap.log
```
kesselb, line 187:
```suggestion
            run: cat nextcloud/data/mail*.log
```

To also print smtp and sieve like before?

lib/Command/DebugAccount.php
```diff
@@ -0,0 +1,75 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DebugAccount extends Command {
+	protected const ARGUMENT_ACCOUNT_ID = 'account-id';
+	protected const OPTION_IMAP_DEFAULT = 'imap';
+	protected const OPTION_IMAP_FULL = 'imap-full';
+	protected const OPTION_SMTP_DEFAULT = 'smtp';
```
kesselb, line 25: What about Sieve? ;)

It's using the SieveClientFactory

lib/IMAP/IMAPClientFactory.php
```diff
@@ -117,8 +117,16 @@ public function getClient(Account $account, bool $useCache = true): Horde_Imap_C
 				'backend' => $this->hordeCacheFactory->newCache($account),
 			];
 		}
-		if ($this->config->getSystemValue('debug', false)) {
-			$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde_imap.log';
+		$debug = array_merge(
```
kesselb, line 120: If we decide to keep the protocol option, then this code should go into a small service or into the account object. 

```mermaid
classDiagram
    class DebugHelper
    DebugHelper : +debugImap(account) bool
    DebugHelper : +debugSmtp(account) bool
    DebugHelper : +debugSieve(account) bool
    DebugHelper : +fileNameImap(account) string
    DebugHelper : +fileNameSmtp(account) string
    DebugHelper : +fileNameSieve(account) string
```

lib/IMAP/IMAPClientFactory.php
```diff
@@ -117,8 +117,16 @@ public function getClient(Account $account, bool $useCache = true): Horde_Imap_C
...
+			explode('|', $this->config->getSystemValueString('MAIL_DEBUG')),
```
kesselb, line 121: I'm aware that getSystemValueString is used here as a workaround for reading environment variables, but I have some concerns:

- Although we don't have strict conventions for naming configuration variables, all variables in [config.sample.php](https://github.com/nextcloud/server/blob/master/config/config.sample.php) use lowercase. I’d prefer to keep this consistent.
- For configuration options related to SMTP in [config.sample.php](https://github.com/nextcloud/server/blob/master/config/config.sample.php), we currently use the mail_ prefix. While smtp_ might be more accurate, it’s not feasible to change this now :disappointed: 
- Our standard format is usually app.mail.xyz, but this could be problematic when working with NC_ environment variables due to naming restrictions in Linux (see [allowed characters in Linux environment variable names](https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names)).

SebastianKrupinski:
> I’d like to suggest removing the option to enable debugging separately for each protocol (like IMAP or SMTP). This would make the setup simpler, with just one debug option instead. I think enabling debugging for individual protocols adds extra complexity without much benefit.

Well the logic is that you only diagnose issues in one protocol at a time anyway, so there was no point in logging protocols you are not interested in. Lets see what @ChristophWurst and @st3iny think.

.github/workflows/test.yml
SebastianKrupinski: This would kind of defeat the purpose of per account logging, as all account smtp transmissions would end up in the same file?

lib/IMAP/IMAPClientFactory.php
SebastianKrupinski: Sure, that can be done. Let see what the rest think.

lib/IMAP/IMAPClientFactory.php
SebastianKrupinski: - 1 sure I can change it to lower case

- 2 yes I could make it support mail_ and smtp_ or anything else and this would enable global debugging, but the idea behind this PR, is to NOT touch the config file at all for mail communication issues, also to NOT enable debugging for all user(s) mail account(s) at the same time and to separate the transmission logs so that its easier and faster to find issues. Just imagine if you have a system with 100 mail accounts and only 2 accounts have an intermittent problem a few times a day. 
yes I know that by adding the 'getSystemValueString' you can now set the debugging in the config.php, but the purpose was for one time environmental variables.

- 3 yes that is why it has to stay with an under sore, the only other option is one word

.github/workflows/test.yml
SebastianKrupinski: NVM. I miss understood this, sure I can change this.

lib/IMAP/IMAPClientFactory.php
kesselb: > For configuration options related to SMTP in [config.sample.php](https://github.com/nextcloud/server/blob/master/config/config.sample.php), we currently use the mail_ prefix. While smtp_ might be more accurate, it’s not feasible to change this now 😞

I was referring to the symfony mailer configuration in Nextcloud using `mail_` as prefix and thus consider mail_debug not an ideal choice.

kesselb: > yes I know that by adding the 'getSystemValueString' you can now set the debugging in the config.php, but the purpose was for one time environmental variables.

In a previous version, you were using $_ENV / getenv, why did you withdraw this approach? I think that would be cleaner than hijacking SystemConfig :pirate_flag:

SebastianKrupinski: > I was referring to the symfony mailer configuration in Nextcloud using `mail_` as prefix and thus consider mail_debug not an ideal choice.

Ah, Okay, yeah we can still change it to mail_app_debug or anything else.

> In a previous version, you were using $_ENV / getenv, why did you withdraw this approach? I think that would be cleaner than hijacking SystemConfig 🏴‍☠️

I agree it was cleaner but abandoned the approach, ironically because I thought it would get flagged during the review process, for one of two reasons,

- A. Because I was overriding account data in the MailAccount class with information from a different source. Didn't think this was best practice.
- B. Because I was using raw system values

ChristophWurst:
Let's simplify and get this in

1. Only one debug mode for both IMAP and SMTP and no "full" mode, it's rarely needed (https://github.com/nextcloud/mail/pull/10301#issuecomment-2463033797)
2. Revive an option to trigger debug mode from an env var, maybe NC_MAIL_DEBUG=true

SebastianKrupinski:
> Let's simplify and get this in
> 
>     1. Only one debug mode for both IMAP and SMTP and no "full" mode, it's rarely needed ([feat: per account imap and smtp debugging #10301 (comment)](https://github.com/nextcloud/mail/pull/10301#issuecomment-2463033797))
> 
>     2. Revive an option to trigger debug mode from an env var, maybe NC_MAIL_DEBUG=true

Done. Rebase also since this was 5 months old now

lib/Command/DebugAccount.php
```diff
@@ -0,0 +1,71 @@
...
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DebugAccount extends Command {
+	protected const ARGUMENT_ACCOUNT_ID = 'account-id';
+	protected const OPTION_DEBuG_ON = 'on';
```
ChristophWurst, line 23: U

ChristophWurst (commented):
image

ChristophWurst (approved):
Tested & works

lib/Command/DebugAccount.php
```diff
@@ -0,0 +1,71 @@
...
+	) {
+		parent::__construct();
+	}
+
+	/**
+	 * @return void
+	 */
+	protected function configure(): void {
+		$this->setName('mail:account:debug');
+		$this->setDescription('Enable or Disable IMAP/SMTP debugging on a account');
```
kesselb, line 38:
```suggestion
		$this->setDescription('Enable or Disable IMAP/SMTP debugging for an account');
```

lib/IMAP/IMAPClientFactory.php
```diff
@@ -117,8 +117,9 @@ public function getClient(Account $account, bool $useCache = true): Horde_Imap_C
 				'backend' => $this->hordeCacheFactory->newCache($account),
 			];
 		}
-		if ($this->config->getSystemValue('debug', false)) {
-			$params['debug'] = $this->config->getSystemValue('datadirectory') . '/horde_imap.log';
+		if ($account->getDebug() || $this->config->getSystemValueBool('mail_debug') || $this->config->getSystemValueBool('MAIL_DEBUG')) {
```
kesselb, line 120:
```suggestion
		if ($account->getDebug() || $this->config->getSystemValueBool('app.mail.debug') || $this->config->getSystemValueBool('MAIL_DEBUG')) {
```

As mentioned before, `mail_` in config.php is used for the server's symfony mailer configuration. We are using `app.mail.` for the [other options](https://github.com/nextcloud/mail/blob/main/doc/admin.md), so we should use it here as well to maintain a bit of consistency. 😉

st3iny (approved):
Tested and works. I applied all suggestions, rebased and squashed.

SebastianKrupinski:
> Tested and works. I applied all suggestions, rebased and squashed.

This app.mail.debug will not work form the command line!!!!

lib/Command/DebugAccount.php
SebastianKrupinski: Done. https://github.com/nextcloud/mail/pull/11011

PR #10321, opened 2024-11-05
fix(contactsintegration): Limit number of search results

ChristophWurst:
Searching with few characters in large address books or system address books can yield lots of results. This can exhaust the memory and a user likely will search more specifically anyway.

Fixes ```Allowed memory size of 1073741824 bytes exhausted (tried to allocate 4096 bytes) at /var/www/html/nextcloud/apps/dav/lib/CardDAV/CardDavBackend.php```


st3iny (approved):
LGTM!

Issue #10159, opened 2024-09-16
Unicode tags not supported


the-djmaze:
### Steps to reproduce

1. Create tag/keyword/label in a different app
2. Open NC Mail app

### Expected behavior

See tags like "🧪" or "ກ ຂ ຄ"

### Actual behavior

![afbeelding](https://github.com/user-attachments/assets/529b2d28-50da-45f1-b7cc-013bf26785e9)

### Mail app version

3.7.8

### PHP engine version

None

### Web server

None

### Database

None

kesselb:
Thanks @the-djmaze for reaching out :+1: 

I was just looking for another IMAP client that supports tags the other day. It's good to know that snappymail does.

Our handling of tags/flags in Nextcloud Mail is strange. The reason is, that horde/imap_client lowercases all flags by default and breaks the encoding. We are trying to be smart and uppercase everything before converting from utf7-imap to utf-8, but that doesn't work too well[^1].

As a first step, we need to change, or add a new setter/getter, to read the unmodified flags: https://github.com/horde/Imap_Client/blob/98a4ba591bc4fa12341684d4051b344579a34e6f/lib/Horde/Imap/Client/Data/Fetch.php#L328-L331

I don't know if there's a reason that horde lowercases all flags by default.

cc @ChristophWurst @miaulalala 

[^1]: https://3v4l.org/A67V9#v8.3.11

miaulalala:
that's also kinda bad with registered IANA flags such as `$Junk` and `$NotJunk` - they're registered properties and should be treated as such by any mail client... related: https://github.com/nextcloud/mail/issues/10131

the-djmaze:
> that's also kinda bad with registered IANA flags such as `$Junk` and `$NotJunk` - they're registered properties and should be treated as such by any mail client... related: #10131

Yes, but what if there are tags in use `$junk` and `$notjunk`?

IMAP Formal Syntax says:
> Except as noted otherwise, all alphabetic characters are case-insensitive.  The use of upper or lower case characters to define token strings is for editorial clarity only.  Implementations MUST accept these strings in a case-insensitive fashion.

So $MDNSent, $mdnsENT, $MdNsEnT, $MdnSent are compliant.

Lowercase flags is ok, but should be done AFTER `Horde_Imap_Client_Utf7imap::Utf7ImapToUtf8()`

To prevent issues with servers that are case-sensitive (so wrong), setting a flag should be done with IANA registration `$MDNSent`

PR #10328, opened 2024-11-05
fix: Junk/NotJunk flags

SebastianKrupinski:
### Issue

Junk and NotJunk flags where not being set due to a invalid comparison. UI was sending "$junk" but it was being compared to "junk".
 
Screenshot 2024-11-05 181610

Allowed flags was also missing "notjunk"

image

### DIFF
--- a/lib/Service/MailManager.php
+++ b/lib/Service/MailManager.php
@@ -48,17 +48,15 @@ use function array_values;
 
 class MailManager implements IMailManager {
 	/**
-	 * https://tools.ietf.org/html/rfc3501#section-2.3.2
+	 * https://datatracker.ietf.org/doc/html/rfc9051#name-flags-message-attribute
 	 */
-	private const ALLOWED_FLAGS = [
+	private const SYSTEM_FLAGS = [
 		'seen' => [Horde_Imap_Client::FLAG_SEEN],
 		'answered' => [Horde_Imap_Client::FLAG_ANSWERED],
 		'flagged' => [Horde_Imap_Client::FLAG_FLAGGED],
 		'deleted' => [Horde_Imap_Client::FLAG_DELETED],
 		'draft' => [Horde_Imap_Client::FLAG_DRAFT],
 		'recent' => [Horde_Imap_Client::FLAG_RECENT],
-		'junk' => [Horde_Imap_Client::FLAG_JUNK, 'junk'],
-		'mdnsent' => [Horde_Imap_Client::FLAG_MDNSENT],
 	];
 
 	/** @var IMAPClientFactory */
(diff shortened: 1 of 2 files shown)


ChristophWurst:
How do we test this?

Found https://github.com/nextcloud/mail/pull/5370 to change ``junk`` to ``$junk``

src/store/actions.js
```diff
@@ -998,8 +998,8 @@ export default {
 
 			try {
 				await setEnvelopeFlags(envelope.databaseId, {
-					$junk: !oldState,
-					$notjunk: oldState,
```
ChristophWurst, line 1002: https://github.com/nextcloud/mail/blob/1d5e5b49c1485e9be70f22c4df9526805f5238a7/src/store/actions.js#L991, https://github.com/nextcloud/mail/blob/1d5e5b49c1485e9be70f22c4df9526805f5238a7/src/store/actions.js#L1011-L1021 and similar stay?

SebastianKrupinski: Yes, The UI handles the "$junk" but the backend compares the it as "junk", so in the UI they can all stay as "$junk" but need to have the $ sign stripped of when sent to the backend. The second image of allowed flags is comparison table, the first image is how they are compared/found.

SebastianKrupinski:
> How do we test this?

The best way to test is to insert a break point in the OCA\Mail\Service\MailManager::filterFlags() function. That way you can see the values for the table and flag.

> Found #5370 to change `junk` to `$junk`

This correctly added the front end logic,

```
commit('flagEnvelope', {
	envelope,
	flag: '$notjunk',
	value: !oldState,
})
```

But incorrectly changed the back end logic,

```
setEnvelopeFlag(envelope.databaseId, '$junk', !oldState).catch((e) => {
```

src/store/actions.js
SebastianKrupinski: Sure, Its actual the line you highlighted,

Before:
```
await setEnvelopeFlags(envelope.databaseId, {
	$junk: !oldState,
	$notjunk: oldState,
})
```
After:
```
await setEnvelopeFlags(envelope.databaseId, {
	junk: !oldState,
	notjunk: oldState,
})
```

kesselb:
Imap Flags, my favorite topic ;)

Please don't spend too much time on this one. We are aware about a couple of issues with tags (e.g. https://github.com/nextcloud/mail/issues/10159), and it's on my to-do to work on it when I'm done with files_emailviewer and sieve. 

Your finding is correct, $junk is not part of the allowed flags and therefore the check will never return true. The list of allowed flags should only contain the system flags, which are mandatory from the RFC: https://www.ietf.org/rfc/rfc9051.html#name-flags-message-attribute. 

For system flags, we can skip the check if permflags are enabled and save a few bits. 

If Junk/NotJunk was not added on your system, I assume permflags are not enabled. Can you check that with the debugger?

kesselb:
> Which brings me to the second logic error, the function then checks if '\ *' exists which is NOT an indicator for system flags or keywords, '\ *' 'indicates that custom flags/keywords (tags/labels) are supported.

:+1: 

We currently don’t handle cases where custom keywords are not allowed, but the keyword already exists.

Here’s what I suggest:

- Keep $junk as is.
- Remove non-system flags from allowed flags.
- Introduce a new method in MailManager to check if a given keyword exists or if we can create a new one.

Let me know your thoughts or if you have any suggestions on this approach!

SebastianKrupinski:
> * Keep $junk as is.
> 
> * Remove non-system flags from allowed flags.
> 
> * Introduce a new method in MailManager to check if a given keyword exists or if we can create a new one.

Here is my thoughts,

1. We should change '$junk' in the UI and send it as 'junk', because all the other flags get sent without the $ or \ signs, this was the only one that was send to the back end differently. Otherwise we need to have all the flags sent with the sign from the UI.
2. I agree we show remove the keywords from the allowed flags. But we could also leave them as they are RFC defined and defined in horde (This would basically saves a request to the server for PERMANENTFLAGS check on every flag change, but there is no guarantee they are supported, although I would put my money that most systems support them).
3. This can be addressed pretty easily in the same function with a simple if statement, pull the supported flags then check for '\\ *' and if that fails check if the specific flag/keyword it supported. (I'll make these changes, it will be a quick modification)

I also have some thoughts on how we can make this way more efficient and work better. But I'll DM you with the idea, so that we don't pollute this chat.

SebastianKrupinski:
@kesselb made the changes we discussed. The filterFlags method now handles all 3 cases, system flags, server defined/supported keywords and server custom keywords

kesselb (approved):

PR #10342, opened 2024-11-12
fix(deps): Apply npm audit fix

nextcloud-command:
Auto-generated fix of npm audit


ChristophWurst:
@GretaD @kesselb please give this a smoke test for nextcloud-vue, and then let's get this in

GretaD (approved):
looks good

PR #10390, opened 2024-11-22
fix(contactsintegration): Limit number of matches

ChristophWurst:
Prevents the composer autocompletion from listing *all* matches. For large address books or system address books this can slow down the browser significantly.


ChristophWurst:
@GretaD @kesselb :wave:

PR #10391, opened 2024-11-22
fix(release): Specify release committer

ChristophWurst:
I'm doing this as `fix` not `chore` so it also shows up in changelogs.


GretaD (approved):

kesselb:
What about stable4.1?

PR #10403, opened 2024-11-23
fix: disable async vue components

kesselb:
Screenshot From 2024-11-23 22-18-48

I'm seeing the famous "can't redefine non-configurable property $router" on my dev setup (main) and on c.nc.c (stable4.1).

Downside is that mail-main.js is now 9,803.38 kB :fearful: 

Before: 
```
./js/Upload-CiIThrPS.chunk.mjs                              1.56 kB │ gzip:   0.84 kB │ map:     1.30 kB
./js/Setup-B65pp2sL.chunk.mjs                               2.78 kB │ gzip:   1.21 kB │ map:     2.79 kB
./js/randomId-BnxyHsj7.chunk.mjs                            5.99 kB │ gzip:   1.78 kB │ map:    14.04 kB
./js/NcCheckboxRadioSwitch-l-6YVD9V-DzPQuco_.chunk.mjs     31.13 kB │ gzip:   6.20 kB │ map:    50.59 kB
./js/AccountForm-BmpEnP8z.chunk.mjs                        32.84 kB │ gzip:   6.75 kB │ map:    48.36 kB
./js/FilePicker-ajWx2Snh-mBxkLcb8.chunk.mjs                90.94 kB │ gzip:  20.73 kB │ map: 3,063.28 kB
./js/NewMessageModal-D4XCzgv-.chunk.mjs                   121.14 kB │ gzip:  26.76 kB │ map:   187.80 kB
./js/AccountSettings-CDiuapQM.chunk.mjs                   228.25 kB │ gzip:  40.23 kB │ map:   330.50 kB
./js/Nextcloud-COIclUaX.chunk.mjs                         315.16 kB │ gzip:  79.50 kB │ map:   690.04 kB
./js/index-BvQHV1fZ.chunk.mjs                           1,370.16 kB │ gzip: 231.08 kB │ map: 2,976.78 kB
./js/mail-main.mjs                                      1,536.75 kB │ gzip: 343.43 kB │ map: 3,134.61 kB
./js/Home-pbw81Vm5.chunk.mjs                            1,553.83 kB │ gzip: 293.28 kB │ map: 2,621.17 kB
./js/TextEditor-n_Vsblu4.chunk.mjs                      2,021.52 kB │ gzip: 401.91 kB │ map: 4,356.37 kB
./js/Navigation-C6biF-u5.chunk.mjs                      4,276.88 kB │ gzip: 724.46 kB │ map: 5,887.16 kB
```

After:
```
[...]

PR #10406, opened 2024-11-25
fix(build): Disable async components with Vite

ChristophWurst:


ChristophWurst:
Duplicate of https://github.com/nextcloud/mail/pull/10403

Issue #10392, opened 2024-11-22
Mail envelope and titles design is off


thstyl2000:
### Steps to reproduce

1. Click on a message
2. Look at the email

### Expected behavior

Mail body, Subject and Reply button should be nicely aligned.

### Actual behavior

Things currently look like that:

![Image](https://github.com/user-attachments/assets/41df6387-338c-42c4-af30-d7e66f2ae021)

![Image](https://github.com/user-attachments/assets/4ff0e82d-b93b-45ef-99f9-92506337d415)

### Mail app version

4.0.3

### Nextcloud version

30.0.2

### Mailserver or service

imap (dovecot)

### Operating system

freebsd

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database

MariaDB

ChristophWurst:
Thanks for this report too @thstyl2000! I noticed it on my phone yesterday. This is a regression and not a deliberate design change.

thstyl2000:
Most welcome! Glad to be of any help.

ChristophWurst:
This was caused by https://github.com/nextcloud/mail/pull/10268. Some css selectors no longer apply.

ChristophWurst:
Fix is at https://github.com/nextcloud/mail/pull/10410

thstyl2000:
Sorry to bump, the Reply button is still (4.0.7) off it now moves up and down depending on the length of the message.

![Image](https://github.com/user-attachments/assets/20345d82-368b-48b3-8f23-dc8c16c7a03a)

![Image](https://github.com/user-attachments/assets/af0a48f2-596b-48d8-a2db-3405c791110f)



Also still thing that the title is way too big. The screenshot in github looks miles better in my opinion...

thstyl2000:
Just noticed that the subject used to be aligned with the message text, and now not (that's why it looked so off to me). Should I open a new issue/feature for this and the reply button?

thstyl2000:
Still (4.0.9.) an issue when phishing warning is shown:

![Image](https://github.com/user-attachments/assets/9c8785b7-669e-4f18-ac93-761a6382cc5d)

PR #10410, opened 2024-11-26
fix(ui): Restore message/thread styling

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/10392

Reverts https://github.com/nextcloud/mail/pull/10268 and adds a more specific fix for the leaked popover styles.

This is still a bodge. Styles should be scoped. We have to move some rules around.

### DIFF
--- a/src/components/RecipientBubble.vue
+++ b/src/components/RecipientBubble.vue
@@ -265,7 +265,6 @@ export default {
 }
 .contact-popover {
 	display: flex;
-	padding: 2px;
 
 	&__email {
 		text-align: center;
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@GretaD @kesselb let's get this in and released

GretaD:
the recipient bubble is cut a bit because of this padding:
`
.contact-popover[data-v-31ad87b1] {
    display: flex;
    padding: 2px;
}`

Screenshot from 2024-11-28 15-08-16

ChristophWurst:
Indeed! Looks better without the padding :)

GretaD (approved):

PR #10437, opened 2024-12-02
feat: make multiselect menu more discoverable

### LINKED ISSUE #10231: More discoverable multi-select and bulk operations
**Problem:**

Users didn't know that it is possible to select multiple messages and perform bulk operations such as move, archive, marks as spam or delete.

GretaD:
fixes #10231

hovered
Screenshot from 2024-12-09 14-47-01

selected
Screenshot from 2024-12-09 14-49-25

To clarify
- [x] Should the checkmark be shown when the whole message is hovered, or when the avatar is hovered. Right now, the check mark is shown when the avatar is hovered.

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -6,208 +6,101 @@
 	<div>
 		<transition name="multiselect-header">
 			<div v-if="selectMode" key="multiselect-header" class="multiselect-header">
-				<div class="button primary" @click.prevent="markSelectedSeenOrUnseen">
-					<span id="action-label">{{
-						areAllSelectedRead
-							? n(
-								'mail',
-								'Mark {number} unread',
-								'Mark {number} unread',
-								selection.length,
-								{
-									number: selection.length,
-								}
-							)
-							: n(
-								'mail',
-								'Mark {number} read',
-								'Mark {number} read',
-								selection.length,
-								{
-									number: selection.length,
-								}
-							)
-					}}</span>
-				</div>
-				<Actions class="app-content-list-item-menu" menu-align="right">
-					<ActionButton v-if="isAtLeastOneSelectedUnimportant"
-						:close-after-click="true"
+				<div class="action-buttons">
+					<NcButton v-if="isAtLeastOneSelectedUnread"
+						type="tertiary"
+						:title="n('mail', 'Mark {number} read', 'Mark {number} read', selection.length, { number: selection.length })"
+						@click.prevent="markSelectedRead">
+						<EmailRead :size="16" />
+					</NcButton>
+
+					<NcButton v-if="isAtLeastOneSelectedRead"
+						type="tertiary"
+						:title="n('mail', 'Mark {number} unread', 'Mark {number} unread', selection.length, { number: selection.length })"
+						@click.prevent="markSelectedUnread">
+						<EmailUnread :size="16" />
+					</NcButton>
+
+					<NcButton v-if="isAtLeastOneSelectedUnimportant"
+						type="tertiary"
+						:title="n('mail', 'Mark {number} as important', 'Mark {number} as important', selection.length, { number: selection.length })"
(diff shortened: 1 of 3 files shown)


GVodyanov (commented):
image

On dark mode the select icon is the wrong color

GretaD:
> image
> 
> On dark mode the select icon is the wrong color

oh its on both modes actually, sorry, will fix soon

GVodyanov (approved):

src/components/EnvelopeList.vue
```diff
@@ -674,4 +568,7 @@ div {
 		display: block;
 	}
 }
+.action-buttons {
```
ChristophWurst, line 571: this is a child element of `.multiselect-header`. Move it in there to get into a BEM-like structure?

src/components/EnvelopeList.vue
```diff
@@ -6,208 +6,91 @@
...
-								}
-							)
-					}}</span>
-				</div>
-				<Actions class="app-content-list-item-menu" menu-align="right">
-					<ActionButton v-if="isAtLeastOneSelectedUnimportant"
-						:close-after-click="true"
+				<div class="action-buttons">
+					<NcButton type="tertiary"
+						:title="areAllSelectedRead ? n('mail', 'Mark {number} unread', 'Mark {number} unread', selection.length, { number: selection.length }) : n('mail', 'Mark {number} read', 'Mark {number} read', selection.length, { number: selection.length })"
```
ChristophWurst, line 11: nit: the previous multi-line wrap made the `n` function call more readable than this one-liner

ChristophWurst:
* If all selected messages are read, there is a bulk *Mark us unread*. If at least one message is unread there is only a bulk *Mark as read* and no more *Mark as unread*.
* There is no option to un-flag (un-star) a message (was not there before either)
* Before it was possible to clear the selection. This feature is gone.
* The expansion of the multiselect buttons makes the list move and at least on Firefox it's a bit jittery.
* The hover state of the avatar of a selected message is identical to a message that is not selected. This means you get no click feedback when selecting a message until you leave the element again.

image

image

GretaD:
> * If all selected messages are read, there is a bulk _Mark us unread_. If at least one message is unread there is only a bulk _Mark as read_ and no more _Mark as unread_.

Same scenario as favorite, unfavorite, explained below

> * There is no option to un-flag (un-star) a message (was not there before either)

It is there, i added it, but its one button that replaces each other. I just changed the icon for both states.
Screenshot from 2024-12-10 12-56-06
Screenshot from 2024-12-10 12-54-57

> * Before it was possible to clear the selection. This feature is gone.

You mean _unselect_, right? Its the second icon before last.

> * The expansion of the multiselect buttons makes the list move and at least on Firefox it's a bit jittery.

I'll test it there and fix it.

> * The hover state of the avatar of a selected message is identical to a message that is not selected. This means you get no click feedback when selecting a message until you leave the element again.
> 
Yes, you need to select the avatar to actually select it. Selecting the message will open the message and should not select it( :heavy_check_mark:  as an action). These 2 actions should be split, you shouldnt select :heavy_check_mark: a message when you only want to open it. So, i would say the hover state of the avatar should be the same, regardless of the state of the message, but, i can change it if you have an idea how.

I havent changed anything from the logic that was there. The behavior is the same as the original menu actions, i only added new icons. So, you propose to have both buttons out like in the _mark as important_ example? When you have both actions selected, like for example, one favorite, and one not favorite, to show both icons, favorite and unfavorite?
In general, you dont expect people to do that.

> image

ChristophWurst:
> When you have both actions selected, like for example, one favorite, and one not favorite, to show both icons, favorite and unfavorite?
> In general, you dont expect people to do that.

Yes. For *Mark important/unimportant* we also have both options

SebastianKrupinski (approved):
Tested. Looks and works great. Much better visually, I think.

image

Issue #10386, opened 2024-11-21
Translate email


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud user I want to be able to understand emails sent in languages I don't speak.

### Describe the solution you'd like

Once I open a message, add an action *Translate* that puts the email text into a translation modal. The modal should be the one that is also used in Talk.
I only need to see the translated text in this modal, with the option to copy, without any further actions in mail.

### Describe alternatives you've considered

N/a

### Additional context

N/a

GVodyanov:
What should I do about HTML messages? I could possibly use some funky regex and extract the text into the modal, then translate it in there as plain text. Or maybe I could try to reinsert the translations back into the HTML structure and try to display that (either in the modal, or in the original `ThreadEnvelope`). Or I could just not give the option for them to be translated (what I currently set).

ChristophWurst:
If it doesn't exist already we can add an API endpoint that gives access to a text-only version of an email. AFAIK that is also the input for thread summaries, so we have that, but possibly not exposed to the frontend.

ChristophWurst:
@GVodyanov we should also cover this new feature in the user manual and add a note that it's only available for instances that have translation apps/services. In the admin manual we can mention it too with a pointer to the translation services admin documentation (I hope it exists).

ChristophWurst:
Reopening as a reminder for documentation.

PR #10453, opened 2024-12-08
Add mail translation

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/10386


src/service/translationService.js
```diff
@@ -0,0 +1,68 @@
...
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+import store from '../store/index.js'
+
+const checkTranslationServiceAvailability = function() {
+	/// TODO
```
GVodyanov, line 11: @kesselb @SebastianKrupinski Hey! I got this endpoint from Marcel, but I don't have the means to test the translation API yet. Could you please do me a favor: check out the endpoint response and adjust the function appropriately? 

Apart from that this PR is ready, all the front-end logic works

SebastianKrupinski: Fixed it.

SebastianKrupinski:
Tested Works.

image

There are a few improvements I would make. 
- The from selector, I would put in a Detect Automatically option, as a user might not know what the original language is.
- The best option would be to get rid of the from text selector, and just make the AI determine what the original language is.
- Since we are using AI for this, don't limit the translation languages, I only have to option to translate to 3 languages, German, French and English. Also German is missing in the from selector.
- Lower the time interval that checks if the task processing job is completed. On my dev system the job is completed in 2s but the UI takes almost 30s to display the translation.
- Hide the original source once the translation is completed. I need to scroll past the original text first before I can see the translation.

src/service/translationService.js
```diff
@@ -0,0 +1,68 @@
...
+			input: {
+				origin_language: fromLanguage ?? null,
+				input: text,
+				target_language: toLanguage,
+			},
+			type: 'core:text2text:translate',
+			appId: 'mail',
+		})
+		const task = scheduleResponse.data.ocs.data.task
+		const getTaskOutput = async (task) => {
```
hamza221, line 54: I'm not a big fan of the recursion here, But I'm not sure how to go around it. I tried utilising the `completionExpectedAt` locally but it's unreliable.
Also the task can take really long, maybe we should allows user to reduce the modal and get a notification once its done , the way that NC assistant works ? 
But maybe as a followup

hamza221:
I Like the UI ✅

src/service/translationService.js
```diff
@@ -0,0 +1,68 @@
...
+import store from '../store/index.js'
+
+const checkTranslationServiceAvailability = function() {
+	/// TODO
+	// const response = await axios.post(generateOcsUrl('taskprocessing/tasktypes'), {
+	// 	appId: 'mail',
+	// })
+
+	const isAvailable = true
+	const languages = [
```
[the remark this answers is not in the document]
GVodyanov: @hamza221 Yes, that's what my previous ping for Sebastian and Daniel was, if you could please see the response and adjust the function accordingly you would do me a huge favor, I don't have an LLM set up yet and no key.

src/service/translationService.js
GVodyanov: This recursion was copied from the text app https://github.com/nextcloud/text/pull/6567, I don't know if there is some better standard nextcloud way of doing it, but it doesn't seem so.

Sure, the modal would be good as a followup

src/store/mutations.js
```diff
@@ -519,4 +519,10 @@ export default {
 	showSettingsForAccount(state, accountId) {
 		state.showAccountSettings = accountId
 	},
+	enableTranslation(state, enabled) {
```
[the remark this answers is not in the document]
GVodyanov: @hamza221 Does this require a backend addition though?

src/service/translationService.js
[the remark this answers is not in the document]
hamza221: on it

src/service/translationService.js
hamza221: sure,I'll dismiss the review until a better approach arrises

src/service/translationService.js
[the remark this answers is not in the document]
SebastianKrupinski: > @hamza221 Yes, that's what my previous ping for Sebastian and Daniel was, if you could please see the response and adjust the function accordingly you would do me a huge favor, I don't have an LLM set up yet and no key.

Hey. I already did this for you.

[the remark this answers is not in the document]
hamza221: did you push it, languages are still hardcoded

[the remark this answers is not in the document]
hamza221: fixed

src/service/translationService.js
```diff
@@ -0,0 +1,48 @@
...
+			if (task.output) {
+				return task.output.output
+			}
+			await new Promise(resolve => setTimeout(resolve, 2000))
+			const taskResponse = await axios.get(generateOcsUrl(`taskprocessing/task/${task.id}`))
+			return getTaskOutput(taskResponse.data.ocs.data.task)
+		}
+		return await getTaskOutput(task)
+	} catch (e) {
+		console.error('Failed to translate', e)
```
ChristophWurst, line 44: I'm missing either a fallback return value or rethrowing the error. The caller will just receive `undefined`

ChristophWurst (commented):
Looks good

src/service/translationService.js
hamza221: got rid of the try catch since the whole function is inside a try catch

ChristophWurst (approved):

PR #10486, opened 2024-12-12
perf: reduce number of avatar requests

hamza221:


lib/IMAP/PreviewEnhancer.php
```diff
@@ -34,14 +35,24 @@ class PreviewEnhancer {
 	/** @var LoggerInterface */
 	private $logger;
 
+	/** @var AvatarService */
+	private $avatarService;
+
+	/** @var string */
+	private $UserId;
```
ChristophWurst, line 42:
```suggestion
	/** @var string|null */
	private $userId;
```

lib/IMAP/PreviewEnhancer.php
```diff
@@ -34,14 +35,24 @@ class PreviewEnhancer {
...
+
 	public function __construct(IMAPClientFactory $clientFactory,
 		ImapMapper $imapMapper,
 		DbMapper $dbMapper,
-		LoggerInterface $logger) {
+		LoggerInterface $logger,
+		AvatarService $avatarService,
+		string $UserId) {
```
ChristophWurst, line 49:
```suggestion
		string $userId) {
```

https://docs.nextcloud.com/server/latest/developer_manual/basics/dependency_injection.html#predefined-core-services

lib/IMAP/PreviewEnhancer.php
```diff
@@ -50,9 +61,13 @@ public function __construct(IMAPClientFactory $clientFactory,
 	 * @return Message[]
 	 */
 	public function process(Account $account, Mailbox $mailbox, array $messages): array {
-		$needAnalyze = array_reduce($messages, static function (array $carry, Message $message) {
+		$needAnalyze = array_reduce($messages, function (array $carry, Message $message) {
 			if ($message->getStructureAnalyzed()) {
 				// Nothing to do
+				if ($message->getAvatar() === null) {
+					$avatar = $this->avatarService->getAvatar($message->getFrom()->first()->getEmail(), $this->UserId);
```
ChristophWurst, line 68: getFrom might return an empty error. will that lead to a NPE when you call `getEmail` on null?

ChristophWurst (commented):
Quick review

hamza221:
I broke RecipientBubble 😨, turning to draft

hamza221:
The avatar controller and frontend service are still needed for Recipient bubble (Thread.vue) , because the message object contains only from's avatar ( to, cc, bcc ) still need to be fetched , and assuming that we need them and pre-fetching is counterproductive because we can't assume that all threads are gonna be opened

lib/IMAP/PreviewEnhancer.php
```diff
@@ -50,9 +61,14 @@ public function __construct(IMAPClientFactory $clientFactory,
 	 * @return Message[]
 	 */
 	public function process(Account $account, Mailbox $mailbox, array $messages): array {
-		$needAnalyze = array_reduce($messages, static function (array $carry, Message $message) {
+		$needAnalyze = array_reduce($messages, function (array $carry, Message $message) {
```
ChristophWurst, line 64: The purpose of this `array_reduce` is to find messages that need processing.

Placing the avatar logic in it is convenient but a hack. Please put this somewhere else.

I also believe that the preview enhancer is run for background job and CLI syncs. Doesn't this slow them down while the volatile avatar result is not used? Only REST calls to get the message/thread lists should have this avatar pre-fetching.

lib/IMAP/PreviewEnhancer.php
```diff
@@ -59,6 +70,17 @@ public function process(Account $account, Mailbox $mailbox, array $messages): ar
 			return array_merge($carry, [$message->getUid()]);
 		}, []);
 
+		// If we are in the API call, we need to fetch the avatar for the sender
+		if ($mode === 'apiCall') {
+			foreach ($messages as $message) {
+				$from = $message->getFrom()->first() ;
```
ChristophWurst, line 76:
```suggestion
				$from = $message->getFrom()->first();
```

lib/IMAP/PreviewEnhancer.php
```diff
@@ -34,22 +35,32 @@ class PreviewEnhancer {
...
+		$this->userId = $userId;
 	}
 
 	/**
 	 * @param Message[] $messages
 	 *
 	 * @return Message[]
 	 */
-	public function process(Account $account, Mailbox $mailbox, array $messages): array {
+	public function process(Account $account, Mailbox $mailbox, array $messages, string $mode = 'sync'): array {
```
ChristophWurst, line 63:
```suggestion
	public function process(Account $account, Mailbox $mailbox, array $messages, bool $preLoadAvatars = false): array {
```

lib/IMAP/PreviewEnhancer.php
```diff
@@ -34,22 +35,32 @@ class PreviewEnhancer {
...
-		LoggerInterface $logger) {
+		LoggerInterface $logger,
+		AvatarService $avatarService,
+		string $userId) {
 		$this->clientFactory = $clientFactory;
 		$this->imapMapper = $imapMapper;
 		$this->mapper = $dbMapper;
 		$this->logger = $logger;
+		$this->avatarService = $avatarService;
+		$this->userId = $userId;
```
ChristophWurst, line 55: nitpick: I see injection of data as a code smell. The only exception for a valid userId injection is a controller.

Would it be possible to move the transport of the user context (userId) to the method parameters?

ChristophWurst:
How will this handle worst case scenarios? Imagine you load 20 messages and each of them is sent by someone else, and there is no contacts picture so gravatar favicon sources are tapped. These make external HTTP requests that could take several seconds, e.g. when the favicon request times out because the domain does not correspond to a HTTP server.

Will this slow down the loading of messages?

If so, we have to fine-tune the approach. Let me know, I have some ideas.

hamza221:
So I did a little benchmarking and here are the results 

**Main**
1. /api/messages?mailboxId=1&limit=20 => 501.22 ms  
2. First batch of requests  /api/avatars/url/{email}=> 9583 ms 
total = 10084 ms 
3. external avatars (3 email addresses ) api/avatars/image/{email}  => 5158 ms

**This branch**

* /api/messages?mailboxId=1&limit=20 (combines 1 and 2 from above)   => 1.4 minutes / 2.4 minutes / 2.3 minutes 
* fetching additional messages => 16.94 seconds 

Doesn't look like a good approach 
Ps: tests done with caching disabling and no additional throttling / timeouts

ChristophWurst:
Okay, that is what I was afraid of.

Then we do it like follows: check the **cache** for an avatar URL. If it exists, attach it to the recipient. Also attach a **false** if there is one, it indicates that there is no avatar for the user.
Leave the rest of the fetching in tact.

This means that the initial loading with empty cache is still done in many requests, but those help us offload the slow fetching into async operations that don't block the main app functionality. For page reloads, mailbox switches, messages of senders seen previously etc we will preload the avatar URLs and save some requests.

hamza221:
I'm not sure how to return state ( Avatar => avatar exists , null => avatar doesn't exist, ? => we want to fetch it from the frontend) 

https://github.com/nextcloud/mail/pull/10486/files#diff-a7e5eb4728a4ad105af312d21ceb1bd6dd7a70068ba67d4b1881099dcae1076bR99

ChristophWurst:
Avatar (value) vs `null` vs `false`? That's how we store the ternary state in the backend

hamza221:
> Avatar (value) vs `null` vs `false`? That's how we store the ternary state in the backend

oh, it wouldn't throw an error when the return type is ?Avatar ?

lib/Db/Message.php
```diff
@@ -286,6 +293,27 @@ public function setFlag(string $flag, bool $value = true) {
...
+	/**
+	 * @param Avatar|null $avatar
+	 * @return void
+	 */
+	public function setAvatar(?Avatar $avatar): void {
+		$this->avatar = $avatar;
+	}
+
+	/**
+	 * @param bool $fetchAvatarFromClient
```
ChristophWurst, line 305: nit: phpdoc duplicates the native php typehint. it can go

lib/Db/Message.php
```diff
@@ -286,6 +293,27 @@ public function setFlag(string $flag, bool $value = true) {
...
+	 */
+	public function setFetchAvatarFromClient(bool $fetchAvatarFromClient): void {
+		$this->fetchAvatarFromClient = $fetchAvatarFromClient;
+	}
+
+	/**
+	 * @return Avatar
```
ChristophWurst, line 312: here the typehint contradicts the native one. Is the return value nullable or not? :wink:

lib/Db/Message.php
```diff
@@ -331,6 +359,8 @@ static function (Tag $tag) {
 			'summary' => $this->getSummary(),
 			'encrypted' => ($this->isEncrypted() === true),
 			'mentionsMe' => $this->getMentionsMe(),
+			'avatar' => $this->avatar ? $this->avatar->jsonSerialize() : null,
```
ChristophWurst, line 362:
```suggestion
			'avatar' => $this->avatar?->jsonSerialize(),
```

https://php.watch/versions/8.0/null-safe-operator

lib/Service/AvatarService.php
```diff
@@ -101,6 +102,11 @@ public function getAvatar(string $email, string $uid): ?Avatar {
 			return null;
 		}
 
+		if($cachedOnly){
```
ChristophWurst, line 105:
```suggestion
		if ($cachedOnly){
```

ChristophWurst (commented):
Code looks good!

Do you have any comparisons for nr of request after page reload from main vs this branch?

ChristophWurst:
Mind using merges over rebases until the PR is reviewed? I can't see what you have pushed since my last review when history is rewritten.

image

hamza221:
> Mind using merges over rebases until the PR is reviewed? I can't see what you have pushed since my last review when history is rewritten.
> 
> image

I usually do, I had to rebase to resolve conflicts, and squashing made it way easier

lib/IMAP/PreviewEnhancer.php
```diff
@@ -59,6 +66,22 @@ public function process(Account $account, Mailbox $mailbox, array $messages): ar
 			return array_merge($carry, [$message->getUid()]);
 		}, []);
 
+		if ($preLoadAvatars) {
+			foreach ($messages as $message) {
+				$from = $message->getFrom()->first();
+				if ($message->getAvatar() === null && $from !== null && $from->getEmail() !== null && $userId !== null) {
```
ChristophWurst, line 72: nit: negate the condition and use `continue` to skip one loop iteration. That avoids at least one level of indentation in this deep indentation tree.

ChristophWurst (approved):
Big :+1: 

Tested and works

Here are my numbers:
* main: takes 48 requests to load the unified inbox with three accounts
* this branch: takes 29 requests to load the unified inbox with three accounts

ChristophWurst:
@GretaD @st3iny good for you too? I guess this can be seen as part of https://github.com/nextcloud/mail/issues/10668

GretaD (approved):

lib/Service/AvatarService.php
```diff
@@ -89,9 +89,10 @@ private function hasAllowedMime(Avatar $avatar): bool {
 	/**
 	 * @param string $email
 	 * @param string $uid
-	 * @return Avatar|null
+	 * @param bool $cachedOnly
+	 * @return Avatar|null|false
 	 */
-	public function getAvatar(string $email, string $uid): ?Avatar {
+	public function getAvatar(string $email, string $uid, bool $cachedOnly = false): mixed {
```
kesselb, line 95: Could we add a new method getCacheAvatar instead of adding a new argument here and changing the return type to mixed?

src/components/OutboxMessageListItem.vue
```diff
@@ -10,7 +10,10 @@
 		:details="details"
 		@click="openModal">
 		<template #icon>
-			<Avatar :display-name="avatarDisplayName" :email="avatarEmail" />
+			<Avatar :display-name="avatarDisplayName"
+				:email="avatarEmail"
+				::fetch-avatar="data.fetchAvatarFromClient"
```
st3iny, line 15:
```suggestion
				:fetch-avatar="data.fetchAvatarFromClient"
```

lib/Service/AvatarService.php
st3iny: I kind of agree with Daniel. I don't like methods returning either null or false (or an object) because it is hard to understand the semantics without knowing the method's code.

st3iny (changes requested):
Looks good otherwise. The amount of requests went down *significantly* on my personal mail account :rocket:

hamza221:
added Unit tests -> codecov is now happy I'll squash, rebase and merge

PR #10516, opened 2024-12-17
Remove php-mail transport

### LINKED ISSUE #10066: Mail attachment does not work / header problems
### Steps to reproduce

1. Open mail app (web interface)
2. Go to "Neue Nachricht" (new message)
3. Select a recipient, write subject and message body
4. Go to "..." at the bottom of the mail interface, then select "Anhang hochladen" (upload attachment)
5. Select an attachment from the File Selection Box
6. Press "SEND"
7. Look at the mail that was sent

I am using Nextcloud Hub 8 29.0.5 with Mail 3.7.7, but this also happened with older nextcloud versions 28.* and 29.0.x with corresponding older mail clients. Language is German.

### Expected behavior

After step 7, I should see a mail with the attachment in another mail client (or in the mail app, if I sent a mail to myself)

### Actual behavior

I see a mail, but the attachment is not there. I see a text "This message is in MIME format"

### Mail app version

3.7.7

### Mailserver or service

exim as mailserver; happens with SMPT transpart, as well as with sendmail interface

### Operating system

Debian Linux

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

Looking at the mail in maildiir format, I can see that the headers seem not to be formatted correctly:

``From: XXXredactedXXX <XXXredactedXXX@XXXredactedXXX>``
``  Cc: ``
`` Message-ID: <20240828123116.Horde.GesoMkoURhU5Gojp60h8exU@XXXredactedXXX>``
`` User-Agent: Horde Application Framework 5``
`` Date: Wed, 28 Aug 2024 12:31:16 +0000``
`` Content-Type: multipart/mixed; boundary="=_gPXHtDyO8TUAmLNl8o7IgH_"``
`` MIME-Version: 1.0``
``Message-Id: <E1sjHpU-00DBWN-1B@XXXredactedXXX>``

Please note the space as first char in the lines Cc:, Message-ID: (the first one), User-Agent:, Date:, Content-Type: and MIME-Version:. Somehow, it seems the lines are not properly ended (perhaps, some CR/LF vs LF vs CR issue?)

That's also why the MTA creates a new Message-ID for this mail. This explains why the MIME data is not processed, because the headers are missing.

kesselb:
Fix #10066 

image

I will backport the setup check with a different wording for stable4.1, stable4.0 and stable3.7. 
For stable3.7 we need a conditional to register the setup check, or we skip the backport.


SebastianKrupinski (approved):

ChristophWurst:
I think this would qualify as a breaking change. We should do a major version bump for this. Do we want to have the setup check only as a first step and the actual removal a bit later like half a year?

kesselb:
> I think this would qualify as a breaking change.

Yes. I will **only** backport the setup check. The removal is done for main / upcoming 4.2. 

I don’t mind keeping the code for a bit longer. However, I’d be surprised if we received many complaints about its removal. This kind of configuration is problematic: a user wouldn’t expect that, even with SMTP credentials set, the message is still delivered by the webserver.

kesselb:
We agreed to keep the php-mail transport for Mail 4.2. 

Here's the extracted setup check: https://github.com/nextcloud/mail/pull/10574

kesselb:
With Mail 4.2 released and Mail 4.3 branched off, we should be good to get this on the way.

ChristophWurst (approved):
:saluting_face:

ChristophWurst (changes requested):
make it a `fix!` please

PR #10523, opened 2024-12-20
fix(notifications): Notifier::prepare() threw \InvalidArgumentExcepti…

nickvergessen:
…on which is deprecated

If a user has any notifications this log is generated all the time:

> OCA\Mail\Notification\Notifier::prepare() threw \InvalidArgumentException which is deprecated. Throw \OCP\Notification\UnknownNotificationException when the notification is not known to your notifier and otherwise handle all \InvalidArgumentException yourself


kesselb:
Nice, Thank you :pray:

kesselb (approved):

PR #10540, opened 2025-01-01
[stable4.0] fix(dns): Update public suffix list

nextcloud-command:
Auto-generated update of https://publicsuffix.org/


ChristophWurst:
4.0 is no longer maintained

PR #10580, opened 2025-01-18
Fix reply button spacing #10525

### LINKED ISSUE #10525: Reply button spacing bottom is off
### Steps to reproduce

1. Click on a message
2. Scroll down to see reply button

### Expected behavior

Button has the same spacing right and bottom.

### Actual behavior

Button has more spacing bottom than right.

### Mail app version

4.1.0

### Nextcloud version

30.0.4

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

PHP 8.2

### Nextcloud memory caching

_No response_

### Web server

Nginx

### Database

MariaDB

### Additional info

.mail-message-body has margin-bottom 30px instead of 10px.

keremsemiz:
fixes #10525 
Fix reply button spacing

Screenshot from 2025-01-22 17-54-07

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -177,7 +177,7 @@ export default {
 	}
 }
 .reply-buttons {
-	margin: 0 10px 0 50px;
+	margin: 0 30px 0 50px;
 	display: flex;
 	flex-wrap: wrap;
 	gap: 5px;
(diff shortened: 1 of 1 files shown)


src/components/Message.vue
```diff
@@ -200,6 +202,12 @@ export default {
 
 	&__notsuggested {
 		margin-left: auto;
+		padding-top: 3px;
+		padding-right: 10px;
+		padding-bottom: 3px;
+		display: flex;
+		align-items: center;
+		justify-content: center;
```
GretaD, line 210: In general, trying to style only the nonsuggested(a.k.a reply button) is the right way to do it, but in this case the issue is caused by a margin that already exists on the parent class `.reply-buttons`. So try to change the margin-right, to be same as bottom margin :)

GretaD (approved):

Issue #10561, opened 2025-01-13
Envelope list has text shifting and hover jitters


GVodyanov:
### Steps to reproduce

1. Open envelope list
2. Hover on envelopes

### Expected behavior

Text doesn't shift on hover 

Extra margin doesn't appear under envelopes on hover

### Actual behavior

https://github.com/user-attachments/assets/b4bfa0fe-c6d4-441f-966d-b68e2ca90791

Found by @nimishavijay 

### Mail app version

main

### PHP engine version

None

### Web server

None

### Database

None

PR #10583, opened 2025-01-19
EnvelopeSkeleton: add filler on hover to prevent text shifting

GVodyanov:
Fix 1/2 https://github.com/nextcloud/mail/issues/10561

When hovering on envelopes in the envelope list, the preview text shouldn't expand when date is hidden

### DIFF
--- a/src/components/EnvelopeSkeleton.vue
+++ b/src/components/EnvelopeSkeleton.vue
@@ -83,6 +83,10 @@
 				<div v-show="forceDisplayActions || displayActionsOnHoverFocus"
 					class="list-item__actions"
 					@focusout="handleBlur">
+					<!-- These details aren't actually displayed, they are just there to prevent text jittering on hover -->
+					<div class="list-item__actions__filler">
+						<slot name="list-item__actions__filler">{{ details }}</slot>
+					</div>
 					<NcActions ref="actions"
 						:primary="isActive || active"
 						:aria-label="computedActionsAriaLabel"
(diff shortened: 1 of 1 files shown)


src/components/EnvelopeSkeleton.vue
```diff
@@ -633,6 +637,15 @@ export default {
 		justify-content: center;
 		align-self: start;
 		margin-top: 0;
+		display: flex;
+		flex-direction: column-reverse;
+		align-items: flex-end;
```
GretaD, line 642: these will not be applied because the parent is display none. It still jumps for me

ChristophWurst: @GretaD ?

GretaD:
i tested it aging, for in me all screen sizes it jumps like this.

https://github.com/user-attachments/assets/b9d6306f-ef64-461c-84ac-a1cf85003ca2

src/components/EnvelopeSkeleton.vue
ChristophWurst: Continued at https://github.com/nextcloud/mail/pull/10583#issuecomment-2838605655

GVodyanov:
Found the issue! Thanks a lot

src/components/EnvelopeSkeleton.vue
```diff
@@ -595,6 +599,10 @@ export default {
 		.list-item__actions {
 			align-self: center;
 			margin-top: 0;
+
+			&__filler {
+				margin-bottom: -10.25px;
```
ChristophWurst, line 604: nitpick: Is this a stable solution in a sense that it won't break with any small pixel shifts of future changes of this component?

GVodyanov: Hey @ChristophWurst! Yeah, I guess that it might change 😅, I changed it now to get the number through a dynamic calculation, the formula is a little weird because of the complexity of the formatting here (which is why I didn't think of using it initially, but I left a comment explaining it, so people don't get too confused in the future), but it does make sense.

ChristophWurst: Whohoo

ChristophWurst (approved):
Looks sane. @GretaD please test if the flickering is fixed for you as well

PR #10571, opened 2025-01-15
fix: message content escaping message element

GretaD:
fixes #10570

PR #10587, opened 2025-01-20
fix: no selected message empty content misalignment

### LINKED ISSUE #10586: empty content when no message select is misaligned
### Steps to reproduce

1. Go to mail app and dont select any messages
2. See the empty content

### Expected behavior

centered text

### Actual behavior

Image

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

regression https://github.com/nextcloud/mail/pull/10571

GretaD:
fixes #10586

### DIFF
--- a/src/components/NoMessageSelected.vue
+++ b/src/components/NoMessageSelected.vue
@@ -52,7 +52,7 @@ export default {
 }
 .app-content {
 	display: flex;
-	height: 100%;
+	height: 100% !important;
 	justify-content: center;
 	&__empty {
 		flex-grow: unset;


src/components/NoMessageSelected.vue
```diff
@@ -52,7 +52,7 @@ export default {
 }
 .app-content {
 	display: flex;
-	height: 100%;
+	height: 100% !important;
```
ChristophWurst, line 55: is there a conflicting rule that set a different height?

GretaD: yes, the app-content-details, that breaks the message content.

https://github.com/nextcloud/mail/pull/10571/files#diff-2dc54a20fa4ab6845b4b0f252336e107d33a13f5c05d966e5c104770a1b292fdR204

ChristophWurst: Wasn't that just added with https://github.com/nextcloud/mail/commit/8089ccb0f3915d57520458ecf0ce8729341cfbd4?

This doesn't smell like a maintainable solution if the v-deep is leaking to places it shouldn't and we have to pull the `!important` overwrite to push back.

Would it be possible to get rid of the deep selector?

ChristophWurst: Is Contacts missing `scoped` for the component styles?

GretaD: it comes from a css file: https://github.com/nextcloud/contacts/blob/main/src/css/contacts.scss#L25
I dont see any vue component on contacts not to be scoped and has this class

ChristophWurst: Can we move the css file into the app component that uses the lib component that is being styled?

GretaD:
close because its fixed on contacts https://github.com/nextcloud/contacts/pull/4302

PR #10609, opened 2025-01-27
fix: issues with AI integration

hamza221:
`convertAxiosError`converts an existing error, and we're not passing any error we're throwing a custom one, mb 🙈

### DIFF
--- a/src/service/AiIntergrationsService.js
+++ b/src/service/AiIntergrationsService.js
@@ -13,7 +13,9 @@ export const summarizeThread = async (threadId) => {
 
 	try {
 		const resp = await axios.get(url)
-		if (resp.status === 204) throw convertAxiosError()
+		if (resp.status === 204) {
+			throw new Error('Thread summary failed, error in the llm service')
+		}
 		return resp.data.data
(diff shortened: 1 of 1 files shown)


hamza221:
I just realized that the features were introduced in different versions , splitting PR

SebastianKrupinski (approved):

ChristophWurst (approved):

src/service/AiIntergrationsService.js
```diff
@@ -13,7 +13,7 @@ export const summarizeThread = async (threadId) => {
 
 	try {
 		const resp = await axios.get(url)
-		if (resp.status === 204) throw convertAxiosError()
+		if (resp.status === 204) throw new Error('Thread summary failed, error in the llm service')
```
ChristophWurst, line 16: **always** use blocks for conditions. We don't use single-line ifs at Nextcloud

ChristophWurst: https://docs.nextcloud.com/server/latest/developer_manual/getting_started/coding_standards/php.html#control-structures

Issue #10534, opened 2024-12-27
Aphistrophe not working in welcome screen


kesselb:
### Steps to reproduce

1. Set instance name to `Nextcloud'`
2. Open mail
3. See `Nextcloud&#39;`

### Expected behavior

`Nextcloud'`

### Actual behavior

![Image](https://github.com/user-attachments/assets/bcff3ad1-a0ec-4bc1-92e2-2c493c1a430b)

### Mail app version

main

### Nextcloud version

main

GretaD:
same for "

![Image](https://github.com/user-attachments/assets/5ef3f7ff-4dcf-48bd-959b-349f4c3cbb55)

PR #10614, opened 2025-01-28
fix: apostrophe issue

SebastianKrupinski:
Fix https://github.com/nextcloud/mail/issues/10534

Fixes and issue with apostrophes in system name

image

### DIFF
--- a/src/components/NoMessageSelected.vue
+++ b/src/components/NoMessageSelected.vue
@@ -5,14 +5,11 @@
 
 <template>
 	<AppContentDetails class="app-content">
-		<NcEmptyContent class="app-content__empty">
+		<NcEmptyContent class="app-content__empty"
+			:name="t('mail', 'Welcome to {productName} Mail', { productName }, null, {escape: false})"
+			:description="t('mail', 'Start writing a message by clicking below or select an existing message to display its contents')">
 			<template #icon>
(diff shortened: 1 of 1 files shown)


src/components/NoMessageSelected.vue
```diff
@@ -37,7 +37,7 @@ export default {
 
 	computed: {
 		welcomeMessage() {
-			return t('mail', 'Welcome to {cloudName} Mail', { cloudName: window?.OC?.theme?.name ?? 'Nextcloud' })
+			return t('mail', 'Welcome to {cloudName} Mail', { cloudName: window?.OC?.theme?.name ?? 'Nextcloud' }, null, {escape: false})
```
ChristophWurst, line 40: https://github.com/nextcloud/mail/blob/ae4502ac5c082b0e39722641ac07b096e9aff524/src/components/NoMessageSelected.vue#L13-L15 puts `welcomeMessage` into the DOM as-is. Doesn't that open the door for code injection?
Is `window?.OC?.theme?.name` HTML encoded already?!

SebastianKrupinski: window?.OC?.theme?.name is not HTML encoded, the encoding happens in the translation function.

Made changes, what do you think?

ChristophWurst (approved):

Issue #10521, opened 2024-12-19
Setup check for IMAP connection speed of provisioned accounts


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a nextcloud Mail admin I want to get warned when provisioned accounts may be slow.

### Describe the solution you'd like

Add a *setup check* for this app that takes three random accounts per provisioning config, performs a simple reading operation on IMAP and measures the time of that. If the time is above x (to be determined), show a warning.

### Describe alternatives you've considered

N/a

### Additional context

We want to make it clearer when performance problems in Nextcloud are not caused by Nextcloud itself but infrastructure or external endpoints.

PR #10659, opened 2025-02-05
feat: check connection performance of mail service

SebastianKrupinski:
Resolves: #10521

### DIFF
--- a/lib/Db/ProvisioningMapper.php
+++ b/lib/Db/ProvisioningMapper.php
@@ -138,4 +138,19 @@ class ProvisioningMapper extends QBMapper {
 			return null;
 		}
 	}
+
+	/**
+	 * @since 4.2.0
+	 *
+	 * @return array<int,string>
+	 */
+	public function findUniqueImapHosts(): array {
+		$query = $this->db->getQueryBuilder();
+		$query->selectDistinct('imap_host')
+			->from('mail_provisionings');
+		$result = $query->executeQuery();
+		$data = $result->fetchAll(\PDO::FETCH_COLUMN);
+		$result->closeCursor();
+		return $data;
+	}
 }
--- a/lib/SetupChecks/MailConnectionPerformance.php
+++ b/lib/SetupChecks/MailConnectionPerformance.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\SetupChecks;
+
+use OCA\Mail\Account;
+use OCA\Mail\Db\MailAccountMapper;
+use OCA\Mail\Db\ProvisioningMapper;
+use OCA\Mail\IMAP\FolderMapper;
+use OCA\Mail\IMAP\IMAPClientFactory;
+use OCP\IL10N;
+use OCP\SetupCheck\ISetupCheck;
+use OCP\SetupCheck\SetupResult;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+class MailConnectionPerformance implements ISetupCheck {
+	public function __construct(
+		private IL10N $l10n,
+		private LoggerInterface $logger,
+		private ProvisioningMapper $provisioningMapper,
+		private MailAccountMapper $accountMapper,
+		private IMAPClientFactory $clientFactory,
+		private FolderMapper $folderMapper,
+		private MicroTime $microtime,
+	) {
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('Mail connection performance');
+	}
+
+	public function getCategory(): string {
+		return 'mail';
+	}
+
+	public function run(): SetupResult {
+		// retrieve unique imap hosts for provisionings and abort if none exists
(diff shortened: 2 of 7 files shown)


lib/Db/ProvisioningMapper.php
```diff
@@ -138,4 +138,17 @@ public function get(int $id): ?Provisioning {
...
+	 */
+	public function uniqueImapHosts(): array {
+		$query = $this->db->getQueryBuilder();
+		$query->selectDistinct('imap_host')
+			->from('mail_provisionings');
+		$result = $query->executeQuery();
+		$data = $result->fetchAll(\PDO::FETCH_COLUMN);
+		$result->closeCursor();
+		return $data;
+	}
```
ChristophWurst, line 153: ^ as Codecov says, there is no test coverage. Please extend `\OCA\Mail\Tests\Integration\Db\ProvisioningMapperTest`

lib/SetupChecks/MailConnectionPerformance.php
```diff
@@ -0,0 +1,111 @@
...
+					$this->l10n->t('Slow mail service detected (%1$s) an attempt to perform a mail box list operation on several accounts took an average of %2$s seconds per account', [$host, round($operation, 3)])
+				);
+			}
+		}
+		return SetupResult::success();
+	}
+
+	public function microtime(bool $asFloat = false): mixed {
+		return microtime($asFloat);
+	}
```
ChristophWurst, line 109: I'm assuming this was added to mock time in tests. I'd advise to refactor to either of the two common options:
1. Use `\OCP\AppFramework\Utility\ITimeFactory::getTime`. You will get less precision but for a comparison and check if the connection takes more than a second, you're probably fine.
2. Move this method into a new class.

The reason I don't like the current approach is that you are mocking an internal of a class while testing. That's an anti pattern to avoid. Only external classes should be mocked. Additionally, a public method is exposed, which serves no purpose otherwise.

tests/Unit/SetupChecks/MailConnectionPerformanceTest.php
```diff
@@ -0,0 +1,191 @@
...
+	private MailAccountMapper&MockObject $accountMapper;
+	private IMAPClientFactory&MockObject $clientFactory;
+	private FolderMapper&MockObject $folderMapper;
+	private MailConnectionPerformance&MockObject $check;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->l10n = $this->createMock(L10N::class);
+		$this->logger = new NullLogger();
```
ChristophWurst, line 40: You can also just use a mock. This comes with the bonus that you would also be able to assert logging when it is important for the test case.

tests/Unit/SetupChecks/MailConnectionPerformanceTest.php
```diff
@@ -0,0 +1,191 @@
...
+			->willReturn($client);
+	
+		$this->check->method('microtime')
+			->willReturnOnConsecutiveCalls(0, .2, .2);
+
+		$result = $this->check->run();
+
+		$this->assertEquals(SetupResult::SUCCESS, $result->getSeverity());
+		
+	}
```
ChristophWurst, line 136: Tip for readable unit test: structure them in three blocks:
1. arrange - create any mock objects needed, initialize stubs
2. act - call the method(s) you are testing
3. assert - test the result of the method for the expected result. this last block may be missing for void methods or when the return value is not important for the test.

Use a blank line between. You will find this pattern in other tests too.

This is a standard practice. See https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/ for an introduction with more details.

This pattern will make your test a lot easier to read. This will help the reviewers, and also the next engineer to touch this code.

tests/Unit/SetupChecks/MailConnectionPerformanceTest.php
kesselb: There's also a TestLogger: https://github.com/nextcloud/mail/blob/4b1b6ca8c5873e8630bbbaac48bfd9cd9726ca58/tests/Unit/Listener/MoveJunkListenerTest.php#L32

(it provides some useful methods to get the number of log records)

tests/Unit/SetupChecks/MailConnectionPerformanceTest.php
SebastianKrupinski: Done.

tests/Unit/SetupChecks/MailConnectionPerformanceTest.php
SebastianKrupinski: Done. Used the TestLogger, added another test for it also

lib/SetupChecks/MailConnectionPerformance.php
```diff
@@ -0,0 +1,108 @@
...
+		private ProvisioningMapper $provisioningMapper,
+		private MailAccountMapper $accountMapper,
+		private IMAPClientFactory $clientFactory,
+		private FolderMapper $folderMapper,
+		private MicroTime $microtime,
+	) {
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('mail connection performance');
```
ChristophWurst, line 36:
```suggestion
		return $this->l10n->t('Mail connection performance');
```

sorry, I wasn't precise. The start of the sting stays capitalized. Amend after your rebase+squash ;-)

ChristophWurst (approved):
:clap:

PR #10661, opened 2025-02-06
fix: detect imip messages from outlook.com

### LINKED ISSUE #10660: Email from Outlook not detected as iMIP message
### Steps to reproduce

1. Send a calendar invitation from outlook.com.
2. Accept the invitation in Nextcloud Mail.
3. Cancel the event in outlook.com and choose to notify the attendees.
4. Check the email in Nextcloud Mail and observe that the event is still present in your calendar.

### Expected behavior

With the background job added via https://github.com/nextcloud/mail/pull/6812, such messages should be processed, and the event should be canceled/removed from the calendar.

### Actual behavior

The email from outlook.com is not flagged as an iMIP message, and therefore no background processing occurs.

### Additional info

The email is not flagged as an iMIP message because the message parts are being parsed incorrectly.

The implementation loops over the message parts and then tries to fetch another part. This works for Gmail because the iMIP message is wrapped in multipart/alternative, but fails for Outlook.com, where the text/calendar part is not nested. https://github.com/nextcloud/mail/blob/ea7392942af81b66eac7e743b2c18c04393d17c7/lib/IMAP/MessageMapper.php#L892-L905

Gmail:

```
body  
└─ multipart/alternative  
   ├─ text/plain  
   ├─ text/html  
   └─ text/calendar  
└─ application/ics  
```

Outlook.com:

```
body  
├─ text/plain  
├─ text/html  
└─ text/calendar  
```

kesselb:
Fix #10660 

Adjust the parsing logic in MessageMapper to correctly detect iMIP messages in flat email structures where the text/calendar part is not wrapped in multipart/alternative.

The implementation loops over the message parts and then tries to fetch another part. This works for Gmail because the iMIP message is wrapped in multipart/alternative, but fails for Outlook.com, where the text/calendar part is not nested. 

partIterator returns all parts, including nested ones. 

Gmail:

```
body  
└─ multipart/alternative  
   ├─ text/plain  
   ├─ text/html  
   └─ text/calendar  
└─ application/ics  
```

Outlook.com:

```
body  
├─ text/plain  
├─ text/html  
└─ text/calendar  
```


ChristophWurst (approved):
I'm assuming writing a unit test for this is not trivial, is it?

kesselb:
> I'm assuming writing a unit test for this is not trivial, is it?

We should have tests for it :pray:

Richard did some tests for smime, I will have a look if we can start from there.

ChristophWurst (approved):
😻

SebastianKrupinski:
RFC examples of possible iMip Messages

https://datatracker.ietf.org/doc/html/rfc2447#section-4

Issue #9853, opened 2024-07-15
Printing edge case emails still does not always work


GVodyanov:
### Steps to reproduce

1. Open a long email with images, EG from supabase, facebook, instagram.
2. Click on print email or do `ctrl + p` in Chromium

### Expected behavior

The email content should start at the start of the page

### Actual behavior

![image](https://github.com/user-attachments/assets/943f2152-dc1b-47e3-af3c-76c009c36e1d)

The print preview shows something like this

[Email error Chromium.pdf](https://github.com/user-attachments/files/16233974/Email.error.pdf)

### Steps to reproduce

1. Open any email while having a large screen size
2. Click on print email or do `ctrl + p` in Firefox

### Expected behavior

The email should take up the whole width of the page

### Actual behavior 

![image](https://github.com/user-attachments/assets/4c01ce9d-475d-430d-bbd8-6b3cd44c997d)

[Email error Firefox.pdf](https://github.com/user-attachments/files/16234117/Email.error.Firefox.pdf)

> Note: zooming in a lot or zooming out a lot fixes this

### Mail app version

Main

GVodyanov:
It might be useful to use this after the Vue 3 migration https://vue-to-print.netlify.app/

SebastianKrupinski:
> It might be useful to use this after the Vue 3 migration [vue-to-print.netlify.app](https://vue-to-print.netlify.app/)

Soooo... why haven't you completed the V3 migration yet? 😆😆😆😆😆

GVodyanov:
Here are some notes on the topic in case anyone tries to dig into this again in the future:
- Just printing the page with some custom CSS has issues. The main one has to do with the fact that the `iframes` that represent messages get cut off between pages, seeing as the browser isn't able to distinguish where one element starts or ends inside an `iframe` for some reason. This results in cut images and text over different PDF pages.
    - This is not considering the fact that the custom CSS has formatting issues. It requires hiding a lot of elements and that is often buggy, classes change, and different browsers do printing very differently.

- Using the `files_emailviewer` app to generate the PDF is an idea, but I assume that most instances don't have it configured.

[...]

PR #10670, opened 2025-02-09
Fix text cutting off when printing and differences between clicking print and doing ctrl + P

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/9853

Previously for printing all we did was hide unnecessary DOM elements with CSS

What this PR actually does:
- distinguish between printing singular messages (through the actions menu) and printing the whole thread with `Ctrl + P`
- If printing a specific message only the iframe is printed, with thread info injected up top
- If printing a whole thread with `Ctrl + P` some DOM elements are hidden with CSS, and some are removed forcefully with javascript. This is needed because browser printing APIs aren't the best, and even hidden elements mess up the layout with Chrome. The downside of this is that the browser cannot read elements inside of iframes, so text and images are split between multiple pages. After this is done the page has to be reloaded to restore the deleted elements.
- Fix bug related to the print content being narrow in Firefox, this was caused by the actions menu still existing in the DOM even if hidden for a few seconds after being closed
- Update iframe-resizer to version 5: now it split between two libraries, child and parent
- Handle printing of plain text messages by creating an iframe and inserting them into it


GVodyanov:
Note: narrow messages in firefox seem to reproducible only after having clicked the menu buttons in the thread element, even though the menu was subsequently closed..?

ChristophWurst:
> After this is done the page has to be reloaded to restore the deleted elements.

This feels a bit bumpy, but I also lack alternative ideas off the top of my head …

GVodyanov:
> > After this is done the page has to be reloaded to restore the deleted elements.
> 
> This feels a bit bumpy, but I also lack alternative ideas off the top of my head …

I tried reinserting the deleted elements back afterwards, but I can't figure out how to insert them back into the right place, they kept getting appended to the end of the component.

src/components/Thread.vue
```diff
@@ -324,57 +328,123 @@ export default {
...
+					await new Promise(resolve => setTimeout(resolve, 100))
 				}
 
-				document.querySelectorAll('iframe').forEach((iframe, index) => {
+				const virtualIframe = document.createElement('iframe')
+				virtualIframe.style.display = 'none'
+				document.body.appendChild(virtualIframe)
+				const virtualIframeDocument = virtualIframe.contentDocument || virtualIframe.contentWindow.document
+				virtualIframeDocument.open()
+				virtualIframeDocument.write('<html><head><title>Print</title></head><body></body></html>')
```
ChristophWurst, line 351: L10n por favor

PR #10674, opened 2025-02-10
fix: user must scroll to view most recent message in thread

### LINKED ISSUE #7552: User must scroll to view most recent message in thread
### Steps to reproduce

1. Exchange multiple emails with someone using the same subject to generate a thread with many emails
2. Open the Mail app and select the long message thread
3. The most recent message(s) in the thread are displayed at, or past, the bottom of the screen.
4. The user must scroll to look for the most recent message in the thread

### Expected behavior

When a user clicks on an unread message in the inbox, it should be displayed without requiring the user to search for it.

Note that this would never be an issue if the user were able to disable threading of messages.

### Actual behavior

The oldest messages in the thread are given top placement, while the most recent, and presumably relevant messages require scrolling down to the bottom of the list.

### Mail app version

2.1.0

### Mailserver or service

Personal domain

### Operating system

Ubuntu 22.04.1 LTS x86_64

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

_No response_

GretaD:
fixes #7552

### DIFF
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -601,7 +601,7 @@ export default {
 
 			// Only one envelope is expanded at the time of mounting so we can
 			// assume that this is the relevant envelope to be scrolled to.
-			this.$nextTick(() => this.scrollToCurrentEnvelope())
+			this.$nextTick(() => this.handleThreadScrolling())
 		}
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Tested & works

ChristophWurst:
Please trigger the backports

PR #10694, opened 2025-02-14
fix: change status code from 404 to 204 for missing avatars

kesselb:
Similar to https://github.com/nextcloud/contacts/issues/3021 and https://github.com/nextcloud/server/pull/49839.

Switching the response code from 404 to 204 improves compatibility with security appliances like CrowdSec.
Since avatar lookups (including external sources) can validly result in a non-existing avatar.

Bonus: This will not trigger the debug log for missing avatars anylonger. 

https://github.com/nextcloud/mail/blob/bcd523917605779f84e4944d1455bee505e28ef1/src/components/Avatar.vue#L62

There's another action in the controller also using 404. 
I'm not sure where we are using it and therefore left it unchanged: https://github.com/nextcloud/mail/blob/bcd523917605779f84e4944d1455bee505e28ef1/lib/Controller/AvatarsController.php#L106

### DIFF
--- a/lib/Controller/AvatarsController.php
+++ b/lib/Controller/AvatarsController.php
@@ -54,7 +54,7 @@ class AvatarsController extends Controller {
 		$avatar = $this->avatarService->getAvatar($email, $this->uid);
 		if (is_null($avatar)) {
 			// No avatar found
-			$response = new JSONResponse([], Http::STATUS_NOT_FOUND);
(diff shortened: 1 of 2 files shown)


SebastianKrupinski (approved):

SebastianKrupinski:
> Bonus: This will not trigger the debug log for missing avatars anylonger.

This will be such a nice improvement!!!

ChristophWurst (approved):

PR #10696, opened 2025-02-14
fix(imap): Avoid OOM when syncing sparse mailboxes

ChristophWurst:
UIDs of a mailbox can be fragmented. In that case we overfetch when building chunk pages. Before doing the actual FETCH this does a test on the number of results. If the result is too big, the range will be halved. That is repeated until a reasonable page is built. 
See https://github.com/nextcloud/mail/pull/10696#issuecomment-2735653743 for a production run and data.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -146,6 +146,7 @@ class MessageMapper {
 		} else {
 			$max = ((int)$metaResults['max']);
 		}
+		unset($metaResults);
 
 		// The inclusive range of UIDs
 		$totalRange = $max - $min + 1;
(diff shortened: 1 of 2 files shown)


ChristophWurst:
From testing with a real email account I see the initial sync of Inbox+Sent to need 0.179% more IMAP communication.

ChristophWurst:
From the affected production instance:

image

^ there is a huge range of UIDs and the assumption that they are distributed evenly is wrong. It turns out there is a hotspot at the start of the range. It could also be anywhere else. This does the trick neatly.

SebastianKrupinski (approved):

SebastianKrupinski:
Tested Works

Issue #10688, opened 2025-02-13
Integrations tests against nextcloud master are failing when creating a nextcloud user


kesselb:
### Steps to reproduce

1. Run CI
2. Spot an error when creating a nextcloud user

### Expected behavior

✅ 

### Actual behavior

❌ 

### Additional info

Something in server changed since yesterday, causing our integration tests to fail.

Last successful run: https://github.com/nextcloud/mail/actions/runs/13292360986 (2025-02-12).

```
 1) OCA\Mail\Tests\Integration\Db\InternalAddressMapperTest::testIndividualExists
TypeError: OCA\DAV\Events\CardCreatedEvent::__construct(): Argument #2 ($addressBookData) must be of type array, null given, called in /home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/CardDAV/CardDavBackend.php on line 649

/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/Events/CardCreatedEvent.php:30
/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/CardDAV/CardDavBackend.php:649
/home/runner/actions-runner/_work/mail/mail/nextcloud/lib/public/AppFramework/Db/TTransactional.php:45
/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/CardDAV/CardDavBackend.php:611
/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/CardDAV/SyncService.php:246
/home/runner/actions-runner/_work/mail/mail/nextcloud/lib/public/AppFramework/Db/TTransactional.php:45
/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/CardDAV/SyncService.php:241
/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/Listener/UserEventsListener.php:73
/home/runner/actions-runner/_work/mail/mail/nextcloud/apps/dav/lib/Listener/UserEventsListener.php:51
/home/runner/actions-runner/_work/mail/mail/nextcloud/lib/private/EventDispatcher/ServiceEventListener.php:68
/home/runner/actions-runner/_work/mail/mail/nextcloud/3rdparty/symfony/event-dispatcher/EventDispatcher.php:220
/home/runner/actions-runner/_work/mail/mail/nextcloud/3rdparty/symfony/event-dispatcher/EventDispatcher.php:56
[...]

PR #10706, opened 2025-02-17
fix(deps): Do not ship psr/log

ChristophWurst:
It is provided by server's 3rdparty. Shipping another copy can lead to a dependency hell.

psr/log is now installed only for Psalm to have stubs. This avoids vendoring stubs in our repo, which would get outdated one day. Renovate will help us keep the stub package updated.

Fixes https://github.com/nextcloud/mail/issues/1065

## How to test

* `composer i` to wipe the old psr/log

The app should work the same as on `main`.


.github/workflows/test.yml
```diff
@@ -63,16 +63,16 @@ jobs:
       strategy:
           matrix:
               php-versions: ['8.3']
-              nextcloud-versions: ['master']
+              nextcloud-versions: ['chore/mail-bisect-6e1d9a26209ec5524fbc2fb9c7cbb53315e64d72']
```
ChristophWurst, line 66: This is because of https://github.com/nextcloud/mail/issues/10688#issuecomment-2662550235. I need a slightly outdated server checkout for CI. Right now that means we don't get the latest server master results, but I'm mostly interested in getting this to stable4.2

kesselb (approved):
:sunglasses:

PR #10709, opened 2025-02-17
ci: Test stable4.2 against server stable31, not master

ChristophWurst:

PR #10710, opened 2025-02-17
ci: Do not force-enable the app

ChristophWurst:
This hides incompatibility issues.

Discovered in https://github.com/nextcloud/mail/pull/10709/files#r1958107277.


ChristophWurst:
@kesselb did we just forget to revert the temporary solution https://github.com/nextcloud/mail/pull/10005?

kesselb (approved):

kesselb:
> did we just forget to revert the temporary solution https://github.com/nextcloud/mail/pull/10005?

I guess

PR #10728, opened 2025-02-20
fix: handle 204 response

kesselb:
Follow-up for https://github.com/nextcloud/mail/pull/10694

- A 404 is no longer an expected response, so the special handling has been removed.
- A 204 response has an empty body, which means avatar.url would return undefined. This update makes it more explicit what is happening and what’s expected, though the behavior itself remains unchanged.

### DIFF
--- a/src/service/AvatarService.js
+++ b/src/service/AvatarService.js
@@ -17,22 +17,18 @@ export const fetchAvatarUrl = (email) => {
 	})
 
 	return Axios.get(url, { adapter: 'fetch', fetchOptions: { priority: 'low' } })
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
I was actually surprised for this to work without frontend modifications in the original PR :see_no_evil:

kesselb:
> I was actually surprised for this to work without frontend modifications in the original PR 🙈

It does work as expected even without this change. Coincidentally, the expected return for 404 response was undefined and `"".url` is also undefined :rofl:

PR #10756, opened 2025-02-26
perf: show the loading spinner only when sync takes a while

GretaD:
fixes #[10716](https://github.com/nextcloud/mail/issues/10716)


GretaD:
closed as the initial issue is not reproducible

PR #10763, opened 2025-02-27
fix: html5 tag errors

SebastianKrupinski:
Resolves: Client Issue

### Summery
- Adds DOCTYPE to html wrapper to prevent libxml errors for html 5 tag "DOMDocument::loadHTML(): Tag figure invalid in Entity, line: 1"
- Adds limitation to prevent network access for DOMDocument

### DIFF
--- a/lib/Service/MimeMessage.php
+++ b/lib/Service/MimeMessage.php
@@ -67,13 +67,13 @@ class MimeMessage {
 			
 			// determine if content is wrapped properly in a html tag, otherwise we need to wrap it properly
 			if (mb_strpos($contentHtml, '<html') === false) {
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Looks good

Is only main affected by this or should it be backported to stable releases?

SebastianKrupinski:
> Looks good
> 
> Is only main affected by this or should it be backported to stable releases?

Yes I would back port it to all the versions that support NC30+

SebastianKrupinski:
> Looks good
> 
> Is only main affected by this or should it be backported to stable releases?

Wait that might not be possible, I made changes to that code block recently, I'll check how far back it went

SebastianKrupinski:
Yes, we would need to backport this also...

PR #10779, opened 2025-02-28
chore: better help text when sieve is not enabled

kesselb:
We received the feedback that the autoresponder help text, when sieve not enabled, is a bit technical. It's hopefully easier to understand and includes a button to jump to the sieve settings ;)

image


ChristophWurst (approved):
Nice

:point_up::nerd_face: this commit changes the product for the user. IMO it should be a `fix` commit. `chore`s alone wouldn't even allow me to do an automated release.

kesselb:
Sure, changed the commit type to fix.

Issue #10785, opened 2025-03-04
Unable to fully view summarized preview.


KraiiFox:
### Steps to reproduce

1. Have a low resolution screen / High scaling factor (mine is 1920x1200 at 150% scaling) or just be zoomed in.
2. Observe an email with a summarized preview in the list.
3. The summarized preview is cut off.

### Expected behavior

Not really a bug so no expected behaviour, just a suggested improvement.

### Actual behavior

Unable to fully read the summarized preview. Hovering over the item list just shows a tooltip with "This summary was AI generated." 

![Image](https://github.com/user-attachments/assets/247ecdb0-2ebf-485c-b16e-cdba374d52e9)

![Image](https://github.com/user-attachments/assets/32a9ea25-bda5-44d1-a540-6051fa57f476)

![Image](https://github.com/user-attachments/assets/35b622c8-e3f4-455a-a83e-cc3873bc3059)
(Fully expanded)

### Mail app version

4.2.3

### Nextcloud version

31.0.0

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

As far as I know, there is no way to fully view the summarized preview in the UI normally except to zoom all the way out or via the inspect element.

I feel like not being able to fully read the whole thing defeats the whole purpose of it being a "preview". So my suggestion is to include full text inside of the tooltip with "This summary was AI generated." appended to the bottom. 

![Image](https://github.com/user-attachments/assets/a05be535-99a0-4b66-aa01-86ec1ce3bc11)

ChristophWurst:
Thank you for the report, @KraiiFox!

This is fixed by https://github.com/nextcloud/mail/pull/10833. Old summaries will stay, but new ones should have shorter text.

PR #10833, opened 2025-03-10
fix: shorten ai summaries to fit in message list better

SebastianKrupinski:
Resolves: AI summaries do not fit in to message list column properly


ChristophWurst:
There is an existing ticket for the problem. Could you link it please?

Do you have screenshots showing the before/after?

SebastianKrupinski:
Shortened summary in the first one,

image

ChristophWurst (approved):
That's better :pray:

ChristophWurst:
> There is an existing ticket for the problem. Could you link it please?

https://github.com/nextcloud/mail/issues/10785

Issue #10416, opened 2024-11-27
Calendar invites sent through mail providers have incorrect headers for the ics attachment


ChristophWurst:
### Steps to reproduce

1. Invite another user to an event

### Expected behavior

I receive an email where my mail client offers to import an event and respond to the RSVP

```
Content-Type: text/calendar; method=REQUEST; name=event.ics
Content-Transfer-Encoding: base64
Content-Disposition: inline; name=event.ics; filename=event.ics
```

^ from an invite before we had mail providers.

### Actual behavior

```
Content-Type: text/calendar; name=event.ics; method=REQUEST
Content-Disposition: attachment; filename=event.ics
Content-Transfer-Encoding: 8bit
```

^ at least for Gnome Evolution, this doesn't make the ics file show as invite. It shows as attachment.

### Mail app version

4.1.0-beta2

### PHP engine version

None

### Web server

None

### Database

None

SebastianKrupinski:
### Tested

Confirmed

### Issue
Mail app sending does not have support for inline attachments

### Resolution
Alter OCA\Mail\Service\MailTransmission::sendMessage() line 88 and 124 to handle inline attachments
Alter OCA\Mail\Db\LocalMessageAttachement to handle inline attachments
Alter OCA\Mail\Provider\Command\MessageSend::perform() to handle inline attachments

SebastianKrupinski:
After doing some extensive testing and RFC reading, calendar invitations should NOT be sent as inline attachments.

RFC Examples
https://datatracker.ietf.org/doc/html/rfc2447#section-4

Therefore the current implementation is correct and the previous implementation with inline attachments seems to have been flawed.

ChristophWurst:
> ^ at least for Gnome Evolution, this doesn't make the ics file show as invite. It shows as attachment.

@SebastianKrupinski can you explain why previous invitations were recognized as such by Evolution and the ones sent through the provider mechanism aren't?

SebastianKrupinski:
Please see #10851 for a possible solution.

After testing several other software suites, I noticed that the content disposition on 'text/calendar' event attachments is not set to nothing (missing). Although according to the RFC a missing Content-Disposition should be treated as Content-Disposition: attachment some clients seem to treat a missing disposition differently which in turn causes the event attachments to not be recognized properly.

PR #10851, opened 2025-03-11
fix: allow to send attachment without setting filename

SebastianKrupinski:
Resolves: #10416
Requires: https://github.com/nextcloud/server/pull/51441

This changes the logic during message construction when sending to allow attachments without a file name

### DIFF
--- a/lib/Provider/Command/MessageSend.php
+++ b/lib/Provider/Command/MessageSend.php
@@ -60,8 +60,8 @@ class MessageSend {
 		// validate that all attachments have a name, type, and contents
 		$entries = $message->getAttachments();
 		array_walk($entries, function ($entry) {
-			if (empty($entry->getName()) || empty($entry->getType()) || empty($entry->getContents())) {
-				throw new SendException('Invalid Attachment Parameter: MUST contain values for Name, Type and Contents');
+			if (empty($entry->getType()) || empty($entry->getContents())) {
+				throw new SendException('Invalid Attachment Parameter: MUST contain values for Type and Contents');
 			}
 		});
 		// retrieve user mail account details
--- a/lib/Service/TransmissionService.php
+++ b/lib/Service/TransmissionService.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
  */
 namespace OCA\Mail\Service;
 
+use Horde_Mime_Part;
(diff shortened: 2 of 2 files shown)


kesselb:
I'm hesitant to overwrite the Horde classes just to unset the Content-Disposition header. It might be better to avoid setting it altogether. I've found that this issue is more complex than initially thought. However, moving setDisposition and setName into the if-block resolved the issue for me. 

Nevertheless, we should extend the mail provider API to allow the initiator to specify their preference rather than relying on guesswork.

```diff
Index: lib/Service/TransmissionService.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/Service/TransmissionService.php b/lib/Service/TransmissionService.php
--- a/lib/Service/TransmissionService.php	(revision f4eef6e93599f5cb293b4ad5c4492004cefb3bcb)
+++ b/lib/Service/TransmissionService.php	(date 1741728908653)
@@ -81,10 +81,10 @@
 
 		try {
 			[$localAttachment, $file] = $this->attachmentService->getAttachment($account->getMailAccount()->getUserId(), (int)$attachment['id']);
+			$omitContentDisposition = false;
+
 			$part = new \Horde_Mime_Part();
 			$part->setCharset('us-ascii');
-			$part->setDisposition('attachment');
-			$part->setName($localAttachment->getFileName());
 			$part->setContents($file->getContent());
 
 			/*
@@ -99,9 +99,17 @@
 
 			$part->setType($contentTypeHeader->value);
 			foreach ($contentTypeHeader->params as $label => $data) {
+				if ($contentTypeHeader->value === 'text/calendar' && $label === 'method') {
+					$omitContentDisposition = true;
+				}
 				$part->setContentTypeParameter($label, $data);
 			}
 
+			if ($omitContentDisposition === false) {
+				$part->setName($localAttachment->getFileName());
+				$part->setDisposition('attachment');
+			}
+
 			return $part;
 		} catch (AttachmentNotFoundException $e) {
 			$this->logger->warning('Ignoring local attachment because it does not exist', ['exception' => $e]);
```

ChristophWurst (approved):

Issue #10788, opened 2025-03-04
Microsoft 365 appointment invitations not working


Luncheon3462:
### Steps to reproduce

See closed issue 703 at https://github.com/nextcloud/mail/issues/703
Nextcloud mail 4.2.3 still cannot handle calendar invitations send from Microsoft 365 accounts.  There is no way to add them to the nextcloud calendar.   Even if one tries to use 'create event', it does not handle invitations for recurring events. 

### Expected behavior

Nextcloud should properly handle invitations. 

### Actual behavior

There is no way to add events to the calendar.   Even when using 'create event', recurring events from 365 cant be added. 

### PHP engine version

None

### Web server

None

### Database

None

SebastianKrupinski:
hi @Luncheon3462 

Can you post a screen capture of the issue?

SebastianKrupinski:
@Luncheon3462 a screen shot of what you seen on the screen, with the MS invitaion

Luncheon3462:
@SebastianKrupinski 

<img width="768" alt="Image" src="https://github.com/user-attachments/assets/425be0b1-751d-4ee6-adc7-224c9c74c013" />

kesselb:
Hi @Luncheon3462, 

Thanks for reaching out. 

Is it possible to create an event at your Microsoft 365 instance and invite "danielk@nextcloud.com" and "sebastiank@nextcloud.com"? That's the easiest way to debug calendar invitations.

Luncheon3462:
@kesselb @SebastianKrupinski Regretfully, I would not be comfortable doing that because of the IT policies in place.   I know that is frustrating to you ( and me) since you are trying to help troubleshoot.   Other alternatives?

SebastianKrupinski:
@Luncheon3462 Thank you for the image. That helps.

Can i ask you one more question, do you know if the message was created with Microsoft Outlook Desktop? Or Office 365 Outlook browser edition?

SebastianKrupinski:
@kesselb I have access to a O365 service and Desktop outlook, I'll send us a message from them.

Luncheon3462:
@SebastianKrupinski the invite for the screenshot was O365 browser.  But I've also tried with desktop app (windows) and have same result.

SebastianKrupinski:
@Luncheon3462 Thank you for confirming

SebastianKrupinski:
After doing some testing the issue seems to be a malformed trigger duration string in O365 calendar invitations.

[...]

PR #10852, opened 2025-03-11
fix: replace broken trigger duration string before prasing

SebastianKrupinski:
Resolves: #10788

### DIFF
--- a/src/components/Imip.vue
+++ b/src/components/Imip.vue
@@ -247,8 +247,12 @@ export default {
 		attachedVCalendar() {
 			const parserManager = getParserManager()
 			const parser = parserManager.getParserForFileType('text/calendar')
-			parser.parse(this.scheduling.contents)
-
+			// Temporarily fix the Microsoft invitation with a P trigger until upstream can be fixed
+			let contents = this.scheduling.contents
+						.replace(/TRIGGER:P[\r\n]/, "TRIGGER:PT0S")
+						.replace(/TRIGGER;RELATED=START:P[\r\n]/, "TRIGGER;RELATED=START:PT0S")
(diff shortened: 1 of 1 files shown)


kesselb:
The reason for https://github.com/nextcloud/mail/issues/10788 is therefore https://github.com/nextcloud/calendar-js/issues/790?

Can you still reproduce the issue on main and if yes, is it a different error message? If so, then the patch for calendar-js did not cover all cases.

Could you please invite Richard and me to such an outlook event?

SebastianKrupinski:
> The reason for #10788 is therefore [nextcloud/calendar-js#790](https://github.com/nextcloud/calendar-js/issues/790)?

Ah... I didn't know that you had already been working on this.

> Can you still reproduce the issue on main and if yes, is it a different error message? If so, then the patch for calendar-js did not cover all cases.

Yes, I reproduced it yesterday, while looking in to another issue.

> Could you please invite Richard and me to such an outlook event?

Sure.

kesselb:
Your invitation works for me.

Could you retry after running git switch main; git pull; npm ci; npm run dev?

SebastianKrupinski:
> Your invitation works for me.
> 
> Could you retry after running git switch main; git pull; npm ci; npm run dev?

Npm CI made the difference. I don't usually run this.

All good not, closing this PR.

kesselb:
@SebastianKrupinski  :+1: 

So it's fixed for main and stable4.3, but not for stable4.2. I assume it's possible to also backport the calendar-js update to stable4.2, but that's something for @ChristophWurst to decide. If backporting it not an option, we can still apply this change to stable4.2.

ChristophWurst:
Let's backport

PR #10904, opened 2025-03-18
fix: go back warning discards the reply text

### LINKED ISSUE #10818: Missing attachments warning discards reply
### Steps to reproduce

1. Open the app
2. Open a message
3. Click reply
4. Enter some reply text and include "attachment attached"
5. Click *Send*
6. See the attachment warning
7. Click *Go back*

### Expected behavior

My reply text is still there

### Actual behavior

My reply text is gone

### Mail app version

4.3

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes #10818

this was a nasty one. 
We have 5 states here:
1. error
2. loading for attachment uploading
3. Loading for sending
4. warning
5. composer

When it switches to warning/error/loading and back, it completely destroys modal content because of v-if, including Composer instance.

Restructuring the component, and keeping alive the states plus the composer, fixes the issue. We shouldnt replace the entire modal content with loading/warning but only the composer.

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -8,70 +8,61 @@
 		:name="modalTitle"
 		:additional-trap-elements="additionalTrapElements"
 		@close="$event.type === 'click' ? onClose() : onMinimize()">
-		<EmptyContent v-if="error"
-			:name="t('mail', 'Error sending your message')"
-			class="empty-content"
-			role="alert">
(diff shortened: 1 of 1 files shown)


st3iny (approved):
Tested and works, good job!

Issue #9944, opened 2024-07-30
Element spacing should use multiples of --default-grid-baseline


ChristophWurst:
### Steps to reproduce

1. Inspect the app DOM

### Expected behavior

Spacing uses multiples of var(--default-grid-baseline)

### Actual behavior

Spacing in this app uses "magic" numbers like 24px, 10px, etc

### PHP engine version

None

### Web server

None

### Database

None

PR #10919, opened 2025-03-22
Use css variables instead of magic numbers

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/9944

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -682,11 +682,11 @@ export default {
 	flex-grow: 1;
 	text-align: center;
 	color: var(--color-text-lighter);
-	margin-bottom: 10px;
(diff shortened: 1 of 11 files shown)


src/components/AccountForm.vue
```diff
@@ -682,11 +682,11 @@ export default {
 	flex-grow: 1;
 	text-align: center;
 	color: var(--color-text-lighter);
-	margin-bottom: 10px;
+	margin-bottom: calc(var(--default-grid-baseline) * 2);
```
ChristophWurst, line 685: 4x2=10? :thinking:

ChristophWurst (approved):
Tested and can't see any alignment issues -> :shipit: 

nit: commit these changes as `refactor`. You are changing code without intending to change functionality.

PR #10924, opened 2025-03-25
fix: delete tag modal

### LINKED ISSUE #10101: Polish confirmation modal
### Steps to reproduce

1. Select a message
2. Pick edit tags
3. Delete a tag

### Expected behavior

The modal looks a bit unfinished. 
For example there's a lot of padding to the top but no spacing to the action items. 

### Actual behavior

image

### Mail app version

629a9fa693648d9414ecc8cb0fc3f315305dbba8

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

keremsemiz:
fixes #10101 
Styling fixes to confirmation modal.


src/components/ConfirmationModal.vue
```diff
@@ -67,11 +67,31 @@ export default {
...
 <style lang="scss" scoped>
 .confirm-modal {
-	padding: 20px;
+	padding: 24px 32px;
+
+	h2 {
+		margin-bottom: 20px;
+		font-size: 24px;
+		font-weight: 600;
+	}
```
GretaD, line 76: you dont need this part, the buttons have the styling from nc/vue library and it should stay that way, so it looks the same everywhere

src/components/ConfirmationModal.vue
```diff
@@ -67,11 +67,31 @@ export default {
...
 
 	&__buttons {
 		display: flex;
-		justify-content: space-between;
+		justify-content: flex-end;
+		gap: 16px;
+		margin-top: 30px;
+		padding-top: 20px;
+
+		:deep(button) {
```
GretaD, line 85: also, you dont need the v:deep. You only need to add space on top of the buttons, and you need to make sure the buttons have space in between.

src/components/ConfirmationModal.vue
```diff
@@ -67,11 +67,31 @@ export default {
...
+	}
 
 	&__buttons {
 		display: flex;
-		justify-content: space-between;
+		justify-content: flex-end;
+		gap: 16px;
+		margin-top: 30px;
+		padding-top: 20px;
+
```
GretaD, line 84: this part is correct, you want to add either a margin or a padding, not both as they both add space. 

The buttons should have space between so they are aligned left and right. So keep that as it is.

src/components/ConfirmationModal.vue
```diff
@@ -60,18 +60,23 @@ export default {
...
+	padding: 0 20px 20px 20px;
 	&__buttons {
 		display: flex;
 		justify-content: space-between;
+		margin-top: 30px;
 	}
 }
+:deep(.dialog__name){
+	display: none;
+}
```
ChristophWurst, line 80: Could this be problematic for a11y?

ChristophWurst:
Stale and conflicting

PR #10928, opened 2025-03-26
feat: save the composer state per account

### LINKED ISSUE #10627: Composer modal should remember the state per account

The default view of composer is that when you add a recipient, the right pane with the info is shown. If i hide that pane, i would like to keep it that way until i decide to show it up again. I would like to choose when i show the info panel.

As proposed by designers, remember the state per account.

_No response_

### Additional context

_No response_

GretaD:
This is a bit tricky because those 2 buttons were supposed to maximize and collapse the composer modal. But now with the recipient details, its about showing or hiding the details. I believe this pr makes it more intuitive what those buttons do.

https://github.com/user-attachments/assets/6e510fd8-6e6d-4d24-945d-5b180eba1555

fixes #10627

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -13,7 +13,7 @@
 				<NcButton class="maximize-button"
 					type="tertiary-no-background"
 					:aria-label="t('mail', 'Maximize composer')"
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Is this really a feat or just a fix?

st3iny (approved):
Tested and works.

I have no preference regarding fix/feat as we will have a minor release soon anyway.

Issue #5913, opened 2021-12-30
Disable Mail Threading


Lucksmith369:
## Feature request

As a Nextcloud user I prefer to use my email program without threads and look at messages separately.

## Work packages

* Add a checkbox per Nextcloud account (or Nextcloud Mail account - to be decided) to turn threading off. It's opt-out and, feature will be on by default.
* When threading is *enabled* do everything like before. Make sure previously not threaded accounts are fully rethreaded if necessary.
* When threading is *disabled* set thread_root_id=message_id for all messages and skip threading after synchronization
* User documentation

## Follow-ups

* Add an admin toggle for user defaults
  * Admin leaves thread on -> threading is on for users unless they disable it
  * Admin opts out -> threading is off for users unless they enable it
  * This will then require admin documentation

## Original request

<details>

Can you please add a Setting to disable Threading like roundcube do?

</details>

ChristophWurst:
mind elaborating why that is desirable?

you seem to have ignored the issue template with your feature request. please follow it next time.

Lucksmith369:
i got this question many times from diverent Customers, because they dont want the Threading feature, sometimes Emails have the same Subject but are not for the same purpose.

in Roundcube you can choose between the two options.
If you can tell me where i can find the part of Code for Threading, i can implement it like in Roundcube.

i read actually you do Threading in Backend, i think a Frontend implementation is a better Choice for this.


sry for the issue template

G-T100:
Any news about this...

ChristophWurst:
No

nordxONE:
Just came across this thread after searching for my "missing mail problem". I couldnt find a mail i was searching in he Mail-app of nextcloud, but it was there in roundcube.

For me this threading is very annyoing and makes me avoid using the Mail app at all. Please add this feature so i can see all my mails chronolical.

Big thumbs up for this feature!!

ChristophWurst:
OK

Scott07uk:
+1 I find it much harder to manage my mailbox with threadding, so would prefer to turn it off.

ChristophWurst:
Just add your :+1: to https://github.com/nextcloud/mail/issues/5913#issue-1091134366 to give the ticket more weight

arnowelzel:
I check the mail app from time to time and want to add - yes, please add an option to disable threading!

One major drawback: with threading the UI does **not** use the available browser viewport but limit the display to a certain design with a local scroll bar inside the box which is used for the mail content:

[...]

PR #10945, opened 2025-03-29
feat: Allow user to control message view mode

SebastianKrupinski:
Resolves: #5913

## Summary

- Adds the ability for a user to control message viewing mode (Threaded View or Singleton View)
- Adds the ability for a administrator to preset default message viewing mode (Threaded View or Singleton View)
- Changes behavior of drag and dropping message(s) to folder depending on viewing mode (Threaded View or Singleton View)
- Changes behavior of Message List -> Message Actions menu (Move, Delete, Snooze) depending on viewing mode (Threaded View or Singleton View)

## User Setting
Screenshot 2025-03-29 164914

## Administrator Settings
Screenshot 2025-03-29 164801

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -124,6 +124,7 @@ class MessagesController extends Controller {
 	 * @param int $cursor
 	 * @param string $filter
 	 * @param int|null $limit
+	 * @param string $view returns messages in requested view ('singleton' or 'threaded')
 	 *
 	 * @return JSONResponse
 	 *
@@ -134,7 +135,8 @@ class MessagesController extends Controller {
 	public function index(int $mailboxId,
 		?int $cursor = null,
 		?string $filter = null,
-		?int $limit = null): JSONResponse {
+		?int $limit = null,
+		?string $view = null): JSONResponse {
 		try {
 			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $mailboxId);
 			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
@@ -143,17 +145,21 @@ class MessagesController extends Controller {
 		}
 
 		$this->logger->debug("loading messages of mailbox <$mailboxId>");
+		$sort = $this->preferences->getPreference($this->currentUserId, 'sort-order', 'newest') === 'newest' ? IMailSearch::ORDER_NEWEST_FIRST: IMailSearch::ORDER_OLDEST_FIRST;
+		
+		$view = $view === 'singleton' ? IMailSearch::VIEW_SINGLETON : IMailSearch::VIEW_THREADED;
 
-		$order = $this->preferences->getPreference($this->currentUserId, 'sort-order', 'newest') === 'newest' ? 'DESC': 'ASC';
+		$messages = $this->mailSearch->findMessages(
+			$account,
+			$mailbox,
+			$sort,
+			$filter === '' ? null : $filter,
+			$cursor,
+			$limit,
+			$view
+		);
 		return new JSONResponse(
-			$this->mailSearch->findMessages(
-				$account,
-				$mailbox,
-				$order,
-				$filter === '' ? null : $filter,
-				$cursor,
-				$limit
-			)
+			$messages
 		);
 	}
 
--- a/lib/Service/Search/SearchQuery.php
+++ b/lib/Service/Search/SearchQuery.php
@@ -12,6 +12,8 @@ namespace OCA\Mail\Service\Search;
 class SearchQuery {
 	/** @var int|null */
 	private $cursor;
+	
+	private bool $threaded = true;
 
 	/** @var Flag[] */
 	private $flags = [];
@@ -70,6 +72,14 @@ class SearchQuery {
 	 */
 	public function setCursor(int $cursor): void {
 		$this->cursor = $cursor;
+	}
+
+	public function getThreaded(): bool {
+		return $this->threaded;
+	}
+	
+	public function setThreaded(bool $threaded): void {
+		$this->threaded = $threaded;
 	}
 
 	public function getMatch(): string {
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -346,6 +346,11 @@ return [
 			'verb' => 'PUT'
 		],
 		[
+			'name' => 'settings#setLayoutMessageView',
+			'url' => '/api/settings/layout-message-view',
+			'verb' => 'PUT'
+		],
+		[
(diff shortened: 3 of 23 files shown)


GretaD:
before



after

GretaD (approved):
it works as described

ChristophWurst (changes requested):
Tested and doesn't work. It's only possible to open the latest received message of a thread. I can not open earlier ones, because they are not shown in the message list. The list is still threaded.

SebastianKrupinski:
> Tested and doesn't work. It's only possible to open the latest received message of a thread. I can not open earlier ones, because they are not shown in the message list. The list is still threaded.

Correct! I was not aware we stacked messages in the list, I have already made the backed and front end changes, just need to updated some backend tests, will push the update soon.

SebastianKrupinski:
> Tested and doesn't work. It's only possible to open the latest received message of a thread. I can not open earlier ones, because they are not shown in the message list. The list is still threaded.

Corrected, message are no longer stacked in message list.

image

lib/Controller/MessagesController.php
```diff
@@ -134,7 +134,9 @@ public function __construct(string $appName,
 	public function index(int $mailboxId,
 		?int $cursor = null,
 		?string $filter = null,
-		?int $limit = null): JSONResponse {
+		?int $limit = null,
+		?string $sort,
```
ChristophWurst, line 138: parameter and allowed values is missing in phpdoc

lib/Controller/MessagesController.php
```diff
@@ -134,7 +134,9 @@ public function __construct(string $appName,
...
+		?string $view): JSONResponse {
```
ChristophWurst, line 139: same

ChristophWurst (commented):
Looks sane

Are there any plans to skip building threads if the user doesn't use them anyway?

lib/Controller/MessagesController.php
SebastianKrupinski: Done

lib/Controller/MessagesController.php
SebastianKrupinski: Done

lib/Controller/MessagesController.php
```diff
@@ -124,6 +124,8 @@ public function __construct(string $appName,
 	 * @param int $cursor
 	 * @param string $filter
 	 * @param int|null $limit
+	 * @param string $sort		returns messages in requested order ('newest' or 'oldest')
```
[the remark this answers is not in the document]
SebastianKrupinski: We can't use IMailSearcH::sort_* here the UI sends 'newest' or 'oldest' and the value of IMailSearch::ORDER_* is 'DESC' and 'ASC'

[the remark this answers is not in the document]
ChristophWurst: And this can't be streamlined?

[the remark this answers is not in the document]
SebastianKrupinski: This was existing code for the most part, the only change to the sort was that the UI now sends the sort order vs the controller pulling the value from the database every time.

I didn't want to refactor sort as it was out of scope for this PR and would make the review more difficult. 

I can address this in a follow-up if you want, or do you want it as part of this PR?

lib/Controller/MessagesController.php
```diff
@@ -143,17 +147,25 @@ public function index(int $mailboxId,
...
-		$order = $this->preferences->getPreference($this->currentUserId, 'sort-order', 'newest') === 'newest' ? 'DESC': 'ASC';
+		if ($sort === null) {
+			$sort = $this->preferences->getPreference($this->currentUserId, 'sort-order', 'newest') === 'newest' ? IMailSearch::ORDER_NEWEST_FIRST: IMailSearch::ORDER_OLDEST_FIRST;
+		} else {
+			$sort = $sort === 'newest' ? IMailSearch::ORDER_NEWEST_FIRST: IMailSearch::ORDER_OLDEST_FIRST;
+		}
+		if ($view !== null) {
+			$view = $view === 'singleton' ? IMailSearch::VIEW_SINGLETON : IMailSearch::VIEW_THREADED;
+		}
+		$messages = $this->mailSearch->findMessages(
```
ChristophWurst, line 158: Inline the var and you have a smaller diff :sparkles:

lib/Controller/MessagesController.php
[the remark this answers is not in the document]
ChristophWurst: Do a follow-up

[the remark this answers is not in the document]
SebastianKrupinski: Done, removed the sort code changes.

SebastianKrupinski:
Will fix lint errors on final approval

ChristophWurst:
Apologies if my communication wasn't clear from the ticket and task description (https://github.com/nextcloud/mail/issues/5913#issue-1091134366). I should have communicated this more actively.

Do you see an issue with handling "dethreading" on the database level? With the latest changes we see lots of conditional logic for threaded vs non threaded views. When every message forms a thread of one, we don't need the conditional logic, do we?

Building threads needs a check to be skipped with any solution. With the current state they are still built:

```
...
[debug] Building threads for account 4634
[debug] Account 4634 has 30 messages with threading information
[debug] Threading 30 messages - build ID table took 0s. 15/16MB memory used
[debug] Threading 30 messages - build root container took 0s. 15/16MB memory used
[debug] Threading 30 messages - free ID table took 0s. 15/16MB memory used
[debug] Threading 30 messages - prune containers took 0s. 15/16MB memory used
[debug] Threading 30 messages - group by subject took 0s. 16/16MB memory used
[debug] Threading 30 messages took 0s
...
```

SebastianKrupinski:
> Do you see an issue with handling "dethreading" on the database level? With the latest changes we see lots of conditional logic for threaded vs non threaded views. When every message forms a thread of one, we don't need the conditional logic, do we?
> 
> Building threads needs a check to be skipped with any solution. With the current state they are still built:

I see what you mean on the building threads part, this should now be addressed.

As for the conditional logic on the front end, if I understand your comments properly, yes we do (but I might be missing something), can we discuss on Monday?

ChristophWurst:
Okay

lib/Service/Search/SearchQuery.php
```diff
@@ -12,6 +12,8 @@
 class SearchQuery {
 	/** @var int|null */
 	private $cursor;
+	
+	private bool $stacked = true;
```
ChristophWurst, line 16: to keep terminology consistent please use `threaded`

ChristophWurst (approved):
Let's use the current approach

PR #10938, opened 2025-03-27
feat(deps): Add PHP8.4 support

neufeind:

PR #10959, opened 2025-04-02
feat(deps): Add PHP8.4 support

kesselb:
Close #10938 

Rebased version of https://github.com/nextcloud/mail/pull/10938


ChristophWurst (approved):
Thanks for the fix @neufeind and @kesselb for the clean-up :)

PR #10960, opened 2025-04-02
feat(deps): Add PHP8.4 support

neufeind:


neufeind:
There is already a cleanup pull-reqest by @kesselb at #10959

PR #10963, opened 2025-04-03
[stable4.3] fix: printing email threads and singular emails (manual backport)

GVodyanov:
fix: printing email threads and singular emails


st3iny (approved):
Tested and works (Firefox).

PR #10957, opened 2025-04-02
Fix: Remove extraneous JSON preamble from LLM responses

sto-j:
The most of the LLM's return in JSON markdown format, example:
\```json
{
  "reply1": "Received, thanks...",
  "reply2": "Thanks for forwarding this..."
}
\```
Adding additional instructions didn't help. 

This is solution to strip:
\```json
\```
from the answer.

PR #10966, opened 2025-04-04
Fix: Remove extraneous JSON preamble from LLM responses

hamza221:
Added unit test for #10957
Was not able to push to @sto-j 's branch

### DIFF
--- a/tests/Unit/Service/AiIntegrationsServiceTest.php
+++ b/tests/Unit/Service/AiIntegrationsServiceTest.php
@@ -158,6 +158,44 @@ class AiIntegrationsServiceTest extends TestCase {
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


tests/Unit/Service/AiIntegrationsServiceTest.php
```diff
@@ -157,6 +157,43 @@ public function testSmartReply(): void {
 			$result
 		);
 	}
+	public function testSmartReplyMarkdownFormat(): void {
```
ChristophWurst, line 160:
```suggestion

	public function testSmartReplyMarkdownFormat(): void {
```

ChristophWurst (approved):

PR #10969, opened 2025-04-04
feat:support darkmode in html messages

hamza221:
| b | a |
|--------|--------|
| image | image |


ChristophWurst (commented):
Does the font style apply to nested elements too?

I believe we experimented with this in the past. It's tricky.

hamza221:
If the HTML message is highly customised, not a simple text message like the screenshot attached. 
There's a high chance that the colors set by the senders will be unreadable, an altering all colors of the message can hinder the viewing experience, not easy to maintain, not easy to implement. 
Closing this for now until a better idea arises

PR #10970, opened 2025-04-06
fix: handle recurring events and show better message to user

SebastianKrupinski:
Resolves:

- Show message when invitation attendees do not match any know mail account instead of just nothing
- Check if the event is recurring and has any dates in the future. At the moment if the user does not accept the recurring event before the start of the event there is no way to accept it afterwards
- Improved messages on why the invitation accept/decline controls are not displayed

image

image


ChristophWurst (approved):
:clap: very nice

PR #10987, opened 2025-04-11
perf(imap): avoid double login during mailbox sync

ChristophWurst:
Sentry profiles showed that we log in twice within one request. This is easy to avoid.

https://blackfire.io/profiles/compare/182c94b7-a971-471d-b49a-125a54660523/graph


st3iny (approved):
:elephant:

Issue #7475, opened 2022-10-21
Unread Mail widget gives different results through Dashboard API (lists trash too)


AlvaroBrey:
### Steps to reproduce

1. Have some unread mails in the Trash, and one in the Inbox
2. See the Unread Mail widget in the web dashboard, which shows only the unread mail in the Inbox
3. Get the unread mail through the API:
`curl -u user:pass "http:/nextcloud.address/ocs/v2.php/apps/dashboard/api/v1/widget-items?widgets[]=mail-unread" 
    -H "Accept: application/json" `

### Expected behavior

Dashboard API should return the same unread mails as seen in the web dashboard, i.e. only the ones in Inbox.

### Actual behavior

API response contains the unread mail in Inbox, followed by the unread mails in Trash.

### Additional info

Versions are whatever we have in our company server

Originally reported in https://github.com/nextcloud/android/issues/10873

ChristophWurst:
@st3iny mind having a look?

ChristophWurst:
Because of https://github.com/nextcloud/mail/pull/8831

ChristophWurst:
I don't think logs help here.

Best would be to attach a debugger and step through the code to find out where the messages come from. E.g. by analyzing the db query.

PR #11002, opened 2025-04-15
fix: don't show important or unread emails in trash

kesselb:
Fix https://github.com/nextcloud/mail/issues/7475
Fix https://github.com/nextcloud/mail/issues/8849

To exclude messages in the trash or junk, we need to pass the IDs of those mailboxes to the search. Previously, MailSearch.findMessagesGlobally expected a filter string like "is:unread" and fed it through a parser to create a SearchQuery object. I've changed the API to expect a SearchQuery object directly. The parsing is now handled a step earlier when processing the incoming search request.

The SearchQuery is used for searching across all mailboxes and for a given mailbox. To distinguish these cases more clearly, we now have GlobalSearchQuery. For example excluding certain mailbox IDs doesn't make sense when searching within a specific mailbox anyway. The MailSearch API still feels a bit unfinished, but this PR is already large enough, and the current state looks good to me.

Note: The special_use (e.g., trash or junk) is stored as a JSON string in the database. It might also work to select those rows using LIKE and toggle a flag like ignoreJunk for the SearchQuery object instead of using the actual mailbox ID.

### DIFF
--- a/tests/Unit/Dashboard/ImportantMailWidgetTest.php
+++ b/tests/Unit/Dashboard/ImportantMailWidgetTest.php
@@ -1,0 +1,118 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace Unit\Dashboard;
+
+use ChristophWurst\Nextcloud\Testing\TestCase;
+use OCA\Mail\Account;
+use OCA\Mail\Dashboard\ImportantMailWidget;
+use OCA\Mail\Db\MailAccount;
+use OCA\Mail\Db\Message;
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\Search\GlobalSearchQuery;
+use OCA\Mail\Service\Search\MailSearch;
+use OCA\Mail\Service\Search\SearchQuery as MailSearchQuery;
+use OCP\AppFramework\Services\IInitialState;
+use OCP\IL10N;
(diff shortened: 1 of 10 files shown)


ChristophWurst:
> simply requests 14 items instead of 7 and filters out the items in the trash, hoping for the best.

I'm not sold on this. If you just cleaned up your inbox you easily have more than 14 deleted emails in your trash.

tests/Unit/Dashboard/ImportantMailWidgetTest.php
```diff
@@ -0,0 +1,128 @@
...
+		$account = new Account($mailAccount);
+
+		$this->accountService->expects($this->once())
+			->method('findByUserId')
+			->willReturn([$account]);
+
+		$items = $this->widget->getItems('bob', '1745340000', 7);
+
+		$this->assertCount(2, $items);
+	}
```
ChristophWurst, line 127: Tip for readable unit tests: structure them into three blocks: *arrange*, *act* and *assert*. Separate the blocks by one newline.

ChristophWurst (approved):
:wrench: :broom:

tests/Unit/Dashboard/ImportantMailWidgetTest.php
kesselb: You mentioned AAA a couple of times, yes 😉 I've changed the tests to follow the AAA pattern, even though I'm not a fan of it and find them rather difficult to read. Most of our tests have complex "Arrange" sections, and understanding them is quite hard when the code isn't at least somewhat structured.

ChristophWurst (approved):

st3iny (approved):
The new API looks good. Long live OOP :smile:

Issue #11000, opened 2025-04-15
Typo in function name: patchAccountiMutation is not defined


Phillipxh:
### Steps to reproduce

1. Log in to Nextcloud
2. Open the Mail app
3. Go to account settings (gear icon)
4. Open the "Automatic replies" section
5. Select "Use auto-response settings from system settings"
6. Click "Save"
7. Observe the error: `this.mainStore.patchAccountiMutation is not a function`

### Expected behavior

The automatic reply setting should save successfully without errors. All required functions should be defined and executed properly.

### Actual behavior

An error appears in the browser console:  
`this.mainStore.patchAccountiMutation is not a function`

I inspected the file `js/mail.account-settings.4aeed50dd21069e5aaaa.js` and found a typo in the function name:  
`patchAccountiMutation` (with an extra "i")

Changing it to `patchAccountMutation` resolves the issue.

### Mail app version

5

### Nextcloud version

31.0.3

### PHP engine version

8.2-fpm

### Web server

Apache2

### Database

MariaDB

kesselb:
Hi @Phillipxh, Thanks 👍 

Nice issue number you got 😃

PR #11003, opened 2025-04-15
fix: patchAccountiMutation is not defined

kesselb:
Fix https://github.com/nextcloud/mail/issues/11000

Regression from the pinia migration :see_no_evil:


ChristophWurst (approved):
:it:

PR #11015, opened 2025-04-22
fix: define appName and app version

GretaD:
on package.js we have this: 
```
"scripts": {
    "build": "NODE_ENV=production webpack --progress --config webpack.prod.js",
```

so i didnt add a new file, i edited the existing common.js webpack


st3iny (approved):
Fixes the app.

ChristophWurst:
^ so that any npm audit fix of nc/vue doesn't break the app again

PR #11026, opened 2025-04-24
fix enabeling sieve

kesselb:
https://github.com/nextcloud/mail/pull/11011 broke enabling sieve for an account.

SieveClientFactory.createClient was used when updating the configuration to validate if the credentials work before updating the account object.

### DIFF
--- a/lib/Sieve/SieveClientFactory.php
+++ b/lib/Sieve/SieveClientFactory.php
@@ -33,6 +33,8 @@ class SieveClientFactory {
 	}
 
 	/**
+	 * @param Account $account
+	 * @return ManageSieve
 	 * @throws ManageSieve\Exception
 	 */
 	public function getClient(Account $account): ManageSieve {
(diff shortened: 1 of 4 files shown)


GretaD (approved):
tested it and it works

SebastianKrupinski:
Hi,

I don't have anything against the additions you've made to the SieveController that use the imap login details as an alternative.

But why do we need two client creation functions (SieveClientFactory::getClient and SieveClientFactory::createClient), the code in #11011 creates the ManageSieve object the same way and the client connection check is in the constructor of ManageSieve therefore the connection was being checked either way, therefore they are functionally the same thing. Having additional code path in the SieveClientFactory for a single use is pointless in my opinion, it unnecessarily complicates the code.

lib/Sieve/SieveClientFactory.php
```diff
@@ -41,43 +33,62 @@ public function getClient(Account $account): ManageSieve {
...
-
 			if ($account->getDebug() || $this->config->getSystemValueBool('app.mail.debug')) {
-				$fn = 'mail-' . $account->getUserId() . '-' . $account->getId() . '-sieve.log';
-				$params['logger'] = new SieveLogger($this->config->getSystemValue('datadirectory') . '/' . $fn);
+				$logFile = $this->config->getSystemValue('datadirectory') . '/mail-' . $account->getUserId() . '-' . $account->getId() . '-sieve.log';
+			} else {
+				$logFile = null;
 			}
 
-			$this->cache[$account->getId()] = new ManageSieve($params);
```
SebastianKrupinski, line 77: The connection is checked in the constructor of ManageSieve. Therefore a separate function just to check the connection is really not needed.

kesselb:
To recap: On the main branch, it's currently not possible to enable Sieve for an account.

The createdClient method was, and is now again, used in [SieveController](https://github.com/nextcloud/mail/blob/6bf0776a0b9772789442d66281f9b52d65c70563/lib/Controller/SieveController.php#L153-L174) to validate that we can establish a connection with the given credentials without altering the Account/MailAccount object. I assume we could also fix the issue by moving the getClient call a few lines down, but it's a valid use case to just validate the credentials. Therefore, I went with reintroducing the method.

 > I don't have anything against the additions you've made to the SieveController that use the IMAP login details as an alternative.

The option to use the IMAP login details has existed since the initial implementation in 2021. I took the opportunity to make the code flow a bit clearer by introducing a helper variable and rearranging the crypt/decrypt calls. The functionality remains unchanged.

Issue #10545, opened 2025-01-06
Missing option for subfolders in actions (filters)


dasbenjo:
### Steps to reproduce

1. create subfolder in mailapp (e.g. subfolder for inbox folder)
2. create new filter and add action "fileinto"

### Expected behavior

you should be able to select subfolders

### Actual behavior

only mail folders in the root directory are displayed, no subfolders

### Mail app version

4.1.1

### Nextcloud version

30.0.4

### Mailserver or service

self-hosted (Postfix and Dovecot)

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database

MariaDB

GretaD:
> create new filter and add action "fileinto"

sorry, can you please clarify what exactly is this action? Are you trying to move a thread?

dasbenjo:
sure, I meant the action part in the filter settings (adding a new filter)
![Image](https://github.com/user-attachments/assets/4882ad84-57be-4351-8da7-3b0d49c457a6)

kesselb:
Thanks again for your report 👍 

We will merge patch later today and release within the next days.

Issue #10934, opened 2025-03-27
Improve filtering UX


ChristophWurst:
### Is your feature request related to a problem? Please describe.

Tbd

### Describe the solution you'd like

tbd

### Describe alternatives you've considered

tbd

### Additional context

tbd

ChristophWurst:
Duplicate of https://github.com/nextcloud/mail/issues/10655

PR #11031, opened 2025-04-25
fix: show submailboxes for filtering

GretaD:
Fix https://github.com/nextcloud/mail/issues/10545

Screenshot from 2025-04-28 15-56-10

- [x] requires [11026](https://github.com/nextcloud/mail/pull/11026)

### DIFF
--- a/src/components/MailboxInlinePicker.vue
+++ b/src/components/MailboxInlinePicker.vue
@@ -92,6 +92,7 @@ export default {
 input.vue-treeselect__input {
 	margin: 0;
 	padding: 0;
+	border: 1px solid var(--color-border-maxcontrast) !important;
 }
 .vue-treeselect__menu {
 	background: var(--color-main-background);
@@ -106,7 +107,8 @@ input.vue-treeselect__input {
 	border-radius: var(--border-radius-large);
 	}
 .vue-treeselect__placeholder, .vue-treeselect__single-value {
-	line-height: 44px;
+	line-height: 34px;
+	color: var(--color-main-text);
 }
 
 </style>
--- a/src/components/mailFilter/ActionFileinto.vue
+++ b/src/components/mailFilter/ActionFileinto.vue
@@ -3,23 +3,18 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<NcSelect ref="select"
-		:value="mailbox"
-		:options="mailboxes"
-		:required="true"
-		@input="onInput" />
+	<MailboxInlinePicker :account="account" :value="mailbox" @input="onInput" />
 </template>
 <script>
 
-import NcSelect from '@nextcloud/vue/components/NcSelect'
-import { mailboxHasRights } from '../../util/acl.js'
 import { mapStores } from 'pinia'
 import useMainStore from '../../store/mainStore.js'
+import MailboxInlinePicker from '../MailboxInlinePicker.vue'
 
 export default {
 	name: 'ActionFileinto',
 	components: {
-		NcSelect,
+		MailboxInlinePicker,
 	},
 	props: {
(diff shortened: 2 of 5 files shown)


src/components/mailFilter/ActionFileinto.vue
```diff
@@ -36,13 +37,28 @@ export default {
...
+				result.push({
+					value: mailbox,
+					label: mailbox.displayName,
+				})
+
+				const subMailboxes = this.mainStore.getSubMailboxes(mailbox.databaseId)
+				for (const subMailbox of subMailboxes) {
+					result.push({
+						value: subMailbox,
+						label: ` > ${subMailbox.displayName}`,
```
ChristophWurst, line 56: What about mailboxes of level two or higher?

For example

* INBOX
* Archive
  * 2025
    * Inbox
    * Sent

How will that show?

GretaD: I didnt test that scenario, and i am not sure if its shown at all. Does that mean that we need to show all the levels here? I will check it in a bit

GretaD: we discussed on a call, we should show all available mailboxes

GretaD: i used mailboxinlinepicker,

src/components/MailboxInlinePicker.vue
```diff
@@ -92,6 +92,7 @@ export default {
 input.vue-treeselect__input {
 	margin: 0;
 	padding: 0;
+	border: 1px solid var(--color-border-maxcontrast) !important;
```
ChristophWurst, line 95: What makes this necessary?

src/components/mailFilter/ActionFileinto.vue
```diff
@@ -31,18 +27,24 @@ export default {
...
-
-			return mailboxes.map((mailbox) => {
-				return mailbox.displayName
-			})
+		selectedMailbox: {
+			get() {
+				return this.currentSelectedMailboxId
+			},
+			set(selectedMailboxId) {
+				logger.debug('Selected mailbox set to', selectedMailboxId)
```
ChristophWurst, line 45: debug log?

src/components/MailboxInlinePicker.vue
GretaD: treeselect borders are 2px by default. Ours are 1px.

src/components/mailFilter/ActionFileinto.vue
GretaD: i added this because thats what we do on accountdefaultsettings.vue where we also use the MailboxInlinePicker.

ChristophWurst (approved):

kesselb (changes requested):
blocking

kesselb:
Screenshot From 2025-05-06 12-55-23

The border looks a bit odd in dark mode :disappointed:

GretaD:
> The border looks a bit odd in dark mode 😞

we can fix that with other design issues that Nimisha has added in the overview ticket #10934

ChristophWurst:
Fix it here so we have a releasable solution after main.

GretaD:
the second box is slightly bigger and i cannot change that
But i fixed the rest. I hope its good enough for merging

Screenshot from 2025-05-13 13-23-22

kesselb:
Hi @GretaD,

I'm still seeing that weird border :disappointed: 

It appears that p from the help-text is overflowing and then overlapping with the treeselect. 

What do you think about adding

```css
.help-text {
	margin-bottom: calc(var(--default-grid-baseline) * 2);
}
```

to UpdateModal.vue as a workaround?

kesselb (approved):

PR #11041, opened 2025-04-29
fix(imap): handle password decryption exception

ChristophWurst:
Fixes a crashing setup check when the password HMAC mismatches

1. ICrypto::decrypt can throw, which was not documented in PHPDoc
2. Move getting a client for the setup check into the try-catch to catch this exception too


kesselb (approved):

ChristophWurst:
I'm adding more tests to satisfy codecov

Issue #11040, opened 2025-04-29
ability to disable "mark as read" after 3 seconds


crtvrmn:
### Is your feature request related to a problem? Please describe.

id like to move from roundcoube to nextcloud mail due to threads beeing a pretty nice feature - but the automatic "mark as read after 3 (?) seconds" causes a lot of confusion in a team with "non techy" people, that are used to marking emails as read if they actually read them (not just scan the content).

### Describe the solution you'd like

a bool-checkbox in the settings
[ ] mark email automatically as read after __ seconds 

### Describe alternatives you've considered

- manually "unreading" emails via gui (not a viable option, would need the user to actually realize the mail was suddenly marked read )
- manually "unreading" emails via 'hidden' shortkey "u" (again, not a viable option - and does not work)
- keep using on roundcube (sadly, for now my solution)

### Additional context

- looked for and tried this on two instances (test and prod with 2x different imap mailaccounts on google and strato), 
- githubing, googleing, duckduckgoing and random llm prompts suggest that this really is not an option/feature at this point in the project.

kesselb:
Hi,

We are glad to hear that you are considering switching to Nextcloud Mail. The seen flag is automatically applied when a message is opened, with a slight delay of 2 seconds. Currently, there's no way to change the delay or turn it off. It seems like a reasonable feature request, especially because this feature is available in Roundcube and Thunderbird.

crtvrmn:
as a workaround, one could set the slight delay from 2000 to 3600000 ms (basically 10 hours of chilling) :) 

[https://github.com/nextcloud/mail/blob/2fd5134c23160eb20bb3c8bb140ce9115437ed82/src/components/ThreadEnvelope.vue#L660](in ThreadEnvelope.vue)

PR #11042, opened 2025-04-29
Fix toggle seen/unseen via shortcut 

kesselb:
Toggling seen/unseen via shortcut was broken due to a mixed up check if storing the state is allowed.

How to test:

- Open mailbox
- Select message
- Press `u`

Main: nothing happens
Here: seen/unseen toggles

Thanks to @crtvrmn for reporting it in https://github.com/nextcloud/mail/issues/11040.

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -386,17 +386,23 @@ export default {
 
 				break
 			case 'unseen':
-				if (this.hasSeenAcl()) {
+				logger.debug('marking as seen/unseen via shortcut')
+
+				if (!this.hasSeenAcl()) {
+					showWarning(t('mail', 'Your IMAP server does not support storing the seen/unseen state.'))
 					return
 				}
-				logger.debug('marking message as seen/unseen via shortkey', { env })
-				this.mainStore.toggleEnvelopeSeen({ envelope: env }).catch((error) =>
+
+				logger.debug('marking as seen/unseen', { env })
+				try {
+					await this.mainStore.toggleEnvelopeSeen({ envelope: env })
+				} catch (error) {
 					logger.error('could not mark envelope as seen/unseen via shortkey', {
 						env,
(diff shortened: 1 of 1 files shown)


src/components/Imip.vue
```diff
@@ -401,7 +401,7 @@ export default {
 				}
 
 				const defaultCalendar = calendarsForPicker.find(
-					(cal) => cal.url === this.currentUserPrincipal.scheduleDefaultCalendarUrl
+					(cal) => cal.url === this.currentUserPrincipal.scheduleDefaultCalendarUrl,
```
ChristophWurst, line 404: :eyes:

src/components/Mailbox.vue
```diff
@@ -386,17 +385,23 @@ export default {
 
 				break
 			case 'unseen':
-				if (this.hasSeenAcl()) {
+				logger.debug('marking as seen/unseen via shortcut')
```
ChristophWurst, line 388: do you want to keep these debug logs?

ChristophWurst (approved):

src/components/Mailbox.vue
kesselb: I've added a couple of debug statements, similar to the archive case, to make debugging easier. In the previous version, it was a bit hard to see if handleShortcut is even called. It might be a bit nicer to move the logic up and have a general debug statement like "shortcut x received." I don't have a strong opinion on this one, happy to remove them if you want me to.

src/components/Imip.vue
kesselb: Good catch! Not sure what's going on with eslint. I will just drop the cs commit, it does not even belong here.

src/components/Mailbox.vue
ChristophWurst: :shipit:

PR #11043, opened 2025-04-29
fix: use alias name for from header

### LINKED ISSUE #10087: Display name on email alias is not used
### Steps to reproduce

Sending email from an added alias address will use the main display name set for the mail account when sending.

Example:
I have one mail account with 2 email aliases on it:
- main@example.com (main address that was used to log in, configured with the name "Main")
- extra@example.com (This is added as an alias in the options of the mail account in the app, display name set to "Extra")

Sending mail from "main" works as expected, however sending as "extra" will use the display name "Main".

This has been tested on two separate Nextcloud instances (one self-hosted and one managed provider).

I have verified that the mail server itself does not do anything weird.
I've done this adding the same account to the email client in Vivaldi Browser, and here it allows sending mail from the same mailbox account but using separate alias addresses with different display names. And the receiver will see the display name set for the alias.

### Expected behavior

Sending from aliases will use the defined display name set on the alias. And the receiver will see the display name set for the alias.

### Actual behavior

Sending from aliases uses the primary display name. The receiver is always getting the primary display name set for the mailbox in the mail app.

### Mail app version

3.7.8

### Mailserver or service

Stalwart mail-server (self-hosted)

### Operating system

Ubuntu 22.04.4 LTS (for the self hosted instance)

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

MySQL

### Additional info

_No response_

### LINKED ISSUE #11027: Mail Alias Name not used
### Steps to reproduce

1.Mail app -> Account settings -> Aliases -> add Alias
2.fill in Name and email address fields for new alias
3.New message -> in From field, select new alias from drop down box
4.Send 

### Expected behavior

mail account: main, main@main.com
alias:  test, test@test.com

Send email using alias, recipient should see in "From" field the alias details for Name and Email address, e.g.:
From: test <test@test.com>

### Actual behavior

mail account: main, main@main.com
alias:  test, test@test.com

Send email using alias, recipient received in "From" field the Name defined in main account and email address defined in alias, e.g.:
From: main <test@test.com>

### Mail app version

5.0.2

### Nextcloud version

31.0.4

### Mailserver or service

braxmail.net

### Operating system

windows

### PHP engine version

PHP 8.3

### Nextcloud memory caching

_No response_

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

_No response_

kesselb:
Fix #11027
Fix #10087

### DIFF
--- a/lib/Service/MailTransmission.php
+++ b/lib/Service/MailTransmission.php
@@ -76,14 +76,23 @@ class MailTransmission implements IMailTransmission {
 		$bcc = $this->transmissionService->getAddressList($localMessage, Recipient::TYPE_BCC);
 		$attachments = $this->transmissionService->getAttachments($localMessage);
 
-		$alias = null;
+		$name = $account->getName();
+		$emailAddress = $account->getEMailAddress();
+
 		if ($localMessage->getAliasId() !== null) {
-			$alias = $this->aliasesService->find($localMessage->getAliasId(), $account->getUserId());
+			try {
+				$alias = $this->aliasesService->find($localMessage->getAliasId(), $account->getUserId());
+				$name = ($alias->getName() ?? $name);
+				$emailAddress = $alias->getAlias();
+			} catch (DoesNotExistException) {
+				$this->logger->debug('The assigned alias no longer exists. Falling back to the default name and email address. It is likely that the alias was deleted or deprovisioned in the meantime.', [
+					'aliasId' => $localMessage->getAliasId(),
+					'accountId' => $account->getId(),
+				]);
+			}
 		}
-		$fromEmail = $alias ? $alias->getAlias() : $account->getEMailAddress();
-		$from = new AddressList([
-			Address::fromRaw($account->getName(), $fromEmail),
-		]);
+
+		$from = Address::fromRaw($name, $emailAddress);
 
 		$attachmentParts = [];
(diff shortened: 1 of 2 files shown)


kesselb:
> Argument 1 of OCA\Mail\Address::fromRaw cannot be null, possibly null value provided

Indeed, that database column might be null. I doubt that one can create such an alias without messing in the database, but I will push a fixup with some additional hardening.

lib/Service/MailTransmission.php
```diff
@@ -76,14 +76,19 @@ public function sendMessage(Account $account, LocalMessage $localMessage): void
...
+		$emailAddress = $account->getEMailAddress();
+
 		if ($localMessage->getAliasId() !== null) {
-			$alias = $this->aliasesService->find($localMessage->getAliasId(), $account->getUserId());
+			try {
+				$alias = $this->aliasesService->find($localMessage->getAliasId(), $account->getUserId());
+				$name = ($alias->getName() ?? $name);
+				$emailAddress = $alias->getAlias();
+			} catch (DoesNotExistException) {
+			}
```
ChristophWurst, line 88: nit for debuggability: log at debug level

ChristophWurst (approved):

PR #11055, opened 2025-05-02
fix: normalize email addresses when retrieving from server

SebastianKrupinski:
resolves: client ticket - issue with email addresses being wrapped in single quotes


ChristophWurst (approved):
:elephant: did not test

lib/Address.php
```diff
@@ -117,4 +124,13 @@ public function equals($object): bool {
...
 			&& $this->getLabel() === $object->getLabel();
 	}
+
+	private static function normalizeAddress(string $address): string {
+		// remove single quotes and whitespace the might exist
+		// Examples:
+		// user1@example.com
+		// 'user1@example.com'
+		// ' user1@example.com'
+		return strtolower(trim(trim($address, "'")));
```
kesselb, line 134: As mentioned at https://github.com/nextcloud/mail/pull/10980#issuecomment-2787257686.

- `'test@example.org` is a valid email address.
- The current normalizer incorrectly changes it to `test@example.org`

Example: 

```php
if (str_starts_with($address, "'") && str_ends_with($address, "'")) {
	$address = substr($address, 1, -1); // or trim($address, "'"); whatever you prefer
}
return strtolower(trim($address));
```

Given that we need the same logic in the repair job, I'd suggest moving the normalization to a small helper or utility service to make it reusable and more testable.

lib/BackgroundJob/RepairRecipients.php
```diff
@@ -0,0 +1,58 @@
...
+	protected function run($argument): void {
+		// fetch all quoted emails
+		$select = $this->db->getQueryBuilder();
+		$select->select('id', 'email')
+			->from('mail_recipients')
+			->where(
+				$select->expr()->like('email', $select->createNamedParameter('\'%\'', IQueryBuilder::PARAM_STR))
+			)
+			->setMaxResults(1000);
+		$recipients = $select->executeQuery()->fetchAll();
```
kesselb, line 38:
```suggestion
		$result = $select->executeQuery();
		$recipients = $result->fetchAll();
		$result->closeCursor();
```

lib/BackgroundJob/RepairRecipients.php
```diff
@@ -0,0 +1,58 @@
...
+		// update emails
+		$update = $this->db->getQueryBuilder();
+		$update->update('mail_recipients')
+			->set('email', $update->createParameter('email'))
+			->where($update->expr()->in('id', $update->createParameter('id'), IQueryBuilder::PARAM_STR));
+		foreach ($recipients as $recipient) {
+			$id = $recipient['id'];
+			$email = $recipient['email'];
+			$email = trim(str_replace('\'', '', (string)$email));
```
kesselb, line 47:
```suggestion
			$email = trim(trim($email, "'"));
```

The best approach would be to reuse the logic from the address class. If we don't go with the helper/utility, we should still use trim to remove the first and last character (even if we can be very sure, due to the query, that we're only dealing with emails starting and ending with a `'`).

kesselb (changes requested):

PR #11113, opened 2025-05-08
fix(mime): ignore HTML parsing errors consistently

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/11112

While ignoring XML errors in MimeMessage I noticed that we use the same trick in other places already. Added a comment with the learning that PHP8.4 handles this better, and unified resetting the XML error handling to its previous value and the parsing options.

### DIFF
--- a/lib/Html/Parser.php
+++ b/lib/Html/Parser.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+namespace OCA\Mail\Html;
+
+use DOMDocument;
+use function libxml_clear_errors;
+use function libxml_use_internal_errors;
+
+class Parser {
+
+	/**
+	 * Parse a DOM document from a string
+	 *
+	 * @psalm-param non-empty-string $html
+	 *
+	 * DOMDocument uses libxml to parse HTML and that expects HTML4. It
+	 * is likely that some markup cause an error, which is otherwise
+	 * logged. Therefore, we ignore any error here.
+	 * @todo Migrate to \Dom\HTMLDocument::createFromString when this app uses PHP8.4+
+	 */
+	public static function parseToDomDocument(string $html): DOMDocument {
+		$document = new DOMDocument();
+		$previousLibxmlErrorsState = libxml_use_internal_errors(true);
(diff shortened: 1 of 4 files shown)


lib/Service/MimeMessage.php
```diff
@@ -71,9 +73,19 @@ private function buildMessagePart(?string $contentPlain, ?string $contentHtml):
...
+			/**
+			 * DOMDocument uses libxml to parse HTML and that expects HTML4. It
+			 * is likely that some markup cause an error, which is otherwise
+			 * logged. Therefore, we ignore any error here.
+			 * @todo Migrate to \Dom\HTMLDocument::createFromString when this app uses PHP8.4+
+			 */
+			$previousLibxmlErrorsState = libxml_use_internal_errors(true);
 			$doc = new DOMDocument();
 			$doc->loadHTML($source, LIBXML_NONET | LIBXML_HTML_NODEFDTD | LIBXML_HTML_NOIMPLIED);
+			libxml_clear_errors();
```
kesselb, line 86: It's intentional to only clear here?

kesselb (approved):
What about a small service / helper function?

kesselb (dismissed):
blocking

lib/Service/MimeMessage.php
ChristophWurst: Nope, bad copy-paste

ChristophWurst:
> What about a small service / helper function?

Good idea. I've pushed a minimal version of an abstraction

lib/Html/Parser.php
```diff
@@ -0,0 +1,35 @@
...
+	/**
+	 * Parse a DOM document from a string
+	 *
+	 * DOMDocument uses libxml to parse HTML and that expects HTML4. It
+	 * is likely that some markup cause an error, which is otherwise
+	 * logged. Therefore, we ignore any error here.
+	 * @todo Migrate to \Dom\HTMLDocument::createFromString when this app uses PHP8.4+
+	 */
+	public static function parseToDomDocument(string $html): DOMDocument {
+		$document = new DOMDocument();
```
kesselb, line 27: I think it should be "safe" to add something like $html === '' then return $document to avoid the psalm type coercion warning.

ChristophWurst: I've pushed domdocument's type up to my little abstraction's parameter. Psalm is happy. Are you? :wink:

kesselb (approved):

ChristophWurst:
/backport 0b682eb1ee1d5b61d818f478c4e457dd5391c3f7 to stable5.0

Issue #4557, opened 2021-02-18
Not able to configure a pure HTML signature without getting it defaced by the editor.


stevearbour:
### Expected behavior
I should be able to have a radio box like this : 
![image](https://user-images.githubusercontent.com/4413101/108425819-03dcb480-7209-11eb-9879-7fb298fa1a14.png)

### Actual behavior
The users are not able to set their email signature properly using optimized and properly formatted HTML. When they paste it inside there no option to not get their signature defaced by the CKEditor. Even by using javascript hack to fill the ckinstance end-up getting filtered. I can understand that you don't want to have PHP code and stuff in there, but sanitizing html markup has been something quite documented by now, we should be able to use a standard set of html markup tags, without being bothered by the editor. It should be concat with the message body. This is actually important because there no way any users want to be limited by the CKEditor for generating a signature and moving them from any others mail client such as thunderbird, it quite clear that this is a required (not optional but absolutely necessary) thing to fix this issues. Otherwise the employees prefer to continue to use desktop version of thunderbird and have the proper business image sent to customers when they send out email, not just deficient RTF signature.

### Mail app

**Mail app version:** 1.8.2

**Mailserver:** Unrelated, Exim.

### Server configuration

**Operating system**: CentOS

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 7.4

**Nextcloud Version:** 20.0.7

#### Client configuration
**Browser:** Chrome, latest.
**Operating system:** Mac, Windows and Linux, it all the same problem.

ChristophWurst:
> it quite clear that this is a required (not optional but absolutely necessary) thing to fix this issues.

Noted.

stevearbour:
I absolutely appreciate your understanding. If there are anything I can do to help, please let me know. 

Do you guys have a Patreon ?

ChristophWurst:
> I absolutely appreciate your understanding. If there are anything I can do to help, please let me know.

Not specific to this issue but in general we always need some good testers and people that can help triage and verify reported bugs :D



> Do you guys have a Patreon ?

Nope, as a company we don't work like that. But we have some very active community people like @StCyr and @st3iny, so maybe they are interested in some free lancing or similar :)

stevearbour:
I am fine with helping to test and report bugs when I see them and / or verifying some if it is possible. However, I was more aiming my though about this precise issue.

[...]

Issue #4599, opened 2021-02-26
Use editor toolbar instead of balloon editor


diemelone:
Fix the formatting tools to the top of the message input area (like most richtext editing fields)

Why? The marking and editing is hard to find if you don't know where to look

ChristophWurst:
@diemelone I appreciate your burst of new tickets but mind respecting the issue template?

ChristophWurst:
Let's see if anyone else is interested in this feature, I'm on the fence, this might be cool but it also might be some visual clutter

diemelone:
Sorry, didn't quite know what to write in to that template.
> 
> 
> @diemelone I appreciate your burst of new tickets but mind respecting the issue template?



I assume thats a quite easy feature, why not make a checkbox for fixed or dynamic formatting tools?
> 
> 
> Let's see if anyone else is interested in this feature, I'm on the fence, this might be cool but it also might be some visual clutter

ChristophWurst:
> I assume thats a quite easy feature, why not make a checkbox for fixed or dynamic formatting tools?

Every setting creates one more path that needs to be tested. Hence we try to have good default that work for the majority rather than tons of configuration that we then can barely maintain.

StCyr:
I would also prefer a toolbar. But maybe it is not convenient on small screens? Do we still really want to support small screens btw?

ChristophWurst:
@jancborchardt @nimishavijay what do you think?

nimishavijay:
Agreed with @jancborchardt, 
I think having it shown all the time is better. Since we are moving to a composer modal the text area is lesser, so we could not show it by default, but have a button that toggles it.

ChristophWurst:
Thanks for the quick feedback!

thstyl2000:
I would also like to add that it is a bit cumbersome to write lists with the floating toolbar, because you first need to write sth down and then convert it to a list.

ChristophWurst:
In discussion with @jancborchardt we found out that what we want is neither a fixed toolbar nor the current balloon (popover) but a floating toolbar that shows at the bottom of the editor, only if formatting is turned *on*.

jancborchardt:
Yup, and also only showing when toggling it via the icon – that is it of course stays open while you edit the text. Here is how Gmail does it – and actually they remember the last open/close state of the formatting bar, even across reloads:
Gmail default | Gmail with toolbar toggled open
-|-
[...]

PR #11125, opened 2025-05-09
fix: add html and source editing support

GretaD:
fixes [#4557](https://github.com/nextcloud/mail/issues/4557)

To add source editing support on our ckeditor, we need to change the editor itself.
The SourceEditing plugin in CKEditor 5 is currently supported only in the ClassicEditor.  Using the DecoupledEditor is not supported and will result in errors.

ref: https://github.com/ckeditor/ckeditor5/issues/12150#issuecomment-1200388379


st3iny (commented):
I can see the "Edit source" button but clicking it does nothing.

image

According to the official CKEditor demos it should show the source code in a slightly different editor pane: https://ckeditor.com/docs/ckeditor5/latest/features/source-editing/source-editing.html

GretaD:
To have the source editing, we need to change the position of the toolbar. It takes a lot of hacking to make it work in the bottom of the modal. I will consult with other fontend-ers but having like the ckeditor by default is better for us too.

Screenshot from 2025-05-14 11-43-19
Screenshot from 2025-05-14 11-42-50
Screenshot from 2025-05-14 11-42-22

ChristophWurst:
why does source editing not work for the balloon editor?

we placed the editor at the bottom intentionally.

GretaD:
> why does source editing not work for the balloon editor?

My understanding is that the source editing doesnt work for balloon editor for free versions. On their docs they say:
> We’ve recently released the premium [Enhanced source code editing](https://ckeditor.com/docs/ckeditor5/latest/features/source-editing/source-editing-enhanced.html) plugin in which we wanted to improve the source editing within CKEditor 5.

Enhanced source code editing, is only on paid version.

Also, see this ticket here: https://github.com/ckeditor/ckeditor5/issues/12150

Source editing has 2 states - html code style and the formatted style. Our composer doesnt work well with both of these states. To make it work, we need to force styling for both of these states. I tried and things kept breaking. If you want me to spend time to make it work for the bottom toolbar, im fine with that. But we have a better solution, without too much hacking.

GretaD:
> Okay, I see.
> 
> Top editor is not an option IMO, as it would revert #4599.

I think its not a revert, but rather a change in position. The balloon version is randomly positioned depending on the text you select. This pr and your pr, has a fixed toolbar, but one is on top, and one is on the bottom.

ChristophWurst:
Alright then we do it your way. Ignore my input.

st3iny (commented):
Tested and works.

Although the new editor does not seem to have a dark mode anymore. I'm not sure if this is the case on main too so it could not be related.

image

GretaD:
fixed, for the signature part and the composer 
Screenshot from 2025-06-03 10-05-09

src/components/TextEditor.vue
```diff
@@ -13,15 +17,14 @@
...
 			@ready="onEditorReady" />
-		<div ref="container" class="toolbar" />
 	</div>
 </template>
 
 <script>
 import CKEditor from '@ckeditor/ckeditor5-vue2'
 import AlignmentPlugin from '@ckeditor/ckeditor5-alignment/src/alignment.js'
 import { Mention } from '@ckeditor/ckeditor5-mention'
-import Editor from '@ckeditor/ckeditor5-editor-decoupled/src/decouplededitor.js'
```
ChristophWurst, line 24: Is the decoupled editor unused now? Can you share the latest details on why the decoupled editor was no longer suitable for this? Judging by the CSS below I take that you are pushing the classic editor to look like a decoupled one.

GretaD: yes, added details to the descr

ChristophWurst: okay then let's remove the package

PeopleInside:
When this fix will be available? Thank you!

hamza221:
image

PR #11150, opened 2025-05-15
feat: add context chat provider

edward-ly:


ChristophWurst:
It was decided that we should move on with this without https://github.com/nextcloud/server/issues/52852.

I'll look for a way to make Psalm understand the foreign code.

lib/AppInfo/Application.php
```diff
@@ -89,6 +91,7 @@
 
 final class Application extends App implements IBootstrap {
 	public const APP_ID = 'mail';
+	public const CONTEXT_CHAT_MESSAGE_MAX_AGE = intval(60 * 60 * 24 * 365.25); // 1 year
```
marcelklehr: Ah, nice, good idea to limit htis!

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,202 @@
...
+							$imapMessage->getSubject(),
+							$fullMessage['body'] ?? '',
+							'E-Mail',
+							$imapMessage->getSentDate(),
+							[$userId],
+						);
+
+						// Submit 100 items at a time
+						if (count($items) < 100) {
+							continue;
```
marcelklehr: Could move this up and use `break;` to short circuit the entire loop

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,202 @@
...
+		return $this->urlGenerator->linkToRouteAbsolute('mail.page.thread', [ 'mailboxId' => $mailboxId, 'id' => $id ]);
+	}
+
+	/**
+	 * Starts the initial import of content items into context chat
+	 *
+	 * @return void
+	 * @since 5.2.0
+	 */
+	public function triggerInitialImport(): void {
```
marcelklehr, line 121: This is guaranteed to be called in a background job, but it may be a good idea to distribute the work across multiple jobs instead of doing it all in one go.

I recommend to have one job per user perhaps? Maybe even chunked by message id so that one job run does 100 messages and then sleeps.

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,202 @@
...
+		private IMAPClientFactory $clientFactory,
+		private MailManager $mailManager,
+		private MessageMapper $messageMapper,
+		private IURLGenerator $urlGenerator,
+		private IUserManager $userManager,
+		private IContentManager $contentManager,
+	) {
+	}
+
+	public function handle(Event $event): void {
```
marcelklehr, line 50: We should have a guard here to check if context chat is actually available to avoid the performance penalty of running the below stuff for nothing when it is not available. (See julien's comment on the server pr)

lib/ContextChat/ContextChatProvider.php
edward-ly: Wouldn't that mean we only submit 100 emails maximum per mailbox, or was that your intention?

lib/ContextChat/ContextChatProvider.php
edward-ly: Same for `triggerInitialImport()` too, or is that guaranteed to only run when context chat is enabled?

lib/ContextChat/ContextChatProvider.php
edward-ly: A per-user `QueuedJob` sounds nice, but I don't know if that's actually possible right now. Or maybe we can [disable running jobs in parallel](https://docs.nextcloud.com/server/latest/developer_manual/basics/backgroundjobs.html#configuring-parallelism) if CPU consumption is a concern.

Also, if we split the job into chunks, I'm not sure what is the best way to figure out which emails have or haven't been imported already to avoid duplicate imports.

lib/ContextChat/ContextChatProvider.php
marcelklehr: Ah, you're right, I misread this

lib/ContextChat/ContextChatProvider.php
marcelklehr: Yes, triggerInitialImport is only run when context chat is enabled

lib/ContextChat/ContextChatProvider.php
marcelklehr: I envision a per-mailbox QueuedJob with a mailbox ID and a last message UUID argument. The job would then fetch n message UUIDs in said mailbox sorted in some way (alphanumerically, or by date), index those messages and reschedule itself with the new last message UUID argument.

edward-ly: Initial implementation is now done. Currently, the list of messages is being passed as a job argument and each message is removed from the array as they are marked as a new content item or skipped (this avoids duplicate fetches for the same message and minimizes the number of required jobs to process all messages). If we reach 100 content items before the entire message list is processed, we reschedule the job with the remaining unprocessed messages as the new argument.

For performance reasons, I decided to disable parallel runs for this job too.

lib/ContextChat/ContextChatProvider.php
```diff
@@ -54,45 +54,14 @@ public function handle(Event $event): void {
...
-
-				// Skip encrypted messages
-				if ($imapMessage->isEncrypted()) {
-					continue;
-				}
+			$this->jobList->add(SubmitContentJob::class, [
+				'userId' => $event->getAccount()->getUserId(),
+				'account' => $event->getAccount(),
+				'mailbox' => $event->getMailbox(),
+				'messages' => $event->getMessages(),
```
marcelklehr: I'm not sure this will work, as the job arguments will be JSON-serialized in the database as far as I know

edward-ly: Talking about the entire `$argument` array or just the messages array? Either way, I'm not sure if I see where the serializing would need to happen (unless the [docs](https://docs.nextcloud.com/server/latest/developer_manual/basics/backgroundjobs.html#registering-manually) are somehow outdated).

marcelklehr: The entire $arguments array is JSON-serialized. You're right, this is not mentioned in the docs. I think we should make sure that the input to IJobList#add is serializable and not too large. Adding all message IDs in there seems too large, I think, for example (see https://github.com/nextcloud/server/blob/daeb1e3a67121262c2a6d7b6fd96a9fdb11ad6f7/lib/private/BackgroundJob/JobList.php#L43 )

marcelklehr: So, as a job argument we can use arrays, numbers, strings, but no objects. And only 4000 characters JSON-serialized.

That is why I propose to pass some sort of sentinel message ID instead of all remaining messages. Then we can get all remaining messages from that. e.g. `'lastMessageId' => 502`, then we would fetch 100 messages that have IDs higher than 502.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,103 @@
...
+			// Skip older messages
+			if ($nextMessage->getSentAt() < $argument['startTime']) {
+				goto next;
+			}
+
+			$imapMessage = $this->mailManager->getImapMessage($client, $account, $mailbox, $nextMessage->getUid(), true);
+
+			// Skip encrypted messages
+			if ($imapMessage->isEncrypted()) {
+				goto next;
```
marcelklehr: Mmh, I'd feel better if we weren't using goto, even if it seems safe here.

marcelklehr (approved):
Looks really good to me! If we can finish the tests, I'd look forward to a review from the groupware team 💙

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,170 @@
...
+
+		if ($event instanceof NewMessagesSynchronized) {
+			$account = $event->getAccount();
+			$mailbox = $event->getMailbox();
+			$userId = $account->getUserId();
+			$accountId = $account->getId();
+			$mailboxId = $mailbox->getId();
+
+			// Check if there is a pending job for this mailbox already to avoid duplicates
+			$jobs = $this->jobList->getJobsIterator(SubmitContentJob::class, null, 0);
```
marcelklehr: This will not catch QueuedJobs that are currently running, sadly :/

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,103 @@
...
+
+		if (count($items) > 0) {
+			$this->contentManager->submitContent($this->contextChatProvider->getAppId(), $items);
+		}
+
+		// Schedule next job to process remaining messages
+		if ($nextMessage !== false) {
+			$newArgument = $argument;
+			$newArgument['nextMessageId'] = $nextMessage->getId();
+			$this->jobList->add(SubmitContentJob::class, $newArgument);
```
marcelklehr: This would prevent other jobs from running. Perhaps we should turn this into a TimedJob to also make sure it's caught by the check below. The next messageId argument would have to be stored in a database table perhaps. :/

lib/BackgroundJob/ContextChat/SubmitContentJob.php
edward-ly: Sure, maybe it's not worth saving a few lines of repetitive code in this case.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
edward-ly: As in, newly added jobs go to the front of the queue? I don't see why that would be a problem if that's the case.

lib/ContextChat/ContextChatProvider.php
edward-ly: That might be fine since the running job might not have retrieved the latest messages.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
marcelklehr: We had this in ContextChat as well. If one app schedules a lot of jobs that reschedule themselves to the front of the queue all the time, it blocks other jobs from being executed in a timely manner :/

lib/ContextChat/ContextChatProvider.php
marcelklehr: But it might lead to duplicate QueuedJob cascades that do the same thing, I think, right? I'm not sure how bad this would be, as context_chat would just deduplicate the entries anyway. Perhaps we can live with this.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
marcelklehr: I recommend using `IJobList#scheduleAfter` to reschedule after like 5mins or perhaps even more.

edward-ly: As long as the messages are being processed correctly, we'll reach the end of the mailbox eventually and the job list would continue with the next mailbox. But I guess in case something goes wrong with a mailbox, putting it further back in the queue wouldn't hurt anyway.

marcelklehr: > As long as the messages are being processed correctly, we'll reach the end of the mailbox eventually and the job list would continue with the next mailbox.

Yes, the messages are processed correctly. But we're not being a nice citizen :) Other apps use the jobs system too. We don't know how many mails there are and how many mailboxes, it could be thousands in a large instance. Until those are processed, these QueuedJobs will keep inserting themselves before all the other jobs in Nextcloud that also need to run.That's not good.

lib/ContextChat/ContextChatProvider.php
edward-ly: As long as there is only one pending job max per mailbox, it should be fine. As the message list is retrieved at the beginning of the job, it wouldn't know about new messages that may have came in since then. Maybe when the running job schedules the next job, we check if there's a pending job already there too and replace it if needed.

marcelklehr: Good idea!

marcelklehr (approved):

marcelklehr:
Looking forward to reviews from the Groupware team :)

lib/BackgroundJob/ContextChat/SubmitContentJob.php
ChristophWurst: Assume users will have up to 200 mailboxes and a total of 100k emails. There are still a few accounts with more data but that is the rough scale we have to expect.

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	public function getItemUrl(string $id): string {
+		// Get mailbox ID from message ID
+		$messages = $this->messageMapper->findByIds('', [(int)$id], '');
+		if (!$messages) {
+			return '';
+		}
+		$mailboxId = $messages[0]->getMailboxId();
+
+		return $this->urlGenerator->linkToRouteAbsolute('mail.page.thread', [ 'mailboxId' => $mailboxId, 'id' => $id ]);
```
kyteinsky: perhaps the item id could be constructed like this: `"$mailboxId::$id"` so this lookup is not necessary

lib/BackgroundJob/ContextChat/SubmitContentJob.php
kyteinsky: I agree with Marcel,  a non-concurrnet TimedJob might be better suited here. In addition to that a separate persistent DB table can be added to keep track of the messages/mails being indexed. Timed jobs are better tracked since they're persistent in the job list and we also don't do multiple joblist table lookups. We had this issue recently with context chat.
And the upsides of having a table would be to able to reference it later when we might need to lookup if an email has been indexed in context chat or not without reaching the context chat backend for it, and even show overall progress wrt. the the mail app.

lib/ContextChat/ContextChatProvider.php
edward-ly: By "lookup", you mean calling the URL generator or looking up the mailbox ID? I'm not sure if I follow.

marcelklehr (approved):

codecov.yml
```diff
@@ -3,4 +3,9 @@
 comment: false
 ignore:
   - "tests"
+  - "lib/BackgroundJob/ContextChat"
+  - "lib/ContextChat"
+  - "lib/Db/ContextChat"
+  - "lib/Migration/Version5200Date20250728000000.php"
+  - "lib/Service/ContextChat"
```
ChristophWurst: Please revert. We don't ignore production code from code coverage.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		// Remove job from the database while it is running
+		// If new messages are received while the job is running,
+		// allow the new job to be added and update it at the end if needed
+		$this->jobsService->delete($job->getId());
+
+		$account = $this->accountService->findById($job->getAccountId());
+		$mailbox = $this->mailManager->getMailbox($job->getUserId(), $job->getMailboxId());
+		$messageIds = $this->messageMapper->findAllIds($mailbox);
```
ChristophWurst: This operation is prone to run out of memory with large mailboxes

ChristophWurst: Can't you push the cursor and limit down to the DB and only fetch the data set you need?

lib/AppInfo/Application.php
```diff
@@ -89,6 +91,9 @@
 
 final class Application extends App implements IBootstrap {
 	public const APP_ID = 'mail';
+	public const CONTEXT_CHAT_MESSAGE_MAX_AGE = 60 * 60 * 24 * 365.25; // 1 year
+	public const CONTEXT_CHAT_IMPORT_MAX_ITEMS = 100;
+	public const CONTEXT_CHAT_JOB_INTERVAL = 60 * 5; // 5 minutes
```
ChristophWurst: Please put the context chat specific constant to a context chat class. Maybe the jobs service, or the provider?

lib/BackgroundJob/ContextChat/SubmitContentJob.php
ChristophWurst: To illustrate the problem: you are processing 100 items max. If a user has 100k emails in their mailbox there will be 1000 jobs. Each junk loads 100k rows to discard 99900.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,118 @@
...
+		$messageIds = array_filter($messageIds, fn (int $id): bool => $id >= $job->getNextMessageId());
+
+		if (count($messageIds) === 0) {
+			return;
+		}
+
+		$messages = $this->messageMapper->findByIds($job->getUserId(), $messageIds, 'asc');
+		// Ensure messages are sorted by ID
+		usort($messages, static fn (Message $a, Message $b): int => $a->getId() <=> $b->getId());
```
ChristophWurst: again, something the DB can do for you very efficiently because there is an index on ID

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,118 @@
...
+
+		$startTime = time() - Application::CONTEXT_CHAT_MESSAGE_MAX_AGE;
+		$nextMessage = reset($messages);
+		$client = $this->clientFactory->getClient($account);
```
ChristophWurst, line 112: Client connection is opened but never closed. Please wrap the following lines with a try-finally and close the connection in finally.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			$fullMessage = $imapMessage->getFullMessage($imapMessage->getUid(), true);
+
+			$items[] = new ContentItem(
+				(string)$nextMessage->getId(),
+				$this->contextChatProvider->getId(),
+				$imapMessage->getSubject(),
+				$fullMessage['body'] ?? '',
+				'E-Mail',
```
ChristophWurst: no l10n needed?

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,141 @@
...
+				foreach ($mailboxes as $mailbox) {
+					$messageIds = $this->messageMapper->findAllIds($mailbox);
+
+					if (count($messageIds) > 0) {
+						$this->jobsService->updateOrCreate($userId, $account->getId(), $mailbox->getId(), 0);
+					}
+				}
+			}
+		});
+	}
```
ChristophWurst: Has this been tested with moderate to large setups? I'm afraid we will run out of memory

You are loading all IDs of a mailbox without really using the result. Again, this has to be pushed to the db with a limit=1.

tests/Unit/ContextChat/ContextChatProviderTest.php
```diff
@@ -0,0 +1,183 @@
...
+			->method('linkToRouteAbsolute')
+			->with('mail.page.thread', ['mailboxId' => 1, 'id' => '2'])
+			->willReturn('http://localhost/apps/mail/box/1/thread/2');
+		$itemUrl = $this->contextChatProvider->getItemUrl('2');
+		$this->assertEquals($itemUrl, 'http://localhost/apps/mail/box/1/thread/2');
+	}
+
+	public function testTriggerInitialImport() {
+		// TODO: finish test
+		$this->markTestIncomplete();
```
ChristophWurst: Unfinished, skipped tests

lib/Migration/Version5200Date20250728000000.php
```diff
@@ -0,0 +1,59 @@
...
+				'notnull' => true,
+			]);
+			$table->addColumn('user_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 255,
+			]);
+			$table->addColumn('account_id', Types::INTEGER, [
+				'notnull' => true,
+				'length' => 4,
+			]);
```
ChristophWurst: This points to a foreign table. Make it a foreign key with CASCADE DELETE. Without that there will be dangling pointers after account deletion.

ChristophWurst (commented):
First round of review

There are some critical aspects to be fixed to make this production ready

lib/BackgroundJob/ContextChat/SubmitContentJob.php
marcelklehr: [Not needed](https://github.com/nextcloud/context_chat/blob/main/doc/Implementing_a_provider.md#the-content-item) indeed.

lib/ContextChat/ContextChatProvider.php
marcelklehr: Currently, you're passing the message Id only when submitting items to context chat. These will be used here to get the item url. If you make the mailbox ID part of the context chat item ID, it will be passed here and you can skip the db queries for the message and the mailbox.

edward-ly: Ah OK, since the content item ID is a string, it doesn't have to be a number and we can set it to anything we want? Something like`{mailboxId}:{messageId}`, I guess?

lib/BackgroundJob/ContextChat/SubmitContentJob.php
edward-ly: OK, so it sounds like the filter/sort operations are best done at the database level instead (through a new function in the `MessageMapper` class, perhaps?).

lib/ContextChat/ContextChatProvider.php
edward-ly: Ideally, we'd check if there is at least one message that satisfies all conditions for import (e.g. less than a year old), but for simplicity, we can check if there is at least one message in the mailbox, period. For that, I'm guessing we can use `MessageMapper::findHighestUid` and check for `null` rather than use `findAllIds`.

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,133 @@
...
+		return $this->urlGenerator->linkToRouteAbsolute('mail.page.thread', [ 'mailboxId' => $mailboxId, 'id' => $messageId ]);
+	}
+
+	/**
+	 * Starts the initial import of content items into context chat
+	 *
+	 * @return void
+	 * @since 5.2.0
+	 */
+	public function triggerInitialImport(): void {
```
ChristophWurst, line 121: Is this only called from https://github.com/nextcloud/context_chat/blob/f9cf8c773fa2dbe8dc2422fcc4b8aed34538df11/lib/BackgroundJobs/InitialContentImportJob.php#L66?

I'm worried about the complexity of the method and how long it will probably take to execute. Looping through all users is already slow, then you have n queries for accounts, and for each account there is a query for mailboxes. For an average setup with 50k users there will be 250k mailboxes loaded at least.

tests/Integration/Db/ContextChat/TaskMapperTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+	/** @var IDBConnection */
+	private $db;
+
+	private TaskMapper $mapper;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->db = \OC::$server->getDatabaseConnection();
```
ChristophWurst: nit: ocp\server::get(Idbconnection)? then no private API is needed

lib/ContextChat/ContextChatProvider.php
marcelklehr: > Is this only called from [nextcloud/context_chat@f9cf8c7/lib/BackgroundJobs/InitialContentImportJob.php#L66](https://github.com/nextcloud/context_chat/blob/f9cf8c773fa2dbe8dc2422fcc4b8aed34538df11/lib/BackgroundJobs/InitialContentImportJob.php#L66)?

Yes, this is guaranteed to be run in a background job

tests/Integration/Db/ContextChat/TaskMapperTest.php
marcelklehr: Good point!

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,133 @@
...
+use OCP\IURLGenerator;
+use OCP\IUser;
+use OCP\IUserManager;
+
+/**
+ * @implements IEventListener<Event>
+ */
+class ContextChatProvider implements IContentProvider, IEventListener {
+
+	public const CONTEXT_CHAT_MESSAGE_MAX_AGE = 60 * 60 * 24 * 365;
```
edward-ly, line 35: The .25 is to account for leap years.

Also, the comments should be added back for clarity.

lib/AppInfo/Application.php
```diff
@@ -91,9 +91,8 @@
 
 final class Application extends App implements IBootstrap {
 	public const APP_ID = 'mail';
-	public const CONTEXT_CHAT_MESSAGE_MAX_AGE = 60 * 60 * 24 * 365.25; // 1 year
```
edward-ly: The .25 is to account for leap years.

Also, the comments should be kept for clarity.

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -47,76 +55,112 @@ protected function run($argument): void {
...
 		}
 
 		if (count($items) > 0) {
 			$this->contentManager->submitContent($this->contextChatProvider->getAppId(), $items);
 		}
 
-		if ($nextMessage !== false) {
-			$this->jobsService->updateOrCreate($job->getUserId(), $job->getAccountId(), $job->getMailboxId(), $nextMessage->getId());
+		try {
+			$this->jobsService->updateOrCreate($task->getMailboxId(), $message?->getId() ?? $messageIds[0]);
```
edward-ly: Is there any situation where `$message` would not have an ID? And even then, why should we loop back to the start of the list?

lib/AppInfo/Application.php
marcelklehr: :+1:

lib/BackgroundJob/ContextChat/SubmitContentJob.php
marcelklehr: This is about `$message` being null and mainly to appease static analysis, I think. Feel free to improve this.

ChristophWurst:
@edward-ly is this ready for review?

edward-ly:
It looks like all unit tests have been added for the new methods now, but the codecov checks are still marking them as missing. I'm not sure if the integration tests is not being run or if there is actually something wrong with the checker.

The failing psalm check on stable31 can be ignored.

tests/Integration/Db/MessageMapperTest.php
```diff
@@ -30,19 +30,25 @@ class MessageMapperTest extends TestCase {
 	/** @var IDBConnection */
 	private $db;
 
+	/** @var ITimeFactory */
+	private $time;
+
+	private int $timestamp;
```
ChristophWurst: missing initialization

tests/Integration/Service/ContextChat/TaskServiceTest.php
```diff
@@ -0,0 +1,98 @@
...
+	public function testUpdateOrCreateNoUpdate(): void {
+		$this->service->updateOrCreate(2, 3);
+		$task = $this->service->findNext();
+		$this->assertEquals(2, $task->getMailboxId());
+		$this->assertEquals(3, $task->getLastMessageId());
+	}
+
+	public function testDelete(): void {
+		$this->insert(1, 1);
+		$this->insert(2, 2);
```
ChristophWurst: you insert mailbox_id=2, last_message_id=2 …

tests/Integration/Service/ContextChat/TaskServiceTest.php
```diff
@@ -0,0 +1,98 @@
...
+		$this->insert(3, 3);
+		$task = $this->service->delete(2);
```
ChristophWurst: … and here you delete primary id=2

that row does not exist most likely. I think what you want to do is use the mapper to insert the rows, then delete by the actual primary id.

tests/Integration/Db/MessageMapperTest.php
edward-ly: @marcelklehr does this need initialization or no?

ChristophWurst: One of the previous test results failed on this

edward-ly:
> @edward-ly could you please add instructions for how to test this? I set up context_chat and assistant. I ran cron many times. The initial import is not triggered.

Hm, the [docs](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/context_chat.html) say that "the triggerInitialImport method is called when Context Chat is first set up", but it's not clear what "set up" means, for sure. Maybe context_chat needs to be installed after the mail app so that it can find and trigger all existing providers?

@marcelklehr do you know the specifics around this?

ChristophWurst:
> @ChristophWurst Did you install context_chat_backend as well?

Bingo! That one is missing

marcelklehr:
mmh, this may be a bug in context_chat then :thinking: 

This method here should be called whenever an app interacts with context_chat: https://github.com/nextcloud/context_chat/blob/main/lib/Public/ContentManager.php#L95 and triggers an event for apps to register themselves.

ChristophWurst:
Looks like I'm getting closer

Bildschirmfoto vom 2025-11-05 08-58-05

lib/BackgroundJob/ContextChat/SubmitContentJob.php
```diff
@@ -0,0 +1,171 @@
...
+					[$account->getUserId()],
+				);
+			}
+		} catch (\Throwable $e) {
+			$this->logger->warning('Exception occurred when trying to fetch messages for context chat', ['exception' => $e]);
+		} finally {
+			try {
+				$client->close();
+			} catch (\Horde_Imap_Client_Exception $e) {
+				// pass
```
ChristophWurst: Please log at debug level at least

ChristophWurst:
I think it's still stuck at no docs, even though that error no longer bubbles up

```
$ php occ context_chat:stats
ContextChat statistics:
The indexing is not complete yet.
Total eligible files: 4
Files in indexing queue: 4
New files in indexing queue (without updates): 4
Queued documents (without files):array (
  'mail__mail' => 12,
)
Files successfully sent to backend: 0
Indexed documents: array (
)
Actions in queue: 0
File system events in queue: 0
```

I'm not sure how indexing is triggered so I execute the background jobs manually. Somehow this gets stuck for a very long time too

```
]$ php occ background-job:execute 10896
Job class:            OCA\ContextChat\BackgroundJobs\IndexerJob
Arguments:            {"storageId":1747,"rootId":38717}
Type:                 timed

Last checked:         2025-11-05T07:55:03+00:00
Reserved at:          2025-11-05T07:55:03+00:00
Last executed:        2025-11-05T07:55:03+00:00
Last duration:        0
Next execution:       2025-11-05T08:25:03+00:00

```

^ waiting 15 mins again.

kyteinsky:
> Indexed documents: array (
> )

weird, it doesn't show anything after indexing all the "mail__mail" docs.
context chat has a lot of logs so it uses a separate logging file.
are there any logs in the "data/context_chat.log*" files, maybe some errors about indexing or when the stats are fetched? Same for context_chat_backend, are there any error logs from when the mail docs were indexed? We can also check the db to see if the docs are actually there.
```bash
docker exec -it nc_app_context_chat_backend psql -U ccbuser -d ccb -p 5001
```
```sql
select * from docs;
\d
```

lib/Service/ContextChat/TaskService.php
```diff
@@ -0,0 +1,76 @@
...
+
+			return $this->taskMapper->insert($entity);
+		}
+
+		if ($lastMessageId >= $entity->getLastMessageId()) {
+			// Existing job already starts at an earlier message, so updating the database is not needed
+			return $entity;
+		}
+
+		$entity->setLastMessageId($lastMessageId);
```
ChristophWurst, line 69: Something appears to be wrong here. `last_message_id` is initialized with `0`. It will never be updated

Bildschirmfoto vom 2025-11-07 14-19-16
image

and that will lead to the same message being queued again and again

Bildschirmfoto vom 2025-11-07 14-17-03

ChristophWurst:
Sending context chat queries works too. Results don't

```json
{
  "reqId": "6Ulnm9L9uSvf4GrSUtG3",
  "level": 2,
  "time": "2025-11-13T08:29:34+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "no app in context",
  "method": "",
  "url": "--",
  "message": "A TaskProcessing context_chat:context_chat task with id 539 failed with the following message: Parameter \"id\" for route \"mail.page.thread\" must match \"[^/]++\" (\"\" given) to generate a corresponding URL.",
  "userAgent": "--",
  "version": "33.0.0.1",
  "data": []
}
```

- Requires https://github.com/nextcloud/context_chat_backend/pull/232 (released, update CCB)
- Requires https://github.com/nextcloud/context_chat/pull/190

---

@edward-ly here are the major todos:

1. Fix the infinite indexing loop.
2. Fix the route generation.
3. Add an opt-in switch. We don't want the context chat integration to be on by default for now. Admins have to turn this on explicitly.

src/components/AppSettingsMenu.vue
```diff
@@ -350,10 +360,11 @@ export default {
 			loadingOptOutSettings: false,
 			loadingInternalAddresses: false,
 			loadingReplySettings: false,
-
+			contextChatText: t('mail', 'Make mails available to Context Chat'),
```
edward-ly, line 370: `emails` would be clearer, maybe?

lib/BackgroundJob/ContextChat/ScheduleJob.php
```diff
@@ -0,0 +1,81 @@
...
+		private AccountService $accountService,
+		private MailManager $mailManager,
+		private LoggerInterface $logger,
+		private IJobList $jobList,
+		private ContextChatSettingsService $contextChatSettingsService,
+		private IContentManager $contentManager,
+	) {
+		parent::__construct($time);
+
+		$this->setInterval(60 * 60 * 24);
```
edward-ly, line 40:
```suggestion
		$this->setInterval(60 * 60 * 24); // 24 hours
```

.github/workflows/psalm-matrix.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        ocp-version: [ 'dev-stable31', 'dev-stable32', 'dev-master' ]
+        ocp-version: [ 'dev-stable32', 'dev-master' ]
```
ChristophWurst: this app still works with 31. leave 31 in.

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         php-versions: ['8.2', '8.3']
-        nextcloud-versions: ['master', 'stable31']
+        nextcloud-versions: ['master']
```
ChristophWurst, line 18: same

.github/workflows/psalm-matrix.yml
marcelklehr:
```suggestion
        ocp-version: [ 'dev-stable31', 'dev-stable32', 'dev-master' ]
```

.github/workflows/test.yml
marcelklehr, line 18:
```suggestion
        nextcloud-versions: ['stable31', 'stable32', 'master']
```

DaphneMuller:
@ChristophWurst I think we're done with your to do list from 18 November (https://github.com/nextcloud/mail/pull/11150#issuecomment-3526445875)

marcelklehr:
Actually, this PR does not fully work with stable31 (see psalm issues), should we adjust it to work with stable31?

PR #7179, opened 2022-09-06
Fix element alignment of thread message

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/7175
Fixes https://github.com/nextcloud/mail/issues/7160

Bildschirmfoto vom 2022-09-06 14-51-51

Bildschirmfoto vom 2022-09-06 14-52-07

The only thing I didn't manage to do is align with the contents of a HTML body. The styling doesn't appear consistent across all email programs/browsers.

PR #11164, opened 2025-05-16
fix: html message alignment

hamza221:
| b | a |
|--------|--------|
|image | image|


ChristophWurst:
Please see https://github.com/nextcloud/mail/pull/7179. The text indentation was fine. What needs better alignment is the phishing detection. Maybe check the external images warning as well while you move pixels around.

ChristophWurst (approved):

Issue #11185, opened 2025-05-21
Nexcloud Mail app still marks most incoming messages as important in Thunderbird email client, even when set (locally and globally) not to do that


dean-ollis:
### Steps to reproduce

1. Under Nextcloud Admin Settings, ensure that "Enable classification of important mails by default" (under Groupware - Enable classification by importance by default) is toggled to the off position
2. In the Mail settings of the Nextcloud Mail app, under "Assistance features", ensure that "Mark as important" is unchecked.  
3. Open Mozilla Thunderbird email client and check email for address linked to Nextcloud account

### Expected behavior

No incoming emails should be automatically tagged as Important

### Actual behavior

Most incoming emails are tagged as Important

### Mail app version

5.1.0

### Nextcloud version

30.0.10 (Nextcloud Hub 9)

### Operating system

Linux

### PHP engine version

None

### Web server

Other

### Database

None

### Additional info

This issue is driving me nuts to the point where I am thinking of cancelling my Nextcloud instance with Hetzner! It started without the Nexcloud Mail app even being installed (but with the global setting toggled to not allow automatic tagging of emails as important), so I installed / enabled the Mail app to change the relevant setting within the app but it made no difference. From what I have read, this issue seems to affect and cause a lot of frustration for many Nextcloud users.

ChristophWurst:
Thank you for your report. I'm having a look.

ChristophWurst:
> * Under Nextcloud Admin Settings, ensure that "Enable classification of important mails by default" (under Groupware - Enable classification by importance by default) is toggled to the off position
> 
> * In the Mail settings of the Nextcloud Mail app, under "Assistance features", ensure that "Mark as important" is unchecked.

This is not reproducible. The user preference is always followed.

What I could find is that the global setting is not always honored. Fix is at https://github.com/nextcloud/mail/pull/11187.

dean-ollis:
> It's a pull request. If you are unfamiliar with this I suggest you wait for the changes to be released. Thanks.

Now that this bug report has been closed, do you know how long it should take until I can expect to see a positive change with regard to this issue on my side? Sorry for my ignorance in not understanding how these fixes work!

dean-ollis:
> The changes will be released soon.

The problem is still there. Have the changes been made yet?

st3iny:
@dean-ollis Not yet unfortunately. The backport was missing. I triggered it again and the next release of `5.1.x` should contain the fix.

[...]

PR #11187, opened 2025-05-22
fix(classification): use global default for new messages too

ChristophWurst:
Resolves https://github.com/nextcloud/mail/issues/11185.

Leftover of https://github.com/nextcloud/mail/pull/9609


kesselb (approved):
:elephant:

PR #11191, opened 2025-05-22
fix: close modal when send is clicked

GretaD:


src/components/NewMessageModal.vue
```diff
@@ -387,9 +387,13 @@ export default {
 		async onSend(data, force = false) {
 			logger.debug('sending message', { data })
 
+			if (this.sending) {
+				return
+			}
 			await this.attachmentsPromise
 			this.uploadingAttachments = false
 			this.sending = true
+			this.$emit('close')
```
ChristophWurst, line 396: L467 emits too. Is that obsolete?

PR #11195, opened 2025-05-23
fix(smime): only accept certificates and keys in file input

hamza221:
Make sure the input only accepts the required files for easier file selection


hamza221:
@st3iny  Did I add all possible extensions ?

ChristophWurst:
I agree that `fix` works here too. You're not adding a new feature

ChristophWurst (approved):

st3iny (dismissed):

PR #11238, opened 2025-06-04
refactor: mark classes final

ChristophWurst:
99% sure tests will explode without https://packagist.org/packages/dg/bypass-finals

Extracted from #11224


lib/Controller/AutoCompleteController.php
```diff
@@ -17,7 +17,7 @@
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\IRequest;
 
-#[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
+final #[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
```
ChristophWurst, line 20: :facepalm:

PR #11252, opened 2025-06-06
fix(outbox): outbox do not render

kesselb:
Follow-up for https://github.com/nextcloud/mail/pull/10486

- data is undefined
- fetchAvatarFromClient is (currently) available for messages, but not local messages


hamza221 (approved):
mb

PR #11272, opened 2025-06-15
Update multiselect to latest design review

GVodyanov:
Adjust to https://github.com/nextcloud/contacts/pull/4477#issuecomment-2973697641

| Before | After |
| - | - |
| image | image |

The animation to show this is now faster 

 image


kesselb:
Truth be told, before and after look the same to me, but lgtm :laughing:

ChristophWurst:
> Truth be told, before and after look the same to me, but lgtm 😆

it's the checkmark in the avatar :)

ChristophWurst (approved):

PR #11338, opened 2025-07-01
fix(l10n): remove trailing space and allow RTL for drafts

ChristophWurst:


package.json
```diff
@@ -66,6 +66,7 @@
     "core-js": "^3.43.0",
     "debounce-promise": "^3.1.2",
     "dompurify": "^3.2.6",
+    "escape-html": "^1.0.3",
```
kesselb, line 69: What about using dompurify?

kesselb (approved):

PR #11347, opened 2025-07-02
fix: quota spacing

GretaD:
The problem:
`NcActionText` has a [br](https://github.com/nextcloud-libraries/nextcloud-vue/blob/31d28c228f6d8dbc09d875f8053e50d49824ec84/src/components/NcActionText/NcActionText.vue#L25) to split action name from action_long text, which breaks our quota and quota text on our action menu as seen below on the screenshot. 
NcActionButton has a section for long text that doesnt break: https://github.com/nextcloud-libraries/nextcloud-vue/blob/ae1b3c2ef908985a4052ac03d9a6310593030545/src/components/NcActionButton/NcActionButton.vue#L346-L364

so i used that to fix our problem. Another solution would be if we remove the `br` and add description prop, but im not sure if that works on every scenario.

before
Screenshot from 2025-07-02 12-35-21
after
Screenshot from 2025-07-02 12-35-41

### DIFF
--- a/src/components/NavigationAccount.vue
+++ b/src/components/NavigationAccount.vue
@@ -21,12 +21,12 @@
 					</ActionText>
 				</template>
 				<template v-else>
-					<ActionText v-if="!account.isUnified" :name="t('mail', 'Quota')">
+					<ActionButton v-if="!account.isUnified" :name="t('mail', 'Quota')">
 						<template #icon>
 							<IconInfo :size="16" />
 						</template>
 						{{ quotaText }}
-					</ActionText>
+					</ActionButton>
 					<ActionButton :close-after-click="true"
 						@click="showAccountSettings(true)">
 						<template #icon>


ChristophWurst (commented):
Works visually until you hover, because buttons have a pointer cursor. This wrongly makes the element appear as if it were clickable.

Can you share more insights about the impact of the br in other usages of the text action? Is the additional space desired everywhere but Mail, or should we fix all usages vi an upstream fix?

Also, the fix only covers the accounts actions. Mailbox actions still have the additional space. Please fix both in one go.

GretaD:
> Can you share more insights about the impact of the br in other usages of the text action? Is the additional space desired everywhere but Mail, or should we fix all usages vi an upstream fix?

I didnt find any other usage of having a name with a long text, but i feel like im missing something because there is this comment on the ncactiontext component: `<!-- white space is shown on longtext, so we can't put {{ text }} on a new line for code readability -->`
So, wondering why it was there in the first place, and what can it break if we remove it? So, while wondering, i saw the ncactionbutton change and went with that :).

As i noted on in the ticket, we can also add a desc prop and see how that fits our scenario.

ChristophWurst:
> As i noted on in the ticket, we can also add a desc prop and see how that fits our scenario.

There is no such prop on https://nextcloud-vue-components.netlify.app/#/Components/NcActions?id=ncactiontext

GretaD:
I guess now this is obvious that needs to be changed upstream. So i will do that

ChristophWurst:
Yep sounds like the best approach if there is not a single use where the current style looks good :)

Issue #11348, opened 2025-07-02
specify which mailbox to load more from


hamza221:
"Load more" --> "Show more follow ups" "Show more important messages"

PR #11350, opened 2025-07-02
fix: specify which mailbox to load more from

hamza221:
fix [#11348](https://github.com/nextcloud/mail/issues/11348)

| b | a |
|--------|--------|
| image |image |


src/components/EnvelopeList.vue
```diff
@@ -128,7 +128,7 @@
 				class="load-more"
 				@click="$emit('load-more')">
 				<AddIcon :size="16" />
-				{{ t('mail', 'Load more') }}
+				{{ name ? t('mail', 'Load more {name}', { name }): t('mail', 'Load more') }}
```
ChristophWurst: I'm doubting that concatenating the two parts will work universally for all languages, and think that the full UI string should be translated.

ChristophWurst: https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#step-1-string-split confirms that

src/components/Mailbox.vue
```diff
@@ -61,6 +62,10 @@ export default {
 	},
 	mixins: [isMobile],
 	props: {
+		name: {
```
ChristophWurst: make it a `loadMoreLabel` or similar

src/components/EnvelopeList.vue
hamza221: I'm not sure how to fix this, If I don't translate separately, I don't see another way of translating them . 
Maybe just add comment for translators on both ends with all possible values for the placeholder ?

hamza221: or nvm, maybe I can pre- generate the whole thing as a computed conditional value

src/components/MailboxThread.vue
```diff
@@ -53,6 +53,7 @@
 							</Popover>
 						</div>
 						<Mailbox v-show="hasFollowUpEnvelopes"
+							:load-more-label="t('mail', 'follow ups')"
```
ChristophWurst: here you translate …

src/components/EnvelopeList.vue
```diff
@@ -237,6 +241,19 @@ export default {
 		sortOrder() {
 			return this.mainStore.getPreference('sort-order', 'newest')
 		},
+		// This is intential to get proper translation
+		loadMoreText() {
+			switch (this.loadMoreLabel) {
+			case 'follow ups':
```
ChristophWurst: … and here you compare with English. So this will only work for lang=en :wink:

src/components/MailboxThread.vue
```diff
@@ -81,6 +82,7 @@
 						</div>
 						<Mailbox v-show="hasImportantEnvelopes"
 							class="nameimportant"
+							:load-more-label="t('mail', 'important messages')"
```
ChristophWurst: how about passing the full string here and use it as is in the list?

src/components/EnvelopeList.vue
hamza221: 🙈

src/components/EnvelopeList.vue
```diff
@@ -196,6 +196,10 @@ export default {
 			type: Object,
 			required: true,
 		},
+		loadMoreLabel: {
+			type: String,
+			default: '',
```
ChristophWurst: nit: fallback to `t('mail', 'Load more')`?

src/components/Mailbox.vue
```diff
@@ -61,6 +62,10 @@ export default {
 	},
 	mixins: [isMobile],
 	props: {
+		loadMoreLabel: {
+			type: String,
+			default: '',
```
ChristophWurst: nit: default undefined so the EnvelopeList default is used when nothing is passed?

ChristophWurst (approved):

ChristophWurst (approved):
Still good :)

Issue #11329, opened 2025-06-30
Unable to attach shared files to an email


chung03:
### Steps to reproduce

1. Have one user (A) share a folder and all files within to another user (B). Ensure that user B has all permissions, including download.
2. Login as user B and go to the mail app
3. Make a new email and go to "Add attachment from Files"
4. Check the "All files" list

### Expected behavior

The folder and files which were shared to user B will be visible and the user can select a file within the shared folder.

### Actual behavior

The shared folder and the files within are not options which appear for user B. So user B cannot attach any files within the shared folder to an email.

### Mail app version

5.1.5

### Nextcloud version

31.0.5

### PHP engine version

None

### Web server

Apache (supported)

### Database

None

chung03:
> Thank you for your report
> 
> I'm not able to reproduce with 31.0.6.2

Hello Christoph,

Thanks for your reply. I tried updating my Nextcloud Server version to 31.0.6 but I still have the same issue as before.

I can see the ticket is now tagged "needs info". What kind of info do you need?
What kind of logs or other info would be helpful? Are there any specific tests or procedures you'd like me to do to get more info?

Also, I want to rule out a bad configuration on my end, though I haven't found much info when I google search. What kind configurations could have possibly resulted in attachments from a shared folder being disabled in the mail app?

ChristophWurst:
I tried again, this time on my development environment and it is reproducible. It happens with direct user shares and with group shares.
The PROPFIND to list the root directory (where the share is located) does contain the share, so it must be an issue in the frontend filtering.
The file picker checks for download permission via the share attributes. The current check takes the download permission and it's `enabled` property. But there is only a `value: true`.

I'll check if there was a change in the files API.

ChristophWurst:
This is a regression of the changes from https://github.com/nextcloud/server/commit/9b84831c8d0c1d715cd42fa3e4e01ac8a59fa369

ChristophWurst:
https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

PR #11352, opened 2025-07-02
fix(composer): allow attaching shared files again

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/11329

Breaking change of https://github.com/nextcloud/server/pull/46007 that was not addressed in this app.

> IAttributes enabled key have bee renamed to value and supports more than boolean.

https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1


kesselb (approved):
🐘

Issue #11353, opened 2025-07-02
AI responses and reply button


hamza221:
- [x] Reply button right aligned with the 3 dot of the attachment 
  - [x] Spacing below the ai responses and reply button should be the same as the right spacing of the reply button

PR #11354, opened 2025-07-02
fix: reply buttons spacing

hamza221:
fix [#11353](https://github.com/nextcloud/mail/issues/11353)

| b | a |
|--------|--------|
|image |image |


ChristophWurst:
Bottom right still looks a bit odd because the white body and the button do not align … maybe the body needs the same amount of space as the button on the right? :thinking:

PR #11356, opened 2025-07-02
fix: thread heading alignment and participants

### LINKED ISSUE #11353: AI responses and reply button

### LINKED ISSUE #11355: Thread heading
- [x] Alignment of text is off: align with text of the message 
  - [ ] Don't show own email address in the thread header

hamza221:
fix #11355
fix #11353
| b | plain/text | html |
|--------|--------|--------|
|image | image| image |

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -106,7 +106,7 @@
 			</div>
 		</template>
 
-		<Thread v-if="showThread" @delete="deleteMessage" />
+		<Thread v-if="showThread" :current-account-email="account.emailAddress" @delete="deleteMessage" />
 		<NoMessageSelected v-else-if="hasEnvelopes" />
 	</AppContent>
 </template>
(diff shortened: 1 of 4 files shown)


ChristophWurst:
Edge case: emails to yourself now show some "void" underneath the thread heading: 
image

I'm undecided if we should add special handling or make the header height responsive and smaller in that case

ChristophWurst:
Bildschirmfoto vom 2025-07-02 18-37-36
Bildschirmfoto vom 2025-07-02 18-37-14
image

I think the message header (sender + email) should be our reference. Thread header has to move left a few px. Plain text emails as well. The white box of HTML emails should start at the same virtual line.

hamza221:
> Edge case: emails to yourself now show some "void" underneath the thread heading: image
> 
> I'm undecided if we should add special handling or make the header height responsive and smaller in that case

I faced the same edge case, @nimishavijay  what do you think ?

GretaD:
@hamza221 should i take over this? I have the time. I can also take over the other ones you have started as part of the desgn review we had :)

Issue #11358, opened 2025-07-02
"Load more" button


hamza221:
- [ ] Text should be left aligned with the text of the thread

PR #11359, opened 2025-07-02
fix: load-more button alignment

hamza221:
fix [#11358](https://github.com/nextcloud/mail/issues/11358)
| b | a |
|--------|--------|
|image |image |


ChristophWurst:
In my interpretation of the design feedback the text of the thread (sender, subject, preview) should be aligned with the text of *Load more*, not the icon. @nimishavijay could you please clarify?

ChristophWurst (approved):

PR #11389, opened 2025-07-15
fix: wrong text colour

GretaD:
Screenshot from 2025-07-15 18-03-04


ChristophWurst (commented):
Please make the commit message more specific to have a meaningful changelog entry

ChristophWurst (approved):

Issue #11065, opened 2025-05-05
Show attachments in thread list


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud Mail user I want to quickly open the (single) attachment of an email. E.g. see the attached PDF for an invoice.

### Describe the solution you'd like

Show the attachment name in the thread list and make it clickable.

### Describe alternatives you've considered

N/a

### Additional context

N/a

nimishavijay:
Design of each attachment:

![Image](https://github.com/user-attachments/assets/e00b52af-b6d1-4dd8-bb6e-55131177a52c)

- [ ] 24px height
- [ ] border: 1px solid var(--color-border-dark)
- [ ] border-radius: var(--border-radius-element) (8px)
- [ ] inside we show a little 16px thumbnail with border-radius: 4px and the name of the attachment with a 4px gap

For the position there are 2 ways in which this can work out.

First one requires bit more rearranging, but has more parity with Outlook, nothing moves around when tags are added, and more attachments can be shown:

![Image](https://github.com/user-attachments/assets/e6f17eaa-c3cf-4de3-8841-1f43c83d3db7)

- [ ] Move the tags row into the second row of the body, resulting in the body ellipsizing earlier if tags are added
- [ ] The attachments are shown in the row below, where the tags currently are

Second one is less effort, but still needs some improvements to the way tags are shown:

![Image](https://github.com/user-attachments/assets/c534cc8f-7769-4240-bf4b-dd066abcaabb)

- [ ] Just add the attachments to the same row as the tags
- [ ] Improve the design of the tags:
  - [ ] text font-size: 13px (instead of the current font-size*0.8)
  - [ ] height of each tag should be 24px
  - [ ] add an 8px margin above the tags

Either way, we can make some changes to the thread list to make it more spaced out and nicer to look at, here are the things I tried just from the inspector:

- [ ] Date time font size reduced to 13px
- [ ] Body font size reduced to 13px
- [ ] Padding around list item increased to 8px (In the screenshots I did `.list-item { padding: calc(var(default-grid-baseline) * 2) } `)
- [ ] Padding on the left of the sender/subject/body increased to 16px (I did  `.list-item-content { padding-inline-start: calc(var(default-grid-baseline) * 4) } `)
- [ ] The width of the body should not change as we hover. I added a width:100% to .envelope__preview-text as this seemed to work (?)

What do you think? Would either of these work?

jancborchardt:
[...]

PR #11400, opened 2025-07-17
feat: make thread elements have the same space around them

### LINKED ISSUE #11316: Improve element spacing, grouping and text densitiy

As a Nextcloud Mail user I want to be able to visually distinguish groups of contents, and see the app make use of available space.

_No response_

_No response_

### Additional context

Needs a design review

GretaD:
fixes #11316 

Only worked on these points as added by Jan on the ticket:

- The mail heading & participants should be aligned to the content box.
- and the "Reply" button as well as the action buttons on the top right
- When scrolling the content, the container / top divider line vanishes and gets overlapped partly by the heading, making it look strange
- Now with the message list having more content, horizontal divider lines could make sense to have some nicer separation.
- Mails have different heights in the message list if there’s shorter ones – but all should have the same consistent height
- Whitespace is sometimes not the same across the board, for example the space left and right in the message list is much less (half?) than in the left navigation.

before
Screenshot from 2025-07-24 18-34-16
after

Screenshot from 2025-07-24 18-30-27

after(reply button aligned with the action menu)
Screenshot from 2025-07-24 18-29-59

Subject and recipient aligned with the recipient in the message header. plus added the line to split subject from the message
Screenshot from 2025-08-11 13-12-19

more space for the preview, but this is a bit tricky to add more space for the preview because the details, the unread and the action menu wouldnt allow it to go further.
before
Screenshot from 2025-07-24 18-37-18

after
Screenshot from 2025-07-24 17-27-46

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -579,17 +579,27 @@ export default {
 	position: -webkit-sticky; // ios/safari fallback
 	position: sticky;
 	top: 0;
-	background: -webkit-linear-gradient(var(--color-main-background), var(--color-main-background) 80%, rgba(255,255,255,0));
-	background: -o-linear-gradient(var(--color-main-background), var(--color-main-background)  80%, rgba(255,255,255,0));
-	background: -moz-linear-gradient(var(--color-main-background), var(--color-main-background)  80%, rgba(255,255,255,0));
-	background: linear-gradient(var(--color-main-background), var(--color-main-background)  80%, rgba(255,255,255,0));
+	margin-bottom: 5px;
+
+	&::before {
+		content: '';
+		position: absolute;
+		top: 0;
+		inset-inline-start: 50%;
+		transform: translateX(-50%);
+		width: 100vw;
+		height: 100%;
+		background: var(--color-main-background);
+		border-bottom: var(--border-width-input-focused) solid var(--color-border);
+		z-index: -1;
+	}
 }
 
 #mail-thread-header-fields {
 	// initial width
 	width: 0;
 	// while scrolling, the back button overlaps with subject on small screen
-	padding-inline-start: calc(var(--default-grid-baseline) * 21);
+	padding-inline-start: calc(var(--border-radius-container-large) + var(--header-height));
 	// grow and try to fill 100%
 	flex: 1 1 auto;
 	h2,
@@ -605,7 +615,6 @@ export default {
 		overflow: hidden;
 		text-overflow: ellipsis;
 	}
-
 	.transparency {
 		opacity: 0.6;
 		a {
@@ -669,7 +678,7 @@ export default {
 	display: flex;
 	align-items: stretch;
 
-	::deep(.v-popper--theme-dropdown.v-popper__popper .v-popper__inner) {
+	:deep(.v-popper--theme-dropdown.v-popper__popper .v-popper__inner) {
 		height: 300px;
 		width: 250px;
 		overflow: auto;
--- a/src/components/EnvelopeSkeleton.vue
+++ b/src/components/EnvelopeSkeleton.vue
@@ -429,7 +429,7 @@ export default {
 	--list-item-padding: calc(var(--default-grid-baseline) * 2);
 	// The content are two lines of text and respect the 1.5 line height
 	--list-item-border-radius: var(--border-radius-element, 32px);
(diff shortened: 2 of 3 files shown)


src/components/Thread.vue
```diff
@@ -579,17 +579,20 @@ export default {
...
+	margin-bottom: 5px;
+	box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
 }
 
 #mail-thread-header-fields {
 	// initial width
 	width: 0;
 	// while scrolling, the back button overlaps with subject on small screen
-	padding-inline-start: calc(var(--default-grid-baseline) * 21);
+	padding-inline-start: 68px;
```
ChristophWurst: nit: magic number. can it be based on the grid baseline?

src/components/Thread.vue
```diff
@@ -579,17 +579,20 @@ export default {
...
 	top: 0;
-	background: -webkit-linear-gradient(var(--color-main-background), var(--color-main-background) 80%, rgba(255,255,255,0));
-	background: -o-linear-gradient(var(--color-main-background), var(--color-main-background)  80%, rgba(255,255,255,0));
-	background: -moz-linear-gradient(var(--color-main-background), var(--color-main-background)  80%, rgba(255,255,255,0));
-	background: linear-gradient(var(--color-main-background), var(--color-main-background)  80%, rgba(255,255,255,0));
+	background: var(--color-main-background);
+	background: var(--color-main-background);
+	background: var(--color-main-background);
+	background: var(--color-main-background);
+	border-bottom: 2px solid var(--color-border);
```
ChristophWurst: nit: is there no var for border width?

ChristophWurst: the border is limited to the width of the thread header and messages. shouldn't it be as wide as the third column?

GretaD: I can double check with design, but i dont think they said specifically it should be all way. Also, it looks nicer this way.

But anyhow, its limited to the width of the header because of the component, N matter where i add that border, including as a separate div between sections, it always has the same width. I even tried to force width, and nothing happened. Im sure i can find a hacky way to do it, but i dont think we should :)

ChristophWurst: Check the header border of Talk. It's also taking the full width while the messages are limited.

ChristophWurst: image

It should be possible without hacks. Here's a quick poc. You have to move the min/max width and centering from outer element to the inner elements.

GretaD: ah, thank you. I still think it looks chaotic and busy. Talk doesnt have message borders and their other borders are 1px. 

Our borders are 2px.

ChristophWurst: Then let's go back to design. @jancborchardt @nimishavijay please decide.

nimishavijay: Since it is separating the thread header from the rest like in Talk, it can be all the way. Would it be possible to only show it if we have scrolled a bit? don't show it if we are all the way at the top 

Also then would be good to left align the contents of the header with the text of each message. In the original screenshots of Greta there also seems to be a shadow? We don't need to have a shadow here, just 2px border in --color-border would be enough

GretaD:
> Nice improvements :) I am wondering if the changes in the last part of this comment can be incorporated as well? Could be as a separate PR too [#11065 (comment)](https://github.com/nextcloud/mail/issues/11065#issuecomment-2990471449) also cc @hamza221 as you are assigned to that issue

they will be done for sure, but in a separate pr :)

src/components/Thread.vue
```diff
@@ -579,17 +579,27 @@ export default {
...
+		position: absolute;
+		top: 0;
+		left: 50%;
+		transform: translateX(-50%);
+		width: 100vw;
+		height: 100%;
+		background: var(--color-main-background);
+		border-bottom: var(--border-width-input-focused) solid var(--color-border);
+		z-index: -1;
+	}
```
ChristophWurst: IMO adjusting the width of the elements would have been preferable over this additional element

ChristophWurst (approved):

src/components/Thread.vue
GretaD: i went for this because the border would lose the sticky position if i would do it like that. Either im doing something wrong with your proposal, or i dont know.

PR #11411, opened 2025-07-23
fix(imip): process imip messages more frequently

kesselb:
When a message is flagged as imip, it takes up to one hour until we try to process it. Given that the job processes every message just once, it seems reasonable to run it more often.

### DIFF
--- a/lib/Service/IMipService.php
+++ b/lib/Service/IMipService.php
@@ -47,7 +47,7 @@ class IMipService {
 	public function process(): void {
 		$messages = $this->messageMapper->findIMipMessagesAscending();
 		if ($messages === []) {
-			$this->logger->info('No iMIP messages to process.');
+			$this->logger->debug('No iMIP messages to process.');
 			return;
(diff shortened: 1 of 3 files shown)


ChristophWurst (approved):
Fine by me

lib/Service/IMipService.php
```diff
@@ -47,7 +47,7 @@ public function __construct(
 	public function process(): void {
 		$messages = $this->messageMapper->findIMipMessagesAscending();
 		if ($messages === []) {
-			$this->logger->info('No iMIP messages to process.');
+			$this->logger->debug('No iMIP messages to process.');
```
ChristophWurst, line 50: > Expectation failed for method name is "info" when invoked 1 time(s).
> Method was expected to be called 1 times, actually called 0 times.

there is a test for this 🫣

kesselb: :see_no_evil:

PR #11419, opened 2025-07-25
Fix show/collapse folders is not saved

kesselb:
STR:

- Open mail
- Click show all folders
- See a failing xhr request

I think it's good to go. I'm considering a follow-up to add a toast when storing the settings faild, it's currently very hard to spot.


ChristophWurst (approved):

ChristophWurst:
https://github.com/nextcloud/mail/pull/11381 was not backported. I'm doing a partial backport

PR #11425, opened 2025-07-28
Feat: add OCS routes for mailbox and mail listing

janepie:
adds OCS routes to
- get the list of mailboxes for an account
- get the list of mails in a mailbox


lib/Controller/MailboxesApiController.php
```diff
@@ -0,0 +1,120 @@
...
+			$account,
+			$mailbox,
+			$sort,
+			$filter === '' ? null : $filter,
+			$cursor,
+			$limit,
+			$userId,
+			$view
+		);
+		return new DataResponse(json_encode($messages), Http::STATUS_OK);
```
marcelklehr: json_encode should not be necessary when using DataResponse, I think 🤔

janepie: You're right, removed :)

lib/Controller/MailboxesApiController.php
```diff
@@ -0,0 +1,120 @@
...
+	/**
+	 * List all mailboxes of an account of the user which is currently logged-in
+	 *
+	 * @param int $accountId the mail account id
+	 * @return DataResponse<Http::STATUS_OK, array<string, mixed>, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array{}, array{}>
+	 *
+	 * 200: Mailbox list
+	 * 404: User was not logged in or account doesn't exist
+	 */
+	#[ApiRoute(verb: 'GET', url: '/mailbox/list')]
```
ChristophWurst: Nitpick: not happy about the URL design. `/api/mailboxes` is already taken by the internal API. How about `/api/ocs/mailboxes` or similar to have a new, dedicated sub path for public ocs APIs?

janepie: The `api/` part is left out for the OCS routes for now, but I'm open to anything here. Could put ocs instead of api, so `/ocs/mailboxes/list`?

ChristophWurst: I'm fine with leaving out the api part for ocs. But please make it `ocs/mailboxes` without the list

janepie: Sure, done!

lib/Controller/MailboxesApiController.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @param int|null $limit the number of messages to be returned, can be left ampty to get all messages
+	 * @param string $view returns messages in requested view ('singleton' or 'threaded')
+	 * @param string|null $v Cache buster version to guarantee unique urls (will trigger HTTP caching if set)
+	 * @return DataResponse<Http::STATUS_OK, array<string, mixed>, array{}>|DataResponse<Http::STATUS_NOT_FOUND, array{}, array{}>|DataResponse<Http::STATUS_FORBIDDEN, array{}, array{}>
+	 *
+	 * 200: Message list
+	 * 403: User cannot access this mailbox
+	 * 404: User was not logged in
+	 */
+	#[ApiRoute(verb: 'GET', url: 'ocs/mailboxes/messages')]
```
ChristophWurst:
```suggestion
	#[ApiRoute(verb: 'GET', url: 'ocs/mailboxes/{mailboxId}/messages')]
```

sorry, missed that one before. let's encode the mailbox ID into the URL

ChristophWurst (commented):
:+1: otherwise :)

PR #11440, opened 2025-08-01
feat(ai-integration): detect if message needs translation

hamza221:
| HTML | PLAIN/TEXT |
|--------|--------|
| image | image |

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -913,6 +913,50 @@ class MessagesController extends Controller {
 	}
 
 	/**
+	 * @NoAdminRequired
+	 *
+	 * @param int $messageId
+	 *
+	 * @return JSONResponse
+	 */
+	#[TrapError]
+	public function needsTranslation(int $messageId): JSONResponse {
+		if ($this->currentUserId === null) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $messageId);
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		if (!$this->aiIntegrationService->isLlmProcessingEnabled()) {
+			$response = new JSONResponse([], Http::STATUS_NOT_IMPLEMENTED);
+			$response->cacheFor(60 * 60 * 24, false, true);
+			return $response;
+		}
+
+		try {
+			$requiresTranslation = $this->aiIntegrationService->requiresTranslation(
+				$account,
+				$mailbox,
+				$message,
+				$this->currentUserId
+			);
+			$response = new JSONResponse(['requiresTranslation' => $requiresTranslation === true]);
+			$response->cacheFor(60 * 60 * 24, false, true);
+			return $response;
+		} catch (ServiceException $e) {
+			$this->logger->error('Translation check failed: ' . $e->getMessage(), [
+				'exception' => $e,
+			]);
+			return new JSONResponse([], Http::STATUS_NO_CONTENT);
+		}
+	}
+
+	/**
 	 * @param int $id
 	 * @param array $attachment
 	 *
--- a/lib/Service/AiIntegrations/AiIntegrationsService.php
+++ b/lib/Service/AiIntegrations/AiIntegrationsService.php
@@ -20,6 +20,7 @@ use OCA\Mail\IMAP\IMAPClientFactory;
 use OCA\Mail\Model\EventData;
 use OCA\Mail\Model\IMAPMessage;
 use OCP\IConfig;
+use OCP\IL10N;
 use OCP\TaskProcessing\IManager as TaskProcessingManager;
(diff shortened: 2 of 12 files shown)


lib/Controller/MessagesController.php
```diff
@@ -912,6 +912,32 @@ public function smartReply(int $messageId):JSONResponse {
 
 	}
 
+	public function needsTranslation(int $messageId): JSONResponse {
```
ChristophWurst, line 923: * this only works for admins
* no error trap

lib/Controller/MessagesController.php
```diff
@@ -912,6 +912,32 @@ public function smartReply(int $messageId):JSONResponse {
...
+		try {
+			$message = $this->mailManager->getMessage($this->currentUserId, $messageId);
```
ChristophWurst, line 928: null handling missing to make Psalm happy

lib/Controller/MessagesController.php
```diff
@@ -912,6 +912,32 @@ public function smartReply(int $messageId):JSONResponse {
...
+			$mailbox = $this->mailManager->getMailbox($this->currentUserId, $message->getMailboxId());
+			$account = $this->accountService->find($this->currentUserId, $mailbox->getAccountId());
+		} catch (DoesNotExistException $e) {
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		if (!$this->aiIntegrationService->isLlmProcessingEnabled()) {
+			return new JSONResponse([], Http::STATUS_NOT_IMPLEMENTED);
```
ChristophWurst: cache for a day to prevent another HTTP request when the message is revisited

lib/Controller/MessagesController.php
```diff
@@ -912,6 +912,32 @@ public function smartReply(int $messageId):JSONResponse {
...
+		}
+
+		try {
+			$requiresTranslation = $this->aiIntegrationService->requiresTranslation(
+				$account,
+				$mailbox,
+				$message,
+				$this->currentUserId
+			);
+			return new JSONResponse(['requiresTranslation' => $requiresTranslation]);
```
ChristophWurst: cache the response for a week. the result won't change. going back to the same message should not even cause another HTTP request

lib/Controller/MessagesController.php
```diff
@@ -912,6 +912,32 @@ public function smartReply(int $messageId):JSONResponse {
...
+		} catch (ServiceException $e) {
+			return new JSONResponse([], Http::STATUS_NO_CONTENT);
```
ChristophWurst, line 955: add logging, else someone will have a hard life trying to find out what went wrong

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -346,6 +348,88 @@ public function requiresFollowUp(
...
+	 */
+	public function requiresTranslation(
+		Account $account,
+		Mailbox $mailbox,
+		Message $message,
+		string $currentUserId,
+	): ?bool {
+		try {
+			$manager = $this->container->get(IManager::class);
+		} catch (ContainerExceptionInterface $e) {
```
ChristophWurst: the interface exists since 30. you can inject it, I think

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -346,6 +348,88 @@ public function requiresFollowUp(
...
+emailText: \"$messageBody\"
+language: \"$language\"
+The JSON output should be in the form: {\"needsTranslation\": true}
+Never return null or undefined.";
+		$task = new Task(FreePromptTaskType::class, $prompt, Application::APP_ID, $currentUserId);
+
+		$manager->runTask($task);
+
+		$this->logger->error('Translation check for message',
+			['messageId' => $message->getId(), 'language' => $language, 'output' => $task->getOutput()]);
```
ChristophWurst:
```suggestion
		$this->logger->error(
			'Translation check for message',
			[
				'messageId' => $message->getId(),
				'language' => $language,
				'output' => $task->getOutput(),
			]
		);
```

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -346,6 +348,88 @@ public function requiresFollowUp(
...
+
+		// Can't use json_decode() here because the output contains additional garbage
+		$result = preg_match('/{\s*"needsTranslation"\s*:\s*true\s*}/i', $task->getOutput()) === 1;
```
ChristophWurst: as Psalm complains needs a null check for the output

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -346,6 +348,88 @@ public function requiresFollowUp(
...
+		} catch (ContainerExceptionInterface $e) {
+			throw new ServiceException(
+				'Text processing is not available in your current Nextcloud version',
+				0,
+				$e,
+			);
+		}
+
+		if (!in_array(FreePromptTaskType::class, $manager->getAvailableTaskTypes(), true)) {
+			throw new ServiceException('No language model available for checking translation needs');
```
ChristophWurst: nit: you could also return `null`, which is a bit cheaper than throwing an exception

lib/Service/AiIntegrations/AiIntegrationsService.php
hamza221: This is a leftover from debugging 🥲🙈

ChristophWurst (approved):

PR #11449, opened 2025-08-05
fix: avoid polluting the logs in the task processing event listner

hamza221:
The log in question 
```
{"reqId":"5TGGnAJAJ8aHBRa8uULc","level":2,"time":"2025-08-05T13:08:24+00:00","remoteAddr":"192.168.21.5","user":"admin","app":"PHP","method":"GET","url":"/index.php/apps/mail/api/messages/3/needsTranslation","message":"Undefined array key 1 at /var/www/html/apps/mail/lib/Listener/TaskProcessingListener.php#49","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36","version":"32.0.0.1","data":{"app":"PHP"}}
```

https://github.com/nextcloud/server/blob/master/lib/public/TaskProcessing/Task.php#L83 custome Id is nullable and defaults to empty string


ChristophWurst:
> I don't like this fix either cause customeId doesn't need to contain ':'

explicitly test the value to be not null and contains a (single) `:`?

PR #11454, opened 2025-08-06
fix(ai-integrations): TextProcessingManager is available on all maintained versions

hamza221:
`TextProcessingManager` is available on all maintained mail versions >= Nextcloud 30


ChristophWurst (approved):
My only remark is that this should be a `refactor` commit. There is no fix :)

:clap: otherwise and thanks for extracting this to a separate PR. That makes it a lot easier to review!

PR #11497, opened 2025-08-11
fix(text-blocks): allow editing in composer and open api for non admin users

### LINKED ISSUE #11496: Make text blocks editable from composer

As a user I want to be able to edit my Text block from the composer action menu 

_No response_

_No response_

### Additional context

_No response_

hamza221:
fix #11496


src/components/textBlocks/ListItem.vue
```diff
@@ -366,24 +366,10 @@ export default {
...
-	&__preview {
-		overflow: hidden;
-		text-overflow: ellipsis;
-		white-space: nowrap;
-	}
-	&__shares {
-		width: 100%;
+	padding-inline-start: 0px !important;
+	padding-inline-end: 0px !important;
+	:deep(.list-item-content){
```
ChristophWurst:
```suggestion
	:deep(.list-item-content) {
```

src/components/textBlocks/TextBlockModal.vue
```diff
@@ -5,12 +5,19 @@
 <template>
 	<NcDialog id="text-block-picker" :name="t('mail','Text blocks')" @closing="handleClose">
 		<p>{{ t('mail','Choose a text block to insert at the cursor') }}</p>
-		<ListItem v-for="textBlock in textBlocks"
+		<ListItem v-for="textBlock in getMyTextBlocks()"
```
ChristophWurst, line 8: I think it should be without `()` because the mapped state is also a computed property, right?

hamza221: they are needed here

Issue #11063, opened 2025-05-05
Single click message actions


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud Mail user I have to process lots of emails. Deleting, archiving, etc currently takes two steps:

1. Click on message actions menu toggle
2. Click action

### Describe the solution you'd like

Offer single click for important actions like *delete*

### Describe alternatives you've considered

N/a

handkerchief333:
I'm upvoting this because the other issue is closed. As described in #5775 it's already possible in the content view, but NOT in the message list/preview list, with one click. I suggesting this solution:
- https://github.com/nextcloud/mail/issues/5775#issuecomment-2851155952

ChristophWurst:
Thank you for your input

nimishavijay:
Gmail:

![Image](https://github.com/user-attachments/assets/f98f748d-3fad-4344-a19d-448fe68f171a)

Outlook: 

![Image](https://github.com/user-attachments/assets/ce428950-b640-4789-a44b-39b819f67ba2)

Thunderbird has no actions (🤔)

The ones that seem to be common to both are mark as unread, mark as important and delete, so we can show that on hover on the left of the 3 dot menu in a message :)

PR #11500, opened 2025-08-11
Add single click actions to envelope

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/11063

Oneline mode:
swappy-20250811_190143

Vertical mode:
swappy-20250811_174612

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -19,9 +19,14 @@
 		:name="addresses"
 		:details="formatted()"
 		:one-line="oneLineLayout"
+		:is-read="showImportantIconVariant"
+		:is-important="isImportant"
 		@click.exact="onClick"
 		@click.ctrl.exact.prevent="toggleSelected"
 		@click.shift.exact.prevent="onSelectMultiple"
+		@delete="onDelete"
+		@toggle-important="onToggleImportant"
+		@toggle-seen="onToggleSeen"
 		@update:menuOpen="closeMoreAndSnoozeOptions">
 		<template #icon>
 			<Star v-if="data.flags.flagged"
@@ -35,8 +40,7 @@
 				:size="18"
 				class="app-content-list-item-star icon-important"
 				:class="{ 'important-one-line': oneLineLayout, 'icon-important': !oneLineLayout }"
-				data-starred="true"
-				@click.prevent="hasWriteAcl ? onToggleImportant() : false" />
+				data-starred="true" />
 			<JunkIcon v-if="data.flags.$junk"
 				:size="18"
 				class="app-content-list-item-star junk-icon-style"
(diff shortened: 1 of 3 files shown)


src/components/EnvelopeQuickActions.vue
```diff
@@ -0,0 +1,80 @@
...
+			default: false,
+		},
+		isImportant: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	data() {
+
+	},
```
ChristophWurst: :fire: ?

src/components/EnvelopeQuickActions.vue
```diff
@@ -0,0 +1,80 @@
...
+
+}
+</script>
+
+<style lang="scss" scoped>
+.list-item-content__quick-actions {
+	display: none;
```
ChristophWurst: I'm not an a11y expert but this feels like it could be tricky for screen readers

src/components/EnvelopeQuickActions.vue
```diff
@@ -0,0 +1,80 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<div class="list-item-content__quick-actions">
+		<EmailUnread v-if="isRead"
+			:size="22"
```
ChristophWurst: I think it should be 20px as per https://github.com/nextcloud/mail/issues/11438

ChristophWurst (commented):
Tested & works :clap: 

The important/unimportant is action duplicates the action of the clickable important icon on the avatar. Perhaps we should remove the avatar icon clickability in favor of the new action

src/components/EnvelopeQuickActions.vue
GVodyanov: I set it to 18px for now per Nimisha's request... maybe we can discuss this in the team call later?

src/components/EnvelopeQuickActions.vue
GVodyanov: I looked into this, screen readers [don't read things with `display: none`](https://www.nomensa.com/blog/how-improve-web-accessibility-hiding-elements/), meaning that users with a screen reader won't be able to see the single click actions, seeing as I don't think they have the ability to hover.

Having said that, I wouldn't consider it a problem since all the actions are still accessible from the actions menu.

GVodyanov:
> Tested & works 👏
> 
> The important/unimportant is action duplicates the action of the clickable important icon on the avatar. Perhaps we should remove the avatar icon clickability in favor of the new action

Yeah I agree, especially considering that we are using clicking the avatar for multiselect now, and it becomes a little confusing.

GVodyanov:
New look!

swappy-20250814_133957

Issue #11328, opened 2025-06-30
Design review v5


hamza221:
## Thread list 

* [x] All items should have the same height if possible 
* [x] When you hover over a mail, the width of the preview text should not change (now the text jumps/ellipsises on hover)
* [x] In priority inbox, the whitespace above important is too much, reduce it a bit (the height above "Other" seems better)
* [x] Reduce whitespace above "Follow up"
* [x] Improve element spacing, grouping and text densitiy #11316 

## Message content

* [x] Change the header design to look more like Talk conversation header 
* [x] Avatar: Space on the top and left should be equal
* [x] Content text should be aligned with the text of the header 
* [x] Message content itself (needs prototypes)
  - [x] Only 2 states: one collapsed fully, one expanded fully, no max-height for each message
  - [x] Full width boxes?
  - [X] Limit width of text to 700 px? check others
* [x] Thread summary https://github.com/nextcloud/mail/pull/11575
  - [x] Style it like a regular message in a thread
  - [x] avatar --> sparkle icon
  - [x] Sender name --> "Thread summary"
  - [x] sender email --> Nextcloud assistany
  - [x] body --> summary content
* [x] Bug: clicking on a message should scroll to the start of the message not ? middle?

## Navigation

* [x] #11918
* [x] #11380
  - [x] text of "Show only subscribed..." text is indented slightly by like 1px
  - [x] "mailbox" --> "folder" in all cases
* [x] chevron-down icon for selected account should be white

## Design changes that will not be delivered in the upcoming release

-  Borders and separation?
    - all containers
    - only lines between each message?
    - shadows?
  - Don't center align
  - Add horizontal lines between the list items

## Not reproducible anymore
* text of any selected folder should be white (when you select send for example )

GretaD:
all the planned ones are done :)

PR #11508, opened 2025-08-14
feat: improve the height and width of threads

GretaD:
ref #11328 

fixed in this pr:

- Full width boxes
- Only 2 states: one collapsed fully, one expanded fully, no max-height for each message
- Avatar: Space on the top and left should be equal

before
Screenshot from 2025-08-14 17-27-20

after
Screenshot from 2025-08-14 17-23-53


ChristophWurst:
can this be converted to a fix so it makes sense to backport? it's not adding a new feature

PR #11517, opened 2025-08-18
fix: ckeditor dropdown overlaping

GretaD:
The problem:
Screenshot from 2025-08-18 15-21-35

The solution i propose in this pr is to make the recipient info column smaller by 30px. Since the problem only occurs when the right pane is shown. 
My initial solution was to do it like we do it on the signature section, that when the width is reached, the elements start creating a third row. Here that will not work because it will be cut on the bottom(same problem we have now on the left) and it glitches. The recipient info doesnt need that much space anyway(in my opinion)

after
Screenshot from 2025-08-18 15-27-13

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -658,7 +658,7 @@ export default {
 }
 
 .right-pane {
-	flex: 0 0 400px;
+	flex: 0 0 370px;
 	overflow-y: auto;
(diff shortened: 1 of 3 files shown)


ChristophWurst:
> The solution i propose in this pr is to make the recipient info column smaller by 30px.

Doesn't that mean the problem will return if we add more formatting options?

PR #11540, opened 2025-08-20
fix(ui): use gradient as fallback for themed fluid design

ChristophWurst:
The png looks odd when the instance is not blue. Therefore we want to show a simple gradient as fallback for now. Later we might be able to use an svg instead of the png and adjust gradient stop colors.

|   | Light | Dark |
|--------|--------|--------|
| Default | Bildschirmfoto vom 2025-08-20 11-03-30 | Bildschirmfoto vom 2025-08-20 11-03-16 |
| Custom | Bildschirmfoto vom 2025-08-20 11-04-42 | Bildschirmfoto vom 2025-08-20 11-04-30 | 

This is a follow-up to https://github.com/nextcloud/mail/pull/11504.

### DIFF
--- a/src/components/NoMessageSelected.vue
+++ b/src/components/NoMessageSelected.vue
@@ -5,7 +5,8 @@
 
 <template>
 	<AppContentDetails class="app-content no-message-selected"
-		:style="{ 'backgroundImage': backgroundImgSrc }">
+		:class="{ 'no-message-selected--themed': isThemed, }"
+		:style="{ 'backgroundImage': isThemed ? undefined: backgroundImgSrc, }">
 		<div class="no-message-selected__heading">
 			{{ t('mail', 'Welcome to {productName} Mail', { productName }, null, {escape: false}) }}
 		</div>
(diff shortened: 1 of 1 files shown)


src/components/NoMessageSelected.vue
```diff
@@ -62,6 +72,14 @@ export default {
 	background-repeat: no-repeat;
 	background-position: right 100% bottom 40%;
 
+	/** fallback gradient when the theme color isn't standard blue */
+	&--themed {
+		background: radial-gradient(100% 100% at 100% 100%, var(--color-primary-element) 0%, rgba(var(--color-main-background-rgb), 0) 100%), var(--color-main-background);
```
ChristophWurst, line 77: @nimishavijay I went with the same gradient for light and dark because we can't use `background: radial-gradient(100% 100% at 100% 100%, rgba(var(--color-primary-element-rgb), 0.5) 0%, rgba(var(--color-main-background-rgb), 0) 100%), var(--color-main-background)` right now without `--color-primary-element-rgb`

PR #11568, opened 2025-08-23
fix(ai-integration): set the language explicitly for message summary

hamza221:
Generate message summaries in system language

### DIFF
--- a/lib/Service/AiIntegrations/AiIntegrationsService.php
+++ b/lib/Service/AiIntegrations/AiIntegrationsService.php
@@ -70,7 +70,7 @@ PROMPT;
 			$this->logger->info('No text summary provider available');
 			return;
 		}
-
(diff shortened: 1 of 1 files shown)


lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -97,6 +97,7 @@ public function summarizeMessages(Account $account, array $messages): void {
 				// construct prompt and task
 				$messageBody = $message->getPlainBody();
 				$prompt = "You are tasked with formulating a helpful summary of a email message. \r\n"
+						  . "The summary should be in the language of this language code ".$language.". \r\n"
```
ChristophWurst:
```suggestion
						  . "The summary should be in the language of this language code ".$language.". \r\n"
```

cs

PR #11575, opened 2025-08-25
feat: make summary thread similar to normal thread style

GretaD:
ref #11328 

smart reply:
Screenshot from 2025-09-01 18-40-23

thread summary:
Screenshot from 2025-09-02 17-08-38

Envelope summary:
Screenshot from 2025-09-02 16-38-05

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -83,7 +83,7 @@
 				<div v-if="data.encrypted || data.previewText"
 					class="envelope__preview-text"
 					:title="data.summary ? t('mail', 'This summary was AI generated') : null">
-					<SparkleIcon v-if="data.summary" :size="15" />
+					<NcAssistantIcon v-if="data.summary" :size="15" class="envelope__preview-text__icon" />
 					{{ isEncrypted ? t('mail', 'Encrypted message') : data.summary ? data.summary.trim() : data.previewText.trim() }}
 				</div>
 				<EnvelopeSingleClickActions v-if="oneLineLayout"
@@ -357,13 +357,12 @@ import {
 	NcActionLink as ActionLink,
 	NcActionSeparator,
 	NcActionInput,
-	NcActionText as ActionText,
+	NcActionText as ActionText, NcAssistantIcon,
 } from '@nextcloud/vue'
 import EnvelopeSkeleton from './EnvelopeSkeleton.vue'
 import AlertOctagonIcon from 'vue-material-design-icons/AlertOctagonOutline.vue'
 import Avatar from './Avatar.vue'
 import IconCreateEvent from 'vue-material-design-icons/CalendarOutline.vue'
-import SparkleIcon from 'vue-material-design-icons/CreationOutline.vue'
 import ClockOutlineIcon from 'vue-material-design-icons/ClockOutline.vue'
 import CheckIcon from 'vue-material-design-icons/Check.vue'
 import ChevronLeft from 'vue-material-design-icons/ChevronLeft.vue'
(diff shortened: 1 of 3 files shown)


nimishavijay:
Looks really nice! Couple of points: 

- [ ] we decided that the assistant components should have tertiary buttons unless it's generating something using AI, so we can change the "Go to latest message" button to a tertiary-no-background button 
 
- [ ] forgot to mention in the original issue, but in dark mode we should use the following colours: 
Dark mode:
Background: #221D2B 
border: linear-gradient(125deg, #0C3A65 50%, #6204A5 125%)

- [ ] nothing to change for now but just a heads up that we will adopt color variables for the assistant elements to be consistent across different apps so once they are in we can do a follow up to use the variables instead of gradient values (ref https://github.com/nextcloud/server/issues/54665)

nimishavijay (changes requested):
Newest screenshot looks really nice! only 3 small notes:
- [x] I see that in the German screenshot the "Summary is AI generated" notice is still in english, it should also be translated to whatever language
- [ ] The strings seem to wrap and ellipsize a lot because they're all too long in German :( Maybe we could do "Newest message" instead of "Go to newest message"?
- [ ] The assistant variables are available now just as an FYI https://github.com/nextcloud/server/pull/54679 they should be used as much as possible in case they are changed later on closer to the release.

src/components/Envelope.vue
```diff
@@ -83,7 +83,7 @@
 				<div v-if="data.encrypted || data.previewText"
 					class="envelope__preview-text"
 					:title="data.summary ? t('mail', 'This summary was AI generated') : null">
-					<SparkleIcon v-if="data.summary" :size="15" />
+					<NcAssistantIcon v-if="data.summary" :size="15" class="summary-icon"/>
```
ChristophWurst:
```suggestion
					<NcAssistantIcon v-if="data.summary" :size="15" class="envelope__preview-text__icon" />
```

src/components/ThreadSummary.vue
```diff
@@ -81,47 +88,65 @@ export default {
...
+			flex-direction: column;
+			line-height: 1.2;
+		}
+
+		&__title {
+			font-weight: bold;
+		}
+
+		&__brand {
+			color:var(--color-text-maxcontrast)
```
ChristophWurst:
```suggestion
			color: var(--color-text-maxcontrast)
```

src/components/ThreadSummary.vue
```diff
@@ -81,47 +88,65 @@ export default {
...
-                &__icon{
-                    color:var(--color-primary-element);
-                    margin-inline: 5px
-                }
-            }
-        }
+		&__info {
+			display: flex;
+			align-items: center;
+			gap: 4px;
```
ChristophWurst: nit: https://github.com/nextcloud/mail/issues/9944

GretaD:
> * In the AI replies, we can remove the icon (the gradient outline is already very eye catching) and make the text color --main-text (I'm concerned about a11y of the gradient on unbolded text)

I can change the colour of the text, but removing the icon means changing the component that we just created for the assistantbuttons: https://nextcloud-vue-components.netlify.app/#/Components/Nextcloud%20Assistant?id=ncassistantbutton

I mean, i can remove it just for mail with :deep(), but that means that the assistant buttons will be different in different part of the software.

marcoambrosini:
I think it is fine to use Text only buttons too

GretaD:
Screenshot from 2025-09-03 15-14-42

PR #11529, opened 2025-08-19
feat: store envelopes as list of threads

hamza221:
Ref https://github.com/nextcloud/mail/issues/11065#issuecomment-3083354991

PR revived with assistance from Claude:claude-opus-4-8

## Summary by CodeRabbit

* **New Features**
  * Threaded mail views now support conversation-aware grouping for sync and search results.
  * Thread message retrieval and ordering are now configurable (defaulting to newest-first).

* **Bug Fixes**
  * Improved threaded attachment display by aggregating attachments across the full conversation.
  * Updated thread ordering for more consistent newest-first behavior.

* **Tests**
  * Updated unit/integration tests to reflect grouped (nested) message results and threaded state handling.

PR #11596, opened 2025-08-28
feat: provide attachement names for messages

hamza221:


hamza221:
I realized this can be merged independently from #11529, rebasing

lib/Service/Attachment/AttachmentService.php
```diff
@@ -249,6 +259,31 @@ public function handleAttachments(Account $account, array $attachments, \Horde_I
...
+				true
+			);
+			$attachments = $imapMessage->getAttachments();
+		} catch (ServiceException $e) {
+			$this->logger->error('Could not get attachment names', ['exception' => $e, 'messageId' => $message->getUid()]);
+		}
+		$result = array_map(static function (array $attachment) {
+			return $attachment['fileName'];
+		}, $attachments);
+		$this->cache->set($message->getUid(), $result);
```
ChristophWurst: UIDs are unique in the scope of a single mailbox. You'll run into conflicts across mailboxes, and worse, across user accounts. Use the primary key or find another unique identifier.

ChristophWurst (commented):
Works, but I'm not a big fan. It's using the distributed cache as storage. Cons:

1. Does not work when there is no distributed cache. It will still do the expensive fetching, but the results are losts.
2. For small installations that use APCu for local+distributed caching this will inflate cache memory quite a bit and risk that the cache is fully evicted because it's running full.

PR #11608, opened 2025-09-01
enh: thread attachement list front end

hamza221:
image

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -142,6 +142,8 @@ class Message extends Entity implements JsonSerializable {
 
 	/** @var bool */
 	private $fetchAvatarFromClient = false;
+	/** @var array */
+	private $attachments = [];
 
 	public function __construct() {
 		$this->from = new AddressList([]);
@@ -312,6 +314,14 @@ class Message extends Entity implements JsonSerializable {
 		return $this->avatar;
 	}
 
+	public function setAttachments(array $attachments): void {
+		$this->attachments = $attachments;
+	}
+
+	public function getAttachments(): array {
+		return $this->attachments;
+	}
+
 	#[\Override]
 	#[ReturnTypeWillChange]
 	public function jsonSerialize() {
@@ -359,6 +369,7 @@ class Message extends Entity implements JsonSerializable {
 			'mentionsMe' => $this->getMentionsMe(),
 			'avatar' => $this->avatar?->jsonSerialize(),
 			'fetchAvatarFromClient' => $this->fetchAvatarFromClient,
+			'attachments' => $this->getAttachments(),
 		];
 	}
 }
--- a/lib/Service/Attachment/AttachmentService.php
+++ b/lib/Service/Attachment/AttachmentService.php
@@ -18,7 +18,10 @@ use OCA\Mail\Contracts\IMailManager;
 use OCA\Mail\Db\LocalAttachment;
 use OCA\Mail\Db\LocalAttachmentMapper;
 use OCA\Mail\Db\LocalMessage;
+use OCA\Mail\Db\Mailbox;
+use OCA\Mail\Db\Message;
 use OCA\Mail\Exception\AttachmentNotFoundException;
+use OCA\Mail\Exception\ServiceException;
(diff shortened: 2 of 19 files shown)


lib/Db/Message.php
```diff
@@ -143,6 +143,8 @@ class Message extends Entity implements JsonSerializable {
 
 	/** @var bool */
 	private $fetchAvatarFromClient = false;
+	/** @var array */
```
ChristophWurst, line 146: nit: you can move `array` down as native property type and use the phpdoc for a stricter type. is that an array of strings?

lib/Db/Message.php
```diff
@@ -313,6 +315,14 @@ public function getAvatar(): ?Avatar {
 		return $this->avatar;
 	}
 
+	public function setAttachments(array $attachments): void {
```
ChristophWurst, line 318: nit: add phpdoc with the specific array type

lib/Db/Message.php
```diff
@@ -313,6 +315,14 @@ public function getAvatar(): ?Avatar {
...
+		$this->attachments = $attachments;
+	}
+
+	public function getAttachments(): array {
```
ChristophWurst, line 322: nit: third time - specific type via phpdoc

lib/Service/Attachment/AttachmentService.php
```diff
@@ -51,6 +55,10 @@ class AttachmentService implements IAttachmentService {
 	 * @var LoggerInterface
 	 */
 	private $logger;
+	/**
+	 * @var ICache
```
ChristophWurst, line 61: psalm is right. misses an import

src/components/Envelope.vue
```diff
@@ -923,6 +944,49 @@ export default {
 			return shortRelativeDatetime(new Date(this.data.dateInt * 1000))
 		},
 
+		addVisibleAttachements() {
```
ChristophWurst: nit: is that not possible with a conditional class? I'd prefer a modifier class over direct style manipulation

src/components/icons/FileIcon.vue
```diff
@@ -0,0 +1,107 @@
...
+			if (FILE_EXTENSIONS_WORD_PROCESSING.includes(this.extension)) {
+				this.color = '--color-info-text'
+			} else if (FILE_EXTENSIONS_SPREADSHEET.includes(this.extension)) {
+				this.color = '--color-border-success'
+			} else if (FILE_EXTENSIONS_PRESENTATION.includes(this.extension)) {
+				this.color = '--color-favorite'
+			} else if (this.extension === 'pdf') {
+				this.color = '--color-text-error'
+			}
+		},
```
ChristophWurst, line 92: another nit: might be achieveable with conditional modified classes

ChristophWurst (commented):
Looks good so far

lib/Db/Message.php
hamza221: Backend changes are to be removed in favor of https://github.com/nextcloud/mail/pull/11596 , fixing tests now in the other Pr

ChristophWurst:
```
 TypeError: can't access property "split", this.fileName is null
    mounted FileIcon.vue:77
    VueJS 19
    addEnvelopesMutation actions.js:2039
    addEnvelopesMutation actions.js:2035
    r pinia.mjs:1399
    fetchEnvelopes actions.js:693
    ge tap.js:26
    a _dispatchable.js:38
    t _curry2.js:19
    t _curry1.js:16
    promise callback*49519/pe< andThen.js:46
    t _curry2.js:19
    t _curry1.js:16
    C _pipe.js:3
    r _arity.js:10
    fetchEnvelopes actions.js:699
    s sessionExpiryHandler.js:13
    fetchEnvelopes actions.js:673
    r pinia.mjs:1399
    loadEnvelopes
```

hamza221:
> * [ ]  BUG I see attachments (sometimes also just a +1 when it's the sole attachment) 

Screen shot look like you have a tiny viewport, attachements would only show if there's enough space

ChristophWurst:
- [ ] BUG clicking on an attachment name should preview the file, but not open that email.

PR #11618, opened 2025-09-03
enh: introduce quick actions

hamza221:


lib/Controller/ActionStepController.php
```diff
@@ -0,0 +1,123 @@
...
+	 * @NoAdminRequired
+	 *
+	 * @return JsonResponse
+	 */
+	#[TrapError]
+	public function findAllStepsForAction(int $actionId): JsonResponse {
+		if ($this->uid === null) {
+			return JsonResponse::error('User not found', Http::STATUS_UNAUTHORIZED);
+		}
+		$actionSteps = $this->actionStepService->findAll($actionId);
```
ChristophWurst: pass the user ID too to prevent https://en.wikipedia.org/wiki/Insecure_direct_object_reference

ChristophWurst (commented):
Just raising the red flag early

The rest looks very clean so far. Keep the direction :)

src/components/quickActions/Settings.vue
```diff
@@ -0,0 +1,324 @@
...
+				</Container>
+				<NcActions :menu-name="t('mail', 'Add another action')">
+					<template #icon>
+						<PlusIcon :size="20" />
+					</template>
+					<NcActionButton @click="addQuickAction('markAsSpam')">
+						<template #icon>
+							<AlertOctagonIcon :size="20" />
+						</template>
+						Mark as spam
```
GretaD: missing translation

src/components/quickActions/Settings.vue
```diff
@@ -0,0 +1,324 @@
...
+					</NcActionButton>
+					<NcActionButton @click="addQuickAction('applyTag')">
+						<template #icon>
+							<TagIcon :size="20" />
+						</template>
+						Tag
```
GretaD: same here, and couple of more below :)

nimishavijay (changes requested):
Looks pretty neat! Some comments and questions: 

- [x] any rationale for the screw (?) icon? I tried to find something that looked like quick actions on material design symbols but couldn't. However this [email-fast-outline](https://pictogrammers.com/library/mdi/icon/email-fast-outline/) community mdi icon seems pretty fitting 
- [x] in the action menu, add a chevron-right icon on the right to indicate there is a submenu
- [x] "Back to actions" --> "Back to all actions"
- [ ] inside the modal
  - [x] Add a heading to the modal "Edit quick action" or "Create quick action" depending on what you're doing
  - [x] Add a h3 after the title field "Do the following actions"
  - [x] Use the correct strings for the actions like "Mark as unread" "Move thread to the folder" etc instead of "markAsUnread" "moveThread" etc
  - [x] Nice to have: show the icon for each action in the modal if they are added (like if "mark as unread" is selected show its icon in addition to the text)


- [ ] follow ups: 
  - [ ] allow reorder and set one action as "primary" or "first" action for parity with outlook

ChristophWurst:
@hamza221 please tick off Nimisha's requested changes when they are done for better clarity of the PR status

hamza221:
> Looks pretty neat! Some comments and questions:
> 
> * [x]  any rationale for the screw (?) icon? I tried to find something that looked like quick actions on material design symbols but couldn't. However this [email-fast-outline](https://pictogrammers.com/library/mdi/icon/email-fast-outline/) community mdi icon seems pretty fitting
> * [x]  in the action menu, add a chevron-right icon on the right to indicate there is a submenu
> * [x]  "Back to actions" --> "Back to all actions"
> * [ ]  inside the modal
>   
>   * [ ]  Add a heading to the modal "Edit quick action" or "Create quick action" depending on what you're doing
>   * [ ]  Add a h3 after the title field "Do the following actions"
>   * [ ]  Use the correct strings for the actions like "Mark as unread" "Move thread to the folder" etc instead of "markAsUnread" "moveThread" etc
>   * [ ]  Nice to have: show the icon for each action in the modal if they are added (like if "mark as unread" is selected show its icon in addition to the text)
> * [ ]  follow ups:
>   
>   * [ ]  allow reorder and set one action as "primary" or "first" action for parity with outlook

@nimishavijay  reordering is already possible,and do you mean to set the first action to primary color or is there a functionality change

appinfo/routes.php
```diff
@@ -505,6 +505,16 @@
 			'url' => '/api/textBlocks/{id}/shares',
 			'verb' => 'GET',
 		],
+		[
+			'name' => 'actionStep#findAllStepsForAction',
+			'url' => '/api/actionStep/{actionId}/steps',
```
ChristophWurst: IMO camel case in URLs are uncommon. Kebab case might be better here. See out-of-office. I missed this for textBlocks

lib/Service/ActionStepService.php
```diff
@@ -0,0 +1,141 @@
...
+ */
+
+namespace OCA\Mail\Service;
+
+use OCA\Mail\Db\ActionStep;
+use OCA\Mail\Db\ActionStepMapper;
+use OCA\Mail\Exception\ServiceException;
+use OCP\AppFramework\Db\DoesNotExistException;
+
+class ActionStepService {
```
ChristophWurst: Action steps are part of the quick actions feature. Therefore ActionStepService should be integrated into QuickActionsService.

Services are generally more high level. Only for low level mappers and similar we separate by entity

lib/Migration/Version5005Date20250903114906.php
```diff
@@ -0,0 +1,66 @@
...
+			return null;
+		}
+		$table = $schema->createTable('mail_actions');
+
+		$table->addColumn('id', Types::INTEGER, [
+			'autoincrement' => true,
+			'notnull' => true,
+			'length' => 4,
+		]);
+		$table->addColumn('owner', Types::STRING, [
```
ChristophWurst: this table has a foreign key on mail_accounts. mail_accounts already has the userid. Doesn't this make this column redundant? You can join the mail_accounts table when checking ownership.

lib/Migration/Version5005Date20250903114909.php
```diff
@@ -0,0 +1,95 @@
...
+				]
+			);
+		}
+		if ($schema->hasTable('mail_mailboxes')) {
+			$table->addForeignKeyConstraint(
+				$schema->getTable('mail_mailboxes'),
+				['mailbox_id'],
+				['id'],
+				[
+					'onDelete' => 'CASCADE',
```
ChristophWurst: In our last discussion it was decided that steps should not silently vanish when their mailbox vanishes, but we set the mailbox_id to `null`. What's the reason for going with CASCADE DELETE?

lib/Service/ActionStepService.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public function update(ActionStep $action, string $name, int $order, ?int $tagId, ?int $mailboxId): ActionStep {
+		$action->setName($name);
+		$action->setOrder($order);
+		if ($tagId !== null) {
+			$action->setTagId($tagId);
+		}
+		if ($mailboxId !== null) {
+			$action->setMailboxId($mailboxId);
+		}
```
ChristophWurst: don't these `if`s make it impossible to unset the tag/mailbox IDs?

lib/Service/ActionStepService.php
```diff
@@ -0,0 +1,141 @@
...
+			throw new ServiceException('Cant perform actions after deleteThread');
+		}
+		$this->actionStepMapper->swapOrder($actionId, $newOrder);
+	}
+
+	/**
+	 * @throws ServiceException
+	 */
+	private function validateActionStep(string $name, int $order, int $actionId, ?int $tagId, ?int $mailboxId): void {
+		if (!in_array($name, self::AVAILABLE_ACTION_STEPS)) {
```
ChristophWurst:
```suggestion
		if (!in_array($name, self::AVAILABLE_ACTION_STEPS, true)) {
```

Please always use strict mode for in_array

src/App.vue
```diff
@@ -52,6 +52,7 @@ export default {
 		this.sync()
 		await this.mainStore.fetchCurrentUserPrincipal()
 		await this.mainStore.loadCollections()
+		await this.mainStore.fetchAllQuickActions()
```
ChristophWurst, line 55: perf: this should be preloaded as initial state

src/components/Envelope.vue
```diff
@@ -740,6 +784,117 @@ export default {
...
+				})
+				if (check) {
+					this.filteredQuickActions.push({
+						...action,
+						steps,
+						icon: steps[steps.length - 1]?.name,
+					})
+				}
+			}
+		},
```
ChristophWurst: looks like a good candidate for a computed prop, no?

src/components/quickActions/Action.vue
```diff
@@ -0,0 +1,92 @@
...
+			type: Object,
+			required: true,
+		},
+	},
+	computed: {
+		mainStore() {
+			return useMainStore()
+		},
+		needsSelection() {
+			return ['applyTag', 'moveThread'].includes(this.action.name)
```
ChristophWurst, line 60: this tells me that action.name is the technical identifier …

src/components/quickActions/Action.vue
```diff
@@ -0,0 +1,92 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<div class="action">
+		<p v-if="!needsSelection">
+			{{ action.name }}
```
ChristophWurst: … which is printed here. Is this missing a conversion to a translated UI string?

src/components/quickActions/Action.vue
```diff
@@ -0,0 +1,92 @@
...
+		</p>
+		<NcSelect v-else
+			:input-label="action.name"
+			:options="options"
+			label="value"
+			:model-value="selectedOption"
+			@update:modelValue="update" />
+		<NcButton aria-label="delete" variant="tertiary-no-background" @click="$emit('delete')">
```
ChristophWurst: l10n missing for aria label

src/components/quickActions/Settings.vue
```diff
@@ -0,0 +1,324 @@
...
+
+.modal-content__action{
+	padding: 9px;
+}
+
+.modal-content__save {
+	position: absolute;
+	bottom: 10px;
+	inset-inline-end: 10px;
+}
```
ChristophWurst: style: reformat using scss using the BEM convention

src/components/quickActions/Settings.vue
```diff
@@ -0,0 +1,324 @@
...
+
+:deep(.v-select){
```
ChristophWurst, line 334: nit: add code comment explaining why a deep selector is necessary here

src/components/quickActions/Action.vue
ChristophWurst: image

probably the reason for this ^

ChristophWurst:
> * no feedback when clicking _Save_ for a new quick action.

- [ ] BUG when clicking on *Edit* of an existing action there is also a lag with no feedback. Missing loading state? Actions menu also stays open

image

src/components/Envelope.vue
```diff
@@ -320,6 +326,29 @@
...
+				<ActionButton :close-after-click="false"
+					@click="closeQuickActionsMenu()">
+					<template #icon>
+						<ChevronLeft :size="20" />
+					</template>
+					{{ t('mail', 'Back to all actions') }}
+				</ActionButton>
+				<ActionButton v-for="action in filteredQuickActions"
+					:key="action.id"
+					@click="executeQuickAction(action)">
```
ChristophWurst:
```suggestion
					:key="action.id"
					:close-after-click="true"
					@click="executeQuickAction(action)">
```

otherwise the action menu stays open

src/components/EnvelopeList.vue
```diff
@@ -137,11 +138,17 @@
 			:account="account"
 			:envelopes="selectedEnvelopes"
 			@close="onCloseTagModal" />
+
+		<NcDialog v-if="showQuickActionsSettings"
+			:name="t('mail', 'manage quick actions')"
```
ChristophWurst:
```suggestion
			:name="t('mail', 'Manage quick actions')"
```

src/components/Envelope.vue
```diff
@@ -103,7 +103,7 @@
 				fill-color="var(--color-primary-element)" />
 		</template>
 		<template #actions>
-			<EnvelopePrimaryActions v-if="!moreActionsOpen && !snoozeOptions">
+			<EnvelopePrimaryActions v-if="!moreActionsOpen && !snoozeOptions" id="primary-actions">
```
st3iny, line 106:
```suggestion
			<EnvelopePrimaryActions v-if="!moreActionsOpen && !snoozeOptions">
```

Appears to be unused. Otherwise, please try not to duplicate the id (as multiple menus may be mounted simultaneously).

src/components/Envelope.vue
```diff
@@ -740,6 +794,103 @@ export default {
...
+				}
+			}
+		},
+		async executeQuickAction(action) {
+			this.quickActionLoading = true
+			try {
+				for (const step of action.steps) {
+					console.debug('Executing quick action step', step)
+					switch (step.name) {
+					case 'markAsSpam':
```
st3iny, line 828: Would be nice to have them as constants some where instead of magic strings.

lib/Migration/Version5005Date20250903114906.php
```diff
@@ -0,0 +1,62 @@
...
+		$schema = $schemaClosure();
+		if ($schema->hasTable('mail_actions')) {
+			return null;
+		}
+		$table = $schema->createTable('mail_actions');
+
+		$table->addColumn('id', Types::INTEGER, [
+			'autoincrement' => true,
+			'notnull' => true,
+			'length' => 4,
```
st3iny:
```suggestion
```

Length is not required for int columns.

lib/Migration/Version5005Date20250903114906.php
```diff
@@ -0,0 +1,62 @@
...
+		]);
+		$table->addColumn('name', Types::STRING, [
+			'notnull' => true,
+			'length' => 64,
+		]);
+		$table->addColumn('account_id', Types::INTEGER, [
+			'notnull' => true,
+			'length' => 4,
```
st3iny:
```suggestion
```

Length is not required for int columns.

lib/Migration/Version5005Date20250903114906.php
```diff
@@ -0,0 +1,62 @@
...
+		if ($schema->hasTable('mail_accounts')) {
+			$table->addForeignKeyConstraint(
+				$schema->getTable('mail_accounts'),
+				['account_id'],
+				['id'],
+				[
+					'onDelete' => 'CASCADE',
+				]
+			);
+		}
```
st3iny, line 56: Very nice!

st3iny (commented):
Code looks good otherwise.

src/components/Envelope.vue
hamza221: Would that work for an async functions ?

hamza221: There's a logic flow here 🥲 anyways , this is run for every envelope in the ui , for every quick action. So many duplicated api calls, I'm trying to also add this to the store

ChristophWurst: Right, it would not. Leave it for now

hamza221:
- [ ] followup , add steps to store

ChristophWurst:
- [ ] BUG Apply tag still needs a translatable UI string

image

- [ ] After executing a quick action the action menu of the envelope should reset so that you see the default actions when opening the next time. It now still shows the quick actions list at re-open.

lib/Controller/QuickActionsController.php
```diff
@@ -0,0 +1,112 @@
...
+	 *
+	 * @return JsonResponse
+	 */
+	#[TrapError]
+	public function create(string $name, int $accountId): JsonResponse {
+		if ($this->uid === null) {
+			return JsonResponse::error('User not found', Http::STATUS_UNAUTHORIZED);
+		}
+		try {
+			$quickAction = $this->quickActionsService->create($name, $accountId);
```
ChristophWurst, line 72: This allows me to create an action for someone else's account. Please check if the current user owns the account.

ChristophWurst (commented):
Tested and works :clap: 

One last blocker, then we can :shipit:

lib/Controller/ActionStepController.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @param string $name
+	 *
+	 * @return JsonResponse
+	 */
+	#[TrapError]
+	public function create(string $name, int $order, int $actionId, ?int $tagId = null, ?int $mailboxId = null): JsonResponse {
+		if ($this->uid === null) {
+			return JsonResponse::error('User not found', Http::STATUS_UNAUTHORIZED);
+		}
+		$actionStep = $this->quickActionsService->createActionStep($name, $order, $actionId, $tagId, $mailboxId);
```
ChristophWurst, line 75: Missing ownership check for `$actionId`, so you can create steps for any (guessed) IDs

lib/Controller/ActionStepController.php
```diff
@@ -0,0 +1,122 @@
...
+	public function update(int $id, string $name, int $order, ?int $tagId, ?int $mailboxId): JsonResponse {
+
+		if ($this->uid === null) {
+			return JsonResponse::error('User not found', Http::STATUS_UNAUTHORIZED);
+		}
+
+		$actionStep = $this->quickActionsService->findActionStep($id, $this->uid);
+
+		if ($actionStep === null) {
+			return JsonResponse::error('Action step not found', Http::STATUS_NOT_FOUND);
```
ChristophWurst: ^ Psalm is right. `findActionStep` either returns a value or it throws. You use `TrapError`, which translates `DoesNotExistException` to HTTP 400 already.

lib/Controller/ActionStepController.php
```diff
@@ -0,0 +1,122 @@
...
+			return JsonResponse::fail('Action step not found', Http::STATUS_NOT_FOUND);
+		}
+	}
+
+	public function swapOrder(int $id, int $newOrder): JsonResponse {
+		if ($this->uid === null) {
+			return JsonResponse::error('User not found', Http::STATUS_UNAUTHORIZED);
+		}
+		try {
+			$this->quickActionsService->swapOrder($id, $newOrder);
```
ChristophWurst: No check for ownership. You can swap any step's order

PR #11623, opened 2025-09-04
chore(CI): Adjust testing matrix for Nextcloud 32 on main

nickvergessen:


ChristophWurst:
I'm taking a cheap shortcut and won't add stable30 to the Psalm matrix now because
1. It worked so far, so the type error is more of a theoretical concern than a practical issue
2. We will drop stable30 from the app in 1-2 weeks as it is EOLing

PR #11624, opened 2025-09-04
feat(deps): Add Nextcloud 33 support

nickvergessen:


ChristophWurst (approved):
I just bumped main to 5.5 after branching off 5.4 so I reverted the second bump

Thanks!

PR #11632, opened 2025-09-08
fix: endless burst of requests on uncached mailboxes

### LINKED ISSUE #11024: Opening a mailbox that isn't cached causes an endless burst of requests
### Steps to reproduce

1. Open mail app
2. Open a mailbox that is not cached
3. See the network tab on console

### Expected behavior

It should perform a single fetch and caching the result

### Actual behavior

Opening a mailbox that isn't cached causes an endless burst of requests

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

The logic should be to start an initial sync of the mailbox when we run into this error

Image

GretaD:
fixes #11024

Another approach would be if we handle initialization in sync() instead. Smth like:

`	if (init) {
			await wait(5000)
			await this.loadEnvelopes()
		}`

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -201,17 +201,21 @@ export default {
 		this.stopInterval()
 	},
 	methods: {
-		initializeCache() {
+		async initializeCache() {
 			this.loadingCacheInitialization = true
 			this.error = false
 
-			logger.debug(`syncing folder ${this.mailbox.databaseId} (${this.query}) during cache initalization`)
-			this.sync(true)
-				.then(() => {
-					this.loadingCacheInitialization = false
-
-					return this.loadEnvelopes()
-				})
+			logger.debug(`syncing folder ${this.mailbox.databaseId} (${this.query}) during cache initialization`)
+			try {
+				await this.sync(true)
+				await wait(5000)
+				await this.loadEnvelopes()
+			} catch (error) {
+				logger.error(`Could not initialize cache of folder ${this.mailbox.databaseId} (${this.searchQuery})`, { error })
+				this.error = error
+			} finally {
+				this.loadingCacheInitialization = false
+			}
 		},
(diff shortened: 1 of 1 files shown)


ChristophWurst:
What caused the endless loop before?

src/components/Mailbox.vue
```diff
@@ -201,17 +201,20 @@ export default {
 		this.stopInterval()
 	},
 	methods: {
-		initializeCache() {
+		async initializeCache() {
```
ChristophWurst, line 204: This method is called when the previous error was MailboxNotCachedError …

src/components/Mailbox.vue
```diff
@@ -201,17 +201,20 @@ export default {
...
-				.then(() => {
-					this.loadingCacheInitialization = false
-
-					return this.loadEnvelopes()
-				})
+			logger.debug(`syncing folder ${this.mailbox.databaseId} (${this.query}) during cache initialization`)
+			try {
+				await this.sync(true)
+				await this.loadEnvelopes()
+			} catch (error) {
```
ChristophWurst, line 216: … so what error is possibly thrown here? MailboxLockedError?

GretaD: yes, when sync is true, it may throw `MailboxLockedError`. Why are you asking? Do you think is unnecessary to catch the error here again?

ChristophWurst: Asking because this will catch *every* error

GretaD:
> Seems intentional?
> 
> > // Keep trying

to keep trying, yes probably. To keep trying forever, i doubt its intentional. Do you think this approach is not the solution we want?

GretaD:
not relevant anymore

Issue #11507, opened 2025-08-14
Envelope height jumping when hovering avatar in one line mode


GVodyanov:
### Steps to reproduce

1. Enter one line mode
2. Hover avatar
3. Envelope becomes thinner

### Expected behavior

Height should remain constant

### Actual behavior

Height shrinks

### PHP engine version

None

### Web server

None

### Database

None

PR #11641, opened 2025-09-09
fix(Envelope): fix oneline mode envelopes jumping when hovering avatars

GVodyanov:
Fix this:

[recording_20250909-142914.webm](https://github.com/user-attachments/assets/e846cf54-e3da-489c-8111-5ae144034e7f)

Fix https://github.com/nextcloud/mail/issues/11507

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -1227,10 +1227,16 @@ export default {
 	}
 }
 
+.hovering-status {
+	// Needs to be the same height as the check-icon and the avatar to prevent automatic resizing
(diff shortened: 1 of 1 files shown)


src/components/Envelope.vue
```diff
@@ -1227,10 +1227,14 @@ export default {
 	}
 }
 
+.hovering-status {
+	height: calc(var(--default-grid-baseline) * 10);
```
ChristophWurst, line 1233: nit: leave a code comment explaining why this is here. someone might accidentally drop the rule if the purpose is not obvious

Issue #11646, opened 2025-09-10
Stop the scrolling to a middle of a (long) email or thread


tflatka:
### Steps to reproduce

1. Open a configured Mail App
2.Click on an email where the content is longer than one screen page
3.It will open the email and scroll automatically to the middle of it

<img width="1747" height="1234" alt="Image" src="https://github.com/user-attachments/assets/9d81d98c-a058-4c02-81a1-e7a34a6dfee2" />

### Expected behavior

Just open the email and do not scroll at all.

### Actual behavior

The email opens and then automatically scrolls to the middle of the content of the email

### Mail app version

5.4.0

### Nextcloud version

31.0.7

### Mailserver or service

Dovecot

### Operating system

Debian

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

This "auto scrolling" was introduced about 4 versions ago

GretaD:
Hello, thank you for your report, this shouldnt happen, you're right. I will work on it

PR #11648, opened 2025-09-10
fix: thread scrolls to middle of iframe

GretaD:
fixes [#11646](https://github.com/nextcloud/mail/issues/11646)


ChristophWurst (approved):
:see_no_evil: 

ok, makes sense. guess we tested with short messages in the past where top/middle makes not a lot of difference?

PR #11651, opened 2025-09-10
CKEditor darkmode and image align

CreatorOfBiontium:
I've made changes that fix the CKEditor colors in dark mode and add the option to align images (left/center/right) when sending emails. I want to extend this option to the signature settings as well, but I struggled with that and will try to implement it later. For now, I'm happy with these changes.

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -682,7 +682,6 @@ https://github.com/ckeditor/ckeditor5/issues/1142
 
 .ck.ck-toolbar .ck-button {
     color: var(--color-main-text) !important;
-    background: transparent !important;
 }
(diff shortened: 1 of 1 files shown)


src/components/textBlocks/ListItem.vue
```diff
@@ -352,6 +354,8 @@ export default {
 		async saveTextBlock() {
 			this.saveLoading = true
 			try {
+				// převod na inline styly
```
ChristophWurst, line 357: leftover? :)

CreatorOfBiontium: Yes, I will remove it in the splitted pr :D

PR #11665, opened 2025-09-11
Fix CKEditor color in darkmode

CreatorOfBiontium:
This is the splitted pr for ckeditor changes. This pr contains fix for the CKEditor in darkmode. 

Before (on hover):
image

After (on hover):
image


CreatorOfBiontium:
I noticed there’s an extra commit from Vodyanov in this branch, which shouldn’t be here. I’ll clean it up and force-push so that only my changes remain. Hope its ok

PR #11666, opened 2025-09-11
Fix image alignment in email composer

CreatorOfBiontium:
This PR focuses on fixing image alignment.
The alignment works when composing and sending emails, but in SignatureSettings.vue it doesn’t display the aligment menu (neither does the button descriptions). That part still needs work.

Additionally:
- When saving a draft or sending a message, CKEditor ignores the added alignment classes.
- The same issue might require adjustments in the email display view.
- CKEditor also ignores image resizing — images are always scaled to the maximum width.


CreatorOfBiontium:
What do you suggest here - the ckeditor doesn't have the plugins for image aligement avalible when running lint. How do we include them?

kesselb:
Thanks for your pr 👍

I think we should evaluate to implement it as creditor plugin. I will have a closer look when time allows.

> How do we include them?

Commit the changes to package.json and package-lock.json.

CreatorOfBiontium:
Adding to package.json not needed – just adjusted imports to satisfy lint and it works :o

PR #11670, opened 2025-09-12
fix: align quick action settings button properly

hamza221:
| b | a |
|--------|--------|
|image | image |

### DIFF
--- a/src/components/quickActions/Settings.vue
+++ b/src/components/quickActions/Settings.vue
@@ -315,9 +315,6 @@ export default {
 }
 </script>
 <style lang="scss" scoped>
-.quick-actions-settings{
-	padding: 10px;
-}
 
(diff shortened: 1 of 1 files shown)


src/components/quickActions/Settings.vue
```diff
@@ -315,9 +315,6 @@ export default {
 }
 </script>
 <style lang="scss" scoped>
-.quick-actions-settings{
-	padding: 10px;
-}
 
 .modal-content{
```
ChristophWurst, line 319:
```suggestion
.modal-content {
```

unrelated, don't fix it here, but I thought we had stylelint for this :smiling_face_with_tear:

hamza221: maybe it has to be adjusted

PR #11691, opened 2025-09-16
fix: sorting for older grouping

### LINKED ISSUE #11567: Sorting broken with new grouping feature
### Steps to reproduce

1.Have a folder with only a few messages preferably less than 25 ( from last month , last week, older ... ) 
2. open folder 
3. change sort order (oldest first) 

### Expected behavior

only older messages are shown

### Actual behavior

newer groups are shown 

Image

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

This only happens when newer messages are loaded into the store that's why you have to make sure you use a smaller folder for testing)

GretaD:
fixes #11567

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -297,7 +297,7 @@ export default {
 		},
 		groupEnvelopes() {
 			const allEnvelopes = this.mainStore.getEnvelopes(this.mailbox.databaseId, this.searchQuery)
-			return this.groupEnvelopesByDate(allEnvelopes, this.mainStore.syncTimestamp)
+			return this.groupEnvelopesByDate(allEnvelopes, this.mainStore.syncTimestamp, 'oldest')
 		},
 	},
 	watch: {
@@ -335,7 +335,7 @@ export default {
 		clearTimeout(this.startMailboxTimer)
 	},
 	methods: {
-		groupEnvelopesByDate(envelopes, syncTimestamp) {
+		groupEnvelopesByDate(envelopes, syncTimestamp, sortOrder = 'newest') {
 			const now = new Date(syncTimestamp)
 			const oneHourAgo = new Date(now.getTime() - 60 * 60 * 1000)
 			const startOfToday = new Date(now.getFullYear(), now.getMonth(), now.getDate())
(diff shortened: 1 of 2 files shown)


src/components/MailboxThread.vue
```diff
@@ -372,8 +375,22 @@ export default {
...
+				groups[key].sort((a, b) => {
+					return sortOrder === 'newest'
+						? b.dateInt - a.dateInt
+						: a.dateInt - b.dateInt
+				})
+			}
+
+			const groupOrder = sortOrder === 'newest'
+				? ['lastHour', 'today', 'yesterday', 'lastWeek', 'lastMonth', 'older']
+				: ['older', 'lastMonth', 'lastWeek', 'yesterday', 'today', 'lastHour']
```
ChristophWurst, line 388: for sorting by oldest it will not show the year of the messages, will it? it will start with *Older*.

GretaD: yes, i think _older_ is good enough to fix the issue reported, we dont have to change it, unless someone has a strong opinion. In the ticket itself I proposed _year_, because i thought thats what we already have.

GretaD: also, changing the whole logic of showing years, then months, then week/recent, is more of a feature request, and here im just fixing a bug :)

ChristophWurst: IMO it's very strange to start with *Older*, because it's missing a context of what it's compared to. If anything, it should start with *Old*. But that's weird.

What is shown in Outlook and others?

GretaD: Outlook has it with Years, then Months(Jan..Feb..), than last week, and today. When we worked on the initial feature, Older was proposed for messages older than last month.

I can change the scope of the PR, because yes, its odd to start with Older, but that will need a bit of more work than just fixing the bug reported.

ChristophWurst: Sounds like a plan

ChristophWurst: The feature was always about covering all sorting options. See https://github.com/nextcloud/mail/issues/11057.

GretaD:
Started from scratch here: https://github.com/nextcloud/mail/pull/11708

Issue #2425, opened 2019-12-30
Ctrl-Enter does not send the mail


jancborchardt:
Usually Ctrl-Enter submits forms or sends messages or sends mails. It’s a quick shortcut to not need to click the "Send" button.

Currently it just inserts a line break, just like Enter.

StCyr:
how should it work on macOS? CTRL-Enter or Meta-Enter?

StCyr:
small study shows that it's not that easy to intercept keypress events with CKEditor.

when using ckeditor's @input event, the event contains the actual ckeditor's content and not the information about the key pressed

Here's what I did:

```javascript
                onTextEditorInputChanged(e) {
                        console.log("onTextEditorInputChanged", e)
                        if (e.keycode == 13 && (navigator.platform.match('Mac') ? e.metaKey : e.ctrlKey)) {
                                this.onSend()
                                return
                        }
                        this.saveDraftDebounced(this.getMessageData())
                },
```

And the result is:

```
1:41:22.965 onTextEditorInputChanged <p>test</p> undefined Composer.vue:378
11:41:29.571 onTextEditorInputChanged <p>test test</p><p>&nbsp;</p> undefined Composer.vue:378
11:41:32.530 onTextEditorInputChanged <p>test test</p><p>test</p> undefined Composer.vue:378
11:41:33.726 onTextEditorInputChanged <p>test test</p><p>testte</p> undefined Composer.vue:378
11:41:34.159 onTextEditorInputChanged <p>test test</p><p>testtest</p> undefined Composer.vue:378
11:41:36.274 onTextEditorInputChanged <p>test test</p><p>testtest</p><p>&nbsp;</p> undefined
```
see also https://ckeditor.com/old/forums/FCKeditor-2.x/changing-behavior-for-Enter-key

ChristophWurst:
Known regression of #1988 https://github.com/nextcloud/mail/blob/4e777e6fbcb3be22ca8ba893fa70f80ff055cc37/src/components/Composer.vue#L104
https://github.com/nextcloud/mail/blob/4e777e6fbcb3be22ca8ba893fa70f80ff055cc37/src/components/Composer.vue#L384-L389

Thanks for having a look @StCyr!

jancborchardt:
> how should it work on macOS? CTRL-Enter or Meta-Enter?

Should be Cmd-Enter (I assume that’s what you mean by Meta-Enter?)

StCyr:
[...]

PR #11710, opened 2025-09-17
feat(composer): send message on Ctrl+Enter

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/2425

Ref https://ckeditor.com/docs/ckeditor5/latest/framework/tutorials/crash-course/keystrokes.html#adding-keyboard-shortcuts


ChristophWurst:
@hamza221 as discussed please test with Mac and see if it works or needs a specific handler

src/components/TextEditor.vue
```diff
@@ -396,6 +396,11 @@ export default {
 				}
 			}, { priority: 'high' })
 
+			editor.keystrokes.set('Ctrl+Enter', (event) => {
+				logger.debug('Detected Ctrl+Enter', event)
```
hamza221: Maybe should be either vague or translate it on macos

hamza221: Detected submit keystroke ?

Issue #11635, opened 2025-09-08
Suggestions for to and cc don’t search through email addresses


jancborchardt:
### Steps to reproduce

1. Remember the email address, but not the name, of a person or organization
2. Type it in the "to" field

### Expected behavior

The email address & contact shows up.

### Actual behavior

It doesn’t show up, seemingly because it doesn't search over the email address but only the display name?
<img width="1126" height="225" alt="Image" src="https://github.com/user-attachments/assets/b6ef68e8-4adc-4217-bb0b-8e5e33d5d2e2" />

Even though it's there:
<img width="1126" height="225" alt="Image" src="https://github.com/user-attachments/assets/02cce648-18a9-417c-9860-807270dc968f" />

### Mail app version

Latest daily

### Nextcloud version

Latest daily

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
Thanks for the report! This is a front-end issue. The network requests show that search results are there for matches on the email. It must be NcSelect that filters out entries that do not match the first line.

PR #11712, opened 2025-09-17
fix: check email in recipient picker

kesselb:
Fix https://github.com/nextcloud/mail/issues/11635

STR:

- Create a contact Bob with email "alice@example.org"
- Enter alice in recipient select

B: Bob not shown
A: Bob shown


jancborchardt:
Nice, thank you @kesselb! :)

PR #11718, opened 2025-09-18
fix: error in mounted hook

### LINKED ISSUE #11711: kitinerary-bin extractor not working
### Steps to reproduce

1. Open an email that contains an iMip calendar request.
2. The event is not extracted.

### Expected behavior

The event should be extracted and the corresponding UI should be shown.

### Actual behavior

The UI for accepting the invitation is not shown.

### Mail app version

5.5.1

### Nextcloud version

31.0.9

### Mailserver or service

_No response_

### Operating system

Fedora Linux 42

### PHP engine version

PHP 8.3

### Nextcloud memory caching

APCu, Redis (but valkey is used instead)

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

On my system, SELinux is in enforcing mode, but I have added rules that allow Nextcloud to execute the `kitinerary-extractor` and these have worked well in the past. This also does not seem to be the issue now, as Nextcloud can still execute the binary (see below) and I don't see any SELinux alerts.

Still, in the Nextcloud logs, the message `KItinerary binary adapter is not available, can't extract information` is printed.
I have found out that this is due to the following code:

https://github.com/ChristophWurst/kitinerary-bin/blob/472f7fa30416cb5b7877ba332d20cf55077fbdc3/src/BinaryAdapter.php#L83-L91

The return value `$ret` is 13, i.e. SIGPIPE. I could only fix this by inserting an `fread` in between the two invocations of `fclose` so that `kitinerary-extractor` does not try to write into a closed pipe. See the following diff:

```
diff --git a/src/BinaryAdapter.php b/src/BinaryAdapter.php
index 6039910..1635b59 100644
--- a/src/BinaryAdapter.php
+++ b/src/BinaryAdapter.php
@@ -86,7 +86,8 @@ class BinaryAdapter implements Adapter, LoggerAwareInterface
                        $this->logger->warning('Could not open ' . realpath($binPath));
                        return false;
                }
                fclose($pipes[0]);
+               fread($pipes[1], 1024);
                fclose($pipes[1]);
                $ret = proc_close($proc);
```

This is probably not good style, but I unfortunately don't have any experience with php.

With this change, I don't get any errors that `kitinerary-extractor` is not executable any more. However, still no extracted events are shown. I have found out that `$content` seems to be always empty in the following code:

https://github.com/nextcloud/mail/blob/bf33e4fa570bfdd04311e4834cba3f3637cc854f/lib/Integration/KItinerary/ItineraryExtractor.php#L62-L80

I know that it worked before I updated my system yesterday, but unfortunately, I updated multiple things at the same time. I updated my OS to Fedora 42 and the Nextcloud to 31.0.9 (from 31.0.8). During this, Mail was updated from version 5.3.2 to 5.5.0. I am now on 5.5.1. So it could be multiple things causing this bug.

If you need any additional information, I'm happy to provide them.

kesselb:
Fix #11711

Regression from https://github.com/nextcloud/mail/pull/11618

The error `Error in mounted hook (Promise/async): "TypeError: this.mainStore.fetchAllQuickActions is not a function"` prevents hasCurrentUserPrincipalAndCollectionsMutation becoming true and thus the imip preview do not work.

B: Opening an email with an imip invitation does not show the preview
A: Opening an email with an imip invitation does show the preview

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -52,7 +52,6 @@ export default {
 		this.sync()
 		await this.mainStore.fetchCurrentUserPrincipal()
 		await this.mainStore.loadCollections()
-		await this.mainStore.fetchAllQuickActions()
 		this.mainStore.hasCurrentUserPrincipalAndCollectionsMutation(true)
 	},
 	methods: {


ChristophWurst (approved):
store does not have a `fetchAllQuickActions` :heavy_check_mark:

PR #11754, opened 2025-09-27
enh: redesign email display on phones

### LINKED ISSUE #11663: Use full screen width on mobile

Image

The email content is not using the full screen width. (Side note: the unsubscribe button seems to overlap.) 

For small screens (mobile), please adjust the layout so that the email content expands to use the full screen width, making it easier to read and interact with on limited space.
(Side note: there’s also an issue with the unsubscribe button that needs attention.)

_No response_

### Additional context

_No response_

CreatorOfBiontium:
Fixes #11663  

### Changes
- Email content now takes the full screen width on phones  
- Topbar partly hides itself on scroll  
- Removed blank space under the email content  
- Added 5px border-radius to the email content  
- Added 5px margin between the email content and the reply button  

### Notes
- Tested with various messages, looks fine so far  
- Needs more testing before marking as ready for review

### Before
https://github.com/user-attachments/assets/7aef0da7-65f9-4c2d-ae3d-21c34a3770db

### What I've done so far
https://github.com/user-attachments/assets/77cf37be-3a7a-4695-9fd4-744061f76eba

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -824,10 +824,10 @@ export default {
 	}
 }
 
-@media only screen and (max-width: var(--breakpoint-mobile)) {
+@media only screen and (max-width: 1024px) {
     #mail-thread-header {
         position: sticky !important;
-        top: 29px !important;;
+        top: 29px !important;
     }
 }
 
@@ -881,7 +881,7 @@ export default {
         padding-inline-start: 48px;
     }
 }
-@media only screen and (max-width: var(--breakpoint-mobile)) {
+@media only screen and (max-width: 1024px) {
 	#mail-thread-header-fields {
 		margin-top: -32px;
 	}


nimishavijay:
Looks really nice! :) thanks for cleaning up the sticky header and the bottom spaces in addition. I see that right now the participants are sticky while I think it would make more sense of the title is sticky, don't you think? 

So I would suggest some small changes:
- [ ] make the title sticky on scroll instead of the participants 
- [ ] for multi line titles ellipsize the title on scroll. So we are showing only ine line of the title on scroll, but to indicate that there is more we show the "..." Would that be possible?

CreatorOfBiontium:
Okay, sure, makes sense to me :) So on scroll, it will display the first line of the subject and the parcipiants below, like it does when not scrolling. Then I have to fix the issue with unsubscribe button overlapping actions menu and that should be all

CreatorOfBiontium:
Should be done! What do you think @ChristophWurst

src/components/Message.vue
```diff
@@ -225,6 +232,11 @@ export default {
 	&__notsuggested {
 		margin-inline-start: auto;
 	}
+
+	@media (max-width: 600px) {
```
ChristophWurst, line 236: I think there is a CSS var for the mobile breakpoint. Perhaps that would be a better option that the hard-coded 600px

src/components/MessageAttachments.vue
```diff
@@ -287,12 +287,16 @@ export default {
...
-	padding: 10px 6px 10px 46px;
+	padding: 10px 12px 10px 46px;
 	margin-top: 4px;
 	margin-bottom: 0;
 	position:sticky;
-	bottom:0;
+	bottom: -2px;
 	background: linear-gradient(0deg, var(--color-main-background), var(--color-main-background) 90%, rgba(255, 255, 255, 0));
+
+	@media (max-width: 600px) {
```
ChristophWurst, line 297: same

src/components/Thread.vue
```diff
@@ -219,9 +228,28 @@ export default {
 		window.addEventListener('keydown', this.handleKeyDown)
 	},
 
+	mounted() {
+		this.scrollEl = this.$el.closest('.app-content-wrapper') || window
```
ChristophWurst, line 232: nit: this gives me jquery vibes. is there any other way to achieve this?

src/components/Thread.vue
```diff
@@ -242,6 +270,150 @@ export default {
 			}
 		},
 
+		_measureHeightsOnce() {
+			const h2 = this.$el.querySelector('#mail-thread-header-fields h2')
```
ChristophWurst: use a vue ref

src/components/ThreadEnvelope.vue
```diff
@@ -1295,6 +1299,38 @@ export default {
...
 	}
 
+	.envelope__header {
+		display: flex;
+		align-items: center;
+
+		.left {
+			display: flex;
+			align-items: center;
+			min-width: 0;     /* umožní textu správně zkrátit */
```
ChristophWurst: přeložit nebo odstranit komentář

src/components/Message.vue
CreatorOfBiontium: There is the _--breakpoint-mobile: 1024px_ but 600-1024 px is wide enough to have an email displayed properly with the margin on the left. I'd keep it just for screens < 600px.

src/components/Thread.vue
CreatorOfBiontium: It's not jquery, it's native js.

$el is root element of the Vue component
.closest() looks for the closest parent

I'm not sure how else to do this, but I can take a look.

CreatorOfBiontium:
@ChristophWurst I made a few changes (like using ref instead of query selector) but some, like isMobile that you suggested isn't usable here, as we need to calculate weather to wrap the subject or not (it's not just phone or not). But still I have to make just a few small adjustements before merging - the main is when the subject is way too long (like 100 chars, and yes I received a subject this long for some reason) the animation keeps looping and the second one, uhh, I maybe, just maybe forgot to change 2 lines (in code) in the last commit back to what they should have been after testing something 😀

src/components/Thread.vue
ChristophWurst: Apologies that my message was confusing. I meant that the coding style reminded me of jQuery due to the element lookup by selector and direct DOM manipulation.

src/components/Message.vue
ChristophWurst: Fair point. But for consistency I'd suggest we use the same breakpoint as in other places. There is no downside of the reduced left margin at 600-1024px.

ChristophWurst (commented):
The responsive margin left works well. Let's get it in!

Let's leave out the header bar and its animation for now. I'm not fully convinced this is the best/only possible approach, and the animation is jittery right now.

CreatorOfBiontium:
Okay, so I'll create a new pr with just the margin fix. We then need to figure out the animation and the unsub button.

src/components/Message.vue
CreatorOfBiontium: Okay so I'll fix this else in the next commit or in the diffrent pr.

Issue #7750, opened 2022-12-08
Add invites to calendar automatically (tentative without respond)


sascha08-15:
### Is your feature request related to a problem? Please describe.

As a nextcloud mail and calendar user,
I want that incoming mails which include an invitation are automatically added to a default calendar (per Mail Account).
Such that I don't miss any incoming invite anymore.

### Describe the solution you'd like

There's two kind of mails I get with invites. One with .ics attachment, one with just the invite, mime type text/calendar. 
Both should be processed by mail and added to a calendar that is specifically set as default calendar for each mail account. 
MS Outlook processes events too. The behavior there is that it adds the invites to the calendar but does not trigger RSVP and leaves the item in the inbox, yet, it adds it to the calendar. Good default.

### Describe alternatives you've considered

Add every new invite to a dedicated calendar just for mail (unprocessed)

ChristophWurst:
> MS Outlook processes events too. The behavior there is that it adds the invites to the calendar but does not trigger RSVP and leaves the item in the inbox, yet, it adds it to the calendar. Good default.

Is there any filtering to prevent spammers entering your calendar?

Luncheon3462:
+1 for this request.

ChristophWurst:
Dear @Luncheon3462! Thank you for your contributions. A +1 comment causes a notification for all 46 who watch this repository, and everyone who is subscribed to the ticket. This is not helpful because we receive too many notifications already. Your affirmation is best expressed with a 👍 expression on https://github.com/nextcloud/mail/issues/7750#issue-1483925292 (like 14 people before you did). Hope this makes sense.

PR #11756, opened 2025-09-27
External invitation calendar

amirmkia1997:
Summary
This pull request fixes [issue #7750](https://github.com/nextcloud/mail/issues/7750) which reports that calendar invitations sent from external domains (e.g., Gmail) are not properly recognized or actionable inside the Nextcloud Mail app.

Background
Users receiving calendar event invitations from outside their Nextcloud domain experienced two key problems:

Missing Tentative Event: Invitations were not automatically added to the user’s Nextcloud Calendar as tentative events, unlike invitations originating from the internal domain.

Non-functional RSVP Buttons: The “Yes”, “No”, and “Maybe” buttons embedded in external invitations were not clickable, preventing users from responding directly within the Mail app.

Proposed Changes

Improved parsing of iCalendar/ICS attachments from external domains to ensure correct detection and automatic creation of tentative events in the user’s calendar.

Fixed the handling of RSVP action links to enable immediate responses (“Yes”, “No”, “Maybe”) from within the Mail app.

Added tests to cover common external providers (e.g., Gmail, Outlook) to prevent future regressions.

Testing

Verified invitations from multiple external providers now appear as tentative events upon receipt.

Confirmed that clicking any RSVP option updates the calendar status and sends the appropriate reply to the event organizer.

Confirmed that internal invitations continue to function as before.

Notes
This change improves interoperability with external email/calendar providers and ensures a consistent experience for users managing calendar events directly from the Mail app.

### DIFF
--- a/src/components/Imip.vue
+++ b/src/components/Imip.vue
@@ -157,6 +157,22 @@ function findAttendee(vEvent, email) {
 	return undefined
 }
 
+function findAttendeeByEmails(vEvent, emails) {
+	if (!vEvent || !Array.isArray(emails) || emails.length === 0) {
+		return undefined
+	}
+
+	const emailSet = new Set(emails.map(e => removeMailtoPrefix(e).toLowerCase()))
+	for (const attendee of [...vEvent.getPropertyIterator('ORGANIZER'), ...vEvent.getAttendeeIterator()]) {
+		const normalized = removeMailtoPrefix(attendee.email).toLowerCase()
+		if (emailSet.has(normalized)) {
+			return attendee
+		}
+	}
+
+	return undefined
+}
+
 export default {
 	name: 'Imip',
 	components: {
@@ -170,6 +186,10 @@ export default {
 	},
 	props: {
 		scheduling: {
+			type: Object,
+			required: true,
+		},
+		message: {
 			type: Object,
 			required: true,
 		},
@@ -197,6 +217,7 @@ export default {
 			currentUserPrincipalEmail: 'getCurrentUserPrincipalEmail',
 			clonedWriteableCalendars: 'getClonedWriteableCalendars',
 			currentUserPrincipal: 'getCurrentUserPrincipal',
+			accounts: 'getAccounts',
 		}),
 
 		/**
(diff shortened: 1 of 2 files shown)


src/components/Imip.vue
```diff
@@ -420,6 +493,22 @@ export default {
...
+					&& !this.wasProcessed
+					&& this.userIsAttendee
+					&& this.eventIsInFuture
+					&& this.existingEventFetched
+					&& !this.isExistingEvent
+				) {
+					await this.saveEventWithParticipationStatus(TENTATIVE)
+				}
+			} catch (e) {
+				// ignore auto-create failures
```
ChristophWurst: Log to the console for debuggability

nikbinmarjan:
Please fix this bug: We need to add multiple domains as trusted domains so they automatically display in the calendar, exactly like our main domain. This is important because invitation emails can get lost among many messages in the inbox, but when events appear in the calendar, nobody misses them.

y
```diff
@@ -0,0 +1,7 @@
+-----BEGIN OPENSSH PRIVATE KEY-----
```
ChristophWurst: :warning: drop this and revoke the key ASAP

src/components/Imip.vue
```diff
@@ -208,6 +229,14 @@ export default {
 			return this.scheduling.method
 		},
 
+		isFromDigikala() {
+			if (!this.message.from || !this.message.from[0]) {
+				return false
+			}
+			const fromEmail = this.message.from[0].email?.toLowerCase() || ''
+			return fromEmail.endsWith('@digikala.com')
+		},
```
ChristophWurst, line 238: This has to be generalized

ChristophWurst (changes requested):

src/components/Imip.vue
```diff
@@ -420,6 +493,22 @@ export default {
 		async decline() {
 			await this.saveEventWithParticipationStatus(DECLINED)
 		},
+		async autoCreateTentativeIfNeeded() {
```
kesselb, line 496: I'd say that approach isn’t ideal from a conceptual standpoint. Creating the tentative event during the mounting of the frontend component requires the user to open the email. It should be done in the imip background job and only for contacts listed as internal addresses or trusted sender.

PR #11758, opened 2025-09-27
Fix request type impl

amirmkia1997:
Summary
This pull request continues the work from [issue #6997](https://github.com/nextcloud/mail/issues/6997)
 by improving how incoming calendar invitations are processed and ensuring that event data is written to the correct calendar.
It also fixes a separate bug where certain incoming emails were not rendered completely in the Mail app.

Background
Currently, when a user receives an iTIP/IMip “REQUEST” message (calendar invitation), the system has no clear rule for selecting the target calendar. As a result, invitations may not appear automatically, and users must manually import them. In addition, some emails—especially those with complex multipart content—were displayed only partially, causing users to miss information.

Proposed Changes

Calendar Invitation Handling

Added an if clause in IMipService::process to check for messages of type REQUEST.

Implemented logic to:

Detect and use a user-defined default “invitation calendar” (stored in oc_preferences) or, if not set, fall back to the personal calendar.

Call ICreateFromString::createFromString to create the event in the chosen calendar.

Mark the message as processed, or as erroneous if storing fails.

Added a frontend settings field so users can configure their preferred invitation calendar.

Email Display Fix

Corrected message parsing to ensure full rendering of emails with multipart or complex MIME structures, resolving the issue where some messages were only partially shown.

Testing

Verified that invitations from both internal and external domains are correctly stored in the selected calendar and reflected in the Calendar app.

Tested fallback behavior when no default calendar is set.

Confirmed that previously problematic emails now display their entire content across major browsers.

Notes
This update streamlines the handling of calendar invitations, reduces manual steps for users, and improves reliability in email rendering.

### DIFF
--- a/src/components/Imip.vue
+++ b/src/components/Imip.vue
@@ -157,6 +157,22 @@ function findAttendee(vEvent, email) {
 	return undefined
 }
 
+function findAttendeeByEmails(vEvent, emails) {
+	if (!vEvent || !Array.isArray(emails) || emails.length === 0) {
+		return undefined
+	}
+
+	const emailSet = new Set(emails.map(e => removeMailtoPrefix(e).toLowerCase()))
+	for (const attendee of [...vEvent.getPropertyIterator('ORGANIZER'), ...vEvent.getAttendeeIterator()]) {
+		const normalized = removeMailtoPrefix(attendee.email).toLowerCase()
+		if (emailSet.has(normalized)) {
+			return attendee
+		}
+	}
+
+	return undefined
+}
+
 export default {
 	name: 'Imip',
 	components: {
@@ -170,6 +186,10 @@ export default {
 	},
 	props: {
 		scheduling: {
+			type: Object,
+			required: true,
+		},
+		message: {
 			type: Object,
 			required: true,
 		},
@@ -197,6 +217,7 @@ export default {
 			currentUserPrincipalEmail: 'getCurrentUserPrincipalEmail',
 			clonedWriteableCalendars: 'getClonedWriteableCalendars',
 			currentUserPrincipal: 'getCurrentUserPrincipal',
+			accounts: 'getAccounts',
 		}),
 
 		/**
@@ -206,6 +227,14 @@ export default {
 		 */
 		method() {
 			return this.scheduling.method
+		},
+
+		isFromDigikala() {
+			if (!this.message.from || !this.message.from[0]) {
+				return false
+			}
+			const fromEmail = this.message.from[0].email?.toLowerCase() || ''
+			return fromEmail.endsWith('@digikala.com')
 		},
 
 		/**
@@ -305,7 +334,7 @@ export default {
 		 * @return {boolean}
 		 */
 		userIsAttendee() {
-			return !!findAttendee(this.attachedVEvent, this.currentUserPrincipalEmail)
+			return !!findAttendeeByEmails(this.attachedVEvent, this.allUserEmails)
 		},
 
 		/**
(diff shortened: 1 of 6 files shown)


src/components/Imip.vue
```diff
@@ -208,6 +229,14 @@ export default {
 			return this.scheduling.method
 		},
 
+		isFromDigikala() {
+			if (!this.message.from || !this.message.from[0]) {
+				return false
+			}
+			const fromEmail = this.message.from[0].email?.toLowerCase() || ''
+			return fromEmail.endsWith('@digikala.com')
```
ChristophWurst: Could you please the purpose of this check?

amirmkia1997: We wanted to have a place to define trusted domains — so that whenever we receive a calendar event from one of those domains, it can automatically appear in the calendar as tentative.
In this case, this domain was just an example of one of our trusted sources.

ChristophWurst: While this fits your personal use case it needs a more generic approach to be usable for others too. How about changing this so it's based on trusted domains?

src/components/MessagePlainTextBody.vue
```diff
@@ -50,7 +50,7 @@ export default {
 	computed: {
 		enhancedBody() {
 			return this.body.replace(/(^&gt;.*\n)+/gm, (match) => {
-				return `<details class="quoted-text"><summary>${t('mail', 'Quoted text')}</summary>${match}</details>`
+				return `<details class="quoted-text" open><summary>${t('mail', 'Quoted text')}</summary>${match}</details>`
```
ChristophWurst, line 53: This appears to be unrelated, no?

src/components/Imip.vue
```diff
@@ -157,6 +157,22 @@ function findAttendee(vEvent, email) {
 	return undefined
 }
 
+function findAttendeeByEmails(vEvent, emails) {
```
kesselb, line 161: Hey, we recently changed the logic with https://github.com/nextcloud/mail/pull/11915. Please ensure to rebase your branch and also re-test if that is still necessary.

src/components/Imip.vue
amirmkia1997: Hi @ChristophWurst,

Thanks for the feedback! I’ve updated this part to make it more generic:

The trusted domains are now configurable via Admin Settings instead of being hardcoded.

The code reads the list of trusted domains from the settings, so any domain can be added and calendar events from those domains will appear automatically as tentative.

This makes the feature usable for all users, not just a single example domain.

ChristophWurst: The app already has trusted senders. That is the list of senders for which the app shows external images by default. Reuse that, please.

amirmkia1997: We tested the Trusted Senders feature, but there’s no button or visible option to add a trusted domain, even though we’re using the latest version of your app.

We also tried adding our domain (@digikala.com) directly to the database as admin, but that didn’t achieve what we’re aiming for — which is to have all emails from that domain automatically appear in the calendar.

It seems that the current trusted senders mechanism only affects image loading, not event handling.
Could you please guide us on how to properly achieve this behavior?

Thanks a lot for your help! 🙏

src/components/MessagePlainTextBody.vue
amirmkia1997: You're right, it's not directly related. I noticed that in some cases incoming emails were not fully displayed — parts of the message body were being cut off. I added this change to fix that rendering issue and ensure the full email content is shown properly.

src/components/Imip.vue
amirmkia1997: thanks

src/components/Imip.vue
ChristophWurst: Good point. You could add a UI option to edit trusted senders.

The auto-import of trusted senders is also something you have to add.

amirmkia1997: Screenshot 2025-10-25 at 14 58 30
Screenshot 2025-10-25 at 15 02 06 
As you can see in the screenshots, I’ve added the azarstan email from the database, and I can also see it in the panel under Trusted Senders.
However, when an email is sent from azarstan, it doesn’t get added to the calendar.

lib/Settings/AdminSettings.php
```diff
@@ -145,6 +145,12 @@ public function getForm() {
 			$this->classificationSettingsService->isClassificationEnabledByDefault(),
 		);
 
+		$this->initialStateService->provideInitialState(
+			Application::APP_ID,
+			'digikala_domain',
+			$this->config->getAppValue('mail', 'digikala_domain', '@digikala.com'),
```
SebastianKrupinski, line 151: Hard coded domain names and vendor specific variables, are not a good solution

src/components/Imip.vue
```diff
@@ -190,13 +211,15 @@ export default {
 			existingEventFetched: false,
 			targetCalendar: undefined,
 			comment: '',
+			digikalaDomain: loadState('mail', 'digikala_domain', '@digikala.com'),
```
SebastianKrupinski, line 214: Hard coded domain names and vendor specific variables, are not a good solution

src/components/Imip.vue
```diff
@@ -208,6 +231,14 @@ export default {
...
 		},
 
+		isFromDigikala() {
+			if (!this.message.from || !this.message.from[0]) {
+				return false
+			}
+			const fromEmail = this.message.from[0].email?.toLowerCase() || ''
+			return fromEmail.endsWith(this.digikalaDomain)
+		},
+
```
SebastianKrupinski, line 241: vendor specific variables and method names, are not a good solution

src/components/Imip.vue
amirmkia1997: @ChristophWurst

amirmkia1997: @ChristophWurst  "Can we set up a time to chat on a platform so that we can solve our problem?"

Issue #11317, opened 2025-06-26
Width of message list jumps after resizing


jancborchardt:
### Steps to reproduce

1. Have an existing install where you used Mail for some time but didn’t resize the message list
2. See that the text doesn’t take the proper amount of space (too much free space to the right)
3. Try to drag the width, the message list width jumps to be much wider

### Expected behavior

Regarding 2: Text should use the proper amount of space, and 3: the width should not jump on resizing.

### Actual behavior

Before resizing:
![Image](https://github.com/user-attachments/assets/1da9cff0-b510-44b3-b478-1926f84fe9f4)

After resizing (still a lot of whitespace on the right):
![Image](https://github.com/user-attachments/assets/a6101162-e893-41ad-95d1-22d9e2723eda)

### Mail app version

5.2.0-alpha.1 on daily, same on current cloud.nc

### Nextcloud version

32.0.0 dev, same on stable

### PHP engine version

None

### Web server

None

### Database

None

PR #11763, opened 2025-09-27
fix: mail pane resetting between layout changes

GVodyanov:
- [x] requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/7573

Fixes https://github.com/nextcloud/mail/issues/11317


kesselb (approved):

ChristophWurst:
I guess we can also merge without the `@nextcloud/vue` bump because there is nothing breaking here

PR #11785, opened 2025-10-02
fix: leftover design changes for the composer

GretaD:
fix pr fixes:

- close icon for the attachment should not be outlined
- border of ckeditor bar and the split of recipient info
- recipient info show more alighnment
-  the color of the right pane split

after
Screenshot from 2025-10-02 12-32-51

### DIFF
--- a/src/components/ComposerAttachment.vue
+++ b/src/components/ComposerAttachment.vue
@@ -34,7 +34,7 @@
 
 <script>
 import { generateUrl } from '@nextcloud/router'
-import Close from 'vue-material-design-icons/CloseOutline.vue'
+import Close from 'vue-material-design-icons/Close.vue'
 import Cloud from 'vue-material-design-icons/CloudOutline.vue'
 import EmailArrowRightIcon from 'vue-material-design-icons/EmailArrowRightOutline.vue'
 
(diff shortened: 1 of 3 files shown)


kesselb (approved):

GretaD:
> @GretaD nice! Regarding the borders (to the right pane, and around and in the ckeditor bar), shouldn't they be color-border rather than maxcontrast? Cause they are still a bit intense, and more oresent than the input fields.

`color-border`, is too dark compared to other borders we have in composer. So, i went with the closest one to `from` or `to` field
Screenshot from 2025-10-02 14-53-42

jancborchardt:
@GretaD it’s the same border color we use between subject and body, and it looks good as you show in the screenshot. :) If it’s too dark for accessibility, then we should adjust color-border instead. But `color-border` is the correct one to use here – color-text-maxcontrast is for text, not for decorations.

GretaD:
> @GretaD it’s the same border color we use between subject and body, and it looks good as you show in the screenshot. :) If it’s too dark for accessibility, then we should adjust color-border instead. But `color-border` is the correct one to use here – color-text-maxcontrast is for text, not for decorations.

ahh, i see, will it fix it asap. When we discussed it you mentioned that is very bright compare to `from` and `to` and i assumed you meant it should look like them. :)

PR #11793, opened 2025-10-06
refactor(rector): apply PHP74 set

ChristophWurst:


st3iny (approved):
In Rector we trust :pray:

ChristophWurst:
Psalm is unhappy

PR #11794, opened 2025-10-06
fix(text-block): preserve line breaks on plain text

### LINKED ISSUE #11729: Line breaks in text blocks are not inserted into mail
### Steps to reproduce

1. Create a text block with linebreaks in the mail app settings
2. insert the block into a new mail
3.

### Expected behavior

The text block is copied 1:1 like entered in the settings

### Actual behavior

The block is inserted without linebreaks.

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

hamza221:
fix #11729

### DIFF
--- a/src/ckeditor/smartpicker/InsertItemCommand.js
+++ b/src/ckeditor/smartpicker/InsertItemCommand.js
@@ -46,9 +46,11 @@ export default class InsertItemCommand extends Command {
 						const modelFragment = editor.data.toModel(viewFragment)
 						editor.model.insertContent(modelFragment)
 					} else {
-						const itemElement = writer.createElement('paragraph')
-						writer.insertText(item.content, itemElement)
-						editor.model.insertContent(itemElement)
+						const lines = item.content.split('\n')
+						const htmlContent = lines.map(line => `<p>${line}</p>`).join('')
+						const viewFragment = editor.data.processor.toView(htmlContent)
+						const modelFragment = editor.data.toModel(viewFragment)
+						editor.model.insertContent(modelFragment)
 					}
 				} else {
 					const itemElement = writer.createElement('paragraph')


src/ckeditor/smartpicker/InsertItemCommand.js
```diff
@@ -46,9 +46,11 @@ export default class InsertItemCommand extends Command {
 						const modelFragment = editor.data.toModel(viewFragment)
 						editor.model.insertContent(modelFragment)
 					} else {
-						const itemElement = writer.createElement('paragraph')
-						writer.insertText(item.content, itemElement)
-						editor.model.insertContent(itemElement)
+						const lines = item.content.split('\n')
```
ChristophWurst, line 49: Can you explain the logic behind this? I have created a test text block with both types of line breaks: a regular one and a paragraph. The database `content` is this: `<p style="margin:0;">Paragraph 1 Line 1<br>Paragraph 1 Line 2</p><p style="margin:0;">Paragraph 2 Line 1</p>`.

src/ckeditor/smartpicker/InsertItemCommand.js
```diff
@@ -46,9 +46,11 @@ export default class InsertItemCommand extends Command {
...
+						const htmlContent = lines.map(line => `<p>${line}</p>`).join('')
```
ChristophWurst, line 50: I'm skeptical about the use of paragraphs for line breaks. `<br>` and `</p><p>` is not the same semantically and also often looks different visually. You will have a regular line height between line breaks and more space between paragraphs.

hamza221: that's how CkEditor creates line breaks internally, we can reconfigure it to `<br>` https://ckeditor.com/docs/ckeditor4/latest/features/enterkey.html

src/ckeditor/smartpicker/InsertItemCommand.js
hamza221: The text is created in Ckeditor intially in the text block config so it will look like 

`<p> line one </p> <p> line 2</p>`
then on insert it will get converted to 
`line one\nline 2`
which here will get reconverted to 
`<p> line one </p> <p> line 2</p> `

src/ckeditor/smartpicker/InsertItemCommand.js
hamza221: I couldn't find docs for ckeditor5 but if you inspect composer and press enter you can see it live

src/ckeditor/smartpicker/InsertItemCommand.js
ChristophWurst: Test with this input in the text block:
1. <kbd>a</kbd>
2. <kbd>Shift+Enter</kbd>
3. <kbd>b</kbd>
4. <kbd>Enter</kbd>
5. <kbd>c</kbd>

Then you will have two paragraphs. The first one has a line break. Do you still have this format after converting to line breaks and back?

hamza221: yes that works, because the intermediate to Plain conversion would have `\n`for both paragraphs and `<br>`would look like this `a\nb\nc`

hamza221: visually it's the same, but it would only create `<p>` which is okay imo for plain text, because at the end of day the html wont be sent and visually it's exactly the same

ChristophWurst: I would expect `\n` for a `<br>` and `\n\n` for a `<p>`

hamza221: That would be a separate bug in https://github.com/nextcloud/mail/blob/fix/text-block-multiline/src/util/text.js#L96 
no?

ChristophWurst: Fair enough

ChristophWurst: Bildschirmfoto vom 2025-11-20 14-04-59

HTML works as expected. Paragraphs are preserved.

Bildschirmfoto vom 2025-11-20 14-05-28

Plain text is a mess. The semantics are lost.

I agree it's an issue with the text helper -> off topic

PR #11889, opened 2025-10-07
fix: ensure drafts folder exists

kesselb:


lib/Service/MailTransmission.php
```diff
@@ -258,6 +255,20 @@ public function saveLocalDraft(Account $account, LocalMessage $message): void {
 		$perfLogger->end();
 	}
 
+	private function findOrCreateDraftsMailbox(Account $account): Mailbox {
+		$draftsMailboxId = $account->getMailAccount()->getDraftsMailboxId();
+
+		if ($draftsMailboxId === null) {
+			return $this->mailManager->createMailbox(
+				$account,
+				'Drafts',
```
kesselb, line 264: We should probably translate it, but if the server doesn’t support storing the special-use flag, guessing the special use from the folder name will fail since we don’t know that “Entwürfe” means “Drafts”. So using the untranslated version is the safer option.

PR #11890, opened 2025-10-07
feat: show mail version in settings

kesselb:
image

1) Add mail version to app settings
2) Flipped keyboard shortcuts and about (for me about is always the last section and I've found it weird to have it otherwise in mail, however I'm fine to change it back)


ChristophWurst (approved):
Eslint is not happy but I am :heart_eyes_cat:

ChristophWurst:
> should maybe be a thing across all of our apps

Absolutely! It was even discussed to have this in a standardized component. Nothing happened yet so I'm okay with doing this for Mail only to kick it off

Issue #11931, opened 2025-10-14
Specified key was too long; max key length is 3072 bytes; migration 5006Date20250927130132 - mail_mailboxes.name


blizzz:
### Steps to reproduce

1. Upgrade to 5.5.9

### Expected behavior

Upgrade succeeds

### Actual behavior

> Exception: Database error when running migration 5006Date20250927130132 for app mail
An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 3072 bytes
Update failed

This is about modifying column `name` of `mail_mailboxes`. 

```json
{
  "reqId": "G5RuqTtIOPSh39aW2nCM",
  "level": 3,
  "time": "2025-10-14T21:49:36+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "updater",
  "method": "",
  "url": "--",
  "message": "\\OC\\Updater::failure: Exception: Database error when running migration 5006Date20250927130132 for app mail\nAn exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 3072 bytes",
  "userAgent": "--",
  "version": "31.0.9.1",
  "data": {
    "app": "updater"
  }
}
```

### Mail app version

5.5.9

### Nextcloud version

31.0.9

### PHP engine version

None

### Web server

None

### Database

MySQL 8.0.43 (yes, sorry)

csriebel08:
Could you explain the workaround or fix to this, I am not sure that I follow.  I am having the same issue with the mail app upgrade with basically the same specs as OP.

ChristophWurst:
Caused by https://github.com/nextcloud/mail/pull/11761

ChristophWurst:
@blizzz @csriebel08 could we compare our table definition? I wonder why the migration succeeded for my personal prod.

```sql
MariaDB [nextcloud]> show create table oc_mail_mailboxes \G;
*************************** 1. row ***************************
       Table: oc_mail_mailboxes
Create Table: CREATE TABLE `oc_mail_mailboxes` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(1024) NOT NULL,
  `account_id` int(11) NOT NULL,
  `attributes` varchar(255) NOT NULL DEFAULT '[]',
  `delimiter` varchar(1) DEFAULT NULL,
  `messages` int(11) NOT NULL,
[...]

PR #11938, opened 2025-10-15
fix(db): clean-up old mailboxes account_id+name index

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/11931

### DIFF
--- a/lib/Migration/Version0161Date20190902103701.php
+++ b/lib/Migration/Version0161Date20190902103701.php
@@ -75,11 +75,14 @@ class Version0161Date20190902103701 extends SimpleMigrationStep {
 			'default' => true,
 		]);
 		$mailboxTable->setPrimaryKey(['id']);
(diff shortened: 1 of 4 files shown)


blizzz (approved):

kesselb (approved):

blizzz:
> `UNIQ_22DEBD839B6B5FBA5E237E06` is the generated name for `oc_mail_mailboxes` .
> `UNIQ_45754FF89B6B5FBA5E237E06` is the generated name for `mail_mailboxes`.

Since it seems this is calculated somehow, can we just use the configured prefix (if any) and determine the expected name?

blizzz:
> Therefore it's hopefully enough to check for both. We should consider adding a note in our developer documentation to avoid adding indexes without an explicit name.

And perhaps enforce it in new code?

PR #11943, opened 2025-10-19
Enforce NOT NULL and Remove Default Value for mail_attachments.created_at Column

dhairyajangir:
Fixes https://github.com/nextcloud/mail/issues/5901

### Overview

This pull request addresses an important data integrity issue in the Nextcloud Mail app: previously, attachments could be created with a `created_at` timestamp of `0` or `NULL` due to a default value on the database column and missing assignment in the application logic. This made it difficult to reliably clean up old attachments and could lead to inconsistent data.

#### 1. **Database Migration**

- **File:** `lib/Migration/Version5007Date20251019000000.php`
- **Purpose:**  
  - Alters the `mail_attachments` table to ensure the `created_at` column is always set.
  - Sets the `created_at` column to `NOT NULL` and removes any default value.
  - This enforces that every new attachment must have an explicit creation timestamp set by the application.

#### 2. **How the Migration Works**

- Checks if the `mail_attachments` table exists.
- Checks if the `created_at` column exists.
- Alters the column to:
  - Require a value (`NOT NULL`)
  - Remove any default value (`default => null`)
- This ensures that the application must always provide a valid timestamp when inserting new attachments.

#### 3. **Why This Is Important**

- **Data Integrity:**  
  Prevents the creation of attachments with missing or invalid timestamps.
- **Cleanup Reliability:**  
  Accurate `created_at` values are essential for reliably cleaning up old attachments and managing storage.
- **Prevents Future Bugs:**  
  This change would have prevented the original bug by making it impossible to insert attachments with a default or missing timestamp.

#### 4. **How to Test**

1. **Run the migration:**  
   - The migration will update the schema as described.
2. **Add a new attachment:**  
   - Confirm that the `created_at` field is set to the current timestamp and is never `0` or `NULL`.
3. **Try inserting an attachment without a `created_at` value:**  
   - The database should reject the insert, enforcing the application logic.

#### 5. **Additional Notes**

- This migration is safe to run multiple times; it checks for the existence of the table and column before making changes.
- No existing data is deleted or modified, but any rows with a `created_at` of `0` or `NULL` should be reviewed separately if present.
- This change is a prerequisite for reliable cleanup jobs and future features that depend on accurate attachment timestamps.

---

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -34,7 +34,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>5.6.0-alpha.3</version>
+	<version>5.6.0-alpha.4</version>
 	<licence>agpl</licence>
 	<author homepage="https://github.com/ChristophWurst">Christoph Wurst</author>
 	<author homepage="https://github.com/GretaD">GretaD</author>


lib/Service/Attachment/AttachmentService.php
```diff
@@ -52,6 +53,11 @@ class AttachmentService implements IAttachmentService {
 	 */
 	private $logger;
 
+	/**
+	 * @var ITimeFactory
+	 */
+	private $timeFactory;
```
ChristophWurst: nitpick: you can use a php property type instead of the phpdoc annotation

dhairyajangir: sure :)

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
```
kesselb:
```suggestion
 * SPDX-License-Identifier: AGPL-3.0-or-later
```

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
...
+class Version5007Date20251019000000 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	#[\Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
```
kesselb:
```suggestion
```

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
...
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version5007Date20251019000000 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
```
kesselb, line 20:
```suggestion
	/**
	 * @param Closure(): ISchemaWrapper $schemaClosure
	 */
```

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version5007Date20251019000000 extends SimpleMigrationStep {
```
kesselb:
```suggestion
class Version5006Date20251019000000 extends SimpleMigrationStep {
```

Also don't forget to rename the file so class- and filename match.

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
```
kesselb, line 3: Please increase the `version` in `appinfo/info.xml` to `5.6.0-alpha.4` (otherwise the migration will not run).

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
...
+
+		if (!$schema->hasTable('mail_attachments')) {
+			return $schema;
+		}
+
+		$attachments = $schema->getTable('mail_attachments');
+
+		// Ensure created_at is NOT NULL and has no default, so app must set it.
+		if ($attachments->hasColumn('created_at')) {
+			$attachments->changeColumn('created_at', [
```
kesselb:
```suggestion
			$attachments->modifyColumn('created_at', [
```

changeColumn is deprecated; please use modifyColumn instead.

lib/Migration/Version5007Date20251019000000.php
```diff
@@ -0,0 +1,45 @@
...
+				'notnull' => true,
```
kesselb:
```suggestion
```

The column is created with notnull = true, we only want to change the default.

kesselb (commented):
Thank you :pray:

lib/Migration/Version5007Date20251019000000.php
dhairyajangir: sure :)

dhairyajangir: https://github.com/nextcloud/mail/pull/11943/commits/91ade169c9d907abc018cebffb49976836162dcb

PR #11945, opened 2025-10-20
perf(quick-actions): eliminate duplicate api calls for each envelope

### LINKED ISSUE #11746: Loading quick action steps triggers one xhr request per message/envelope
### Steps to reproduce

Open the mail app and observer that there's one xhr request for the quick actions steps per message/envelope,

### Expected behavior

The steps should be fetched once per account or once per user.

### Actual behavior

Steps are fetched once per envelope.

### Mail app version

main / stable5.5

hamza221:
fix #11746

### DIFF
--- a/lib/Service/QuickActionsService.php
+++ b/lib/Service/QuickActionsService.php
@@ -35,13 +35,24 @@ class QuickActionsService {
 		private ActionStepMapper $actionStepMapper,
 	) {
 	}
-
 	/**
 	 * @param string $userId
 	 * @return Actions[]
 	 */
 	public function findAll(string $userId): array {
-		return $this->actionsMapper->findAll($userId);
+		$actions = $this->actionsMapper->findAll($userId);
+		$actionIds = array_map(fn (Actions $action) => $action->getId(), $actions);
+		$actionSteps = $this->actionStepMapper->findStepsByActionIds($actionIds, $userId);
+		return array_map(function (Actions $action) use ($actionSteps) {
+			$steps = array_values(array_filter($actionSteps, function (ActionStep $step) use ($action) {
+				return $step->getActionId() === $action->getId();
+			}));
+			$action->setActionSteps($steps);
+			if (!empty($steps)) {
+				$action->setIcon($steps[0]->getName());
+			}
+			return $action;
+		}, $actions);
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


src/components/quickActions/Settings.vue
```diff
@@ -228,8 +228,8 @@ export default {
 				this.localAction = { id: null, name: '' }
 				this.actions = []
 			} else {
-				this.localAction = { ...action }
-				this.actions = await findAllStepsForAction(action.id)
+				this.localAction = { id: action.id, name: action.name, accountId: action.accountId }
```
ChristophWurst: is there no generic way to clone an object in vue? the manual destruct/construct is fragile. when we change properties of the object we have to keep this line in sync.

hamza221: we need specific properties from the Object

ChristophWurst: can you elaborate? what I see above we map three properties to identical properties.

hamza221: Sorry, I meant the object contains both quickActions and action steps to avoid multiple requests to the servers
so it looks like `{id, name, accountId, actionsSteps : [array of steps] }`
I split them back here so I don't break the CRUD logic for both 
If you look at the next line you will see that I'm also storing the steps in a local variable.

ChristophWurst: Right. But what I'm objecting to is that you are manually writing down the properties to copy to the local action. If a new property is added to it, you have to also update the code here. It's prone to errors.
I like the old approach better where you used the generic object destruct to take all properties of the action object.

lib/Service/QuickActionsService.php
```diff
@@ -35,13 +35,18 @@ public function __construct(
...
-
 	/**
 	 * @param string $userId
 	 * @return Actions[]
 	 */
 	public function findAll(string $userId): array {
-		return $this->actionsMapper->findAll($userId);
+		$actions = $this->actionsMapper->findAll($userId);
+		foreach ($actions as $action) {
+			$actionSteps = $this->actionStepMapper->findAllStepsForOneAction($action->getId(), $userId);
```
ChristophWurst: perf: *n+1 problem*

you could fetch the steps in bulk with a WHERE action_id IN (...)

PR #11957, opened 2025-10-21
fix: also await onToggleJunk

kesselb:
We should also await onToggleJunk ;)

I'm migrating to the more verbose if-else instead of the ternary, as the eslint 9 update will emit a warning for ternaries, when not using the result, and thus to keep it in sync with the lines below.


kesselb:
> Should we also migrate the consequent lines ?

I will do when updating eslint, as we don't use async/await there, we don't have any issue to fix.

hamza221 (approved):

PR #11962, opened 2025-10-23
fix: l10n fr traduction

koug44:
«Effacer le dossier» is unclear wether it deletes the folder or only the content. «Vider le dossier» is better traduction


kesselb:
Thank you for reporting :+1: 

We are using Transifex to manage translations. Please join https://explore.transifex.com/nextcloud/ to help. It's also possible to sign up with your GitHub account.

Translations are synced daily from Transifex to GitHub. Any change related to translation must be done there, or they are lost.

koug44:
Sure - I’ll have a look. Please close the MR

PR #11968, opened 2025-10-24
fix:Add Opt-out classification settings to account settings and setup

### LINKED ISSUE #11716: Offer classification opt-out at account setup for better transparency and control

As a Nextcloud Mail user I might not want Mail to classify my emails into important and not important. Right now I have to add an account and then turn classification off (classical opt-out). This assumes I'm aware of the feature and its toggle as a user, which is not realistic. Instead I will be surprised by the classification when it shows up on other clients.

Make the feature and an opt-out more transparent by placing a checked checkbox to the setup form. Users who are fine with the classification will learn about it there and not get surprised later on. Those who don't want the classification can turn it off from the start.

N/a

### Additional context

* https://github.com/nextcloud/mail/issues/5323

hamza221:
fix #11716

Todo: 

- [x]  Migrate old value ( from oc_preference in bulk )

### DIFF
--- a/tests/phpunit.unit.xml
+++ b/tests/phpunit.unit.xml
@@ -11,6 +11,7 @@
 			<directory suffix=".php">../lib</directory>
 		</include>
 		<exclude>
+			<directory suffix=".php">../mail/lib/Migration</directory>
 			<directory suffix=".php">../mail/lib/Vendor</directory>
 		</exclude>
 		<report>


lib/Migration/Version5006Date20251024153423.php
```diff
@@ -35,4 +40,33 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
+			->andWhere($qb->expr()->eq('configkey', $qb->createNamedParameter('tag-classified-messages')));
+
+		$res = $qb->executeQuery();
+
+		$falseUsers = [];
+		while ($row = $res->fetch()) {
+			if ($row['configvalue'] === 'false') {
+				$falseUsers[] = $row['userid'];
+			}
+		}
```
ChristophWurst: perf: move the false check into the db query

lib/Migration/Version5007Date20251024153423.php
```diff
@@ -35,4 +40,33 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
+		}
+
+		$res->closeCursor();
+
+		$qb = $this->db->getQueryBuilder();
+		$qb->update('mail_accounts')
+			->set('classification_enabled', $qb->createNamedParameter(false, IQueryBuilder::PARAM_BOOL))
+			->where($qb->expr()->in('user_id', $qb->createParameter('users')));
+		foreach (array_chunk($falseUsers, 1000) as $chunk) {
+			$qb->setParameter('users', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
```
ChristophWurst, line 70: please use the output to show the admins the process isn't stuck. only relevant for very big installations, but still relevant

lib/Migration/Version5007Date20251024153423.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	#[Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+		$accountsTable = $schema->getTable('mail_accounts');
+		if (!$accountsTable->hasColumn('classification_enabled')) {
+			$accountsTable->addColumn('classification_enabled', Types::BOOLEAN, [
+				'default' => true,
```
ChristophWurst, line 39: in previous migrations we also added notnull=>true. this could be important because Oracle treats null and false as the same thing. only by setting notnull=true you will get false back, else it will be null.

src/components/AccountForm.vue
```diff
@@ -37,6 +37,12 @@
 					:label="t('mail', 'Password')"
 					:required="!hasPasswordAlternatives"
 					@change="clearFeedback" />
+				<NcCheckboxRadioSwitch
+					id="auto-classification-enabled"
+					v-model="classificationEnabled"
+					:disabled="loading">
+					{{ t('mail', 'Enable classification') }}
```
ChristophWurst: this text …

src/components/AccountForm.vue
```diff
@@ -218,6 +224,12 @@
 					:disabled="loading"
 					required
 					@change="clearFeedback" />
+				<NcCheckboxRadioSwitch
+					id="auto-classification-enabled"
+					v-model="classificationEnabled"
+					:disabled="loading">
+					{{ t('mail', 'Enable mark as important classification') }}
```
ChristophWurst, line 231: … is different from this. is this intentional?

lib/Migration/Version5006Date20251024153423.php
```diff
@@ -0,0 +1,76 @@
...
+		$qb->update('mail_accounts')
+			->set('classification_enabled', $qb->createNamedParameter(false, IQueryBuilder::PARAM_BOOL))
+			->where($qb->expr()->in('user_id', $qb->createParameter('users')));
+		foreach (array_chunk($users, 1000) as $chunk) {
+			$output->advance();
+			$qb->setParameter('users', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
+			$qb->executeStatement();
+		}
+		$output->finishProgress();
+
```
ChristophWurst: preferences tag-classified-messages clean-up is missing

ChristophWurst (approved):
Sorry, another round

Looks good, tho :smiling_face_with_tear:

PR #12007, opened 2025-11-08
fix(emoji-picker): ignore if first character is a space

### LINKED ISSUE #11649: Emoji picker is hard to get rid of when surronding the colon ponctuation with spaces
### Steps to reproduce

Hi,

When typing a mail containing a colon, the emoji picker immediately pops up, which is a great addition (I guess as I'm old school and prefer formal correspondence, mail is not a chat platform). 

English sentence spacing conventions does not require a space *before* the colon punctuation, so the picker is kind of smart and does not trigger when adding a new line.

Example:
```
Please find attached: 
- file1
```
However, English convention [isn't the only one](https://en.wikipedia.org/wiki/Sentence_spacing), and other languages (namely French but probably others) suffers from this behavior. As such it's really hard to get rid of the picker when making the same list:
```
Merci de trouver ci-joint : 
- file 1
```
- Multiple spaces doesn't work
- `<enter>` insert the emoji
- navigation keys navigate the picker
- `<esc>` quit the whole mail

This simple feature makes the whole email writing unfortunately really annoying.

As a comparison, use the Github new issue modal and try the same two examples (`Please find attached: <enter>` and `Merci de trouver ci-joint : <enter>`). Then do the same exercise with the email text box in Nextcloud.

The only workaround I've found so far is to click elsewhere in the text and get back to my colon to continue writing. 

Best regards,

### Expected behavior

Emoji picker should disappear when the character following the colon is not part of a suggestion (a space character for example), or should be disable-able somewhere 

### Actual behavior

Emoji picker pops and stays until clicked somewhere else

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

hamza221:
fix #11649

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -226,6 +226,10 @@ export default {
 		},
 
 		getEmoji(text) {
+			// Disable the emoji picker if a [space] is the first character after the colon ':'
+			if (text[0] === ' ') {
+				return []
+			}
 			const emojiResults = emojiSearch(text)
 			if (this.textSmiles.includes(':' + text)) {
 				emojiResults.unshift(':' + text)


src/components/TextEditor.vue
```diff
@@ -226,6 +226,9 @@ export default {
 		},
 
 		getEmoji(text) {
+			if (text[0] === ' ') {
```
ChristophWurst, line 230: Could you leave code comment? I think the next dev looking at the code will not get why this condition exists

PR #12030, opened 2025-11-14
fix: add the warning icon when account cannot connect

GretaD:
Continuation of https://github.com/nextcloud/mail/pull/12027

sorry, i didnt push the last fixup with the icon, before creating the pull request

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -26,9 +26,17 @@
 					:is-first="isFirst(group.account)"
 					:is-last="isLast(group.account)"
 					:is-disabled="isDisabled(group.account)" />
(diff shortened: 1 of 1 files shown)


src/components/Navigation.vue
```diff
@@ -235,5 +245,10 @@ to {
 .v-popper__inner {
 	height: unset !important;
 }
+.error-message {
```
ChristophWurst: the usual remark

GretaD: i was not sure if it was the BEM, or the space above, so i did both

ChristophWurst: yes

Issue #11998, opened 2025-11-05
Improved wording and structure for folder actions


jancborchardt:
### Is your feature request related to a problem? Please describe.

Some of the folder actions area unclear, could get wording improvements, or don’t seem to be important enough to warrant direct display in the menu:

<img width="443" height="578" alt="Image" src="https://github.com/user-attachments/assets/44c560c2-86f0-4760-b44c-7d87c3acd27b" />

### Describe the solution you'd like

- **Folder name and unread count:** This is good
- **Mark all as read:** Subline duplicates info of the main line. "Mark all as read" is understandable and also used by other mail apps like Gmail
- **Add subfolder:** Good wording, but better icon would be a simple "+"
- **Rename:** Good wording, but better icon would be a simple edit pencil like we use elsewhere for renaming
- **Clear folder:** The wording here could be clearer, like "Delete all messages". And also the icon could be the trashbin icon we normally use for deletion, as the "eraser" is the one and only time we use it in all of Nextcloud.

And the very confusing/technical ones which we could either remove, automate, or put in a submenu:
- **Repair folder:** This makes it look like it’s broken all the time. When clicked, nothing seems to happen, and it still shows up. When is this needed and what does it do?
- **Subscribed:** What does this mean? Why would it be disabled?
- **Sync in background:** Wouldn’t I want this enabled for all folders?

PR #12039, opened 2025-11-17
fix(NavigationMailbox): reword

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/11998

| Before | After |
| - | - |
|image| swappy-20251117_124106 |

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -109,13 +109,12 @@
 				<template #icon>
 					<IconEmailCheck :size="20" />
 				</template>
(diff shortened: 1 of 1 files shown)


src/components/NavigationMailbox.vue
```diff
@@ -827,4 +824,8 @@ export default {
 .counter-bubble__counter {
 	max-width: initial;
 }
+
+:deep(.action-item__menutoggle) {
```
ChristophWurst, line 828: isn't this an upstream bug then?

ChristophWurst: If this is a general bug in nc-vue then please send a fix there

PR #12043, opened 2025-11-17
feat: manually refresh folder list

hamza221:
With this change when pressing the message refresh button it also refreshes folders


hamza221:
rebased to fix Ci && conflicts

PR #12031, opened 2025-11-14
Settings redesign to use new vue components

GVodyanov:
Fix https://github.com/nextcloud/mail/issues/11910

swappy-20251115_000130
swappy-20251115_000150
swappy-20251115_000159

And with mailvelope enabled:
swappy-20251115_000239

PR #12046, opened 2025-11-18
fix(Settings): followup on redesign

GVodyanov:
Address issues from followup on Settings redesign https://github.com/nextcloud/mail/pull/12031#pullrequestreview-3476131039

What this PR solves
- [x] Plus icons are missing from <kbd>+ New text block</kbd> and <kbd>+ Add internal address</kbd>
- ~The <kbd>+ Add mail account</kbd> button has rounded top corners, as opposed to on the mockup.~ - Will be done in Nextcloud Vue
- ~The text in lists is not aligned with that of form components cc ShGKme (how) can this be done?~ - Will be done in Nextcloud Vue
- [x] There are wording discrepancies that I outlined
- [x] > The icon on the account settings feels like a link file it acts like a button (it changes the settings dialog content) 
The current dialog content? Originally, it opened a separate dialog. If so, an icon like arrow_forward can be used to indicate navigation instead.

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -28,7 +28,7 @@
 							:aria-label="t('mail', 'Account settings')"
 							@click="openAccountSettings(account.id)">
 							<template #icon>
-								<IconLink :size="20" />
+								<IconArrow :size="20" />
 							</template>
 							{{ account.emailAddress }}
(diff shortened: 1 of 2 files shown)


kra-mo (approved):

GVodyanov:
image

image

image

image


image

PR #12056, opened 2025-11-19
feat: improve search filter buttons

### LINKED ISSUE #11798: Improved quick filters

Our current quick filters are the following:
- **Has attachment:** Is nice, we should keep this
- **Last 7 days:** Just truncates the list as we have sorted by date by default, is not really useful
- **From me:** Not useful in most of the inboxes. @GretaD wrote in https://github.com/nextcloud/mail/issues/9182#issuecomment-1991362440 that Outlook has "Sent to me, or cc: me", which is much more useful

Image

Better quick filters would be:
- **Has attachment:** We already have it
- **Unread:** Is available in some other apps as well, and we have it in the search details already
- **To me**: Like Outlook – question is if they include cc?

We could also include "Priority" but that won’t fit in the bar, and also we already have the priority inbox. Similar for "Favorites".

### Additional context

_No response_

GretaD:
fixes #11798

used to be:
Screenshot from 2025-11-21 12-20-06

- Has attachment
- Last 7 days
- From me

now:
Screenshot from 2025-11-21 12-19-34

- Has attachment
- Unread
- To me

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,7 +121,6 @@
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
       "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
       "dev": true,
-      "peer": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.5",
         "@jridgewell/trace-mapping": "^0.3.24"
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -121,7 +121,6 @@
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
       "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
       "dev": true,
-      "peer": true,
```
ChristophWurst, line 124: if the lock file format file changed feel free to update and check it into git, but not as part of the feature.

PR #12058, opened 2025-11-19
fix(ui): improve color contrast of tags

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/5149

|   | List | Modal |
|--------|--------|--------|
| Before | Bildschirmfoto vom 2025-11-19 16-43-24 | Bildschirmfoto vom 2025-11-19 16-43-38 |
| Dark | Bildschirmfoto vom 2025-11-19 16-40-12 | Bildschirmfoto vom 2025-11-19 16-40-06 |
| Bright | Bildschirmfoto vom 2025-11-19 16-41-05 | Bildschirmfoto vom 2025-11-19 16-41-28 |

### DIFF
--- a/src/components/Tag.vue
+++ b/src/components/Tag.vue
@@ -3,6 +3,18 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
+<template>
+	<div class="tag-group" :class="{ 'tag-group--dark': isDark, 'tag-group--bright': !isDark }">
+		<div
+			class="tag-group__bg"
+			:style="{ 'background-color': tag.color }" />
+		<span
+			class="tag-group__label">
+			{{ translateTagDisplayName(tag) }}
+		</span>
+	</div>
+</template>
+
 <script setup>
 import { defineProps } from 'vue'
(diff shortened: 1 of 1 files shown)


ChristophWurst:
The rounding is a side effect of deduplicating four copies of the more or less same code for tags 🌝

The tags will look like this with the suggested CSS:

Bildschirmfoto vom 2025-11-20 14-53-39

This works in terms of text readability but we'll lose most of the color

src/components/Tag.vue
```diff
@@ -0,0 +1,106 @@
...
+import { translateTagDisplayName } from '../util/tag.js'
+
+const props = defineProps({
+	tag: {
+		type: Object,
+		required: true,
+	},
+})
+
+// Source - https://stackoverflow.com/a
```
kesselb, line 29:
```suggestion
// Source - https://stackoverflow.com/a/9733420
```

nimishavijay:
You're right, but design wise this seems to be in the right direction (ref outlook as well) 
image

How about 
background: `color-mix(in srgb, [tag color] 40%, var(--color-main-background) 100%)`
This should make it brighter in the background

Issue #10384, opened 2024-11-21
Opening messages is slow


ChristophWurst:
### Steps to reproduce

1. Open the app
2. Click on a message
3. Click on another message

### Expected behavior

Fast switching

### Actual behavior

Mediocre to slow switching. HTML messages are worse than plain text messages.

### Mail app version

4.0

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
Thank you for the feedback, @9-FS. Could you please open the browser developer console, switch to the network tab, reload the page and try to find the specific requests which are slow, if any? That should help narrowing it down.

solderdev:
FYI - had the same problem both with Nextcloud Mail AND Snappymail.
The cause was a bad DNS configuration:
* Nextcloud running in Docker
* local DNS server (Technitium/Unbound/pihole) running in Docker on the same machine
* Docker internal network: 172.16.0.0
* LAN-IP: 192.168.1.10 --> the local router sets this (server IP) as DNS for all clients in LAN
* Server host-OS and therefore Docker daemon gets it's own IP assigned as DNS server (via DHCP)
* DNS lookup of mailserver-URL from Nextcloud via 192.168 IP fails (something about docker networking)

Solution:
in the docker-compose file of the DNS server, make sure port 53 is bound to the correct host-IP:
```
ports:
      - 192.168.1.10:53:53/udp
      - 192.168.1.10:53:53/tcp
```

Could you please check if you run a similar setup?
Easy test: set s/t like 1.1.1.1 as host-OS DNS and restart Docker daemon.

9-FS:
@ChristophWurst Unfortunately, it's incosistent. The requests that take multiple seconds all have in common that they are all JSON data from Google Mail and are not displayed in a pure text fashion. In contrast, the mails I have imported from my university server seem to be text only and load instantaneously.

@solderdev This is a pretty similar setup to mine, I just have my Technitium DNS server in its own network `br0` under `192.168.0.3` and my router advertises that IP to all devices in the network as DNS server. From the console of other containers like Nextcloud, that are in the default `bridge` network, I've tried some `nslookup` and `ping 192.168.0.3` and they were all instantaneous. I post my docker composes for Technitium and Nextcloud down here just in case it might help.

```
services:
    technitium:
        container_name: "technitium"
        image: "technitium/dns-server:13.3.0"
        environment:
            HOST_OS: "Unraid"
            TZ: "UTC"
[...]

PR #12082, opened 2025-11-25
fix(imap): cache message body

kesselb:
Should help with https://github.com/nextcloud/mail/issues/10384

For HTML messages we already know that after the "body" request there is also an "html" request which also needs the body part, so keeping them in cache makes it a bit faster.

I went with distributed because even in these cases I would expect a small improvement. We skip the usual IMAP handshake and just fetch what we need.

### DIFF
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -45,6 +45,8 @@ use OCP\Files\Folder;
 use OCP\Files\GenericFileException;
 use OCP\Files\IMimeTypeDetector;
 use OCP\Files\NotPermittedException;
+use OCP\ICache;
+use OCP\ICacheFactory;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\IURLGenerator;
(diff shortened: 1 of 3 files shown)


lib/Controller/MessagesController.php
```diff
@@ -1001,4 +1021,8 @@ private function attachmentIsImage(array $attachment): bool {
 	private function attachmentIsCalendarEvent(array $attachment): bool {
 		return in_array($attachment['mime'], ['text/calendar', 'application/ics'], true);
 	}
+
+	private function getCacheForAccount(int $accountId): ICache {
+		return $this->cacheFactory->createDistributed('mail_account_' . $accountId);
```
ChristophWurst, line 1026: Missing check for cache availability?

kesselb: The check for cache availability is not missing. createDistributed might return a null cache instance; that's fine and we just go ahead and fetch the part.

ChristophWurst:
Okay, got it. Cache is definitely faster than IMAP :+1: As usual I would just consider the local cache due to lower latency.

ChristophWurst (approved):

Issue #11064, opened 2025-05-05
Pin email/thread


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Nextcloud Mail user I want specific messages to not move out of sight.

### Describe the solution you'd like

Allow me to pin a message or thread, which will make it stay on top of the thread list.

In a technical sense we will sort by pinned first, then the actual sort attribute (date).

### Describe alternatives you've considered

N/a

### Additional context

N/a

Sev6n:
It would make sense to implement this, as it is present in Microsoft Outlook. This screenshot is from the web application.

![Image](https://github.com/user-attachments/assets/5d2ca3c5-0ce2-4584-8288-f9404c4fcdd6)

ChristophWurst:
What I'm unsure about is interoperability. There are standard IMAP flags like *seen*, *flagged*, *junk*, *deleted*, and there are some common perflags `$label1` to `$label5` (important, later, work, etc). But nothing for *pinned* as far as I know. So pinning would only show in Mail, and not in any other client connected to the same IMAP server.

@Sev6n do you also have a non-microsoft email account inside Outlook maybe? I wonder if they still offer pinning, and if they write that to IMAP.

ChristophWurst:
Thanks for checking!

PR #12111, opened 2025-12-01
feat: pin favorites

hamza221:
part of: [#11064](https://github.com/nextcloud/mail/issues/11064)

### DIFF
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -29,17 +29,85 @@
 					role="heading"
 					:aria-level="2"
 					@shortkey.native="onShortcut">
-					<Mailbox
-						v-if="!mailbox.isPriorityInbox"
-						:account="account"
-						:mailbox="mailbox"
-						:search-query="query"
-						:bus="bus"
-						:open-first="mailbox.specialRole !== 'drafts'"
-						:group-envelopes="groupEnvelopes"
-						:initial-page-size="messagesOrderBydate"
-						:collapsible="true" />
+					<template v-if="!mailbox.isPriorityInbox">
+						<div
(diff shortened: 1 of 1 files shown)


src/components/MailboxThread.vue
```diff
@@ -192,6 +260,7 @@ export default {
 		return {
 
 			importantInfo: t('mail', 'Messages will automatically be marked as important based on which messages you interacted with or marked as important. In the beginning you might have to manually change the importance to teach the system, but it will improve over time.'),
+			FavoritesInfo: t('mail', 'Messages that your mark as favorite will be shown at the top of folders, you can disable this behavior in the app settings'),
```
kesselb:
```suggestion
			favoritesInfo: t('mail', 'Messages that your mark as favorite will be shown at the top of folders, you can disable this behavior in the app settings'),
```

ChristophWurst:
```suggestion
			FavoritesInfo: t('mail', 'Messages that your mark as favorite will be shown at the top of folders. You can disable this behavior in the app settings.'),
```

grammar: these are two sentences

src/components/MailboxThread.vue
```diff
@@ -368,6 +448,8 @@ export default {
 	},
 
 	async mounted() {
+		// TODO: check the user preference
```
ChristophWurst, line 447: to do when? :smile:

ChristophWurst (approved):

src/components/MailboxThread.vue
hamza221: In the separate setting Pr, wip rn

PR #12114, opened 2025-12-01
add unit tests for UserDeletedListener functionality

printminion-co:
Added comprehensive unit tests for the UserDeletedListener to ensure proper handling of user deletion events, including scenarios with no accounts, single and multiple accounts, client exceptions, and partial failures. This enhances test coverage and reliability of the mail account deletion process.

In order for us to extend the logic in the future

### DIFF
--- a/tests/Unit/Listener/UserDeletedListenerTest.php
+++ b/tests/Unit/Listener/UserDeletedListenerTest.php
@@ -1,0 +1,190 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Tests\Unit\Listener;
+
+use ChristophWurst\Nextcloud\Testing\TestCase;
+use OCA\Mail\Account;
+use OCA\Mail\Db\MailAccount;
+use OCA\Mail\Exception\ClientException;
+use OCA\Mail\Listener\UserDeletedListener;
+use OCA\Mail\Service\AccountService;
+use OCP\EventDispatcher\Event;
+use OCP\IUser;
+use OCP\User\Events\UserDeletedEvent;
+use PHPUnit\Framework\MockObject\MockObject;
+use Psr\Log\LoggerInterface;
+
+class UserDeletedListenerTest extends TestCase {
+	private AccountService&MockObject $accountService;
+	private LoggerInterface&MockObject $logger;
+	private UserDeletedListener $listener;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->accountService = $this->createMock(AccountService::class);
+		$this->logger = $this->createMock(LoggerInterface::class);
+
+		$this->listener = new UserDeletedListener(
+			$this->accountService,
+			$this->logger
+		);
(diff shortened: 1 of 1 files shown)


tests/Unit/Listener/UserDeletedListenerTest.php
```diff
@@ -0,0 +1,199 @@
...
+		$account = $this->createMock(Account::class);
+		$account->method('getId')->willReturn($id);
+		return $account;
+	}
+
+	public function testImplementsIEventListener(): void {
+		$this->assertInstanceOf(\OCP\EventDispatcher\IEventListener::class, $this->listener);
+	}
+
+	public function testHandleMethodHasOverrideAttribute(): void {
```
kesselb: Why would you have a test for that?

tests/Unit/Listener/UserDeletedListenerTest.php
```diff
@@ -0,0 +1,199 @@
...
+		$this->accountService = $this->createMock(AccountService::class);
+		$this->logger = $this->createMock(LoggerInterface::class);
+
+		$this->listener = new UserDeletedListener(
+			$this->accountService,
+			$this->logger
+		);
+	}
+
+	private function createUserMock(string $userId = 'test-user'): IUser&MockObject {
```
kesselb:
```suggestion
	private function createUserMock(string $userId): IUser&MockObject {
```

Not using a default makes the test easier to read because the value is given when used.

tests/Unit/Listener/UserDeletedListenerTest.php
```diff
@@ -0,0 +1,199 @@
...
+		$user = $this->createMock(IUser::class);
+		$user->method('getUID')->willReturn($userId);
+		return $user;
+	}
+
+	private function createAccountMock(int $id): Account&MockObject {
+		$account = $this->createMock(Account::class);
```
kesselb: Account is mostly an DTO. Nicer would be, because we are not working with a test double, to instantiate a new object and only mock the MailAccount object.

tests/Unit/Listener/UserDeletedListenerTest.php
printminion-co: In this commit
https://github.com/nextcloud/mail/commit/ed9c2cc92c39bec34ff581426d2e806f1efaf0c1#diff-cf5b1bf33773bd1c2479f1d55e2fa15a16cd2095711a62454ef94b08f59c4fa4R32-R35
the `#[\Override]` was introduced.  I am sure it was important otherwise you would not have it in codebase.
 
Now if one removes the `#[\Override]` in the source file the test will fail signaling  the file editor - do not touch this.

kesselb: https://stitcher.io/blog/override-in-php-83

Please remove the test case. Psalm, our static analysis tool, will fail if the override attribute is missing.

printminion-co: done

tests/Unit/Listener/UserDeletedListenerTest.php
printminion-co: done

tests/Unit/Listener/UserDeletedListenerTest.php
printminion-co: done

PR #12115, opened 2025-12-01
feat: add occ command to list mail accounts

SebastianKrupinski:
### Summary
- Added command to list mail accounts

### Usage
List for single user
occ mail:account:list 'user name'

List for all users
occ mail:account:list --all

List with full details
occ mail:account:list --all --full

### Output

List all users
image

List single user
image


SebastianKrupinski:
> With the exception of a _list all accounts on the system_ this duplicates [`main`/lib/Command/ExportAccount.php](https://github.com/nextcloud/mail/blob/main/lib/Command/ExportAccount.php) more or less

Yeah I agree, I didn't realize the export command displayed the information. We can kill it, I didn't spent that much time on it... or we can also add the ability to search for specific values, which might be handy... Like list account that not using SSL, or using specific ports... etc... it wouldn't be much effort

ChristophWurst:
For these specific queries one can always resort to the database. I don't think it's worth developing and maintaining a command for this.

Issue #10475, opened 2024-12-10
Check coverage matrix


kesselb:
**To do:** Check if we could reduce the coverage matrix. 

Why? The CI runs with coverage are usually a bit slower, and it takes some time for Codecov to process the results. 

The initial reason to look into this topic were some issues when updating the Codecov action, leading to a "flanky" coverage score. This was resolved by settings the working-directory for the action: https://github.com/nextcloud/mail/pull/9401#issuecomment-2422269830

Example: https://app.codecov.io/gh/nextcloud/mail/commit/63f31a2009e3c0a4aef9e56b9a471ddcaf40ea26

![Image](https://github.com/user-attachments/assets/10f3c29c-cf93-4763-9eba-c31e6c5aa241)

Workflow: https://github.com/nextcloud/mail/blob/63f31a2009e3c0a4aef9e56b9a471ddcaf40ea26/.github/workflows/test.yml

**Unit:**

https://github.com/nextcloud/mail/blob/63f31a2009e3c0a4aef9e56b9a471ddcaf40ea26/.github/workflows/test.yml#L13-L18

The coverage is calculated if php = 8.3
=> 2x coverage

**Integration:**

https://github.com/nextcloud/mail/blob/63f31a2009e3c0a4aef9e56b9a471ddcaf40ea26/.github/workflows/test.yml#L64-L81

The coverage is calculated if db = mysql
=> 4x coverage

PR #12125, opened 2025-12-02
ci: less coverage runs

kesselb:
Fix https://github.com/nextcloud/mail/issues/10475

This PR:

- unit
- integration with redis
- integration without cache

Main:

- 2x unit (if php8.3)
- 4x integration (if mysql)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,20 +17,23 @@ jobs:
         php-versions: ['8.2', '8.3']
         nextcloud-versions: ['master', 'stable31']
         include:
+          - php-versions: '8.4'
+            nextcloud-versions: 'master'
+            coverage: true
           - php-versions: '8.1'
             nextcloud-versions: 'stable32'
           - php-versions: '8.4'
             nextcloud-versions: 'stable32'
           - php-versions: '8.5'
             nextcloud-versions: 'master'
-    name: Nextcloud ${{ matrix.nextcloud-versions }} php${{ matrix.php-versions }} unit tests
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -74,6 +70,11 @@ jobs:
               db: ['sqlite', 'mysql', 'pgsql']
               cache: ['nocache']
               include:
+                - php-versions: 8.4
+                  nextcloud-versions: 'master'
+                  db: 'mysql'
+                  cache: 'redis'
```
kesselb: Coverage decrases when running integration tests without cache.

.github/workflows/test.yml
```diff
@@ -17,20 +17,23 @@ jobs:
...
-    name: Nextcloud ${{ matrix.nextcloud-versions }} php${{ matrix.php-versions }} unit tests
+    name: Nextcloud ${{ matrix.nextcloud-versions }} php${{ matrix.php-versions }} unit tests ${{ matrix.coverage && '(coverage)' || ''}}
     steps:
     - name: Set up Nextcloud env
       uses: ChristophWurst/setup-nextcloud@fc0790385c175d97e88a7cb0933490de6e990374 # v0.3.2
       with:
         nextcloud-version: ${{ matrix.nextcloud-versions }}
         php-version: ${{ matrix.php-versions }}
-        php-coverage: 'xdebug'
+        php-coverage: ${{ matrix.coverage && 'xdebug' || 'none' }}
```
ChristophWurst, line 36: > GitHub provides a way to create conditional logic in expressions using binary logical operators (&& and ||). This pattern can be used to achieve similar functionality to the ternary operator (?:) found in many programming languages, while actually using only binary operators.

TIL

PR #12135, opened 2025-12-04
feat: compact view for mail layouts

### LINKED ISSUE #9713: Add option to disable body preview in list pane / offer "compact" format

It's not a "problem" but IMHO the preview is a waste of space. The first line of an email usually is "Dear <<recipient>>", nowadays often followed by a lot of blah before people get to the point. I don't think having a preview adds a lot of value. Additionally it seems to have problems with some encodings.

Screenshot_20240606_090254

Screenshot_20240606_090319

I would like to see an option to turn off the preview line so that only the sender and the subject are shown. Maybe even a compact mode like in the original Horde webmail client or in Roundcube, with only sender, subject and no avatar.

Only alternative is using another MUA

### Additional context

_No response_

GretaD:
fixes #9713 

Screenshot from 2026-01-09 11-24-37

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -79,6 +79,13 @@
 					</NcRadioGroupButton>
 				</NcRadioGroup>
 
+				<NcFormBox>
+					<NcFormBoxSwitch
+						v-model="compactMode"
+						:label="t('mail', 'Use compact mode')"
+						@update:modelValue="compactMode = $event" />
+				</NcFormBox>
+
 				<NcRadioGroup :model-value="sortOrder" :label="t('mail', 'Sorting')" @update:modelValue="onSortByDate">
 					<NcRadioGroupButton :label="t('mail', 'Newest first')" value="newest" />
 					<NcRadioGroupButton :label="t('mail', 'Oldest first')" value="oldest" />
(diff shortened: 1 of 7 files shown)


nimishavijay:
> The black bullet point in the beginning of the row its not an unread message, but a bullet point instead of the avatar :) I did it like that because thats how roundcube has it. I can do it primary, but maybe it will be confusing with the unread market that is shown in the right.

Oh I see! In that case it is definitely confusing 🙈 Would it be possible to show the multiselect checkbox all the time instead of the bullet? 
image

It can be color-maxcontrast in order to make sure it doesn't stand out too much in the screen, and when you select an email, the checkbox-filled icon should be primary.
Nice to have: once you select an email they can all change to primary to show we "entered" into selection mode

image

What do you think? :)

GretaD:
nimisha is sick and i dont have a mockup how important or favorite should look like, but i imporvized something like gmail does
Screenshot from 2026-01-08 19-43-59

src/components/AppSettingsMenu.vue
```diff
@@ -79,6 +79,13 @@
 					</NcRadioGroupButton>
 				</NcRadioGroup>
 
+				<NcFormBox>
+					<NcFormBoxSwitch
+						v-model="compactMode"
+						:label="t('mail', 'Set compact mode')"
```
ChristophWurst: grammar: should it be `Use compact mode`?

PR #12144, opened 2025-12-08
feat: automated appointment creation

### LINKED ISSUE #7750: Add invites to calendar automatically (tentative without respond)

SebastianKrupinski:
### Summary
- Resolves: https://github.com/nextcloud/mail/issues/7750
- Requires: https://github.com/nextcloud/server/pull/56924
- Added option in account setting to enable feature per account
- Modified imip service to use new imip handler

### Testing
- Navigate to mail account settings, turn on "Automatically create tentative appointments"
- Using another external account/service send a calendar invitation to the configure nc mail account
- Trigger cron job to run background jobs

image

### Documentation
https://github.com/nextcloud/documentation/pull/14009

### DIFF
--- a/lib/Migration/Version5007Date20251208000000.php
+++ b/lib/Migration/Version5007Date20251208000000.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version5007Date20251208000000 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+		$accountsTable = $schema->getTable('mail_accounts');
+		if (!$accountsTable->hasColumn('imip_create')) {
+			$accountsTable->addColumn('imip_create', Types::BOOLEAN, [
+				'default' => false,
+				'notnull' => false,
+			]);
+		}
+		return $schema;
+	}
+}
--- a/lib/Service/IMipService.php
+++ b/lib/Service/IMipService.php
@@ -16,6 +16,7 @@ use OCA\Mail\Db\Message;
 use OCA\Mail\Db\MessageMapper;
 use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\Model\IMAPMessage;
+use OCA\Mail\Util\ServerVersion;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\Calendar\IManager;
 use Psr\Log\LoggerInterface;
@@ -30,6 +31,7 @@ class IMipService {
 	private MailboxMapper $mailboxMapper;
 	private MailManager $mailManager;
 	private MessageMapper $messageMapper;
+	private ServerVersion $serverVersion;
 
 	public function __construct(
 		AccountService $accountService,
@@ -38,6 +40,7 @@ class IMipService {
 		MailboxMapper $mailboxMapper,
 		MailManager $mailManager,
 		MessageMapper $messageMapper,
+		ServerVersion $serverVersion,
 	) {
 		$this->accountService = $accountService;
 		$this->calendarManager = $manager;
@@ -45,6 +48,7 @@ class IMipService {
 		$this->mailboxMapper = $mailboxMapper;
 		$this->mailManager = $mailManager;
 		$this->messageMapper = $messageMapper;
+		$this->serverVersion = $serverVersion;
 	}
 
 	public function process(): void {
@@ -115,8 +119,10 @@ class IMipService {
 				continue;
 			}
 
-			$principalUri = 'principals/users/' . $account->getUserId();
+			$userId = $account->getUserId();
 			$recipient = $account->getEmail();
+			$imipCreate = $account->getImipCreate();
+			$systemVersion = $this->serverVersion->getMajorVersion();
 
 			foreach ($filteredMessages as $message) {
 				/** @var IMAPMessage $imapMessage */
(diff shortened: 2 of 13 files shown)


src/components/CalendarSettings.vue
```diff
@@ -0,0 +1,69 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<div>
+		<NcCheckboxRadioSwitch
```
ChristophWurst, line 8: missing loading state, which is problematic for multiple clicks in a short time

ChristophWurst (commented):
Looks good so far

Is the server change a hard requirement? If so, add a version check

src/components/CalendarSettings.vue
SebastianKrupinski: Done

ChristophWurst (approved):

lib/Migration/Version5007Date20251208000000.php
```diff
@@ -0,0 +1,37 @@
...
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+		$accountsTable = $schema->getTable('mail_accounts');
+		if (!$accountsTable->hasColumn('imip_create')) {
+			$accountsTable->addColumn('imip_create', Types::BOOLEAN, [
+				'default' => '0',
```
kesselb:
```suggestion
				'default' => false,
```

lib/Migration/Version5007Date20251208000000.php
```diff
@@ -0,0 +1,37 @@
...
+				'notNull' => false,
```
kesselb:
```suggestion
				'notnull' => false,
```

c.f. https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#portable-options

lib/Migration/Version5007Date20251208000000.php
```diff
@@ -0,0 +1,37 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+		$accountsTable = $schema->getTable('mail_accounts');
+		if (!$accountsTable->hasColumn('imip_create')) {
+			$accountsTable->addColumn('imip_create', Types::BOOLEAN, [
+				'default' => false,
+				'notnull' => false,
```
DerDreschner, line 32: Shouldn't this be true, as the default is false anyway?

src/components/CalendarSettings.vue
DerDreschner: What did you mean with "missing loading state"? Just to learn :)

DerDreschner (approved):
Tested successfully on my machine, ~~but does work only on messages delivered to the `Inbox`. When creating a rule on Webmin/Virtualmin to put it in another folder, it doesn't work anymore. But I guess that's expected behaviour. 😃 Their implementation doesn't use Sieve, so, it may differ there.~~

Looking forward to use it on a daily basis!

lib/Migration/Version5007Date20251208000000.php
SebastianKrupinski: no the default database value should be false.... imip_create = false, means do not automatically create events... this is a user opt in feature

src/components/CalendarSettings.vue
ChristophWurst: The line with `:disabled="saving"`. Before, clicking another time while the first click already triggered an asynchronous action led to a second concurrent action. By disabling the input element while we process the click we avoid this situation

lib/Migration/Version5007Date20251208000000.php
ChristophWurst: David means null vs notnull.

We can't make it notnull despite the default because Oracle stores boolean false as `null` (as it does with empty string)

(resubmitting this comment as somehow github has lost my reply from 3min ago)

DerDreschner: Yeah, that's what I meant @ChristophWurst :smile: Good to know, that's pretty important to remember. Thanks! :+1:

SebastianKrupinski:
> Tested successfully on my machine, but does work only on messages delivered to the `Inbox`. When creating a rule on Webmin/Virtualmin to put it in another folder, it doesn't work anymore. But I guess that's expected behaviour. 😃 Their implementation doesn't use Sieve, so, it may differ there.

This should work on other mail boxes, we process all the mail boxes, the issue might be that the alternate mailbox has not been sync'd with NC yet, we do not sync all mailboxes all the time.

lib/Service/IMipService.php
```diff
@@ -138,20 +144,35 @@ public function process(): void {
...
+							if (!method_exists($this->calendarManager, 'handleIMip')) {
+								$this->logger->error('iMIP handling is not supported by server version installed.');
+								continue;
+							}
+							$processed = $this->calendarManager->handleIMip(
+								$userId,
+								$schedulingInfo['contents'],
+								[
+									'recipient' => $recipient,
+									'absent' => $imipCreate ? 'create' : 'ignore',
```
kesselb, line 168: [`options.absent`  is defined as not set or "create". "ignore" is not a valid value.](https://github.com/nextcloud/server/blob/0ff8b35578827797f1bbc0f0a5f27ff52a0b610a/lib/public/Calendar/IManager.php#L146)

SebastianKrupinski: Server PR is merged... will rebase in a couple hours will see what psalm says

kesselb: `static-psalm-analysis dev-master` should pass 🥳 
 
`static-psalm-analysis dev-stable32`  will still fail 💥 

We could change `OCP\Calendar\IManager.handleIMip` in stable32 like

`@param array{absent?: "create"|"ignore", recipient?: string, ...} $options`

To avoid the warning about "ignore" and make psalm more relax about the unknown "absentCreateStatus" with the "...". 

That isn't ideal, but at least we don't have to suppress some errors or bump the baseline only for one supported version.

SebastianKrupinski: Yeah... I agree we need to adjust stable32 cause psalm will wine again

SebastianKrupinski: https://github.com/nextcloud/server/pull/57582

ChristophWurst: I'm skeptical about this approach. That means you have the public API (interface) changes partially backported without the implementation.
That feels worse than ignoring the psalm warning here until we have dropped off 32 support :thinking:

kesselb: Adding "ignore" as value to absent is syntactical sugar. It's already possible as of today by just omitting absent. There's no implementation to be done, the code only runs if you set it to "create" on 32 or newer.  

absentCreateStatus is not documented for stable32 and will not be. We are just telling psalm, by adding `...`, to be more relaxed about the additional absentCreateStatus array element. Psalm will not warn you about a typo like "absentCreateStatu", but it will warn you if you set absentCreateStatus to an invalid value.

ChristophWurst: Ok, that makes sense. Thanks!

davidlohner:
@SebastianKrupinski 
Thank you! Just got pointed to this feature since [I've asked over on Mastodon](https://higher-edu.social/@nextcloud@mastodon.xyz/116526566115321132).

It would be nice to select a specific calendar (per email account) next to the checkbox shown in the initial post.

SebastianKrupinski:
> @SebastianKrupinski Thank you! Just got pointed to this feature since [I've asked over on Mastodon](https://higher-edu.social/@nextcloud@mastodon.xyz/116526566115321132).
> 
> It would be nice to select a specific calendar (per email account) next to the checkbox shown in the initial post.

Hi @davidlohner 

Thank you for the complement! 

As for the additional enhancement request, we can add that to our very long list of feature requests

PR #12145, opened 2025-12-09
[stable5.6] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (3 unfixable, 16 only fixable manually using --force)


ChristophWurst:
Nothing fixed

PR #12183, opened 2025-12-17
fix(ui): bind attr to popover trigger

ChristophWurst:
Fixes

```
[Vue warn]: It looks like you are using a custom button as a <NcPopover> or other popover #trigger. If you are not using <NcButton> as a trigger, you need to bind attrs from the #trigger slot props to your custom button. See <NcPopover> docs for an example.

found in

---> <NcPopover>
       <NcActions>
         <RouterLink>
           <EnvelopeSkeleton> at src/components/EnvelopeSkeleton.vue
             <Envelope> at src/components/Envelope.vue
               <TransitionGroup>
                 <EnvelopeList> at src/components/EnvelopeList.vue
                   <Mailbox> at src/components/Mailbox.vue
                     <NcAppContentList>
                       <Pane>
                         <Splitpanes>
                           <NcAppContent>
                             <MailboxThread> at src/components/MailboxThread.vue
                               <NcContent>
                                 <Home> at src/views/Home.vue
                                   <App> at src/App.vue
                                     <Root>
```

after page load

From https://nextcloud-vue-components.netlify.app/#/Components/NcPopover

> If you are using your own custom button as a trigger make sure to bind attrs from the trigger slot props.

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -27,8 +27,10 @@
 						<Popover
 							v-if="threadParticipants.length > participantsToDisplay"
 							class="avatar-more">
-							<template #trigger>
-								<span class="avatar-more">
+							<template #trigger="{ attrs }">
+								<span
+									class="avatar-more"
+									v-bind="attrs">
 									{{ moreParticipantsString }}
 								</span>
 							</template>


src/components/mailFilter/Operator.vue
```diff
@@ -14,8 +14,8 @@
 			:clearable="false"
 			@input="updateOperator($event)" />
 		<NcPopover class="operator__popover" :no-focus-trap="true" popup-role="dialog">
-			<template #trigger>
-				<NcButton variant="tertiary-no-background" :aria-label="t('mail', 'Help')">
+			<template #trigger="{ attrs }">
+				<NcButton variant="tertiary-no-background" :aria-label="t('mail', 'Help')" v-bind="attrs">
```
kesselb: Thanks 👍 

Why's that a custom button?

ChristophWurst: You are right, it doesn't need attr binding.

> When <NcButton> is used as a <NcPopover> trigger, it injects required for a11y attributes to the button.

ChristophWurst: Reverted the changes for anc NcButton triggers

kesselb:
Thanks a lot :partying_face: 

I was looking into that before, but I didn't find that one place where we were not using an NcButton :see_no_evil: 

Finally that warning is gone :tada:

Issue #11169, opened 2025-05-19
Make use of the `$Phishing` flag to detect phishing emails


hamza221:
### Is your feature request related to a problem? Please describe.

Delivery agents can mark a message as phishing with `$Phishing` flag.
https://datatracker.ietf.org/doc/html/rfc9051#name-flags-message-attribute

### Describe the solution you'd like

Make use of the `$Phishing` flag to reinforce the mail app's Phishing detection.
* Add a new check in  [PhishingDetectionService](https://github.com/nextcloud/mail/blob/main/lib/Service/PhishingDetection/PhishingDetectionService.php) 
* Display message eg ' Message is highly likely to be a phishing email ' 
* Follow the rfc ( placing the message in the \Junk special-use mailbox ...)

PR #12187, opened 2025-12-18
feat(PhishingDetection): Add support for the `$Phishing` IMAP flag

DerDreschner:
This PR introduces support for the `$Phishing` IMAP flag defined since IMAP4rev2 (see [RFC9051](https://datatracker.ietf.org/doc/html/rfc9051#name-flags-message-attribute)). When the flag is set by the delivery agent (e.g., Dovecot), the following warning is being shown to the user:

Screenshot 2025-12-18 at 22-21-40 Mail - Nextcloud

**Important**: For compatibility reasons, the warning is only being shown when both the `$Junk` / `Junk` and `$Phishing` flags are present. That matches the behavior described in the RFC with the exception that Thunderbird is being supported as well, although it uses a non-standard flag for junk/non-junk flagging.

Additionally, I've added some unit tests for the phishing warning front-end component.

Closes https://github.com/nextcloud/mail/issues/11169

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -84,6 +84,7 @@ class Message extends Entity implements JsonSerializable {
 		'forwarded',
 		'$junk',
 		'$notjunk',
+		'$phishing',
 		'mdnsent',
 		Tag::LABEL_IMPORTANT,
 		'$important' // @todo remove this when we have removed all references on IMAP to $important @link https://github.com/nextcloud/mail/issues/25
--- a/lib/Service/PhishingDetection/ImapFlagCheck.php
+++ b/lib/Service/PhishingDetection/ImapFlagCheck.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Service\PhishingDetection;
+
+use Horde_Imap_Client;
+use OCA\Mail\PhishingDetectionResult;
+use OCP\IL10N;
+
+class ImapFlagCheck {
+	protected IL10N $l10n;
+
+	public function __construct(IL10N $l10n) {
+		$this->l10n = $l10n;
+	}
+
+	/**
+	 * @param string[] $messageFlags
+	 */
+	public function run(array $messageFlags): PhishingDetectionResult {
+		$flaggedAsSpam = in_array(Horde_Imap_Client::FLAG_JUNK, $messageFlags, true) || in_array('junk', $messageFlags, true);
+		// TODO: Use Horde const once the flag is implemented there
+		//  (https://github.com/bytestream/Imap_Client/blob/master/lib/Horde/Imap/Client.php#L153).
+		$flaggedAsPhishing = in_array('$phishing', $messageFlags, true);
+
+		if ($flaggedAsSpam && $flaggedAsPhishing) {
+			return new PhishingDetectionResult(PhishingDetectionResult::IMAP_FLAG_CHECK, true, $this->l10n->t('Mail server marked this message as phishing attempt'));
+		}
+
+		return new PhishingDetectionResult(PhishingDetectionResult::IMAP_FLAG_CHECK, false);
+	}
+}
--- a/lib/Service/PhishingDetection/PhishingDetectionService.php
+++ b/lib/Service/PhishingDetection/PhishingDetectionService.php
@@ -21,10 +21,19 @@ class PhishingDetectionService {
 		private DateCheck $dateCheck,
 		private ReplyToCheck $replyToCheck,
 		private LinkCheck $linkCheck,
+		private ImapFlagCheck $imapFlagCheck,
 	) {
 	}
 
-	public function checkHeadersForPhishing(Horde_Mime_Headers $headers, bool $hasHtmlMessage, string $htmlMessage = ''): array {
+	/**
+	 * @param Horde_Mime_Headers $headers
+	 * @param string[] $flags
+	 * @param bool $hasHtmlMessage
+	 * @param string $htmlMessage
+	 * @return array
+	 * @throws \Exception
+	 */
+	public function checkHeadersForPhishing(Horde_Mime_Headers $headers, array $flags, bool $hasHtmlMessage, string $htmlMessage = ''): array {
 		/** @var string|null $fromFN */
 		$fromFN = null;
 		/** @var string|null $fromEmail */
@@ -66,6 +75,9 @@ class PhishingDetectionService {
 		if ($date !== null) {
 			$list->addCheck($this->dateCheck->run($date));
 		}
+
+		$list->addCheck($this->imapFlagCheck->run($flags));
+
 		if ($hasHtmlMessage) {
 			$list->addCheck($this->linkCheck->run($htmlMessage));
 		}
--- a/src/tests/unit/components/PhishingWarning.vue.spec.js
+++ b/src/tests/unit/components/PhishingWarning.vue.spec.js
@@ -1,0 +1,165 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { createLocalVue, shallowMount } from '@vue/test-utils'
+import PhishingWarning from '../../../components/PhishingWarning.vue'
+import Nextcloud from '../../../mixins/Nextcloud.js'
+
+const localVue = createLocalVue()
+localVue.mixin(Nextcloud)
+
+describe('PhishingWarning', () => {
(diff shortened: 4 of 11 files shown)


lib/Migration/Version5700Date20251218204405.php
```diff
@@ -0,0 +1,39 @@
...
+	#[Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$messagesTable = $schema->getTable('mail_messages');
+		$messagesTable->addColumn('flag_phishing', Types::BOOLEAN, [
+			'notnull' => false,
+			'default' => false,
+		]);
```
ChristophWurst: The messages table is rather big already, so a migration is expensive. The flag data for the phishing service is coming right from IMAP, not the DB. Therefore I suggest we remove the migration and $phishing flag persistence.

lib/Controller/MessagesController.php
```diff
@@ -472,7 +472,7 @@ public function mdn(int $id): JSONResponse {
 
 		try {
 			$this->mailTransmission->sendMdn($account, $mailbox, $message);
-			$this->mailManager->flagMessage($account, $mailbox->getName(), $message->getUid(), 'mdnsent', true);
+			$this->mailManager->flagMessage($account, $mailbox->getName(), $message->getUid(), '$mdnsent', true);
```
ChristophWurst: nitpick: I would prefer to have the mdnsent to $mdnsent fix in a separate PR. That keeps the change set small. Additionally we are able to *backport* the flag change to an older stable branch.

lib/Service/PhishingDetection/ImapFlagCheck.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
ChristophWurst: 2025 :)

lib/Service/PhishingDetection/ImapFlagCheck.php
```diff
@@ -0,0 +1,37 @@
...
+use OCP\IL10N;
+
+class ImapFlagCheck {
+	protected IL10N $l10n;
+
+	public function __construct(IL10N $l10n) {
+		$this->l10n = $l10n;
+	}
+
+	public function run(array $messageFlags) : PhishingDetectionResult {
```
ChristophWurst: CS + more specific typing

```suggestion
	/**
	 * @param string[] $messageFlags
	 */
	public function run(array $messageFlags): PhishingDetectionResult {
```

lib/Service/PhishingDetection/ImapFlagCheck.php
```diff
@@ -0,0 +1,37 @@
...
+		$flaggedAsSpam = in_array(Horde_Imap_Client::FLAG_JUNK, $messageFlags) || in_array('junk', $messageFlags);
```
ChristophWurst: nitpick: always add `true` as third argument of `in_array` to enforce strict type comparisons. We *shouldn't* receive anything but strings for the flags but better safe than sorry. Non-strict PHP comparisons are unpredictable.

lib/Service/PhishingDetection/ImapFlagCheck.php
```diff
@@ -0,0 +1,37 @@
...
+		// TODO: Use Horde const once the flag is implemented there
+		//  (https://github.com/bytestream/Imap_Client/blob/master/lib/Horde/Imap/Client.php#L153).
+		$flaggedAsPhishing = in_array('$phishing', $messageFlags);
```
ChristophWurst: same here

lib/Service/PhishingDetection/ImapFlagCheck.php
```diff
@@ -0,0 +1,37 @@
...
+
+		if ($flaggedAsSpam && $flaggedAsPhishing) {
+			return new PhishingDetectionResult(PhishingDetectionResult::IMAP_FLAG_CHECK, true, $this->l10n->t('Mail server marked this message as phishing attempt'));
+		}
+
+		return new PhishingDetectionResult(PhishingDetectionResult::IMAP_FLAG_CHECK, false);
+
```
ChristophWurst: CS

```suggestion
		return new PhishingDetectionResult(PhishingDetectionResult::IMAP_FLAG_CHECK, false);
```

lib/Service/PhishingDetection/PhishingDetectionService.php
```diff
@@ -21,10 +21,11 @@ public function __construct(
 		private DateCheck $dateCheck,
 		private ReplyToCheck $replyToCheck,
 		private LinkCheck $linkCheck,
+		private ImapFlagCheck $imapFlagCheck,
 	) {
 	}
 
-	public function checkHeadersForPhishing(Horde_Mime_Headers $headers, bool $hasHtmlMessage, string $htmlMessage = ''): array {
+	public function checkHeadersForPhishing(Horde_Mime_Headers $headers, array $flags, bool $hasHtmlMessage, string $htmlMessage = ''): array {
```
ChristophWurst, line 36: nitpick: specific type documentation for `$flags` via phpdoc

src/tests/unit/components/PhishingWarning.vue.spec.js
```diff
@@ -0,0 +1,165 @@
+/**
+ * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
ChristophWurst: 2025

tests/Unit/Service/Phishing/ImapFlagCheckTest.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
ChristophWurst: 2025

ChristophWurst (commented):
Thanks a lot!

This looks very good!

I've added some comments for clean-up and coding style

lib/Controller/MessagesController.php
DerDreschner: Gotcha! I had this in mind and put these changes in separate commits to make it easy to revert (and cherry-pick in a new branch). Should I revert the change to hide the Thunderbird `junk` flag as well and put that into the same PR as the `$mdnsent` fix (or a second fix-PR)?

Sorry, my habit is still "one supported release (e.g., what's live) only". But I'll eager to learn and change. :)

lib/Migration/Version5700Date20251218204405.php
DerDreschner: I'm on your side. I've only put the flag persistence into the PR to be consistent with the other flag implementations, which are stored there as well ($junk, $nonjunk, $mdnsent). But that only happens because the other flags can be actively manipulated via the mail app and have to be synced to the IMAP server later, right?

The other question I have is what happens to the flag (and my code) when the IMAP server is down when the user opens a message, as these are being fetched directly from the server? I would guess that this case is (or should be) supported somehow as the mail messages are stored in the database? Or is it just for performance reasons?

lib/Db/Message.php
```diff
@@ -84,7 +84,8 @@ class Message extends Entity implements JsonSerializable {
 		'forwarded',
 		'$junk',
 		'$notjunk',
-		'mdnsent',
+		'$phishing',
```
DerDreschner, line 87: My understanding is that this should still be added, although the flag isn't persisted into the database, right? Otherwise, the flag could be added by the user manually...?

lib/Controller/MessagesController.php
DerDreschner: I moved the changes to the `mdnsent` flag as well as hiding the non-standard `junk` flag into this PR: #12189

lib/Migration/Version5700Date20251218204405.php
ChristophWurst: The flags are in the database for quick filtering and for the sections of the *priority inbox*. This allows us to build a list of threads for a specific search query without involving the (slow) IMAP server.
For phishing I'd see filtering as problematic because users would only see emails flagged by the server. They would not see emails that are only considered as phishing by our app.
Technically, we could process every email and check it for phishing and persist that state (e.g. via that flag). Then the user filtering would be accurate. Something for another day :wink: 

Offline support is limited. The oc_mail_messages contents alone are insufficient to show an email. You still need IMAP to fetch the body. So either that works and you can also do the phishing processing or you won't see the email anyway.

lib/Db/Message.php
ChristophWurst: Yes let's go this way for now

ChristophWurst:
One last challenge, and sorry for forgetting to tell you about this convention before the PR, could you clean up the git history?

I prefer a linear git history with somewhat atomic commits. The reason is that it allows for easier reverts if something goes wrong, and that we use release automation, which generates changelogs based on semantic commits.

I can also clean it up for you if it turns out to be tricky with the current history :)

PR #12189, opened 2025-12-19
fix: Fix incorrect flag for MDN sent and hide junk flag used by Thunderbird

DerDreschner:
This is a follow-up PR from #12187 and fixes the following points that were discovered during development there:
- **Fix**: Use correct flag for MDNSent throughout the code base (`mdnsent` -> `$mdnsent` - $ sign was missing). There is no case known to me where this flag is being used without the $ sign (even Thunderbird does it, I've tested that).
- **Fix**: Thunderbird uses the non-standard `junk` flag without $ sign, which was previously shown to the user as tag. This is being fixed here.

### DIFF
--- a/lib/Db/Message.php
+++ b/lib/Db/Message.php
@@ -85,7 +85,7 @@ class Message extends Entity implements JsonSerializable {
 		'$junk',
 		'$notjunk',
 		'$phishing',
-		'mdnsent',
+		'$mdnsent',
 		Tag::LABEL_IMPORTANT,
 		'$important' // @todo remove this when we have removed all references on IMAP to $important @link https://github.com/nextcloud/mail/issues/25
 	];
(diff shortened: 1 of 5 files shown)


lib/Db/Message.php
```diff
@@ -338,7 +338,7 @@ public function jsonSerialize() {
 				'important' => ($this->getFlagImportant() === true),
 				'$junk' => ($this->getFlagJunk() === true),
 				'$notjunk' => ($this->getFlagNotjunk() === true),
-				'mdnsent' => ($this->getFlagMdnsent() === true),
+				'$mdnsent' => ($this->getFlagMdnsent() === true),
```
kesselb, line 344: If we change it there, we should also change for consistency: 

https://github.com/nextcloud/mail/blob/6de0c2a0df34804c008308cb2c7b490376f650d0/src/components/MdnRequest.vue#L53-L55

https://github.com/nextcloud/mail/blob/6de0c2a0df34804c008308cb2c7b490376f650d0/lib/Model/IMAPMessage.php#L165

DerDreschner: Good point! That's done.

PR #12222, opened 2026-01-09
feat(attachments): Add support for changed API due to introduction of snowflake ids in server

DerDreschner:
The underlying server is introducing [snowflake ids](https://en.wikipedia.org/wiki/Snowflake_ID). This means that the API will change when using them (`id` will be from type `string` in the future, instead of `int`).

Although we don't use them right now, there were some [breaking changes](https://github.com/nextcloud/server/commit/7c1a8a4060413b85448dc61c315a9f03368e750a#diff-b5b13a13f1cff8ae531cbba72e8a905e01be6515108f6c450ab19b06ca421426L20) in the entity model (that are going [to be reverted partially](https://github.com/nextcloud/server/pull/57442)). As we have to use them in the future anyway, I think it's a good idea to merge the changes from this PR anyway to be a bit more prepared.

### DIFF
--- a/lib/Service/Attachment/AttachmentService.php
+++ b/lib/Service/Attachment/AttachmentService.php
@@ -82,6 +82,10 @@ class AttachmentService implements IAttachmentService {
 		$attachment->setMimeType($file->getMimeType());
 
 		$persisted = $this->mapper->insert($attachment);
+		if ($persisted->id === null) {
+			throw new UploadException('Given attachmentId is null');
+		}
+
 		try {
 			$this->storage->save($userId, $persisted->id, $file);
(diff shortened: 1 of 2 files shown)


DerDreschner:
I think we're not affected by any other changes beside the ones that will be reverted in https://github.com/nextcloud/server/pull/57442. No tables get converted automatically, which means our own tables are fine for now. Aside from them I'm not aware of any other tables we're currently access/use? Or what do you mean with `file id`?

Just don't want to do anything wrong or missing any task for myself. :sweat_smile:

ChristophWurst:
Yes, CI is green again in other new PRs

Issue #12057, opened 2025-11-19
Don't provisioning an account when user is not able to use the mail app


kesselb:
### Steps to reproduce

1. Enable the mail app only for some groups
2. Have a provisioning configuration
3. See accounts are also created for user that are unable to use the app

### Expected behavior

No account is created when the user is unable to use mail.

### Actual behavior

Account created.

PR #12226, opened 2026-01-09
fix(provisioning): Do not provision mail account for users without access to mail app

DerDreschner:
When using the provisioning functionality, only the configured pattern where checked before pushing the mail account configuration into the database. This is fine as long as all users can use the mail app. But as the access to specific apps can be limited to specific groups, this can lead to accounts that are configured and synced in the background, yet being never used and only eat up precious resources.

To ensure this does not happen anymore, I use the `IAppManager` to check if the specific user has the right to use the mail app. If not, the provisioning for the user will be stopped.

This fixes https://github.com/nextcloud/mail/issues/12057.

### DIFF
--- a/lib/Db/AliasMapper.php
+++ b/lib/Db/AliasMapper.php
@@ -106,12 +106,7 @@ class AliasMapper extends QBMapper {
 	/**
 	 * Delete all provisioned aliases for the given uid
 	 *
-	 * Exception for Nextcloud 20: \Doctrine\DBAL\DBALException
-	 * Exception for Nextcloud 21 and newer: \OCP\DB\Exception
-	 *
-	 * @TODO: Change throws to \OCP\DB\Exception once Mail does not support Nextcloud 20.
-	 *
-	 * @throws \Exception
+	 * @throws \OCP\DB\Exception
 	 */
 	public function deleteProvisionedAliasesByUid(string $uid): void {
 		$qb = $this->db->getQueryBuilder();
--- a/lib/Service/Provisioning/Manager.php
+++ b/lib/Service/Provisioning/Manager.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace OCA\Mail\Service\Provisioning;
 
 use Horde_Mail_Rfc822_Address;
+use OCA\Mail\AppInfo\Application;
 use OCA\Mail\Db\Alias;
 use OCA\Mail\Db\AliasMapper;
 use OCA\Mail\Db\MailAccount;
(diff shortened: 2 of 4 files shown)


lib/Service/Provisioning/Manager.php
```diff
@@ -186,6 +192,10 @@ public function ldapAliasesIntegration(Provisioning $provisioning, IUser $user):
 	 * @param Provisioning[] $provisionings
 	 */
 	public function provisionSingleUser(array $provisionings, IUser $user): bool {
+		if (!$this->appManager->isEnabledForUser(Application::APP_ID, $user)) {
+			return false;
```
DerDreschner, line 221: @kesselb : You mean to run the code to remove provisioned accounts here? As the `ProvisionAccounts` migration just runs over all users?

kesselb: Yes

lib/Db/AliasMapper.php
```diff
@@ -106,12 +107,7 @@ public function deleteAll($accountId): void {
 	/**
 	 * Delete all provisioned aliases for the given uid
 	 *
-	 * Exception for Nextcloud 20: \Doctrine\DBAL\DBALException
```
DerDreschner, line 109: As we're far away from NC 20/21, I think it's okay to change that here now?

ChristophWurst: Yep

tests/Unit/Service/Provisioning/ManagerTest.php
```diff
@@ -61,6 +61,30 @@ public function testProvisionSkipWithoutConfigurations(): void {
...
+			->method('isEnabledForUser')
+			->willReturn(false);
+		$this->mock->getParameter('aliasMapper')
+			->expects($this->once())
+			->method('deleteProvisionedAliasesByUid')
+			->with($user->getUID());
+		$this->mock->getParameter('mailAccountMapper')
+			->expects($this->once())
+			->method('deleteProvisionedAccountsByUid')
+			->with($user->getUID());
```
ChristophWurst, line 82: RE: https://github.com/nextcloud/mail/pull/12226#issuecomment-3759957084

This is sufficient for the unit test. You are asserting that the method is called correctly. Whatever happens inside that method is not subject of this unit test

lib/Service/Provisioning/Manager.php
```diff
@@ -186,6 +193,26 @@ public function ldapAliasesIntegration(Provisioning $provisioning, IUser $user):
 	 * @param Provisioning[] $provisionings
 	 */
 	public function provisionSingleUser(array $provisionings, IUser $user): bool {
+		$unprovisionUser = function (string $userUid) : void {
```
kesselb: Please make that a class method.

DerDreschner: Done :smile:

DerDreschner:
@kesselb Done!

PR #12227, opened 2026-01-09
feat(deps): add PHP8.5 support

ChristophWurst:
We have run tests with PHP8.5 for a long time. Now that server supports 8.5 we can allow it too.

The test workflow is already adjusted for 8.6 once we add it to the matrix.


ChristophWurst:
I suggest we fix them in the coming weeks as we see them pop up on CI or with local testing

PR #12238, opened 2026-01-13
fix(DKIMValidator): Fix broken canonical handling in upstream dependency

DerDreschner:
There are several bugs in the DKIMValidator library which we use in mail. These affect the canonical flag (`c=`) in the following cases:
- `c=` is missing as described in the RFC (`c=simple/simple` must be used then). This will trigger a `Undefined offset` or `Undefined index` notice and may throw a `TypeError` exception later on.
- `c=` contains only a header algorithm (`c=relaxed` or `c=simple` - `simple` must be used for the body then). This will trigger a `TypeError` exception as described in ticket https://github.com/nextcloud/mail/issues/12206.
- `c=simple` being used for header canonicalization. This will result in an invalid DKIM validation, although the mail contains a valid DKIM signature.

As being discussed with @ChristophWurst, we will introduce this patch until the fix is accepted and released upstream (https://github.com/PHPMailer/DKIMValidator/pull/21 - contains additional unit tests for each case).

Fixes https://github.com/nextcloud/mail/issues/12206


ChristophWurst (approved):
Nice

PR #12255, opened 2026-01-15
fix: action items icons on compact mode

GretaD:
@nimishavijay In your mockup, the important icon is second. Since we have the important icon first in the avatar, I thought  to do the same here.
If you have a strong opinion on that, I'm happy to change the order

list view
Screenshot from 2026-01-26 14-13-14

vertical view
Screenshot from 2026-01-26 14-30-47

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -32,27 +32,30 @@
 		@toggle-seen="onToggleSeen"
 		@update:menuOpen="closeMoreAndSnoozeOptions">
 		<template #icon>
-			<Star
-				v-if="data.flags.flagged"
-				fill-color="#f9cf3d"
-				:size="18"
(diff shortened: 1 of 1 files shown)


GretaD:
I forgot to change the icon-spam

nimishavijay (changes requested):
Thanks for the quick fix! Good point about it being first, I don't have a strong opinion about this. However 2 other points: 
- [ ] could we change the colour of the important icon to primary? Before their positions were very different so it was easy to see distinguish between them, but since both are basically in the same position it is difficult to tell if a message is marked as favourite or important 
- [ ] before these buttons used to be clickable, is that still the case? If so it seems like the clickable area is very small. The size of the icons can also be increased to 16px or even 18px because they seem very small now.

PR #12265, opened 2026-01-16
fix(admin-settings): bring back the setting to disable classification…

hamza221:
Was removed accidentally  in  #11968

### DIFF
--- a/lib/Migration/Version5007Date20251024153423.php
+++ b/lib/Migration/Version5007Date20251024153423.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace OCA\Mail\Migration;
 
 use Closure;
+use OCP\AppFramework\Services\IAppConfig;
 use OCP\DB\ISchemaWrapper;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\DB\Types;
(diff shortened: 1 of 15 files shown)


lib/Migration/Version5007Date20251024153423.php
```diff
@@ -44,26 +44,55 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
+			->from('appconfig')
+			->where(
+				$qb->expr()->andx(
+					$qb->expr()->eq('appid', $qb->createNamedParameter('mail')),
+					$qb->expr()->eq('configkey', $qb->createNamedParameter('importance_classification_default')),
+				)
+			);
+		$res = $qb->executeQuery();
+		$row = $res->fetch();
+		$res->closeCursor();
```
ChristophWurst: nit: use `\OCP\AppFramework\Services\IAppConfig::getAppValueBool`

We have to "break" the abstraction for the mass update. But for this simple fetch we can go the usual route

lib/Migration/Version5007Date20251024153423.php
```diff
@@ -44,26 +44,55 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
 		$res = $qb->executeQuery();
 		$users = $res->fetchAll();
-
 		$res->closeCursor();
+
 		$output->info('Migrating classification user preferences to mail_accounts table');
 		$output->startProgress();
+
+		// If classification is disabled by default we first disable it for all users
+		if(!$isEnabledBydefault){
```
ChristophWurst:
```suggestion
		if (!$isEnabledBydefault) {
```

nit: funky coding style

PR #12266, opened 2026-01-16
chore(ci): Switch to setup-php@verbose version due to stuck runs

DerDreschner:
At least @SebastianKrupinski and myself experienced some workflow runs where the `set up php` step stuck. As discussed in the talk channel, this seem to affect more people from within the company. It also [happened today](https://github.com/nextcloud/mail/actions/runs/21071987240/job/60603837762?pr=12259#step:3:60) with the `setup-nextcloud` workflow action from @ChristophWurst , which tags `@v2` and therefore already uses the latest version:

Screenshot 2026-01-16 at 22-25-19 fix(imap) Add support for mail services that enforce a client id · nextcloud_mail@0f64a8b

To be able to find the root cause for that issue, I've decided to switch the daily workflows to the `@verbose` version (see: https://github.com/shivammathur/setup-php?tab=readme-ov-file#verbose-setup). It might be worth to do that for the `setup-nextcloud` workflow as well, to get more verbose logging on possibly stuck runs faster.

### DIFF
--- a/.github/workflows/lint-php-cs.yml
+++ b/.github/workflows/lint-php-cs.yml
@@ -34,7 +34,7 @@ jobs:
         uses: icewind1991/nextcloud-version-matrix@58becf3b4bb6dc6cef677b15e2fd8e7d48c0908f # v1.3.1
 
       - name: Set up php${{ steps.versions.outputs.php-min }}
-        uses: shivammathur/setup-php@44454db4f0199b8b9685a5d763dc37cbf79108e1 # v2.36.0
+        uses: shivammathur/setup-php@verbose # v2.36.0 with verbose logging due to stuck runs
         with:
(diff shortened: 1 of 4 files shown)


SebastianKrupinski:
@DerDreschner Had a look the changes, they look right... the CI run is the true test tho... I am surprised that some of those action versions where not updated by dependabot

ChristophWurst:
> Edit: I've removed the [last two steps](https://github.com/nextcloud/.github/blob/master/workflow-templates/openapi.yml#L87) (`Check openapi*.json [...]` and `Show changes on failure`) from the `openapi` template as it fails on our repository. They were not included before anyway here (why?).

We try not to commit artifacts. They bloat the repository, you have to keep source and artifact in sync and it can lead to additional conflicts. Therefore we just generate these files when the package is built.

PR #12269, opened 2026-01-19
feat: add html and source editing support

### LINKED ISSUE #4557: Not able to configure a pure HTML signature without getting it defaced by the editor.

GretaD:
fixes #4557

### Todos:

- [ ] Clarify the scope of the feature; what do we want to support?
- [x] Add elements to allow list
- [x] Find out why saving, in the source view, does not work
- [ ] Use proxy for images?

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -30,7 +30,7 @@ import {
 	Base64UploadAdapter,
 	BlockQuote,
 	Bold,
-	DecoupledEditor,
+	ClassicEditor,
 	DropdownView,
 	Essentials,
 	FindAndReplace,
@@ -46,6 +46,7 @@ import {
 	Mention,
 	Paragraph,
 	RemoveFormat,
+	SourceEditing,
 	Strikethrough,
 	Subscript,
 	Superscript,
@@ -153,6 +154,7 @@ export default {
 				RemoveFormat,
 				Base64UploadAdapter,
 				MailPlugin,
+				SourceEditing,
 				TextDirectionPlugin,
 			])
 			toolbar.unshift(...[
@@ -177,6 +179,7 @@ export default {
 				'link',
 				'removeFormat',
 				'findAndReplace',
+				'sourceEditing',
 			])
 		}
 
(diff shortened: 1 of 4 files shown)


src/components/TextEditor.vue
```diff
@@ -220,14 +225,28 @@ export default {
...
+							classes: true,
+							styles: true,
+						},
 						{
 							name: 'img',
-							attributes: {
-								'data-cid': true,
-							},
+							styles: true,
+							attributes: true,
```
kesselb: @kesselb check if forwarding still works? ;)

src/components/TextEditor.vue
```diff
@@ -236,7 +255,43 @@ export default {
...
+		registerSourceEditingInputListener(editor) {
+			const textarea = editor.ui.getEditableElement('sourceEditing:main')
+
+			if (!textarea || textarea === this.sourceEditingTextarea) {
+				return
+			}
+
+			this.unregisterSourceEditingInputListener()
+
+			this.sourceEditingInputHandler = () => {
```
kesselb, line 275: @kesselb debounce?

kesselb (approved):

PR #12272, opened 2026-01-20
fix: catch invalid URL exception in the HTML image proxy

ChristophWurst:
Emails with bogus URLs, e.g. with no hostname, cause a HTTP500. This handles the case more gracefully. The error will be logged as warning, not error, because admins can't do anything about this.

### DIFF
--- a/lib/Controller/ProxyController.php
+++ b/lib/Controller/ProxyController.php
@@ -17,6 +17,7 @@ use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\Attribute\UserRateLimit;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\Http\Client\IClientService;
+use OCP\Http\Client\LocalServerException;
 use OCP\IRequest;
 use OCP\ISession;
 use OCP\IURLGenerator;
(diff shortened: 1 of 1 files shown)


lib/Controller/ProxyController.php
```diff
@@ -108,6 +109,12 @@ public function proxy(string $src): ProxyDownloadResponse {
 		} catch (ClientExceptionInterface $e) {
 			$this->logger->notice('Unable to proxy image', ['exception' => $e]);
 			$content = file_get_contents(__DIR__ . '/../../img/blocked-image.png');
+		} catch (LocalServerException $e) {
+			$this->logger->warning('Prevented image proxy access to forbidden URL', [
```
szaimen, line 113:
```suggestion
			$this->logger->info('Prevented image proxy access to forbidden URL', [
```

ChristophWurst: Why?

szaimen: Because as you said there is nothing an admin can do about this, so it should be printed to info level imho as we do in a lot of other places. Or is this critical from a security standpoint for example?

ChristophWurst: Yes, security is why I would use warning

PR #12277, opened 2026-01-20
fix(MessageMapper): Use message headers as fallback when processing MIME content

### LINKED ISSUE #12228: Message preview with "Content-Transfer-Encoding: base64" wrong?
### Steps to reproduce

Image

The mail has set

```
Content-Type: text/plain; charset="utf-8"
```

and does render correctly when opening, just the preview shows the base64 encoded data?

### Expected behavior

The preview should take base64 encoded data into account and show the actual contents.

### Actual behavior

Base64 encoded data in preview.

### Mail app version

5.7.0-beta.1

### Nextcloud version

32.0.3

### Mailserver or service

Exchange

### Operating system

Debian 13

### PHP engine version

Other

### Nextcloud memory caching

APCu & Redis

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

PHP 8.4 is used. This is the first time the mailbox was connected to mail, so no idea if that is a regression.

DerDreschner:
When generating the message preview, we previously used the MIME headers exclusively to determine the `Content-Type` and `Content-Transfer-Encoding` values. This is fine as long as the MDA doesn't follow the RFC too strictly and automatically fall back to the message headers if the MIME part in question has no headers set (e.g., when using Dovecot).

But other MDAs like Exchange / Office 365 or Stalwart don't do that. In this case, the current code path believe that nothing has to be converted at all - which isn't correct and produce an unreadable message preview. This must be handled by manually checking both the MIME headers **and** the message header for any `Content-Type` / `Content-Transfer-Encoding` values set as described in [RFC 2045 Section 6.4](https://datatracker.ietf.org/doc/html/rfc2045#section-6.4).

Fixes #12228.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -881,7 +881,6 @@ class MessageMapper {
 		$structureQuery = new Horde_Imap_Client_Fetch_Query();
 		$structureQuery->structure();
 		$structureQuery->headerText([
-			'cache' => true,
 			'peek' => true,
 		]);
 		$this->smimeService->addEncryptionCheckQueries($structureQuery);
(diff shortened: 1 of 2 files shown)


lib/IMAP/MessageMapper.php
```diff
@@ -881,7 +881,6 @@ public function getBodyStructureData(Horde_Imap_Client_Socket $client,
 		$structureQuery = new Horde_Imap_Client_Fetch_Query();
 		$structureQuery->structure();
 		$structureQuery->headerText([
-			'cache' => true,
```
DerDreschner, line 884: This setting does nothing at all, as it's not being used in the `headerText()` function and not mentioned in the description as valid value at all. That's why I've removed it here to avoid confusion in the future.

PR #12288, opened 2026-01-21
fix: clean text blocks on user deletion

hamza221:
Deletes all text blocks and their shares once a user gets deletd

### DIFF
--- a/lib/Db/TextBlockMapper.php
+++ b/lib/Db/TextBlockMapper.php
@@ -73,4 +73,11 @@ class TextBlockMapper extends QBMapper {
 		return $this->findEntities($qb);
 	}
 
+	public function deleteByUserId(string $userId) {
+		$qb = $this->db->getQueryBuilder();
+		$delete = $qb->delete($this->getTableName())
+			->where($qb->expr()->eq('owner', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)));
+		$delete->executeStatement();
+	}
+
 }
--- a/lib/Listener/UserDeletedListener.php
+++ b/lib/Listener/UserDeletedListener.php
@@ -11,6 +11,7 @@ namespace OCA\Mail\Listener;
 
 use OCA\Mail\Exception\ClientException;
 use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\TextBlockService;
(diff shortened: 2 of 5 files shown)


lib/Db/TextBlockMapper.php
```diff
@@ -73,4 +73,11 @@ public function findSharedWithMe(string $userId, array $groups) {
 		return $this->findEntities($qb);
 	}
 
+	public function cleanUpUserData(string $userId) {
```
ChristophWurst: nit: we have some other `deleteBy` methods in mappers. This could be a `deleteByUserId` or `deleteByOwner` for consistency

lib/Service/TextBlockService.php
```diff
@@ -149,4 +149,8 @@ public function unshare(int $textBlockId, string $shareWith): void {
 		$share = $this->textBlockShareMapper->find($textBlockId, $shareWith);
 		$this->textBlockShareMapper->delete($share);
 	}
+
+	public function cleanUpUserData(string $userId) : void {
```
ChristophWurst:
```suggestion
	public function cleanUpUserData(string $userId): void {
```

cs fixer might find this

lib/Listener/UserDeletedListener.php
```diff
@@ -51,6 +55,9 @@ public function handle(Event $event): void {
 					'exception' => $e,
 				]);
 			}
+			$this->textBlockService->cleanUpUserData(
```
ChristophWurst: shouldn't it be outside the loop? the result is identical but if a user has x mail accounts there will be x text block clean ups -> x-1 unnecessary queries

tests/Unit/Listener/UserDeletedListenerTest.php
```diff
@@ -31,9 +34,11 @@ protected function setUp(): void {
...
 		$this->accountService = $this->createMock(AccountService::class);
 		$this->logger = $this->createMock(LoggerInterface::class);
+		$this->textBlockService = $this->createMock(TextBlockService::class);
 
 		$this->listener = new UserDeletedListener(
 			$this->accountService,
+			$this->textBlockService,
 			$this->logger
 		);
 	}
```
ChristophWurst, line 44: please assert that the text block clean-up happens

PR #12303, opened 2026-01-23
fix(OutOfOfficeForm): datepicker overlapping with sidebar

GVodyanov:
This is being caused by the main view of the settings having its overflow hidden by the sidebar, and the popup of the datepicker is contained inside the main view and isn't appended somewhere later in the DOM

The ideal solution here would be to have the popup somewhere else where it isn't constrained, but seeing as it comes from a third party library I don't know what we can do about that.

| Before | After |
| - | - |
| image | swappy-20260123_131330 |

### DIFF
--- a/src/components/OutOfOfficeForm.vue
+++ b/src/components/OutOfOfficeForm.vue
@@ -387,4 +387,10 @@ export default {
 		}
 	}
 }
+
+#ooo-first-day {
+	:deep(.mx-datepicker-popup) {
+		inset-inline-start: 0 !important;
+	}
+}
 </style>


src/components/OutOfOfficeForm.vue
```diff
@@ -387,4 +387,10 @@ export default {
 		}
 	}
 }
+
+#ooo-first-day {
+	:deep(.mx-datepicker-popup) {
+		left: 0 !important;
```
ChristophWurst: stylelint wants inset-inline-start. guess that make sense if the input is on the right in RTL mode

PR #12306, opened 2026-01-24
feat(provisioning): add support for Dovecot master user auth

tnxfi:
Nextcloud mail app has an option called "Master password". This name is not to be mixed with Dovecot Master password system.
The functionalitys is slightly misnamed. The correct name should be "Static password for all users".

Dovecot server for pop3, imap and sieve allows various ways of authenticating users. In this case "Static password" means that Dovecot is configured so that every account has the same static password and every account can log in with the same password. That means users like user_ann@example.com and user_dean@example.com will have the same password for their accounts.
Dovecot does allow to limit the usage of static password to specific CIDR. This can be useful in some cases (webmail system and SSO).
But this system is extremely problematic if the users need to be able to use passwords to log in from other system like Thunderbird client software.

Dovecot has a more advanced system called Master passwords, which allows us to configure one or more master users that have their own passwords. For example we could have user masteruser that has password MASTERPASS. When logging into imap, pop3 or sieve, master users can use username in format "user_ann@example.com*masteruser" to log in as any specific user they are allowed to represent. Default separator in the format is * (star sign), but it is configurable in Dovecot. In all other ways the login process is exactly the same as in standard imap, pop3 or sieve login. Only the username has both useraccount and masteraccount specified with the separator.
This configuration is a lot more flexible than static user passwords. Master passwords can also be limited to being accepted from specific CIDRs.
When using this system, the users still can have their own default passwords and the master password system doesn't affect their normal usage in any way. The system also allows granular permissions for master users.

This patch adds the ability to set master user and optionally master user separator character in Nextcloud Mail app. Patch still uses master password field for the password.
If Master user is not set, old style static password is used and only defined user is used as login. If Master user is set, login will use Dovecot style user in logins.

This patch assumes that user can use Dovecot master authentication for all services. Note that Dovecot includes SMTP Submission ability to help with master password login using SMTP.

### DIFF
--- a/lib/Db/Provisioning.php
+++ b/lib/Db/Provisioning.php
@@ -39,6 +39,10 @@ use ReturnTypeWillChange;
  * @method void setMasterPasswordEnabled(bool $masterPasswordEnabled)
  * @method string|null getMasterPassword()
  * @method void setMasterPassword(string $masterPassword)
+ * @method string|null getMasterUser()
+ * @method void setMasterUser(?string $masterUser)
+ * @method string|null getMasterUserSeparator()
+ * @method void setMasterUserSeparator(?string $masterUserSeparator)
  * @method bool|null getSieveEnabled()
  * @method void setSieveEnabled(bool $sieveEnabled)
  * @method string|null getSieveHost()
@@ -72,6 +76,8 @@ class Provisioning extends Entity implements JsonSerializable {
 	protected $smtpSslMode;
 	protected $masterPasswordEnabled;
 	protected $masterPassword;
+	protected $masterUser;
+	protected $masterUserSeparator;
 	protected $sieveEnabled;
 	protected $sieveUser;
 	protected $sieveHost;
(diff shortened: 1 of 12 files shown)


ChristophWurst (commented):
Looks good so far. Thank you very much! We will check it out very soon

tnxfi:
Thank you for pulling this. Sorry for bit of delay on testing this, I've been busy.

I tested pr-12442 in my testing environment and it works correctly against our Dovecot setup. The changes seem ok.

I noticed a small logic issue in the admin UI while testing it: If user deselects "Use master password" and there is content in "Master user" field, settings can't be saved. If the "Master user" field is emptied, save config works. This may be misleading if the user doesn't check for the "Successfully updated..."-toast message. As there is no explicit failure toast, this may lead to some annoyance on wondering why save doesn't work.

ChristophWurst:
The fix is at https://github.com/nextcloud/mail/pull/13076

Issue #12001, opened 2025-11-06
Export and import for user migration


ChristophWurst:
## Feature Request

As a Nextcloud user I want to be able to export my data from Mail and import it on another instance.

## Describe the solution you'd like

Implement a migrator: https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/user_migration.html#register-a-migrator

## Work packages

- [x] MVP implementation (name, email, mail server settings and aliases)
- [x] Migrate special mailboxes
- [x] Migrate account preferences (including internal addresses, trusted senders and text blocks)
- [x] Migrate sieve settings
- [x] Migrate smime
- [x] Migrate quick actions (including all tags)
- [ ] CI workflow to test re-import
- [ ] User manual documentation

## Follow-ups

None

## Out of scope

* Export of provisioned accounts (configuration owned by admin)
* Export of emails and related data (this data is stored on the IMAP server)
* Export of sieve scripts (this data is stored on the Sieve server)

PR #12382, opened 2026-02-06
[WIP] feat(migration): Improve mail import/export functionality

DerDreschner:
# Work in progress!

## Overview
This pull request extends the previous MVP import/export functionality. The following things are now being migrated as well:
- Special mailboxes being set for mail accounts (includes migration of all mailboxes without their messages)
- Sieve settings being set for mail accounts
- Quick actions being created for mail accounts (includes migration of all tags)
- Imported S/MIME certificates and their relation to mail accounts and aliases
- Signatures being set only for specific aliases
- Defined internal addresses
- Defined trusted senders
- Text blocks, but only those not shared by any other group
- General app settings being set by the UI (e.g., view configuration)

## Open tasks
- [ ] Improve the phpdoc `@return array` description regarding the mapping
- [x] Discuss and decide what cases should be tested and how (see this [comment](https://github.com/nextcloud/mail/issues/12001#issuecomment-3860009926))
- [ ] Implement the chosen testing solution (may happen in a separate PR)
- [ ] Squash commits
- [ ] Create user manual documentation

## Open questions by myself
I was confused about the architecture. There are the `mapper` classes, which sometimes are being used via `services`. But sometimes they are used directly (examples: `MailboxMapper`, `ActionStepMapper`, `TagMapper`). When is it advised to create a service class instead of using the mapper directly? :thinking: 

Relates to #12001.


DerDreschner:
I'll close this PR in favor of separate PRs for easier review.

PR #12419, opened 2026-02-13
fix(http): guard against null user in MessagesController

ChristophWurst:
For https://github.com/nextcloud/mail/pull/11224


lib/Controller/MessagesController.php
```diff
@@ -572,6 +608,13 @@ public function export(int $id): Response {
 	 */
 	#[TrapError]
 	public function getHtmlBody(int $id, bool $plain = false): Response {
+		return new TemplateResponse(
```
kesselb: 🤔

PR #12420, opened 2026-02-13
fix(MailTransmission): Handle clients/servers picky about empty headers

joshtrichards:
Came up on the forum:

- https://help.nextcloud.com/t/emails-sent-from-nextcloud-contain-an-empty-cc-field/240195
- https://github.com/simple-login/app/discussions/2676

Changes:

- New function buildHeaders to have same logic applied to sendMessage, saveDraft, saveLocalDraft
- Set Cc and Bcc header on demand
- Drop the detour through ModelMessage
- Drop Date header (done by Horde Mime later)

### DIFF
--- a/lib/Service/MailTransmission.php
+++ b/lib/Service/MailTransmission.php
@@ -14,6 +14,8 @@ use Horde_Imap_Client;
 use Horde_Imap_Client_Data_Fetch;
 use Horde_Imap_Client_Fetch_Query;
 use Horde_Imap_Client_Ids;
+use Horde_Mail_Rfc822_Address;
+use Horde_Mail_Rfc822_List;
 use Horde_Mail_Transport_Null;
 use Horde_Mail_Transport_Smtphorde;
 use Horde_Mime_Exception;
@@ -42,7 +44,6 @@ use OCA\Mail\Exception\ClientException;
 use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\IMAP\IMAPClientFactory;
 use OCA\Mail\IMAP\MessageMapper;
-use OCA\Mail\Model\Message as ModelMessage;
 use OCA\Mail\Model\NewMessageData;
 use OCA\Mail\Service\DataUri\DataUriParser;
 use OCA\Mail\SMTP\SmtpClientFactory;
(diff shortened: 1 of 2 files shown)


lib/Service/MailTransmission.php
```diff
@@ -297,33 +284,22 @@ public function saveDraft(NewMessageData $message, ?Message $previousDraft = nul
...
-		$imapMessage->setContent($message->getBody());
-
-		// build mime body
-		$headers = [
-			'From' => $imapMessage->getFrom()->first()->toHorde(),
-			'To' => $imapMessage->getTo()->toHorde(),
-			'Cc' => $imapMessage->getCC()->toHorde(),
-			'Bcc' => $imapMessage->getBCC()->toHorde(),
-			'Subject' => $imapMessage->getSubject(),
-			'Date' => Horde_Mime_Headers_Date::create(),
```
DerDreschner, line 318: Was the removal of the date incidental or on purpose? If the latter, what happened to it? It's not in the headers with the changed code.

lib/Service/MailTransmission.php
```diff
@@ -197,41 +198,27 @@ public function saveLocalDraft(Account $account, LocalMessage $message): void {
...
 		}
 
 		// build mime body
-		$headers = [
-			'From' => $imapMessage->getFrom()->first()->toHorde(),
-			'To' => $imapMessage->getTo()->toHorde(),
-			'Cc' => $imapMessage->getCC()->toHorde(),
-			'Bcc' => $imapMessage->getBCC()->toHorde(),
-			'Subject' => $imapMessage->getSubject(),
-			'Date' => Horde_Mime_Headers_Date::create(),
```
DerDreschner, line 226: Same question about the date here as well.

lib/Service/MailTransmission.php
kesselb: Horde is taking care of it https://github.com/horde/Mime/blob/94626840e5ceb1aca068ae742d70c984e2874496/lib/Horde/Mime/Mail.php#L424-L426

DerDreschner: Ahh, I see!

PR #12429, opened 2026-02-16
fix: quick search filters

hamza221:
Bugs : 
- Hiding the buttons was quick, which makes the buttons un-clickable most of the time 
- setting `this.searchInTo` to `null`  breaks the app 
- Toggling logic was reversed for attachments and CurrentUser

Changes: 
- Always show active Button (filter) for quick toggling and knowledge about current state


GretaD (approved):
👍

PR #12431, opened 2026-02-17
[main] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (3 unfixable, 33 only fixable manually using --force)


GretaD (commented):
tests need to be adjusted

Issue #10811, opened 2025-03-07
Calendar invite updates, attendance notifications and cancellations are (still) not working


ghost:
I run multiple Nextcloud servers, both private and for work. All my servers run Nextcloud 31.0.0 with Nextcloud Mail 4.2.3. I was very excited for https://github.com/nextcloud/mail/pull/10116 to be merged as it promised to solve all issues regarding calendar invites, and basically all calendar-related bugs were closed by Nextcloud developers saying it would all be resolved by that PR.

Sadly, still not everything works. 

### Steps to reproduce

When I invite a user on my work server from my private server, an e-mail is sent.

Private server:
![Image](https://github.com/user-attachments/assets/1e82e79c-317b-422c-b808-68b3bbcd3210)

Work server:
![Image](https://github.com/user-attachments/assets/32f10916-f006-4e4e-8ec1-e3a7c68bb9dd)

When I click the *Accept* button, the event is added to the calendar of the test account on my work server.

![Image](https://github.com/user-attachments/assets/aa413699-fc39-4389-95ce-712af8931139)

On my private server, I get an e-mail saying the test account on my work server has accepted my invitation.

![Image](https://github.com/user-attachments/assets/7313f0ae-836d-48d6-aa51-29b420d63b6c)

However, in my private calendar, I don't see a checkmark confirming the other party has accepted. It still says "Waiting for reply".

![Image](https://github.com/user-attachments/assets/5cd93a1f-9563-438b-9a6b-7893cf940d40)

When I then change the time of the event by dragging the appointment in my private calendar, an update mail is sent.

![Image](https://github.com/user-attachments/assets/b733172c-bd75-4477-a872-06b4849b8657)

![Image](https://github.com/user-attachments/assets/b999f3f4-3361-4860-bda9-8b320dcb2f0c)

The state remains set to *Accepted*, and there is no way for me to re-accept (or decline) the event now the time has changed.

In the calendar on my work server, the event is also not moved to the new time slot, and remains marked as *Accepted*.

![Image](https://github.com/user-attachments/assets/172a4529-9b72-4e34-9e1f-bdb28367485c)

When I try to add the modified .ics attachment to my calendar, it throws the error `Could not create event`, even though nothing is written to `nextcloud.log`.

![Image](https://github.com/user-attachments/assets/193ce673-5f87-4a5a-b80b-3520d2d1ecdf)

I have to manually delete the event from my calendar, after which the selection bar re-appears in the Mail app.

[...]

PR #12432, opened 2026-02-17
Added email notification upon iMIP message processing failure

Juniar-Rakhman:
This is a follow up changes to a patch done for:
https://github.com/nextcloud/mail/issues/10811

[The patch to address above issue](https://github.com/nextcloud/mail/pull/12129) fixes the processing flow of iMip message processing. When a failure occurs, the batch job logs the error and continues processing the next iMIP message instead of failing the entire job.

**Issue:**
Although the patch resolves the processing problem, failures occur silently from the user’s perspective. The user is not informed when an iMIP message fails to process and therefore cannot take corrective action.

**Solution:**
Created an email notification upon iMIP messaging failure that will be sent to message owner (account's email).

**Added the following changes:**
1. Error recovery with notification instead of failing silently.
2. Added sendErrorNotification() method to generate email body containing instruction for the user on what to do. Also contains technical details on the errors
3. Email notification will be sent to the account's owner.
4. The email contains original message attached so the user can add it manually to their calendar.


Juniar-Rakhman:
Hi @SebastianKrupinski 

That was my initial thoughts too. But not all of our users use nextcloud web.
Most of them are using their own email/calendar clients e.g. thunderbird, mac mail/calendar etc. So email is the only way we could ensure that they get this notification.

ChristophWurst:
Thank you for the pull request @Juniar-Rakhman!

Informing the user makes sense. I would just request that this should happen as a *Nextcloud notification* with user settings. Default should be to receive a *notification* and *no email*.
Admins will be able to change defaults, and users can change the setting to their liking.

https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/notifications.html

Juniar-Rakhman:
@ChristophWurst Thank you for the feedback.

I’ll switch this to a Nextcloud notification with user settings, defaulting to in-app notifications (no email). I will update the PR accordingly.

PR #12481, opened 2026-02-25
fix(imap): Handle unselectable mailboxe

### LINKED ISSUE #12367: IMAP account setup fails on servers returning root "" mailbox and incomplete ACL support (e.g. Xserver)
### Steps to reproduce

When adding an IMAP account in Nextcloud Mail, account setup fails on some IMAP servers
(e.g. Xserver, common in Japan).

The failure happens during folder initialization / STATUS fetching.

### Expected behavior

- Root placeholder / empty mailboxes should be ignored for STATUS
- STATUS should be fetched per mailbox with error isolation
- ACL capability should be treated defensively

### Actual behavior

- IMAP LIST/LSUB returns a root placeholder mailbox `""` (NOSELECT)
- Nextcloud Mail attempts to call STATUS on it
- Server returns NO / BAD
- This aborts account setup

Additionally:
- Server advertises `CAPABILITY ACL`
- But `GETMYRIGHTS` / ACL-related commands are not fully implemented
- This causes further failures during folder ACL fetch

### Mail app version

5.7.0-rc.1

### Nextcloud version

32.0.5

### Mailserver or service

Xserver

### Operating system

Ubuntu24.04.2live server

### PHP engine version

PHP 8.3

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

The modification of FolderMapper.php can fix this issue.

kesselb:
Fix #12367

### DIFF
--- a/lib/IMAP/FolderMapper.php
+++ b/lib/IMAP/FolderMapper.php
@@ -98,11 +98,6 @@ class FolderMapper {
 
 	/**
 	 * @param Folder[] $folders
-	 * @param Horde_Imap_Client_Socket $client
-	 *
-	 * @throws Horde_Imap_Client_Exception
-	 *
-	 * @return void
 	 */
 	public function fetchFolderAcls(array $folders,
 		Horde_Imap_Client_Socket $client): void {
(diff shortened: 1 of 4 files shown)


lib/IMAP/FolderMapper.php
```diff
@@ -111,16 +112,22 @@ public function fetchFolderAcls(array $folders,
 		foreach ($folders as $folder) {
 			$acls = null;
 			if ($hasAcls && !in_array('\\noselect', array_map('strtolower', $folder->getAttributes()), true)) {
-				$acls = (string)$client->getMyACLRights($folder->getMailbox());
+				try {
+					$acls = (string)$client->getMyACLRights($folder->getMailbox());
+				} catch (Horde_Imap_Client_Exception_ServerResponse $e) {
+					// FolderMapper.getFolders can return mailboxes that we are unable to access.
```
ChristophWurst:
```suggestion
					// FolderMapper::getFolders can return mailboxes that we are unable to access.
```

this isn't java :P

ChristophWurst:
```suggestion
					// FolderMapper.getFolders may return mailboxes that could not be accessed.
```

lib/IMAP/FolderMapper.php
```diff
@@ -111,43 +106,31 @@ public function fetchFolderAcls(array $folders,
...
 
-	/**
-	 * @param Horde_Imap_Client_Socket $client
-	 * @param string $mailbox
-	 *
-	 * @throws Horde_Imap_Client_Exception
-	 *
-	 * @return MailboxStats[]
-	 */
-	public function getFoldersStatusAsObject(Horde_Imap_Client_Socket $client,
```
kesselb, line 129: Since https://github.com/nextcloud/mail/pull/12021 we are doing one status command per mailbox. As the performance advantage is gone, there's also no need anymore to have one function that fetches the status for multiple mailboxes.

PR #12489, opened 2026-02-26
fix(cache): expire the IMAP ratelimit keys via TTL

ChristophWurst:
They keys are only needed for 3h but currently stay forever.

### DIFF
--- a/lib/IMAP/HordeImapClient.php
+++ b/lib/IMAP/HordeImapClient.php
@@ -14,6 +14,7 @@ use Horde_Imap_Client_Exception_NoSupportExtension;
 use Horde_Imap_Client_Socket;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\IMemcache;
+use OCP\IMemcacheTTL;
 use function floor;
 
 /**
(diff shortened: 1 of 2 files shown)


tests/Unit/IMAP/HordeImapClientTest.php
DerDreschner, line 1: Are we supposed/allowed to put multiple interfaces and classes into one file? I see it in our projects for the first time 🤔

DerDreschner (approved):
Not sure about the multiple classes in the test file - but besides that I approve!

lib/IMAP/HordeImapClient.php
```diff
@@ -73,11 +80,14 @@ protected function _login() {
 		}
 
 		try {
-			return parent::_login();
+			return $this->imapLogin();
 		} catch (Horde_Imap_Client_Exception $e) {
 			if ($e->getCode() === Horde_Imap_Client_Exception::LOGIN_AUTHENTICATIONFAILED
 				&& $e->getMessage() === 'Authentication failed.') {
 				$this->rateLimiterCache->inc($cacheKey);
+				if ($this->rateLimiterCache instanceof IMemcacheTTL) {
```
kesselb, line 88: I don't know why, but I would prefer something like $counter === null, then cache->add (with the ttl) and inc if != null. 

But we bail out at 3 so the 2 unnecessary set ttl calls are not that bad.

tests/Unit/IMAP/HordeImapClientTest.php
kesselb: Older versions of PHPUnit cannot create a mock from 2 interfaces. Therefore the trick with the interface. For simplicity, we just keep it in one file.

ChristophWurst: Yep, test code is really the exception here

lib/IMAP/HordeImapClient.php
ChristophWurst: Then it is no longer atomic. You could have multiple requests that read null and initialize the counter. This doesn't happen with inc. The concurrent inc would still sum them up.

tests/Unit/IMAP/HordeImapClientTest.php
DerDreschner: Ahh, I see! Understood. Thanks!

PR #12495, opened 2026-02-27
fix(http): prevent NPE by non-null controller parameters

ChristophWurst:
Some controller params were nullable but only in the controller method and not the method passed into. This can lead to a NPE when the param is actually missing from a HTTP request.

For https://github.com/nextcloud/mail/pull/11224

### DIFF
--- a/lib/Controller/AccountsController.php
+++ b/lib/Controller/AccountsController.php
@@ -119,37 +119,21 @@ class AccountsController extends Controller {
 	/**
 	 * @NoAdminRequired
 	 *
-	 * @param int $id
-	 * @param string $accountName
-	 * @param string $emailAddress
-	 * @param string $imapHost
-	 * @param int $imapPort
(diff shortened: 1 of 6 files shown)


lib/Controller/AccountsController.php
```diff
@@ -126,12 +126,12 @@ public function show(int $id): JSONResponse {
...
 	 * @param string $imapSslMode
 	 * @param string $imapUser
-	 * @param string $imapPassword
 	 * @param string $smtpHost
 	 * @param int $smtpPort
 	 * @param string $smtpSslMode
 	 * @param string $smtpUser
-	 * @param string $smtpPassword
+	 * @param string|null $imapPassword
+	 * @param string|null $smtpPassword
```
kesselb: Could also drop the param phpdoc as they don't provide any additional information to the type hints.

Issue #12490, opened 2026-02-26
oc_mail_messages.mail_msg_thrd_root_snt_idx missing on Postgres


ChristophWurst:
### Steps to reproduce

1. Install app on Postgres
2. Run expensive queries

### Expected behavior

Fast results

### Actual behavior

Slow results

The mail_msg_thrd_root_snt_idx was never created on Postgres. The self-join on (mailbox_id, thread_root_id, sent_at) is doing a sequential scan or using only the mailbox_id from the unique index.

Ref bottom of \OCA\Mail\Migration\Version1130Date20220412111833::changeSchema

### PHP engine version

None

### Web server

None

### Database

None

PR #12496, opened 2026-02-27
fix(db): add missing messages table index for Postgres

ChristophWurst:
It's true that Postrges doesn't support prefix indices. But it can just work with the full width. This adds the index also for Postgres, allowing it to run messages table self-join (when listing threads) efficiently.

Fixes https://github.com/nextcloud/mail/issues/12490

### DIFF
--- a/lib/Listener/OptionalIndicesListener.php
+++ b/lib/Listener/OptionalIndicesListener.php
@@ -9,7 +9,7 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Listener;
 
-use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use OCP\DB\Events\AddMissingIndicesEvent;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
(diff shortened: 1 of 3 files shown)


lib/Listener/OptionalIndicesListener.php
```diff
@@ -98,6 +103,25 @@ public function handle(Event $event): void {
 			['user_id', 'email', 'display_name'],
 			false
 		);
+
+		/**
+		 * This index may be missing on Postgres - we add it back on all DBs nevertheless
+		 * @see \OCA\Mail\Migration\Version1130Date20220412111833::changeSchema for the different lengths
+		 */
+		if ($this->connection->getDatabasePlatform() instanceof PostgreSQL94Platform) {
```
DerDreschner: `PostgreSQL94Platform` is deprecated in dbal versions >= 3.2.1. Lets use `PostgreSQLPlatform` instead.

ChristophWurst: yeah, noticed. I was hesitant with replacing because I'd like to backport the change. https://github.com/nextcloud/3rdparty/blob/master/doctrine/dbal/src/Platforms/PostgreSQLPlatform.php is not new, so this should be safe :crossed_fingers:

DerDreschner: It should be safe anyways. The last server version to be incompatible with that change is 25, which uses dbal version 3.1.4. :smile: But it doesn't hurt for the moment, that's right!

PR #12537, opened 2026-03-09
chore: add agents file

ChristophWurst:


AGENTS.md
```diff
@@ -0,0 +1,54 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
kesselb: Hello 2026

ChristophWurst: Damn I even had this in mind when copy pasting

ChristophWurst: I don't like the taste of my own medicine

PR #12542, opened 2026-03-10
fix(ui): restore folder expand state after drag-and-drop

### LINKED ISSUE #5367: Usability issue when moving mail via drag and drop
#### Summary

Currently when dragging a mail the folder view is somehow reacting. It seems that all folders are expanded. When dropping the file, all folders are minimized. The folders that were visible initially are also minified. When moving the next file, all folder get expanded again but the scrollbar is at the very top. Rather cumbersome when you want to move the item to a folder further down in the list.

### Feature Request

Do not change the folder view neither when dragging nor when dropping mails.

A single folder could be expanded when hovering with a drag item over a folder. This is what Windows Explorer or Outlook 365 do, too.

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/5367

cc @powerpaul17 for review and testing

### DIFF
--- a/src/components/NavigationMailbox.vue
+++ b/src/components/NavigationMailbox.vue
@@ -358,6 +358,7 @@ export default {
 			changingSyncInBackground: false,
 			editing: false,
 			showSubMailboxes: false,
+			wasExpandedBeforeDrag: false,
 			menuOpen: false,
 			renameLabel: true,
 			renameInput: false,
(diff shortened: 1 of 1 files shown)


ChristophWurst:
I can still reproduce on main but not here.

1. Open the app
2. Expand a mailbox to see its sub mailboxes
3. Drag a message, don't drop it anywhere

main: sub mailboxes are no longer visible
here: sub mailboxes are still shown

odzhychko (approved):
LGTM

PR #12553, opened 2026-03-12
perf(imap): reuse existing IMAP client/connection for mailbox sync

ChristophWurst:
I've tested new account creation, sync at reload and sync via CLI - works as expected.


kesselb (approved):
Nice 🙏 

I would prefer, for a cleaner API, to make the client required.

PR #12555, opened 2026-03-12
fix: no sent mailbox

hamza221:
Fix https://github.com/nextcloud/mail/issues/4153

### DIFF
--- a/lib/Send/CopySentMessageHandler.php
+++ b/lib/Send/CopySentMessageHandler.php
@@ -7,13 +7,15 @@ declare(strict_types=1);
  */
 namespace OCA\Mail\Send;
 
+use Horde_Imap_Client;
 use Horde_Imap_Client_Exception;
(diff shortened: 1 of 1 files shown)


lib/Send/CopySentMessageHandler.php
```diff
@@ -82,4 +62,18 @@ public function process(
...
 		return $this->processNext($account, $localMessage, $client);
 	}
+
+	private function findOrCreateSentMailbox(Account $account): Mailbox {
+		$sentMailboxId = $account->getMailAccount()->getSentMailboxId();
+
+		if ($sentMailboxId === null) {
+			return $this->mailManager->createMailbox(
+				$account,
+				'Sents',
```
kesselb, line 72:
```suggestion
				'Sent',
```

hamza221: Put that in for testing , I'll change it  after testing

PR #12557, opened 2026-03-12
fix: Avoid undefined array key for address list

kesselb:


lib/Service/PhishingDetection/PhishingDetectionService.php
```diff
@@ -42,7 +42,7 @@ public function checkHeadersForPhishing(Horde_Mime_Headers $headers, array $flag
 		$customEmail = null;
 		$fromHeader = $headers->getHeader('From');
 		if ($fromHeader instanceof Horde_Mime_Headers_Element_Address) {
-			$firstAddr = AddressList::fromHorde($fromHeader->getAddressList(true))?->first();
+			$firstAddr = AddressList::fromHorde($fromHeader->getAddressList(true))->first();
```
kesselb, line 45: Declared self as return type for fromHorder caused a psalm warning about `?` being unless here.

lib/AddressList.php
```diff
@@ -43,14 +42,11 @@ public static function parse($str) {
...
-	 * @param Horde_Mail_Rfc822_List $hordeList
-	 * @return AddressList
 	 */
-	public static function fromHorde(Horde_Mail_Rfc822_List $hordeList) {
-		$addresses = array_map(static fn (Horde_Mail_Rfc822_Address $addr) => Address::fromHorde($addr), array_filter(iterator_to_array($hordeList), static fn (Horde_Mail_Rfc822_Object $obj)
-			// TODO: how to handle non-addresses? This doesn't seem right …
-			=> $obj instanceof Horde_Mail_Rfc822_Address));
+	public static function fromHorde(Horde_Mail_Rfc822_List $hordeList): self {
+		$hordeObjects = iterator_to_array($hordeList, false);
+		$hordeAddresses = array_values(array_filter($hordeObjects, static fn (mixed $obj) => $obj instanceof Horde_Mail_Rfc822_Address));
```
kesselb, line 48: Why changing Horde_Mail_Rfc822_Object to mixed:

Introducing intermediate variables did not only improve the readability, but also made it easier for psalm to unterstand the code. It's currently not documented, that Horde_Mail_Rfc822_List always resolved to Horde_Mail_Rfc822_Objects (it should be, but isn't) and therefore psalm will complain about the typehint. 

The instanceof ensures, as before, that only objects of type Horde_Mail_Rfc822_Address are passed through.

lib/AddressList.php
```diff
@@ -67,12 +63,11 @@ public static function fromRow(array $recipient): self {
 	 *
 	 * @return Address|null
 	 */
-	public function first() {
-		if ($this->addresses === []) {
-			return null;
+	public function first(): ?Address {
+		foreach ($this->addresses as $address) {
```
kesselb, line 67: While 33 has a polyfill for array_first, 32 has not.

tests/Unit/Service/AiIntegrationsServiceTest.php
```diff
@@ -198,7 +196,6 @@ public function testGeneratedMessage(): void {
 		$this->assertEquals($replies, []);
 		$imapMessage->method('isOneClickUnsubscribe')->willReturn(false);
 		$imapMessage->method('getUnsubscribeUrl')->willReturn(null);
-		$addessList->method('first')->willreturn('noreply@test.com');
```
kesselb, line 201:
```
1) OCA\Mail\Tests\Unit\Service\AiIntegrationsServiceTest::testGeneratedMessage
PHPUnit\Framework\MockObject\IncompatibleReturnValueException: Method first may not return value of type string, its declared return type is "?OCA\Mail\Address"
```

The mock is not used -> remove

lib/AddressList.php
ChristophWurst, line 67: I hate that looping and returning is correct for getting the first element 😆

kesselb: Once we drop support for 32, we can use array_first ;)

This will use the same loop for < PHP 8.5, but you won't see it anymore :see_no_evil:

PR #12596, opened 2026-03-18
refactor(ui): make the directives Vue 3 compatible

ChristophWurst:
For https://github.com/nextcloud/groupware/issues/64

Drag-and-drop and infinite scroll still works and this way it's compatible with Vue 2.7 and Vue 3.

AI-assisted: Claude Code + Opus 4.6

### DIFF
--- a/src/directives/infinite-scroll.js
+++ b/src/directives/infinite-scroll.js
@@ -4,222 +4,218 @@
  * SPDX-License-Identifier: MIT
  *
  * Vendored and fixed version of the abandoned https://github.com/ElemeFE/vue-infinite-scroll lib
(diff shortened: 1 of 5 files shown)


src/directives/infinite-scroll.js
```diff
@@ -4,222 +4,212 @@
...
+		if (currentNode.nodeType === 11) {
+			return false
+		}
+		currentNode = currentNode.parentNode
+	}
+	return false
+}
+
+const doBind = function() {
+	if (this.binded) return
```
GretaD: there are couple of if conditions that dont have the {}

GretaD: approved so it doesnt get blocked after fixing these

PR #12615, opened 2026-03-19
fix(ai): handle invalid JSON and possible null for smart replies

ChristophWurst:
Fixes ``Exception: array_values(): Argument #1 ($array) must be of type array, null given in file '/nextcloud/apps/mail/lib/Controller/MessagesController.php' line 926`` on Sentry.

### DIFF
--- a/tests/Unit/Controller/MessagesControllerTest.php
+++ b/tests/Unit/Controller/MessagesControllerTest.php
@@ -1396,4 +1396,239 @@ class MessagesControllerTest extends TestCase {
 
 		$this->controller->index(100, null, null, $limit);
 	}
+
+	public function testSmartReplyNoUser(): void {
+		$controller = new MessagesController(
+			$this->appName,
+			$this->request,
+			$this->accountService,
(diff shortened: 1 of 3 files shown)


tests/Unit/Controller/MessagesControllerTest.php
```diff
@@ -1396,4 +1396,239 @@ public function testRestrictLimit(?int $limit, int $expectedLimit): void {
...
+
+	public function testSmartReplyNoMailbox(): void {
+		$message = new \OCA\Mail\Db\Message();
+		$message->setId(100);
+		$message->setMailboxId(1);
+
+		$this->mailManager->expects($this->once())
+			->method('getMessage')
+			->with($this->userId, 100)
+			->willReturn($message);
```
kesselb, line 1450: No AAA? 😞

ChristophWurst: I stayed consistent with the inconsistency of the rest of the file

PR #12622, opened 2026-03-20
chore: add commit guidelines for agents and users

ChristophWurst:
We don't want to make changes in autopilot mode. Even with agents the user has to review the changes. IMO this is best achieved when the agent only makes the changes and the user handles git.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,32 +1,48 @@
 <!--
-  - SPDX-FileCopyrightText: 2016-2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-FileCopyrightText: 2016-2026 Nextcloud GmbH and Nextcloud contributors
   - SPDX-FileCopyrightText: 2013-2016 ownCloud, Inc.
   - SPDX-License-Identifier: AGPL-3.0-only
 -->
 ## Submitting issues
-
-### Short version
-
- * The [**issue template can be found here**][template] but be aware of the different repositories! See list below. Please always use the issue template when reporting issues.
 
 ### Guidelines
 * Please search the existing issues first, it's likely that your issue was already reported or even fixed.
   - Go to one of the repositories, click "issues" and type any word in the top search/command bar.
   - You can also filter by appending e. g. "state:open" to the search string.
   - More info on [search syntax within github](https://help.github.com/articles/searching-issues)
-* Report the issue using our [template][template], it includes all the informations we need to track down the issue.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -69,3 +69,14 @@ composer test:unit                                    # Run all unit tests
...
+## Git Workflow
+
+Do NOT commit changes unless explicitly asked to.
+
+After completing code changes:
+1. Verify your work is complete and tests pass
+2. Make sure there are no trailing whitespaces
+3. Leave changes in working directory or staged (do not commit)
+4. Provide a summary of what was changed and why
+5. The user will review and commit when ready
```
kesselb: 6. Please suggest the user a commit message usign conventional commits. The last line should be "AI-assisted: Insert name and version of model"

kesselb: Claude told me he would understand, but would recommend to make it more explcit:

```
## Git Workflow
Do NOT commit changes unless explicitly asked to.
After completing code changes:
1. Verify your work is complete and tests pass
2. Make sure there are no trailing whitespaces
3. Leave changes in working directory or staged (do not commit)
4. Provide a summary of what was changed and why
5. Suggest a commit message using Conventional Commits format
6. The user will review and commit when ready

The commit message's last line should be "AI-assisted: Claude Sonnet 4.6"
```

ChristophWurst: The agent can only remind the user, right? This doc is not for users. Maybe something for a contributing.md?

kesselb: We should think about defining a list of valid scopes. That makes it easier for us, but also the models to find then. When it matches use scope, if not leave out.

kesselb: The agent can also suggest a commit message right away?

ChristophWurst: I've pushed something that hopefully helps agents and users

.github/CONTRIBUTING.md
```diff
@@ -1,32 +1,48 @@
...
+* **style**: for coding style changes (formatting only)
+* **ci**: for changes to our CI setup, like workflow files
+* **chore**: for any change that doesn't fit the rest, like development dependency updates
+
+You can also use scopes. Try to have them broad and not specific to a file or ticket. Here are common scopes used:
+
+* **imap**: for changes in the IMAP client
+* **ui**: for changes to the web UI
+
+Usage of AI agents has to be made transparent. Therefore, the commit message's last line before sign-off has to be `AI-assisted: <agent> (model)` for agenting contributions. For example: `AI-assisted: Claude Code (Claude Haiku 4.5)`
```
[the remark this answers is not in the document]
ChristophWurst: Commits have to be signed off.

Issue #10646, opened 2025-02-02
`PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 11605029 bytes) in /var/www/html/nextcloud/apps/mail/lib/Db/MessageMapper.php on line 583`


hboetes:
### Steps to reproduce

I regularly get emails like that from the cronjob: `/usr/bin/php -f /var/www/html/nextcloud/cron.php`. Please tell if you want me to run some debug commands to get more informative output.

### Expected behavior

No memory overload. :-}

### Actual behavior

Memory overload.

### Mail app version

4.1.2

### Nextcloud version

30.0.5

### Mailserver or service

postfix

### Operating system

Fedora

### PHP engine version

PHP 8.3

### Nextcloud memory caching

  'memcache.local' => '\\OC\\Memcache\\Redis',   'memcache.locking' => '\\OC\\Memcache\\Redis',   'memcache.distributed' => '\\OC\\Memcache\\Redis',

### Web server

Nginx

### Database

PostgreSQL

SebastianKrupinski:
Hi @hboetes 

Is there any indication to which background job is causing this? Is there more details in the nextcloud.log? Otherwise this error might be hard to track down.

hboetes:
Would this help? I simply searched for `MessageMapper`  in the log, and then grepped for the matching log entry strings. I'm not a noob, but I have no idea how to debug this problem. Please make some suggestions.

```
root@bran /var/log/nextcloud # grep 1FUn7BlKfr8bgkdCHu8u log
{"reqId":"1FUn7BlKfr8bgkdCHu8u","level":0,"time":"2025-02-02T11:20:00+00:00","remoteAddr":"","user":"--","app":"cron","method":"","url":"--","message":"CLI cron call has selected job OCA\\Notifications\\BackgroundJob\\GenerateUserSettings (id: 10, arguments: null)","userAgent":"--","version":"30.0.5.1","data":{"app":"cron"}}
{"reqId":"1FUn7BlKfr8bgkdCHu8u","level":3,"time":"2025-02-02T11:20:33+00:00","remoteAddr":"","user":"--","app":"PHP","method":"","url":"--","message":"iconv(): Detected an illegal character in input string at /var/www/html/nextcloud/apps/mail/lib/IMAP/Charset/Converter.php#43","userAgent":"--","version":"30.0.5.1","data":{"app":"PHP"}}
[...]

Issue #10179, opened 2024-09-22
Mailbox sync huge memory consumption


stafra:
### Steps to reproduce

1. Install fresh nextcloud from snap (Ubuntu 24.04.1 LTS)
2. Add nextcloud user with "big" mailbox (19 GB) with INBOX size 13 GB, configure it's mailbox usint imap, wait for sync
3. The mailbox sync process dies on low memory even if there is 284 GB memory (incl. swap) configured and available

### Expected behavior

The sync process should take resonable amount of memory.

### Actual behavior

The sync process dies on low memory even if a huge swap used:

```
root@nextcloud:~# /snap/bin/nextcloud.occ mail:account:sync -vvv 3
[debug] Skipping mailbox sync for 37
[debug] Skipping mailbox sync for 45
[debug] Skipping mailbox sync for 41
[debug] Skipping mailbox sync for 8
[debug] Skipping mailbox sync for 40
[debug] Syncing 9
[debug] Locking mailbox 9 for new messages sync
[debug] Locking mailbox 9 for changed messages sync
[debug] Locking mailbox 9 for vanished messages sync
[debug] Running partial sync for 9
[debug] partial sync 3:INBOX - get all known UIDs took 0s. 45/73MB memory used
[debug] partial sync 3:INBOX - get new messages via Horde took 187s. 2411/3093MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 457s. 12818/12841MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 549s. 23179/23202MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 471s. 33575/33597MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 464s. 43970/43993MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 472s. 54345/54368MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 654s. 64680/64703MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 465s. 75076/75099MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 481s. 85473/85496MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 718s. 95794/95817MB memory used
[debug] partial sync 3:INBOX - classified a chunk of new messages took 483s. 106190/106213MB memory used
[...]

Issue #8545, opened 2023-06-14
Account sync performance issue


Mer0me:
### Steps to reproduce

1. Configure several imap accounts with 2GB of emails under MS Exchange 2007
2. Let Nextcloud perform default background sync of these account with system cron
3. See multiple cron processes on the server, as it takes more than 30 minutes to execute

### Expected behavior

Background sync of mail accounts should be quicker

### Actual behavior

A single account with 2GB of mails sync can take several minutes to be executed. Even if we have just synced it a few minutes ago.
When you have 400 accounts and each one take 3 minutes to sync, the background job may takes hours to finish and the nextcloud cronjob cannot be finished before the next one begins, five minutes later.

The old mail server (Exchange 2007) may be partly responsible of the slow sync, but other client like Thunderbird, Outlook or some other webmails (SnappyMail) are displaying this very same mailboxes instantly after the first sync. Each future refresh takes then a few milliseconds to perform.

Syncing with occ command does not show relevant informations (see below)
Disabling the Mail app allow the nextcloud cronjob to be executed in a couple of minutes as it should do for this instance scale I guess.

### Mail app version

3.2.0

### Mailserver or service

Exchange 2007

### Operating system

Debian 11

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

```
root@cloud:/var/www/nextcloud# time sudo -u www-data php8.1 ./occ mail:account:sync -vvv 190
[debug] account is up to date, skipping mailbox sync
[debug] Skipping mailbox sync for 12558
[debug] Skipping mailbox sync for 12560
[debug] Skipping mailbox sync for 12561
[debug] Skipping mailbox sync for 12562
[debug] Skipping mailbox sync for 12563
[debug] Skipping mailbox sync for 12564
[debug] Skipping mailbox sync for 12565
[debug] Skipping mailbox sync for 12566
[debug] Skipping mailbox sync for 16009
[debug] Skipping mailbox sync for 16037
[debug] Skipping mailbox sync for 16369
[debug] Skipping mailbox sync for 16073
[debug] Skipping mailbox sync for 16021
[debug] Skipping mailbox sync for 16022
[debug] Skipping mailbox sync for 16065
[debug] Skipping mailbox sync for 16011
[debug] Skipping mailbox sync for 16019
[debug] Skipping mailbox sync for 16028
[debug] Skipping mailbox sync for 16023
[debug] Skipping mailbox sync for 16018
[...]

Issue #8318, opened 2023-03-30
Unable to Synchronize Large Mailbox


tiongsengis:
### Steps to reproduce

1. Mail users have mailbox greater than 9985 mail in single folder. Nextcloud Mail enter endless loading state when access folder.
2. Configuration is based on IMAP and SMTP
3. When issue command on console i.e. mail:account:sync -vvv <uid>, it shows "4. Range for findAll min=1 max=9985 found= 5026 messages, 5000 left after filtering. Highest UID to fetch is 5000" "Mailbox INBOX has more messages to fetch: 1:5000"

Is it there is a limit set for read 5000 email only? If yes, is it possible to adjust up or remove this limit?

### Expected behavior

Able to sync and present all amount of email supported by email server

### Actual behavior

When click on folders contain large amount of email, Nextcloud mail will present in infinite looping state with word "Loading".
When click on folders contain small amount fo email, able to load and view all email inside accurately.

### Mail app version

2.25

### Mailserver or service

Lotus Domino

### Operating system

Debian 11

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

Nextcloud 25.0.5

ChristophWurst:
> Is it there is a limit set for read 5000 email only? If yes, is it possible to adjust up or remove this limit?

We are testing the app with mailboxes of up to 6 million messages. There is no limit.

tiongsengis:
> Could you share the redacted but full output of the cli sync?
> 
> Sync and indexing happens in chunks of 5k messages to not saturate the database with too many writes at a time. The app should fetch roughly 5k messages, then another 5k and repeat until it has seen the highest UID reported by the server.

Noted. Below is a list. Unable to produce the full because it went on for hours:

root@sghqrmisvsvr001:/var/www/nextcloud# sudo -u www-data php -f occ mail:account:sync -vvv 508
[debug] account is up to date, skipping mailbox sync
[debug] Skipping mailbox sync for 5164
[debug] Skipping mailbox sync for 5165
[debug] Skipping mailbox sync for 5166
[debug] Skipping mailbox sync for 5179
[debug] Syncing 5167
[debug] Locking mailbox 5167 for new messages sync
[debug] Locking mailbox 5167 for changed messages sync
[debug] Locking mailbox 5167 for vanished messages sync
[debug] Running initial sync for 5167
[debug] Initial sync 508:INBOX - mailbox meta search took 1s. 30/32MB memory used
[...]

PR #12633, opened 2026-03-21
feat: Add configurable sync window (mail.max_sync_days) to limit database growth

Rikdekker:
## Summary

Add a new admin setting `mail.max_sync_days` that limits how far back the Mail app synchronizes messages from IMAP to the local database. Default `0` (unlimited) — **no breaking change** for existing installations.

At scale (10,000+ accounts), the Mail database grows unbounded. In our production deployment:
- ~1 billion rows in `mail_messages` (100K messages/account × 10K accounts)
- Database exceeds 200 GB and grows daily
- Initial sync for large mailboxes fails with PHP memory exhaustion
- Background `SyncJob` blocks cron for 80–190 minutes

A 90-day sync window reduces database size by **80–95%** while preserving full access to older messages via IMAP search and on-demand fetching.

### Related issues

- #10179 — Mailbox sync huge memory consumption
- #8545 — Account sync performance issue with many accounts
- #10646 — PHP Fatal error: memory exhausted during sync
- #8318 — Unable to synchronize large mailbox
- #2910 — Loading messages takes "forever"

### 1. Admin setting + UI

Numeric input in **Admin Settings → Mail → Synchronization window**:

```bash
occ config:app:set mail max_sync_days --value=90
```

### 2. Sync filter

When `max_sync_days > 0`, `ImapToDbSynchronizer::runInitialSync()` adds `SEARCH SINCE <date>` (RFC 3501 §6.4.4) via `Horde_Imap_Client_Search_Query::dateSearch()`. Only messages within the window are fetched.

### 3. Daily cleanup job

`SyncWindowCleanupJob` (TimedJob, every 24h) deletes records where `sent_at < now - max_sync_days`. Batched at 500 rows per transaction. No-op when `max_sync_days = 0`.

### 4. Search fallback

When IMAP search returns UIDs outside the local cache, `BackfillSearchResultsJob` fetches those specific messages asynchronously (batches of 200 UIDs).

### 5. On-demand scroll

When a user scrolls past cached messages, `OnDemandSyncJob` fetches the next batch from IMAP asynchronously.

### 6. Database migration

Upgrades the existing `(mailbox_id, uid)` index on `mail_messages` from non-unique to **UNIQUE**. This is required for `INSERT IGNORE` / `ON CONFLICT DO NOTHING` deduplication used by the on-demand sync and search backfill jobs. The migration is idempotent and runs automatically via `occ upgrade`.

### Architecture

```
Admin UI ──→ mail.max_sync_days (oc_appconfig)
                    │
    ┌───────────────┼───────────────────┐
    ▼               ▼                   ▼
Initial Sync    Cleanup Job      Search/Scroll
(SEARCH SINCE)  (daily purge)    (backfill on demand)
```

### Impact

| Deployment | Accounts | Without sync window | With 90-day window |
|---|---|---|---|
| Small | 100 | 5M rows | 1M rows |
| Medium | 1,000 | 50M rows | 10M rows |
| Large | 10,000 | 500M–1B rows | 30–50M rows |

## Files changed

| File | Change |
|---|---|
| `lib/Settings/AdminSettings.php` | Load `max_sync_days` config for admin UI |
| `lib/Controller/SettingsController.php` | `setMaxSyncDays()` endpoint |
| `src/components/settings/AdminSettings.vue` | Number input field |
| `lib/Service/Sync/ImapToDbSynchronizer.php` | `SEARCH SINCE` filter in `runInitialSync()` |
| `lib/BackgroundJob/SyncWindowCleanupJob.php` | **New** — daily cleanup job |
| `lib/Service/Search/MailSearch.php` | On-demand sync + search backfill scheduling |
| `lib/BackgroundJob/OnDemandSyncJob.php` | **New** — async older message fetch |
| `lib/BackgroundJob/BackfillSearchResultsJob.php` | **New** — async search result backfill |
| `lib/BackgroundJob/DeepSyncJob.php` | **New** — background deep-sync |
| `lib/Migration/Version5200Date20260309000000.php` | **New** — UNIQUE constraint on `(mailbox_id, uid)` |
| `appinfo/info.xml` | Job registration |

## Production context

We operate a Nextcloud 32 instance with in the future **10,000+ accounts** syncing against Exchange Online. This feature was developed to address the database growth issues we are experiencing at this scale. It has been tested in our staging environment but is not yet deployed to production — we would like to align with upstream before rolling it out.

## Test plan

- [ ] Set `max_sync_days=0` (default) — verify no behavioral change
- [ ] Set `max_sync_days=90` — verify only recent messages sync
- [ ] Verify `SyncWindowCleanupJob` deletes old records in batches
- [ ] Search for a term that matches old messages — verify `BackfillSearchResultsJob` runs
- [ ] Scroll past cached messages — verify `OnDemandSyncJob` fetches older messages
- [ ] Verify admin UI input saves and loads correctly
- [ ] Verify migration upgrades index to UNIQUE without errors
- [ ] Test with PostgreSQL and MySQL/MariaDB

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/BackgroundJob/OnDemandSyncJob.php
```diff
@@ -0,0 +1,77 @@
...
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\QueuedJob;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Async on-demand sync triggered when a user scrolls past cached messages.
+ *
+ * Instead of blocking the HTTP request with a synchronous IMAP fetch,
+ * MailSearch schedules this job. The frontend retries after a short delay
+ * and picks up the newly synced messages from the local DB.
```
ChristophWurst, line 25: A standard deployment has a cron interval of 5 minutes. That means the asynchronous job will be delayed up to 5 minutes if the cron queue is empty, potentially longer if there are more time critical jobs in the queue.

lib/Db/MessageMapper.php
```diff
@@ -347,6 +364,139 @@ public function insertBulk(Account $account, Message ...$messages): void {
...
+		}
+
+		$existing = [];
+		foreach (array_chunk($uids, 1000) as $chunk) {
+			$query = $this->db->getQueryBuilder();
+			$query->select('uid')
+				->from($this->getTableName())
+				->where($query->expr()->eq('mailbox_id', $query->createNamedParameter($mailbox->getId(), IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT))
+				->andWhere($query->expr()->in('uid', $query->createNamedParameter($chunk, IQueryBuilder::PARAM_INT_ARRAY)));
+			$existing = array_merge($existing, $this->findUids($query));
```
ChristophWurst, line 494: perf: array_merge in a loop is very memory hungy

lib/Service/Sync/ImapToDbSynchronizer.php
```diff
@@ -583,4 +589,151 @@ public function repairSync(
...
 		$perf->end();
 	}
+
+	/**
+	 * Sync older messages on-demand when user scrolls past the sync window.
+	 *
+	 * Uses iterative date ranges (7d -> 14d -> 30d -> 90d) to find messages
+	 * efficiently. Only messages not already in the local DB are persisted.
+	 * Messages are marked as structure_analyzed to skip the PreviewEnhancer's
+	 * second IMAP connection -- structure is loaded lazily when opened.
```
ChristophWurst, line 599: The structure is loaded, true, but the meta data acquired and persisted during preprocessing will then be missing. That is
* Flag if email has attachments
* Preview text
* Flag if it's an IMIP message
* Flag if it's encrypted
* Flag if it's mentioning the user

tests/Integration/Db/MessageMapperTest.php
```diff
@@ -221,7 +221,13 @@ public function testDeleteByUid(): void {
 		self::assertCount(8, $messages);
 	}
 
+	/**
+	 * Skipped: the UNIQUE constraint on (mailbox_id, uid) now prevents
+	 * duplicate inserts at the database level, making this test impossible
+	 * to set up and the deleteDuplicateUids() method unnecessary.
+	 */
 	public function testDeleteDuplicateUids(): void {
+		$this->markTestSkipped('UNIQUE(mailbox_id, uid) constraint prevents duplicate inserts');
```
ChristophWurst, line 230: in that case the test can go

ChristophWurst (commented):
Some comments about the code

ChristophWurst:
> Heads-up: I have some ideas for an alternative approach that reduces database size while keeping UX, sorting and threading intact. I'll draft a ticket shortly.

This is currently blocked by the lack of data. We are trying to find specific hotspots in the current schema to eliminate data selectively. Good candidates are index prefixes for IMAP message IDs, or having preview text and summary expire while keeping all data relevant for threading.

ChristophWurst (changes requested):
Thanks for the contributions.

I understand the concern and the database usage issue. However, this is not an acceptable solution UX-wise nor is it technically sound for general usage of Nextcloud.

The most critical issues are
1. Loading of messages outside the sync window is handled via a background job. These run every 5 minutes typically. At high load times this can be longer. We can not let the user wait that long for search results. Also, there is no frontend for the re-try described in the backend code. This approach could only work with a low latency background task worker. There is a task processing API, but again, the majority of installations won't have that set up. It will run via cron and introduce significant delays.
2. Threading is based on this algorithm: https://www.jwz.org/doc/threading.html. It assumes that we have all emails and their relevant headers. That is currently the case, it will no longer be that way with the proposed changes. That means that threads would be incomplete and split up when the ancestor nodes are missing in the thread tree. It also means that a user will see incomplete threads in the rightmost column.
3. The changes assume the sort order to be *newest first*. That is one of the two sort options of this app. They will no longer see the actually oldest emails but the oldest emails in the sync window. Users will be able to scroll to newer emails, but not to older ones.

If the database usage is a concern, we should look into the actual usage of the tables and their columns. Depending on the mail server there could be some bloat caused be long IMAP message IDs. Those would show in the messages table. https://github.com/nextcloud/mail/issues/12862
Another source for bloat is the message preview text and AI summary. These two columns could definitely have a hot/cold window depending on sort order, where values get set to null when they are cold and we fill the columns only when an email is shown. Tracked in https://github.com/nextcloud/mail/issues/12860.

PR #12635, opened 2026-03-22
style: Move methods

kesselb:


kesselb:
It's the committed result of `npm run lint:fix`

Issue #8351, opened 2023-04-11
Forwarded emails don't contain embedded pictures


LostinSpacetime:
### Steps to reproduce

1. Open an email containing pictures.
2. Forward the email to yourself.
3. In the editor you can see the pictures.
4. Send the email.
5. Wait until you receive the email.
6. Open the email you just received.

### Expected behavior

The email contains text and pictures.

### Actual behavior

The email contains only the text of the forwarded email, while the pictures are missing.

### Mail app version

2.2.5

### Mailserver or service

IMAP/SMTP

### Operating system

Ubuntu 18.04

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

I also tested answering emails with embedded pictures as well by choosing "Edit as new message" and the result was the same. In the message editor, the images were there, while the received email didn't contain any images.

LostinSpacetime:
I tested with the Mail app 3.1.0 and had the same issue.

Frodoriko14:
Still having the issue with the Mail app 3.3.1

alexskrms:
Same here same version 3.3.1.

marcribugent:
The same thing happens to me on version 3.3.1.

ChristophWurst:
As long as the ticket is open it is save to assume nobody has addressed this restriction. Therefore I encourage you to add your :+1: to the report and refrain from commenting.

alexskrms:
@ChristophWurst Any news about this bug? I think this is a very important one that made my colleagues leave nextcloud mail app because if we can't forward images it makes no sense to use it.

miaulalala:
Resulting email:

![image](https://github.com/nextcloud/mail/assets/7427347/4e0480a7-804b-4142-8a97-f8236a087512)

The email in the Sent mailbox (removed the from and to):

```
From:
To: 
Cc: 
Subject: Test
Message-ID: <20240312184619.Horde.XFbyRp1zcOIXy5dE4is1sID@localhost>
User-Agent: Horde Application Framework 5
Date: Tue, 12 Mar 2024 18:46:19 +0000
Content-Type: multipart/related; boundary="=_Q9aV6NBmOL1aycK-hXFzTpT"
MIME-Version: 1.0
Content-Transfer-Encoding: 8bit

This message is in MIME format.

--=_Q9aV6NBmOL1aycK-hXFzTpT
[...]

PR #12637, opened 2026-03-22
test: Add tests for getAttachment and S/MIME

kesselb:
I need to change getAttachment for  https://github.com/nextcloud/mail/issues/8351 to return the actual content type and hence, some coverage for this path would be good. 

AI-assisted: Claude Sonnet 4.6

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -102,6 +102,12 @@ SPDX-FileCopyrightText = "2024 Nextcloud GmbH and Nextcloud contributors"
 SPDX-License-Identifier = "AGPL-3.0-or-later"
 
 [[annotations]]
+path = "tests/data/decrypted-message-body-with-attachment.txt"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "2026 Nextcloud GmbH and Nextcloud contributors"
+SPDX-License-Identifier = "AGPL-3.0-or-later"
+
+[[annotations]]
 path = ["tests/data/mail-filter/*"]
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2024 Nextcloud GmbH and Nextcloud contributors"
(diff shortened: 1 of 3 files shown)


tests/Unit/IMAP/MessageMapperTest.php
```diff
@@ -864,4 +865,67 @@ public function testGetAttachmentAppliesContentTransferEncoding(): void {
...
+		$attachment = $this->mapper->getAttachment(
+			$imapClient,
+			'INBOX',
+			$messageUid,
+			$attachmentId,
+			'alice',
+		);
+
+		$this->assertEquals('report.pdf', $attachment->getName());
+		$this->assertEquals('application/octet-stream', $attachment->getType());
```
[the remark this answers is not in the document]
kesselb: https://github.com/nextcloud/mail/blob/28414219b5bc1b34c04a99d4623b986c7fd2514d/lib/IMAP/MessageMapper.php#L838-L840

REUSE.toml
```diff
@@ -101,6 +101,12 @@ precedence = "aggregate"
 SPDX-FileCopyrightText = "2024 Nextcloud GmbH and Nextcloud contributors"
 SPDX-License-Identifier = "AGPL-3.0-or-later"
 
+[[annotations]]
+path = "tests/data/decrypted-message-body-with-attachment.txt"
```
kesselb, line 105: I'm thinking about adding a new "tests/fixtures" folder and mark the whole folder in resue.toml. So we can just drop fixtures there and don't have to care about the reuse check.

ChristophWurst: We would then just have to make sure nobody drops files with another license

kesselb: Yeah, too much hustle for now :see_no_evil:

Issue #12530, opened 2026-03-06
IMAPMessage violating the IMessage contract


kesselb:
The interface is implemented by `OCA\Mail\Model\Message` and `OCA\Mail\Model\IMAPMessage`.

https://github.com/nextcloud/mail/blob/52214dca31a4be8b3a166350f4ae8b91a0b24d30/lib/Model/IMessage.php#L119-L122

### `OCA\Mail\Model\Message`  is okay:

https://github.com/nextcloud/mail/blob/52214dca31a4be8b3a166350f4ae8b91a0b24d30/lib/Model/Message.php#L291-L305

### `OCA\Mail\Model\IMAPMessage` is not:

https://github.com/nextcloud/mail/blob/52214dca31a4be8b3a166350f4ae8b91a0b24d30/lib/Model/IMAPMessage.php#L385-L391

https://github.com/nextcloud/mail/blob/52214dca31a4be8b3a166350f4ae8b91a0b24d30/lib/IMAP/ImapMessageFetcher.php#L304-L312

https://github.com/nextcloud/mail/blob/52214dca31a4be8b3a166350f4ae8b91a0b24d30/lib/IMAP/ImapMessageFetcher.php#L337-L345

### Solution

Drop the interface?

It looks like we are always using the actual classes instead of the interface.

kesselb:
Wdyt @ChristophWurst @DerDreschner?

PR #12651, opened 2026-03-24
Forward inline images

### LINKED ISSUE #8351: Forwarded emails don't contain embedded pictures

kesselb:
Fix #8351

### Todo

- [x] Merge https://github.com/nextcloud/mail/pull/12676
- [x] Enable also for replies?

### Findings

- `MessageMapper::getAttachment()` previously forced the Content-Type of all attachments to application/octet-stream, except for text/calendar. I assume this was originally introduced as a security hardening.
For forwarding messages with inline attachments, however, preserving the original Content-Type is necessary. Mail clients generally do not render inline parts correctly when they are marked as application/octet-stream, so forwarded inline images would no longer be displayed inline.
- In `AttachmentService::handleForwardedAttachment()`, we previously tried to guess the Content-Type` from the attachment payload. Now that the original MIME metadata is preserved, this is no longer necessary. The trade-off is that we now rely on the provider-supplied MIME headers instead of re-deriving the type from the content, so this effectively trusts the original message metadata without additional validation.
- I added type annotations for attachments and inline attachments, but for now they are only applied to inlineAttachments. Applying the same annotations to attachments currently triggers Psalm warnings because the existing attachment shape is less precise and not yet fully aligned. I think it is better to address that cleanup in a follow-up PR rather than expanding the scope of this change (see also https://github.com/nextcloud/mail/issues/12530).

### getAttachment() consumers

| Consumer | Assesment |
|--------|--------|
| `AttachmentService.handleForwardedAttachment` | No direct browser exposure |
| `MessageController.downloadAttachment` | DownloadResponse forces Content-Disposition: attachment |
| `MessageController.saveAttachment` | No direct browser exposure in Mail |
| `MessageApiController.getAttachment` | Returns attachment data embedded in JSON | 

### Test cases

1. Forward email with attachment (not inline)
2. Forward email with at least two inline images
3. Calendar invitations via mail provider are shown in evolution (The workaround added for https://github.com/nextcloud/mail/issues/10416 was moved)
4. Download attachment
5. Download inline attachment
6. Store attachments to cloud
7. Fetch attachment via api

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -20,7 +20,7 @@ use Horde_Imap_Client_Search_Query;
 use Horde_Imap_Client_Socket;
 use Horde_Mime_Exception;
 use Horde_Mime_Headers;
-use Horde_Mime_Headers_ContentParam;
+use Horde_Mime_Headers_ContentParam_ContentDisposition;
 use Horde_Mime_Headers_ContentParam_ContentType;
 use Horde_Mime_Headers_ContentTransferEncoding;
 use Horde_Mime_Part;
@@ -804,24 +804,15 @@ class MessageMapper {
 
 		$mimePart = new Horde_Mime_Part();
 
-		// Serve all files with a content-disposition of "attachment" to prevent Cross-Site Scripting
-		$mimePart->setDisposition('attachment');
+		$contentId = $mimeHeaders['content-id']?->value_single;
+		if ($contentId !== null) {
+			$mimePart->setContentId($contentId);
+		}
 
-		// Extract headers from part
-		$cdEl = $mimeHeaders['content-disposition'];
-		$contentDisposition = $cdEl instanceof Horde_Mime_Headers_ContentParam
-			? array_change_key_case($cdEl->params, CASE_LOWER)
-			: null;
-		if (!is_null($contentDisposition) && isset($contentDisposition['filename'])) {
-			$mimePart->setDispositionParameter('filename', $contentDisposition['filename']);
-		} else {
-			$ctEl = $mimeHeaders['content-type'];
-			$contentTypeParams = $ctEl instanceof Horde_Mime_Headers_ContentParam
-				? array_change_key_case($ctEl->params, CASE_LOWER)
-				: null;
-			if (isset($contentTypeParams['name'])) {
-				$mimePart->setContentTypeParameter('name', $contentTypeParams['name']);
-			}
+		$contentDisposition = $mimeHeaders['content-disposition'];
+		if ($contentDisposition instanceof Horde_Mime_Headers_ContentParam_ContentDisposition) {
+			$mimePart->setDisposition($contentDisposition->value_single);
+			$mimePart->setDispositionParameter('filename', $contentDisposition['filename'] ?? null);
 		}
 
 		// Content transfer encoding
@@ -830,17 +821,15 @@ class MessageMapper {
 			$mimePart->setTransferEncoding($tmp);
 		}
 
-		/* Content type */
-		$contentType = $mimeHeaders['content-type']?->value_single;
-		if (!is_null($contentType) && str_contains($contentType, 'text/calendar')) {
-			$mimePart->setType('text/calendar');
-			if ($mimePart->getContentTypeParameter('name') === null) {
-				$mimePart->setContentTypeParameter('name', 'calendar.ics');
+		$contentType = $mimeHeaders['content-type'];
+		if ($contentType instanceof Horde_Mime_Headers_ContentParam_ContentType) {
+			if (str_contains($contentType->value_single, 'text/calendar')) {
+				$mimePart->setType('text/calendar');
+				$mimePart->setContentTypeParameter('name', $contentType['name'] ?? 'calendar.ics');
+			} else {
+				$mimePart->setType($contentType->value_single);
+				$mimePart->setContentTypeParameter('name', $contentType['name'] ?? null);
 			}
-		} else {
-			// To prevent potential problems with the SOP we serve all files but calendar entries with the
-			// MIME type "application/octet-stream"
-			$mimePart->setType('application/octet-stream');
 		}
 
 		$mimePart->setContents($body);
--- a/lib/Migration/Version5008Date20260320125737.php
+++ b/lib/Migration/Version5008Date20260320125737.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\DB\Types;
+use OCP\IDBConnection;
+use OCP\Migration\Attributes\ModifyColumn;
+use OCP\Migration\IOutput;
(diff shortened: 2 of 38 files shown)


lib/IMAP/MessageMapper.php
```diff
@@ -685,11 +684,11 @@ public function getAttachments(Horde_Imap_Client_Socket $client,
 		foreach ($structure->partIterator() as $key => $part) {
 			/** @var Horde_Mime_Part $part */
 
-			if (!$part->isAttachment()) {
+			if (count($attachmentIds) === 0 && !$part->isAttachment()) {
```
kesselb: Todo: Double-Check / Revert

src/service/MessageService.js
```diff
@@ -208,6 +208,18 @@ export async function fetchMessageHtmlBody(id) {
 	}
 }
 
+export async function fetchMessageHtmlBodyForCompose(id) {
```
kesselb: Todo: Remove 

_first iteration had a own endpoint and therefore the new method_

lib/Service/HtmlPurify/TransformURLScheme.php
```diff
@@ -125,4 +93,16 @@ private function filterHttpFtp(&$uri, $context) {
 			null
 		);
 	}
+
+	private function replaceCidWithUrl(HTMLPurifier_URI $uri): HTMLPurifier_URI {
+		$inlineAttachment = array_find($this->inlineAttachments, static function ($attachment) use ($uri) {
+			return $attachment['cid'] === $uri->path;
+		});
```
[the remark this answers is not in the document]
kesselb: min-version 32

https://github.com/nextcloud/mail/blob/14e87a15b56ba40daeef8ff328637a80e93afc95/appinfo/info.xml#L56

polyfill for 32

https://github.com/nextcloud/3rdparty/blob/3bf6291ce069bffbaa7b254938f9f048c8f46c38/symfony/polyfill-php84/Php84.php#L68

lib/Model/IMAPMessage.php
```diff
@@ -33,6 +33,15 @@
 
 /**
  * @psalm-import-type MailIMAPFullMessage from ResponseDefinitions
+ *
+ * @psalm-type InlineAttachment = array{
+ *      id: string|null,
```
kesselb, line 38: Todo: remove null for id, add safe-guard in fetcher. Not having a mime id is not a valid state.

lib/Migration/Version5008Date20260320125737.php
```diff
@@ -0,0 +1,51 @@
...
+]
+class Version5008Date20260320125737 extends SimpleMigrationStep {
+	/**
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 */
+	#[Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('mail_attachments')) {
```
ChristophWurst: nit: invert condition and use early null return. pro: one indentation less

lib/Migration/Version5008Date20260320125737.php
```diff
@@ -0,0 +1,51 @@
...
+			if (!$attachmentsTable->hasColumn('content_id')) {
+				$attachmentsTable->addColumn('content_id', Types::STRING, [
+					'notnull' => false,
+				]);
+			}
+			if (!$attachmentsTable->hasColumn('disposition')) {
+				$attachmentsTable->addColumn('disposition', Types::STRING, [
+					'notnull' => false,
+				]);
+			}
```
ChristophWurst: can they have lengths to avoid unnecessarily wide columns?

ChristophWurst (commented):
The approach and the code changes make sense

Preliminary 👍 

Giving it a test now

ChristophWurst:
- [x] BUG triggered by plain-text emails. I clicked through my inbox. Two had this:

```
OCA\Mail\Controller\MessagesController::enrichAttachments(): Argument #2 ($attachments) must be of type array, null given, called in nextcloud/apps/mail/lib/Controller/MessagesController.php on line 263 in file 'nextcloud/apps/mail/lib/Controller/MessagesController.php' line 1048 
```

ChristophWurst:
image

Possibly not new but this is the visual state for scheduled/stuck emails in the outbox.

ChristophWurst (approved):
Tested regular attachment and inline attachment forwarding. Works as expected except the quirks commented above :)

lib/Migration/Version5008Date20260320125737.php
kesselb: For disposition we can use 10. 

For content_id, there's no limit.

Symfony Mime uses `16 characters@symfony` . 
Horde `23 character@hostname`. 

I'm going with 255, but could also use a lower value.

ChristophWurst: The lower the better :+1:

ChristophWurst:
This is good to go IMO, only the version needs to be rebumped. @hamza221 final testing round?

kesselb:
> Possibly not new but this is the visual state for scheduled/stuck emails in the outbox.

I'm taking care in a follow up.

lib/Service/Attachment/AttachmentService.php
```diff
@@ -365,32 +376,23 @@ private function handleForwardedMessageAttachment(Account $account, array $attac
 	private function handleForwardedAttachment(Account $account, array $attachment, \Horde_Imap_Client_Socket $client): ?int {
```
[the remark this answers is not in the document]
kesselb: True

lib/Service/TransmissionService.php
```diff
@@ -75,10 +75,20 @@ public function handleAttachment(Account $account, array $attachment): ?Horde_Mi
...
 			$part = new Horde_Mime_Part();
 			$part->setCharset('us-ascii');
-			if (!empty($localAttachment->getFileName())) {
-				$part->setDisposition('attachment');
+
+			if ($localAttachment->isDispositionAttachmentOrInline()) {
+				$part->setDisposition($localAttachment->getDisposition());
+				/*
+				 * Setting a name implicitly adds a Content-Disposition header in Horde,
+				 * which would override the intentional omission. Only set it for attachment/inline dispositions.
```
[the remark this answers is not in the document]
kesselb: Migration is now adding disposition attachment for existing attachments.

kesselb:
Follow ups:

- [ ] Possibly not new but this is the visual state for scheduled/stuck emails in the outbox.
- [ ] Version5008Date20260320125737 is annotated #[ModifyColumn(…)], but the changeSchema body calls addColumn twice. The correct attribute is #[AddColumn]. 
- [ ] _DISPOSITION_OMIT = null — the constant works in PHP, but a string sentinel like DISPOSITION_OMIT = 'omit' avoids the edge case of isNull() checks in DB queries needing to agree with PHP-side comparisons. The current migration backfills correctly so it's fine as-is, just slightly unusual._ (undecided)

PR #12657, opened 2026-03-25
fix(mention): custom range for usernames and error handling

hamza221:
Fixed Bugs: 
1. when searching for a user after typing `@` and your query is longer than 5 characters the user is added to `to` but the text is never replaced 
2. If your results contain contacts with `null` emails you get no suggestions and it fails silently 
3. Error handling for the http request 

Should be probably generalized for all triggers mainly Text blocks and links

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -179,6 +179,7 @@ export default {
 		return {
 			linkTribute: null,
 			emojiTribute: null,
+			contactMentionQuery: null,
 			textSmiles: [],
 			ready: false,
 			editor: DecoupledEditor,
@@ -245,8 +246,15 @@ export default {
 			if (text.length === 0) {
 				return []
 			}
-			let contactResults = await autoCompleteByName(text)
-			contactResults = contactResults.filter((result) => result.email.filter((email) => email.length > 1).length > 0)
+			this.contactMentionQuery = text
+			let contactResults
+			try {
+				contactResults = await autoCompleteByName(text)
+				contactResults = contactResults.filter((result) => result.email?.filter((email) => email.length > 0))
+			} catch (error) {
+				logger.error('could not fetch contacts to mention', { error })
+				return []
+			}
 			return contactResults
 		},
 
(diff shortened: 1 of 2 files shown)


src/components/TextEditor.vue
```diff
@@ -245,8 +246,15 @@ export default {
...
 			}
-			let contactResults = await autoCompleteByName(text)
-			contactResults = contactResults.filter((result) => result.email.filter((email) => email.length > 1).length > 0)
+			this.contactMentionQuery = text
+			let contactResults
+			try {
+				contactResults = await autoCompleteByName(text)
+				contactResults = contactResults.filter((result) => result.email?.filter((email) => email.length > 0))
+			} catch (error) {
+				console.error(error)
```
ChristophWurst: use the logger with a meaningful log message. log the error as context

src/components/TextEditor.vue
```diff
@@ -464,7 +472,8 @@ export default {
 						})
 				}
 				if (eventData.marker === '@') {
-					this.editorInstance.execute('insertItem', { email: item.email[0], label: item.label }, '@')
+					const lookback = this.contactMentionQuery?.length + 1 || 5 // +1 for the '@' itself
```
ChristophWurst: Works, but adding +1 to a possible undefined and then falling back to another value is hard to read

```js
> const obj = {}
undefined
> obj?.foo + 1 || 5
5
```

src/components/TextEditor.vue
```diff
@@ -464,7 +472,8 @@ export default {
 						})
 				}
 				if (eventData.marker === '@') {
-					this.editorInstance.execute('insertItem', { email: item.email[0], label: item.label }, '@')
+					const lookback = this.contactMentionQuery?.length ? this.contactMentionQuery.length + 1 : 5// +1 for the '@' itself
```
ChristophWurst, line 475:
```suggestion
					const lookback = this.contactMentionQuery?.length ? this.contactMentionQuery.length + "@".length : 5 //
```

ChristophWurst (approved):

Issue #9311, opened 2024-02-09
Not all recipient addresses are displayed


handkerchief333:
### Steps to reproduce

1. Send an e-mail to approx. 6 or 7 different recipients.
2. View the sent e-mail after sending
3. Only about 3 or 4 e-mail addresses are displayed above the e-mail text. In addition, you cannot see who was addressed as the main recipient and which of them was addressed as Cc or Bcc.

### Expected behavior

When viewing e-mail messages, all e-mail addresses involved and the exact use (e.g. Cc or Bcc) should be visible.

### Actual behavior

The remaining recipients and more detailed information can only be seen if you look at the source or download the message. 

### Mail app version

3.5.4

### Operating system

FreeBSD 13.2

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

It wouldn't be a bad idea to handle it like Gmail. All email addresses are visible there and can be easily copied and reused.

handkerchief333:
Bcc addresses are also not visible in the source and as an eml file.

ChristophWurst:
Done via https://github.com/nextcloud/mail/pull/12666

PR #12666, opened 2026-03-27
feat(thread): show per-message To/Cc/Bcc recipients in thread view

### LINKED ISSUE #11305: Display all recipients and basic information for each individual email in the email view.

With a single email message, it is not immediately apparent who the email was actually sent to, e.g. Cc, without downloading the email and opening the .eml file.

When you can view the entire email header at a glance or with a single click (for space reasons) without having to download the .eml file.

Here, too, I would suggest the same as for Gmail: simple but very effective.

Image

1. In an email conversation, the email addresses involved are displayed at the top of the right-hand column, but these are usually based only on the first email in the entire conversation. However, it is common for certain emails in the conversation to have different recipients in the Cc field than others, etc. This means that this display is not always accurate and does not refer to a single email in the entire conversation. This address display is therefore not particularly useful.
2. If there are multiple recipients, the “Reply All” option (button) is displayed, indicating that there are multiple recipients, but you can only see them when you click on “Reply All” and a draft message opens. This is quite a workaround if you just want to know the recipients of a single message.
3. View source or download message (eml file) – both are very inconvenient if you just want to quickly read the basic information in the email and are only a workaround.

### Additional context

_No response_

sturlan:
## Summary

  Closes #11305, #9311, #4258.

  Recipients (To, Cc, Bcc) are already returned by the API for every message
  in a thread but were never displayed in `ThreadEnvelope`. This PR surfaces
  them in the expanded message header using the existing `RecipientBubble`
  component (which provides the NcPopover contact card with reply/copy/add
  contact actions).

  Also removes the aggregate participant bubble list from the thread header
  (`Thread.vue`) per the design spec in #4258 — per-message recipients in
  the expanded view replace that approach.

  Prior attempt: #4315 (2021, closed 2023).

  ## Changes

  - `ThreadEnvelope.vue`: render To/Cc/Bcc as `RecipientBubble` rows when
    message is expanded; show sender email only when expanded; apply
    `--color-text-maxcontrast` to sender name in expanded state
  - `Thread.vue`: remove aggregate participant bubble header; remove unused
    `currentAccountEmail` prop
  - `MailboxThread.vue`: remove unused `:current-account-email` binding

  ## Test plan

  - [ ] Open a thread with multiple messages
  - [ ] Expand a message → To/Cc/Bcc rows appear with clickable bubbles
  - [ ] Collapse the message → recipient rows disappear
  - [ ] Click a recipient bubble → NcPopover opens (reply, copy email, add contact)
  - [ ] `npm run test:unit` — all 220 tests pass
  - [ ] `npm run lint` — no errors

  AI-assisted: Claude Code (claude-sonnet-4-6)


src/components/ThreadEnvelope.vue
```diff
@@ -1366,6 +1398,28 @@ export default {
 		inset-inline-start: var(--default-grid-baseline);
 	}
 
+	.envelope__recipients {
```
ChristophWurst, line 1438: style: this could be broken up into a nested scss structure

src/components/ThreadEnvelope.vue
```diff
@@ -1366,6 +1398,28 @@ export default {
...
+		padding-inline: 60px 38px;
+		padding-block: 4px;
```
ChristophWurst: nit: magic numbers. ideally we can leverage the grid spacing var for this. See https://github.com/nextcloud/mail/issues/9944.

ChristophWurst (commented):
Thank you for the contribution!

Code looks good so far. Please provide before/after screenshots

nimishavijay:
Sorry for the delay in response, but this is really cool! Something I would definitely use! My main concern with this is the space it's taking up. If I understood correctly, this information is shown all the time when the message is expanded, and it may not be relevant always. What do you think of the following changes to address that issue?

- [ ] We don't show this by default, instead we have a small indication next to the recipient details like a chevron. Clicking on the recipient email or the chevron toggles showing this info
- [ ] We use smaller user bubbles, with height as `var(--clickable-area-small)` (24px), remove the padding and add a border-radius-pill to make it rounded completely

Result would look something like this: 

image

What do you think? :)

sturlan:
> What do you think? :)

I think it's a great idea. Here it the screenshot of the new design

Screenshot 2026-04-07 at 14-59-29 Mail - Nextcloud

nimishavijay:
Nice! :) Only some small points:
- [ ] I suspect having the sender name and email on the same line will obscure a lot of info on mobile, so we can move it to 2 lines as it was before (there's definitely the space as the avatar is big anyway)
- [ ] The height of each item in the to/cc/bcc section is still more than ideal. Any way to reduce it?
- [ ] We can add a little bit more spacing between the to/cc/bcc section and the mail body for better separation. Maybe 8px more? (`calc(var(--default-grid-baseline) * 2)`)

sturlan:
> Nice! :) Only some small points:
> 
>     * [ ]  I suspect having the sender name and email on the same line will obscure a lot of info on mobile, so we can move it to 2 lines as it was before (there's definitely the space as the avatar is big anyway)
> 
>     * [ ]  The height of each item in the to/cc/bcc section is still more than ideal. Any way to reduce it?
> 
>     * [ ]  We can add a little bit more spacing between the to/cc/bcc section and the mail body for better separation. Maybe 8px more? (`calc(var(--default-grid-baseline) * 2)`)

Thanks for the feedback. I agree with your insights, will get to them in a few days as i got some stuff that needs to be done.

sturlan:
@nimishavijay about you referenced --clickable-area-small (24px) as the target height, pill border-radius, and no trailing padding.

I've added the CSS overrides for NcUserBubble, but there's a wrinkle: NcUserBubble sets height as an inline style (height: ${size}px), which takes precedence over our scoped CSS. So the chip stays at the height determined by the :size prop rather than our CSS variable.
A few options:
1. Pass :size="24" to RecipientBubble — this makes the inline style height: 24px, which matches --clickable-area-small. Simple, but hardcodes the value instead of using the CSS variable.
2. Upstream a fix to @nextcloud/vue — add a prop or CSS custom property to NcUserBubble that lets consumers control chip height cleanly.

What would you prefer? Happy to go with option 1 for now if option 2 is a longer-term thing. Also could leave it as is.
I changed the other points you mentioned, here are the screenshots.

Screenshot 2026-04-15 at 20-27-23 Mail - Nextcloud
Screenshot 2026-04-15 at 20-27-56 Mail - Nextcloud

Looking forward to hear from you!
Regards,

jancborchardt (commented):
@sturlan in the philosophy of the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), it would make sense to go for the direct approach given in solution 1 for now.

sturlan:
Hi all,
I hope this is the end of it :)

Bubble and avatar at 24px, with 5px padding
Screenshot 2026-04-21 at 08-38-14 Mail - Nextcloud
Screenshot 2026-04-21 at 08-37-59 Mail - Nextcloud

jancborchardt (approved):
Seems good! FYI @nimishavijay if further polishing is needed, it can be done in a follow-up.

ChristophWurst (commented):
Code Quality & Issues

  1. Bug — Stale method call (blocking)
  - Thread.vue:220 — this.updateParticipantsToDisplay() must be removed. This is the CI failure.

  2. Duplicate :key risk in recipient lists
  - The v-for loops use :key="recipient.email" but the same email can appear multiple times in a To/Cc/Bcc list (unlikely but possible with malformed messages). Consider using the index as part of the key: :key="recipient.email + '-' +
  index" or just :key="index".

  3. !important overrides in scoped styles
  - The .envelope__recipients section uses several !important declarations to override NcUserBubble internal styles. This is fragile and could break when @nextcloud/vue updates. The existing RecipientBubble.vue already had :size="26"
  removed in this PR — worth checking if there's a prop-based approach instead of deep style hacking.

  4. Unrelated change in RecipientBubble.vue
  - The removal of :size="26" from RecipientBubble.vue affects all usages of RecipientBubble across the app (e.g., in the composer), not just the thread view. This should either be reverted (and the size override kept local to the new
  recipient section) or explicitly noted as an intentional global change.

  5. Button inside <router-link> (a11y concern)
  - The toggle button (lines 67-76) is nested inside a <router-link> that has @click.native.prevent. While @click.stop.prevent on the button prevents propagation, nesting interactive elements (<button> inside <a>) is invalid HTML and can
  cause accessibility issues with screen readers.

sturlan:
@ChristophWurst bump

ChristophWurst:
Please have some patience. I'm doing my best to provide human reviews.

ChristophWurst:
Have the earlier review points been addressed?

The force push made it impossible to see a diff to the old version

sturlan:
> Please have some patience. I'm doing my best to provide human reviews.

I understand, sorry for pushing.

sturlan:
> Have the earlier review points been addressed?
> 
> The force push made it impossible to see a diff to the old version

yes, in commit dcbf341

fix(thread): address PR review comments
- Remove stale updateParticipantsToDisplay() call (CI fix)
- Add size prop to RecipientBubble (default 26) to restore pre-PR behaviour
- Pass :size="24" to recipient bubbles via prop instead of !important CSS hacks
- Fix duplicate :key risk in recipient v-for loops with index suffix
- Replace router-link.left with div + role="button" to fix button-in-<a> a11y violation
- Add keyboard handlers (Enter/Space) on envelope header div
- Fix @click.stop on unsubscribe button (latent click-propagation bug)

sturlan:
@ChristophWurst  no worries if you're swamped, but curious — what's the typical turnaround for PRs here? Mine's been open a while and I want to make sure I haven't missed a step in the process.

ChristophWurst:
Thanks and sorry for the delay

sturlan:
@ChristophWurst thank you for closing the PR. Can you close the other two issues with this PR?
https://github.com/nextcloud/mail/issues/9311
https://github.com/nextcloud/mail/issues/4258

PR #12605, opened 2026-03-18
feat: add delegation backend

hamza221:
ref https://github.com/nextcloud/mail/issues/12403

Use https://github.com/nextcloud/mail/pull/12698 for testing 

image

- [x] Unit tests 
- [x] Fix integration tests 
- [x] Add possibility to leave a delegation (followup- not part of mvp) 
- [x] Notifications
- [x] Audit logs 
- [ ] To discuss : Allow account deletion ?

Issue #12403, opened 2026-02-12
Account delegation


ChristophWurst:
## Is your feature request related to a problem? Please describe.

As a Nextcloud Mail user I want to delegate my full account to another Nextcloud user. This allows them to act on behalf of me. This feature could be used for assistants and vacation replacements.

## Describe the solution you'd like

Add a UI to pick other users as delegatees. These users will then see my account in their left sidebar, and use it as if it were their own. That means they can send emails under my name.
For transparency it should be somehow indicated that it was acted on behalf of someone else.

## Work packages

* [ ] Account delegation
  * UI for delegator (delegate, list, undelegate) #12727
  * Backend for delegation #12727
  * UI for delegatee (indicator that an account is delegated) #12727
* [ ] Audit log (for protocol when acting on behalf) Backend for delegation #12729
* [ ] Notifications #12728

## Describe alternatives you've considered

* https://github.com/nextcloud/mail/issues/572

## Additional context

* This is the Mail equivalent of https://github.com/nextcloud/calendar/issues/2706
* https://github.com/nextcloud/server/issues/51803 could become relevant

kra-mo:
The main UI for adding a delegate could look like so:

<img width="500" height="276" alt="Image" src="https://github.com/user-attachments/assets/a5e85f11-6eee-4d84-b34f-4b234d35c5ff" />

- [ ] There is no placeholder if no accounts are delegated. The context is enough.
- [ ] The tooltip for the `X` button is "Revoke access".
- [ ] When the `X` button is clicked, a confirmation dialog appears: **Revoke access?** {email address} will no longer be able to act on your behalf

When clicking "Add delegate", the following dialog appears:

<img width="449" height="202" alt="Image" src="https://github.com/user-attachments/assets/2c5bf4a7-cc05-4a3c-973f-447826fc3b7e" />

Then, after selecting someone, the user receives an email informing them about the delegation.

Delegated mail accounts would show up similarly to other accounts for the delegatee, but they would have the subline "Delegated" in Mail settings.

If there are more requirements, like custom permissions, etc., please tell me :)

hamza221:
Should we make it possible to leave a delegation ?

ChristophWurst:
Let's consider it a follow-up. We have to check if this is a common feature in other groupware.

johnnyq:
We should also be able to delegate to a group / team for example shared mailboxes support@domain.ext info@domain.ext

johnnyq:
[...]

PR #12698, opened 2026-04-06
feat: add delegation frontend

hamza221:
STATUS: I'll wait for the backend to get merged before continuing with this 

requires #12605
ref https://github.com/nextcloud/mail/issues/12403

| 1 | 2 | 3 |4 |
|--------|--------|--------|--------|
| image | image | image |image |
image| 

- [x] Design review 
- [x] hide action on delegated accounts


nimishavijay (changes requested):
Awesome! Looks really nice! Only some small changes. Overall:
- [x] The icon for delegation (across the board) can be [supervisor account](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:supervisor_account:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=super). The one you've chosen is also very similar but we use that for indicating users and groups
- [x] Instead of "Delegation", I'm wondering if we can use the wording "Delegate account" in the 3 dot menu and the modal title
- [x] Use tertiary buttons for all the "Cancel" buttons and don't use the icon.

In the "Delegation" modal:
- [x] There is some whitespace on top which shouldn't be there (somewhow not there in the "Add delegates" modal)
- [x] We can also center align the "Delegates" heading. Ik in @kra-mo's mockups it was left aligned, but because in the component we always have the heading in the center it looks inconsistent now. 
- [x] I see that there's a `h3 { font-weight: bold; }` which makes the "Delegates" heading stronger than its parent heading. We could remove that

In the "Add delegates" modal:
- [x] the select component should be full width

kra-mo:
> We can also center align the "Delegates" heading.

I was going off of form group components, not a dialog, in this context, I think the two different headings don't make sense either, we could just have one "Delegates" title :)

kra-mo (changes requested):
Adding to what @nimishavijay said, the "Revoke" button should also probably not have an icon, a checkmark with danger styling is always confusing :)

Additionally, the user selection dropdown should be full-width and there should be a bit more padding between the explanation text beneath and the buttons, also in the revoke access dialog.

hamza221:
will continue in #12727

PR #12706, opened 2026-04-09
fix(l10n): Accidentally translated placeholder name

helmo:


ChristophWurst (changes requested):
Thank you for your contribution!

Translations are submitted exclusively via the Transifex platform. Any direct modifications to git would be overwritten and lost at the next nightly sync.

So please submit the fix there.

ChristophWurst:
Closing per https://github.com/nextcloud/mail/pull/12706#pullrequestreview-4080484385

PR #12711, opened 2026-04-12
feat: jmap support - part 1

### LINKED ISSUE #12609: JMAP Protocol Abstraction and Routing Infrastructure
## Summary

Introduce a protocol abstraction layer that enables the service layer to route operations to 
either IMAP or JMAP client factories based on the account's configured protocol.

## Problem

The service layer is currently tightly coupled to IMAP. To support multiple protocols, 
the codebase needs a way to route client instantiation and key operations based on account configuration.

## Solution

- Add `protocol_type` column to `mail_accounts` table (default: 'imap' for existing accounts)
- Create `ProtocolClientFactory` dispatcher
- Refactor existing code to use dispatcher for client instantiation
- Update account creation API to accept protocol parameter
- Ensure existing IMAP accounts continue working

## Implementation

Database migration adds `protocol_type` enum column ('imap' | 'jmap', default: 'imap').

Services query account protocol before instantiating clients:
```php
$client = $this->protocolClientFactory->getClient($account);
```

The dispatcher checks `$account->getProtocol()` and returns the appropriate client factory.

Account API endpoint accepts optional `protocol` parameter (defaults to 'imap').

## Affected Components

- Database: migration for `mail_accounts` table
- `MailAccount` entity: add protocol field
- `MailManager`, `SyncService`: use dispatcher
- `AccountsController`: accept protocol parameter
- New: `ProtocolClientFactory` dispatcher

## Out of Scope

- Authentication method selection (OAuth, etc.)
- Protocol auto-detection

SebastianKrupinski:
### Summary
- Resolves: https://github.com/nextcloud/mail/issues/12609
- Added php-jmap-client dependency
- Modified test workflow to start additional mail server with jmap support
- Added jmap specific console command to create account (mail:account:create-jmap)
- Added jmap specific console command to create account (mail:account:create-jmap)
- Modified existing imap console command to create account (mail:account:create-imap) with alias to old command (mail:account:create)
- Modified existing imap console command to update account (mail:account:update-imap) with alias to old command (mail:account:create)
- Added unified console test command (mail:account:test) with alias to old (mail:account:diagnose)
- Added a protocol factory to create the correct client and connectors
- Added database migrations


ChristophWurst (approved):

lib/Command/CreateImapAccount.php
```diff
@@ -5,7 +5,7 @@
 /**
  * SPDX-FileCopyrightText: 2016-2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-FileCopyrightText: 2014-2016 owncloud, Inc.
- * SPDX-License-Identifier: AGPL-3.0-only
+ * SPDX-License-Identifier: AGPL-3.0-or-later
```
ChristophWurst: Revert. I don't think we can do this retroactively

ChristophWurst (changes requested):

lib/Protocol/ProtocolFactory.php
```diff
@@ -0,0 +1,113 @@
...
+	/**
+	 * @template T
+	 * @param Account $account
+	 * @param class-string<T> $interface
+	 * @return T
+	 * @throws ServiceException
+	 */
+	private function resolveConnector(Account $account, string $interface): mixed {
+		$protocol = $account->getMailAccount()->getProtocol();
+		$class = self::CONNECTOR_MAP[$protocol][$interface] ?? null;
```
kesselb, line 105: Does that work with the map commented out? 🤔

lib/Migration/Version5011Date20260401000001.php
```diff
@@ -0,0 +1,52 @@
...
+		$accountsTable = $schema->getTable('mail_accounts');
+		if (!$accountsTable->hasColumn('protocol')) {
+			$accountsTable->addColumn('protocol', Types::STRING, [
+				'length' => 16,
+				'default' => 'imap',
+				'notnull' => true,
+			]);
+		}
+		if (!$accountsTable->hasColumn('path')) {
+			$accountsTable->addColumn('path', Types::STRING, [
```
kesselb, line 45:
```suggestion
			$accountsTable->addColumn('jmap_path', Types::STRING, [
```

Prefix the fields for jmap if they are unused for imap?

lib/Protocol/ProtocolFactory.php
SebastianKrupinski: Yes it works with the map commented out, 

Part 2 has the actual references.

lib/Migration/Version5011Date20260401000001.php
SebastianKrupinski: I left this generic on purpose, if we add another protocol this field can be reused.

lib/Command/CreateImapAccount.php
SebastianKrupinski: That's odd it says I already changed this, 2 days ago

image

SebastianKrupinski: Or do you mean it should be "AGPL-3.0-only"

lib/Migration/Version5800Date20260401000002.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	#[\Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+		$mailboxesTable = $schema->getTable('mail_mailboxes');
+		if (!$mailboxesTable->hasColumn('rpid')) {
+			$mailboxesTable->addColumn('rpid', Types::STRING, [
```
kesselb: I'm a fan of self-speaking variable names, and hence I would prefer a name that is easy to match to the specification.

rpid = remote_parent_id
r = remote_id

Looking at the sample at https://jmap.io/spec/rfc8621/#section-2.6, state is outside the list and should be the same for all mailboxes, or does that state depends on the requested mailboxes?

lib/Migration/Version5011Date20260401000001.php
```diff
@@ -0,0 +1,52 @@
...
+use OCP\DB\Types;
+use OCP\Migration\Attributes\AddColumn;
+use OCP\Migration\Attributes\ColumnType;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * @psalm-api
+ */
+#[AddColumn(table: 'mail_accounts', name: 'protocol', type: ColumnType::STRING)]
```
kesselb, line 23: Please also list path so the list is complete

kesselb:
Please correct the license in your composer.json: https://github.com/SebastianKrupinski/jmap-client-php/blob/30dce48d854abc7d416f3784bc796b379775bdd3/composer.json#L7

kesselb:
jmap-client currently requires guzzlehttp as dependency.

Issue 1: While guzzle is already a dependency for server via 3rdparty, guzzle will be installed another time here leadiing to another copy of guzzle.

Issue 2: Not using the nextcloud http client means our protections like ssrf, dns pinning, configuration of proxies through config.php are not applied.

Solution:

- jmap-client: Add https://packagist.org/packages/psr/http-client, remove guzzle, change change to code to use the interface.
- mail: Currently only the http client in Nextcloud 34 also implements the client interface, hence we need a wrapper in mail to exposes a http client implementing ClientInterface. For example like https://github.com/nextcloud/bookmarks/blob/master/lib/Http/Client.php.

lib/Command/CreateImapAccount.php
kesselb: To change the license of a file we need the permissions from all contributors, hence it might be easier to just stay with AGPL-3.0-only.

That might be Christoph, Joas, Richard, Anna, me, Andy, Hamza according to the commit history. I cannot guarantee the corrects of that information.

- https://github.com/nextcloud/mail/commits/main/lib/Command/CreateAccount.php
- https://github.com/nextcloud/mail/commits/4531545c6102a06fee7ea9a31f65375a866c257c/lib/command/createaccount.php?browsing_rename_history=true&new_path=lib/Command/CreateAccount.php&original_branch=main
- https://github.com/owncloud-archive/mail/commits/master/lib/command/createaccount.php

I'm okay with changing the license ;)

lib/Db/MailAccount.php
```diff
@@ -337,6 +354,8 @@ public function toJson() {
 			'debug' => $this->getDebug(),
 			'classificationEnabled' => $this->getClassificationEnabled(),
 			'imipCreate' => $this->getImipCreate(),
+			'protocol' => $this->getProtocol() ?? self::PROTOCOL_IMAP,
```
kesselb:
```suggestion
			'protocol' => $this->getProtocol(),
```

(default already in declaration, proctol cannot be null)

lib/Command/CreateImapAccount.php
SebastianKrupinski: Ah! I see what you mean. Okay, I can change is back if its a big deal but it seems that all the contributors are NC employees, so idk

lib/Migration/Version5800Date20260401000002.php
SebastianKrupinski: Okay, I can change the field names.

As for the state, yes jmap has a unified state for all changes. You can sync all messages from all mailboxes with a single delta request or per mailbox BUT we handle sync on a per mailbox basis that means I need to know the last sync state for that mailbox, hence the per mailbox state.

lib/Command/CreateImapAccount.php
SebastianKrupinski: I changed it back to avoid issues

lib/Migration/Version5011Date20260401000001.php
SebastianKrupinski: Done

lib/Migration/Version5800Date20260401000002.php
SebastianKrupinski: Changed the field names

lib/Db/MailAccount.php
SebastianKrupinski: Done

SebastianKrupinski:
> jmap-client currently requires guzzlehttp as dependency.
> 
> Issue 1: While guzzle is already a dependency for server via 3rdparty, guzzle will be installed another time here leadiing to another copy of guzzle.
> 
> Issue 2: Not using the nextcloud http client means our protections like ssrf, dns pinning, configuration of proxies through config.php are not applied.
> 
> Solution:
> 
>     * jmap-client: Add https://packagist.org/packages/psr/http-client, remove guzzle, change change to code to use the interface.
> 
>     * mail: Currently only the http client in Nextcloud 34 also implements the client interface, hence we need a wrapper in mail to exposes a http client implementing ClientInterface. For example like https://github.com/nextcloud/bookmarks/blob/master/lib/Http/Client.php.

I agree this is a good upgrade for the client, but this will have to wait, due to the time available for the implementation of this I cannot redesign the jmap-client at the moment.

ChristophWurst (approved):

ChristophWurst (changes requested):
On hold for https://github.com/nextcloud/mail/pull/12711#issuecomment-4314068117

SebastianKrupinski:
> On hold for [#12711 (comment)](https://github.com/nextcloud/mail/pull/12711#issuecomment-4314068117)

@ChristophWurst @kesselb 

Done. Implemented

ChristophWurst (changes requested):
⚠️ Some of the new files are `AGPL-3.0-only` but have to be `AGPL-3.0-or-later`
⚠️ New column remote_id is not indexed but used in WHERE -> full table scan on our biggest table
⚠️ Version is migration class names is outdated
:warning: `php-http/discovery` plugin has to be allowed at installation. Change should be committed

Lots of dead code, but it will be used in part 2, ignoring.

lib/Command/CreateJmapAccount.php
```diff
@@ -0,0 +1,93 @@
...
+
+	protected function configure(): void {
+		$this->setName('mail:account:create-jmap');
+		$this->setDescription('creates a JMAP mail account');
+		$this->addArgument(self::ARGUMENT_USER_ID, InputArgument::REQUIRED, 'user to add the account for');
+		$this->addArgument(self::ARGUMENT_NAME, InputArgument::REQUIRED, 'display name of the account');
+		$this->addArgument(self::ARGUMENT_EMAIL, InputArgument::REQUIRED, 'email address');
+		$this->addArgument(self::ARGUMENT_HOST, InputArgument::REQUIRED, 'JMAP server hostname (e.g. mail.example.com)');
+		$this->addArgument(self::ARGUMENT_PORT, InputArgument::REQUIRED, 'JMAP server port (e.g. 443)');
+		$this->addArgument(self::ARGUMENT_SSL_MODE, InputArgument::REQUIRED, 'SSL mode (ssl or none)');
```
odzhychko: This CLI help might be misleading:

> SSL mode (ssl or none)

To enable SSL the expected argument is "yes" and not "ssl".

SebastianKrupinski: Ooops, looks like somewhere along the line I changed my mind on the name.

SebastianKrupinski: Fixed

SebastianKrupinski:
> ⚠️ Some of the new files are `AGPL-3.0-only` but have to be `AGPL-3.0-or-later`
* Done, the remaining two files  (CreateImapAccount and UpdateImapAccount)  where renamed... So the "AGPL-3.0-only" should probably stay 

> ⚠️ New column remote_id is not indexed but used in WHERE -> full table scan on our biggest table
* Done

> ⚠️ Version is migration class names is outdated
* Done

> ⚠️ `php-http/discovery` plugin has to be allowed at installation. Change should be committed
* Done

ChristophWurst (changes requested):
> Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.

SebastianKrupinski:
> > Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.

Done

ChristophWurst (approved):

lib/Contracts/IMessageConnector.php
```diff
@@ -0,0 +1,115 @@
...
+use OCA\Mail\Service\Quota;
+use OCA\Mail\Service\Search\SearchQuery;
+use Psr\Log\LoggerInterface;
+
+interface IMessageConnector {
+
+	/**
+	 * Synchronize all relevant mailboxes for the account.
+	 */
+	public function syncAll(Account $account, bool $force = false): void;
```
kesselb, line 28: Does that belong here?

lib/Command/TestAccount.php
```diff
@@ -0,0 +1,153 @@
...
+
+		try {
+			$imapClient->login();
+			$output->writeln('<info>Login successful</info>');
+
+			$capabilities = array_keys(
+				json_decode($imapClient->capability->serialize(), true)
+			);
+			sort($capabilities);
+			$output->writeln('Capabilities: <info>' . implode(', ', $capabilities) . '</info>');
```
kesselb, line 91: The old formatting, using newlines, was much better to read.

lib/Command/TestAccount.php
```diff
@@ -0,0 +1,153 @@
...
+
+			$output->writeln('<info>IMAP connection test passed</info>');
```
kesselb, line 93: Can we add "Account has $messages messages in $nMailboxes mailboxes" back, why the removal?

lib/Contracts/IMessageConnector.php
SebastianKrupinski: Yes, it sync messages for an entire account

lib/Command/TestAccount.php
SebastianKrupinski: You need to run this command on feat/jmap-2, the command is truncated in this PR, because its missing code to make the rest work

lib/Command/TestAccount.php
SebastianKrupinski: You need to run this command on feat/jmap-2, the command is truncated in this PR, because its missing code to make the rest work

ChristophWurst:
@kesselb ready to merge?

lib/Command/TestAccount.php
SebastianKrupinski: image

kesselb (approved):

lib/Contracts/IMailboxConnector.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace OCA\Mail\Contracts;
+
+use OCA\Mail\Account;
+use OCA\Mail\Db\Mailbox;
+
+interface IMailboxConnector {
+
+	/**
+	 * Synchronise the list of mailboxes for the given account.
```
kesselb:
```suggestion
	 * Synchronize the list of mailboxes for the given account.
```

SebastianKrupinski: Done

SebastianKrupinski: Done

SebastianKrupinski:
Thank you @kesselb

PR #12715, opened 2026-04-13
fix(autocomplete): respect privacy settings for sharing

hamza221:
How to test: 
1. Admin settings -> sharing -> Privacy settings for sharing
2. edit the settings and make sure that both : 
- Adress autocompletion in composer 
- `@` the mention shortcut in composer

respect them

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -14,6 +14,7 @@ use OCP\Contacts\IManager;
 use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IUserManager;
+use function is_array;
 
 class ContactsIntegration {
 	/** @var IManager */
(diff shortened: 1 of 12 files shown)


lib/Service/ContactsIntegration.php
```diff
@@ -262,14 +281,14 @@ private function doSearch(string $term, array $fields, bool $strictSearch) : arr
 	 * @param string $mailAddr
```
ChristophWurst: now phpdoc is out of sync. either have both params or preferrably drop both if there is no additional info

tests/Unit/Service/Phishing/ContactCheckTest.php
```diff
@@ -56,9 +56,9 @@ public function testCaseInsensitiveEmail(): void {
 		$this->contactsIntegration
 			->expects(self::once())
 			->method('getContactsWithName')
-			->with($fn)
+			->with('', $fn)
 			->willReturn($contacts);
-		$result = $this->service->run($fn, $email);
+		$result = $this->service->run('', $fn, $email);
```
ChristophWurst: can you make this meaningful?

tests/Unit/Service/PhishingDetection/ContactCheckTest.php
```diff
@@ -62,7 +62,7 @@ public function testDifferentEmailMultipleContactsReturnsPhishing(): void {
 			['uid' => 'contact1', 'email' => ['email1@example.com', 'email2@example.com']],
 		]);
 
-		$result = $this->check->run('Sender Name', 'sender@example.com');
+		$result = $this->check->run('', 'Sender Name', 'sender@example.com');
```
ChristophWurst: same for these

tests/Unit/Service/ContactsIntegrationTest.php
```diff
@@ -540,7 +574,7 @@ public function testGetContactsWithNameNoEmail(): void {
 			],
 		];
 
-		$actual = $this->contactsIntegration->getContactsWithName($name);
+		$actual = $this->contactsIntegration->getContactsWithName('', $name);
```
ChristophWurst: same for these

ChristophWurst (changes requested):

hamza221:
@ChristophWurst I've been actually wondering if we want to always allow full sab access to phishing detection check? maybe without exposing the actual contact ?

ChristophWurst (approved):

PR #12442, opened 2026-02-17
feat(provisioning): Add support for Dovecot master user authentication

kesselb:
Local copy of https://github.com/nextcloud/mail/pull/12306 with conflicts resolved

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a Dovecot master user alongside the master password.
  * IMAP, SMTP, and Sieve login previews now show master-user-aware usernames and configuration status.
  * Master-user settings can be enabled or disabled with the master password.

* **Bug Fixes**
  * Improved validation for IMAP port and master-password settings.
  * Prevented empty master passwords from overriding user passwords.

* **Migration**
  * Existing installations are upgraded to store the master user setting.

PR #12727, opened 2026-04-14
feat: Account delegation

hamza221:
Mvp for https://github.com/nextcloud/mail/issues/12403
| 1 | 2 | 3 | 4 | 
|--------|--------|--------|--------|
| image | image | image | image |

## Summary by CodeRabbit

* **New Features**
  - Add mail delegation: modal UI to add/remove delegates and API endpoints to list/create/remove delegations
  - Backend delegation service with DB migration to store delegations

* **Improvements**
  - Delegated accounts shown alongside your own and labeled "(delegated)"
  - Server settings and account removal hidden for delegated accounts; "Delegate account" added to account menu
  - App version bumped to 5.9.0-dev.3

* **Tests**
  - New and updated unit/integration tests covering delegation flows


lib/Controller/DelegationController.php
```diff
@@ -0,0 +1,115 @@
...
+		private AccountService $accountService,
+		private IUserManager $userManager,
+		?string $UserId,
+	) {
+		parent::__construct($appName, $request);
+		$this->currentUserId = $UserId;
+	}
+
+	/**
+	 * @NoAdminRequired
```
kesselb: Please use `#[NoAdminRequired]` attribute

lib/Controller/DelegationController.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param int $accountId
+	 * @return JSONResponse
```
kesselb:
```suggestion
```

Please remove redundant (=> dont provide any additional knoweldge to the type hints) phpdoc

lib/Controller/DelegationController.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param int $accountId
+	 * @param string $userId
+	 * @return JSONResponse
+	 */
+	#[TrapError]
+	public function delegate(int $accountId, string $userId): JSONResponse {
+
+		$account = $this->accountService->findById($accountId);
+		if ($this->currentUserId === null) {
```
kesselb, line 62: Flip with the line above

lib/Controller/DelegationController.php
```diff
@@ -0,0 +1,115 @@
...
+	 * @NoAdminRequired
+	 *
+	 * @param int $accountId
+	 * @param string $userId
+	 * @return JSONResponse
+	 */
+	#[TrapError]
+	public function delegate(int $accountId, string $userId): JSONResponse {
+
+		$account = $this->accountService->findById($accountId);
```
kesselb: Would recommend find with try+catch

hamza221:
Copied over design review from the other pr 

Nimisha 
---- 
Awesome! Looks really nice! Only some small changes. Overall:
- [x] The icon for delegation (across the board) can be [supervisor account](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:supervisor_account:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=super). The one you've chosen is also very similar but we use that for indicating users and groups
- [x] Instead of "Delegation", I'm wondering if we can use the wording "Delegate account" in the 3 dot menu and the modal title
- [x] Use tertiary buttons for all the "Cancel" buttons and don't use the icon.

In the "Delegation" modal:
- [x] There is some whitespace on top which shouldn't be there (somewhow not there in the "Add delegates" modal)
- [x] We can also center align the "Delegates" heading. Ik in kra-mo's mockups it was left aligned, but because in the component we always have the heading in the center it looks inconsistent now. 
- [x] I see that there's a `h3 { font-weight: bold; }` which makes the "Delegates" heading stronger than its parent heading. We could remove that

In the "Add delegates" modal:
- [x] the select component should be full width

Laura
----

> We can also center align the "Delegates" heading.

I was going off of form group components, not a dialog, in this context, I think the two different headings don't make sense either, we could just have one "Delegates" title :)

Laura 
---- 

Adding to what nimishavijay said, the "Revoke" button should also probably not have an icon, a checkmark with danger styling is always confusing :)

Additionally, the user selection dropdown should be full-width and there should be a bit more padding between the explanation text beneath and the buttons, also in the revoke access dialog.

kra-mo (approved):

kesselb:
For provisioned accounts I would suggest to hide the ui option to create a delegation (and maybe even api wise) for now. 

--- 

We have currently two (soon 3) way how authentication works: 1) by using the password from the user session, 2) static password, 3) master password (which is similar to static password).

Case 1: This needs the password of the owner and hence, delegating would only work if the user has been logged in once. That should be the case anyway, however this still seems a bit fragile (think of password change cases, etc).

Case 2: Static password should work, but in the current version wa also use the middleware to update the password which probably wont work currently.

Case 3: Master password should work but is currently not merged. Community PR that I took over https://github.com/nextcloud/mail/pull/12442. It needs a bit of polishing and we wanted to move the password updating from the middleware to the factory.

hamza221:
> For provisioned accounts I would suggest to hide the ui option to create a delegation (and maybe even api wise) for now.
> 
> We have currently two (soon 3) way how authentication works: 1) by using the password from the user session, 2) static password, 3) master password (which is similar to static password).
> 
> Case 1: This needs the password of the owner and hence, delegating would only work if the user has been logged in once. That should be the case anyway, however this still seems a bit fragile (think of password change cases, etc).
> 
> Case 2: Static password should work, but in the current version wa also use the middleware to update the password which probably wont work currently.
> 
> Case 3: Master password should work but is currently not merged. Community PR that I took over #12442. It needs a bit of polishing and we wanted to move the password updating from the middleware to the factory.

Done, It's enough to check if the account has a provisioning id right ?

hamza221:
@copilot+claude-sonnet-4.6 Any undesired behaviors or bugs introduced by this PR to the SMIME encryption feature

lib/Controller/AccountsController.php
```diff
@@ -307,7 +326,8 @@ public function updateSignature(int $id, ?string $signature = null): JSONRespons
 	 */
 	#[TrapError]
 	public function destroy(int $id): JSONResponse {
-		$this->accountService->delete($this->currentUserId, $id);
+		$effectiveUserId = $this->delegationService->resolveAccountUserId($id, $this->currentUserId);
+		$this->accountService->delete($effectiveUserId, $id);
```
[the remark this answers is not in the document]
ChristophWurst: ^ I agree with this. I don't think delegatees should delete accounts.

lib/Controller/DelegationController.php
```diff
@@ -0,0 +1,117 @@
...
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private DelegationService $delegationService,
+		private AccountService $accountService,
+		private IUserManager $userManager,
+		?string $UserId,
+	) {
+		parent::__construct($appName, $request);
+		$this->currentUserId = $UserId;
```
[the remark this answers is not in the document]
ChristophWurst: Capital UserId is deprecated: https://docs.nextcloud.com/server/latest/developer_manual/basics/dependency_injection.html#predefined-core-services

lib/Controller/AccountsController.php
```diff
@@ -164,9 +179,10 @@ public function update(int $id,
 		}
 
 		try {
-			return MailJsonResponse::success(
-				$this->setup->createNewAccount($accountName, $emailAddress, $imapHost, $imapPort, $imapSslMode, $imapUser, $imapPassword, $smtpHost, $smtpPort, $smtpSslMode, $smtpUser, $smtpPassword, $this->currentUserId, $authMethod, $id)
+			$result = MailJsonResponse::success(
+				$this->setup->createNewAccount($accountName, $emailAddress, $imapHost, $imapPort, $imapSslMode, $imapUser, $imapPassword, $smtpHost, $smtpPort, $smtpSslMode, $smtpUser, $smtpPassword, $effectiveUserId, $authMethod, $id)
 			);
+			return $result;
```
ChristophWurst: nit: can be a one-liner again

ChristophWurst (dismissed):

lib/Controller/AccountsController.php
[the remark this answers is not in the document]
hamza221: any other features we should restrict ?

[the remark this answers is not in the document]
ChristophWurst: I think the other destructive ones like email and mailbox deletion are fine. So this should be it.

hamza221:
resolved conflicts, proceeding with the reviews now

lib/Db/DelegationMapper.php
```diff
@@ -0,0 +1,83 @@
...
+
+		$result = $qb->executeQuery();
+		$row = $result->fetch();
+		$result->closeCursor();
+
+		if ($row === false) {
+			throw new DoesNotExistException("No delegation found for account $accountId and user $delegatedUserId");
+		}
+
+		return (string)$row['user_id'];
```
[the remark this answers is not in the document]
hamza221: Stable5.9 will support Nextcloud32 afaik and `fetchAssociative()` / `fetchOne()` are introduced in 33

lib/Migration/Version5009Date20260312152753.php
```diff
@@ -0,0 +1,62 @@
...
+			$table->addColumn('account_id', Types::INTEGER, [
+				'notnull' => true,
+				'length' => 20,
+			]);
+			$table->addColumn('user_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->setPrimaryKey(['id']);
+			$table->addUniqueIndex(['account_id', 'user_id'], 'mail_deleg_acc_user_uniq');
```
ChristophWurst, line 45: perf: also add an index on `user_id` because `findDelegatedAccountsForUser` needs it

hamza221: I removed the method, it now lives in AccountMapper to address https://github.com/nextcloud/mail/pull/12727#discussion_r3188488268

ChristophWurst: The index is equally necessary for the DB to find the oc_mail_delegation rows to join with oc_mail_accounts. Only d.user_id and d.account_id are accessed. If you create the index on `user_id`+`account_id` it will be a *covering index* and very efficient.

hamza221: so switch to 
````
$table->addUniqueIndex(['user_id', 'account_id'], 'mail_deleg_acc_user_uniq');

Comment (ChristophWurst) [reply lib/Migration/Version5009Date20260312152753.php:45]: No, you need both. There are also a couple of lookup by only account_id. An index with two columns can be used to find just the first column or both, but not the second one.

**Comment by kesselb** [comment]
For the final commit, I would prefer to have one "feat: Account delegation". 

If you want to keep the current structure, please use the right types.

`feat: add delegation backend` => `feat(delegation): Add delegation backend`
`feat(delegation): add unit tests` => `test(delegation): Add unit tests for delegation backend`
`feat: add delegation frontend` => `feat(delegation): Add delegation frontend`

**Comment by hamza221** [comment]
> For the final commit, I would prefer to have one "feat: Account delegation".
> 
> If you want to keep the current structure, please use the right types.
> 
> `feat: add delegation backend` => `feat(delegation): Add delegation backend` `feat(delegation): add unit tests` => `test(delegation): Add unit tests for delegation backend` `feat: add delegation frontend` => `feat(delegation): Add delegation frontend`

I'll squash into one commit once ready

File: lib/Migration/Version5008Date20260312152753.php [hunk 10]
```diff
@@ -0,0 +1,62 @@
...
+
+		if (!$schema->hasTable('mail_delegations')) {
+			$table = $schema->createTable('mail_delegations');
+			$table->addColumn('id', Types::INTEGER, [
+				'autoincrement' => true,
+				'notnull' => true,
+			]);
+			$table->addColumn('account_id', Types::INTEGER, [
+				'notnull' => true,
+				'length' => 20,
```
kesselb:
```suggestion
```

To match https://github.com/nextcloud/mail/blob/70bb3d351c3dedfeae43f39bf9f1fbc66cfe53d7/lib/Migration/Version0100Date20180825194217.php#L38-L42

The length has no effect for integers.

lib/Controller/AccountsController.php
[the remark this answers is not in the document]
kesselb: I would also restrict the AccountController.update and hide the section in the ui.

lib/Controller/AccountApiController.php
```diff
@@ -54,17 +56,36 @@ public function list(): DataResponse {
...
 					'id' => $alias->getId(),
 					'email' => $alias->getAlias(),
 					'name' => $alias->getName(),
 				], $aliases),
 			];
-		}, $accounts));
+		}, $accounts);
+
+		$delegatedAccounts = $this->accountService->findDelegatedAccounts($userId);
+		foreach ($delegatedAccounts as $account) {
```
kesselb: nit: regular accounts uses array_map for the result and delegated foreach, would be nice to have one pattern ;) Maybe abstract building the result shape with a little helper transformAccountList($account, $isDelegated)

lib/Controller/AccountApiController.php
```diff
@@ -54,17 +56,36 @@ public function list(): DataResponse {
...
+				'id' => $account->getId(),
+				'email' => $account->getEmail(),
+				'isDelegated' => true,
+				'aliases' => array_map(static fn (Alias $alias) => [
+					'id' => $alias->getId(),
+					'email' => $alias->getAlias(),
+					'name' => $alias->getName(),
+				], $aliases),
+			];
+		}
```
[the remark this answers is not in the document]
kesselb: Sounds good

lib/Db/MessageMapper.php
```diff
@@ -131,6 +131,21 @@ public function findByUserId(string $userId, int $id): Message {
...
 
+	/**
+	 * @throws DoesNotExistException
+	 */
+	public function findAccountIdForMessage(int $messageId): int {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('mb.account_id')
+			->from($this->getTableName(), 'm')
+			->join('m', 'mail_mailboxes', 'mb',
+				$qb->expr()->eq('m.mailbox_id', 'mb.id', IQueryBuilder::PARAM_INT))
```
kesselb:
```suggestion
			->join('m', 'mail_mailboxes', 'mb', $qb->expr()->eq('m.mailbox_id', 'mb.id', IQueryBuilder::PARAM_INT))
```

lib/Service/DelegationService.php
```diff
@@ -0,0 +1,106 @@
...
+		$this->delegationMapper->delete($delegation);
+	}
+
+	/**
+	 * @throws DoesNotExistException
+	 */
+	public function resolveAccountUserId(int $accountId, string $currentUserId): string {
+		// Check if the current user owns the account
+		try {
+			$account = $this->mailAccountMapper->find($currentUserId, $accountId);
```
kesselb: It's a bit sad to construct an account object to just throw it away. We cannot use the regular account service here (the one with caching applied)?

Could we introduce a method similar to findAccountOwnerForDelegatedUser to fetch only the user id? select user_id where id = x and user_id = y?

src/components/DelegationModal.vue
```diff
@@ -0,0 +1,322 @@
...
+		:buttons="revokeButtons"
+		@closing="closeDialog">
+		<p class="revoke-dialog__text">
+			{{ revokeText }}
+		</p>
+	</NcDialog>
+</template>
+
+<script>
+import IconCheck from '@mdi/svg/svg/check.svg?raw'
```
kesselb, line 95:
```suggestion
import IconCheck from 'vue-material-design-icons/Check.vue'
```

Like the other's or do we need a different one here?

lib/Service/DelegationService.php
```diff
@@ -0,0 +1,106 @@
...
+
+	public function findDelegatedToUsersForAccount(int $accountId): array {
+		return $this->delegationMapper->findDelegatedToUsers($accountId);
+	}
+
+	/**
+	 * @throws DoesNotExistException
+	 */
+	public function unDelegate(int $accountId, string $userId): void {
+		$delegation = $this->delegationMapper->find($accountId, $userId);
```
kesselb: Catch DoesNotExistException and call it a day ;)

hamza221: Why ?  we want it to fail (caught by middleware)  on not found

kesselb: The method should remove a delegation for a accountId and userId. DoesNotExistException = the delegation does not exist. Why bother the user with an error? Other exceptions should still bubble up, but that one we dont care. In prod that will probably not happen anyway.

kesselb:
@claude could you review that one please, just to make copilot and coderrabbitai didn't miss anything? Please use mythos if possible.

hamza221:
> @claude could you review that one please, just to make copilot and coderrabbitai didn't miss anything? Please use mythos if possible.

Rabbit was autorun 😅

src/components/DelegationModal.vue
```diff
@@ -0,0 +1,322 @@
...
+			return [
+				{
+					label: t('mail', 'Cancel'),
+					type: 'tertiary',
+					callback: () => { this.closeDialog() },
+				},
+				{
+					label: t('mail', 'Revoke'),
+					type: 'error',
+					callback: () => { this.revokeDelegate() },
```
kesselb: Use async await to have a loading state?

src/components/DelegationModal.vue
```diff
@@ -0,0 +1,322 @@
...
+					type: 'tertiary',
+					disabled: this.delegating,
+					callback: () => { this.closeDialog() },
+				},
+				{
+					label: t('mail', 'Delegate access'),
+					type: 'primary',
+					icon: IconCheck,
+					disabled: !this.selectedUser || this.delegating,
+					callback: () => { this.addDelegate() },
```
kesselb: Also async await

src/components/DelegationModal.vue
```diff
@@ -0,0 +1,322 @@
...
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcModal
+		v-if="view === 'main'"
+		size="normal"
+		label-id="delegation-modal-title"
+		@close="$emit('close')">
+		<div class="delegation-modal">
```
kesselb, line 12: super nitpick: Loadign state for the list

Like src/components/mailFilter/MailFilters.vue

(I refreshed my page and opened modal right away and was confused by the empty list of delegations because the request was still qeued)

src/components/DelegationModal.vue
```diff
@@ -0,0 +1,322 @@
...
+
+<script>
+import { getCurrentUser } from '@nextcloud/auth'
+import axios from '@nextcloud/axios'
+import { showError, showSuccess } from '@nextcloud/dialogs'
+import { generateOcsUrl } from '@nextcloud/router'
+import { ShareType } from '@nextcloud/sharing'
+import { NcAvatar, NcButton, NcDialog, NcListItem, NcModal, NcSelectUsers } from '@nextcloud/vue'
+import debounce from 'lodash/fp/debounce.js'
+import IconCheck from 'vue-material-design-icons/Check.vue'
```
kesselb:
```suggestion
import IconCheck from '@mdi/svg/svg/check.svg?raw'
```

😆 

(Your initial version was correct, NcDialogButton expects the svg as string and throws an error with the objects from vue-mterial-design-icons)

lib/Controller/AccountApiController.php
```diff
@@ -57,35 +57,30 @@ public function list(): DataResponse {
...
-		}
+			return $this->transformAccountList($account, true, $ownerUserId);
+		}, $delegatedAccounts);
+		array_push($result, ... $delegatedList);
 
 		return new DataResponse($result);
 	}
+
+	private function transformAccountList(Account $account, bool $isDelegated, string $userId): array {
+		$aliases = $this->aliasesService->findAll($account->getId(), $userId);
```
kesselb: We cannot re-use the userId from the account object?

hamza221: we can, nice catch

lib/Controller/AccountsController.php
```diff
@@ -113,7 +126,8 @@ public function index(): JSONResponse {
 	 */
 	#[TrapError]
 	public function show(int $id): JSONResponse {
-		return new JSONResponse($this->accountService->find($this->currentUserId, $id));
+		$effectiveUserId = $this->delegationService->resolveAccountUserId($id, $this->currentUserId);
```
kesselb, line 129: resolveAccountUserId throws a DoesNotExistException when we didn't find a matching account (no match regular, no match delegation). 

DoesNotExistException = 404
ClientException = 400

That might be okay for some cases. 

Was it intentional to change the api or an oversight?

lib/Controller/DelegationController.php
```diff
@@ -0,0 +1,117 @@
...
+			return new JSONResponse([], Http::STATUS_UNAUTHORIZED);
+		}
+
+		if ($account->getMailAccount()->getProvisioningId() !== null) {
+			return new JSONResponse(['message' => 'Cannot delegate provisioned accounts'], Http::STATUS_FORBIDDEN);
+		}
+
+		if ($userId === $this->currentUserId) {
+			return new JSONResponse(['message' => 'Cannot delegate to yourself'], Http::STATUS_BAD_REQUEST);
+		}
```
kesselb: Make it the second check in the function?

ChristophWurst:
@hamza221 let's get this in ASAP but please have a look at the response codes and how they are handled.

lib/Migration/Version5007Date20260312152753.php
```diff
@@ -0,0 +1,62 @@
...
+namespace OCA\Mail\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+use Override;
+
+class Version5007Date20260312152753 extends SimpleMigrationStep {
```
kesselb: Please rename to Version5009 ;)

kesselb (approved):

lib/Controller/AccountsController.php
hamza221: Oversight, it might be useful but I'll change back to the pre-existing behaviour

kesselb:
I just realized that Copilot may have flagged operations like managing filters, sieve, etc. because the help text in the delegatee modal (as seen in screenshot 2) says "Allow users to send, receive, and delete mail on your behalf." With that in mind, the warnings make more sense and we should either prevents account actions like filters or update the help text.

On a side note, "receive mail on your behalf" sounds a bit odd, since you are not really receiving emails on someone's behalf but rather reading emails that the owner has already received - but that is just semantics I guess. :)

PR #12764, opened 2026-04-16
feat(aliases): add admin setting to disable alias creation

bromiesTM:
Introduces an `allow_new_mail_aliases` app config flag (default: yes) that lets administrators prevent users from creating new mail aliases. This is useful for providers who want to prevent alias creation or manage it elsewhere.

- Backend: guard in AliasesService::create() throws ClientException when disabled
- Admin UI: toggle switch in AdminSettings, mirroring the existing "allow new mail accounts" setting
- Frontend: hides the "Add alias" button when disabled + (warning) description for setting
- Exposed via PageController initial state and storable via occ config:app:set

Extends existing unit tests for AliasesService, AdminSettings, and PageController to cover the new setting.

### DIFF
--- a/src/components/AccountSettings.vue
+++ b/src/components/AccountSettings.vue
@@ -12,6 +12,7 @@
 		:name="t('mail', 'Account settings')"
 		@update:open="updateOpen">
 		<AppSettingsSection
+			v-if="allowNewAliases"
 			id="alias-settings"
 			:name="t('mail', 'Aliases')">
 			<AliasSettings :account="account" @rename-primary-alias="scrollToAccountSettings" />
@@ -208,6 +209,10 @@ export default {
 
 		email() {
 			return this.account.emailAddress
+		},
+
+		allowNewAliases() {
+			return this.mainStore.getPreference('allow-new-aliases', true)
 		},
 	},
 
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -346,6 +346,11 @@ return [
 			'verb' => 'POST'
 		],
 		[
+			'name' => 'settings#setAllowNewMailAliases',
+			'url' => '/api/settings/allownewaliases',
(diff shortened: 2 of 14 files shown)


ChristophWurst:
Thank you for sending a pull request

ChristophWurst:
Please point your agent to https://github.com/nextcloud/mail/blob/main/AGENTS.md#git-workflow in case you didn't open it from this repo's root. Agentic AI has to be transparent in commit messages.

src/components/AccountSettings.vue
```diff
@@ -209,6 +214,16 @@ export default {
...
+		allowNewAliases() {
+			return this.mainStore.getPreference('allow-new-aliases', true)
+		},
+		showProviderAppPassword() {
+			// Show the password reset section if:
+			// 1. Account is managed by a provider (managedByProvider is set)
+			// 2. Provider supports app passwords
+			return this.account.managedByProvider
+				&& this.account.providerCapabilities?.appPasswords === true
+		},
```
ChristophWurst, line 226: I believe this is unrelated to your feature. The property does not exist

jancborchardt (commented):
Just giving the design-ok for the record. This is useful for providers who want to prevent alias creation or manage it elsewhere.

@bromiesTM for future pull requests please also always include the use-case or the issue the pull request tackles. That helps with deciding how to go ahead with it. Thank you! :)

src/components/AccountSettings.vue
bromiesTM: Yes sorry, slipped through during my rebase

ChristophWurst:
What about the *Alias to S/MIME certificate mapping* section? It will still be shown when aliases are disabled by the admin. Should it be hidden?

ChristophWurst:
Tested and works otherwise. Thank you!

PR #12775, opened 2026-04-17
chore: Move TestLogger to phpunit vendor-bin

kesselb:
The motivation for this change is an obvious "me problem" :)

I run a script to update my dev instance that also executes git pull and composer i --no-dev (to avoid pollution from dev dependencies, probably less likely due to vendor-bin nowadays, but still a good default I think).

Every time I run the unit tests for the mail app, they fail because TestLogger is missing and I forgot to run composer i for the mail app. This can be resolved by moving the logger to phpunit vendor-bin, or we ditch it completely and stub it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,6 @@
 		"psr/log": "^1.0.4|^2|^3"
 	},
 	"require-dev": {
-		"fig/log-test": "^1.2.1",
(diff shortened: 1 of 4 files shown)


ChristophWurst (approved):
Makes sense to you

PR #12777, opened 2026-04-20
feat(UserMigration): Overwork migration to include all parts

DerDreschner:
This PR is the first of a series of PRs that will enhance the user migration to include all settings.

Included changes:
- Smoke test via GitHub Action
- Removes code/responsibilities that will be moved to separate files
- Move background jobs to end of user migration

Ref #12001

## New Features
* Added user migration support for the Mail app, enabling users to export and import email accounts, app settings, S/MIME certificates, trusted senders, text blocks, tags, and quick actions when migrating to a new Nextcloud instance.

## Chores
* Enhanced workflow testing for user migration scenarios.


.github/workflows/test.yml
```diff
@@ -304,13 +304,65 @@ jobs:
     env:
       CI: true
 
+  user-migration-smoke-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [ '8.4' ]
+        nextcloud-versions: [ 'stable33' ]
```
DerDreschner, line 312: This is necessary as the `user_migration` app doesn't support the latest `master` branch without additional code changes as it calls deprecated code that was removed in Nextcloud 34:

```
| Importing from /tmp/user1_2026-04-20_12-16-51.zip…
| Importing user information from user.json…
| An unhandled exception has been thrown:
| Error: Call to undefined method OC\Server::getSecureRandom() in /home/ddreschner/git/mail/nextcloud/apps/user_migration/lib/Service/UserMigrationService.php:261
```

Was looking for an error in my changes before suspecting the `user_migration` app, so... 😅

lib/UserMigration/MailAccountMigrator.php
```diff
@@ -42,140 +46,31 @@ public function export(IUser $user,
...
-					$alias['name'],
-				);
-			}
-		}
+		$output->writeln("Importing mail accounts for user {$user->getUID()}", OutputInterface::VERBOSITY_VERBOSE);
+
+		$this->deleteExistingData($user, $output);
+
+		$this->accountMigrationService->scheduleBackgroundJobs($user, $output);
+	}
```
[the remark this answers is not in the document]
DerDreschner: Irrelevant, as this is only the first of a series of PRs

lib/UserMigration/MailAccountMigrator.php
```diff
@@ -42,140 +46,31 @@ public function export(IUser $user,
...
-				$data = $alias->jsonSerialize();
-				return $data;
-			}, $aliases);
-
-			$exportDestination->addFileContents($accountFilePath, json_encode($accountData));
-			$index[$account->getId()] = $accountFilePath;
-		}
-
-		$exportDestination->addFileContents('mail/accounts/index.json', json_encode($index));
+		$output->writeln("Exporting mail accounts for user {$user->getUID()}", OutputInterface::VERBOSITY_VERBOSE);
```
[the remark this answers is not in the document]
DerDreschner: Irrelevant, as this is only the first of a series of PRs

tests/Unit/UserMigration/MailAccountMigratorTest.php
```diff
@@ -113,127 +113,4 @@ public function testCanImportOlder(): void {
...
-			->willReturnArgument(0);
-
-		$this->migrator->import(
-			$user,
-			$importSource,
-			$this->output,
-		);
-	}
-
 }
```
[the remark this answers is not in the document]
DerDreschner: Irrelevant, as this is only the first of a series of PRs

tests/Unit/UserMigration/Service/AccountMigrationServiceTest.php
```diff
@@ -0,0 +1,93 @@
...
+			});
+
+		$this->serviceMock->getParameter('crypto')
+			->method('decrypt')
+			->willReturnCallback(function (string $encryptedValue) {
+				if (!str_ends_with($encryptedValue, '_encrypted')) {
+					throw new Exception('Invalid encrypted value');
+				}
+				return substr($encryptedValue, 0, strlen($encryptedValue) - strlen('_encrypted'));
+			});
```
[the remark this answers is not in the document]
DerDreschner: Irrelevant, as this is only the first of a series of PRs

.github/workflows/test.yml
```diff
@@ -304,13 +304,65 @@ jobs:
...
+          nextcloud-version: ${{ matrix.nextcloud-versions }}
+          php-version: ${{ matrix.php-versions }}
+          patch-php-version-check: ${{ matrix.php-versions == '8.6' }}
+          node-version: 'false'
+          install: true
+      - name: Checkout mail
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          path: nextcloud/apps/mail
+          fetch-depth: 2
```
DerDreschner, line 331: Used the same `fetch-depth` as in other tests before. But I wonder why that's necessary in those cases? 🤔

lib/UserMigration/Service/AccountMigrationService.php
```diff
@@ -0,0 +1,68 @@
...
+		$accountCount = count($allAccounts);
+		$uid = $user->getUID();
+
+		$output->writeln("Deleting {$accountCount} mail account(s) for user {$uid}", OutputInterface::VERBOSITY_VERBOSE);
+
+		foreach ($allAccounts as $account) {
+			$accountId = $account->getId();
+			$this->accountService->deleteByAccountId($accountId);
+			$output->writeln("Deleted mail account {$accountId}", OutputInterface::VERBOSITY_VERBOSE);
+		}
```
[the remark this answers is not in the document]
DerDreschner: Good point. But I'll move it to the same `foreach` to prevent duplicate loops

.github/workflows/test.yml
ChristophWurst: Please report at https://github.com/nextcloud/user_migration/issues.

ChristophWurst (approved):

lib/UserMigration/Service/AccountMigrationService.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @param IUser $user
+	 * @param OutputInterface $output
+	 * @return void
+	 */
+	public function scheduleBackgroundJobs(IUser $user, OutputInterface $output): void {
+		$accounts = $this->accountService->findByUserId($user->getUID());
+		$accountCount = count($accounts);
+
+		$output->writeln("Scheduling background jobs for {$accountCount} mail account(s)", OutputInterface::VERBOSITY_VERBOSE);
+
```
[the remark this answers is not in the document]
DerDreschner: Mhm, valid point... Although the impact may be limited to use-cases outside of the user migration.

@ChristophWurst: I'll call `unset()` in the `delete()`, `deleteByAccountId()`, `save()` and `updateSignature()` methods as the cache content has to be renewed then. What do you think about it? :thinking: The caching isn't persisted across several requests anyway right now as far as I understand it? :thinking: Sure, removing/adding/updating only the affected entries is more correct, but the change is kinda unrelated to this PR and I would move it to memcache anyway then once the persistence under longer lifetimes is given.

.github/workflows/test.yml
DerDreschner: Done! https://github.com/nextcloud/user_migration/issues/1088

lib/UserMigration/MailAccountMigrator.php
```diff
@@ -42,140 +39,31 @@ public function export(IUser $user,
...
-					$user->getUID(),
-					$mailAccount->getId(),
-					$alias['alias'],
-					$alias['name'],
-				);
-			}
-		}
+		$output->writeln($this->l10n->t("Importing mail accounts for user {$user->getUID()}"), OutputInterface::VERBOSITY_VERBOSE);
+
+		$this->deleteExistingData($user, $output);
```
kesselb: Why is it necessary? 

The user_migration app says "Please note that existing data may be overwritten", so it might be a bit unexpected that existing data is removed. 

At worst, an account is added a second time and the user needs to remove it manually?

lib/Service/AccountService.php
```diff
@@ -149,6 +149,10 @@ public function delete(string $currentUserId, int $accountId): void {
 		}
 		$this->aliasesService->deleteAll($accountId);
 		$this->mapper->delete($mailAccount);
+
+		// Invalidate cache to ensure deleted account is not included
+		// in subsequent `findByUserId` and `findByUserIdAndAddress` calls
+		unset($this->accounts[$currentUserId]);
```
kesselb, line 153: That looks like a bugfix we should backport (even it's an edge case)

DerDreschner: Alright, then I'll remove it here and put it into a separate PR.

.github/workflows/test.yml
ChristophWurst: Fixed, so we can also include master :)

lib/UserMigration/MailAccountMigrator.php
DerDreschner: Well, I initially wrote the code as my development instance crashed due to an enormous number of duplicate mail accounts being added during development. But I left it in there for the following reasons:

1. Some of the data must be unique, which means I must check for collisions and handle them properly (e.g., check if it's really the same data or not - what happens if it's different - get the correct ID of the existing entry and use that to link it with the mail account, and so on...). That's something I tried to prevent here for the second reason.
2. In my opinion, the main goal of the user_migration app is to backup and import a specific **state**, not only data. At least that's what I would expect from a sysadmin point of view and how I understand the "Do you really want to overwrite this user with the imported data?" question that [must be confirmed](https://github.com/nextcloud/user_migration/blob/main/lib/Command/Import.php#L68C25-L68C90) during import of an existing user account.
3. After all PRs are merged, there would be duplicate everything (if it not must be unique in the database) which would be annoying (and crashed my instance easily after a few imports).

The documentation of the user_migration app doesn't mention how that case should be handled exactly. As I had the code written anyway, I was hoping to discuss that point here. In an ideal world, we could get additional option flags from the CLI and decide if that should be done or not... But yeah :smile:

.github/workflows/test.yml
DerDreschner: Additional or instead of? 😄

lib/Service/AccountService.php
DerDreschner: Please also see my question here regarding that point: https://github.com/nextcloud/mail/pull/12777/changes#r3112940249

lib/UserMigration/MailAccountMigrator.php
ChristophWurst: That is indeed something that should be clarified at https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/user_migration.html#register-a-migrator. It's probably relevant for many migrators.

.github/workflows/test.yml
ChristophWurst: Instead should be sufficient for this smoke test

.github/workflows/test.yml
ChristophWurst: Don't remember. Leave it like this for now

lib/UserMigration/Service/AccountMigrationService.php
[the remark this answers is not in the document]
ChristophWurst: Obsolete with the changes from https://github.com/nextcloud/mail/pull/12777/changes#r3119948729, right?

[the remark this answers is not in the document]
DerDreschner: Yep, correct :smile:

lib/Service/AccountService.php
DerDreschner: Bugfix PR: https://github.com/nextcloud/mail/pull/12803

Will remove the change here later.

lib/UserMigration/MailAccountMigrator.php
kesselb: > Do you really want to overwrite this user with the imported data?

This is not shown when the import is started as a regular user via web (settings -> data migration). 

The **migrator** for the files app may overwrite existing data, but data not in the archive is not removed.

DerDreschner: I've created the following discussion in the `user_migration` repository to clarify how it should be handled: https://github.com/nextcloud/user_migration/discussions/1090

Afterwards, the documentation can be updated. Don't want to just edit it without further discussion.

DerDreschner: > This is not shown when the import is started as a regular user via web (settings -> data migration).

Yeah, I was talking about the CLI. Sorry that I didn't clarified that.

We should continue that in the discussion I've created at the same time you typed your comment. :smile:

kesselb: > Yeah, I was talking about the CLI. Sorry that I didn't clarified that.

You did include a reference to a CLI command in your post; hence, I was aware of that.

.github/workflows/test.yml
DerDreschner: I have to patch the nextcloud `max-version` in the info.xml as user_migration still uses `max-version="33"` in the default branch.

lib/UserMigration/MailAccountMigrator.php
DerDreschner: Alright, I'll patch out the removal of existing data. Côme answered in the discussion that existing data should not be removed. I want to follow that path and will add a note in the documentation. 

Additionally, I realized that - for using it in real-world scenarios - I would have to add a check to only remove stuff not used in provisioned accounts. That's being removed ruthlessly right now. So, that's another point for dropping it until there is a "remove existing data" switch on import.

DerDreschner: Issue for such a "remove existing data" switch (use-case backup and restore): https://github.com/nextcloud/user_migration/issues/1096

PR for clarification the `user_migration` documentation: https://github.com/nextcloud/documentation/pull/14680

ChristophWurst (approved):

DerDreschner:
All parts should now have the same structure.

TODO: Test if all duplicate entry-cases are being handled correctly.
TODO: Unit tests for accounts.

kesselb:
Claude

- [ ] High AccountMigrationService.php:183 — Mailbox mapping overwritten in loop — multi-account users lose mailbox→quick-action links
- [ ] High AccountMigrationService.php:146 — Wrong JSON key ooFollowsSystem (should be outOfOfficeFollowsSystem) — OOO setting silently lost
- [ ] Medium AccountMigrationService.php:319 — Missing pathExists() guard before getFolderListing() — crashes on old/partial archives
- [ ] Medium AccountMigrationService.php:173 — No transaction: failed alias/mailbox import leaves orphaned account rows
- [ ] Medium AppConfigMigrationService.php:949 — All config keys imported blindly — stale account IDs from source instance written to dest
- [ ] Low AccountMigrationService.php:413 — (int)null = 0 without ?? null on alias SMIME lookup — inconsistent with account-level pattern
- [ ] Cleanup AccountMigrationService.php:199 — Redundant findByUserId() query — accounts already available in the mapping
- [ ] Cleanup AccountMigrationService.php:400 — Double UPDATE per alias (4 DB ops instead of 2)
- [ ] Cleanup test.yml:322 — user_migration checkout not pinned to a SHA

lib/Service/AliasesService.php
```diff
@@ -133,4 +134,16 @@ public function updateSignature(string $userId, int $aliasId, ?string $signature
...
 		return $this->aliasMapper->update($entity);
 	}
+
+	/**
+	 * Update the S/MIME certificate for an alias.
+	 *
+	 * @throws DoesNotExistException
+	 * @throws Exception
+	 */
+	public function updateSMIMECertificateId(string $userId, int $aliasId, ?int $sMimeCertificateId = null): Alias {
```
kesselb:
```suggestion
	public function updateSmimeCertificateId(string $userId, int $aliasId, ?int $smimeCertificateId = null): Alias {
```

kesselb:
@coderabbitai review

lib/UserMigration/Service/AppConfigMigrationService.php
```diff
@@ -0,0 +1,134 @@
...
+
+		foreach ($appConfig as $appSetting) {
+			$output->writeln(
+				$this->l10n->t('Importing mail app configuration key %s for user %s', [ $appSetting['key'], $user->getUID() ]),
+				OutputInterface::VERBOSITY_VERBOSE
+			);
+
+			/** @noinspection PhpUnhandledExceptionInspection */
+			$this->config->setUserValue($user->getUID(), Application::APP_ID, $appSetting['key'], $appSetting['value']);
+		}
```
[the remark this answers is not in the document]
DerDreschner: True, very good point!

tests/Unit/UserMigration/Service/AccountMigrationServiceTest.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Tests\Unit;
```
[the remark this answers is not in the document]
DerDreschner: Good point, those files are not mapped to the `OCA\Mail` namespace.

tests/Unit/UserMigration/Service/InternalAddressesMigrationServiceTest.php
```diff
@@ -0,0 +1,152 @@
...
+			->method('addFileContents')
+			->with(InternalAddressesMigrationService::INTERNAL_ADDRESSES_FILE, json_encode($trustedSendersList));
+
+		$this->serviceMock->getParameter('internalAddressService')
+			->method('getInternalAddresses')
+			->with(self::USER_ID)
+			->willReturn($trustedSendersList);
+
+		$this->migrationService->exportInternalAddresses($this->user, $this->exportDestination, $this->output);
+	}
```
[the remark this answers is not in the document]
DerDreschner: That's already applied? Ignore.

tests/Unit/UserMigration/Service/TextBlocksMigrationServiceTest.php
```diff
@@ -0,0 +1,171 @@
...
+			->method('getFileContents')
+			->with(TextBlocksMigrationService::TEXT_BLOCKS_FILE)
+			->willReturn($fileContents);
+
+		$this->serviceMock->getParameter('textBlockService')
+			->expects(self::never())
+			->method('create');
+
+		$this->migrationService->importTextBlocks($this->user, $this->importSource, $this->output);
+	}
```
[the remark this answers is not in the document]
DerDreschner: That is already applied? Ignore.

tests/Unit/UserMigration/Service/AppConfigMigrationServiceTest.php
```diff
@@ -0,0 +1,177 @@
...
+				string $key) use (&$calls): string {
+				$calls[] = [$userId, $appId, $key];
+				return $this->getAppValue($key);
+			});
+
+		$this->migrationService->exportAppConfiguration($this->user, $this->exportDestination, $this->output);
+
+		$expected = array_map(fn (string $key) => [self::USER_ID, Application::APP_ID, $key], $this->getAppKeys());
+		self::assertEqualsCanonicalizing($expected, $calls);
+	}
```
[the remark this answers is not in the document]
DerDreschner: That is already being applied. Ignore.

tests/Unit/UserMigration/Service/AppConfigMigrationServiceTest.php
```diff
@@ -0,0 +1,177 @@
...
+				string $key,
+				string $value) use (&$calls): void {
+				$calls[] = [$userId, $appId, $key, $value];
+			});
+
+		$this->migrationService->importAppConfiguration($this->user, $this->importSource, $this->output);
+
+		$expected = array_map(fn (array $c) => [self::USER_ID, Application::APP_ID, $c['key'], $c['value']], $this->getAppConfig());
+		self::assertEqualsCanonicalizing($expected, $calls);
+	}
```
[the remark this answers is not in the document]
DerDreschner: That is already being applied. Ignore.

tests/Unit/UserMigration/Service/AccountMigrationServiceTest.php
```diff
@@ -0,0 +1,93 @@
...
+				$scheduledIds[] = $accountId;
+			});
+
+		$this->migrationService->scheduleBackgroundJobs(
+			$this->user,
+			$this->output,
+		);
+
+		self::assertSame([101, 102], $scheduledIds);
+	}
```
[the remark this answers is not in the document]
DerDreschner: That is already being applied. Ignore.

PR #12778, opened 2026-04-20
fix(composer): Don't show 'No results' dropdown when To/Cc/Bcc focused, only with search term put in

jancborchardt:
When opening the composer, the "To" field is automatically focused. This is nice.

But before, this also automatically opened the search result dropdown, which of course shows "No results" for no search term. This change hides the dropdown until there is an actual search term given:

Before | After
-|-
Screenshot From 2026-04-20 15-00-29|Screenshot From 2026-04-20 14-58-57

AI-assisted: GitHub Copilot (OpenAI GPT-5.4)

> **Prompt**
> When opening the composer via "New message", the "Contact or email address" field is focused by default. This is good. However, the dropdown immediately shows "No results", referring to the search-as-you-type suggestions which would show there from users and contacts. This should only show as soon as there is any content typed in the field.

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -52,11 +52,13 @@
 					id="to"
 					ref="toLabel"
 					:model-value="selectTo"
+					class="select"
 					:options="selectableRecipients.filter(recipient => !selectTo.some(to => to.email === recipient.email))"
 					:get-option-key="(option) => option.email"
 					:taggable="true"
 					:aria-label-combobox="t('mail', 'Select recipient')"
 					:filter-by="(option, label, search) => filterOption(option, label, search, 'to')"
+					:dropdown-should-open="shouldOpenRecipientDropdown"
 					:multiple="true"
 					:clear-search-on-select="true"
 					:loading="loadingIndicatorTo"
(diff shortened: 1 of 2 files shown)


src/components/Composer.vue
```diff
@@ -52,11 +52,13 @@
 					id="to"
 					ref="toLabel"
 					:model-value="selectTo"
+					class="select"
```
ChristophWurst, line 55: nitpick: unrelated, stray edit

ChristophWurst (approved):
Welcome back as committer

src/components/Composer.vue
jancborchardt: Yep sorry for that – I initially wanted to also have the issue fixed of the subject and body jumping when the "To" field is focused vs unfocused. Will continue with that later. :)

jancborchardt (commented):
So is the Copilot review useful and should be applied? If so it needs someone with more technical understanding than me. :) If not, it is good to go, right?

src/components/Composer.vue
```diff
@@ -1177,6 +1181,10 @@ export default {
 				|| (option?.email || '').toLocaleLowerCase().includes(searchInLowerCase)
 		},
 
+		shouldOpenRecipientDropdown({ noDrop, open, search }) {
+			return !noDrop && open && search.trim() !== ''
```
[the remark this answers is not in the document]
kesselb: Good hardening. 

search is initialized as an empty string, so it seems okay: https://github.com/nextcloud-libraries/vue-select/blob/f5ef266619fb80df945840326ea2e9ea69499f34/src/components/Select.vue#L795-L796

Issue #11009, opened 2025-04-16
Mail is not parsing calendar invites


geransmith:
### Steps to reproduce

1. Using a third party system (such as Google Calendar), send an invite to your configured email account in NextCloud
2. After the invitation email has been received, click the email in the Mail app
3. Observe that there are no options to accept or decline the invite
4. Go to the Calendar app
5. Observe that the invite does not appear automatically in the calendar

### Expected behavior

The Mail app renders a section asking me to accept or decline the event.

The Mail app creates an unaccepted calendar invite in the Calendar app that can be accepted via connected CalDAV clients.

### Actual behavior

The Mail app does not display an option to accept or decline an event.

The Calendar app does not show any tentative events.

### Mail app version

5.0.0

### Nextcloud version

31.0.3

### Mailserver or service

Proton Mail (via ProtonMail-Bridge)

### Operating system

Unraid

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

I am not sure if this is similar / a duplicate of #10811 

The portion around the Calendar app getting an invite as tentative does seem to duplicate #7750 

[example_email.txt](https://github.com/user-attachments/files/19785062/example_email.txt)

geransmith:
> Hi [@geransmith](https://github.com/geransmith)
> 
> Can you attach a screen capture of what you are seeing when you click on the message?
> 
> Also, two things to double check,
> 
> A) the invitation needs to be in the future, B) the invitation must contain an attendee with the same email address as one of your configured email accounts

Attached is what I see when I open the message (with blanking out my actual email addresses):

![Image](https://github.com/user-attachments/assets/ec8cda16-e802-4d4b-b846-ce658ae722fa)

For sanity, I included the date/time in the screenshot to confirm the meeting invite is in the future.

The from address is from my work email. The to address matches what I have as the primary email on my user profile in Nextcloud and also matches one of the aliases in my mail account.

SebastianKrupinski:
> The from address is from my work email. The to address matches what I have as the primary email on my user profile in Nextcloud and also matches one of the aliases in my mail account.

Okay, its most likely something to do with non matching email addresses. I recently made some upgrades to the display prompt.

https://github.com/nextcloud/mail/pull/10970

This might not be released yet, but can you test this again after this patch has been installed

geransmith:
> Okay, its most likely something to do with non matching email addresses. I recently made some upgrades to the display prompt.

The to address does match though.

[...]

PR #12790, opened 2026-04-22
fix(imap): detect iMIP when method= is missing or mime is application/ics

lubo92:
## Summary

Two upstream-detection gaps let calendar invitations slip through unprocessed:

**1. Proton Mail Bridge strips the `method=` parameter.** During E2E re-assembly Bridge loses the Content-Type parameter on the text/calendar part:

```
Content-Type: text/calendar; filename="invite.ics"; name="invite.ics"
```

instead of

```
Content-Type: text/calendar; charset=UTF-8; method=REQUEST
```

The ICS body still contains `METHOD:REQUEST`, but \`MessageMapper::getBodyStructureData()\` and \`ImapMessageFetcher::getPart()\` both key off the Content-Type parameter and ignore the body. Result: every single Proton-Bridge user silently loses inbound invitations from Gmail, M365, Fastmail, …

**2. `application/ics` is ignored.** Google's invitations ship two calendar parts — one `text/calendar`, one `application/ics`. If the calling code hands the fetcher the `application/ics` part (some MUAs/filters prefer the alphabetically-second attachment), it is treated as a regular attachment and no scheduling entry is built.

## Changes

- \`MessageMapper::getBodyStructureData()\`: accepts both mime types as iMIP candidates, drops the `method=` precondition (the method resolution happens further down).
- \`ImapMessageFetcher::getPart()\`: accepts both mime types; if the Content-Type has no `method=` parameter, parses \`METHOD:\` from the ICS body instead. Body is only loaded in the fallback path, so the common case keeps its I/O pattern unchanged.

## Tests

Two new fixtures + data-provider entries in \`MessageMapperTest::testGetBodyStructureIsImipMessage()\`:

- \`request_proton_bridge.txt\` — text/calendar without `method=` parameter, METHOD:REQUEST in body
- \`request_application_ics.txt\` — application/ics instead of text/calendar

Both must now evaluate to \`isImipMessage() === true\`.

## Real-world impact

Empirically verified on a production Proton-Business + Nextcloud 32 + Mail-App 5.7.12 setup: before this change, Gmail and M365 invitations delivered via Bridge never reached CalDAV. After applying the same fix in-place, the iMIP pipeline works end-to-end for requests, replies, and cancels. Repro MIME samples available on request.

## Checklist

- [x] Conventional-Commits commit message
- [x] DCO signed-off commit
- [x] Unit tests for new branch and fixture files added
- [x] No behaviour change for messages that *do* carry a valid `method=` parameter

Co-authored-with Claude Opus 4.7.


SebastianKrupinski:
Hi @lubo92 

Thanks for the PR!

I think the method discovery can actually be removed, the back end code no longer has split handling for calendar invitations, like it did before,

https://github.com/nextcloud/server/blob/a3e6ea0e836cbb554c87af18924b50d05f25786a/lib/private/Calendar/Manager.php#L229

There for the external (pre ics parsing) method discovery in the mail app, no longer really has a use. The method discovery now happens after the calendar parses the event.

https://github.com/nextcloud/server/blob/a3e6ea0e836cbb554c87af18924b50d05f25786a/apps/dav/lib/CalDAV/CalendarImpl.php#L232

The only thing I would leave is the ics file discovery based on type text/calendar and application/ics

@kesselb what do you think?

kesselb:
Thanks for your PR!

We've discussed this a while ago: https://github.com/nextcloud/mail/issues/11009

I reached out to Proton back then, and they confirmed the bug and told me they would add it to their roadmap. I would kindly ask you to reach out and flag it again, as that hopefully raises the chances of having this issue resolved properly at some point. My request ticket number was 3648275.

---

Why did you also add application/ics? The RFC is quite strict on this, it should be text/calendar. Is there a provider generating those with application/ics?

---

I think we still need the code path, as it adds the invitation to the scheduling array which is later used to render the iMIP component. I'm not aware that we want to get rid of it.

https://github.com/nextcloud/mail/blob/95442ad75577309e73401959b8a956f89729bb61/lib/IMAP/ImapMessageFetcher.php#L306-L338

SebastianKrupinski:
> Why did you also add application/ics? The RFC is quite strict on this, it should be text/calendar. Is there a provider generating those with application/ics?

A lot of providers add this.

lubo92:
Thanks @SebastianKrupinski and @kesselb for the careful read. Quick update plus answers below.

**Update on the PR itself**

- Rebased on `main`.
- Fixed the psalm `RedundantCondition` failure (`is_string($contents)` was redundant — `loadBodyData()` is `: string`-typed).
- Added the `AI-assisted:` trailer required by [AGENTS.md](https://github.com/nextcloud/mail/blob/main/AGENTS.md#commit-message-format).

**On scope (re @SebastianKrupinski)**

Agreed that the calendar backend now does its own METHOD discovery after ICS parsing in `Calendar/Manager.php:229` / `CalDAV/CalendarImpl.php:232`. But the `$this->scheduling[]` array populated here in `ImapMessageFetcher::getPart()` isn't consumed by the server calendar — it's exposed via `ImapMessageFetcher::getScheduling()` and fed into the mail app's own iMIP UI rendering (the pointer @kesselb gave at [ImapMessageFetcher.php:306-338](https://github.com/nextcloud/mail/blob/95442ad75577309e73401959b8a956f89729bb61/lib/IMAP/ImapMessageFetcher.php#L306-L338)). Removing the path wholesale would regress the in-app iMIP component.

So I kept the codepath and only widened the discovery in two minimally invasive ways:

1. Accept `application/ics` alongside `text/calendar` at both detection sites.
2. When the Content-Type `method=` parameter is missing, parse the `METHOD:` line out of the ICS body once — body data is only loaded in the fallback path, so the common case has identical I/O to before.

Happy to revisit if you'd rather drop the REQUEST/REPLY/CANCEL filter as well and let the backend decide — but it felt safer to keep the filter as-is for this fix.

**On `application/ics` (re @kesselb)**

You're right that RFC 5546 is strict on `text/calendar`. The in-the-wild reality differs: Google's invitations occasionally ship the same event twice — once as `text/calendar`, once as `application/ics`. The existing `tests/data/imip/request_google.txt` fixture shows the dual-part pattern; the new `request_application_ics.txt` exercises the case where the application/ics part is the one the MUA prefers. We've seen the same `application/ics`-only delivery from a few smaller IMAP-fronted setups too.

**On Proton (re @kesselb)**

Will reach out to Proton again referencing ticket 3648275 — I agree the proper fix belongs upstream in Bridge. Treating this PR as the defensive fallback so Bridge users aren't silently losing invitations until that lands.

PTAL when you have a minute.

lib/IMAP/ImapMessageFetcher.php
```diff
@@ -325,18 +327,29 @@ private function getPart(Horde_Mime_Part $p, string $partNo, bool $isFetched): v
...
+			// Try the Content-Type method= parameter first — that's the common
+			// case. If it is missing, fall back to the METHOD: line inside
+			// the ICS body: Proton Mail Bridge strips the parameter during
+			// E2E re-assembly, and requiring it here would mean every
+			// Proton-Bridge user silently loses inbound invitations.
+			$method = $allContentTypeParameters['method'] ?? null;
+			$contents = null;
+			if ($method === null) {
+				$contents = $this->loadBodyData($p, $partNo, $isFetched);
+				if (preg_match('/^METHOD:\s*(\S+)/mi', $contents, $m)) {
```
kesselb:
```suggestion
				if (preg_match('/^METHOD:([A-Z]+)/mi', $contents, $m) === 1) {
```

kesselb:
Thanks for the update.

With the current information (provided by you and the original report), I am okay with looking into the attachment to check if it looks like an event invitation.

For a Google invitation damaged by Proton, the actual fix is obtaining the method from the attachment. Nothing else needed.

Looking at application/ics seems nice, but in this case it is a nice-to-have on the side. I would prefer to not do that. If someone reaches out with a provider that only includes application/ics, we can reevaluate (in case they are not fixing their implementation). That matches the current state in Thunderbird.

> The existing tests/data/imip/request_google.txt fixture shows the dual-part pattern.

It does not. I created that sample based on an original Google invitation, but dropped all unnecessary parts including the base64 encoded application/ics part.

lubo92:
@kesselb thanks. Pushed a fixup commit (e4990e6) that scopes the PR down per your guidance:

- dropped the `application/ics` widening at both detection sites (ImapMessageFetcher + MessageMapper); we can revisit if a real-world report comes in for a provider that only ships `application/ics`
- tightened the METHOD regex to `'/^METHOD:([A-Z]+)/mi'` with the `=== 1` check
- removed the now-obsolete `tests/data/imip/request_application_ics.txt` fixture and its data-provider row; the Proton Bridge fixture stays as the regression case for the actual fix

Per AGENTS.md this went in as `--fixup` rather than amend. Heads-up for the squash: the original commit body and subject still mention `application/ics` — when squashing, please trim that down so the message only covers the Proton Bridge / METHOD-from-body fallback.

Will re-ping Proton with their ticket 3648275 in parallel so the proper Bridge-side fix gets attention again.

lubo92:
FYI: the two failing checks on `e4990e6` (`Block fixup and squash commits` + `Block unconventional commits`) both fire on the `fixup!` commit-subject prefix and are the ["intentional and expected" failures](https://github.com/nextcloud/mail/blob/main/AGENTS.md#pr-review-workflow) per AGENTS.md — they'll clear automatically when the branch is rebased + autosquashed before merge. Substantive checks (DCO, lint, static analysis, tests) are still waiting on first-time-contributor approval.

lubo92:
Pushed a second fixup ([`404bb93`](https://github.com/lubo92/mail/commit/404bb931)) adding an integration test for the METHOD-from-body fallback. Uses the existing `request_proton_bridge.txt` fixture and the `saveMimeMessage` / `fetchMessage` pattern from the surrounding S/MIME tests in `ImapMessageFetcherIntegrationTest`, so it exercises the new branch through the real Horde+IMAP stack — which is what the MessageMapperTest path doesn't reach.

This is what was driving the codecov patch-coverage failure (9.09%): the new `if (preg_match…)` line in `ImapMessageFetcher::getPart()` had no test exercising it before. Should clear codecov once CI runs.

Note: I don't have the integration-test docker stack locally so I couldn't verify the assertions; CI will be the source of truth.

kesselb (approved):
Thanks a lot

kesselb:
Hi @lubo92, could you please rebase and squash?

PR #12791, opened 2026-04-22
feat(preview-enhancement): make background-job interval configurable

lubo92:
## Why

\`PreviewEnhancementProcessingJob\` writes the \`imip_message\` flag. \`IMipMessageJob\` only picks up rows with that flag set, so the enhancement interval is effectively the ceiling on how long an incoming calendar invitation can take to turn into a CalDAV event.

The current hard-coded \`setInterval(3600)\` is reasonable for bulk preview-data enrichment, but it's the wrong number if you want iMIP to feel live. On a small Proton/Fastmail/self-hosted install the box has plenty of headroom to run the job every minute or two; on a hoster with 10k mailboxes the same number would be painful.

That's a per-setup trade-off. So: expose the knob.

## What

New system config \`app.mail.preview-enhancement-interval\` (integer seconds).

\`\`\`
occ config:system:set app.mail.preview-enhancement-interval --value=300 --type=integer
\`\`\`

Unset ⇒ keeps the current 3600s default. A \`MIN_INTERVAL = 60\`s floor prevents a misconfig from hammering the DB. The job is instantiated via DI per tick, so the new value picks up on the next scheduler run — no restart required.

## Companion PR

Related: #12790 (iMIP detection when \`method=\` is missing). Together they make Proton-Bridge-delivered invitations flow end-to-end. This PR is independent and valuable on its own.

## Checklist

- [x] Conventional-Commits commit message
- [x] DCO signed-off commit
- [x] Default unchanged (backwards-compatible)
- [x] Lower-bound clamp to avoid tight loops

Co-authored-with Claude Opus 4.7.

### DIFF
--- a/lib/BackgroundJob/PreviewEnhancementProcessingJob.php
+++ b/lib/BackgroundJob/PreviewEnhancementProcessingJob.php
@@ -14,11 +14,16 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\BackgroundJob\IJobList;
 use OCP\BackgroundJob\TimedJob;
+use OCP\IConfig;
 use OCP\IUserManager;
 use Psr\Log\LoggerInterface;
+use function max;
 use function sprintf;
 
 class PreviewEnhancementProcessingJob extends TimedJob {
+	private const DEFAULT_INTERVAL = 3600;
+	private const MIN_INTERVAL = 60;
+
 	private IUserManager $userManager;
 	private AccountService $accountService;
 	private LoggerInterface $logger;
@@ -30,7 +35,8 @@ class PreviewEnhancementProcessingJob extends TimedJob {
 		AccountService $accountService,
 		PreprocessingService $preprocessingService,
 		LoggerInterface $logger,
-		IJobList $jobList) {
+		IJobList $jobList,
+		IConfig $config) {
 		parent::__construct($time);
 
 		$this->userManager = $userManager;
@@ -39,7 +45,16 @@ class PreviewEnhancementProcessingJob extends TimedJob {
 		$this->jobList = $jobList;
 		$this->preprocessingService = $preprocessingService;
 
-		$this->setInterval(3600);
+		// Allow admins to tighten the interval on setups where preview data
+		// directly gates user-visible behaviour (e.g. the imip_message flag
+		// that IMipMessageJob depends on to turn incoming invitations into
+		// calendar events). A floor of MIN_INTERVAL keeps a misconfigured
+		// value from hammering the DB.
+		$configured = $config->getSystemValueInt(
+			'app.mail.preview-enhancement-interval',
+			self::DEFAULT_INTERVAL,
+		);
+		$this->setInterval(max(self::MIN_INTERVAL, $configured));
 		$this->setTimeSensitivity(self::TIME_SENSITIVE);
 	}
 
--- a/lib/BackgroundJob/PreviewEnhancementProcessingJob.php
+++ b/lib/BackgroundJob/PreviewEnhancementProcessingJob.php
@@ -8,19 +8,21 @@ declare(strict_types=1);
 
 namespace OCA\Mail\BackgroundJob;
 
+use OCA\Mail\AppInfo\Application;
 use OCA\Mail\Service\AccountService;
 use OCA\Mail\Service\PreprocessingService;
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):
Looks good, thanks!

This should also be reflected at https://docs.nextcloud.com/server/latest/admin_manual/groupware/mail.html.

ChristophWurst:
Agent and model is missing from the commit. See https://github.com/nextcloud/mail/blob/main/AGENTS.md#commit-message-format

lubo92:
Thanks @ChristophWurst for the review.

Rebased on `main` and amended the commit with the required `AI-assisted:` trailer per [AGENTS.md](https://github.com/nextcloud/mail/blob/main/AGENTS.md#commit-message-format). No code changes since approval.

I'll open a follow-up PR against `nextcloud/documentation` to mirror this in [admin_manual/groupware/mail](https://docs.nextcloud.com/server/latest/admin_manual/groupware/mail.html) once this lands so the docs reference a released config key.

lib/BackgroundJob/PreviewEnhancementProcessingJob.php
```diff
@@ -39,7 +45,16 @@ public function __construct(ITimeFactory $time,
...
 		$this->preprocessingService = $preprocessingService;
 
-		$this->setInterval(3600);
+		// Allow admins to tighten the interval on setups where preview data
+		// directly gates user-visible behaviour (e.g. the imip_message flag
+		// that IMipMessageJob depends on to turn incoming invitations into
+		// calendar events). A floor of MIN_INTERVAL keeps a misconfigured
+		// value from hammering the DB.
+		$configured = $config->getSystemValueInt(
+			'app.mail.preview-enhancement-interval',
```
kesselb:
```suggestion
			'app_mail_preview_enhancement_interval',
```

New configuration options (that go into `config.php`) should not use `.` anymore.

kesselb: For new code I would in general prefer IAppConfig rather then config.php. For example like `abuse_number_of_recipients_per_message_threshold`

kesselb (changes requested):

lubo92:
@kesselb thanks for the review. Pushed a fixup commit (1cbb893) addressing both points:

- switched to `IAppConfig::getValueInt('mail', 'preview_enhancement_interval', …)`, modelled on `AntiSpamService` / the `abuse_number_of_recipients_per_message_threshold` example you pointed at
- exposed `PreviewEnhancementProcessingJob::CONFIG_KEY_INTERVAL` constant so other call sites (CLI, settings page) can reference the key without re-typing it
- updated `PreviewEnhancementProcessingJobTest` to mock `IAppConfig` and pass through the default

Per AGENTS.md I went with `--fixup` rather than amending, but heads-up for whoever does the squash: the original commit body still references the old `occ config:system:set app.mail.preview-enhancement-interval` example. The new equivalent is:

```
occ config:app:set mail preview_enhancement_interval --value=300 --type=integer
```

Happy to reword inline once the squash happens, or to push a `git commit --squash` instead if you'd prefer the rewording be reviewable now.

lubo92:
Docs companion: opened **nextcloud/documentation#14845** as a draft against `master`. It mirrors this PR's review-state config key (`preview_enhancement_interval` via `occ config:app:set`) and stays draft until this PR lands so we don't ship docs for a key that doesn't exist yet.

lubo92:
Heads-up: the just-failed Test workflow ([run 25744545267](https://github.com/nextcloud/mail/actions/runs/25744545267)) ran against the **pre-fixup** HEAD `2432e924`. The 4 errors are all the same:

```
ArgumentCountError: Too few arguments to function PreviewEnhancementProcessingJob::__construct(), 6 passed … and exactly 7 expected
```

That's the original commit having added the constructor param without updating the test — my bad. The fixup commit `1cbb893` already fixes it (test now mocks `IAppConfig` and passes 7 args), but the workflow on that head is still gated on the first-time-contributor approval. Could you approve the run on `1cbb893` so we can confirm green?

lubo92:
Pre-empting a likely false alarm: once CI runs on `1cbb893`, the `Block fixup and squash commits` + `Block unconventional commits` checks will fail because the commit subject starts with `fixup!`. Both are the [expected-to-fail](https://github.com/nextcloud/mail/blob/main/AGENTS.md#pr-review-workflow) clean-history checks per AGENTS.md — they clear at rebase/squash. The substantive checks (lint, static analysis, tests) are the ones to watch.

PR #12800, opened 2026-04-23
feat: jmap support - part 2

### LINKED ISSUE #12610: Read-only JMAP Support
## Summary

Implement read-only JMAP support: connect to JMAP servers, sync mailbox structure, and fetch messages.

## Problem

Users with JMAP-enabled accounts cannot access their mail via Nextcloud Mail. 
We need to add JMAP as a supported protocol with feature parity to IMAP for reading mail.

## Solution

- Create JMAP client factory using SebastianKrupinski/jmap-client-php
- Implement mailbox structure sync and conversion to domain models
- Implement message fetching and conversion to domain models
- Handle JMAP Session endpoint discovery and state-based sync
- Ensure output uses same domain models as IMAP

## Implementation

Create `lib/JMAP/` namespace for JMAP-specific logic:
- `JmapClientFactory` - creates and configures JMAP client
- `JmapMailboxMapper` - converts JMAP Mailbox objects to domain models
- `JmapMailboxSync` - syncs folder structure into database
- `JmapMessageMapper` - converts JMAP Message objects to MessageDTO
- `JmapSync` - orchestrates message sync using JMAP state tokens

Credential handling reuses existing encryption patterns.

Session discovery fetches `.well-known/jmap` for JMAP capabilities.

Sync strategy:
- Full sync on first run
- Incremental sync using JMAP state tokens on subsequent runs
- Detects new/changed/deleted messages

Message extraction:
- Headers (To, From, Subject, Date, Message-ID, References, In-Reply-To)
- Flags and keywords
- Attachment metadata
- Message IDs for threading

## Affected Components

- New: `lib/JMAP/` directory with mappers and sync logic
- `ProtocolClientFactory`: register JmapClientFactory (from https://github.com/nextcloud/mail/issues/12609)
- `SyncService`: use dispatcher to call appropriate sync
- `composer.json`: add SebastianKrupinski/jmap-client-php dependency

## Acceptance Criteria

- JMAP accounts can be created (via https://github.com/nextcloud/mail/issues/12609 infrastructure)
- Connection test validates JMAP server availability
- Mailbox structure syncs correctly
- Messages fetch with all required properties
- Output matches IMAP mapper format (same domain models)
- Existing IMAP sync unaffected

## Out of Scope

- Message mutations (flags, move, delete) - see https://github.com/nextcloud/mail/issues/12611
- UI changes - see https://github.com/nextcloud/mail/issues/12612
- Auto-detection

### LINKED ISSUE #12611: JMAP Write Operations
## Summary

Implement write operations for JMAP accounts: flag changes, move, copy, and delete.

## Problem

JMAP accounts are read-only without write operations. Users cannot change flags, organize messages, 
or delete mail.

## Solution

- Implement flag operations using JMAP Message/Email API
- Implement move/copy using JMAP mailboxIds property
- Implement delete
- Ensure all operations work identically to IMAP operations from user perspective

## Implementation

Extend `lib/JMAP/` namespace with operations:
- Operation implementations for flag changes, move, copy, delete
- Reuse existing service layer abstractions via dispatcher pattern

JMAP-specific details:
- Flags use JMAP `keywords` property (no separate read/unread flag)
- Move uses `mailboxIds` (JMAP allows multi-mailbox; simulate single mailbox behavior)
- Copy adds mailbox ID to `mailboxIds`
- Delete moves to trash or sets destroyed flag

Service layer remains protocol-agnostic via dispatcher pattern.

Error handling catches JMAP errors and translates to service exceptions.

## Affected Components

- `lib/JMAP/`: operation implementations
- Existing service layer: use dispatcher or extend as needed

## Acceptance Criteria

- Flag operations (read/unread, custom flags) work for JMAP accounts
- Move/copy/delete operations work for JMAP accounts
- Operations persist on server
- Operations sync back correctly on next sync
- Behavior matches IMAP for equivalent operations

## Out of Scope

- UI changes
- JMAP-specific optimizations

SebastianKrupinski:
### Summary
Adds a protocol abstraction so Mail can talk to both IMAP and JMAP servers through a
common set of interfaces, and reworks the existing IMAP code to implement them.

- Resolves: https://github.com/nextcloud/mail/issues/12610
- Resolves: https://github.com/nextcloud/mail/issues/12611
- Requires: https://github.com/nextcloud/mail/pull/12711
- Implements `ProtocolFactory` — selects the right connector for an account's protocol
- Implements `IMailboxConnector` (+ `ImapMailboxConnector`, `JmapMailboxConnector`)
- Implements `IMessageConnector` (+ `ImapMessageConnector`, `JmapMessageConnector`)

### Scope
- This PR covers **mailbox operations** and **message operations** ONLY.
  Sending operations are in https://github.com/nextcloud/mail/pull/12871.
- DB changes (remote id / sync state columns) required for https://github.com/nextcloud/mail/pull/12711

### Automated tests
- [x] `composer psalm` clean
- [x] `composer test:unit` green
- [x] `composer test:integration` green
- [x] `composer cs:check` clean

### Testing
Every operation that talks to an external server must be tested **twice** — once against an IMAP account/server and once against a JMAP account/server.

Copy a block below into a comment and tick the boxes for the server you tested against.
Each operation has one checkbox per protocol.

#### Mailbox operations

- List / sync folders
  - [x] IMAP
  - [x] JMAP
- Create in root folder
  - [x] IMAP
  - [x] JMAP
- Create in another folder
  - [x] IMAP
  - [x] JMAP
- Rename in root folder
  - [x] IMAP
  - [x] JMAP
- Rename in another folder
  - [x] IMAP
  - [x] JMAP
- Delete in root folder
  - [x] IMAP
  - [x] JMAP
- Delete in another folder
  - [x] IMAP
  - [x] JMAP
- Subscribe / unsubscribe
  - [x] IMAP
  - [x] JMAP
- Empty / clear mailbox
  - [x] IMAP
  - [x] JMAP

#### Message operations

- List / sync messages
  - [x] IMAP
  - [x] JMAP
- Fetch message
  - [x] IMAP
  - [x] JMAP
- Download raw message (show source / .eml)
  - [x] IMAP
  - [x] JMAP
- Download attachments
  - [x] IMAP
  - [x] JMAP
- Search messages
  - [ ] IMAP
  - [ ] JMAP
- Move message
  - [x] IMAP
  - [x] JMAP
- Move thread
  - [x] IMAP
  - [x] JMAP
- Delete message
  - [x] IMAP
  - [x] JMAP
- Delete thread
  - [x] IMAP
  - [x] JMAP
- Flag message (read/unread, etc.)
  - [x] IMAP
  - [x] JMAP
- Tag message (set/unset)
  - [x] IMAP
  - [x] JMAP
- Quota (usage reported)
  - [x] IMAP
  - [ ] JMAP

#### Console operations
- [x] `mail:account:create` — alias of `mail:account:create-imap`
- [x] `mail:account:create-imap`
- [x] `mail:account:create-jmap`
- [x] `mail:account:debug`
- [x] `mail:account:delete`
- [x] `mail:account:diagnose` — alias of `mail:account:test`
- [x] `mail:account:export`
- [x] `mail:account:export-threads`
- [x] `mail:account:run-meta-estimator`
- [x] `mail:account:sync`
- [x] `mail:account:test`
- [x] `mail:account:train`
- [x] `mail:account:update`
- [x] `mail:clean-up`
- [ ] `mail:predict-importance`
- [ ] `mail:repair:system-autoresponders`
- [x] `mail:repair:tags`
- [x] `mail:tags:migration-jobs`
- [x] `mail:thread`


ChristophWurst (commented):
This is a more than challenging review give then amount of changes, slicing technique and architectural decisions, so I used Claude Fable for assistance. Here is the result:

---

# JMAP PR stack: connector architecture exceeds ticket scope

Applies to #12711 (part 1), #12800 (part 2), #12871 (part 2b). Written as review input for the architecture discussion.

## What the tickets scoped

- #12609: *"Create `ProtocolClientFactory` dispatcher … The dispatcher checks `$account->getProtocol()` and returns the appropriate client factory."* — a dispatcher at the **client** level. Services keep their shape and ask one factory for the right client.
- #12610: *"Create `lib/JMAP/` namespace … `JmapMailboxSync`, `JmapSync` … `SyncService`: use dispatcher to call appropriate sync."* — JMAP **twins of the existing IMAP-side classes**, selected at the sync entry point.
- #12611: *"Reuse existing service layer abstractions via dispatcher pattern. Service layer remains protocol-agnostic."* — **reuse**, not replace.

Under that design, `MailManager`, `IMailManager`, and their consumers stay untouched; new code is additive in `lib/JMAP/`.

## What was built instead

A strategy-pattern connector layer that replaces the service layer's internals:

- Three new interfaces: `IMailboxConnector` (6 methods), `IMessageConnector` (15), `ITransmissionConnector` (3)
- `ProtocolFactory` resolving implementations from a class map through the DI container
- Four connector classes (~1,300 lines) plus a 991-line `JmapOperationsService`
- A rewritten `MailManager`
- Deletion of the `IMailManager` contract (349 lines of the app's central service boundary) — mentioned in no ticket

## Findings

### 1. The `IMailManager` deletion inflates #12800 without buying anything

Dozens of files changed *only* to swap the import and type hint from `Contracts\IMailManager` to the concrete `Service\MailManager` (e.g. `PageController`, `TagsController` — two-line changes each; 46 `IMailManager` reference lines across the diff). The consumer diffs show the method signatures did not change — keeping the interface would have cost nothing and erased nearly all of that churn from a 111-file PR. Deleting a single-implementation interface is defensible housekeeping *on its own*; inside the highest-risk feature PR of the release it is noise that buries the real changes, and it downgrades every consumer from interface injection to concrete-class coupling as a side effect.

### 2. The "protocol-neutral" interface is IMAP wearing a new name

`IMessageConnector`:

- returns `IMAPMessage[]` from `fetchMessages()`
- takes Horde-style `int $criteria` and `int[] $knownUids` in `syncMailbox()`
- carries a `LoggerInterface` parameter through a domain interface
- includes `isPermflagsEnabled()` — an IMAP capability with no JMAP meaning

This is the classic failure mode of abstracting from a single implementation: `MailManager`'s old internals were hoisted into an interface and declared protocol-agnostic. JMAP is then squeezed through IMAP-shaped holes — its string ids ride in a bolted-on `remote_id` column while the interface still speaks integer UIDs. Once part 2 merges, this surface calcifies; every future protocol and every JMAP bug fix pays for it.

### 3. The abstraction is not where the work happens

`JmapMessageConnector` is a forwarding shell: every method is `$this->jmapOperationsService->connect($account)` plus one delegation (`entityDelta`, `entityFetchMessage`, `entityMove`, `entityDelete`, `entityModifyFlags`, … — lines 65–262 on `feat/jmap-2`). The real JMAP logic lives in `JmapOperationsService`, *beside* the abstraction, not behind it. The request path is now:

```
MailManager → ProtocolFactory (container lookup via class map)
            → JmapMessageConnector → JmapOperationsService
            → JmapClient → adapter → IClient
```

Two new indirection layers where the tickets asked for zero — and one of them demonstrably adds nothing but `connect()` calls.

### 4. After all that, the goal is not reached

The rewritten `MailManager` — the thing the exercise was meant to make protocol-agnostic — still injects `ImapMessageMapper` and `ImapFlag`, and still documents `Horde_Imap_Client_Exception` in its PHPDoc (`lib/Service/MailManager.php` lines 12–14, 47, 54, 411–414 on `feat/jmap-2`). Line 463 runs `$this->imapFlag->create($displayName)` to build tag labels regardless of account protocol — at best dead weight, at worst wrong behavior for JMAP accounts (needs an answer in the part-2 review). The abstraction is simultaneously bigger than scoped *and* incomplete on its own terms.

### 5. It is an unticketed architecture decision

AGENTS.md asks for complex, multi-subsystem, architecture-level changes to go through a ticket for alignment *before* implementation. The connector layer, the container-map resolution, and the `IMailManager` removal appear in no ticket; reviewers are being asked to ratify a decade-long maintenance commitment inside a feature diff.

## The steelman

Dispatching in one seam beats `if ($protocol === …)` sprinkled through ~30 `MailManager` methods, and the ticket design *would* have produced some of that branching. The connector style matches other Nextcloud integration apps and leaves a socket for a hypothetical third protocol. Those are real arguments — for a design that should have been proposed on #12609 and, if accepted, landed as an IMAP-only prep refactor. The problem is not that a connector layer is indefensible; it is that its cost/benefit was never weighed outside the implementation, and the version that arrived is IMAP-flavored, double-layered, and bundled with unrelated interface deletion.

## Proposed conditions (no re-slice of the PR stack)

Before part 2 merges — the last moment any of this is cheap:

1. **Ratify or reject the connector architecture explicitly on #12609/#12610.** One comment thread, one decision. Realistically: ratify with the conditions below.
2. **Restore `IMailManager`** (same methods, `MailManager` implements it). Signatures did not change, so this is mechanical, deletes most of the consumer churn from the diff, and preserves interface injection.
3. **De-IMAP-ify `IMessageConnector` now**: neutral message DTO or a documented rationale for `IMAPMessage`; drop the `LoggerInterface` parameter; protocol-neutral sync criteria; decide the uid-vs-remote-id story at the interface level.
4. **Collapse `JmapMessageConnector` into `JmapOperationsService`** (or have the ops service implement the interface directly) — one layer, not two.
5. **Resolve the `ImapFlag`/Horde residue in `MailManager`** — either it is protocol-agnostic or it is not; the tag-creation path for JMAP accounts needs an answer.

Items 2–4 are hours, not days.

---

cc @kesselb for these fundamental changes to the codebase

kesselb:
Hey,

12800 nice pr number ;)

Something I've overlooked in the first jmap pr:

Several methods in IMessageConnector take the message list as a variadic (`Message ...$messages`). Could we change these to a plain `array $messages` (with a `@param Message[] $messages` docblock) instead?

Both implementations treat `$messages` as an array right away internally (`array_map`, `foreach`, or re-spreading into another call, there's no use of the variadic for anything positional.

At the call sites we already hold an array in most cases and just spread it (`...$messages`), and in a couple of places we even wrap a single value to spread it (`...[$message]`), which a plain array param avoids entirely.

SebastianKrupinski:
Hey 

> Several methods in IMessageConnector take the message list as a variadic (`Message ...$messages`). Could we change these to a plain `array $messages` (with a `@param Message[] $messages` docblock) instead?

Sure I can change this... back to arrays (I had arrays initially)... buy I decided on using variadic notation because it gives actual type safety not just while using psalm.

> wrap a single value to spread it (`...[$message]`), which a plain array param avoids entirely.

This was a dumb over site on my part... from the original array iteration.

 
Any chance we can make this change after this is merged? As it make is messy right now with multiple PR's based on the already merged code to change and alter all the tests again.

I can make this change while I wait for the front end changes to be approved

lib/BackgroundJob/SyncJob.php
```diff
@@ -116,8 +116,12 @@ protected function run($argument) {
...
 
 		try {
-			$this->mailboxSync->sync($account, $this->logger, true);
-			$this->syncService->syncAccount($account, $this->logger);
+			$this->protocolFactory
+				->mailboxConnector($account)
+				->syncAll($account, true);
+			$this->protocolFactory
+				->messageConnector($account)
+				->syncAll($account, true);
```
kesselb:
```suggestion
				->syncAll($account, false);
```

lib/Command/SyncAccount.php
```diff
@@ -89,9 +85,5 @@ private function sync(Account $account, bool $force, OutputInterface $output): v
 			$output->writeln("<info>Batch of new messages sync'ed. " . $mbs . 'MB of memory in use</info>');
 			$this->sync($account, $force, $output);
 		}
-
-		foreach ($this->clientFactory->getLoginStats() as $host => $count) {
-			$consoleLogger->debug(sprintf('%d IMAP connection(s) to %s', $count, $host));
```
kesselb, line 94: Is that coming back for imap? ;)

lib/Controller/MessagesController.php
```diff
@@ -214,24 +212,19 @@ public function getBody(int $id): JSONResponse {
...
-			$json = $imapMessage->getFullMessage($id);
-		} finally {
-			$client->logout();
+		if ($imapMessage->hasHtmlMessage()) {
+			$cacheInstance->set($imapMessageCacheKey, $imapMessage->getHtmlBody($id), 600);
 		}
 
+		$json = $imapMessage->getFullMessage($id);
+
 		$itineraries = $this->itineraryService->getCached($account, $mailbox, $message->getUid());
```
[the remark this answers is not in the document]
SebastianKrupinski: Fixed

lib/Db/Mailbox.php
```diff
@@ -127,7 +127,8 @@ public function isSpecialUse(string $specialUse): bool {
 	public function isCached(): bool {
 		return $this->getSyncNewToken() !== null
 			&& $this->getSyncChangedToken() !== null
-			&& $this->getSyncVanishedToken() !== null;
+			&& $this->getSyncVanishedToken() !== null
+			|| $this->getState() !== null;
 	}
```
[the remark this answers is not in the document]
SebastianKrupinski: Changed

lib/Db/MessageMapper.php
```diff
@@ -751,6 +753,74 @@ public function deleteByUid(Mailbox $mailbox, int ...$uids): void {
...
+		$selectMessageIdsQuery = $this->db->getQueryBuilder();
+		$deleteRecipientsQuery = $this->db->getQueryBuilder();
+		$deleteMessagesQuery = $this->db->getQueryBuilder();
+
+		$selectMessageIdsQuery->select('id')
+			->from($this->getTableName())
+			->where(
+				$selectMessageIdsQuery->expr()->eq('mailbox_id', $selectMessageIdsQuery->createNamedParameter($mailbox->getId())),
+				$selectMessageIdsQuery->expr()->in('remote_id', $deleteMessagesQuery->createParameter('remote_ids')),
+			);
```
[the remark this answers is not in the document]
SebastianKrupinski: identical pattern exists one function up, at [MessageMapper.php:723 (deleteByUid), using $deleteMessagesQuery->createParameter('uids') inside $selectMessageIdsQuery

lib/Command/SyncAccount.php
SebastianKrupinski: This will have to be reworked for both protocols. We can't really have protocol specific things in any commands or controllers. I've added it to my list of follow up PR's

lib/BackgroundJob/SyncJob.php
SebastianKrupinski: Done

lib/IMAP/MailboxSync.php
```diff
@@ -72,13 +73,13 @@ public function sync(Account $account,
 		LoggerInterface $logger,
 		bool $force = false,
 		?Horde_Imap_Client_Socket $client = null): void {
-		if (!$force && $account->getMailAccount()->getLastMailboxSync() >= ($this->timeFactory->getTime() - 7200)) {
+		if (!$force && $account->getMailAccount()->getLastMailboxSync() >= ($this->timeFactory->getTime() - 900)) {
```
kesselb, line 76: ?

lib/Command/TestAccount.php
```diff
@@ -36,118 +69,414 @@ protected function configure(): void {
...
 	}
 
-	private function testJmap(\OCA\Mail\Account $account, OutputInterface $output): int {
-		$output->writeln('Testing JMAP connection...');
+	private function testJmap(Account $account, SymfonyStyle $io, int $mailboxLimit, int $messageLimit): int {
+		$io->section('JMAP Test');
 
 		$mailAccount = $account->getMailAccount();
 		$sslMode = $mailAccount->getInboundSslMode();
 		$scheme = ($sslMode === 'none') ? 'http' : 'https';
```
kesselb:
```suggestion
		$scheme = ($sslMode === 'yes') ? 'https' : 'http';
```

Like the factory: https://github.com/nextcloud/mail/blob/2bf5a3b6b6e7af32202dcdbe120333fad79b268c/lib/JMAP/JmapClientFactory.php#L48

lib/Command/TestAccount.php
```diff
@@ -36,118 +69,414 @@ protected function configure(): void {
...
+		$messages = $this->jmapOperationsService->entityList(
+			$inbox->getRemoteId(),
+			null,
+			[['attribute' => 'received', 'direction' => true]],
+			['anchor' => 'absolute', 'position' => 0, 'tally' => $messageLimit]
+		);
+
+		/** @var list<Message> $messageList */
+		$messageList = $messages['list'] ?? [];
+		$this->renderMessageTable($io, $this->buildJmapMessageRows($messageList), 'No recent messages found in INBOX.');
```
kesselb:
```suggestion
		$this->renderMessageTable($io, $this->buildJmapMessageRows($messageList));
```

Inline the "no recent messages found" message?

lib/BackgroundJob/RepairSyncJob.php
```diff
@@ -65,6 +68,19 @@ protected function run($argument): void {
 			return;
 		}
 
+		$this->protocolFactory
+			->mailboxConnector($account)
+			->syncAll($account, true);
+
+		if ($account->getMailAccount()->getProtocol() !== MailAccount::PROTOCOL_IMAP) {
```
kesselb: Move the check/early-exit before canAuthenticateImap

lib/BackgroundJob/RepairSyncJob.php
```diff
@@ -65,6 +68,19 @@ protected function run($argument): void {
 			return;
 		}
 
+		$this->protocolFactory
+			->mailboxConnector($account)
+			->syncAll($account, true);
```
kesselb, line 82: Please explain

lib/Service/MailManager.php
```diff
@@ -310,339 +246,146 @@ public function deleteMessageWithClient(
 		} catch (DoesNotExistException $e) {
 			throw new ServiceException('No trash folder', 0, $e);
 		}
+		$operation = $sourceMailbox->getId() === $trashMailbox->getId() ? 'delete' : 'move';
```
kesselb, line 295: Const for delete and move

lib/Service/MailManager.php
```diff
@@ -228,79 +194,49 @@ public function getSource(Horde_Imap_Client_Socket $client,
...
-		Account $account,
-		Mailbox $mailbox,
-		int $messageUid,
-		Horde_Imap_Client_Socket $client,
-	): void {
-		$this->eventDispatcher->dispatchTyped(
-			new BeforeMessageDeletedEvent($account, $mailbox->getName(), $messageUid)
-		);
+	public function deleteMessage(Account $account, Mailbox $mailbox, Message $message): void {
+		$this->deleteMessages($account, $mailbox, ...[$message]);
```
kesselb:
```suggestion
		$this->deleteMessages($account, $mailbox, $message);
```

lib/Db/Mailbox.php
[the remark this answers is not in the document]
kesselb: Did you push?

lib/Db/Mailbox.php
```diff
@@ -125,9 +125,13 @@ public function isSpecialUse(string $specialUse): bool {
...
 
 	public function isCached(): bool {
-		return $this->getSyncNewToken() !== null
+		return (
```
kesselb:
```php
		if ($this->remoteId !== null) {
			return $this->getState() !== null;
		}

		return $this->getSyncNewToken() !== null
			&& $this->getSyncChangedToken() !== null
			&& $this->getSyncVanishedToken() !== null;
```

I would prefer splitting the if-branches for readability.

Please also take a look at Mailbox.getCacheBuster as it might need an update as well.

lib/Db/MailboxMapper.php
```diff
@@ -169,6 +169,38 @@ public function findAccountIdForMailbox(int $mailboxId): int {
 		return (int)$row['account_id'];
 	}
 
+	/**
+	 * @throws DoesNotExistException
+	 * @throws ServiceException
+	 */
+	public function findByName(Account $account, string $name): Mailbox {
```
kesselb: Why would you not use find? ;)

Lookup by name should use the name_hash column because of the `inbox ` vs `inbox` issue (some db consider them the same).

lib/Db/TagMapper.php
```diff
@@ -80,6 +80,7 @@ public function tagMessage(Tag $tag, string $messageId, string $userId): void {
 		try {
 			$tag = $this->getTagByImapLabel($tag->getImapLabel(), $userId);
 		} catch (DoesNotExistException $e) {
+			$tag->setUserId($userId);
```
kesselb, line 83: Nice catch. Could you please re-submit as own PR?

lib/IMAP/ImapMessageConnector.php
```diff
@@ -0,0 +1,431 @@
...
+				$criteria,
+				$knownUids,
+				$force,
+			);
+		} finally {
+			$client->logout();
+		}
+
+		return new SyncResult(
+			state: $mailbox->getSyncChangedToken(),
```
kesselb, line 74: What are you plans here?

While for jmap only one state exist, for imap it's usually the combination of sync new, sync changed and sync vanished.

Currently SyncResult is not consumed, hence can we drop the return? Otherwise please use something like new|changed|vanished.

lib/IMAP/PreviewEnhancer.php
```diff
@@ -77,6 +78,19 @@ public function process(Account $account, Mailbox $mailbox, array $messages, boo
 			return $messages;
 		}
 
+		if ($account->getMailAccount()->getProtocol() !== MailAccount::PROTOCOL_IMAP) {
```
kesselb:
```suggestion
		if ($account->getMailAccount()->getProtocol() === MailAccount::PROTOCOL_JMAP) {
```

lib/JMAP/JmapMailboxAdapter.php
```diff
@@ -0,0 +1,169 @@
...
+		if ($role === null) {
+			return null;
+		}
+
+		$role = strtolower(trim($role, '\\'));
+		if ($role === 'flagged') {
+			return 'important';
+		}
+
+		$allowed = ['all', 'archive', 'drafts', 'important', 'inbox', 'junk', 'sent', 'trash'];
```
kesselb, line 119: Const?

lib/JMAP/JmapMailboxAdapter.php
```diff
@@ -0,0 +1,169 @@
...
+
+	private function convertToAcl(MailboxParametersResponse $response): ?string {
+		$rights = $response->rights();
+		if ($rights === null) {
+			return null;
+		}
+
+		/** @var string $acls */
+		$acls = '';
+		if ($rights->readItems() === true) {
```
kesselb:
```suggestion
		if ($rights->readItems()) {
```

lib/JMAP/JmapMailboxAdapter.php
```diff
@@ -0,0 +1,169 @@
...
+		return in_array($role, $allowed, true) ? $role : null;
+	}
+
+	private function normalizeSpecialUse(string $role): string {
+		$role = strtolower($role);
+
+		return $role === 'important' ? 'flagged' : $role;
+	}
+
+	private function convertToAcl(MailboxParametersResponse $response): ?string {
```
kesselb, line 130:
```suggestion
	private function convertToAcl(?MailboxPermission $rights): ?string {
```

lib/JMAP/JmapMailboxConnector.php
```diff
@@ -0,0 +1,316 @@
...
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Db\TTransactional;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\EventDispatcher\IEventDispatcher;
+use Psr\Log\LoggerInterface;
+
+class JmapMailboxConnector implements IMailboxConnector {
+	use TTransactional;
+
+	private const MAILBOX_SYNC_TTL = 300;
```
kesselb, line 28: Move const to the interface or would we use different ttl for jamp/imap?

lib/Db/MailboxMapper.php
kesselb: Why would you not use find?

For selecting by name we  should use name_hash because for some db `INBOX ` = `INBOX`.

kesselb (commented):
Made it till JmapMailboxConnector

kesselb:
### Some context in the description please

109 files, 5467 additions, ~2498 deletions.

That's a lot to take in, and right now the PR doesn't say why any of it looks the way it does. I don't need a walkthrough of every change, the diff already tells me what happened. What I'm missing is the reasoning behind the structural decisions: which boundaries you drew, and what the alternatives were that you rejected. Without that I can't tell an intentional design from an accident, which is what I'd be approving here.

As Christoph already mentioned, dropping IMailManager is the prime example.

### Lost annotations

We're losing a lot of helpful annotations, the thrown exceptions in particular. getImapMessage is the clearest example, see the usage in SubmitContentJob.

On main you can see that IMailManager.getImapMessage may throw a ServiceException or a SmimeDecryptException, and the implementation in MailManager enforces that by wrapping DoesNotExistException, Horde_Mime_Exception and Horde_Imap_Client_Exception into a ServiceException.

On this branch:

- IMessageConnector.fetchMessages annotates no exceptions, so the contract we give API consumers says the method doesn't leak one.
- ImapMessageConnector.fetchMessages wraps nothing, so everything findByIds throws leaks straight through.
- MailManager.getImapMessage has no docblock at all, and it throws a ClientException that nothing documents.
- SubmitContentJob doesn't handle that ClientException either. It extends \Exception directly, so it's a sibling of ServiceException and the existing catch misses it. A message that's gone from the remote server now kills the job instead of skipping one message.

Independent of the IMailManager question, the new interfaces should carry proper throw annotations. I'm quite sure we have tons of missing or incomplete docblocks on main, and I'm not asking you to fix those, but we should maintain the current level for the code you touch here.

 getImapMessages still documents `@param int[] $uids`, which doesn't exist anymore.

### Variadics madness

I'd like to understand the reasoning, because passing lists of objects around seems cleaner to me.

- MailManager.tagMessages -> messageConnector.tagMessages, spread just to spread again, and then array_map'd into a plain UID array anyway
- MailManager.deleteMessages -> messageConnector.moveMessages and messageConnector.deleteMessages
- MessageMapper.deleteByRemoteIds
- ImapMessageConnector.fetchMessages

The usage in MessagesController.mdn, setFlags and setTag is another example why I'm absolutely not a fan of the variadic overuse here. Passing the list of messages last just reads weird:

`$this->mailManager->flagMessages($account, $mailbox, '$mdnsent', true, $message);`
`$this->mailManager->tagMessages($account, $mailbox, $tag, true, $message);`

In fairness, updateBulk and deleteByUid already use variadics on main, so you didn't invent this. But this PR extends the pattern much further, which makes it a good moment to decide deliberately.

### Clients and loops

In a couple of places, TrashRetentionJob for example, we previously re-used one client for a whole list of messages. I understand why the refactoring lost that, and for a background job the trade-off is acceptable, but please log a technical debt issue for it. I assume we can dispatch a single deleteMessages call, since all those messages are in the same mailbox, and optimize the retention mapper code path in the same go.

Also, in flagMessages, tagMessages, moveMessages and deleteMessages the $client->logout() sits after the try/catch instead of inside a finally.

### Unrelated changes

FollowUpClassifierJob currently carries two changes:

1. getByMessageId => getMessagesByMessageId
2. tagMessage => tagMessages

Might be that getMessagesByMessageId is the better name, might be that a single tagMessages is less to maintain. But both are unrelated to this PR and are just noise while reviewing. The second one is only half unrelated, since it also swaps string $mailboxName for a Mailbox, which the abstraction genuinely needs. Please say so in the description. In the same go, keeping MailManager.deleteMessage as a wrapper for the batch function is a bit inconsistent.

Renaming MessagesController.getSource to getRawMessage also seems unrelated to introducing JMAP, and it changes the route id mail.messages.getSource even though the URL stays the same. 

### Two things that look like bugs

- ImapMessageConnector.fetchMessages passes a hardcoded true as the 5th argument to MessageMapper.findByIds, which is $loadBody. Its own $loadBody parameter is ignored.
- DkimService and ItineraryService switched their cache keys from UID to database message id, but kept the key format (accountId_mailboxName_$id) and the 7 day TTL. Bumping a prefix on the cache key?

lib/Command/TestAccount.php
SebastianKrupinski: Sure

lib/Command/TestAccount.php
SebastianKrupinski: Sure. Done

lib/Db/TagMapper.php
SebastianKrupinski: Sure

lib/Db/Mailbox.php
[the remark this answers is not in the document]
SebastianKrupinski: yes

lib/Db/Mailbox.php
SebastianKrupinski: Sure

lib/JMAP/JmapMailboxConnector.php
SebastianKrupinski: I want to use independent mailbox sync values, because jmap has a delta sync for mailboxes. that means we can do a very quick check, very often, unlike imap which is slow due to the need to pull the entire tree.

lib/IMAP/MailboxSync.php
SebastianKrupinski: I can revert this, I lowered it for testing originally to 5min then when putting it back, I was like "we only sync mailboxes every two hours?" so made it 15min

SebastianKrupinski:
> ### Some context in the description please
> 
> 109 files, 5467 additions, ~2498 deletions.
> 
> That's a lot to take in, 

Not just right now, it was also a lot to de-spagettie, 10 years of code, and to implement a new protocol in 4 weeks.

> As Christoph already mentioned, dropping IMailManager is the prime example.

Well, did he really thought? Looks like it was Claude code that made the comments... But I will response never the less...

Before responding point by point, I want to clarify how I treated [#12610](https://github.com/nextcloud/mail/issues/12610) and [#12611](https://github.com/nextcloud/mail/issues/12611). They were a high-level Ai generated breakdown of an earlier Ai generated working plan, intended to define the outcomes and divide the work, rather than freeze the final class structure. #12610 asks for state-based JMAP mailbox and message sync using the same domain models as IMAP. #12611 asks for flag, move, copy, and delete operations, and explicitly says to reuse or extend the existing service-layer abstractions through a dispatcher so that the service layer remains protocol-agnostic. Neither ticket says that `MailManager` and its callers must remain untouched, and the implementation work showed that those existing abstractions were not actually protocol-neutral enough to be reused unchanged.

The implementation did drift from the class layout suggested in #12610. Instead of separate `JmapMailboxSync`, `JmapMessageMapper`, and `JmapSync` twins selected only by `SyncService`, it uses adapters, connectors, and `JmapOperationsService`. This was a deliberate change discovered during implementation: the Horde IMAP client and the JMAP client do not expose a common operational API, and several existing service methods encoded IMAP concepts such as Horde clients, mailbox names, and integer UIDs.

Message identity was the most important issue. The existing request path passed an integer IMAP UID from the controller through the service layer to the client. An IMAP UID is a native, mailbox-scoped remote identifier, whereas a JMAP Email ID is an opaque string that remains stable when the message moves between mailboxes. Preserving the old UID-based boundary for JMAP would have required synthetic integer UIDs to be kept aligned through every operation, including moves, even though those integers have no JMAP meaning. It would also require an additional database lookup for each operation to translate the supplied UID into the JMAP `remote_id`. Passing the persisted `Message` entity through the service boundary instead gives the connector both forms of identity and lets each protocol use its native identifier.

Because of that mismatch, dispatching only client construction would not have made the service layer protocol-agnostic. It would have retained IMAP identity semantics and moved the protocol branches and ID translation into downstream services. Dispatching complete protocol operations through connectors keeps callers independent of the concrete client and its identifier model. This is how I interpreted #12611's requirement that the service layer remain protocol-agnostic. In hindsight, the tickets' assumption that the existing abstractions could simply be reused through a client dispatcher was incomplete because it did not account for this identity-model difference.

I agree that this reasoning and the differences from the suggested class layout should have been explained in the PR description earlier. This context also does not make the incidental renames, the `IMailManager` removal, or the concrete bugs correct; those should be handled on their own merits. My point is only that the connector boundary was part of the plan, even though its concrete implementation evolved while making the IMAP-shaped code work with JMAP.

> ### Lost annotations

Agreed. The connector interfaces should document their exception contracts, and the implementations should consistently translate protocol-specific failures into the service-level exceptions expected by callers. I will also correct the stale `$uids` annotation and ensure that `SubmitContentJob` continues to skip a message that has disappeared from the remote server.

> ### Variadics madness

The variadics were a deliberate choice: they keep each argument typed as `Message` and are consistent with existing APIs such as `updateBulk()` and `deleteByUid()`. I do not think variadics are inherently a design problem here. That said, I see the readability issue when the messages follow several other arguments, and the repeated spreading adds little at the connector boundary. I am open to changing these batch connector methods to accept a consistently documented `Message[]` if that is the preferred convention.

> ### Clients and loops

Agreed on the `logout()` calls; client cleanup should be in a `finally` block, and I will correct those paths in this PR. Moving client ownership into the connector was deliberate so callers remain independent of protocol-specific clients, but an unintended consequence in `TrashRetentionJob` is that calling `deleteMessage()` in a loop creates a client for every message. A single `deleteMessages()` call would avoid that overhead. I agree that this batching and result-reporting path should be tracked as a focused follow-up rather than expanding this PR further.

> ### Unrelated changes

These changes came from making `MailManager` the protocol-independent boundary between IMAP and JMAP. In particular, the tagging path needs a `Mailbox` and `Message` entities so the connector can resolve protocol-specific identifiers rather than assume an IMAP mailbox name and UID.

> ### Two things that look like bugs

Agreed; both are bugs rather than intentional design changes. I will pass `$loadBody` through correctly and change the cache-key prefix so entries created with database message IDs cannot collide with or reuse entries created with IMAP UIDs.

lib/Db/MailboxMapper.php
SebastianKrupinski: Removed. This was a accidental left over from a previous version of the implementation

SebastianKrupinski: Removed. This was a accidental left over from a previous version of the implementation

lib/IMAP/PreviewEnhancer.php
SebastianKrupinski: Sure

lib/JMAP/JmapMailboxAdapter.php
SebastianKrupinski: Why?

lib/JMAP/JmapMailboxAdapter.php
SebastianKrupinski: I prefer the explicit true, but sure

lib/Service/MailManager.php
SebastianKrupinski: Why? the values are only used this one function, and are just a way to identify which operation we are doing. they have no actual meaning in jmap, this can be "m" and "d"

SebastianKrupinski (commented):
.

SebastianKrupinski:
> Made it till JmapMailboxConnector

Thank you, for getting that far

lib/Service/JMAP/JmapOperationsService.php
```diff
@@ -0,0 +1,1024 @@
...
+	 * @throws ServiceException
+	 */
+	public function connect(Account $account): bool {
+		$this->dataStore = $this->protocolFactory->jmapClient($account);
+		try {
+			// evaluate if client was already connected
+			if (!$this->dataStore->sessionStatus()) {
+				$this->dataStore->connect();
+			}
+		} catch (Exception $e) {
```
kesselb:
```suggestion
		} catch (\Throwable $e) {
```

because everything happens in the lib?

lib/JMAP/JmapMessageConnector.php
```diff
@@ -0,0 +1,434 @@
...
+			$filters[] = [
+				'attribute' => 'body',
+				'value' => $textToken,
+			];
+		}
+
+		return $filters;
+	}
+
+	private function mapMessagesByRemoteId(Message ...$messages): array {
```
kesselb, line 326:
```suggestion
   /**
	 * @return array<string, Message>
	 */
	private function mapMessagesByRemoteId(Message ...$messages): array {
```

lib/Protocol/ProtocolFactory.php
```diff
@@ -62,6 +67,28 @@ public function jmapClient(Account $account): JmapClient {
...
 	}
 
+	/**
+	 * @throws ServiceException
+	 */
+	public function testConnection(Account $account): void {
+		$protocol = $account->getMailAccount()->getProtocol();
+
+		if ($protocol === MailAccount::PROTOCOL_IMAP) {
+			$this->imapClient($account)->close();
```
kesselb, line 81: Can throw Horde_Imap_Client_Exception

lib/Protocol/ProtocolFactory.php
```diff
@@ -62,6 +67,28 @@ public function jmapClient(Account $account): JmapClient {
...
+			return;
+		}
+
+		if ($protocol === MailAccount::PROTOCOL_JMAP) {
+			$client = $this->jmapClient($account);
+			if (!$client->sessionStatus()) {
+				$client->connect();
```
kesselb, line 88: Can throw JmapTransportException?

lib/Service/JMAP/JmapOperationsService.php
```diff
@@ -0,0 +1,1024 @@
...
+			$this->dataStore->download($account, $identifier, $data, $type, $name);
+		} catch (Exception $e) {
+			throw new ServiceException('JMAP download failed: ' . $e->getMessage(), 0, $e);
+		}
+	}
+
+	/**
+	 * List of collections in remote storage
+	 *
+	 * @param string|null $location optional location constraint (e.g. parent collection id)
```
kesselb, line 159: Defining the shapes with psalm-type like below would be nice. For example to reuse the `JmapEntityListFilter` shape in JmapMessageConnector. 

```php
/**
 * @psalm-type JmapCollectionListFilter = array{
 *      attribute: 'name'|'role'|'hasRoles'|'subscribed',
 *      value: string,
 *  }
 *
 * @psalm-type JmapCollectionListSort = array{
 *      attribute: 'name'|'order',
 *      direction: 'descending',
 *   }
 *
 * @psalm-type JmapEntityListFilter = array{
 *     attribute: '*'|'in'|'inOmit'|'from'|'to'|'cc'|'bcc'|'subject'|'body'|'attachmentPresent'|'tagPresent'|'tagAbsent'|'before'|'after'|'min'|'max',
 *     value: string,
 * }
 *
 * @psalm-type JmapEntityListSort = array{
 *     attribute: 'from'|'to'|'subject'|'received'|'sent'|'tag',
 *     direction: string,
 * }
 */
```

lib/Service/JMAP/JmapOperationsService.php
```diff
@@ -0,0 +1,1024 @@
...
+	 * @param array<array{attribute:string,value:mixed}>|null $filter optional filter conditions
+	 * @param array<array{attribute:string,direction:string}>|null $sort optional sort conditions
+	 *
+	 * @return Mailbox[]
+	 */
+	public function collectionList(?string $location = null, ?array $filter = null, ?array $sort = null): array {
```
kesselb: $filter and $sort as array to simplify the api

lib/Service/JMAP/JmapOperationsService.php
```diff
@@ -0,0 +1,1024 @@
...
+	 *
+	 * @param string|null $location optional location constraint
+	 * @param array|null $filter optional filter conditions
+	 * @param array|null $sort optional sort conditions
+	 * @param array|null $range optional range conditions
+	 * @param string|null $granularity optional granularity level
+	 *
+	 * @return array{state:string, list:array<string, Message>}
+	 */
+	public function entityList(?string $location = null, ?array $filter = null, ?array $sort = null, ?array $range = null, ?string $granularity = null): array {
```
kesselb: filter & sort should be an array.
range ok, please also add a shape so the expected format is documented.
granularity string, please also add a shape and document the expected values (e.g. basic and full). entityDelta calls entityList with "B" as granularity. B = Basic? With proper typing psalm can flag.

lib/Service/JMAP/JmapOperationsService.php
```diff
@@ -0,0 +1,1024 @@
...
+			// evaluate if client was already connected
+			if (!$this->dataStore->sessionStatus()) {
+				$this->dataStore->connect();
+			}
+		} catch (Exception $e) {
+			throw new ServiceException('Could not connect to JMAP server: ' . $e->getMessage(), 0, $e);
+		}
+		// determine account
+		if ($this->dataAccount === null) {
+			$account = $this->dataStore->sessionAccountDefault('mail');
```
kesselb: Please pick a different variable name. Re-assigning account is not nice.

lib/Service/JMAP/JmapOperationsService.php
```diff
@@ -0,0 +1,1024 @@
...
+		try {
+			// evaluate if client was already connected
+			if (!$this->dataStore->sessionStatus()) {
+				$this->dataStore->connect();
+			}
+		} catch (Exception $e) {
+			throw new ServiceException('Could not connect to JMAP server: ' . $e->getMessage(), 0, $e);
+		}
+		// determine account
+		if ($this->dataAccount === null) {
```
kesselb: Does that work as expected when looping over accounts like in MailManager.deleteTag?

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: yes

lib/BackgroundJob/RepairSyncJob.php
SebastianKrupinski: Done

lib/IMAP/ImapMessageConnector.php
SebastianKrupinski: The idea here was to generate a sync token in both jmap and imap the same way.

lib/JMAP/JmapMessageConnector.php
SebastianKrupinski: done

lib/Protocol/ProtocolFactory.php
SebastianKrupinski: done

lib/Protocol/ProtocolFactory.php
SebastianKrupinski: done

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: You're right that would have failed. I borrowed all the code from another implementation i did and didn't think twice about what happens on reuse.

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: Sure

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: done

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: Sure... not sure how this simplifies the api... but i've changed it

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: done

SebastianKrupinski:
@kesselb

I have addressed / changed all requested via comments... will clean up CI once code is approved.

lib/Service/JMAP/JmapOperationsService.php
kesselb: I was suggesting to use Throwable instead of Exception

lib/Service/JMAP/JmapOperationsService.php
kesselb: Thx, ?array gives you two ways to say "no filter/sort"

kesselb:
I moved the psalm-types to the class level docblock, otherwise the import from other classes wont work.

lib/Service/JMAP/JmapOperationsService.php
SebastianKrupinski: Sure, done

lib/Service/AiIntegrations/AiIntegrationsService.php
```diff
@@ -361,6 +337,64 @@ public function requiresFollowUp(
 		return preg_match('/{\s*"expectsReply"\s*:\s*true\s*}/i', $output) === 1;
 	}
 
+	/**
+	 * Analyze whether an email is written in a specific language.
+	 *
+	 * @throws ServiceException
+	 */
+	public function requiresTranslation(
```
kesselb: requiresTranslation was removed on main by https://github.com/nextcloud/mail/pull/13501/changes#diff-8acc94347e8ed2d2e9cc612df74764cffac805832d6ad430058814d837e2df87

SebastianKrupinski:
> I moved the psalm-types to the class level docblock, otherwise the import from other classes wont work.

Saw that thank you. I address the last comment you made

lib/IMAP/ImapMessageConnector.php
```diff
@@ -0,0 +1,452 @@
...
+		$client = $this->protocolFactory->imapClient($account);
+
+		$mutatedMessages = [];
+		foreach ($messages as $message) {
+			try {
+				$newUid = $this->imapMessageMapper->move($client, $sourceMailbox->getName(), $message->getUid(), $targetMailbox->getName());
+				if ($newUid === null) {
+					// The IMAP server does not support UIDPLUS and the message has no Message-ID
+					// header, so the new UID is unknown. It will be reconciled on the next sync.
+					$this->logger->warning('Moved message but could not determine its new UID', [
```
kesselb: This adds a **new** warning about a state that is handled. The admin cannot do anything to resolve it and it does not require action => this should not be logged as warning.

lib/Service/MailManager.php
```diff
@@ -94,213 +72,207 @@ public function getMailbox(string $uid, int $id): Mailbox {
...
+			->messageConnector($account)
+			->moveMessages($account, $targetMailbox, $sourceMailbox, ...$messages);
 
-		$client = $this->imapClientFactory->getClient($account);
-		try {
-			$this->deleteMessageWithClient($account, $sourceMailbox, $messageUid, $client);
-		} finally {
-			$client->logout();
+		// update local store
+		$this->dbMessageMapper->updateBulk($account, false, ...$mutatedMessages);
```
kesselb: One would assume MessageMapper::updateBulk() would update entities in one go, but that's apparently not the case ;)

I'd agree it would be nice to have more batch methods (like moveMessages), but I'd see that as its own feature, and would really appreciate keeping this PR focused on the necessary changes. So far I don't see why JMAP needs a moveMessages method.

The updateBulk() call here is dead code because it never writes mailbox_id/uid, so it can't do what these two call sites need it to do.

IMAP has two paths calling deleteByUid:

1. lib/Cache/Cache.php deleteMsgs
2. MessageCacheUpdaterListener, in response to a MessageDeletedEvent

moveMessages and moveMessage no longer emit MessageDeletedEvent. That's a regression and should be resolved.

I'd prefer to keep moveMessages out. If the method has to stay, it must emit MessageDeletedEvent.

lib/Service/AiIntegrations/AiIntegrationsService.php
SebastianKrupinski: Is gone now... that was just a rebase glitch... its really difficult to keep this PR clean with the weekly conflicts resolutions

lib/IMAP/ImapMessageConnector.php
SebastianKrupinski: Done changed it to debug

lib/Service/MailManager.php
SebastianKrupinski: > So far I don't see why JMAP needs a moveMessages method.

JMAP operates in bulk by design a single request moves N messages, there's no per-message round trip. Restricting this PR to single-message moves would mean adding a batch capability that the protocol already gives us for free, only to bolt it back on in a follow-up. All the connector interfaces here already operate on Message ...$messages for the other operations (flag, tag, delete); moveMessages being singular would be the outlier, not the norm. Also I don't understand the lets fix it later approach that we never end up doing... since the mail manager had to be re-write to be protocol agnostic... why not do it correctly from the start, why handicap the new protocol with the bad logic from the old one.

> The updateBulk() call here is dead code because it never writes mailbox_id/uid,

Ugh... another thing that makes no sense... why are we matching the rows by mailbox_id and uid when there is a perfectly fine record id...

Okay, I've changed the code to push the changes row by row. This should get refactored into a real bulk update-by-id later, since updateBulk() as written can't do it — it never writes mailbox_id/uid at all.

>  That's a regression and should be resolved.

Fixed

SebastianKrupinski:
@kesselb

I have addressed / changed all requested in your most recent comments

ChristophWurst:
@DerDreschner since you tested this positively, could you please review and approve? Thanks

ChristophWurst (commented):
As far as I can see you removed the explicit try-finally handling of IMAP connections as resource.
There is no replacement mechanism to manage them.

Doesn't this mean that we might end up with a pile of open connections by the end of a cron execution but possibly also during initial sync when we need to hop between caching and non-caching clients?

As a reminder, there is a common per-IP connection limit on mail servers. IIRC it was *five connections per IP* for Dovecot. That would bite us rather quickly.

SebastianKrupinski:
> As far as I can see you removed the explicit try-finally handling of IMAP connections as resource. There is no replacement mechanism to manage them.
> 
> Doesn't this mean that we might end up with a pile of open connections by the end of a cron execution but possibly also during initial sync when we need to hop between caching and non-caching clients?

The last commit addresses the open connections
 
> As a reminder, there is a common per-IP connection limit on mail servers. IIRC it was _five connections per IP_ for Dovecot. That would bite us rather quickly.

As for this connection reuse/caching is on my list of after this is merged follow ups,

kesselb:
Findings from our co-working session:

Bugs:

  1. Moving or deleting a thread acts on the wrong messages (MailManager.php
     fetchThread/moveThread/deleteThread, around lines 389-452)
     - fetchThread collects messages from every mailbox in the thread.
     - moveMessages/deleteMessages then run all of them against the single $srcMailbox name
       (ImapMessageConnector.php:192,237).
     - UIDs from Sent get applied inside INBOX, so unrelated messages are moved or deleted.
     - main used each row's own mailboxName.
  2. Tags get wiped (MailManager.php:345,366 → Db/MessageMapper::updateTags line 541)
     - updateBulk($account, true, …) makes the DB tags match $message->getTags().
     - findByUids and findByMessageId never load tags, so every existing tag gets removed.
       This hits three callers:
       - markFolderAsRead (line 374) strips the tags of the whole folder.
       - FollowUpClassifierJob:58 → :107.
       - deleteTag (line 594) → :613.
     - Only getMessage → findByUserId loads tags, through findRelatedData.


Ideas:

- Add a markFolderAsRead as abstraction. The current impl for imap and jmap selects all uids for a mailbox. On imap we can just flag as whole folder as read without needing the uids.

lib/Service/MailManager.php
```diff
@@ -310,403 +292,223 @@ public function deleteMessageWithClient(
...
-			} catch (Horde_Imap_Client_Exception $e) {
-				throw new ServiceException(
-					'Could not set message keyword on IMAP: ' . $e->getMessage(),
-					$e->getCode(),
-					$e
-				);
-			}
+	public function markFolderAsRead(Account $account, Mailbox $mailbox): void {
+		// find all messages in mailbox with their remote ids
+		$messages = $this->dbMessageMapper->findByUids($mailbox, $this->dbMessageMapper->findAllUids($mailbox));
```
kesselb, line 374: Could we fetch all messages with seen = false from the db, to avoid processing message with that state already?

lib/IMAP/ImapMessageConnector.php
```diff
@@ -0,0 +1,465 @@
...
+	public function fetchMessages(Account $account, Mailbox $mailbox, bool $loadBody = false, Message ...$messages): array {
+		$client = $this->protocolFactory->imapClient($account);
+		$uids = array_map(static fn ($message) => $message->getUid(), $messages);
+		try {
+			return $this->imapMessageMapper->findByIds(
+				$client,
+				$mailbox->getName(),
+				$uids,
+				$account->getUserId(),
+				true
```
kesselb, line 95: https://github.com/nextcloud/mail/blob/81ed4b031e378fff7ece5ce293f0a3e751203dd7/lib/IMAP/MessageMapper.php#L73

SebastianKrupinski:
@kesselb 

> 1. Moving or deleting a thread acts on the wrong messages (MailManager.php

Fixed

> 2. Tags get wiped (MailManager.php:345,366 → Db/MessageMapper::updateTags line 541)

Fixed

> 3. Phishing check dropped (ImapMessageConnector.php:90-96)

Fixed

PR #12803, opened 2026-04-24
fix(AccountService): Invalidate cache after account creation/change/deletion

DerDreschner:
While working on #12777, I found out that the cache isn't being invalidated after any change. This **may** cause some issues, although those will be very rare as the lifecycle of the `AccountService` class is limited to each request. For performance reasons, it may be worth to move the cache to `memcache`, which then will survive the lifecycle of the `AccountService` class.

AI-assisted: Claude Sonnet 4.6


ChristophWurst (approved):
Makes sense!

PR #12182, opened 2025-12-17
fix(ui): add ckeditor body element to editor focus trap

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/12178

1. Restore the undefined `additionalTrapElements` data (regression of https://github.com/nextcloud/mail/pull/12118)
2. Push the ckeditor body element to the trap elements once it's ready

This makes it possible to inserts links with CKEditor again, while also avoiding the focus trap issues on mobile.

## How to test link insertion

1. Open the app
2. Click *New message*
3. Enable formatting
4. Insert link via CKEditor

main: you can't focus link text or URL
here: you can focus the inputs and type a value

## How to test mobile editing

1. Open the app on phone or desktop responsive view
2. Click *New message*
3. Click into the *To* field

main: multiselect opens and closes instantly. You can't enter anything
here: multiselect opens and you can enter a value

PR #12809, opened 2026-04-24
https://github.com/nextcloud/mail/pull/12810

kesselb:
STR:

- Open composer
- Enable formatting
- Try inserting a link or use find an replace

Main: You cannot focus the inputs
Here: You can

AI-assisted: Claude Code (Claude Sonnet 4.6)


ChristophWurst:
What about https://github.com/nextcloud/mail/pull/12182?

kesselb:
> What about #12182?

Ha, that explains why I'd looked so familiar :laughing:

PR #12863, opened 2026-05-04
fix: Adjust relative date display to also use timestamp for yesterday

### LINKED ISSUE #1095: Display mail date and time instead of age in mailbox
### Feature Request

In the mailbox view, mails are sorted by age and their age is shown next to the sender name.

Currently, all mails have their age shown, which might make sense for mails that are minutes or hours old, or even up to a day. But after this, the mail age makes emails almost impossible to find. Mails that are "one month ago" can be hundreds of emails. An email that is 21 hours old requires me to do math instead of just reading the timestamp.

Suggestion:
Mark emails with:
Today's date: "Today, [hh:mm]"
Yesterday's date: "Yesterday, [hh:mm]"
Older: "[YYYY-MM-DD] [hh:mm]" (or other corresponding configurable local date format)

I will monitor this feature request, and if I can get some hints on where to look for this feature in the code, I can consider registering for contributing by making the appropriate code changes. I haven't contributed to Nextcloud Mail before but I wouldn't mind doing.

#### Summary
Display mail date and time instead of age for older emails.

jancborchardt:
Fixes https://github.com/nextcloud/mail/issues/1095

### Message/thread list (center column)
* Emails in *Yesterday* show the time instead of the day abbreviation
* In PI and unified inbox emails of yesterday shows as *Yesterday 9:27*
* Time for today and yesterday now respects user locale (am/pm was not supported before)

### Thread view (right column)
* Time for emails of today respects user locale
* Emails from *Yesterday* show *Yesterday 9:27* instead of date without time
* Date+time for older emails respects user locale instead of date without time

## Old description (outdated)

<details>

The inbox and folders in the Mail app have date headers like "Today", "Yesterday", "Last week", "Last month" etc. Currently we use a shortened weekday for mails in the "Yesterday" section, but instead it would be better to use the timestamp, same as is done for mails in the "Today" section.

Extra challenge: In the detail view it should show "Yesterday [hh:mm]" because otherwise it’s not clear it’s from yesterday, so the function needed to have a parameter added.

This could be good to go, and an additional fix as follow-up could be:
> For mails older than a year, the sender and date can overlap in the message list. At the same time the width of list-item-content__inner__subname which contains subject and some body text preview is unnecessarily made narrow by the very wide list-item-content__inner__subname, which is only on the line of the sender. It is only the sender name which should be ellipsized due to longer dates, but the subject or body text preview should not be shortened unnecessarily.
> The date should be aligned to the right and be shown fully, and the name should take the rest of available space. The name should ellipsize on overflow.

I tried that but couldn’t get the components to comply. :D 

---

### AI-assisted: GitHub Copilot (Claude Sonnet 4.6)

**Prompt:**
The inbox and folders in the Mail app have date headers like "Today", "Yesterday", "Last week", "Last month" etc. Currently we use a shortened weekday for mails in the "Yesterday" section, but instead it would be better to use the timestamp, same as is done for mails in the "Today" section.

**Iteration 1**
2 issues:
1. For mails which are in "Last week" but less than 48 hours ago, so not yesterday, it still shows the timestamp in the list.
2. For mails from yesterday, in the detail view it only shows the timestamp, not "Yesterday" or any other classifier.

**Iteration 2**
Now it shows "Yesterday [timestamp]" both in the detail view as well as in the envelope list. It should only show this long version in the detail view, keeping only the timestamp in the envelope list, because there is a "Yesterday" heading already.

</details>


src/util/shortRelativeDatetime.js
```diff
@@ -5,15 +5,23 @@
 
 import moment from '@nextcloud/moment'
 import curry from 'lodash/fp/curry.js'
+import { translate as t } from '@nextcloud/l10n'
 
-export const shortDatetime = curry((ref, date) => {
+export const shortDatetime = curry((ref, date, withLabel = false) => {
```
kesselb: The withLabel flag has only an effect when the date is yesterday. 

I think this should be done in the component itself, or if required in more places a own function. 

The purpose of the shortDatetime function should stay with shorting a date ;)

ChristophWurst:
1. Fixed kesselb's comment
2. Fixed that not all emails are in a "date group". In PI we have "Favorites/Important" and they will use the long format
3. Fixed that the short format did not use the user's locale setting. E.g. message list showed *13:38* and actions menu *1:38pm*. Now it's consistent.

src/util/shortRelativeDatetime.js
```diff
@@ -26,8 +33,17 @@ export const shortDatetime = curry((ref, date) => {
 	return momentDate.format('MMM D, YYYY')
 })
 
+export const longDatetime = curry((ref, date) => {
+	// Yesterday?
+	if (date < startOfDay(ref) && date >= startOfPreviousDay(ref)) {
+		return t('mail', 'Yesterday') + ' ' + moment(date).format('LT')
```
ChristophWurst: I *think* we have to use one single translation string with a placeholder so that translators can move the position of the time around

ChristophWurst: Pushed

ChristophWurst:
In the right column we now use time for today, yesterday + time for yesterday and date + time for anything older. That makes it very easy to get precise info without any hover.

I think this is good to go

ChristophWurst:
PR description updated to summarize all the changes in this PR

ChristophWurst (approved):

PR #12874, opened 2026-05-05
refactor: add null guards in db mappers and fix JsonResponse return type for Psalm 6

ChristophWurst:
## Summary

- `LocalMessageMapper`: add null guard for `getLocalMessageId()` before using as array key in `getAllForUser`, `findDue`, `findDueDrafts`; use `?? []` for `getRecipients()` in `updateWithRecipients`
- `MessageMapper`: extract `getMessageId()` to a local variable with null guard before using as array key in `updateTags` and `findRelatedData`
- `JsonResponse`: align `@return static` → `@return self` on `errorFromThrowable` to match the actual PHP return type hint

These are Psalm 6 `PossiblyNullArrayOffset` and `MoreSpecificReturnType` errors that remain after the Psalm v6 upgrade in #11224.

## Test plan

- [x] PHP unit tests pass (`composer test:unit`)
- [x] Psalm static analysis passes (`composer psalm`)

Part of #11224

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -545,7 +545,8 @@ class MessageMapper extends QBMapper {
 	 */
 	private function updateTags(Account $account, Message $message, array $tags, PerformanceLoggerTask $perf): void {
 		$imapTags = $message->getTags();
-		$dbTags = $tags[$message->getMessageId()] ?? [];
+		$messageId = $message->getMessageId();
+		$dbTags = $messageId !== null ? ($tags[$messageId] ?? []) : [];
 
 		if ($imapTags === [] && $dbTags === []) {
 			// neither old nor new tags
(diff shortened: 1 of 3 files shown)


ChristophWurst:
@coderabbitai review

lib/Db/MessageMapper.php
```diff
@@ -531,7 +531,8 @@ public function updateBulk(Account $account, bool $permflagsEnabled, Message ...
 	 */
 	private function updateTags(Account $account, Message $message, array $tags, PerformanceLoggerTask $perf): void {
 		$imapTags = $message->getTags();
-		$dbTags = $tags[$message->getMessageId()] ?? [];
+		$messageId = $message->getMessageId();
```
kesselb, line 548: if $messageId === null then return?

kesselb (approved):

PR #12900, opened 2026-05-07
feat: add select-all checkbox and lift selection state to parent

RobinAngele:
Depends on: #12899

Refs #4285
Refs #7880
Refs #6070
Refs #7276
Refs #11526

Follow-up: #12901 adds filter-based mass selection

### Summary

Adds a "Select all X messages" checkbox above the envelope list and lifts
envelope selection state from individual `EnvelopeList` instances to the
`Mailbox` parent, enabling cross-group shift-click range selection and
global select-all/unselect-all.

Together with #12901 this fully closes #4285 and #7880.

### Changes by file

**`Mailbox.vue`**
- Select-all checkbox using `NcCheckboxRadioSwitch` at the top of the list
- Shows "Select all N messages" with the current visible page count
- Hidden once selection mode is active (multiselect header takes over)
- `margin-top: 8px` so it doesn't sit flush against the sticky search bar
- `selection` array in data — single source of truth for selection state
- `selectAll()`, `unselectAll()` methods for global control
- `flatEnvelopeList`, `selectMode`, `allSelected` computed properties
- `onUpdateSelection()`, `onSelectRange()`, `getGroupFlatIndex()` methods
- Passes `selection` prop down to `EnvelopeList` children
- Listens for `update:selection` and `select-range` events from children
- Resets selection on mailbox or filter change
- `.select-all-bar` CSS

**`EnvelopeList.vue`**
- `selection` prop (type `Array`, replaces local `data().selection`)
- `flatIndex` prop (type `Number`, for cross-group shift-click indexing)
- Watch on `selection` prop with `immediate: true` to sync `flags.selected`
- `emitLocalSelection()` method — emits `update:selection` to parent
- Shift-click now emits `select-range` with global flat indices
- Deleted local `selection` from `data()` and `setEnvelopeSelected` mutations

### How to test

1. Open a mailbox → "Select all N messages" checkbox visible at top
2. Click it → all visible messages selected → multiselect header appears
3. Shift-click across messages in different date groups → range selected
4. Change mailbox or filter → checkbox resets to unchecked
5. Unselect all → checkbox reappears

### Screenshots

Select All - Screenshot_20260507_061040

Issue #11973, opened 2025-10-28
Selecting threads should work across date groups


kesselb:
### Steps to reproduce

1. Open mail
2. Select an email from the "last week" and "last month"

### Expected behavior

The action bar should be shown above the groups and only once. The action should be executed on all messages, independent of the grouping. 

### Actual behavior

There's one action bar for each group

<img width="445" height="879" alt="Image" src="https://github.com/user-attachments/assets/ce48213c-ab0b-4a78-9e79-ed1937cbfcc3" />

Issue #11526, opened 2025-08-19
Multiple multi actions headers (one for each date group)


SebastianKrupinski:
### Steps to reproduce

1. Open a mailbox folder with messages for a wide range of dates
2. Multi select dates in 2 date groups
3. See the multiple multi action menus

### Expected behavior

There should only be one multi action menu at the top of the list.

### Actual behavior

Multiple multi action menus one for each date group

### PHP engine version

None

### Web server

None

### Database

None

Issue #11420, opened 2025-07-26
support bulk archive


asandikci:
### Is your feature request related to a problem? Please describe.

There is no bulk action to archive multiple mail

### Describe the solution you'd like

There should be a bulk action for archiving multiple mail

ChristophWurst:
Thanks for the request. This makes sense!

PR #12901, opened 2026-05-07
feat: add filter-based mass selection via search modal

### LINKED ISSUE #4285: Select all messages in folder
### Feature Request

Provide an option to select all mail in the current folder.

#### Summary

Currently, it is possible to select multiple messages one-by-one. Most e-mail clients have an option to select all e-mail e.g. for (re-)moving them. Please add this option to Nextcloud / Mail as well.

### LINKED ISSUE #7880: Implement a "Select All" menu item

I can't select all items in a folder/mailbox to apply the same action to them, typîcally move or delete.

After selecting one element in the messages list, the "Actions on selection" 3-dot menu at the top could benefit from a "Select all" entry, so you can then apply any action on all the emails you have.

There no keybaord  shortcut to enable selection like the "S" does for starring.

### Additional context

It'a quick win for multiple mails handling.  #7276 has more in-depth desires which would be handy.

### LINKED ISSUE #11527: Multi select is broken across date groups
### Steps to reproduce

1. Open a mailbox folder with messages for a wide range of dates
2. Select message in first date group
3. Press shift then select message in another date group

### Expected behavior

All messages in between first selected message and second selected message should be selected.

### Actual behavior

Messages are not selected

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

RobinAngele:
Depends on: #12900

Closes #4285
Closes #7880
Closes #11527
Refs #6070
Refs #7276
Refs #11526
Refs #11973
Refs #11420

## Summary

Built on #12900 which centralised selection state in `Mailbox`, this PR completes #4285 and #7880 by adding three complementary ways to select messages for bulk actions:

1. **Checkbox row** — click the checkbox at the top of the list to select all currently loaded messages at once.

2. **Select-all banner** — after clicking the checkbox to select all loaded messages, if more pages remain a banner appears. Clicking it loads all remaining pages and selects every message in the current view.

3. **Search modal "Select all matching"** (new primary button) — open the search dialog, set any filter combination, and click "Select all matching". The modal closes, all matching messages are loaded across all pages, and the entire result set is selected ready for bulk action.

In all three flows, selection is capped at 500 messages to prevent OOM in unified mailboxes (which fan out to N IMAP accounts per page fetch). When the cap is reached a warning explains it and suggests narrowing the filter or working in batches.

Also fixes **shift-click range selection across date groups** (#11527): shift-clicking a message in a different date group from the last-selected one now correctly selects all messages in between.

Also adds full **Priority inbox support**: each of the four sections (Favorites, Follow-up, Important, Other) gets its own independent select-all flow. Selecting messages in one section automatically clears the selection in any other, since each section has its own action toolbar and cross-section bulk operations are not possible.

### Search modal — new "Select all matching" primary button
Screenshot_20260507_114235_select_feat_Search parameters_ovelay

### Mass-select result — N selected with action toolbar
Screenshot_20260510_102704_selft_filter_with_result_background

### Priority inbox mutual exclusion — section cleared toast
Screenshot_20260510_103559_toast_priority_mailbox

### Component interaction

```
User clicks "Select all matching" in search modal
        │
        ▼
SearchMessages.vue
  emits 'select-all-matching' (query string)
        │
        ▼
MailboxThread.vue
  sets this.searchQuery = query          ← must happen before bus emit
  emits bus('select-all-matching')       ← broadcast to all Mailbox instances
        │
        ├─────────────────────────────────────────────┐
        ▼                                             ▼
Mailbox.vue × 2 (regular / Unified inbox) Mailbox.vue × 4 (Priority inbox only)
  Favorites  (isPriorityInbox=true)     Favorites, Follow-up, Important, Other
  Main       (isPriorityInbox=false)    all have isPriorityInbox=true
  share one mitt bus                    share one mitt bus
  Main owns the select-all banner       each fetches its own filtered pages
  each fetches its own filtered pages   mutual exclusion via 'section-selected'
        │                                             │
        ▼                                             ▼
EnvelopeList.vue (per section)          EnvelopeList.vue (per section)
  receives selection[] as prop            receives selection[] as prop
  emits update:selection (delta)          emits update:selection / select-range
  emits select-range (flat indices)
```

State flows **down** via props, changes flow **up** via Vue events, and cross-instance coordination uses the shared **mitt bus**.

---

### Key design decisions

Every `MailboxThread` view mounts at least two `Mailbox` instances on the same bus — a Favorites section and a Main section. The Priority inbox (`PRIORITY_INBOX_ID`) extends this to four sections (Favorites, Follow-up, Important, Other), all with `isPriorityInbox=true`. The unified inbox (`UNIFIED_INBOX_ID`) uses the same two-section layout as a regular folder and requires no special handling. The decisions below address the non-obvious coordination problems the multi-instance shared-bus pattern creates.

**1. Capturing the settled query before loading**
Priority inbox sections inject a query suffix one Vue render tick _after_ the parent's `searchQuery` prop changes. Starting mass-select immediately would use a stale query and corrupt the cache key. The fix is to raise a lock flag (`loadingAllMatching`) before yielding — which simultaneously blocks the `searchQuery` watcher from triggering a competing reload — then read the settled, section-filtered query after the tick.

**2. Mutual exclusion between sections**
Each section has its own independent action toolbar, so a selection spanning two sections can't be acted upon in a single bulk operation. When a section's selection goes from empty to non-empty (user-initiated), it broadcasts a `section-selected` event on the shared bus. Any sibling holding a selection clears itself with an info toast.

**3. Concurrent mass-selects must not interfere**
The search modal triggers mass-select on all sections simultaneously — each independently loads its own filtered pages. Mutual exclusion must _not_ fire on completion: each finishing section would otherwise clear the others' just-loaded selections. So `section-selected` is emitted only on user-initiated transitions, never at the end of a programmatic mass-load.

**4. OOM protection: 500-message cap**
The unified inbox fans out to N IMAP accounts per page fetch. Without a ceiling, large multi-account selections exhaust client memory. The loop halts at 500 and shows a visible warning rather than silently truncating.

**5. Keeping the spinner visible during mass-select**
Mass-select forces a cache invalidation to guarantee a fresh fetch. Without a guard, this triggers `LoadingSkeleton`, hiding the spinner mid-operation. A `!loadingAllMatching` condition in `loadEnvelopes()` suppresses the skeleton so the spinner stays visible throughout.

### `SearchMessages.vue`
- Added **"Select all matching"** as a primary `NcDialog` button; **"Search"** demoted to tertiary
- `selectAllMatching()` sets `match = 'allof'`, closes the modal, then emits `select-all-matching` with the full computed query string — intentionally **no `sendQueryEvent()`** to avoid a concurrent reload
- `closeSearchModal()` has an idempotency guard — safe to call multiple times
- Fixed `filterData.mentions`: coerces `false` to `''` so `mentions:false` is never appended to the IMAP query string

### `MailboxThread.vue`
- Wires `@select-all-matching="onSelectAllMatching"` on `SearchMessages`
- `onSelectAllMatching(query)` sets `this.searchQuery = query` **before** `this.bus.emit('select-all-matching')` so each Priority inbox section reads the correct parent query when it captures `effectiveQuery` after its own `$nextTick`

### `Mailbox.vue`
- `isPriorityInbox` prop — hides the select-all banner, disables envelope grouping, and shows an empty-section placeholder (mutual exclusion via `section-selected` fires on all instances regardless of this flag)
- `loadingAllMatching` / `selectAllMatching` / `selectionLimitReached` data flags
- `onBusSelectAllMatching()` — full mass-select lifecycle: raises flag, yields tick, clears cache, fetches all pages, caps at 500, selects all; both failure paths (page-load error mid-loop and outer catch) now show an error toast and reset `selectAllMatching` (previously silent — console-only logging)
- `loadEnvelopes()` — `!loadingAllMatching` guard suppresses `LoadingSkeleton` during mass-select
- `selection` watcher — emits `section-selected` on 0 → non-zero transition, guarded by `!loadingAllMatching`
- `onBusSectionSelected(activeQuery)` — clears sibling section selection with info toast ("Selection cleared — only one Priority inbox section can be selected at a time"); self-guard via `activeQuery !== this.searchQuery`
- `selectAllMatchingAction()` — in-page banner entry point (calls `onBusSelectAllMatching()` directly, no bus round-trip)
- `selectAllLabel` computed — context-aware label:

| Condition | Label |
|-----------|-------|
| `loadingAllMatching` | Selecting messages… |
| `selectMode` | N selected |
| `hasFilter` | Select N matching messages |
| `!endReached && count > 0` | Select N loaded messages |
| _(default)_ | Select all N messages |

- `select-all-banner` — prompts to load all matching when all visible are selected and more exist; **hidden for Priority inbox**; button label adapts: "Select all matching messages" when a filter is active, "Select all messages in this folder" otherwise
- `select-all-limit-warning` — shown in warning colour when 500-cap is hit: "Selection limited to N messages — use a filter to narrow results, or process in several batches."
- `onUpdateSelection()` — merged into single array assignment (removes intermediate empty reactive state)
- `unselectAll()` — also clears `loadingAllMatching` and `expanded`
- `bus.on/off('section-selected', ...)` symmetric in `created`/`unmounted`
- Imports `priorityImportantQuery`, `priorityOtherQuery`, and `favoritesQuery` constants instead of hardcoding strings in `hasFilter`

### `EnvelopeList.vue`
- `_localToggleInProgress` guard on both `selection` and `sortedEnvelops` watchers — prevents watchers from overwriting `flags.selected` during a local click; reset via `$nextTick` to survive Vue's watcher flush

## How to test

- [ ] Open any folder → click the checkbox row → all visible messages selected, action toolbar appears
- [ ] Open a large folder (more than one page of messages) → scroll down to load a second page → without selecting anything, verify the checkbox-row label updates to "Select N loaded messages" where N reflects only the messages loaded so far, not the full folder total
- [ ] Select all visible in a large folder (no filter) → **banner appears** → click it → all pages load and select
- [ ] Open the search modal → set a filter → click **"Search"** (tertiary button, not "Select all matching") → filter results load → select all loaded via checkbox → banner appears with filter wording → click it → all matching messages across all pages selected
- [ ] Open the search modal → set any filter → click **"Select all matching"** (primary button) → modal closes, spinner appears, all matching messages across all pages are selected directly (no banner step)
- [ ] While messages are selected, click individual avatar circles to deselect/reselect one at a time → count updates correctly, no spurious multi-deselect
- [ ] Shift-click across messages in different date groups → range selected correctly
- [ ] **Priority inbox** → open Priority inbox → click avatar circles in the **Important** section to select one or more messages → then click an avatar circle in the **Other** section → verify the Important selection clears automatically and a toast appears: "Selection cleared — only one Priority inbox section can be selected at a time"
- [ ] **Priority inbox** → select in Important → click individual avatars to add/remove messages within the same section → sibling sections are not affected
- [ ] Open a folder with > 500 messages → trigger mass-select via the banner or search modal → selection stops at 500 and shows the cap warning in orange with actionable advice
- [ ] Open DevTools → Network tab → start a mass-select on a large folder → set network to Offline mid-load → spinner disappears, error toast appears, no messages remain selected
- [ ] Bulk actions (delete, mark read, move, tag) apply correctly to all selected messages

## Known limitations

- **No cross-section bulk actions in Priority inbox.** Selecting in Important and then clicking in Other clears the first section. There is no toolbar that spans sections, so a cross-section selection cannot be acted upon.
- **500-message cap.** Raising it requires backend pagination performance improvements first.
- **No server-side "select all" flag.** Bulk operations iterate over the `selection` array and fire individual API calls. A future improvement would be a server-side endpoint accepting a filter query and performing the operation without enumerating IDs client-side

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -168,7 +168,7 @@
 import { showError } from '@nextcloud/dialogs'
 import { NcActionButton as ActionButton, NcActions as Actions, NcButton, NcDialog } from '@nextcloud/vue'
 import { mapStores } from 'pinia'
-import { differenceWith } from 'ramda'
+
 import AlertOctagonIcon from 'vue-material-design-icons/AlertOctagonOutline.vue'
 import IconSelect from 'vue-material-design-icons/CloseThick.vue'
 import EmailRead from 'vue-material-design-icons/EmailOpenOutline.vue'
@@ -266,11 +266,20 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+
+		selection: {
+			type: Array,
+			default: () => [],
+		},
+
+		flatIndex: {
+			type: Number,
+			default: 0,
+		},
 	},
 
 	data() {
 		return {
-			selection: [],
 			showMoveModal: false,
 			showTagModal: false,
 			lastToggledIndex: undefined,
@@ -362,14 +371,32 @@ export default {
 	},
 
 	watch: {
-		sortedEnvelops(newVal, oldVal) {
-			// Unselect vanished envelopes
-			const newIds = newVal.map((env) => env.databaseId)
-			this.selection = this.selection.filter((id) => newIds.includes(id))
-			differenceWith((a, b) => a.databaseId === b.databaseId, oldVal, newVal)
-				.forEach((env) => {
-					env.flags.selected = false
+		selection: {
+			handler(newSelection) {
+				if (this._localToggleInProgress) {
+					return
+				}
+				// Sync flags.selected when the selection changes from outside this
+				// instance (e.g. shift-click across groups, Select All / Unselect All).
+				const selectionSet = new Set(newSelection)
+				this.sortedEnvelops.forEach((env) => {
+					env.flags.selected = selectionSet.has(env.databaseId)
 				})
+			},
+			immediate: true,
+		},
+
+		sortedEnvelops(newVal, oldVal) {
+			// Skip if a local toggle is in progress to avoid race conditions
+			if (this._localToggleInProgress) {
+				return
+			}
+			// Unselect vanished envelopes by emitting cleaned selection
+			const newIds = new Set(newVal.map((env) => env.databaseId))
+			const cleanedSelection = this.selection.filter((id) => newIds.has(id))
+			if (cleanedSelection.length !== this.selection.length) {
+				this.$emit('update:selection', cleanedSelection, this.envelopes)
+			}
 		},
 	},
 
@@ -534,20 +561,24 @@ export default {
 			this.unselectAll()
 		},
 
-		setEnvelopeSelected(envelope, selected) {
-			const alreadySelected = this.selection.includes(envelope.databaseId)
-			if (selected && !alreadySelected) {
-				envelope.flags.selected = true
-				this.selection.push(envelope.databaseId)
-			} else if (!selected && alreadySelected) {
-				envelope.flags.selected = false
-				this.selection.splice(this.selection.indexOf(envelope.databaseId), 1)
-			}
-		},
-
 		onEnvelopeSelectToggle(envelope, index, selected) {
 			this.lastToggledIndex = index
-			this.setEnvelopeSelected(envelope, selected)
+			this._localToggleInProgress = true
+			// Compute new local selection from the authoritative prop, applying this
+			// single toggle. flags.selected is unreliable here: the selection watcher
+			// may have been skipped (while _localToggleInProgress was true) leaving
+			// flags stale, and Pinia store updates can replace the flags object
+			// entirely, wiping selected=true during a background sync.
+			const myIds = new Set(this.sortedEnvelops.map((e) => e.databaseId))
+			const currentLocal = this.selection.filter((id) => myIds.has(id))
+			const newLocal = selected
+				? [...new Set([...currentLocal, envelope.databaseId])]
+				: currentLocal.filter((id) => id !== envelope.databaseId)
+			this.$emit('update:selection', newLocal, this.envelopes)
+			// Reset after next tick — Vue batches watchers at end of tick
+			this.$nextTick(() => {
+				this._localToggleInProgress = false
+			})
 		},
 
 		onEnvelopeSelectMultiple(envelope, index) {
@@ -558,12 +589,12 @@ export default {
 				return
 			}
 
-			const start = Math.min(lastToggledIndex, index)
-			const end = Math.max(lastToggledIndex, index)
-			const selected = this.selection.includes(envelope.databaseId)
-			for (let i = start; i <= end; i++) {
-				this.setEnvelopeSelected(this.sortedEnvelops[i], !selected)
-			}
+			// Convert to global flat indices and delegate to the parent
+			const globalFrom = this.flatIndex + lastToggledIndex
+			const globalTo = this.flatIndex + index
+			// If the clicked envelope is already selected, deselect the range
+			const deselect = this.selection.includes(envelope.databaseId)
+			this.$emit('select-range', globalFrom, globalTo, deselect)
 			this.lastToggledIndex = index
 		},
 
@@ -571,7 +602,7 @@ export default {
 			this.sortedEnvelops.forEach((env) => {
 				env.flags.selected = false
 			})
-			this.selection = []
+			this.$emit('update:selection', [], this.envelopes)
 		},
 
 		onOpenMoveModal() {
--- a/src/components/MailboxThread.vue
+++ b/src/components/MailboxThread.vue
@@ -16,7 +16,8 @@
 					<SearchMessages
 						:mailbox="mailbox"
 						:account-id="account.accountId"
-						@search-changed="onUpdateSearchQuery" />
+						@search-changed="onUpdateSearchQuery"
+						@select-all-matching="onSelectAllMatching" />
 				</div>
 				<AppContentList
 					v-infinite-scroll="onScroll"
@@ -217,6 +218,7 @@ import {
 import useMainStore from '../store/mainStore.js'
 import { groupEnvelopesByDate } from '../util/groupedEnvelopes.js'
 import {
+	favoritesQuery,
 	priorityImportantQuery,
 	priorityOtherQuery,
 } from '../util/priorityInbox.js'
@@ -276,7 +278,7 @@ export default {
 
 			priorityImportantQuery,
 			priorityOtherQuery,
-			favoriteQuery: 'is:starred',
+			favoriteQuery: favoritesQuery,
 			favoriteInitialPageSize: 5,
 			startMailboxTimer: undefined,
 			hasContent: false,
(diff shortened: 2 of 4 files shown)


ChristophWurst:
Sorry if the review rabbit caused any confusion. We enabled the integration but the config to turn auto reviews off was missing.

RobinAngele:
Tested in production — all scenarios work. The two-event design handles filter changes and mass loading reliably. Ready for review when you have time.
@ChristophWurst

src/components/MailboxThread.vue
```diff
@@ -602,6 +603,11 @@ export default {
 		onUpdateSearchQuery(query) {
 			this.searchQuery = query
 		},
+
+		onSelectAllMatching() {
+			// Signal the Mailbox component to select all after loading finishes
+			this.bus.emit('select-all-matching')
```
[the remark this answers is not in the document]
RobinAngele: This is intentional. In the Priority inbox, all sections (Favorites, Follow-up, Important, Other) are rendered as separate `Mailbox` instances sharing the same `mitt` bus. When the user clicks "Select all matching" in the search modal, each section should independently load and select its own filtered set of messages.

Each section guards against stale query reads by setting `loadingAllMatching = true` *before* `$nextTick`, then reading `effectiveQuery = this.searchQuery` after the tick — at which point its `appendToSearch()` suffix has settled. So each section correctly uses its own section-filtered query, not a shared one.

For a regular (non-Priority) folder only one `Mailbox` instance exists on the bus, so broadcasting is equivalent to targeting.

RobinAngele:
@coderabbitai review

RobinAngele:
@coderabbitai review

RobinAngele:
@coderabbitai review

RobinAngele:
@coderabbitai review

RobinAngele:
@coderabbitai review

src/components/EnvelopeList.vue
ChristophWurst, line 46: Remnant of https://github.com/nextcloud/mail/pull/12899.

Nextcloud_Mail_Select_All_and_Mass_Selection.md
```diff
@@ -0,0 +1,509 @@
+# Nextcloud Mail — Select All & Mass Selection Feature
```
ChristophWurst: Drop this file.

ChristophWurst (changes requested):

Nextcloud_Mail_Select_All_and_Mass_Selection.md
RobinAngele: Done — dropped in commit b4c53ba62.

src/components/EnvelopeList.vue
RobinAngele: Agreed — this hunk belongs to #12899. It appears here because this PR is stacked on top of that branch. Once #12899 merges and this branch is rebased onto main, the line will no longer be in this diff.

ChristophWurst: Sorry to ask this, but am I talking to Robin or the agent? The PR is merged.

ChristophWurst: @RobinAngele?

RobinAngele: @ChristophWurst it's Robin. Sorry for the noise —  I rebased, squashed, removed the doc file, and dropped the duplicate from #12899. One clean commit now. Ready for review :)

RobinAngele:
Rebased on latest main. Tested on Debian 12 alongside #12900 — stable.

PR #12915, opened 2026-05-11
fix(composer): Fix paragraph breaks getting lost on paste and sending

### LINKED ISSUE #6141: Pasting text deletes newlines
### Steps to reproduce

1. Copy some text with newlines, like some generated [lorem ipsum](https://lipsum.com/)
2. Paste it in the composer

### Expected behavior

The newlines should be kept as they are.

### Actual behavior

**Original text from lipsum.org:**
image
---
**Pasted in the composer:**
image
---
**Pasted in the [CKeditor demo](https://ckeditor.com/ckeditor-5/demo/) for testing:**
image

### Mail app version

Current on cloud.nc.com

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

Also happens when copying from other places like GitHub, Atom editor, etc. Does _not_ happen when pasting into https://ckeditor.com/ckeditor-5/demo/

jancborchardt:
In the mail composer (when writing a new mail), pasting plain text content from elsewhere like a local text editor or a website doesn’t preserve empty lines inbetween lines of text or paragraphs. Currently they are removed on pasting, but of course they should be kept.
This happens in both plain-text composing as well as when using formatting.

As an example, this text:
```
Testing testing
This is second line

Here is another paragraph


And another paragraph
```

Is being pasted wrongly and appears like this:
```
Testing testing
This is second line
Here is another paragraph

And another paragraph
```

Interestingly, when copying the block of text back from the composer into the text editor, it looks fine. So it seems to have been partly a formatting/CSS issue only but the structure was correctly preserved.

Screenshots of:
- Composer on pasting
- Received plaintext mail
- Received HTML mail

Before | After
-|-
Screenshot From 2026-05-11 13-06-12|Screenshot From 2026-05-11 12-55-02
Screenshot From 2026-05-11 13-07-23|Screenshot From 2026-05-11 12-55-14
Screenshot From 2026-05-11 13-07-43|Screenshot From 2026-05-11 12-55-19

---

1. src/components/TextEditor.vue: Visual fix (editor) Added a CSS rule so consecutive `<p>` elements render with visible spacing:
```
 :deep(p + p) { margin-top: 1em !important; }
```

2. src/util/text.js: Plain-text sending fix Added a customParagraph formatter for `<p>` that emits two line breaks (\n\n) instead of one, so paragraph breaks are preserved in sent plain-text emails.

3. lib/Service/MimeMessage.php: HTML→plain-text conversion fix Removed htmlToTextCallback, which was intentionally overriding Horde's default to collapse `<p>` to \n. Horde's default (\n\n) is now used.

4. src/ckeditor/mail/MailPlugin.js: HTML mail fix (deleted) This plugin injected `style="margin:0;"` onto every `<p>` in the serialized HTML, which overrode email client rendering and collapsed paragraph spacing in received HTML emails. Deleted entirely.

5. Tests updated
 - src/tests/unit/util/text.spec.js: updated expectations for `<p>` producing \n\n
 - src/tests/unit/components/MailPlugin.spec.js: deleted (plugin gone)
 - src/tests/unit/components/TextEditor.spec.js: updated to reflect no inline margin styles on `<p>`

AI-assisted: GitHub Copilot (Claude Sonnet 4.6)


src/components/TextEditor.vue
```diff
@@ -600,6 +598,11 @@ export default {
 	cursor: text;
 	margin: 0 !important;
 }
+
+// Show empty lines correctly in composer
+:deep(p + p) {
```
ChristophWurst, line 674: this will only fix it *in the editor*, not for the sent email, no?

jancborchardt: Yeah this part specifically is to fix the appearance when pasted into the composer (both plaintext and HTML), as without it it looks like the "before" pics. Just verified and tested again, removing that line specifically.

I started with only that CSS part, but that wasn’t enough to fix the issue. So the rest is to make it actually work during the mail sending process.

jancborchardt (commented):
@ChristophWurst so any adjustments needed or what do you think?

ChristophWurst:
I need time to test. I'll have a look soon.

ChristophWurst:
Conflicts :s

jancborchardt (commented):
@ChristophWurst was a simple conflict, fixed. :)

jancborchardt (commented):
Also fixed the linter error by removing the now unused `use DOMNode;` from `lib/Service/MimeMessage.php`

jancborchardt:
Rebased and all tests pass, ready for review @GretaD @kesselb @ChristophWurst :)

jancborchardt:
This would fix https://github.com/nextcloud/mail/issues/6141 FYI

PR #12919, opened 2026-05-11
refactor: migrate console calls to logger

ChristophWurst:
Addresses a bunch of eslint warnings.


kesselb (approved):

kesselb:
Enforce it by flagging it as error? https://github.com/nextcloud/mail/blob/17d048a2deac844415f28d98e7bc5cd7bf3d3a3c/eslint.config.mjs#L25

PR #12932, opened 2026-05-13
fix(send): skip IMAP APPEND when server already saved sent message

ChristophWurst:
Some mail servers (e.g. Microsoft Exchange/Office 365) automatically save a copy of sent messages to the Sent folder upon SMTP submission. When Nextcloud Mail also does an IMAP APPEND, the user ends up with duplicates.

After SMTP delivery succeeds, search the Sent mailbox for the outgoing message's Message-ID header. If the server already saved it, skip the APPEND. If the IMAP search fails for any reason, fall through and APPEND as normal (safe degradation).

The check is an overhead for the vast majority of setups so this is an opt-in feature. Admin docs at https://github.com/nextcloud/documentation/pull/15015.

AI-assisted: Claude Code (claude-sonnet-4-6)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate saves to the Sent folder when the mail server has already stored the sent message (handler now skips redundant appends).

* **Tests**
  * Added unit and integration tests to verify skipping behavior and proper fallback when mailbox existence checks fail.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nextcloud/mail/pull/12932)

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -429,6 +429,25 @@ class MessageMapper {
 	}
 
 	/**
+	 * Returns true if a message with the given Message-ID header already exists in $mailbox.
+	 * Used to detect server-side automatic sent-message saving (e.g. Exchange).
+	 *
+	 * @throws Horde_Imap_Client_Exception
+	 */
+	public function existsInMailboxByMessageId(
+		Horde_Imap_Client_Socket $client,
+		Mailbox $mailbox,
+		string $messageId,
+	): bool {
+		$query = new Horde_Imap_Client_Search_Query();
+		$query->headerText('Message-ID', $messageId);
+		$result = $client->search($mailbox->getName(), $query, [
+			'results' => [Horde_Imap_Client::SEARCH_RESULTS_COUNT],
+		]);
+		return ($result['count'] ?? 0) > 0;
+	}
+
+	/**
 	 * @throws Horde_Imap_Client_Exception
 	 */
 	public function addFlag(Horde_Imap_Client_Socket $client,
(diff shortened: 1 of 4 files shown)


ChristophWurst:
@coderabbit review

ChristophWurst:
@coderabbitai review

kesselb (approved):

Rikdekker:
Thanks for tackling this, @ChristophWurst  — this exact issue affects us in a production setup.

Environment: Nextcloud 33.0.6 Enterprise, Mail app 5.9.2, mailboxes connected to Microsoft 365 via OAuth (IMAP + SMTP).

Symptom: sending from Nextcloud Mail results in a single copy in the Mail app's Sent view (deduplicated by Message-ID), but two copies in Outlook/OWA, because Exchange Online saves its own server-side copy on SMTP submission and Mail additionally does an IMAP APPEND. This is exactly the scenario smtp_saves_sent addresses.

Two requests, if possible:

- Could this be squashed and merged? The only failing checks are the fixup!/unconventional-commit hygiene checks — all functional checks (unit, integration, php-cs, static analysis) are green.
- Given the backport-request label, a backport to stable5.9 would be very valuable — that's the line our NC33 Enterprise deployment tracks, and it's clearly still receiving patches.
- 
Happy to test a build against a live Microsoft 365 tenant if that helps.

T0mWz:
Hi @kesselb , @ChristophWurst ,
Do you have any idea yet as to when you'd like to merge this? :-)

Thanks, Tom

kesselb:
Thanks for the reminder! 

The failing unit tests seem related. Otherwise, only the squash is missing.

I'm still trying to convince Christoph not to merge it as-is, but instead to implement it properly as a per-account setting. That's a bit more work (DB state, UI form), but I believe it's actually the correct way to do it. 

As I understand it, "save SMTP to Sent folder" is also a configuration option on Exchange, so having a single global toggle for this feels a bit off.

PR #12936, opened 2026-05-14
Feat: make some events webhook compatible

janepie:
Makes these event types webhook compatible:
- NewMessageReceivedEvent
- MessageSentEvent
- MessageFlaggedEvent
- MessageDeletedEvent

### DIFF
--- a/lib/Events/MessageFlaggedEvent.php
+++ b/lib/Events/MessageFlaggedEvent.php
@@ -11,13 +11,15 @@ namespace OCA\Mail\Events;
 
 use OCA\Mail\Account;
 use OCA\Mail\Db\Mailbox;
+use OCA\Mail\Db\Message;
 use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IWebhookCompatibleEvent;
 
-class MessageFlaggedEvent extends Event {
+class MessageFlaggedEvent extends Event implements IWebhookCompatibleEvent {
 	public function __construct(
 		private Account $account,
 		private Mailbox $mailbox,
-		private int $uid,
+		private Message $message,
 		private string $flag,
 		private bool $set,
 	) {
@@ -32,8 +34,12 @@ class MessageFlaggedEvent extends Event {
 		return $this->mailbox;
 	}
 
+	public function getMessage(): Message {
+		return $this->message;
+	}
+
 	public function getUid(): int {
-		return $this->uid;
+		return $this->message->getUid();
 	}
 
 	public function getFlag(): string {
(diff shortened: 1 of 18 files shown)


lib/Events/MessageFlaggedEvent.php
```diff
@@ -61,4 +62,14 @@ public function getFlag(): string {
 	public function isSet(): bool {
 		return $this->set;
 	}
+
+	public function getWebhookSerializable(): array {
+		return [
+			'accountId' => $this->account->getId(),
+			'mailboxId' => $this->mailbox->getId(),
+			'messageId' => $this->uid,
```
kesselb:
```suggestion
			'messageUid' => $this->uid,
```

The uid is the identifier on the imap server. 
With messageId we usually reference the database id on our end.

kesselb:
@janepie is the target 34?

lib/Events/MessageFlaggedEvent.php
janepie: Fine by me :)

janepie:
> @janepie is the target 34?

Would be nice but I think I can live with 35, let me doublecheck

janepie:
Checked, 35 is fine!

lib/Events/MessageFlaggedEvent.php
```diff
@@ -61,4 +62,14 @@ public function getFlag(): string {
 	public function isSet(): bool {
 		return $this->set;
 	}
+
+	public function getWebhookSerializable(): array {
```
kesselb, line 53: Do you have a use case in mind for this event? 

I'm just wondering because you cannot use the messageUid for our api because the expects the message id (oc_mail_messages.id). The uid is an identifier on the imap server. The accountId and mailboxId are database ids from nextcloud mail. 

https://github.com/nextcloud/mail/blob/38004918a3eba76f8c6ca8f5d4acadf5a321cac9/lib/Controller/MessageApiController.php#L234

With the uid you could directly lookup a message on the imap server, but for that you need the name of the mailbox (which you currently won't pass, only the id but that one the imap server wont know).

janepie: Use case in mind is some hook that reacts whenever a message is flagged. Yeah I guess that would actually need the internal ID to do sth with it, tbh I just forwarded the information already available in the event as I thought that would be fine. Is there an easy way to get the internal id from here?

kesselb:
Sorry @janepie for the slow response :see_no_evil: 

I'd pushed a slight different version of the change.

janepie:
All good, I also left it lying around a bit too long :sweat_smile:

ChristophWurst (approved):

janepie:
What kind of tests would you want to have for this?

kesselb:
I've modified and extended the PR a bit to address some inconsistencies found during review. For example, the messageId in MessageDeletedEvent is actually the message's UID. I hope it's a bit cleaner now, but it needs fresh review. @ChristophWurst @DerDreschner

PR #12948, opened 2026-05-19
chore: agents commit trailer and commit format

ChristophWurst:
Ref https://docs.kernel.org/process/coding-assistants.html for the emerging standard of `Assisted-By` instead of our `AI-Assisted`

cc @GVodyanov for the other groupware app agent files.


GVodyanov (approved):
Looks good!

PR #12632, opened 2026-03-21
feat: Add route to open messages via Message-ID

SoleroTG:
Related to #2269

Hi @ChristophWurst, this is the first part of the deep-linking feature.

## What this PR does

This PR introduces a new frontend route `/apps/mail/open/<message-id>` that allows opening a specific email directly if the Message-ID is known.

The app will search for a message with the given ID across all accounts and mailboxes of the user and open it if found.

This is the foundational work for a future UI element that will allow users to copy such a link. The button for generating the link is not part of this PR and will follow in a separate one.

## How to test

1.  Find a `Message-ID` of an existing email (e.g., from the email source).
2.  Manually construct a URL like `http://<your-nextcloud-instance>/index.php/apps/mail/open/<your-message-id>`.
3.  Make sure to URL-encode the Message-ID (e.g., `<` becomes `%3C`).
4.  Navigate to the URL.
5.  The Mail app should open and display the corresponding email directly.

PR #12959, opened 2026-05-20
fix: disable CSRF check for message-id deeplinks

SoleroTG:
## Description
This PR fixes a CSRF error that occurs when accessing a message-id deeplink directly via the browser's address bar or via external apps. This feature was recently introduced in #12632 .

Since the route is accessed via a standard, stateless `GET` request from outside the Vue frontend, no CSRF token header is present. Adding the `#[NoCSRFRequired]` attribute to the `DeepLinkController::open` method allows the stateless redirect to work as intended.

## Related Issue
Follow-up fix for #12632.

## How to test
1. Log into Nextcloud.
2. Open a new tab and paste a valid deeplink directly into the URL bar, e.g.: `https://<your-nc>/apps/mail/open/<valid-message-id>`
3. Without this PR: Nextcloud throws a CSRF error.
4. With this PR: The request is processed, and you are correctly redirected to the expected email message.

### DIFF
--- a/lib/Controller/DeepLinkController.php
+++ b/lib/Controller/DeepLinkController.php
@@ -14,6 +14,7 @@ use OCA\Mail\Db\MailAccountMapper;
 use OCA\Mail\Db\MessageMapper;
 use OCA\Mail\Service\AccountService;
 use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
 use OCP\AppFramework\Http\RedirectResponse;
 use OCP\IRequest;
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> <NUMMER_DEINES_ALTEN_PR>

FTFY

ChristophWurst (approved):
Thanks for the fix

CSRF can be skipped for a GET 👍

SoleroTG:
@ChristophWurst I accidentally messed up my branches earlier, but I've cleaned it up and force-pushed the correct state. Could you re-trigger the CI for me?

PR #12964, opened 2026-05-21
fix(ui): sub-mailbox drops blocked by parent cascade

### LINKED ISSUE #12951: Cannot move mails into Inbox subfolders since Mail 5.8.0
### Steps to reproduce

Open Nextcloud Mail
Select an email located in the Inbox
Try to drag & drop the email into a subfolder of Inbox
Alternatively, use right click → "Move to"

### Expected behavior

Inbox subfolders should be selectable as valid move targets
Emails should be movable into Inbox subfolders via drag & drop and context menu

### Actual behavior

Since updating to Mail 5.8.0:

Inbox subfolders are greyed out during drag & drop operations
Dropping mails into these folders is not possible
When using right click → "Move to", the Inbox subfolders are not displayed at all

This makes it impossible to move mails from Inbox into its subfolders.

### Mail app version

5.8.0

### Nextcloud version

33.0.3

### Mailserver or service

_No response_

### Operating system

Windows

### PHP engine version

Other

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

The issue started immediately after updating to Mail 5.8.0
The problem specifically affects subfolders below Inbox
The same issue is also present in Mail 5.9.0-alpha
Browser cache refresh did not resolve the issue
This functionality worked correctly before the update

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/12951

Two bugs in the droppable-mailbox directive:

1. [droppable-mailbox="disabled"] applied pointer-events:none and opacity:.3 to the whole element, cascading into sub-mailboxes rendered in the slot and making them unreachable as drop targets.

2. With pointer-events:none unscoped, drag events reached the wrapper div of disabled mailboxes; onDragLeave unconditionally called setStatus('enabled'), visually re-enabling them mid-drag.

Scope the disabled styling to > div > a (matching the dragover state) and add canBeDropped() guards to onDragOver/onDragLeave/onDrop.

Partly introduced by https://github.com/nextcloud/mail/pull/12650. cc @cdvv7788

## How to test

1. Create a mailbox structure mb1 > mb1.1 > mb1.1.1 (parent - child - grand child)
2. Move an email into mb1
3. Open mb1
4. Try to drag and drop that email into mb1.1 or mb1.1.1

main: they are disabled
here: they are enabled

Assisted-by: Claude:claude-sonnet-4-6

### DIFF
--- a/src/directives/drag-and-drop/droppable-mailbox/droppable-mailbox.js
+++ b/src/directives/drag-and-drop/droppable-mailbox/droppable-mailbox.js
@@ -83,7 +83,7 @@ export class DroppableMailbox {
 	}
 
 	onDragOver(event) {
-		if (!this.isCurrentlyDragging) {
+		if (!this.isCurrentlyDragging || !this.canBeDropped()) {
 			return
 		}
 
@@ -102,7 +102,7 @@ export class DroppableMailbox {
 	}
 
 	onDragLeave(event) {
-		if (!this.isCurrentlyDragging) {
+		if (!this.isCurrentlyDragging || !this.canBeDropped()) {
 			return
 		}
 
@@ -111,7 +111,7 @@ export class DroppableMailbox {
 	}
 
 	async onDrop(event) {
-		if (!this.isCurrentlyDragging) {
+		if (!this.isCurrentlyDragging || !this.canBeDropped()) {
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@cdvv7788 @IT-Seezeit could you please help test and review this? Thanks

IT-Seezeit:
Unfortunately I'm not a developer, so I can't properly review it, but I applied the patch in our test environment on 5.9.0 RC and drag and drop works as expected there.

ChristophWurst:
@GretaD can you give this a test? Thanks

GretaD (approved):
tested and works well

PR #12984, opened 2026-05-27
feat(editor): preserve arbitrary font sizes and families in content

dagyu:
### Problem — custom HTML signatures don't work

Setting a **custom HTML signature** (a branded block with specific fonts and
sizes — e.g. a 13pt name, 11pt role/phone, 9pt address, a logo and links) is a
common, reasonable requirement. Today it's effectively **impossible to do
faithfully through Mail**: the signature is stored correctly, looks correct in
account settings, but **loses its typography the moment it passes through the
composer**, and reaches recipients with the wrong size and font.

Concretely, given a signature span like:

`<span style="color:#3c3c3b;font-family:Arial,Helvetica,sans-serif;font-size:13pt">…</span>`

the recipient receives:

`<span style="color:#3c3c3b">…</span>`  ← font-size and font-family gone.

`color` survives; `font-size` and `font-family` are dropped. There is no way to
work around it from the signature HTML — the loss happens inside the editor.

### Root cause

The composer's CKEditor `FontSize` / `FontFamily` plugins run with their
default presets and `supportAllValues` disabled. When signature/pasted HTML is
inserted into the editor, any `font-size` or `font-family` that isn't one of
the built-in presets is stripped during the model round-trip. (`color` is kept
because `FontColor` accepts all values by default — which is exactly the
behaviour size/family should have too.)

### Fix

Enable `supportAllValues: true` on both Font plugins so inline `font-size` and
`font-family` survive:

```js
fontSize: {
    options: [9, 10, 11, 12, 13, 14, 16, 18, 24, 'default'],
    supportAllValues: true,
},
fontFamily: {
    supportAllValues: true,
},

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -248,6 +248,22 @@ export default {
 					],
 				},
 
+				// Preserve arbitrary font sizes/families on inserted or pasted
+				// HTML (e.g. app-generated signatures). Without supportAllValues
+				// the Font plugins drop any value not in their preset list, so
+				// raw-HTML signatures lose font-size/font-family on send.
+				// NOTE: supportAllValues is incompatible with the default *named*
+				// presets ('tiny'/'big'/…) — it requires numeric options, or
+				// CKEditor throws at init and the editor fails to mount.
+				fontSize: {
+					options: [9, 10, 11, 12, 13, 14, 16, 18, 24, 'default'],
+					supportAllValues: true,
+				},
+
+				fontFamily: {
+					supportAllValues: true,
+				},
+
 			},
 		}
 	},


ChristophWurst:
Thanks for the changes

~~How can this be tested? The signature editor doesn't allow arbitrary font sizes either, does it?~~

It would be best if you could provide before and after screenshots that demo the change in addition to textual steps to reproduce.

Please point your agents to AGENTS.md to fix the commit footer.

dagyu:
@ChristophWurst  I update the commit footer as written in `AGENTS.md`. 

Following videos show exactly the bug that I had, without the patch it is impossible for me to customize the font size cause the editor doesn't preserve the values. First video is current main version, second one is the patched version.

https://github.com/user-attachments/assets/af76381b-3c1b-4e3c-ad79-605f2d139a9c


https://github.com/user-attachments/assets/3f188a26-7a72-407f-934c-c8da2a2d2060

GretaD (approved):
tested and it works, thanks a lot

dagyu:
Can you tell me please if you are planning to put this in the next release? Because I need it so if you can't I will install patched version

GretaD:
hi @dagyu sorry for the late reply. Can you please fix the lint issue so we can merge this? I tried to fix it myself, but i dont have pushing rights here :)

dagyu:
@GretaD lint issue fixed

GretaD:
Hi @dagyu theres one more lint to fix and its ready to get merged :)

dagyu:
@GretaD now any lint has been removed, the last time I didn't push the right commit

PR #12987, opened 2026-05-28
refactor: Remove long deprecated IServerContainer usages

CarlSchwan:


lib/AppInfo/Application.php
```diff
@@ -105,10 +105,10 @@ public function register(IRegistrationContext $context): void {
 		$context->registerParameter('hostname', Util::getServerHostName());
 
 		$context->registerService('userFolder', static function (ContainerInterface $c) {
-			$userContainer = $c->get(IServerContainer::class);
 			$uid = $c->get('userId');
-
-			return $userContainer->getUserFolder($uid);
+			/** @var IRootFolder $rootFolder */
+			$rootFolder = $c->get(IRootFolder::class);
```
ChristophWurst, line 110: `interface IRootFolder extends Folder, Emitter`

trading one tech debt with another 🙈 

I don't think we need the emitter. Adjust the Psalm config to ignore the missing dependency

ChristophWurst (changes requested):

lib/AppInfo/Application.php
ChristophWurst: https://github.com/nextcloud/server/pull/61539

PR #12990, opened 2026-05-28
Add AI hint that the message might contain ai generated text

### LINKED ISSUE #12689: Add a tag/header on the sent email. It will signal to the recipient that they are receiving (partly) generated content.

GretaD:
fixes #12689

### DIFF
--- a/lib/Db/LocalMessage.php
+++ b/lib/Db/LocalMessage.php
@@ -53,6 +53,8 @@ use function array_filter;
  * @method setRaw(string|null $raw)
  * @method bool getRequestMdn()
  * @method setRequestMdn(bool $mdn)
+ * @method void setAiGenerated(bool $aiGenerated)
+ * @method bool|null getAiGenerated()
  */
 class LocalMessage extends Entity implements JsonSerializable {
 	public const TYPE_OUTGOING = 0;
@@ -141,6 +143,8 @@ class LocalMessage extends Entity implements JsonSerializable {
 	/** @var bool */
 	protected $requestMdn;
 
+	protected $aiGenerated = false;
+
 	public function __construct() {
 		$this->addType('type', 'integer');
 		$this->addType('accountId', 'integer');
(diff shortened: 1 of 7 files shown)


GretaD:
This is a solution I discussed with the Claude AI agent. It suggested changing the backend as well, because my initial idea was to show the hint at all times - when the message is collapsed and when it’s expanded.

Right now, when an envelope is collapsed, only some data is loaded (sender, subject, preview text, flags, etc.). The full message body, which includes the AiGeneratedContent checks, is only fetched when you expand it.

After thinking about it more, having the hint shown at all times doesn’t seem that necessary. My understanding is that the main goal is to make sure users know a message contains AI-generated content when they open it. If they haven’t opened the message yet, it’s probably not that important for them to know.

I added my own simple solution here: https://github.com/nextcloud/mail/pull/13022

@ChristophWurst what do you think? Should we go for the backend changes or keep it simple on the front end and only have it when the body is loaded?

ChristophWurst:
I would start with the composing side. How is that flag set? Is the user able to remove it? I wonder if an embedding of a hint into the content is sufficient. That would also make this usable for clients that do not understand our custom header. We can of course do both.
But I think the marking of an email as AI generated should stay under user control. E.g. when you insert a generated snippet and remove it again you should be able to remove the AI hint.

GretaD:
>But I think the marking of an email as AI generated should stay under user control. E.g. when you insert a generated snippet and remove it again you should be able to remove the AI hint.

my simple solution here: https://github.com/nextcloud/mail/pull/13022 on the frontend side, shows the hint/header if the message body has "All or part of this reply was generated by AI"

If the user clicks smart reply but removes "All or part of this reply was generated by AI" the message will not be flagged as AI-generated. Is this good enough? 

When you say users should be able to remove it, do you mean via toggle?

ChristophWurst:
Yes that added text is probably good enough.

Yes the toggle (checkbox) would be for a header, because you can't edit a header in ckeditor. It would be added later when the MIME message is built.

GretaD:
closing it here and continuing at https://github.com/nextcloud/mail/pull/13022

PR #12993, opened 2026-05-29
[codex] fix SMTP sender validation for digit-leading domains

### LINKED ISSUE #12992: Horde_Mail_Rfc822 validator rejects valid email domains starting with digits (e.g. 180grad-kiel.de) causing "Error when parsing angle address"
## Bug description

Nextcloud Mail cannot send any emails when the mail domain starts with a digit (e.g. `180grad-kiel.de`). The Horde RFC822 validator rejects these domains as invalid when `validate => true` is set in `Transport.php`, even though RFC 1123 explicitly allows domain labels to start with digits.

## Steps to reproduce

1. Configure a Nextcloud Mail account with an email address on a domain whose first label starts with a digit (e.g. `user@180grad-kiel.de`)
2. Attempt to send any email
3. The message stays in the outbox and the following error appears in the log

## Error message

```text
Horde_Mime_Exception: Error when parsing angle address.
Horde_Mail_Exception: Error when parsing angle address.
at Horde_Mail_Rfc822->_parseAngleAddr()
```

## Root cause

In `vendor/bytestream/horde-mail/lib/Horde/Mail/Transport.php` line ~137:

```php
'validate' => $this->eai ? 'eai' : true
```

The Horde RFC822 parser with `validate => true` rejects domain labels starting with digits, despite RFC 1123 Section 2.1 explicitly relaxing the RFC 1034 restriction.

## Workaround

Change `true` to `false` in `Transport.php` line ~137.

## Environment

- Nextcloud: `33.0.3.2`
- Mail app: `5.8.1`
- PHP: `8.4`
- OS: `Debian 12`

Phillipxh:
## Summary

- route SMTP sends through a small `LenientSmtphordeTransport` wrapper
- disable Horde's strict `From` header validation for sender addresses so domains like `180grad-kiel.de` are accepted
- add a regression test for digit-leading domains and update the SMTP factory test to cover the new transport

## Why

`Horde_Mail_Transport::prepareHeaders()` uses `Horde_Mail_Rfc822` with `validate => true` for the `From` header. That rejects valid domains whose labels start with digits, which breaks message sending for addresses like `user@180grad-kiel.de` even though RFC 1123 permits them.

This keeps the workaround local to the SMTP transport used by Mail instead of patching vendor code in place.

## Validation

- Added a unit regression test for `From: Test User <user@180grad-kiel.de>`
- Ran `git diff --check`
- Could not run PHPUnit locally in this environment because `php` is not installed

Closes #12992

### DIFF
--- a/lib/SMTP/LenientSmtphordeTransport.php
+++ b/lib/SMTP/LenientSmtphordeTransport.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\SMTP;
+
+use Horde_Mail_Exception;
+use Horde_Mail_Rfc822;
+use Horde_Mail_Transport_Smtphorde;
+
+class LenientSmtphordeTransport extends Horde_Mail_Transport_Smtphorde {
+	/**
+	 * Work around Horde's overly strict domain validation for envelope senders.
+	 *
+	 * @throws Horde_Mail_Exception
+	 */
+	public function prepareHeaders(array $headers) {
+		$from = null;
+		$lines = [];
+		$raw = $headers['_raw'] ?? null;
+
+		foreach ($headers as $key => $value) {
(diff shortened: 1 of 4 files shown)


ChristophWurst:
Closing with https://github.com/nextcloud/mail/issues/12992. The validation was never the issue. It's a false negative of `0` in the horde libs.

PR #12994, opened 2026-05-29
fix(smtp): handle non ascii sender names

### LINKED ISSUE #12965: Mail 5.8.0 fails to send emails when sender display name contains Cyrillic characters
### Steps to reproduce

1. Upgrade Nextcloud Mail from 5.7.x to 5.8.0
2. Configure a mail account with a Cyrillic display name, for example:
   Иванов Иван Иванович <ivanov@company.ru>
3. Compose and send an email
4. Sending fails with:
   Horde_Mime_Exception: Address is missing domain.
5. Change the sender display name to Latin characters or to the email address itself
6. Sending works correctly

### Expected behavior

Emails should be sent successfully regardless of whether the sender display name contains non-ASCII characters (UTF-8/Cyrillic).

Properly MIME-encoded From headers should be accepted and processed correctly by the Mail app and Horde transport layer.

### Actual behavior

Sending fails when the sender display name contains Cyrillic characters.

Nextcloud logs:

Horde_Mime_Exception: Address is missing domain.

The generated From header is MIME-encoded correctly:

From: =?utf-8?b?0JjQstCw0L3QvtCyINCY0LLQsNC9INCY0LLQsNC90L7QstC40Yc=?= <ivanov@company.ru>

However, Mail/Horde appears to re-parse the already encoded header incorrectly during sending, causing the encoded display name to be interpreted as an invalid email address without a domain.

Using only Latin characters in the sender display name avoids the issue.

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

### LINKED ISSUE #12985: Email sending is broken (likely broke with 5.8.0)
We're seeing failures sending any email using the mail app.

Somewhat related to https://github.com/nextcloud/mail/issues/12965, except it happens for all email, even when only ASCII is used.

Even sending calendar invitations (as they also go via the Mail app) is broken, so unfortunately this is causing us significant trouble.

Errors on the log are either
`"Exception":"Horde_Mime_Exception","Message":"Address is missing domain."`

or
`"Exception":"Horde_Mime_Exception","Message":"Error when parsing angle address."`

We debugged a little bit, in the `$headers` array that is passed to Horde, the "To" key is empty.

ChristophWurst:
This is a hotfix. I will submit this upstream too.

Fixes https://github.com/nextcloud/mail/issues/12965
Fixes https://github.com/nextcloud/mail/issues/12985

## How to test

* Set your account name to `Christophé`
* Send an email

main: sending fails
here: sending works

Regression of https://github.com/bytestream/Mail/commit/5054666fe1a43f86a85e911111c115a133d80334.

Demo: https://3v4l.org/k9sJM

### DIFF
--- a/.patches/horde-mail-rfc822-isAtext-empty-zero.patch
+++ b/.patches/horde-mail-rfc822-isAtext-empty-zero.patch
@@ -1,0 +1,11 @@
+--- a/lib/Horde/Mail/Rfc822.php
++++ b/lib/Horde/Mail/Rfc822.php
+@@ -802,7 +802,7 @@
+             return strcspn($chr, $validate);
+         }
+
+-        $ord = empty($chr) ? 0 : ord($chr);
++        $ord = ($chr === '' || $chr === false) ? 0 : ord($chr);
+
+         /* UTF-8 characters check. */
+         if ($ord > 127) {
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -180,6 +180,12 @@ SPDX-FileCopyrightText = "angrychimp, Teon d.o.o. - Bostjan Skufca, Marcus Boint
 SPDX-License-Identifier = "MIT"
 
 [[annotations]]
+path = [".patches/horde-mail-rfc822-isAtext-empty-zero.patch"]
+precedence = "aggregate"
+SPDX-FileCopyrightText = "2026 Nextcloud GmbH and Nextcloud contributors"
+SPDX-License-Identifier = "AGPL-3.0-or-later"
+
+[[annotations]]
 path = [".patches/url-normalizer-fix-deprecation-warning-on-PHP-8.5.patch"]
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2013 Glen Scott"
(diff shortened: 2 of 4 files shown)


jonas2515:
Thanks a lot for the quick fix, looks like the actual patch file didn't make it into the commit (perhaps forgotten to `git add`) :)

jonas2515:
Yup, can confirm it fixes the issue, thanks again for the quick fix!

GretaD (approved):

GK-774:
Yes it works, thanks!

A sidenote: #12985 says "even when only ASCII is used" and that was what I could reproduce whith sender and recipient address in plain ascii (no "non ascii" at all involved). This patch fixes it, so the name "handle non ascii sender names" is not correct, at least not for #12985.

ChristophWurst:
True, it's not about ASCII but the `0` character. Encoding of non-ascii can trigger it.

PR #12995, opened 2026-05-31
fix(AccountSettings): overflow bug when using autoresponder checkbox

GVodyanov:
To reproduce:

1. Open account settings
2. Find autoresponder settings
3. Switch between autoresponder on/off

You will experience this:
image


ChristophWurst:
Is this an upstream bug?

GretaD:
I cannot reproduce this..in which version is it present?

ChristophWurst:
v5.9.3 and v5.10.0 have it

GVodyanov:
> Is this an upstream bug?

Yes, this is a quick fix

ChristophWurst (approved):
Tested and works

PR #13006, opened 2026-06-01
fix(EnvelopeList): remove multiple select bars and fix drag and drop

GVodyanov:
Fixes:
- Different sections in the envelope list (like today, yesterday, last week) having separate multiselect action bars
- Dragging and dropping multiselected envelopes would only move one of the emails and not all of them

### DIFF
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -116,7 +116,29 @@
 			</div>
 		</transition>
 
-		<transition-group :name="listTransitionName">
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Same as https://github.com/nextcloud/mail/pull/12900?

ChristophWurst (commented):
Tested and there are still multiple multiselect bars: one for each.

The actions in the bar says "Mark x as important" but only considers the selection of its section, not the others when counting.

PR #13010, opened 2026-06-02
[main] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (3 unfixable, 27 only fixable manually using --force)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -104,20 +104,6 @@
         "npm": "^11.3.0"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.3.0",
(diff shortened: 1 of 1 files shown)


ChristophWurst (changes requested):
Claude: axios 1.17.0 breaks actions.spec.js. The new version adds https-proxy-agent as a dependency, which likely changes axios's adapter initialization in the Node.js/jsdom test environment in a way that causes   
  vi.mock() on MailboxService to no longer intercept HTTP calls.

ChristophWurst:
@coderabbitai review

ChristophWurst:
@coderabbitai review again and look closely at the axios bump

ChristophWurst (approved):

Issue #12538, opened 2026-03-09
LLM doesnt work to fill out event/meeting title, subject and description


GretaD:
### Steps to reproduce

1. Have the Integration AI app activate and LLM on groupware admin settings
2. Open a message and more action menu
3. Then click reply with a meeting

### Expected behavior

the title and description get an ai summary

### Actual behavior

nothing happens, same text as in the message is shown

### Mail app version

latest

### PHP engine version

None

### Web server

None

### Database

None

PR #13018, opened 2026-06-03
fix(ui): harden error handling for AI summaries

GretaD:
For #12538

claude was used to fix this issue


ChristophWurst (changes requested):
This doesn't fix AI summaries. This adds hardening for when summaries can not be generated. https://github.com/nextcloud/mail/issues/12538 will still happen.
Point the agent to AGENTS.md to have the required commit footer added.

ChristophWurst (approved):

ChristophWurst (changes requested):
You have changed the PR title but not the commit message. Change it to `fix(ui): harden error handling for AI summaries`

I also fixed the PR description so it doesn't auto-close the issue that won't be fixed

ChristophWurst (commented):
The hardening for meeting agendas is missing. Is that intentional?

GretaD:
>  Is that intentional?

No. Added it

ChristophWurst (approved):

PR #12646, opened 2026-03-23
feat: add ai hint for smart replies

GretaD:
fixes #12390
 
**rich text**
Screenshot from 2026-03-23 17-12-20

______________________________________
**plain text**
Screenshot from 2026-03-23 17-14-03

I wrote _Part of this reply_ because people can also add more manually as a reply. Happy to change it if someone has a better idea.

buttons
open
Screenshot from 2026-03-27 17-42-16

closed
Screenshot from 2026-03-27 17-42-00

Issue #12390, opened 2026-02-11
Improve clarity of generated content in suggested replies


ChristophWurst:
### Is your feature request related to a problem? Please describe.

As a Mail user, I want to be able to recognize generated content.

### Describe the solution you'd like

*Suggested replies* must have

1. A hint for the user underneath buttons
2. A hint in the generated content/email

Additionally, it should have

1. A tag/header on the sent email. This header needs research. Is there a standard yet? It will signal to the recipient that they are receiving (partly) generated content

### Describe alternatives you've considered

N/a

nimishavijay:
Interesting problem. Do we need to show it in the message list? It's already overloaded with lots of information. If we must, then I would say we do what @GretaD posted, but with different wording "Contains AI generated content" (which then also addresses images which are AI generated), and make the text regular weight to avoid the text wrapping/ellipsizing in other languages

Also another idea I had was to make this look like a teeny tiny ncnotecard component, which would feel a bit more like a warning, what do you think?

<img width="485" height="192" alt="Image" src="https://github.com/user-attachments/assets/a159dcd9-8fff-4254-bffd-d8fe75b93270" />

ChristophWurst:
I don't think it has to be part of the thread list.

ChristophWurst:
It must not be a tag (permflag). These are not part of the email sent out. It has to be a header.

ChristophWurst:
Relevant: https://github.com/nextcloud/server/blob/1ab09ec753f106661beadee1794b390ebec455dd/lib/public/SystemTag/ISystemTag.php#L43

GretaD:
closing as done

PR #13022, opened 2026-06-03
feat: add ai hint that content contains ai generated text

### LINKED ISSUE #12689: Add a tag/header on the sent email. It will signal to the recipient that they are receiving (partly) generated content.

GretaD:
fixes #12689
Screenshot from 2026-06-25 17-36-56


src/components/Composer.vue
```diff
@@ -755,6 +766,7 @@ export default {
 
 			editorMode: (this.body?.format !== 'html') ? EDITOR_MODE_TEXT : EDITOR_MODE_HTML,
 			requestMdnVal: this.requestMdn,
+			isAiGeneratedVal: this.isAiGenerated,
```
kesselb, line 807: Claude: Please also reset isAiGeneratedVal in reset method

src/components/Composer.vue
```diff
@@ -1344,6 +1361,7 @@ export default {
 			}
 			if (this.smartReply) {
 				this.bus.emit('append-to-body-at-cursor', this.smartReply)
+				this.isAiGeneratedVal = true
```
kesselb, line 1431: Claude flagged this block, because a onEditorReady event (e.g the user is changing the alias) would insert the smart reply again, reset isAiGeneratedVal to true, and overwrite the user's choice. I think it's okay for now to ignore it, but we should look into a better way as follow up.

appinfo/info.xml
```diff
@@ -34,7 +34,7 @@ The rating depends on the installed text processing backend. See [the rating ove
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 	]]></description>
-	<version>5.10.0-rc.1</version>
+	<version>5.11.0-dev.1</version>
```
odzhychko: question: Is this change intentional?

lib/Db/LocalMessage.php
```diff
@@ -195,6 +200,7 @@ public function jsonSerialize() {
 			'status' => $this->getStatus(),
 			'raw' => $this->getRaw(),
 			'requestMdn' => $this->getRequestMdn(),
+			'isAiGenerated' => $this->isAiGenerated() === true,
```
odzhychko: issue: Inconsistency. But maybe me wrongly understanding PHP types ^^


`$aiGenerated` declared as `bool`.
` $this->isAiGenerated() === true` handles it as if it could be `null`.

lib/Service/MailTransmission.php
```diff
@@ -138,6 +138,10 @@ public function sendMessage(Account $account, LocalMessage $localMessage): void
 			$fccHeaders->addHeaderOb(new Horde_Mime_Headers_Addresses(Horde_Mime_Mdn::MDN_HEADER, $from->toHorde()));
 		}
 
+		if ($localMessage->isAiGenerated()) {
+			$fccHeaders->addHeader('X-AI-Generated', '1');
```
odzhychko: nit: Inconsistency `'x-ai-generated'` vs `'X-AI-Generated'`. Guess, we should declare it somewhere centrally.

src/components/ThreadEnvelope.vue
```diff
@@ -1189,6 +1199,10 @@ export default {
 <style lang="scss" scoped>
 	.sender {
 		margin-inline-start: calc(var(--default-grid-baseline) * 3);
+		display: flex;
+		align-items: center;
+		gap: 6px;
```
odzhychko: nit: What code abbit said in https://github.com/nextcloud/mail/pull/13022/changes#r3508125797

tests/Unit/Model/IMAPMessageTest.php
```diff
@@ -156,9 +158,11 @@ public function testSerialize() {
 			'hasHtmlBody' => true,
 			'dispositionNotificationTo' => 'disposition',
 			'hasDkimSignature' => false,
+			'aiGenerated' => false,
 			'phishingDetails' => [],
 			'scheduling' => [],
 			'isPgpMimeEncrypted' => false,
+			'aiGenerated' => false,
```
odzhychko: nit: second `'aiGenerated' => false` is redundant

lib/Migration/Version5012Date20260604000000.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+class Version5201Date20260604000000 extends SimpleMigrationStep {
+	#[\Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('mail_local_messages');
+		if (!$table->hasColumn('ai_generated')) {
+			$table->addColumn('ai_generated', Types::BOOLEAN, [
+				'notnull' => false,
```
odzhychko, line 29: issue: This says it might be `null`, but LocalMessage.php declares it as non-null.

```php
/** @var bool */
protected $aiGenerated;
```

odzhychko (commented):
Looked at PHP code. Just some small stuff.

Testing and looking at UI code now.

src/components/Composer.vue
```diff
@@ -400,6 +400,12 @@
 							@uncheck="requestMdnVal = false">
 							{{ t('mail', 'Request a read receipt') }}
 						</ActionCheckbox>
+						<ActionCheckbox
```
odzhychko: just for reference: looks like this

Image

appinfo/info.xml
SebastianKrupinski: Yes its intentional, you need to change the apps version number to trigger the migration

lib/Migration/Version5012Date20260604000000.php
SebastianKrupinski: That is fine this is just a "must have a value enforcement"

odzhychko (changes requested):
issue(blocking): Draft does not "AI generated" setting.

- Start composing a mail
- Add some subject or body text
- Check as AI generated
- Close to save as draft
- Open draft
- Actual: Not marked as AI generated
- Expected: Should be marked as AI generated

~Cause: Seems to be due to inconsistent usage of `isAIGenerated` vs just `aiGenerated`.~
~(I also didn't notice this while reviewing PHP.) 😆~

In some places `aiGenerated` might got mixed up with `isAiGenerated`.

SebastianKrupinski:
> issue(blocking): Draft does not "AI generated" setting.
> 
>     * Start composing a mail
> 
>     * Add some subject or body text
> 
>     * Check as AI generated
> 
>     * Close to save as draft
> 
>     * Open draft
> 
>     * Actual: Not marked as AI generated
> 
>     * Expected: Should be marked as AI generated
> 

Looks like the message is losing its "X-AI-Generated" header, when its pushed to the IMAP server

lib/Service/MailTransmission.php
```diff
@@ -246,6 +250,9 @@ public function saveLocalDraft(Account $account, LocalMessage $message): void {
 
 		$mail = new Horde_Mime_Mail();
 		$mail->addHeaders($headers);
+		if ($message->isAiGenerated()) {
+			$mail->addHeader(LocalMessage::HEADER_AI_GENERATED, '1');
```
odzhychko: note(non-blocking): This is `saveLocalDraft` and I wondered what `saveDraft` does.

`saveDraft` is reachable by `/api/accounts/{id}/draft` with `NewMessageData`. Seems to be used by external clients that want to integrate with the Mail app.

---

`NewMessageData` does not have `isAiGenerated`, so not changing `saveDraft` seems correct. Adding and handling `NewMessageData.isAiGenerated` seem out-of-scope because this in concerned with adding the AI generated notice when sending from the Mail app.

lib/Service/MailTransmission.php
odzhychko, line 227: note(non-blocking): Looked how `saveDraft` and `saveLocalDraft` differ.

- `saveLocalDraft` seems not to emit `SaveDraftEvent`

Might be a bug.

lib/Db/MailAccount.php
```diff
@@ -107,6 +107,10 @@
  * @method void setClassificationEnabled(bool $classificationEnabled)
  * @method bool getImipCreate()
  * @method void setImipCreate(bool $value)
+ * @method string getProtocol()
+ * @method void setProtocol(string $protocol)
+ * @method string|null getPath()
+ * @method void setPath(?string $path)
```
kesselb, line 113: Does not belong to this PR.

lib/Db/MailAccount.php
```diff
@@ -107,10 +107,6 @@
  * @method void setClassificationEnabled(bool $classificationEnabled)
  * @method bool getImipCreate()
  * @method void setImipCreate(bool $value)
- * @method string getProtocol()
```
odzhychko, line 110: issue: Accidentally removed during conflict resolution.

lib/Db/Message.php
```diff
@@ -121,6 +121,8 @@ class Message extends Entity implements JsonSerializable {
 	protected $mentionsMe = false;
 	protected ?string $remoteId = null;
 
+	protected ?string $remoteId = null;
```
odzhychko: issue: Accidentally added during conflict resolution.

src/components/ThreadEnvelope.vue
```diff
@@ -1510,4 +1524,11 @@ export default {
 		display: inline;
 		align-items: center;
 	}
+
+	.ai-generated-label {
+		display: flex;
+		align-items: center;
+		gap: 4px;
```
odzhychko:
```suggestion
		gap: var(--default-grid-baseline);
```

odzhychko (changes requested):

odzhychko:
> issue(blocking): Draft does not "AI generated" setting.

I still have the issue with a reopened draft not being marked as AI-generated.
Not sure, what is the cause as the code looks good.
I will debug a bit to make sure that I didn't break my setup (maybe I didn't execute migrations correctly).

appinfo/info.xml
odzhychko: Now this PR does not have a `appinfo/info.xml`. Is this an issue?

SebastianKrupinski: yes it needs a version bump to trigger the migration on upgrade

lib/Migration/Version5012Date20260604000000.php
odzhychko: Sorry, but I do not understand that.

`LocalMessage#aiGenerated` is declared as `bool` and not `bool | null`.
So `LocalMessage#aiGenerated` should never be `null`.

```php
class LocalMessage extends Entity implements JsonSerializable {

	/** @var bool */
	protected $aiGenerated;
```

But this migration declares the column `ai_generated` and declares it as `notnull`.

```php
if (!$table->hasColumn('ai_generated')) {
	$table->addColumn('ai_generated', Types::BOOLEAN, [
		'notnull' => false,
		'default' => false,
	]);
}
```

---

In my understanding ` LocalMessage#aiGenerated` can be `null` according to this migration.
Needs to be handled as such.
And should be declared as `bool | null`.

SebastianKrupinski:
> > issue(blocking): Draft does not "AI generated" setting.
> 
> I still have the issue with a reopened draft not being marked as AI-generated. Not sure, what is the cause as the code looks good. I will debug a bit to make sure that I didn't break my setup (maybe I didn't execute migrations correctly).

Turn on debug on the account and check if the email pushed to the imap server has the "X-AI-Generated" header, you might need to turn on "literal" logging in the horde client to see the actual message being pushed

lib/Migration/Version5012Date20260604000000.php
SebastianKrupinski: This in the constructor, tells the entity that the field is a boolean:

$this->addType('aiGenerated', 'boolean');

It should then be cast automatically here to a boolean: 

https://github.com/nextcloud/server/blob/c44ef837082c61569c2daba93c0387ffa9e6bf8a/lib/public/AppFramework/Db/Entity.php#L142

And null = false in PHP land

odzhychko:
This is my log. Seems the header is missing when saving.

```
------------------------------
>> Wed, 12 Aug 2026 12:51:06 +0000
>> Connection to: imap://imap.web.de:993/
------------------------------
>> Wed, 12 Aug 2026 12:51:06 +0000
>> Connection to: imap://imap.web.de:993/
>> Server connection took 0.1448 seconds.
>> Server connection took 0.0971 seconds.
S: * OK [CAPABILITY IMAP4rev1 CHILDREN ENABLE ID IDLE LIST-EXTENDED LIST-STATUS LITERAL- MOVE NAMESPACE QUOTA SASL-IR SORT SPECIAL-USE THREAD=ORDEREDSUBJECT UIDPLUS UNSELECT WITHIN AUTH=LOGIN AUTH=PLAIN] IMAP server ready H miweb184 32.3.1 IMAP-1M5QuV-1wtJVL2bQr-001xUc
S: * OK [CAPABILITY IMAP4rev1 CHILDREN ENABLE ID IDLE LIST-EXTENDED LIST-STATUS LITERAL- MOVE NAMESPACE QUOTA SASL-IR SORT SPECIAL-USE THREAD=ORDEREDSUBJECT UIDPLUS UNSELECT WITHIN AUTH=LOGIN AUTH=PLAIN] IMAP server ready H miweb131 32.3.1 IMAP-1MfIQb-1xNhVm2rWN-00g3RF
C: 1 AUTHENTICATE PLAIN [INITIAL CLIENT RESPONSE (username: xxx)]
C: 1 AUTHENTICATE PLAIN [INITIAL CLIENT RESPONSE (username: xxx)]
S: 1 OK AUTHENTICATE completed
S: 1 OK AUTHENTICATE completed
>> Command 1 took 0.1971 seconds.
>> Command 1 took 0.1971 seconds.
C: 2 CAPABILITY
C: 2 CAPABILITY
S: * CAPABILITY IMAP4rev1 CHILDREN ENABLE ID IDLE LIST-EXTENDED LIST-STATUS LITERAL- MOVE NAMESPACE QUOTA SASL-IR SORT SPECIAL-USE THREAD=ORDEREDSUBJECT UIDPLUS UNSELECT WITHIN APPENDLIMIT=27962027
S: 2 OK CAPABILITY completed
>> Command 2 took 0.0326 seconds.
>> CACHE: Using the OCA\Mail\Cache\Cache storage driver.
C: 3 ID NIL
S: * CAPABILITY IMAP4rev1 CHILDREN ENABLE ID IDLE LIST-EXTENDED LIST-STATUS LITERAL- MOVE NAMESPACE QUOTA SASL-IR SORT SPECIAL-USE THREAD=ORDEREDSUBJECT UIDPLUS UNSELECT WITHIN APPENDLIMIT=27962027
S: 2 OK CAPABILITY completed
>> Command 2 took 0.0411 seconds.
>> CACHE: Using the OCA\Mail\Cache\Cache storage driver.
C: 3 ID NIL
S: * ID ("name" "nemesis" "version" "32.3.1")
S: * ID ("name" "nemesis" "version" "32.3.1")
S: 3 OK ID completed
>> Command 3 took 0.0391 seconds.
S: 3 OK ID completed
>> Command 3 took 0.0308 seconds.
C: 4 APPEND Entwurf (\seen \draft) {241}
C: 4 STATUS INBOX (MESSAGES UIDNEXT UIDVALIDITY)
S: + Ready for literal data
Date: Wed, 12 Aug 2026 12:51:06 +0000
Message-ID: <20260812125106.Horde.7QkIba32bWtm1a7c_z8Dq_w@localhost>
From: myweb <*****>
To: 
Subject: Subject a draft with ai
Content-Type: text/plain; charset=utf-8
MIME-Version: 1.0

C: 
S: * STATUS INBOX (MESSAGES 0 UIDNEXT 82 UIDVALIDITY 1782204910)
S: 4 OK STATUS completed
>> Command 4 took 0.0381 seconds.
C: 5 STATUS INBOX (MESSAGES UIDNEXT UIDVALIDITY)
S: * STATUS INBOX (MESSAGES 0 UIDNEXT 82 UIDVALIDITY 1782204910)
S: 5 OK STATUS completed
>> Command 5 took 0.0352 seconds.
C: 6 SELECT INBOX
S: * 0 EXISTS
S: * 0 RECENT
S: * FLAGS (\Answered \Flagged \Deleted \Seen \Draft)
S: * OK [PERMANENTFLAGS (\Answered \Flagged \Draft \Deleted \Seen \*)] Unlimited
S: * OK [UIDNEXT 82] Predicted next UID
S: * OK [UIDVALIDITY 1782204910] UIDs valid
S: 6 OK [READ-WRITE] SELECT completed
>> Command 6 took 0.048 seconds.
C: 7 LOGOUT
S: 4 OK [APPENDUID 1782204910 49] APPEND completed
>> Command 4 took 0.1703 seconds.
C: 5 LOGOUT
S: * BYE Server logging out
S: 7 OK LOGOUT completed
>> Command 7 took 0.0388 seconds.
S: * BYE Server logging out
S: 5 OK LOGOUT completed
>> Command 5 took 0.0374 seconds.
```

lib/Service/MailTransmission.php
```diff
@@ -237,6 +241,10 @@ public function saveLocalDraft(Account $account, LocalMessage $message): void {
 		$draftHeaders = $this->buildMimeHeaders($from, $to, $cc, $bcc, $message->getSubject());
 
 		$mail = new Horde_Mime_Mail();
+		$mail->addHeaders($headers);
+		if ($message->isAiGenerated()) {
+			$mail->addHeader(LocalMessage::HEADER_AI_GENERATED, '1');
```
odzhychko: Seems like this needs to be added to `$draftHeaders`.

src/components/NewMessageModal.vue
```diff
@@ -92,6 +92,7 @@
 						:is-first-open="modalFirstOpen"
 						:is-draft="composerData.draftId !== undefined"
 						:request-mdn="composerData.requestMdn"
+						:is-ai-generated="composerData.isAiGenerated"
```
odzhychko, line 126: issue: `composerData.isAiGenerated` does not exist when opening a draft.

lib/Migration/Version5012Date20260604000000.php
odzhychko: Aaaaah. Thx.

odzhychko (changes requested):
Tested it. Works as expected and code look good.

Following comments still need resolving before merging to `main`:

- https://github.com/nextcloud/mail/pull/13022#discussion_r3759755675
- https://github.com/nextcloud/mail/pull/13022#discussion_r3766010448

SebastianKrupinski:
> Following comments still need resolving before merging to `main`:

And...

https://github.com/nextcloud/mail/pull/13022/changes#r3766010448

odzhychko (changes requested):
There are still removals in `MailAccount.php` that to do not related to this PR.

See https://github.com/nextcloud/mail/pull/13022#discussion_r3766010448

lib/Db/MailAccount.php
odzhychko: This relates to all lines removed in `MailAccount.php`.

GretaD: yes, returning these manually would have been easier than asking claude :)

odzhychko (approved):

kesselb:
I've rebased the branch against main, resolved the conflicts, made some smaller changes:

- Version 5.13.0-alpha.1 => 5.12.0-alpha.2
- Version5201Date20260604000000.php -> Version5012Date20260604000000.php 
- ImapMessageFetcher.php revert of the HEADER_STREAM/parseHeaders rewrite
- Composer.vue 
  - NcActionCheckbox :checked -> :model-value
  - TODO note on the smart-reply block changing isAiGeneratedVal when the editor is recreated on editorMode change
- Split the two flags by role: outgoing stays isAiGenerated on LocalMessage; incoming becomes hasAiGeneratedHeader on IMAPMessage + ImapMessageFetcher

lib/IMAP/ImapMessageFetcher.php
odzhychko, line 533: > ImapMessageFetcher.php revert of the HEADER_STREAM/parseHeaders rewrite

question(non-blocking, out-of-scope): Was it reverted because it was an unrelated change or because the code had issues?

This changed seemed to be an improvement to me.

1. Less code for us
2. Correcter call to `fclose`
    See how ` case self::HEADER_PARSE:` only conditionally calls `fclose`.
    I even suspect that we might actually sometimes cause a bug here by always calling `fclose`.
    
---

Current code:

```php 
$headersStream = $fetch->getHeaderText('0', Horde_Imap_Client_Data_Fetch::HEADER_STREAM);
$parsedHeaders = Horde_Mime_Headers::parseHeaders($headersStream);
fclose($headersStream);
```


Impl:

```php

   public function getHeaderText($id = 0, $format = 0)
    {
        return $this->_getHeaders(
            $id,
            $format,
            Horde_Imap_Client::FETCH_HEADERTEXT
        );
    }
   
    [...]

    protected function _getHeaders($id, $format, $key)
    {
        switch ($format) {
        case self::HEADER_STREAM:
            if (!isset($this->_data[$key][$id])) {
                $data = null;
            } elseif (is_object($this->_data[$key][$id])) {
                switch ($key) {
                case Horde_Imap_Client::FETCH_HEADERS:
                    $data = $this->_getHeaders($id, 0, $key);
                    break;

                case Horde_Imap_Client::FETCH_HEADERTEXT:
                case Horde_Imap_Client::FETCH_MIMEHEADER:
                    $data = $this->_data[$key][$id];
                    break;
                }
            } else {
                $data = $this->_data[$key][$id];
            }

            return $this->_msgText(true, $data);

        case self::HEADER_PARSE:
            if (!isset($this->_data[$key][$id])) {
                return new Horde_Mime_Headers();
            } elseif (is_object($this->_data[$key][$id])) {
                switch ($key) {
                case Horde_Imap_Client::FETCH_HEADERS:
                    return clone $this->_data[$key][$id];

                case Horde_Imap_Client::FETCH_HEADERTEXT:
                case Horde_Imap_Client::FETCH_MIMEHEADER:
                    $hdrs = $this->_data[$key][$id];
                    break;
                }
            } else {
                $hdrs = $this->_getHeaders($id, self::HEADER_STREAM, $key);
            }

            $parsed = Horde_Mime_Headers::parseHeaders($hdrs);
			if (is_resource($hdrs)) {
				// Close the temporary stream
				fclose($hdrs);
			}
			return $parsed;
        }

        if (!isset($this->_data[$key][$id])) {
            return '';
        }

        if (is_object($this->_data[$key][$id])) {
            switch ($key) {
            case Horde_Imap_Client::FETCH_HEADERS:
                return $this->_data[$key][$id]->toString(
                    array('nowrap' => true)
                );

            case Horde_Imap_Client::FETCH_HEADERTEXT:
            case Horde_Imap_Client::FETCH_MIMEHEADER:
                return strval($this->_data[$key][$id]);
            }
        }

        return $this->_msgText(false, $this->_data[$key][$id]);
    }
```

src/components/Composer.vue
```diff
@@ -1407,8 +1424,11 @@ export default {
 			if (!this.isDraft || this.changeSignature) {
 				this.insertSignature()
 			}
+			// TODO: the editor is recreated on editorMode change, so this re-appends the
+			// smart reply and re-arms isAiGeneratedVal even after the user unchecked it.
 			if (this.smartReply) {
```
odzhychko: question: Why is this a TODO? Seems like fixed you fixed it and it reads like a comment.

src/components/Composer.vue
```diff
@@ -1407,8 +1424,11 @@ export default {
...
 				this.bus.emit('append-to-body-at-cursor', this.smartReply)
+				this.isAiGeneratedVal = true
```
odzhychko, line 1431: issue: This checks the box without a user interaction user can miss that this is set (and send) because it is hidden behind a menu.

This might become frustrating/a problem, if:

1. A user use suggested reply but doesn't want message to be marked as AI-generated.
    Not everybody is required by law to indicate this...
2. User uses a smart reply and then manually changes the whole text.

Reminds me of Copilot become silently co author in VSCode.[^1][^2] 

I **mildly** lean towards not setting `this.isAiGeneratedVal = true` for smart replies until we solve this UX visibility issue. We already add the text "(All or part of this reply was generated by AI)" to smart reply. So I'd rather be inconsistent than making unexpected things for the user.

(Sorry, noticing this only so late. You TODO comment made me reread the code ^^)

[^1]: https://news.ycombinator.com/item?id=47958353
[^2]: https://github.com/orgs/community/discussions/194075

odzhychko (commented):
Code looks good. Works as expected.

But want to have a second person think about https://github.com/nextcloud/mail/pull/13022/changes/45f2e2c70c8b34072c80800418a2a3fac28c2872#r3810226996
If they don't see this as blocking issue you can close it.

lib/IMAP/ImapMessageFetcher.php
kesselb: Because unrelated.

src/components/Composer.vue
kesselb: @GretaD @ChristophWurst

GretaD: > We already add the text "(All or part of this reply was generated by AI)" to smart reply. So I'd rather be inconsistent than making unexpected things for the user.

I totally agree with this, and i have already brought it up twice: https://github.com/nextcloud/mail/issues/12390#issuecomment-4170755855

and here: https://github.com/nextcloud/mail/pull/12646#issuecomment-4117392406

GretaD: Since we need this to be screenshotted, should we merge it and I follow up with the change if thats what we decide to do?

odzhychko: > Since we need this to be screenshotted, should we merge it and I follow up with the change if thats what we decide to do?

Totally fine with me. My approval is given ^^

PR #13035, opened 2026-06-08
Change order of sans-serif and Noto Color Emoji

Nihlus:
If Noto Color Emoji is installed, but none of the preceding fonts are, the UI for plaintext emails gets pretty messed up with super-wide spaces as the emoji font takes precedence. The rest of NC has sans-serif placed ahead of the Noto emoji font for this reason.

### DIFF
--- a/css/html-response.css
+++ b/css/html-response.css
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 * {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Thanks for the PR! Could you show before/after screenshots to demo the change?

Nihlus:
Sure thing! I've blacked out some identifying information but the issue should be pretty clear.

# Before
before

# After
after

ChristophWurst (approved):
Makes sense, thanks

Ref https://github.com/nextcloud/server/blob/549790c77295379ed0d2d32eeca4ef6800515492/core/css/variables.scss#L81

ChristophWurst:
Please amend your commit and change the commit message to `fix(ui): prefer sans-serif over Noto Color Emoji for text rendering`

Nihlus:
Done!

ChristophWurst (approved):

PR #13062, opened 2026-06-10
fix(tests): externalize CJS deps in vitest config

ChristophWurst:
https-proxy-agent and agent-base (CJS-only, pulled in by axios 1.17.0) interfere with module mocking in vitest v4's vmForks pool, causing vi.mock() to silently fail on MailboxService/MessageService.

Assisted-by: Claude:claude-opus-4-7

## 🤖 AI (if applicable)

### DIFF
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -15,5 +15,10 @@ export default defineConfig({
 		environment: 'jsdom',
 		// Required for transforming CSS files
 		pool: 'vmForks',
+		server: {
+			deps: {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Closing in favor of #13064 (ci: pin front-end unit tests to ubuntu-24.04). The vitest.config.js change in this PR doesn't address the root cause — frontend tests fail only on self-hosted runners due to vitest v4 + vmForks incompatibility with that environment.

PR #13076, opened 2026-06-12
fix(provisioning): prevent silent save failure around master user

tnxfi:
For #12442

With a master user set, saving the provisioning config could fail with no visible feedback when "use master password" was toggled off, or when the separator was left empty (the separator became required-but-empty, so the submit was blocked by native validation only).

- Clear master user + separator when master password is disabled, so a value that just became hidden cannot block validation or silently survive.
- Default the master user separator to "*" (Dovecot default) when a master user is entered, so the separator field is never required-but-empty.
- Give the master password input a unique id per provisioning config (it was a hardcoded duplicate shared with the "add new config" form).

## 🤖 AI (if applicable)


ChristophWurst:
Thanks for the PR

PR #13083, opened 2026-06-15
fix(psalm): adapt to updated OCP dev-master return types

GretaD:
Tests are failing here: https://github.com/nextcloud/mail/pull/13018
Tried to fix it with claude

- IRequest::getUploadedFile now declared as array|null; use inline isset() so psalm can narrow the type correctly
- IUser::getUID now declared as non-nullable string; remove redundant null guard in Provisioning::buildUserEmail

Assisted-by: ClaudeCode:claude-sonnet-4-6

## 🤖 AI (if applicable)

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal attachment handling logic.
  * Simplified user identification placeholder processing.

### DIFF
--- a/lib/Controller/MessageApiController.php
+++ b/lib/Controller/MessageApiController.php
@@ -455,8 +455,7 @@ class MessageApiController extends OCSController {
 	 */
 	private function handleAttachments(): array {
 		$fileAttachments = $this->request->getUploadedFile('attachments');
-		$hasAttachments = isset($fileAttachments['name']);
-		if (!$hasAttachments) {
+		if (!isset($fileAttachments['name'])) {
 			return [];
 		}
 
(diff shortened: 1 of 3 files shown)


lib/Controller/MessageApiController.php
```diff
@@ -455,8 +455,7 @@ public function getAttachment(int $id, string $attachmentId): DataResponse {
 	 */
 	private function handleAttachments(): array {
 		$fileAttachments = $this->request->getUploadedFile('attachments');
-		$hasAttachments = isset($fileAttachments['name']);
-		if (!$hasAttachments) {
+		if (!isset($fileAttachments['name'])) {
```
ChristophWurst, line 458: I don't get this one. It's the identical code, just inlined?

ChristophWurst:
@coderabbitai review

lib/Controller/MessageApiController.php
ChristophWurst: It's because Psalm has issues detecting the types otherwise.

ChristophWurst (approved):
Pushing a commit to also fix the OOO controller test that fails now because the iuser mock returns an empty string instead of null

PR #13087, opened 2026-06-16
fix(ui): restore virtual start mailbox too

### LINKED ISSUE #13084: "All Inboxes" is not remembered as  the"last used" folder

Hi,

I no longer use the priority inbox, but it's still present and Mail defaults to it. 

I'd prefer to have the "all inboxes" only, or at least set as default.

I understand from this https://github.com/nextcloud/mail/issues/12872 that we won't have an option to disable priority and/or all inboxes, because Mail opens on the last opened folder.

It's true when it's a specific inbox folder, but not with "all inboxes" unified folder.

So would it be possible to generalize that behaviour (Mail opening on the last used folder) *including* the unified "all inboxes" folder please ?

Thanks

_No response_

### Additional context

_No response_

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/13084

The *start mailbox id* is always saved, also for *priority* and *unified*. But since it was always parsed to an int the two virtual mailboxes did not work. Priority only happened to "work" because it was also the fallback.

## 🤖 AI (if applicable)

### DIFF
--- a/src/init.js
+++ b/src/init.js
@@ -57,10 +57,13 @@ export default function initAfterAppCreation() {
 		key: 'sort-favorites',
 		value: preferences['sort-favorites'],
 	})
-	const startMailboxId = preferences['start-mailbox-id']
+	let startMailboxId = preferences['start-mailbox-id']
+	if (startMailboxId && !['unified', 'priority'].includes(startMailboxId)) {
+		startMailboxId = parseInt(startMailboxId, 10)
+	}
 	mainStore.savePreferenceMutation({
 		key: 'start-mailbox-id',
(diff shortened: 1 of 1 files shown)


madbob (approved):

kesselb (approved):

h3n3b0sch:
Why is it so hard to have an option to remove the priority box? It seems to me that implementing the priority box was way more work dan disabling would be... But then, I'm not a developer, I'm just an user who get's irritated when opening the mail.
Maybe people exists who'd like a priority box but it seems there are a lot of people who don't want it and have no use for it.

Please understand that it is not something I say to bother you, but please understand that there are a lot of users (since 2020) who get irritated by this box.

PR #13095, opened 2026-06-17
feat(sieve): implement redirecting action in filter UI

hazveg:
This patch implements an action for redirecting via the Sieve 'redirect' directive.

form to create forward

I've attempted to include a corresponding test case, but was unfortunately unable to figure out how the tests are run - thus being unable to test my test. Sorry about that.

## Summary by CodeRabbit

* **New Features**
  * Added a "Forward to" mail filter action, enabling users to automatically forward emails matching filter rules to a specified recipient address.

* **Tests**
  * Added test fixtures to validate forward mail filter action functionality and Sieve script generation.

### DIFF
--- a/lib/Service/MailFilter/FilterBuilder.php
+++ b/lib/Service/MailFilter/FilterBuilder.php
@@ -90,6 +90,10 @@ class FilterBuilder {
 				if ($action['type'] === 'stop') {
 					$actions[] = 'stop;';
 				}
+				if ($action['type'] === 'forward') {
+					$recipient = SieveUtils::escapeString($action['recipient']);
+					$actions[] = sprintf('redirect "%s";', $recipient);
+				}
 			}
 
 			if (count($tests) > 1) {
--- a/src/components/mailFilter/ActionForward.vue
+++ b/src/components/mailFilter/ActionForward.vue
@@ -1,0 +1,53 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<NcTextField
+		:required="true"
+		:model-value="recipient"
+		:label-outside="true"
+		:placeholder="t('mail', 'Enter recipient')"
+		@update:value="onInput" />
+</template>
+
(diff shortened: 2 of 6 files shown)


ChristophWurst:
Very nice. Thank you! We will check it out soon

ChristophWurst:
@coderabbitai review

src/components/mailFilter/ActionForward.vue
```diff
@@ -0,0 +1,53 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
```
kesselb, line 3: The spdx as html comment is fine
Coderabbit wrong is

src/components/mailFilter/ActionForward.vue
```diff
@@ -0,0 +1,53 @@
...
+-->
+<template>
+	<NcTextField
```
kesselb, line 6: Could you try to use a https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=ncinputfield with type = email? That should give us a bit of validation

tests/data/mail-filter/builder9.json
```diff
@@ -0,0 +1,23 @@
...
+				],
+				"field": "to"
+			}
+		],
+		"actions": [
+			{
+				"type": "forward",
+				"flag": "alice@example.org"
+			}
+		],
```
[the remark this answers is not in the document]
kesselb: Valid finding

lib/Service/MailFilter/FilterBuilder.php
```diff
@@ -90,6 +90,10 @@ public function buildSieveScript(array $filters, string $untouchedScript): strin
 				if ($action['type'] === 'stop') {
 					$actions[] = 'stop;';
 				}
+				if ($action['type'] === 'forward') {
+					$recipient = SieveUtils::escapeString($action['recipient']);
```
kesselb, line 94: I think we should check if recipient is !empty. On empty just go ahead.

kesselb (commented):
Thanks a lot

Issue #7142, opened 2022-09-02
Add image support to the text editor: Use editor mode for signature editor


kesselb:
### Is your feature request related to a problem? Please describe.

Follow up https://github.com/nextcloud/mail/pull/6935

TextEditor for signature does not respect the configured editor mode and always uses richtext mode. 

### Describe the solution you'd like

Set editor mode for the signature text editor. 
Warn user when changing from html to text (loosing the format/images).

kesselb:
Fixed by https://github.com/nextcloud/mail/pull/13388

Issue #4861, opened 2021-03-30
Extra newline after "--" delimiter in signatures


thstyl2000:
### Expected behavior
The signature should start directly below "--"

### Actual behavior
A new line is added after the double dash delimiter, making the signature look a bit off....

### Mail app

**Mail app version:** 1.9.4

GretaD:
Yeah, it also acts very weird when you try to select smth on it.


![Kazam-screencast-00014](https://user-images.githubusercontent.com/12728974/113028607-bea68d80-918b-11eb-83f4-d4f8b54308c8.gif)

kesselb:
For master: https://github.com/nextcloud/mail/blob/ae3646ac5aab93a6d22210941f43bab61ba6f983/src/ckeditor/signature/InsertSignatureCommand.js#L59

thstyl2000:
Ok, that's funny, did someone do that on purpose...? :-)

kesselb:
I dunno. It was always like that.

Fregf:
Also the signature delimiter ideally should be "-- ", with an extra space behind the dashes:
https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter

gnilebein:
However, the problem only occurs in the “Compose Email” mode. In the sent email, the blank line is no longer there.

PR #13098, opened 2026-06-17
fix(signature): insert "-- " separator for plain text only

### LINKED ISSUE #3457: Signature direct below the message and without --
I use the signature as salutation (eg "Greetings, Name") . 
So for me it would be better to have it direct below my message and without the "--"

joeldj-nl:
Fixes https://github.com/nextcloud/mail/issues/3457

## Summary

The signature separator (`--`) was previously always inserted above account signatures, with no way for users to disable it.

This PR adds a per-account setting:

> Add a separator line (`--`) above the signature

The setting is enabled by default to preserve the existing behavior for current users.

## Changes

* Add a per-account `signatureAboveSeparator` setting.
* Store the setting in the database.
* Expose the setting through the account controller.
* Include the setting in the user/account migration.
* Thread the setting through the composer signature command.
* Keep the existing behavior as the default by enabling the separator unless explicitly disabled.

## Testing

Tested successfully in a fresh Nextcloud development installation.

Manual test performed:

* Started a fresh Nextcloud dev container with PHP 8.3.
* Enabled the Mail app from this branch.
* Added/configured a mail account.
* Verified that the separator line is shown by default.
* Disabled the new account setting.
* Verified that the `--` separator is no longer inserted above the signature.
* Re-enabled the setting.
* Verified that the separator is inserted again.

## Notes

This keeps backward compatibility by defaulting the new option to enabled.

## 🤖 AI (if applicable)

* [x] The content of this PR was partly or fully generated using AI

screenshot-signature-separator-setting


ChristophWurst (approved):
Thank you for the contribution

Code looks good. Test coverage is appreciated

ChristophWurst:
DCO is failing. Please sign off your commits. By signing off you guarantee that the contribution is compatible with the project's license.

kesselb:
Thanks for your pull request. 

Thunderbird has a hidden flag to disable the separator too, so it probably makes sense to have it.

That said, I would like to double-check if you are aware that this separator, while not strictly standardized (apart from being mentioned at https://www.rfc-editor.org/info/rfc3676/#section-4.3), is widely used to distinguish content from a signature. Many clients provide an option to split or hide on it. Removing the separator therefore reduces interoperability.

Is there a particular reason why you need that switch?

joeldj-nl:
> Thanks for your pull request.
> 
> Thunderbird has a hidden flag to disable the separator too, so it probably makes sense to have it.
> 
> That said, I would like to double-check if you are aware that this separator, while not strictly standardized (apart from being mentioned at https://www.rfc-editor.org/info/rfc3676/#section-4.3), is widely used to distinguish content from a signature. Many clients provide an option to split or hide on it. Removing the separator therefore reduces interoperability.
> 
> Is there a particular reason why you need that switch?

Thanks for the clarification. Yes, I’m aware that the `-- ` separator is a long-standing convention and that some clients use it to detect, hide or strip signatures in replies.

The intention of this PR is not to change the default behavior. The new setting defaults to enabled, so existing interoperability is preserved unless a user explicitly opts out.

The reason for the switch is that the separator is not desirable for every account or signature style. For modern/HTML or business signatures, the visible `--` line can look odd to recipients, and currently users who do not want it have to manually remove it every time they compose a message.

Since Thunderbird also provides a way to suppress the separator, I think it makes sense for Nextcloud Mail to expose this as an account-level preference as well: keep the interoperable behavior by default, but allow users to disable it when they intentionally prefer not to send the visible separator.

joeldj-nl:
This PR resolves https://github.com/nextcloud/mail/issues/3457 and is also related to https://github.com/nextcloud/mail/issues/4861.

ChristophWurst:
And how about only using the separator for plain text emails? Then we do follow the convention for traditional emails but skip it for HTML.

joeldj-nl:
That might be a good idea. I’ll open a separate pull request to fix https://github.com/nextcloud/mail/issues/7142, since signatures first need a way to switch between the plain text and rich text editors.

Done in https://github.com/nextcloud/mail/pull/13141. Should I now open a new PR with changes to only include the separator for plain text emails, or what would be the recommended approach?

joeldj-nl:
@ChristophWurst let me know how you want it: an option to turn the separator off, or only showing the separator in plain text mode.

ChristophWurst:
Show it only for plain mode.

joeldj-nl:
Done @ChristophWurst

ChristophWurst (approved):
Tested & works

Please squash into one commit and change it to a `fix` commit. IMO this is no longer a "feature".

ChristophWurst:
Renamed the PR title to what the commit message should be

joeldj-nl:
> Tested & works
> 
> Please squash into one commit and change it to a `fix` commit. IMO this is no longer a "feature".

@ChristophWurst ✅

ChristophWurst (approved):

PR #13099, opened 2026-06-17
fix: save correctly any match when closing search modal

### LINKED ISSUE #13068: Search box finds nothing unless I clear every field except "body"
### Steps to reproduce

If I type a word in the search box above the message list, I get zero results,
even when I'm staring at a message that clearly contains that word in its body.

The only way I found to make it work: open the advanced search panel, delete the
word from Subject, From and To by hand, and leave it only in Body. Then the
message shows up. So the search box is basically unusable as-is.

I have "Enable mail body search" turned on for the account, and my IMAP server
(Dovecot with flatcurve) does have a working full-text index. `doveadm search`
on the server finds the message by body just fine, so this is on the Mail app side.

### Steps to reproduce

1. Enable "Enable mail body search" on an account.
2. Open a folder.
3. Search for a word that's in the body of a message but not in its subject,
   sender or recipient.
4. Nothing comes up.
5. Open advanced search, remove the word from Subject / From / To, keep it only
   in Body, search again -> the message is found.

### Expected behavior

Typing one word in the search box should find messages that have that word in any
field (subject, sender, recipient or body). A plain OR.

### Actual behavior

The search box puts the same word into Subject, From, To AND Body at once, and then
combines them with AND. So it ends up asking for a message where the word is in the
sender AND the recipient AND (subject OR body) all at the same time, which never
happens for a normal email.

### Mail app version

5.9.2

### Nextcloud version

33.0.3

### Mailserver or service

IMAP: Dovecot (Mailcow) + flatcurve FTS (mailcow)

### Operating system

AIO

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Where it comes from (I went digging)

- The quick search fills every field with the same query: `searchInSubject = query`,
  `searchInFrom = [{email: query}]`, `searchInTo = [{email: query}]`,
  `searchInMessageBody = searchBody ? query : null`.
- The default match mode is `allof` (AND): `SearchQuery.php` has
  `private string $match = 'allof';`, and the frontend `resetFilter()` sets
  `this.match = "allof"`.
- In `MessageMapper::findIdsByQuery()`, with `allof`, From and To each become their
  own `andWhere(...)`, while Subject and Body are OR'd together. Net result is
  `From~term AND To~term AND (Subject~term OR Body~term)`.

Switching the single-term search to `anyof` would put all four into the OR group and
fix it. `allof` only makes sense in the advanced panel when you fill different fields
with different values.

GretaD:
fixes #13068 

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -928,9 +928,10 @@ class MessageMapper extends QBMapper {
 		}
 		// createParameter
 		if ($uids !== null) {
-			// In the case of body+subject search we need a combination of both results,
-			// thus the orWhere in every other case andWhere should do the job.
-			if (!empty($query->getSubjects())) {
+			// In anyof mode or when subjects are also searched, body results must be
+			// OR'd with other criteria so a match in any one field is sufficient.
+			// In allof mode without subjects, AND is correct: all criteria must match.
+			if (!empty($query->getSubjects()) || $query->getMatch() === 'anyof') {
 				$textOrs[] = $qb->expr()->in('m.uid', $qb->createParameter('uids'));
 			} else {
 				$select->andWhere(
--- a/lib/Service/Search/MailSearch.php
+++ b/lib/Service/Search/MailSearch.php
@@ -150,7 +150,7 @@ class MailSearch implements IMailSearch {
 	 * @throws ServiceException
 	 */
 	private function getIdsLocally(Account $account, Mailbox $mailbox, SearchQuery $query, string $sortOrder, ?int $limit): array {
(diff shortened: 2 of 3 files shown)


ChristophWurst:
Can you please explain what the fix does and why?

GretaD:
> Can you please explain what the fix does and why?

The initial proposal didnt work well. I pushed a different approach.

The problem is that the body was not populated from the simple search box, and now it is.

ChristophWurst:
I don't get it :/

PR #13100, opened 2026-06-17
fix(sieve): fix handling of empty existing sieve scripts

hazveg:
This patch implements a check to more cleanly handle the case where the user's existing Sieve script is empty.
This occurs in the following two cases (from what I've tested):

- User cleared the contents in the Sieve script editor manually
- The Sieve script files on the server were deleted otherwise

This issue was first encountered when I implemented the redirect action in #13095 (See blank line at start of built script).

blank line uptop

I must presume this wasn't an issue before now because every other implemented action in the Filter UI also needed to import something, somehow causing that empty string in `$lines` to disappear. Debugging via Exceptions shows that an empty script file leads to a blank string at the start of `$lines`:

> Before

empty string

> After

after

### DIFF
--- a/lib/Service/MailFilter/FilterBuilder.php
+++ b/lib/Service/MailFilter/FilterBuilder.php
@@ -121,15 +121,17 @@ class FilterBuilder {
 			$lines[] = self::SEPARATOR;
 		}
 
-		/*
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Thank you for the fix! I'll let Sieve engineer @kesselb handle this one :)

PR #13110, opened 2026-06-18
fix(delegation): show delegates' display names instead of usernames

hamza221:
In the account delegation modal list users by their display names instead of usernames 

### DIFF
--- a/lib/Service/DelegationService.php
+++ b/lib/Service/DelegationService.php
@@ -62,7 +62,11 @@ class DelegationService {
 	}
 
 	public function findDelegatedToUsersForAccount(int $accountId): array {
-		return $this->delegationMapper->findDelegatedToUsers($accountId);
+		return array_map(function (Delegation $delegation) {
+			$displayName = $this->userManager->get($delegation->getUserId())?->getDisplayName();
+			$delegation->setDisplayName($displayName);
+			return $delegation;
+		}, $this->delegationMapper->findDelegatedToUsers($accountId));
 	}
 
 	public function unDelegate(Account $account, string $userId, string $currentUserId): void {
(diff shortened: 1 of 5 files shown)


lib/Service/DelegationService.php
```diff
@@ -62,7 +62,11 @@ public function delegate(Account $account, string $userId, string $currentUserId
 	}
 
 	public function findDelegatedToUsersForAccount(int $accountId): array {
-		return $this->delegationMapper->findDelegatedToUsers($accountId);
+		return array_map(function (Delegation $delegation) {
```
ChristophWurst, line 65: JSON encoding in the service layer smells like a bodge

and that the return type allows this just makes it clear that a stricter type is necessary

ChristophWurst (commented):
Code smell

lib/Service/DelegationService.php
hamza221: add display name in the controller then ?

ChristophWurst: Then you are moving logic out of the service.

How about a transient property on the Delegation class? Or a new class that covers delegation and the property?

src/components/DelegationModal.vue
```diff
@@ -260,8 +260,8 @@ export default {
 			this.delegating = true
 			try {
 				const delegation = await delegate(this.account.accountId, this.selectedUser.id)
-				this.delegates.push(delegation)
-				showSuccess(t('mail', 'Delegated access to {userId}', { userId: this.selectedUser.id }))
+				this.delegates.push({ ...delegation, displayName: this.selectedUser.displayName })
```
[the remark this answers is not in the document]
hamza221: POST returns the fallback (aka userid)

ChristophWurst (approved):

PR #13131, opened 2026-06-19
perf(imap): batch body-part fetches in getBodyStructureData

hamza221:
## Before
getBodyStructureData benchmark: 5 UID(s) in 2327.36 ms
getBodyStructureData benchmark: 20 UID(s) in 8091.76 ms
getBodyStructureData benchmark: 8 UID(s) in 4658.03 ms
getBodyStructureData benchmark: 20 UID(s) in 8009.11 ms
getBodyStructureData benchmark: 37 UID(s) in 13710.65 ms

## After
getBodyStructureData benchmark: 5 UID(s) in 3285.25 ms
getBodyStructureData benchmark: 20 UID(s) in 1237.34 ms
getBodyStructureData benchmark: 22 UID(s) in 1126.85 ms
getBodyStructureData benchmark: 20 UID(s) in 1421.08 ms
getBodyStructureData benchmark: 37 UID(s) in 1878.88 ms

Assisted-by: Claude:claude-opus-4-8[1m]

## 🤖 AI (if applicable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how email previews are generated for messages with HTML or plain-text bodies.
  * Handles missing or changed body parts more safely, avoiding incomplete preview content.
  * Better detects when a message mentions the recipient and preserves that in the preview.

* **Performance**
  * Reduced repeated message body lookups by batching body-part retrievals, which should speed up loading previews for multiple emails.

### DIFF
--- a/lib/IMAP/MessageMapper.php
+++ b/lib/IMAP/MessageMapper.php
@@ -890,15 +890,14 @@ class MessageMapper {
 		$structures = $client->fetch($mailbox, $structureQuery, [
 			'ids' => new Horde_Imap_Client_Ids($uids),
 		]);
-		return array_map(function (Horde_Imap_Client_Data_Fetch $fetchData) use ($mailbox, $client, $emailAddress) {
-			$hasAttachments = false;
-			$text = '';
-			$isImipMessage = false;
-			$isEncrypted = false;
 
-			if ($this->smimeService->isEncrypted($fetchData)) {
-				$isEncrypted = true;
-			}
+		$perMessage = [];
+		$bodyPartIds = [];
+		foreach ($structures as $fetchData) {
+			/** @var Horde_Imap_Client_Data_Fetch $fetchData */
+			$hasAttachments = false;
+			$isImipMessage = false;
+			$isEncrypted = $this->smimeService->isEncrypted($fetchData);
 
 			$structure = $fetchData->getStructure();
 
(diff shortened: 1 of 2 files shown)


ChristophWurst:
So it gets slower?

hamza221:
> So it gets slower?

Not really 3/5 times it was faster. other factors might be involved in why it's slower 2/5. I'm not sure how to verify. 
Reusing the same fetch makes sense to me in this case. But I'll leave the decision to approve/ close to the maintainers

ChristophWurst:
@coderabbitai review and validate the optimization

ChristophWurst (dismissed):
Rabbit comments might be worth checking

kesselb:
Asked Fable for a review,  used 70% of my 5h limit :see_no_evil: 

I did not validate the findings.

---

The PR replaces the N+1 pattern in MessageMapper::getBodyStructureData (one body-part fetch per message) with a single combined fetch of the union of all text/html part IDs across the batch. The refactoring preserves correctness — each message still reads only its own part IDs, the result stays UID-keyed for PreviewEnhancer, and the null-part fallback works because Horde's Fetch_Results::offsetGet returns null for missing UIDs. But the shape of the batching has a real flaw that verification confirmed, and it explains the author's own benchmark regression.

Findings (most severe first)

1. lib/IMAP/MessageMapper.php:949 — CONFIRMED — The union fetch requests every collected part ID from every UID, and MIME part IDs are message-relative, so colliding IDs download entire foreign attachments.
In multipart/alternative the HTML body is section "2"; in multipart/mixed section "2" is typically the first attachment. Once any message in the batch has its HTML at "2", the fetch issues BODY.PEEK[2] for all UIDs and every mixed message's attachment (a 25 MB PDF, say) is transferred in full and discarded — verified against Horde's Socket.php FETCH construction: a query's criteria apply to every UID, with no per-UID scoping possible in one fetch() call. Messages with no text/html body at all (S/MIME, calendar-only) also get all union parts fetched, where main skipped them entirely. This — not measurement noise — is almost certainly why the 5-UID benchmark got slower (2327 ms → 3285 ms). Fix at the right altitude: group UIDs by their (textBodyId, htmlBodyId) tuple and issue one fetch per group. Real mailboxes have few distinct tuples, so you keep the round-trip win (O(distinct structures) instead of O(N)) with zero cross-structure overfetch — strictly better than CodeRabbit's suggested small-batch threshold fallback, which only treats the symptom.

2. lib/IMAP/MessageMapper.php:949 — CONFIRMED — All bodies of the whole batch are buffered in memory at once, and the production batch size is unbounded.
PreviewEnhancementProcessingJob → PreprocessingService::process → Db\MessageMapper::getUnanalyzed (lib/Db/MessageMapper.php:1585) has no LIMIT — only mailbox IDs are chunked, not messages — so a fresh or busy account within the 14-day window can push thousands of UIDs into one giant FETCH whose entire response is held in $parts (amplified by finding 1's foreign-attachment payloads). Main already fetched all structures at once, but bodies are the dominant payload and were previously fetched one message at a time. The same file already has the precedent: findByIds wraps its fetch in chunk_uid_sequence() (lib/IMAP/MessageMapper.php:297). Chunk the UIDs here the same way.

3. tests/Unit/IMAP/MessageMapperTest.php:776 — CONFIRMED — The new test can't catch a wrong batched query, and the modified fallback branches are untested.
All three messages get identical single-part structures and the fetch mock returns canned results with no argument assertions, so the test only proves "fetch is called twice" — it would pass even if the union contained wrong part IDs or the second fetch targeted wrong UIDs. The rewritten fallback paths (both-body-ids-null at line 966, $part === null at line 974) are exercised by no test, which the repo's AGENTS.md rule ("Every changed or added code segment must be covered by unit tests") requires. Note CodeRabbit's withConsecutive suggestion is stale advice: it's deprecated in the repo's PHPUnit 9.6 and removed in 10 — capture the queries with willReturnCallback instead, and use two messages with differing structures.

4. lib/IMAP/MessageMapper.php:894-972 — CONFIRMED — The two-pass rewrite carries redundant state that can be deleted.
$perMessage copies fetchData/structure although $structures is already UID-keyed and safely re-iterable; $bodyPartIds plus the array_keys replay loop re-implements dedup Horde_Imap_Client_Fetch_Query already does (bodyPart() keys by MIME id, and the query is Countable, so count($partsQuery) > 0 guards the fetch); the $parts !== null arm at line 972 is unreachable (line 966 guarantees a non-null body ID → non-empty query → $parts assigned); and $text = '' is a constant to inline. Building $partsQuery directly in pass 1 and iterating $structures in pass 2 deletes ~25 lines. (Note: the grouping fix from finding 1 will reshape this code anyway — fold these in then.)

On the PR discussion

- ChristophWurst's "So it gets slower?" — the small-batch slowdown is structural, not noise: it's the foreign-part overfetch in finding 1. Signature-grouped fetches should make all batch sizes faster, resolving the open question properly rather than accepting a trade-off.
- CodeRabbit's empty("0") inline comment — verified pre-existing: the !empty($htmlBody)/!empty($textBody) checks are unchanged context lines identical on main. Not a regression of this PR; fix separately if at all.
- CodeRabbit's "logic is sound" conclusion — its per-message correctness trace agrees with mine, but it missed that colliding part IDs pull full attachment bodies and that the background-job batch is unbounded (findings 1–2), which are the substantive blockers.

My overall take: the direction is right and the correctness of the per-message logic holds, but I'd rework the fetch to group-by-body-part-signature plus chunking before merge — despite the existing approval, findings 1 and 2 are real production risks on the background-job path.

PR #13140, opened 2026-06-22
fix: use From address for reply-to-sender-only on mailing list emails

### LINKED ISSUE #11089: Reply to sender only, when the email is received from a mailing list doesnt work properly
### Steps to reproduce

1. open a group Email thread with at least one reply, e.g. Email subject "[Company] Welcome!"
2. select one email reply from the thread
3. click on the single arrow icon "reply to sender only"
4. in the email draft that will pop-up, the email distribution list "[email@lists.com] will appear as a recipient

### Expected behavior

individual email of the selected sender should appear in the as a recipient

### Actual behavior

the mailing list is the recipient

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

GretaD:
fixes: #11089 
## 🤖 AI (if applicable)

### DIFF
--- a/src/store/mainStore/actions.js
+++ b/src/store/mainStore/actions.js
@@ -523,7 +523,13 @@ export default function mainStoreActions() {
 					if (reply.mode === 'reply') {
 						logger.debug('Show simple reply composer', { reply })
 						const account = this.getAccount(reply.data.accountId)
-						let to = original.replyTo !== undefined ? original.replyTo : reply.data.from
+						// For mailing list emails, "Reply to sender" must use From because
+						// Reply-To points to the list address, not the original sender.
+						// For regular emails, honor Reply-To if the sender set one.
+						const isMailingList = !!(original.unsubscribeUrl || original.unsubscribeMailto)
+						let to = (!isMailingList && original.replyTo !== undefined)
+							? original.replyTo
+							: reply.data.from
 						// Replying to a message we sent ourselves: follow up with the
 						// original recipient(s) instead of addressing ourselves.
 						const isOwnMessage = to.length > 0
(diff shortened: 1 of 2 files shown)


ChristophWurst (commented):
Does this break any of the existing flows? I'm skeptical because replyTo is now generally unused.

GretaD:
> Does this break any of the existing flows? I'm skeptical because replyTo is now generally unused.

I tested `reply all`, and its the same as on the main branch. I tested an email with a couple of recipients but without a mailing list, and it works well too.

ChristophWurst:
Test with an email that has an actual reply-to header. That's the case where it actually makes a difference.

GretaD:
> Test with an email that has an actual reply-to header. That's the case where it actually makes a difference.

all good

ChristophWurst (changes requested):
Tested and breaks the legit reply-to case of a multi recipient email.

How to test
1. Send an email to three recipients and use another reply-to address
2. Open that emails as recipient in Mail
3. Click *Reply to sender*

main: reply-to address is used ✔️ 
original sender is used ❌ 

The list condition needs a dedicated check. Only for lists it can be ambiguous if *Reply to sender* uses the reply-to (list) or original sender. Gnome Evolution shows a dialogue for this case.

ChristophWurst (approved):
nit: would be cool to have the three+ cases covered by unit tests

ChristophWurst (approved):
Looks good

AI attribution trailer missing

PR #13141, opened 2026-06-22
fix(settings): respect the writing mode in the signature editor

### LINKED ISSUE #7142: Add image support to the text editor: Use editor mode for signature editor

joeldj-nl:
Fixes https://github.com/nextcloud/mail/issues/7142

### Summary

The signature editor always used rich text mode, even when the account was configured to compose messages as plain text. As a result, signatures containing images or other HTML content could be exposed as raw HTML after switching to plain text mode.

### Changes

* Use the account’s configured writing mode for the signature editor.
* Re-encode the saved signature when switching between rich text and plain text modes.
* Warn before switching from rich text to plain text, since that conversion drops formatting, inline images, including content in the signature.

https://github.com/user-attachments/assets/1a631918-259a-415b-a71a-8e98b99d3ca9

### DIFF
--- a/src/components/EditorSettings.vue
+++ b/src/components/EditorSettings.vue
@@ -8,20 +8,24 @@
 		<p>
 			<input
 				id="plaintext"
-				v-model="mode"
+				ref="plaintext"
+				:name="radioGroupName"
 				type="radio"
 				class="radio"
-				value="plaintext">
-			<label :class="{ primary: mode === 'plaintext' }" for="plaintext">
+				:checked="account.editorMode === EDITOR_MODE_TEXT"
+				@change="selectMode(EDITOR_MODE_TEXT)">
+			<label :class="{ primary: account.editorMode === EDITOR_MODE_TEXT }" for="plaintext">
 				{{ t('mail', 'Plain text') }}
 			</label>
 			<input
 				id="richtext"
-				v-model="mode"
+				ref="richtext"
+				:name="radioGroupName"
 				type="radio"
 				class="radio"
-				value="richtext">
-			<label :class="{ primary: mode === 'richtext' }" for="richtext">
+				:checked="account.editorMode === EDITOR_MODE_HTML"
+				@change="selectMode(EDITOR_MODE_HTML)">
+			<label :class="{ primary: account.editorMode === EDITOR_MODE_HTML }" for="richtext">
 				{{ t('mail', 'Rich text') }}
 			</label>
 		</p>
@@ -31,6 +35,7 @@
 <script>
 import { mapStores } from 'pinia'
 import Logger from '../logger.js'
+import { EDITOR_MODE_HTML, EDITOR_MODE_TEXT } from '../store/constants.js'
 import useMainStore from '../store/mainStore.js'
 
 export default {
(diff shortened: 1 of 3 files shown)


ChristophWurst:
I wasn't familiar with https://github.com/nextcloud/mail/issues/7142 and I'm a bit unsure about the approach.

I checked Gmail as reference and they just have the rich signature editor like we currently have. The only thing they do better is when you actually compose an email the writing mode dictates if HTML is stripped to plain text for the signature.

cc @kesselb @nimishavijay

nimishavijay:
Makes sense for it to respect the writing mode in general, as the root problem is that people don't know what their signaure will look like in the end. if we are able to communicate this up front as much as possible it is a good thing. 

Additionally, since I think a lot of people don't even look at the "writing mode" section in the settings when they just want to add/edit their signature, I would add the option to enable formatting from the signature editor as well, in that case the writing mode switches to rich text automatically. 

And instead of a blocking modal, I would instead change the UI of the writing mode to include a subline explaining that formatting from signatures will also be removed. That way it's a bit less disruptive. @ChristophWurst What are your thoughts?

ChristophWurst:
> I would add the option to enable formatting from the signature editor as well, in that case the writing mode switches to rich text automatically.

Then we have two buttons to change the writing mode, no?

My idea with the HTML signature is that you can use it in any writing mode. Doesn't matter what you set as default. Doesn't matter what you set for an email.

If the switch is part of the sigature then we will probably have to strip HTML to plain when going to plain mode, and from that point you can not go back, nor will you have formatting in an HTML email.

ChristophWurst:
:ping_pong: @nimishavijay

nimishavijay:
I see! I understand. Your idea is that if we are keeping the rich text signature in the signature editor, then it is seamless to switch from plain text to rich text and vice versa without losing any contents during composing, is that correct @ChristophWurst ? I would agree. I would then add warnings (not necessarily in the form of a blocking modal). 

- [ ] A subline for each option in the writing mode section of the settings modal: 
>Plain text: Unformatted text only, signature will lose styling.
>Rich text: Supports formatting and images, including in the signature.
- [ ] A banner in the signature section of the settings modal using NcNoteCard, as an info card if rich text is selected and warning card if plain text is selected:
> Writing in plain text mode will remove the signature of its formatting

This removes any complicated logic of setting/unsetting writing mode and losing formatting on switching, what do you think @ChristophWurst @joeldj-nl

kesselb:
Sorry, I've overlooked this PR :see_no_evil: 

https://github.com/nextcloud/mail/pull/13388 changed the signature editor to respect the writing mode and improve the ux around signatures.

It works as Christoph suggested, the signature is always html. The available formatting options depend on the writing mode. We transparently change the composer to match the signature (e.g. the user uses an image in the signature, has plain text selected => composer is opened with html enabled).

joeldj-nl:
Thanks for the update! Looks like #13388 covers the original issue and implements the direction discussed here. I'll close this PR as superseded.

PR #13142, opened 2026-06-22
fix(message): render received external SVG images

joeldj-nl:
### Summary

Received emails containing external SVG images did not render. The message showed only the broken-image placeholder and the image's alt text or name. External images are fetched through the image proxy, which served every response as `application/octet-stream`. Browsers happily sniff raster formats (PNG/JPEG/GIF) out of an  tag regardless of content type, but they refuse to render SVG unless it is served as `image/svg+xml`, so SVG logos and graphics stayed blank.

### Changes

- Detect SVG markup in proxied responses and serve it with the `image/svg+xml` content type so browsers render it.
- Add an `SvgSanitizer` service that strips active content (scripts, event handlers, external/javascript: references) and rejects DOCTYPE/entity declarations, as defence in depth against XXE and stored XSS even though the  context does not execute scripts.

The first screenshot shows the issue; the second screenshot shows it has been resolved.

svg-not-rendering
svg-rendering-succeeded

## Release Notes

* **New Features**
  * SVG content is automatically detected during proxying and sanitized before being served.
  * Sanitized SVGs are returned with the correct `image/svg+xml` content type; unsafe/invalid content is replaced with a blocked-image fallback.

* **Bug Fixes**
  * Improved handling of proxy responses that don’t return valid image payloads, reducing failures and ensuring consistent output.

* **Tests**
  * Added unit tests for SVG sanitization security rules and updated controller tests to verify sanitized SVG delivery.


ChristophWurst:
Thanks for the contribution

ChristophWurst:
@coderabbitai review and scan for new vulnerabilities

ChristophWurst:
Yes

joeldj-nl:
The recommended changes have been implemented.

ChristophWurst:
@coderabbitai re-review

lib/Controller/ProxyController.php
```diff
@@ -112,6 +116,22 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
...
+
+		// Browsers sniff raster image formats in <img> tags, but they refuse to
+		// render SVG unless it is served with the image/svg+xml content type.
+		// Detect and sanitise SVG markup so external SVG logos are displayed
+		// instead of staying blank. Sanitising also strips any active content in
+		// case the response is fetched through a direct (non-<img>) navigation.
+		if ($this->svgSanitizer->looksLikeSvg($content)) {
+			$sanitized = $this->svgSanitizer->sanitize($content);
+			if ($sanitized === '') {
+				$content = (string)file_get_contents(__DIR__ . '/../../img/blocked-image.png');
```
ChristophWurst: file_get_contents returns string|false. casting to string feels problematic. handle the (very unlikely) error by throwing a ServiceException when it returns false

ChristophWurst (commented):
👍 otherwise

thanks

ChristophWurst:
The branch needs an update because there is a conflict

lib/Controller/ProxyController.php
joeldj-nl: Should be resolved now.

joeldj-nl:
> The branch needs an update because there is a conflict

Updated!

lib/Controller/ProxyController.php
```diff
@@ -97,15 +100,43 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
...
 		} catch (LocalServerException $e) {
 			$this->logger->warning('Prevented image proxy access to forbidden URL', [
 				'blockedUrl' => $src,
 				'exception' => $e,
 			]);
-			$content = file_get_contents(__DIR__ . '/../../img/blocked-image.png');
+			$content = $this->getBlockedImage();
+		}
+
+		$content = (string)$content;
```
ChristophWurst: why?

joeldj-nl: Do you mean why $content is explicitly cast to a string here?

ChristophWurst: Yes

joeldj-nl: getBody() returns string|resource, while the SVG sanitizer expects a string, so the cast is there to make the types line up.

joeldj-nl: If you have a different approach in mind, let me know.

ChristophWurst: But that won't fly. You can not just cast a resource to a string.

joeldj-nl: You’re right I’ll handle the resource explicitly instead

joeldj-nl: Should be resolved now. Thanks for reviewing!

ChristophWurst:
@coderabbitai review again

lib/Controller/ProxyController.php
```diff
@@ -95,17 +100,46 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
 		try {
 			$response = $client->get($src);
 			$content = $response->getBody();
+			if (!is_string($content)) {
+				throw new ServiceException('Could not read proxied image');
+			}
```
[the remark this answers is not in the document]
joeldj-nl: Fixed in new commit

lib/Controller/ProxyController.php
```diff
@@ -95,17 +100,46 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
...
-			$content = file_get_contents(__DIR__ . '/../../img/blocked-image.png');
+			$content = $this->getBlockedImage();
+		}
+
+		// Browsers sniff raster image formats in <img> tags, but they refuse to
+		// render SVG unless it is served with the image/svg+xml content type.
+		// Detect and sanitise SVG markup so external SVG logos are displayed
+		// instead of staying blank. Sanitising also strips any active content in
+		// case the response is fetched through a direct (non-<img>) navigation.
+		if ($this->svgSanitizer->looksLikeSvg($content)) {
```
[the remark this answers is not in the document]
joeldj-nl: Fixed in new commit

lib/Controller/ProxyController.php
```diff
@@ -95,17 +100,46 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
...
+	/**
+	 * @throws ServiceException
+	 */
+	private function getBlockedImage(): string {
+		$content = file_get_contents(__DIR__ . '/../../img/blocked-image.png');
+		if ($content === false) {
+			throw new ServiceException('Could not read blocked image');
+		}
+
+		return $content;
```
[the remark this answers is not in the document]
joeldj-nl: This fallback handling is specific to the controller and does not represent reusable business logic. Moving such a small private helper into a service would introduce unnecessary indirection, so I’d prefer to keep it within the controller.

joeldj-nl:
@ChristophWurst the new findings by CodeRabbitAI have been resolved.

kesselb:
Thanks for your pr :+1: 

We reject SVG's with XLS/Transform in server (c.f. https://github.com/nextcloud/server/pull/59835) and should apply the same hardening here.

joeldj-nl:
@kesselb thanks for your input. This should be fixed now.

lib/Service/SvgSanitizer.php
```diff
@@ -0,0 +1,163 @@
...
+		libxml_clear_errors();
+		libxml_use_internal_errors($previousErrors);
+
+		if (!$loaded || $dom->documentElement === null) {
+			return '';
+		}
+
+		$xpath = new DOMXPath($dom);
+
+		// Remove processing instructions (e.g. <?xml-stylesheet type="text/xsl"?>).
```
kesselb: The `?>` ends php and make the linter fails.

lib/Service/SvgSanitizer.php
```diff
@@ -0,0 +1,163 @@
...
+	/** Reject payloads larger than this to prevent DoS via oversized documents. */
+	private const MAX_SVG_BYTES = 2 * 1024 * 1024;
+
+	/**
+	 * @param string $svg The raw (decoded) SVG markup
+	 * @return string The sanitised markup, or an empty string if it cannot be
+	 *                parsed safely
+	 */
+	public function sanitize(string $svg): string {
+		if (trim($svg) === '' || strlen($svg) > self::MAX_SVG_BYTES) {
```
kesselb, line 52: Please take a look at https://github.com/nextcloud/server/pull/62162 and check if we have those cases covered.

lib/Controller/ProxyController.php
```diff
@@ -95,17 +101,49 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
 		try {
 			$response = $client->get($src);
 			$content = $response->getBody();
+			if (is_resource($content)) {
+				$content = stream_get_contents($content);
+				if ($content === false) {
+					$content = $this->getBlockedImage();
+				}
+			}
```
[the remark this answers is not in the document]
kesselb: Ignore

lib/Controller/ProxyController.php
```diff
@@ -95,17 +101,49 @@ public function proxy(string $src, ?int $id, ?string $hmac): Response {
 		try {
 			$response = $client->get($src);
 			$content = $response->getBody();
```
kesselb, line 102: $content is only a resource with a call like `$client->get($src, ['stream' => true])`. The actual return for the current request is string.

Please drop is_resource fallback. 

To make static code analysis happy add:

```php
// Not a stream request, so the body is a string, never a resource.
assert(is_string($content));
```

joeldj-nl:
Added new commits to resolve the issues.

PR #13146, opened 2026-06-23
Do not re-insert signature when opening a draft

jancborchardt:
When you have a signature enabled and open a draft, that signature gets inserted. This can even result in multiple insertions of the signature if you open the draft multiple times. When opening a draft, the signature should never be inserted, as this is already done at initial composition.

Initial composer (with signature removed) | Before when reopening draft | Bonus: Multiple draft reopening | After when reopening draft
-|-|-|-
Screenshot From 2026-06-23 12-17-50|Screenshot From 2026-06-23 11-57-49|Screenshot From 2026-06-23 12-10-10|Screenshot From 2026-06-23 11-57-04

## 🤖 AI (if applicable)

Assisted-by: ClaudeCode:claude-opus-4-8

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signature handling when editing draft messages. Signatures no longer auto-insert when reopening a draft, reducing signature duplication issues. Manually changing the signature selection will still insert the new signature correctly. Draft editing now has separate behavior from new message composition, where signatures continue to auto-insert as before.

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -705,6 +705,11 @@ export default {
 			required: true,
 		},
 
+		isDraft: {
+			type: Boolean,
+			default: false,
+		},
+
 		requestMdn: {
 			type: Boolean,
 			default: false,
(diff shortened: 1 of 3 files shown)


jancborchardt:
(Rebased, all tests were passing before also.)

ChristophWurst:
@coderabbitai review and find out if there is existing logic to handle this, because this was fixed before

ChristophWurst (approved):

kesselb:
image

I'm seeing in CKEditor inspector, that detecting the signature is broken. Hence the existing logic, to replace the signature block, does not work anymore.

We should fix the detection of the signature. 
Not workaround the actual problem.

ChristophWurst:
Thanks. I knew we had something like that. Rabbit just didn't find it.

PR #13152, opened 2026-06-23
feat: warn users when sending email with empty to field

### LINKED ISSUE #13094: Warn user when sending email with empty "To" field

While RFC 5322 allows empty **To** headers, some major mail providers (for example GMX and WEB.DE) still reject such messages due to local policy restrictions. Users currently receive SMTP errors after sending (e.g. https://postmaster.gmx.net/en/case?c=r0735, https://postmaster.web.de/en/case?c=r0735), which leads to undelivered emails and poor user experience.

PS: I double checked the emails beeing sent multiple times, the email nextcloud mail have sent was RFC compliant, the To field was given and it was perfectly empty (no comma, no colon, nothing). his is not a nextcloud mail bug about beeing incompliant.

This change proposes adding warning banner / notification when a user attempts to send an email without any visible recipients at all (i.e., recipients only in Cc/Bcc or none in visible recipient fields). When the user clicks the "Send" button and the To field and the Cc field is empty, the UI should show warning dialog or inline confirmation message, e.g.: "The email has no visible recipients. Some mail providers may reject this message. Do you want to continue?"

Optional further enhancement: 
- Add a "Don't show again" option (per account)
- Auto-suggest: Add current sender email address to the To field

steven-mpawulo:
Show a warning when the To field is empty during the sending of an email.

fixes #13094.


src/components/Composer.vue
```diff
@@ -1549,6 +1539,11 @@ export default {
 		},
 
 		async onSend() {
+
+			if (this.selectTo.length === 0) {
+				showWarning(t('mail', 'The email has no visible ‘To’ recipients. Some mail providers may reject this message.'))
```
ChristophWurst: The type of *warning* discussed in https://github.com/nextcloud/mail/issues/13094 is a conditionally rendered text underneath the recipient input.

It would be something like the *helper text* of NcTextField https://nextcloud-vue-components.netlify.app/#/Components/NcFields?id=nctextfield but for the NcSelect component. NcSelect might not have this feature at the moment.

ChristophWurst (changes requested):
Thank you for looking into this, @steven-mpawulo!

src/components/Composer.vue
steven-mpawulo: Oh, I see. I initially thought you meant displaying an actual warning toast.
For this case, would it be okay to use a simple `<p>` element below the NcSelect as helper text and style it accordingly, since NcSelect doesn't seem to support helper text like NcTextField does at the moment.

ChristophWurst: Yes, sounds good as workaround.

I've created https://github.com/nextcloud-libraries/nextcloud-vue/issues/8669 to track the missing feature. If we add this later we can switch.

steven-mpawulo: Awesome.
Let me update

src/components/Composer.vue
```diff
@@ -98,6 +98,9 @@
 					</template>
 				</NcSelect>
 			</div>
+			<div v-if="displayMissingToWarning">
+				<p class="to-warning">{{ t('mail', 'The email has no visible ‘To’ recipients. Some mail providers may reject this message') }}</p>
```
ChristophWurst: We use the BEM structure for classes. Perhaps this could be a `composer-fields__helper-text`

src/components/Composer.vue
```diff
@@ -1549,6 +1544,11 @@ export default {
 		},
 
 		async onSend() {
+
+			if (this.selectTo.length === 0) {
+				this.displayMissingToWarning = true
+			}
```
ChristophWurst: This happens too late. The warning text should be shown every time there are no direct recipients.

ChristophWurst (changes requested):

src/components/Composer.vue
steven-mpawulo: Noted.
Let me update this.

src/components/Composer.vue
steven-mpawulo: I also considered this but initially ignored it. I’ll update it now.

src/components/Composer.vue
```diff
@@ -1029,6 +1031,7 @@ export default {
 		},
 
 		selectTo(val) {
+			this.displayMissingToWarning = false
```
ChristophWurst: Triggers

> [Vue warn]: Computed property "displayMissingToWarning" was assigned to but it has no setter.

Please test your changes

src/components/Composer.vue
```diff
@@ -1549,6 +1547,7 @@ export default {
 		},
 
 		async onSend() {
+
```
ChristophWurst: revert this empty line

src/components/Composer.vue
```diff
@@ -1001,6 +999,10 @@ export default {
 			return this.mainStore.getSharedTextBlocks()?.map((textBlock) => ({ title: textBlock.title, content: textBlock.content }))
 				.concat(this.mainStore.getMyTextBlocks().map((textBlock) => ({ title: textBlock.title, content: textBlock.content })))
 		},
+
+		displayMissingToWarning() {
+			return this.selectTo.length === 0
+		},
```
ChristophWurst, line 1010: UX-wise I don't like that this shows instantly when you open the composer.

Image

The *To* field is auto-focused. How about we only start checking for selectTo.length after the first blur (lost focus) of *To*?

cc @nimishavijay

ChristophWurst (changes requested):

src/components/Composer.vue
```diff
@@ -98,6 +98,9 @@
 					</template>
 				</NcSelect>
 			</div>
+			<div v-if="displayMissingToWarning">
+				<p class="composer-fields__helper-text">{{ t('mail', 'The email has no visible ‘To’ recipients. Some mail providers may reject this message') }}</p>
```
MundMBo:
```suggestion
				<p class="composer-fields__helper-text">{{ t('mail', 'The email has no visible ‘To’ recipients. Some mail providers may reject this message.') }}</p>
```

MundMBo: Also a full stop is missing at the end of the message.

src/components/Composer.vue
```diff
@@ -98,6 +98,11 @@
 					</template>
 				</NcSelect>
 			</div>
+			<div v-if="displayMissingToWarning">
+				<p class="composer-fields__helper-text">
+					{{ t('mail', 'The email has no visible ‘To’ recipients. Some mail providers may reject this message.') }}
```
```suggestion
					{{ t('mail', 'The email has no \'To\' recipients. Some mail providers may reject it.') }}
```

src/components/Composer.vue
```diff
@@ -98,6 +98,11 @@
 					</template>
 				</NcSelect>
 			</div>
+			<div v-if="displayMissingToWarning">
+				<p class="composer-fields__helper-text">
```
ChristophWurst: you can put the v-if on this element and avoid the wrapping div

ChristophWurst (commented):
Tested & works smoothly 😎 

Thank you

ChristophWurst:
Please squash into a single commit and rebase to solve the composer conflict

src/components/Composer.vue
nimishavijay: Would use "message" instead of "email" here as we do in other places.

Edit:
And we can combine the 2 sentences to one:
"Messages with no 'To' recipients may be rejected by some mail providers."

src/components/Composer.vue
nimishavijay: Sounds good, and I would also change the text color to text-maxcontrast and reduce the size to the smaller text. It is not a blocking error and if we make this message appear on the first blur it will be very visible. 

image

ChristophWurst: Image

Adjusting it

ChristophWurst (approved):
Cleaned-up, :shipit:

PR #13168, opened 2026-06-25
feat(ui): Add option to disable auto mark as read

### LINKED ISSUE #13061: Option to fully disable "auto mark as read" when opening a message

Currently, when opening an unread email in Nextcloud Mail, the message is automatically marked as read. The existing implementation (resulting from #3723) only introduces a delay before the message is marked as read, but it does not provide a way to fully disable the auto-mark-as-read behavior.

For some users and workflows, this is problematic:

    Users often open emails just to skim or preview them, but want to come back later and process them properly. With auto-mark-as-read, these messages lose their "unread" status and can easily be forgotten.
    Other mail clients (Thunderbird, Outlook, etc.) offer this as a standard option.
    The current delay-based solution is a workaround, not a real opt-out, also the delay time is pretty short.

This was previously discussed in:

    #3785 (closed as duplicate)
    #3723 (resolved with a delay, not a full opt-out)

As suggested by @ChristophWurst in #3785, this is being opened as a new feature request.

Add a new option in the Nextcloud Mail settings, for example:

    ☐ Automatically mark messages as read when opened

    Default: enabled (current behavior preserved)
    When disabled: opening a message does not change its read/unread status. The user must mark messages as read manually (e.g., via the existing "Mark as read" action, keyboard shortcut, or by marking the conversation).

_No response_

### Additional context

Related issues:

    #3723 – Do not auto read mark (resolved via delay)
    #3785 – Options to not auto mark as read when opening a message (closed as duplicate)

SoleroTG:
Fixes https://github.com/nextcloud/mail/issues/13061

## Summary
 This Pull Request introduces a user preference option to fully disable the "auto mark as read" feature when opening a message in Nextcloud Mail (#13061). When disabled, viewing/skimming messages does not automatically change their status to read, making "mark as read" an explicitly manual action.

## Key Changes
1. **Settings Switch:** Added an option switch labeled `"Automatically mark messages as read when opened"` in the App Settings Menu under the "Messages" section.
2. **Preference Storage:** Registered `auto-mark-as-read` user preference key, integrated it into the backend PHP initial state payload (`PageController`), and synced it with the frontend Vue Pinia store state on application bootstrap (`init.js`).
3. **Seen Timer Check:** Modified the timing logic in `ThreadEnvelope.vue` (`fetchMessage()`) to retrieve and check the value of `auto-mark-as-read` (defaulting to `true`) before starting the 2000ms delay timer that toggles the envelope to `seen`.

## How to Test
1. Compile front-end changes (`npm run dev` or `npm run watch`).
2. Open Nextcloud Mail.
3. Open **Mail settings** -> **Messages**.
4. Disable the toggle: **"Automatically mark messages as read when opened"**.
5. Select/click an unread message in the envelope list.
6. Verify that:
   - The message loads and displays its body.
   - The message remains unread (no timer is started, no `seen` flag is sent to the server).
   - Reloading the page persists the disabled toggle and state correctly.
7. Enable the toggle and verify that opening an unread message automatically marks it as read after the standard 2-second delay.

## Checklist
- [x] Code passes linting (`npm run lint` and `composer run cs:check`)
- [x] Unit tests pass successfully (`npm run test:unit`)
- [x] Signed-off commits (DCO check compliant)

## 🤖 AI (if applicable)

CC: @GretaD (as you [previously expressed interest](https://github.com/nextcloud/mail/issues/13061#issuecomment-4716598486) in this feature/discussion)

## Summary by CodeRabbit

* **New Features**
  * Added a setting to control when messages are automatically marked as read: immediately, after 3 or 30 seconds, or manually.
  * Applied the selected preference when viewing messages.

* **Bug Fixes**
  * Added error handling when saving the automatic read-status preference fails.

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -153,6 +153,13 @@
 					</NcFormBoxSwitch>
 				</NcFormBox>
 
+				<NcRadioGroup :model-value="autoMarkAsRead" :label="t('mail', 'Mark messages as read')" @update:modelValue="onToggleAutoMarkAsRead">
+					<NcRadioGroupButton :label="t('mail', 'Immediately')" value="0" :disabled="hasLoadingState('auto-mark-as-read')" />
+					<NcRadioGroupButton :label="n('mail', 'After %n second', 'After %n seconds', 3)" value="3000" :disabled="hasLoadingState('auto-mark-as-read')" />
+					<NcRadioGroupButton :label="n('mail', 'After %n second', 'After %n seconds', 30)" value="30000" :disabled="hasLoadingState('auto-mark-as-read')" />
+					<NcRadioGroupButton :label="t('mail', 'Manually')" value="-1" :disabled="hasLoadingState('auto-mark-as-read')" />
+				</NcRadioGroup>
+
 				<NcRadioGroup :model-value="useBottomReplies" :label="t('mail', 'Reply position')" @update:modelValue="onToggleButtonReplies">
 					<NcRadioGroupButton :label="t('mail', 'Top')" :value="false" :disabled="hasLoadingState('reply-mode')" />
 					<NcRadioGroupButton :label="t('mail', 'Bottom')" :value="true" :disabled="hasLoadingState('reply-mode')" />
@@ -449,6 +456,16 @@ export default {
 			},
 		},
 
+		autoMarkAsRead: {
+			get() {
+				return this.mainStore.getPreference('auto-mark-as-read', '3000')
+			},
+
+			set(value) {
+				this.onToggleAutoMarkAsRead(value)
+			},
+		},
+
 		useExternalAvatars: {
 			get() {
 				return this.mainStore.getPreference('external-avatars', 'true') === 'true'
@@ -672,6 +689,22 @@ export default {
 				Logger.error('could not save preferences', { error })
 			} finally {
 				this.setLoadingState('search-priority-body', false)
+			}
+		},
+
+		async onToggleAutoMarkAsRead(value) {
+			this.setLoadingState('auto-mark-as-read', true)
+
+			try {
+				await this.mainStore.savePreference({
+					key: 'auto-mark-as-read',
+					value: String(value),
+				})
+			} catch (error) {
+				Logger.error('could not save preferences', { error })
+				showError(t('mail', 'Could not update preference'))
+			} finally {
+				this.setLoadingState('auto-mark-as-read', false)
 			}
 		},
 
--- a/src/components/ThreadEnvelope.vue
+++ b/src/components/ThreadEnvelope.vue
@@ -853,12 +853,17 @@ export default {
 					clearTimeout(loadingTimeout)
 				}
 
-				if (!this.envelope.flags.seen && this.hasSeenAcl) {
-					logger.info('Starting timer to mark message as seen/read')
+				const autoMarkReadSetting = this.mainStore.getPreference('auto-mark-as-read', '3000')
+				const delay = parseInt(autoMarkReadSetting, 10)
+
+				if (!this.envelope.flags.seen && this.hasSeenAcl && delay >= 0) {
+					logger.info(`Starting timer (${delay}ms) to mark message as seen/read`)
 					this.seenTimer = setTimeout(() => {
-						this.mainStore.toggleEnvelopeSeen({ envelope: this.envelope })
+						if (!this.envelope.flags.seen) {
+							this.mainStore.toggleEnvelopeSeen({ envelope: this.envelope })
+						}
 						this.seenTimer = undefined
-					}, 2000)
+					}, delay)
 				}
 
 				if (this.message.hasHtmlBody) {
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -240,6 +240,7 @@ class PageController extends Controller {
 			'sort-favorites' => $this->preferences->getPreference($this->userId, 'sort-favorites', 'false'),
 			'index-context-chat' => $this->contextChatSettingsService->isIndexingEnabled($this->userId) ? 'true' : 'false',
 			'compact-mode' => $this->preferences->getPreference($this->userId, 'compact-mode', 'false'),
+			'auto-mark-as-read' => $this->preferences->getPreference($this->userId, 'auto-mark-as-read', '3000'),
 		]);
 		$this->initialStateService->provideInitialState(
 			'prefill_displayName',
--- a/src/init.js
+++ b/src/init.js
@@ -98,6 +98,10 @@ export default function initAfterAppCreation() {
 		key: 'compact-mode',
 		value: preferences['compact-mode'],
 	})
+	mainStore.savePreferenceMutation({
+		key: 'auto-mark-as-read',
+		value: preferences['auto-mark-as-read'],
+	})
 
 	mainStore.setQuickActions(loadState('mail', 'quick-actions', []))
 
--- a/tests/Unit/Controller/PageControllerTest.php
+++ b/tests/Unit/Controller/PageControllerTest.php
@@ -188,7 +188,7 @@ class PageControllerTest extends TestCase {
 		$account1 = $this->createMock(Account::class);
 		$account2 = $this->createMock(Account::class);
 		$mailbox = $this->createStub(Mailbox::class);
-		$this->preferences->expects($this->exactly(14))
+		$this->preferences->expects($this->exactly(15))
 			->method('getPreference')
 			->willReturnMap([
 				[$this->userId, 'account-settings', '[]', json_encode([])],
@@ -205,6 +205,7 @@ class PageControllerTest extends TestCase {
 				[$this->userId, 'smime-sign-aliases', '[]', '[]'],
 				[$this->userId, 'sort-favorites', 'false', 'false'],
 				[$this->userId, 'compact-mode', 'false', 'false'],
+				[$this->userId, 'auto-mark-as-read', '3000', '3000'],
 			]);
 		$this->accountService->expects($this->once())
 			->method('findByUserId')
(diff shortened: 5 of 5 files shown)


ChristophWurst:
Thank you for the contribution!

ChristophWurst:
@nimishavijay small decision needed. Our friends have a setting for this as well but with a few options

image

(excuse the German UI, you get the idea :speak_no_evil:)

Should we keep it binary (on vs off) or also offer a few options like *off*, *instant*, *3 seconds* and *30 seconds*?

ChristophWurst:
:ping_pong: @nimishavijay

nimishavijay:
Initially I thought adding a timer would be overengineering but I checked outlook and they also seem to have a timer setting ([support link](https://support.microsoft.com/en-us/outlook/mark-a-message-as-read-or-unread-in-outlook)), so I'm guessing it would be a gap if we don't have it. Your timer options also sound great @ChristophWurst so we could go with that :)

SoleroTG:
Thanks @ChristophWurst and @nimishavijay! That makes total sense.

I have updated the PR to replace the binary toggle with the proposed radio group options (Immediately, After 3 seconds, After 30 seconds, and Manually). Backward compatibility for existing preference values has also been included.

It looks like this now:

image

src/components/AppSettingsMenu.vue
```diff
@@ -151,6 +150,13 @@
 					</NcFormBoxSwitch>
 				</NcFormBox>
 
+				<NcRadioGroup :model-value="autoMarkAsRead" :label="t('mail', 'Mark messages as read')" @update:modelValue="onToggleAutoMarkAsRead">
+					<NcRadioGroupButton :label="t('mail', 'Immediately')" value="0" />
+					<NcRadioGroupButton :label="t('mail', 'After 3 seconds')" value="3000" />
+					<NcRadioGroupButton :label="t('mail', 'After 30 seconds')" value="30000" />
```
ChristophWurst: nitpick: use the `n` function to translate plural with a placeholder for the number of seconds. That means there will only be one string to translate and should we ever change a value or have another option then the same strings can be re-used.

SoleroTG: @ChristophWurst, Didn't know that. Fixed it.

src/components/AppSettingsMenu.vue
```diff
@@ -97,7 +97,6 @@
 
 				<NcRadioGroup :model-value="sortOrder" :label="t('mail', 'Sorting')" @update:modelValue="onSortByDate">
 					<NcRadioGroupButton :label="t('mail', 'Newest first')" value="newest" />
-					<NcRadioGroupButton :label="t('mail', 'Oldest first')" value="oldest" />
```
kesselb: 🤔

src/components/AppSettingsMenu.vue
```diff
@@ -151,6 +150,13 @@
 					</NcFormBoxSwitch>
 				</NcFormBox>
 
+				<NcRadioGroup :model-value="autoMarkAsRead" :label="t('mail', 'Mark messages as read')" @update:modelValue="onToggleAutoMarkAsRead">
+					<NcRadioGroupButton :label="t('mail', 'Immediately')" value="0" />
+					<NcRadioGroupButton :label="n('mail', 'After {count} second', 'After {count} seconds', 3, { count: 3 })" value="3000" />
+					<NcRadioGroupButton :label="n('mail', 'After {count} second', 'After {count} seconds', 30, { count: 30 })" value="30000" />
+					<NcRadioGroupButton :label="t('mail', 'Manually')" value="-1" />
```
kesselb:
```suggestion
					<NcRadioGroupButton :label="t('mail', 'Immediately')" value="0" :disabled="loadingAutoMarkAsRead" />
					<NcRadioGroupButton :label="n('mail', 'After {count} second', 'After {count} seconds', 3, { count: 3 })" value="3000" :disabled="loadingAutoMarkAsRead" />
					<NcRadioGroupButton :label="n('mail', 'After {count} second', 'After {count} seconds', 30, { count: 30 })" value="30000" :disabled="loadingAutoMarkAsRead" />
					<NcRadioGroupButton :label="t('mail', 'Manually')" value="-1" :disabled="loadingAutoMarkAsRead" />
```

SoleroTG: Thanks @kesselb, sorry for the accidental deletion of the "Oldest first" radio button. 

I've restored that option and also added `:disabled="loadingAutoMarkAsRead"` to all radio buttons in the new auto-mark-as-read group as suggested.

ChristophWurst:
Conflicts

SoleroTG:
@ChristophWurst, Rebased onto current main and resolved the conflicts.

src/components/AppSettingsMenu.vue
```diff
@@ -153,6 +153,13 @@
 					</NcFormBoxSwitch>
 				</NcFormBox>
 
+				<NcRadioGroup :model-value="autoMarkAsRead" :label="t('mail', 'Mark messages as read')" @update:modelValue="onToggleAutoMarkAsRead">
+					<NcRadioGroupButton :label="t('mail', 'Immediately')" value="0" :disabled="hasLoadingState('auto-mark-as-read')" />
+					<NcRadioGroupButton :label="n('mail', 'After {count} second', 'After {count} seconds', 3, { count: 3 })" value="3000" :disabled="hasLoadingState('auto-mark-as-read')" />
+					<NcRadioGroupButton :label="n('mail', 'After {count} second', 'After {count} seconds', 30, { count: 30 })" value="30000" :disabled="hasLoadingState('auto-mark-as-read')" />
```
ChristophWurst: Use %n

https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#correct-plurals

ChristophWurst (changes requested):

src/components/AppSettingsMenu.vue
SoleroTG: Thanks @ChristophWurst, Fixed it.

src/components/AppSettingsMenu.vue
```diff
@@ -449,6 +456,23 @@ export default {
...
+		autoMarkAsRead: {
+			get() {
+				const value = this.mainStore.getPreference('auto-mark-as-read', '3000')
+				if (value === 'true') {
+					return '3000'
+				}
+				if (value === 'false') {
+					return '-1'
+				}
+				return value
```
ChristophWurst: true/false are unexpected except for *your* dev env. this doesn't need handling.

src/components/ThreadEnvelope.vue
```diff
@@ -853,12 +853,25 @@ export default {
 					clearTimeout(loadingTimeout)
 				}
 
-				if (!this.envelope.flags.seen && this.hasSeenAcl) {
-					logger.info('Starting timer to mark message as seen/read')
-					this.seenTimer = setTimeout(() => {
+				let autoMarkReadSetting = this.mainStore.getPreference('auto-mark-as-read', '3000')
+				if (autoMarkReadSetting === 'true') {
```
ChristophWurst: same here. remnant of an earlier development version.

ChristophWurst (dismissed):

src/components/ThreadEnvelope.vue
SoleroTG: Thanks @ChristophWurst, I remove the legacy true/false check.

ChristophWurst:
@coderabbitai review

src/components/ThreadEnvelope.vue
```diff
@@ -853,12 +853,20 @@ export default {
...
 						this.mainStore.toggleEnvelopeSeen({ envelope: this.envelope })
-						this.seenTimer = undefined
-					}, 2000)
+					} else {
+						logger.info(`Starting timer (${delay}ms) to mark message as seen/read`)
+						this.seenTimer = setTimeout(() => {
+							this.mainStore.toggleEnvelopeSeen({ envelope: this.envelope })
+							this.seenTimer = undefined
+						}, delay)
+					}
```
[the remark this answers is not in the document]
ChristophWurst: Checking if the setting changes is overkill.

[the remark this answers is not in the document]
ChristophWurst: But the manual intervention is an interesting case. If that happened we should not toggle. Example:

1. User set 30s timeout
2. User opens email, reads it quickly and 10s decides to mark it read
3. Use navigates to next email

Email will be marked unread 20s later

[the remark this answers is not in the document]
SoleroTG: @ChristophWurst, Good Point. Added a check `if (!this.envelope.flags.seen)` inside the timer callback to prevent toggling the message back to unread if it was already manually marked as read before the timer expires.

src/components/ThreadEnvelope.vue
```diff
@@ -853,12 +853,22 @@ export default {
...
 				}
 
-				if (!this.envelope.flags.seen && this.hasSeenAcl) {
-					logger.info('Starting timer to mark message as seen/read')
-					this.seenTimer = setTimeout(() => {
+				const autoMarkReadSetting = this.mainStore.getPreference('auto-mark-as-read', '3000')
+				const delay = parseInt(autoMarkReadSetting, 10)
+
+				if (!this.envelope.flags.seen && this.hasSeenAcl && delay >= 0) {
+					if (delay === 0) {
```
kesselb: Could we unify the delay branching here and always go through the same `setTimeout` path? Using a delay of `0` should preserve the behavior while reducing the number of code paths we need to maintain. 

That would make the read-timing logic a bit easier to follow and lowers the chance that one branch drifts from the other over time.

kesselb (approved):

src/components/ThreadEnvelope.vue
SoleroTG: Thanks @kesselb! Unified the branching to always use the `setTimeout` path for `delay >= 0`.

src/components/ThreadEnvelope.vue
```diff
@@ -853,12 +853,17 @@ export default {
 					clearTimeout(loadingTimeout)
 				}
 
-				if (!this.envelope.flags.seen && this.hasSeenAcl) {
-					logger.info('Starting timer to mark message as seen/read')
+				const autoMarkReadSetting = this.mainStore.getPreference('auto-mark-as-read', '3000')
+				const delay = parseInt(autoMarkReadSetting, 10)
+
+				if (!this.envelope.flags.seen && this.hasSeenAcl && delay >= 0) {
+					logger.info(`Starting timer (${delay}ms) to mark message as seen/read`)
```
kesselb, line 860:
```suggestion
					logger.debug(`Starting timer (${delay}ms) to mark message as seen/read`)
```

kesselb (approved):

PR #13169, opened 2026-06-26
fix: remove beforeunload and resize listener when closing composer

### LINKED ISSUE #13031: Mail app shows a confirmation popup before leaving - beforeunload listener not removed when closing composer
### Steps to reproduce

1.Open Nextcloud Mail
2.Start composing a new message
3.Close the composer without sending - draft is saved automatically
4.Navigate to another Nextcloud app 
5.Confirmation popup is displayed

### Expected behavior

The browser should not prompt for confirmation when navigating away after the composer is closed and draft message has been saved

### Actual behavior

Browser shows a popup asking for confirmation before leaving, even after the composer is closed.

Image

Reproducible on many instances

### Mail app version

5.9 - behavior started at around 5.7

### Nextcloud version

33.0.5

### Mailserver or service

Any

### Operating system

Any

### PHP engine version

Any

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

Using `beforeunload` event listener monitor in the browser console, you can see that the listener registered in NewMessageModal.vue (`created()` hook) is never removed after the composer is closed.

Patch used in browser console to identify : 

```
const _orig = window.addEventListener.bind(window);
const _origRemove = window.removeEventListener.bind(window);

window.addEventListener = function(type, fn, opts) {
  if (type === 'beforeunload') {
    console.trace('beforeunload ADDED', fn);
  }
  return _orig(type, fn, opts);
};

window.removeEventListener = function(type, fn, opts) {
  if (type === 'beforeunload') {
    console.trace('beforeunload REMOVED', fn);
  }
  return _origRemove(type, fn, opts);
};
```

The `removeEventListener` call in `beforeUnmount()` is never reached, suggesting the component is not being fully unmounted when the composer is dismissed.

steven-mpawulo:
Removed beforeunload and resize listeners

Fixes #13031

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -631,6 +631,9 @@ export default {
 			this.mainStore.setComposerIndicatorDisabledMutation(true)
 			await this.onMinimize()
 
+			window.removeEventListener('beforeunload', this.onBeforeUnload)
+			window.removeEventListener('resize', this.checkScreenSize)
+
 			// End the session only if all unsaved changes have been saved
 			if (this.canSaveDraft && ((this.changed && this.draftSaved) || !this.changed)) {
 				logger.debug('Closing composer session due to close button click')
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -631,9 +631,6 @@ export default {
 			this.mainStore.setComposerIndicatorDisabledMutation(true)
 			await this.onMinimize()
 
-			window.removeEventListener('beforeunload', this.onBeforeUnload)
-			window.removeEventListener('resize', this.checkScreenSize)
-
 			// End the session only if all unsaved changes have been saved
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Thanks for the fix

ChristophWurst (changes requested):
I think this needs more refinement. The listeners have to be put back when you start composing the next email.
And saving any draft should be awaited, so the lines have to be pushed down after the `if`.

steven-mpawulo:
Ohh,
Let me adjust.

ChristophWurst (commented):
Tested and after sending an email I still can not reload without the warning.

Which use case does this fix address?

steven-mpawulo:
This PR covers the case from #13031 - closing the composer without sending (draft autosaved) where the beforeunload listener wasn't being removed on close.

ChristophWurst:
After https://github.com/nextcloud/mail/pull/13303 the handling should already improve. Should be tested again.

PR #13172, opened 2026-06-28
Fix search provider replace URL with URL with deeplink

taiebot:
I do not know which other apps use this search provider, but this is used by the tables app. This change replace the URL with a deeplink URL (there was a todo)

This has been tested with the tables app, this will allow sharing links between multiple mailboxes.

see the screenshots  below
image

image

## 🤖 AI (if applicable)

@SoleroTG for review

### DIFF
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -116,10 +116,9 @@ class Provider implements IProvider {
 					]),
 					$message->getSubject(),
 					$subline,
-					$this->urlGenerator->linkToRouteAbsolute('mail.page.thread', [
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Makes sense

nitpick: you can inline the url generation

ChristophWurst:
You have to fix the commit. Message should be `fix(search): use deep link`. it needs the AI trailer and a sign-off.

ChristophWurst:
Please rebase so there is only the `fix` commit. Thank you

ChristophWurst (approved):

PR #13173, opened 2026-06-28
fix: validating email addresses from Contacts

### LINKED ISSUE #7121: Mail app doesn't check if the email format of a contact is valid
### Steps to reproduce

1. Create a contact with a domain as email address. E.g. Contact Joe has as email address `joe.com`
2. Go to Mail app and send new email
3. In recipients set the contact Joe as receiver
4. Send email
5. Now the email is stuck in Outbox, and it never alerts you about the problem 

### Expected behavior

Mail app should check if the email has the proper format and alert the user otherwise.

### Actual behavior

The email is never sent and the user isn't prompt with any useful alert

### Mail app version

1.13.8

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

madbob:
The Contacts app now validates email addresses on his own (https://github.com/nextcloud/contacts/pull/4380) so #7121 may be considered obsolete.

Yet, an extra check in `ContactsIntegration` may prevent broken imports or other issues on the Contacts side.

Fixes #7121

### DIFF
--- a/lib/Service/ContactsIntegration.php
+++ b/lib/Service/ContactsIntegration.php
@@ -66,6 +66,12 @@ class ContactsIntegration {
 				if ($e === '') {
 					continue;
 				}
+
+				$mightBeValidEmail = filter_var($e, FILTER_VALIDATE_EMAIL);
+				if ($mightBeValidEmail === false) {
+					continue;
+				}
+
 				$receivers[] = [
 					'id' => $id,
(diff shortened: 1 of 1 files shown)


kesselb:
Thanks for your PR :+1: 

Could you please fix the dco check?

kesselb (approved):

PR #13174, opened 2026-06-28
feat: capturing Ctrl+S key on email composer to save draft

### LINKED ISSUE #634: Ctrl-S for triggering draft save
Currently, pressing Ctrl-S when you are writing an email brings up the »Save Document as« dialog which is not really helpful.

Instead, it should force trigger a draft save if you are in any of the compose or reply fields. (Maybe we should even catch it in all cases, so that the dialog never pops up.)

madbob:
Fixes #634

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -264,6 +264,7 @@
 				@input="onEditorInput"
 				@ready="onEditorReady"
 				@mention="handleMention"
+				@save="onEditorSave"
 				@submit="onEditorSubmit" />
 			<MailvelopeEditor
 				v-else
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Thanks a lot! This will be reviewed and tested soon!

GretaD (approved):
I tested it and works well. Thanks a lot for your contribution.

src/components/Composer.vue
```diff
@@ -1345,6 +1346,10 @@ export default {
 			}
 		},
 
+		onEditorSave(editor) {
+			this.saveDraft();
+		},
```
GretaD, line 1351: please fix the lint here and ready to go :)

GretaD:
hi, thank you for fixing the lint, please rebase and squash the 3 commits.

PR #13175, opened 2026-06-29
feat: ability to cancel a draft

### LINKED ISSUE #2492: Add 'cancel' button to discard draft

madbob:
The ingenuous solution here provided is to explicitely trigger the draft-saving function immediately when the `NewMessageModal` is opened, so to move the whole component in the canonical "draft saved" state (displaying the explicit trash button).

Alternative solutions would imply a more complex integration with `MessageService.deleteMessage()` (instead of `DraftService.deleteDraft()`) when required.

Fixes #2492

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -269,6 +269,10 @@ export default {
 		this.updateCookedComposerData()
 		await this.openModalSize()
 		window.addEventListener('resize', this.checkScreenSize)
+
+		if (this.composerData.draftId !== undefined) {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Thank you for the PR

ChristophWurst:
I'm hesitant about the mechanics of the solution because opening an existing draft and not making any changes should probably not even start the whole draft complexity.

So yeah, I think we need the alternative.

madbob:
Superseeded by #13284

PR #13176, opened 2026-06-29
feat(perf):  remove window event listeners on component unmount

GretaD:
Changes in this PR:

Envelope.vue — deduplicate store getter calls + resize listener leak       
                                                                             
  getEnvelopeTags(this.data.databaseId) was being called separately inside   
  both _isImportant_ and _tags_ on every render. Since _Envelope_ is rendered once 
  per message in a list, this doubles the store lookups per envelope per     
  render cycle. The new _envelopeAllTags_ computed property calls it once and Vue caches the result, both isImportant and tags then read from the cache and only recompute when the store data actually changes. The resize listener had the same leak as above  added in mounted() but never removed. The new beforeUnmount hook fixes that.                      

-----

EnvelopeList.vue—remove unnecessary array copy + deduplicate tag lookups 
                                                                             
  [...this.envelopes] was creating a full array copy on every computed       
  evaluation with no reason to, the spread was defensive but the computed   
  property is read-only. With large mailboxes, this is a pointless allocation on every render.      

-------

isAtLeastOneSelectedImportant and isAtLeastOneSelectedUnimportant were each
  iterating _selectedEnvelopes_ and calling getEnvelopeTags per envelope, so N selected envelopes meant 2N store calls. The new selectedEnvelopeTags computed property fetches tags once for all selected envelopes, and both computed properties read from that cache.   

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -553,11 +553,15 @@ export default {
 
 	mounted() {
 		this.sortOrder = this.mainStore.getPreference('sort-order', 'newest')
-		document.addEventListener.call(window, 'mailvelope', () => this.checkMailvelope())
+		window.addEventListener('mailvelope', this.checkMailvelope)
 		if (!this.mainStore.areTextBlocksFetched()) {
 			this.mainStore.fetchMyTextBlocks()
 			this.mainStore.fetchSharedTextBlocks()
 		}
+	},
+
+	beforeUnmount() {
+		window.removeEventListener('mailvelope', this.checkMailvelope)
 	},
 
 	updated() {
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -560,7 +560,7 @@ export default {
 		}
 	},
 
-	beforeUnmount() {
+	beforeDestroy() {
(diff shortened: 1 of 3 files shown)


src/components/AppSettingsMenu.vue
```diff
@@ -553,13 +553,17 @@ export default {
 
 	mounted() {
 		this.sortOrder = this.mainStore.getPreference('sort-order', 'newest')
-		document.addEventListener.call(window, 'mailvelope', () => this.checkMailvelope())
+		window.addEventListener('mailvelope', this.checkMailvelope)
```
ChristophWurst, line 556: I'd be worried that the method is invoked with the wrong context. Is Vue binding them automagically?

src/components/AppSettingsMenu.vue
```diff
@@ -553,13 +553,17 @@ export default {
...
 		if (!this.mainStore.areTextBlocksFetched()) {
 			this.mainStore.fetchMyTextBlocks()
 			this.mainStore.fetchSharedTextBlocks()
 		}
 	},
 
+	beforeUnmount() {
```
ChristophWurst: does this ever happen?

PR #13178, opened 2026-06-29
fix(sync): tolerate PERMANENTFLAGS IMAP errors

### LINKED ISSUE #12506: Error "Undefined array key "id""
### Steps to reproduce

1. Install Nextcloud Mail app in the newest version (Nextcloud 33 and Mail App v. 5.7.1)
2. Setup a 2FA App Password for an iCloud Mail account
3. Login with this App password in Nextcloud Mail

### Expected behavior

All mails are fetched from the iCloud Mail Servers and displayed within the Nextcloud Mail app

### Actual behavior

No emails are shown. Instead, you can see the following error hundreds of times within the logs:

`Undefined array key "id" at /var/www/html/nextcloud/apps/mail/vendor/bytestream/horde-imap-client/lib/Horde/Imap/Client/Socket.php#1007`

### Mail app version

5.7.1

### Nextcloud version

33.0.0

### Mailserver or service

Apple iCloud

### Operating system

Ubuntu 24.0.4 LTS

### PHP engine version

PHP 8.3

### Nextcloud memory caching

_No response_

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

_No response_

niklasarnitz:
## Summary

- Treat IMAP `STATUS PERMANENTFLAGS` failures as custom flags being unsupported instead of aborting account sync
- Log the server failure at debug level with mailbox context
- Add unit coverage for the fallback

## Motivation

Some IMAP servers can return an error for the `PERMANENTFLAGS` status query while regular mailbox sync otherwise works. On iCloud this causes account sync to abort with `Could not get message flag options from IMAP: IMAP error reported by server` and can also trigger the log spam reported in #12506.

If the server cannot report permanent flag support, Mail can safely continue with custom keyword sync disabled for that mailbox.

Fixes #12506

### DIFF
--- a/lib/Service/MailManager.php
+++ b/lib/Service/MailManager.php
@@ -759,11 +759,12 @@ class MailManager implements IMailManager {
 		try {
 			$capabilities = $client->status($mailbox, Horde_Imap_Client::STATUS_PERMFLAGS);
 		} catch (Horde_Imap_Client_Exception $e) {
-			throw new ServiceException(
-				'Could not get message flag options from IMAP: ' . $e->getMessage(),
-				$e->getCode(),
-				$e
-			);
+			$this->logger->debug('Could not get IMAP PERMANENTFLAGS support; treating custom flags as unsupported', [
+				'exception' => $e,
+				'mailbox' => $mailbox,
+			]);
+
+			return false;
 		}
 		return (is_array($capabilities) === true && array_key_exists('permflags', $capabilities) === true && in_array("\*", $capabilities['permflags'], true) === true);
 	}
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Thanks for the PR

lib/Service/MailManager.php
odzhychko, line 730: This function makes the same call and I would expect it to fail in the same way.
But I haven't tested it and am not sure what user facing issue it would cause.

kesselb:
Thanks a lot for your patch :+1: 

Custom flags are supported by iCloud, hence this function should always return true. Masking the exception leads to a wrong state. 

We documented our current findings about the root cause at https://github.com/nextcloud/mail/issues/13350. 

I'm closing the PR, as that workaround is not an option.

PR #13180, opened 2026-06-30
fix(sending): preserve Bcc header in Sent mailbox copy

ChristophWurst:
Fixes https://github.com/nextcloud/mail/issues/11772.

Horde_Mime_Mail::addHeader('Bcc', ...) silos the Bcc value into a separate $_bcc field instead of $_headers, so getRaw() never includes it. The raw message stored in the Sent mailbox therefore had no Bcc header, unlike Evolution, Thunderbird, and Horde IMP.

Mirror Horde IMP's approach: build Horde_Mime_Headers directly with Bcc included, strip it from a clone for SMTP delivery (RFC 5321), and pass the full headers to $mimePart->toString() for the Sent mailbox copy.

Assisted-by: Claude:claude-sonnet-4-6

## How to test

1. Compose an email
2. Enter a *To* recipient
3. Enter a *Bcc* recipient
4. Send
5. Look at sent email source

main: no *Bcc:* header
here: *Bcc* header

If you look at the email source as *To* recipient you will still not see the Bcc header.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

PR #13182, opened 2026-06-30
fix(drafts): preserve Bcc header when saving drafts

ChristophWurst:
Somewhat related to https://github.com/nextcloud/mail/pull/13180.

## How to test

1. Start a new email
2. Enter a bcc recipient
3. Save the draft
4. Open the draft mailbox
5. Open the new draft email

main: empty Bcc
here: restored Bcc

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/MailTransmission.php
+++ b/lib/Service/MailTransmission.php
@@ -14,7 +14,6 @@ use Horde_Imap_Client;
 use Horde_Imap_Client_Data_Fetch;
 use Horde_Imap_Client_Fetch_Query;
 use Horde_Imap_Client_Ids;
-use Horde_Mail_Rfc822_Address;
 use Horde_Mail_Rfc822_List;
 use Horde_Mail_Transport_Null;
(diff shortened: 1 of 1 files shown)


lib/Service/MailTransmission.php
```diff
@@ -372,6 +344,24 @@ public function saveDraft(NewMessageData $message, ?Message $previousDraft = nul
 		return [$account, $draftsMailbox, $newUid];
 	}
 
+	private function buildMimeHeaders(Address $from, AddressList $to, AddressList $cc, AddressList $bcc, ?string $subject): Horde_Mime_Headers {
```
madbob, line 369: This new function replaces entirely `buildHeaders()`, which is no longer used and can be suppressed.

ChristophWurst: Good catch! Missed because I opened this PR when https://github.com/nextcloud/mail/pull/13180 wasn't in either.

GretaD (approved):
tested and it works

PR #13183, opened 2026-06-30
fix(oauth): generate state token server-side before redirect

ChristophWurst:
Add a dedicated endpoint that produces an HMAC-signed state token binding the account ID and a timestamp to the current user session. The token expires after 10 minutes. Both Google and Microsoft OAuth flows use this token instead of constructing the state client-side.

Assisted-by: Claude:claude-sonnet-4-6

## 🤖 AI (if applicable)

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -476,6 +476,11 @@ return [
 			'verb' => 'GET',
 		],
 		[
+			'name' => 'oauth#generateState',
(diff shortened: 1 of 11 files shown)


miaulalala (approved):

kesselb (approved):

kesselb:
Failing test look related

PR #13184, opened 2026-06-30
fix: fixed "ORGANIZER" in ICS generated by "Reply with meeting" and delivered to DAV endpoint

### LINKED ISSUE #12213: "Reply with meeting" is not sending meeting invitations
### Steps to reproduce

1. Receive an email.
2. Create an event using "Reply with meeting"
3. Event is successfully created. However, no invitation is sent. Examining the event in Calendar shows "Invitation will be sent" and never changes even after editing the event and saving again in Calendar.

### Expected behavior

Reply with meeting should create the meeting and send out invitations to those that were included in the email being replied to.

### Actual behavior

Reply with meeting does NOT send the event invitation to the intended recipients. Examining the event in Calendar shows "Invitation will be sent".

Invitations for events created in Calendar are being sent immediately as expected. 

### Mail app version

5.6.5

### Nextcloud version

32.0.3

### Mailserver or service

Mailcow

### Operating system

Ubuntu 24.04

### PHP engine version

PHP 8.3

### Nextcloud memory caching

apcu and redis

### Web server

Apache (supported)

### Database

MySQL

### Additional info

_No response_

madbob:
Without a "mailto:" prefix in the ORGANIZER email property, the NextCloud server fails ([exactly here](https://github.com/nextcloud/3rdparty/blob/69b1534909f2abd81621d03e3660398743c38ae5/sabre/vobject/lib/ITip/Broker.php#L244)) to properly handle the ICS payload and generate invitations to attendees.

Fixes #12213

### DIFF
--- a/src/components/EventModal.vue
+++ b/src/components/EventModal.vue
@@ -314,7 +314,7 @@ export default {
 						event.addProperty(new AttendeeProperty('ATTENDEE', `mailto:${att}`))
 					})
 
-					event.addProperty(new AttendeeProperty('ORGANIZER', organizerEmail))
+					event.addProperty(new AttendeeProperty('ORGANIZER', `mailto:${organizerEmail}`))
 				}
 
 				for (const vObject of calendar.getVObjectIterator()) {


SebastianKrupinski (approved):
Thanks for the PR, that is the correct fix.

madbob:
I think one of the tests broke just for an internal issue.

Is it possible to re-run the pipeline again manually?

PR #13187, opened 2026-06-30
fix: while creating a new account, check for his existance

### LINKED ISSUE #12042: Don't allow duplicate account setup
### Steps to reproduce

1. Have an account setup 'example@domain.net'
2. Setup the same account again

### Expected behavior

Don't allow to setup the same account more than once 

### Actual behavior

Allowed

madbob:
Unsure if the duplication check should span across all existing accounts on the instance, or just those registered by the current user...

Fixes #12042


ChristophWurst:
Thanks for the PR

GretaD (approved):
it works, thank you 

Image

GretaD:
would it be possible for you to write a quick test about this check?

hamza221 (approved):

ChristophWurst:
What I'm missing

1. Same logic for `occ mail:account:create` or moved check
2. Strategy for existing accounts
3. (Eventually) a UNIQUE INDEX in the db because that is the only way to assure the constraint

madbob:
> 1. Same logic for `occ mail:account:create` or moved check
> 
I'm going to check.

> 2. Strategy for existing accounts
> 
In my opinion, this cannot be retroactive. What is done, is done.

> 3. (Eventually) a UNIQUE INDEX in the db because that is the only way to assure the constraint
> 
This implementation just checks that the current user do not register multiple times the same mail account, but no one specified what happens if different users register the same mail account. If allowed: it cannot be unique in the database. If not allowed: this implementation is wrong.

ChristophWurst:
> > 5\. (Eventually) a UNIQUE INDEX in the db because that is the only way to assure the constraint
> 
> This implementation just checks that the current user do not register multiple times the same mail account, but no one specified what happens if different users register the same mail account. If allowed: it cannot be unique in the database. If not allowed: this implementation is wrong.

The unique index should be on `(userid, email)` so multiple user can set up the same email (service account etc)

madbob:
I've added the check also for the CLI command.

What to do for the database? I've figured out how to create a migration, but if applied it will - of course - fail in instances where duplicated accounts already exist...

kesselb:
@madbob can you check the conflicts?

madbob:
@kesselb I can do that, but... What about backward compatibility with existing database, as pointed out by @ChristophWurst? Is it planned a follow up commit to manage that?

madbob:
While I fight with unit tests, I have one more question: is there any specific reason why `CreateImapAccount` do not use `SetupService::createNewAccount()`?

The latter also performs `testConnectivity()` and - more relevant - calls `TagMapper::createDefaultTags()`. So: accounts created by CLI have no populated tags. Is this an explicit choice?

GretaD:
> While I fight with unit tests, I have one more question: is there any specific reason why `CreateImapAccount` do not use `SetupService::createNewAccount()`?
> 
> The latter also performs `testConnectivity()` and - more relevant - calls `TagMapper::createDefaultTags()`. So: accounts created by CLI have no populated tags. Is this an explicit choice?

unfortunatelly, i dont know the answer to these questions - if its a specific reason or not, but can you please sign the commits so we can get this merged? Thanks again

madbob:
I've signed the last commit: sorry, I overlooked it...

My question about `CreateImapAccount` was mostly aimed to cut some - probably - redundant code from the command, and ultimately remove the need for extra test coverage.

GretaD:
please solve the conflits, and should be ready to merge :)

PR #13199, opened 2026-07-02
fix: whitespaces in visualization of plaintext email

### LINKED ISSUE #10914: Plain text mail spaces are not shown correctly
### Steps to reproduce

1. Send yourself a mail with simple plaintext content with subsequent spaces and indents - do not use Nextcloud Mail for it (see #10912), e.g.:
```
first   << more spaces
  second, indented
```
2. Open the mail. The spaces will be ignored just as they are in HTML source.

### Expected behavior

Spaces are visible, even if the plaintext is shown with non-monospace fonts as it currently is, the spaces should be preserved and shown. It should be possible to copy/paste them.

When I modified the displayed page with DevTools with `white-space` attribute, it looked much better for plain text content, so something in this effect would help a lot for plain text mails or plain text parts of multipart mails:
````
<div id="message-container" ... style="white-space: pre;">
````

### Actual behavior

White spaces are lost by the HTML rendering. Leading spaces are gone, multiple spaces are shrinked to one.

### Mail app version

4.2.6

### Nextcloud version

31.0.1

### Mailserver or service

Zimbra 8.x

### Operating system

Ubuntu 24.04.2 LTS

### PHP engine version

PHP 8.3

### Nextcloud memory caching

_No response_

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

_No response_

madbob:
The right CSS rule was already in place, but with the wrong CSS selector!

Fixes #10914

### DIFF
--- a/src/components/MessagePlainTextBody.vue
+++ b/src/components/MessagePlainTextBody.vue
@@ -106,7 +106,7 @@ export default {
 </style>
 
 <style lang="scss" scoped>
-.message-container,
+#message-container,
 .mail-signature {
 	white-space: pre-wrap;
 }


ChristophWurst (approved):
Thanks! Good catch

PR #13204, opened 2026-07-03
fix(unified-inbox): don't discard every account's envelopes if one account fails

### LINKED ISSUE #9072: *Other* section of unified inbox doesn't load for new accounts
### Steps to reproduce

1. Open the app
2. Add an account
3. Wait for everything to load, don't touch anything

Possibly relevant: I have an existing account. It's provisioned and doesn't work at the moment due to the mail server being down. If this is the reason for the error we have to improve the error handling.

### Expected behavior

Everything loads without issues

### Actual behavior

Priority inbox loads *Important* messages but *Other* loads forever. Console shows an error

image

> Unified list has no tail

The *Other* section always loads after the next page reload. This seems specific to the first time the new account is used for the unified/priority inbox.

### Mail app version

3.5

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

ktogias:
## Summary

`fetchEnvelopes()` and `fetchNextEnvelopes()` combine every constituent account's envelope fetch into the unified/priority inbox via `Promise.all()`, which rejects as soon as any single promise rejects. A single account that's temporarily unreachable (throttled provider, network hiccup, provisioned account with an unavailable server, etc.) was therefore enough to make the whole unified mailbox render nothing at all, discarding every other account's already-successful envelopes too.

This matches the failure shape described in #9072 ("Other section of unified inbox doesn't load for new accounts") — one account's IMAP server being unavailable breaking the combined view for every account.

## Fix

Catches each individual account's fetch before it reaches `Promise.all()`, so a failing account contributes an empty list instead of failing the whole batch. Same reasoning applied to the pagination path in `fetchNextEnvelopes()`.

## Test plan

- [x] Added a unit test with two accounts where one throws — confirms the working account's envelopes still render.
- [x] Verified the new test fails without the fix (reproduces the original bug) and passes with it.
- [x] `npx vitest --run src/tests/unit/store/actions.spec.js` — 24/24 passing.
- [x] `npx eslint src/store/mainStore/actions.js` — 0 errors (pre-existing unrelated jsdoc warnings only).

Fixes #9072

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified inboxes now continue displaying available messages when fetching mail from one account fails.
  * Pagination can proceed using successful mailbox results even if another mailbox encounters an error.

### DIFF
--- a/src/store/mainStore/actions.js
+++ b/src/store/mainStore/actions.js
@@ -722,6 +722,12 @@ export default function mainStoreActions() {
 				const mailbox = this.getMailbox(mailboxId)
 
 				if (mailbox.isUnified) {
+					// One account's fetch rejecting must not discard every
+					// other account's already-successful envelopes. Promise.all()
+					// rejects as soon as any single promise rejects, so a single
+					// slow/unreachable account used to make the whole unified
+					// mailbox render nothing at all instead of everything except
+					// that one account. See #9072.
 					const fetchIndividualLists = pipe(
 						map((mb) => this.fetchEnvelopes({
 							mailboxId: mb.databaseId,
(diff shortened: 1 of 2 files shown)


ktogias:

ChristophWurst:
@coderabbitai review

PR #13208, opened 2026-07-03
fix(search): don't block message listing on a mailbox sync lock

ktogias:
## Summary

`MailSearch::findMessages()` -- the method that serves a folder's message list -- checked `$mailbox->hasLocks(...)` and threw `MailboxLockedException` before doing anything else. But the rest of the method only ever reads already-cached messages from the local DB (`messageMapper->findByIds()` / `findIdsByQuery()`), with no live IMAP call at all outside an unrelated body-text-search path.

The sync lock exists to coordinate concurrent *writes* (two sync attempts racing on the same mailbox) -- it was never a correctness requirement for a *read*, just an overly cautious check that happened to also block them.

On the frontend, `Mailbox.vue`'s `loadEnvelopes()` responds to a `MailboxLockedError` by waiting 15s and recursively retrying -- with `Mailbox::LOCK_TIMEOUT` at 300s, that's up to ~20 silent retry cycles with the loading state up the whole time, for an operation that could have just returned the already-cached list immediately. This was especially visible for any account slow to sync (large mailbox, temporarily rate-limited provider, etc.): every folder's message list would refuse to load for minutes at a time even though the data was sitting right there in the DB the whole time.

## Fix

Removes the `hasLocks()` check from `findMessages()`. `isCached()` (a separate, legitimate check -- has this mailbox ever completed an initial sync) is unaffected and stays in place. Also removed the now-unused `MailboxLockedException` import (this was its only use in the file). Left the now-unused `$timeFactory` constructor dependency alone to keep this PR focused -- happy to clean that up too if preferred.

## Test plan

- [x] Updated the existing test that asserted the old locked-throws behavior (`testFindMessagesLocked`) to instead assert `findMessages()` succeeds normally despite a lock being present (`testFindMessagesIgnoresLock`).
- [x] `php -l` clean on both changed files.
- [x] Ran this exact change against a live production instance for several hours (an account with an intermittently unresponsive IMAP provider) -- every folder's message list now loads instantly from cache regardless of that account's lock/sync state.

### DIFF
--- a/lib/Db/Mailbox.php
+++ b/lib/Db/Mailbox.php
@@ -122,13 +122,13 @@ class Mailbox extends Entity implements JsonSerializable {
 	}
 
 	public function hasLocks(int $now): bool {
-		if ($this->getSyncNewLock() !== null || $this->getSyncNewLock() > ($now - self::LOCK_TIMEOUT)) {
+		if ($this->getSyncNewLock() !== null && $this->getSyncNewLock() > ($now - self::LOCK_TIMEOUT)) {
 			return true;
 		}
-		if ($this->getSyncChangedLock() !== null || $this->getSyncChangedLock() > ($now - self::LOCK_TIMEOUT)) {
+		if ($this->getSyncChangedLock() !== null && $this->getSyncChangedLock() > ($now - self::LOCK_TIMEOUT)) {
 			return true;
 		}
-		if ($this->getSyncVanishedLock() !== null || $this->getSyncVanishedLock() > ($now - self::LOCK_TIMEOUT)) {
+		if ($this->getSyncVanishedLock() !== null && $this->getSyncVanishedLock() > ($now - self::LOCK_TIMEOUT)) {
 			return true;
 		}
 		return false;
--- a/lib/Service/Search/MailSearch.php
+++ b/lib/Service/Search/MailSearch.php
@@ -16,7 +16,6 @@ use OCA\Mail\Db\Mailbox;
 use OCA\Mail\Db\Message;
 use OCA\Mail\Db\MessageMapper;
 use OCA\Mail\Exception\ClientException;
-use OCA\Mail\Exception\MailboxLockedException;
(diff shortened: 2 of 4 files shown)


ChristophWurst (changes requested):
Please stop your agent and the PR burst! See comments in other PRs.

ktogias:

Also pushed a follow-up commit that replaces this PR's original approach.
The diagnosis (locks blocking a read-only, already-cached listing) was
right, but removing hasLocks() entirely was more than needed. The actual
bug is a one-character mistake in `hasLocks()` itself: `||` where it
should be `&&` between "lock is set" and "lock is still within
LOCK_TIMEOUT", so any lock was treated as active forever instead of
expiring after 5 minutes as intended (see commit 00ea427332, which added
that timeout but the refactor kept `||` at the wrong level -- full
details in the new commit's message). Fixed that instead, added direct
test coverage for it, and reverted the earlier change to findMessages().

ktogias:
Sorry for the back-and-forth on this one -- found a real issue with my last
revision that's worth explaining plainly.

I'd replaced the original fix (drop the lock check from findMessages()
entirely) with a smaller one (fix the hasLocks() ||/&& typo instead),
reasoning the smaller fix covered the same problem. It didn't: hasLocks()
fixed correctly still returns true for a lock that's genuinely active right
now, not just a stale one -- and blocking a read-only listing on any active
lock, stale or not, is exactly what this PR set out to fix, since
findMessages() never touches IMAP at all.

Confirmed live: with only the typo fix in place, a real (not stuck) sync
lock still caused repeated 409s on plain message listing.

Pushed a commit restoring the original approach (drop the check from
findMessages()) on top of, not instead of, the hasLocks() typo fix --
they address different failure modes (a stale lock vs. any active lock
blocking a read that never needed to check locks) and aren't alternatives
to each other.

ChristophWurst:
Per https://github.com/nextcloud/mail/pull/13207#issuecomment-4880705823

PR #13210, opened 2026-07-03
fix: unify mail printing

GVodyanov:
Fix mail printing... again

This makes emails with multiple images not be all forced into one page

2026-07-03-181738_2560x1440_scrot

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/MessageHTMLBody.vue
+++ b/src/components/MessageHTMLBody.vue
@@ -52,7 +52,6 @@
 import iframeResize from '@iframe-resizer/parent'
 import { loadState } from '@nextcloud/initial-state'
 import { NcActionButton as ActionButton, NcActions as Actions } from '@nextcloud/vue'
-import PrintScout from 'printscout'
 import IconDomain from 'vue-material-design-icons/Domain.vue'
 import IconMail from 'vue-material-design-icons/EmailOutline.vue'
 import IconImage from 'vue-material-design-icons/ImageSizeSelectActual.vue'
@@ -61,8 +60,7 @@ import NeedsTranslationInfo from './NeedsTranslationInfo.vue'
 import logger from '../logger.js'
 import { needsTranslation } from '../service/AiIntergrationsService.js'
 import { trustSender } from '../service/TrustedSenderService.js'
-
-const scout = new PrintScout()
+import { isPrintShortcut } from '../util/printMessage.ts'
 
 export default {
 	name: 'MessageHTMLBody',
@@ -100,7 +98,6 @@ export default {
 			isSenderTrusted: this.message.isSenderTrusted,
 			needsTranslation: false,
 			enabledFreePrompt: loadState('mail', 'llm_freeprompt_available', false),
-			printOriginalHeight: null,
 		}
 	},
 
@@ -112,11 +109,6 @@ export default {
 		domain() {
 			return this.sender?.split('@').pop()
 		},
-	},
-
-	beforeMount() {
-		scout.on('beforeprint', this.onBeforePrint)
-		scout.on('afterprint', this.onAfterPrint)
 	},
 
 	async mounted() {
@@ -132,19 +124,30 @@ export default {
 	},
 
 	beforeDestroy() {
-		scout.off('beforeprint', this.onBeforePrint)
-		scout.off('afterprint', this.onAfterPrint)
-		this.$refs.iframe.iFrameResizer.close()
+		// The frame's document goes away with the frame, so this is housekeeping
+		// rather than a fix for a leak. It is done because the listener is added
+		// to a document this component does not own: nothing guarantees the frame
+		// is torn down right away, and until it is, a keydown in it would still
+		// reach a destroyed component.
+		this.getIframeDoc()?.removeEventListener('keydown', this.onFrameKeyDown)
+		this.$refs.iframe?.iFrameResizer?.close()
 	},
 
 	methods: {
 		getIframeDoc() {
 			const iframe = this.$refs.iframe
-			return iframe.contentDocument || iframe.contentWindow.document
+			return iframe?.contentDocument ?? iframe?.contentWindow?.document ?? null
 		},
 
 		onMessageFrameLoad() {
 			const iframeDoc = this.getIframeDoc()
+
+			// A frame has a window of its own, and a keydown is delivered to the
+			// window of whatever is focused. So while the reader has the message
+			// focused, the app's own window listener never sees the print
+			// shortcut and the browser would take it instead.
+			iframeDoc.addEventListener('keydown', this.onFrameKeyDown)
+
 			this.hasBlockedContent
 				= iframeDoc.querySelectorAll('[data-original-src]').length > 0
 					|| iframeDoc.querySelectorAll('[data-original-style]').length > 0
@@ -156,17 +159,17 @@ export default {
 			}
 		},
 
-		onBeforePrint() {
-			const iframe = this.$refs.iframe
-			this.printOriginalHeight = iframe.style.height
-			iframe.style.setProperty('height', `${this.getIframeDoc().body.scrollHeight}px`, 'important')
-		},
-
-		onAfterPrint() {
-			if (this.printOriginalHeight !== null) {
-				this.$refs.iframe.style.height = this.printOriginalHeight
-				this.printOriginalHeight = null
+		/**
+		 * Hand the print shortcut to the app, from inside the message frame.
+		 *
+		 * @param {KeyboardEvent} event the frame's keydown event
+		 */
+		onFrameKeyDown(event) {
+			if (!isPrintShortcut(event)) {
+				return
 			}
+			event.preventDefault()
+			this.$emit('print-shortcut')
 		},
 
 		displayIframe() {
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -21,6 +21,7 @@
 			<ThreadSummary v-if="showSummaryBox" :loading="summaryLoading" :summary="summaryText" />
 			<ThreadEnvelope
 				v-for="(env, index) in thread"
+				ref="envelopeRefs"
 				:key="env.databaseId"
 				:envelope="env"
 				:mailbox-id="$route.params.mailboxId"
@@ -29,8 +30,8 @@
 				:full-height="thread.length === 1"
 				:thread-index="index"
 				@delete="$emit('delete', env.databaseId)"
-				@loaded="addLoadedThread"
 				@move="onMove(env.databaseId)"
+				@print-shortcut="printThread"
 				@toggle-expand="toggleExpand(env.databaseId)"
 				@print="print" />
 		</template>
@@ -50,7 +51,33 @@ import logger from '../logger.js'
 import { summarizeThread } from '../service/AiIntergrationsService.js'
 import useMainStore from '../store/mainStore.js'
 import { getRandomMessageErrorMessage } from '../util/ErrorMessageFactory.js'
-import { formatDateTimeFromUnix } from '../util/formatDateTime.js'
+import {
+	BROWSER_PRINT_NOTICE_ID,
+	buildBrowserPrintNotice,
+	buildMessageHeader,
+	isPrintShortcut,
+	PRINT_CONTENT_HEIGHT,
+	PRINT_DOCUMENT_STYLE,
+	PRINT_FRAME_ID,
+	PRINT_STAGING_STYLE,
+	renderHtmlMessage,
+	renderPlainTextMessage,
+	waitForImages,
+} from '../util/printMessage.ts'
+import { wait } from '../util/wait.js'
+
+/**
+ * How long Ctrl/Cmd+P waits for the messages it expanded to render before it
+ * prints the ones that did.
+ */
+const THREAD_RENDER_TIMEOUT = 30000
+const THREAD_RENDER_POLL_INTERVAL = 100
+
+/**
+ * How long a print rendering is kept around in case the browser emits no
+ * `afterprint`. Long enough not to interrupt a print dialog that is still open.
+ */
+const PRINT_CLEANUP_TIMEOUT = 60000
 
 export default {
 	name: 'Thread',
@@ -73,7 +100,7 @@ export default {
 			enabledThreadSummary: loadState('mail', 'llm_summaries_available', false),
 			summaryText: '',
 			summaryError: false,
-			loadedThreads: 0,
+			isolatedPrint: false,
 		}
 	},
 
@@ -135,22 +162,21 @@ export default {
 			return thread[0].subject || this.t('mail', 'No subject')
 		},
 
-		threadParticipants() {
-			const seen = new Set()
-			return this.thread.flatMap((envelope) => [
-				...(envelope.from ?? []),
-				...(envelope.to ?? []),
-			]).filter(({ email }) => {
-				if (seen.has(email)) {
-					return false
-				}
-				seen.add(email)
-				return true
-			})
-		},
-
 		showSummaryBox() {
 			return this.thread.length > 2 && this.enabledThreadSummary && !this.summaryError
+		},
+
+		/**
+		 * Thread indices of the currently expanded messages, in thread order.
+		 * Those are the messages that are rendered, and therefore the ones that
+		 * can be printed.
+		 *
+		 * @return {number[]}
+		 */
+		expandedIndices() {
+			return this.thread.flatMap((envelope, index) => (
+				this.expandedThreads.includes(envelope.databaseId) ? [index] : []
+			))
 		},
 	},
 
@@ -175,8 +201,20 @@ export default {
 		window.addEventListener('keydown', this.handleKeyDown)
 	},
 
+	mounted() {
+		// A print started from the browser itself (its menu, or right-click →
+		// Print) cannot be served, see `buildBrowserPrintNotice`. The notice that
+		// says so is put up once and shown for print media only, rather than
+		// staged from a `beforeprint` handler, so that it is also there for a
+		// print the browser announces late or not at all.
+		if (!document.getElementById(BROWSER_PRINT_NOTICE_ID)) {
+			document.body.appendChild(buildBrowserPrintNotice(document))
+		}
+	},
+
 	beforeDestroy() {
 		window.removeEventListener('keydown', this.handleKeyDown)
+		document.getElementById(BROWSER_PRINT_NOTICE_ID)?.remove()
 	},
 
 	methods: {
@@ -232,7 +270,6 @@ export default {
 				await this.fetchThread()
 			}
 			this.updateSummary()
-			this.loadedThreads = 0
 		},
 
 		async fetchThread() {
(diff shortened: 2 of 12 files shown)


ChristophWurst:
Do you have before/after examples and can tell us what we should test?

Code looks … complicated

GVodyanov:
> Do you have before/after examples and can tell us what we should test?
> 
> Code looks … complicated

Yeah fair enough

Here's a file which the old version for me had trouble printing, it had an extra page and lots of white space:


[Grazie da Italo!.eml](https://github.com/user-attachments/files/29796620/Grazie.da.Italo.eml)


Here is a full changelog:

  - Unifies the two separate print code paths (Ctrl/Cmd+P for a whole thread, and the "⋯ → Print" action for a single message) into one shared pipeline in src/util/printMessage.ts.                                                                                                                            
  - Fixes: wrong message body could be printed for the "⋯ → Print" action in threads mixing plain-text and HTML messages, because the old code picked the DOM node via querySelectorAll(...)[threadIndex * 2], an index that only holds if every prior message is HTML.                                         
  - Fixes: printed header could be enlarged/misstyled by the email's own CSS (the old code inserted it into the same iframe as the email).                                                                                                                                                                      
  - Fixes: brief visible flash in the open message when printing a single message (old code inserted/removed content in the live, on-screen iframe).                                                                                                                                                            
  - Fixes: Ctrl+P could throw/fail on a message with no "To" recipient (e.g. Bcc-only).                                                                                                                                                                                                                         
  - Adds: Cc/Bcc are now shown in the printed header (previously omitted, or folded anonymously into a thread-wide participant list).                                                                                                                                                                           
  - Hardens security: printed content now renders in a sandboxed iframe (allow-same-origin allow-modals, no scripts) instead of an unsandboxed one; <script> tags and iframe-resizer markers are stripped as defense in depth.                                                                                  
  - Cleanup now waits on the afterprint event (with a 60s fallback) instead of guessing with fixed setTimeouts.                                                                                                                                                                                                 
  - Native browser print (outside the app's own Ctrl+P handler) now hides app chrome (nav, message list, composer, attachments) via CSS.

ChristophWurst:
Still hesitant on the review because the the code is hard to grasp. Maybe others are more successful.

src/util/printMessage.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @param root root to look for images in
+ */
+export function waitForImages(root: ParentNode): Promise<void> {
+	const pending = Array.from(root.querySelectorAll<HTMLImageElement>('img')).filter((img) => !img.complete)
+
+	if (pending.length === 0) {
+		return Promise.resolve()
+	}
+
+	return new Promise<void>((resolve) => {
```
odzhychko: suggestion(blocking): use `Promise.all` with a `Promise` per pending image

src/components/Thread.vue
```diff
@@ -278,223 +266,136 @@ export default {
...
-						}
-					})
-
-					while (true) {
-						if (this.loadedThreads === this.thread.length) {
-							break
-						}
-						await new Promise((resolve) => setTimeout(resolve, 100))
-					}
+			if (!(event.ctrlKey || event.metaKey) || event.key !== 'p') {
```
odzhychko: note: Not sure, but I think detecting Mac and only then listening to `metaKey` is better.

Current impl might block native Mac shortcuts from `ctrlKey + p`.

But this is a general issue for shortcuts anyway and did not change in this PR.

src/util/printMessage.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * Base stylesheet for the standalone print document. Kept intentionally
+ * small: the message bodies bring their own (isolated) CSS, this only lays
+ * out the page margins, the spacing between messages of a thread, and keeps
+ * a message header attached to the start of its content.
+ */
+export const PRINT_DOCUMENT_STYLE = `
+	@page { margin: 15mm; }
```
odzhychko: question: Why we set page margins?

How do this margins relate to the margins I select in the print dialog? Are user selected margins added to this margins or do they override?

src/util/printMessage.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @param envelope the envelope/message to build a header for
+ */
+export function buildMessageHeader(doc: Document, envelope: PrintEnvelope): HTMLElement {
+	const header = doc.createElement('div')
+	header.className = 'print-message-header'
+	header.style.cssText = 'margin: 0 0 16px 0 !important; color: #000 !important; font-family: sans-serif !important;'
+
+	const subject = doc.createElement('div')
+	subject.style.cssText = 'font-size: 18px !important; font-weight: bold !important; margin: 0 0 8px 0 !important; color: #000 !important; font-family: sans-serif !important;'
+	subject.textContent = envelope.subject || t('mail', 'No subject')
```
odzhychko: issue: Do we want to show 'No Subject' if the subject is an empty string `""`? If so, than we probably should also show 'No subject' if the string has only whitespace characters.

And can `envelope.subject` be `undefined`?

src/util/printMessage.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @param doc document to create the header elements in
+ * @param envelope the envelope/message to build a header for
+ */
+export function buildMessageHeader(doc: Document, envelope: PrintEnvelope): HTMLElement {
+	const header = doc.createElement('div')
+	header.className = 'print-message-header'
+	header.style.cssText = 'margin: 0 0 16px 0 !important; color: #000 !important; font-family: sans-serif !important;'
+
+	const subject = doc.createElement('div')
+	subject.style.cssText = 'font-size: 18px !important; font-weight: bold !important; margin: 0 0 8px 0 !important; color: #000 !important; font-family: sans-serif !important;'
```
odzhychko: issue: Is there a reason to not add these styles to `PRINT_DOCUMENT_STYLE`?

If not, I would like them to go into `PRINT_DOCUMENT_STYLE`.
Then it would be it make easier to reason about them.

I will look at the styles in detail later with the questions:

- what styles a redundant
- why is each `!important` needed

src/util/printMessage.ts
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * The content is returned as a normal (light DOM) element so that long
+ * emails paginate across print pages. It must only ever be attached to the
+ * document while printing and removed again afterwards, because the copied
+ * email CSS is global and would otherwise affect the rest of the app.
+ *
+ * @param doc document to create the wrapper element in
+ * @param sourceDocument the message iframe's own document
+ */
+export function buildHtmlMessageContent(doc: Document, sourceDocument: Document): HTMLElement {
```
odzhychko: issue: Copying messages into one document seems error prone with regards to styles. Currently style sheet from message A would influence the styles from message B.

I think each message should be put into its own iframe.

src/util/printMessage.ts
```diff
@@ -0,0 +1,160 @@
...
+ * content would fall back to unstyled browser defaults.
+ *
+ * The body HTML is copied verbatim from the already server-sanitized message
+ * iframe. `<script>` and the iframe-resizer marker are stripped as a
+ * safeguard; the print document that this content is placed into must itself
+ * be sandboxed without script execution (see `Thread.printMessages`), because
+ * unlike the live message iframe it is not protected by the backend's
+ * Content-Security-Policy.
+ *
+ * The content is returned as a normal (light DOM) element so that long
```
odzhychko: > It must only ever be attached to the
> document while printing and removed again afterwards, because the copied
 > email CSS is global and would otherwise affect the rest of the app.

issue: But if the styles are global they would affect the app when attached for printing. Seems bad. Not sure, why we don't attach it in the an iframe (or use shadow dom) for isolation.

src/tests/unit/util/printMessage.spec.js
```diff
@@ -0,0 +1,180 @@
...
+		it('falls back to "No subject" when the subject is empty', () => {
+			const header = buildMessageHeader(document, { ...envelope, subject: '' })
+
+			expect(header.firstChild.textContent).toBe('No subject')
+		})
+
+		it('gives the subject an explicit inline font size so email CSS cannot enlarge it', () => {
+			const header = buildMessageHeader(document, envelope)
+
+			expect(header.firstChild.style.fontSize).not.toBe('')
```
odzhychko, line 66: thought: Not a good test ensure it is "enlarged". Would like to see visual snapshot tests for style related stuff.

But maybe we do not have this setup for the app yet and is out-of-scope for this PR. But would be nice longterm.

src/tests/unit/util/printMessage.spec.js
```diff
@@ -0,0 +1,180 @@
...
+				'From: Alice <alice@example.com>',
+				'To: Bob <bob@example.com>',
+				`Date: ${formatDateTimeFromUnix(envelope.dateInt)}`,
+			])
+		})
+
+		it('omits Cc and Bcc lines when there are no such recipients', () => {
+			const header = buildMessageHeader(document, envelope)
+
+			const lines = Array.from(header.querySelectorAll('div')).map((line) => line.textContent)
```
odzhychko: nit(non-blocking): could be simplified and made robuster by asserting `.innerHTML` does not contain `Cc:` and does not contain `Bcc:`

Same for some other tests to:

- 'includes Cc and Bcc lines when present, with all recipients'
- 'omits the To line when there are no recipients'

src/util/printMessage.ts
odzhychko: I will have to look at how we sanatize mail HTML, but if each mail has it's own `<html/>` or `<body/>` or `<head/>` elements, they should not go all into one document.

I don't even know how a browser would handle multiple `<body/>`s in a document ^^

src/components/Thread.vue
```diff
@@ -673,4 +574,35 @@ export default {
 .user-bubble__title {
 	cursor: pointer;
 }
+
+// The "Print message" action and Ctrl/Cmd+P render into an isolated iframe
```
odzhychko: thought: Interesting. We have two ways of printing. Ideally, they should have the same result and share as much code as possible.

Our print shortcut might expand all threads for convenience, but then it should use the same styles and HTML structure for printing. 

I see a way here to heavily simplify the print logic, but I have to think about it more.

odzhychko (commented):
Did a cursory reading. And added what I noticed.

Full review will take some more time. The domain of HTML mails is entirely new to me, and I'm looking at the first time at our mail rendering and printing code. Also I need to read up on how our backend sanitizes mail to make sure the new printing implementation does not create some vulnerability.

src/util/printMessage.ts
GVodyanov: Yep, done

src/util/printMessage.ts
GVodyanov: Added No subject text

src/util/printMessage.ts
GVodyanov: Moved them,

About important it's to not have conflicts with styling that comes with the email

src/util/printMessage.ts
GVodyanov: User selected margins override, this is basically a sensible default from our side in case the browser doesn't add them

src/components/Thread.vue
GVodyanov: Sure yeah can set this to metaKey

src/components/Thread.vue
GVodyanov: So there's a print action on single thread elements, and one to print the whole thread with ctrl + P

The idea of this PR was to make the pipeline as standard as possible yeah so if you have any suggestions they are very welcome

src/tests/unit/util/printMessage.spec.js
GVodyanov: Sure done

src/util/printMessage.ts
GVodyanov: It's complicated

So we are only extracting innerHtml so there is no risk of having multiple bodies

However putting every message into its own iframe doesn't work because we need to understand the full height of the elements in order to lay them out in pages, meanwhile iframes just crop after a certain height

GVodyanov: It's the best solution I could find... but idk if there are better ways

src/util/printMessage.ts
GVodyanov: I'm not sure if the shadow dom would work across all browsers... would need some more testing

A quick fix I had for this was to simply reload the page after printing

src/util/printMessage.ts
odzhychko: Thx, for the reasoning behind the solution. I will give it some thought to.

src/components/Thread.vue
odzhychko: > So there's a print action on single thread elements, and one to print the whole thread with ctrl + P

Ah, okay. I will keep that in mind.

odzhychko: Given, the users has a mail thread open: What is the expected behavior, if the user uses "Print..." through the browser UI?

- Print the whole thread, even hidden messages
- Print the whole thread, but only already expanded messages
- Print latest last mail (same as selecting `... > Print` in a mail actions)
- Print the selected (focused) mail in the thread

Not sure if it is relevant for this PR. But maybe you already thought about it or checked how other clients behave. If this PR does not want to change it, I will just check that our previous behavior did not break.

src/components/Thread.vue
```diff
@@ -673,4 +574,35 @@ export default {
...
+// and print that, so these rules only matter for a native browser print
+// (browser menu / right-click → Print) of the live page. They hide the app
+// chrome and let the document flow so the message body isn't printed inside
+// the surrounding UI. Because the message body keeps its own sandboxed
+// iframe, none of its CSS leaks into the app, and these rules revert
+// automatically once printing is done — nothing needs to be reloaded.
+@media print {
```
odzhychko: issue: Native browser print is broken for single message

Given:

- Show all messages in thread: enabled 
- Mail: [Grazie da Italo!.eml](https://github.com/user-attachments/files/29796620/Grazie.da.Italo.eml)

When:

- Using native browser print

Then

- Expected (result as with Ctrl/Cmd+P)
   Image

- Actual
   

 
  Image

odzhychko: Same Issue with message thread.

src/components/Thread.vue
odzhychko, line 201: issue: Method of handling Cmd/Ctrl+P does not work reliably.

1. `this.handleKeyDown` is not triggered when the message <iframe> is focused.

     
   Image

2. `this.handleKeyDown` is not triggered when the message <body> is focused.
  - Steps to reproduce:
     - Open a message
       - This updates the URL
     - Reload page
       - message is opened automatically because of URL


To be honest, I'm not sure, what what is the root cause. Some guesses are:

- `keydown`events on `iframe` do not bubble up by default
  - we might need an additional listener on each `iframe`
- for the issue with `body`
  - maybe some other listener calls `stopPropagation`.

Seems we have https://nextcloud-vue-components.netlify.app/#/Composables/useHotKey

Maybe it solve some of the issues above.

src/components/Thread.vue
```diff
@@ -278,223 +266,136 @@ export default {
...
-						const iframeDocument = iframe.contentDocument || iframe.contentWindow.document
-						const iframeContent = iframeDocument.body.innerHTML
-						const div = virtualIframeDocument.createElement('div')
-
-						div.innerHTML = iframeContent
-						virtualIframeDocument.body.appendChild(div)
+			try {
+				this.thread.forEach((envelope) => {
+					if (!this.expandedThreads.includes(envelope.databaseId)) {
+						this.expandedThreads.push(envelope.databaseId)
```
odzhychko: issue: printing expands all messages in a thread

From one side this is non-blocking, because it happened before too

On the other the code is rewritten here and ideally a print action should not influence the users UI (not even temporarily).

Proper solution would be, to load message content always separately for printing. (Maybe using a shared cache to reduce requests).

src/components/Thread.vue
odzhychko: Ok, this is already broken in `main`.

> Native browser print (outside the app's own Ctrl+P handler) now hides app chrome (nav, message list, composer, attachments) via CSS.

But I assume this PR should fix it?

odzhychko:
I think I got the gist of mail printing.

Let me summarize requirements:

- User can print a single message through "... > More actions > Print message"
  - Works with "Show all messages in thread" enabled and disabled
 - User can use the browser native print action to print a message
   - If "Show all messages in thread" is disabled
 - User can use the browser native print action to print a thread
   - If "Show all messages in thread" is enabled
- When a thread is printed, we want to print all messages from the thread

Some background about browser native print action:

- Usual shortcut is Ctrl/Cmd+P
 - `beforeprint` can listened for to detect it and we 
   - can prepare something synchronously (e.g. change styles)
   - can not do something asynchronously (e.g. load images)
   - can not abort native print action to execute our handler


From that, my overarching issue with this PR (and solution in `main`) is:

 - Browser native print has different results than our custom handler for  Ctrl/Cmd+P
   - we do not expand all messages in a thread 
   - we do not wait for images
- native print action and custom handler cannot have same results
   - after detecting native print action we cannot execute asynchronous code
   - our custom handle is asynchronous


~Proposed solution~:

- ~Make native browser print work out of the box as expected~
     - ~Have everything that needs to printed in DOM~
     - ~Have hidden DOM-Nodes with the mail content for all messages of a thread, even if the thread is not expanded~
      - ~This should make images load in the background automatically~
      - ~Even if it is hidden, Iframe-resizer should be able to calculate the size~
        
         > But "laid out" is not the same as "visible." The iframe's content does have to be rendered in the layout sense, yet you can keep the element itself visually hidden while that happens — off-screen, opacity: 0, visibility: hidden, positioned outside the viewport, or sized 1×1 — measure it, then reveal it at the correct height.
         >
         > Suggestion from Claud Opus 4.8 for possible solutions.
         
        
        - ~With this it should be possible to have an iframe for each mail and not have styles of mails interfere with each other and our header styles~
   - ~Have hidden DOM-Nodes for header (e.g. CC, BCC)~
   - ~Use `@print` stylesheets to have everything unneeded hidden on native print~
   - ~Use `@print` stylesheets to have hidden DOM-nodes be visible~
- ~I think we then do not even have to implement a custom handler~
- ~If the user clicks "... > More actions > Print message"~
   - ~Use the same code/hidden HTML but only for this instead of all messages (e.g. hide others, remove them temporarily from DOM or use a temporary in-memory document that can be printed)~
   
Edit my proposed solution does not work, because the size width of the page is different then the width of a mail in the UI. Also page size can be changed in the native print dialog. See https://github.com/nextcloud/mail/pull/13210#discussion_r3663648613

src/components/Thread.vue
odzhychko: Fine for now, as we are already doing this.

src/util/printMessage.ts
odzhychko: Not relevant anymore. We append to an iframe anyway.

src/util/printMessage.ts
odzhychko: You were right. I missed that the size we display to the user is not the size that has to be printed.

For the size in the mail app, we can use iframe-resizer. But for printing, the width of the page is not the same width as the content in our UI. This becomes especially evident if the user changes the page size or orientation in the print dialog.

So now the document (with no iframes) seems to have the best long-term solution. But would need good isolation (Shadow DOM seems widely available enough).

GVodyanov:
In the end the iframe in iframe solution works surprisingly well, I tested it with various standard layouts in firefox and I consider it to be pretty functional, it also has less quirks than other solutions

hamza221:
I'm just testing din't look at the code yet. 
Going through the browser print (Chrome, macos) not the whole email gets rendered. 

[Print- mail app .pdf](https://github.com/user-attachments/files/30512067/Print-.mail.app.pdf)

[Print- browser.pdf](https://github.com/user-attachments/files/30512064/Print-.browser.pdf)

Doing a code review ... ⏳

src/components/Thread.vue
```diff
@@ -278,223 +311,214 @@ export default {
 		},
 
 		async handleKeyDown(event) {
```
hamza221, line 325: What happens if The shortcut is pressed a couple times in a row ?

src/components/Thread.vue
```diff
@@ -278,223 +311,214 @@ export default {
...
+		 * `parent` has to be part of a document already, because a message frame is
+		 * written to as soon as it is created.
+		 *
+		 * @param {HTMLElement} parent element to append the message to
+		 * @param {number} index thread index of the message to append
+		 * @return {?HTMLIFrameElement} the message's frame, if it has one
+		 */
+		appendPrintMessage(parent, index) {
+			const envelope = this.thread[index]
+			const envelopeComponent = this.$refs.envelopeRefs?.[index]
```
hamza221: Are we sure Vue keeps rendered refs idx in sync with the array ?  
Maybe a manual lookup would be more robust

src/components/Thread.vue
odzhychko: Works now

GVodyanov:
# **EDIT THE FOLLOWING IS OUTDATED**

## How printing works now                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                
The per-message iframes are gone. Each message body renders into a shadow root instead, so the print output is one ordinary document that the browser paginates normally.

## Why iframes didn't work

Its content lays out against the iframe's own viewport, not the paper. So we had to guess A4, render at that width, measure the height and pin it. The moment you change paper size or orientation in the open dialog, that measurement is wrong and nothing can re-run it.

## What a printed message looks like                                                                                                                                                                                                                                                                             
 
 ```                                                                                                                                                                                                                                                                                                               
  <div class="print-message">                                                                                                                                                                                                                                                                                   
    <div class="print-message-header">…</div>      ← plain DOM, inline !important styles                                                                                                                                                                                                                        
    <div class="print-message-content">            ← shadow host                                                                                                                                                                                                                                                
      #shadow-root                                                                                                                                                                                                                                                                                              
        <html><head><style>…email css…</style></head>                                                                                                                                                                                                                                                           
              <body>…email body…</body></html>                                                                                                                                                                                                                                                                  
  </div>  
  ```
  
 ## Gaps shadow DOM leaves, and what fills them                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                
  A shadow boundary scopes selector matching and nothing else, so three small helpers patch what leaks:                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                
  - Inherited properties still cross in through the host → the host pins color: #000, or a dark theme prints white-on-white.                                                                                                                                                                                    
  - The app's stylesheet no longer reaches plain-text bodies → pre-wrap is re-supplied inside the shadow root.                                                                                                                                                                                                  
  - A shadow root has no <base> → absolutizeUrls() resolves relative URLs against the message's own endpoint before copying (this also applies any <base> the email carried).                                                                                                                                   
  - querySelectorAll doesn't descend into shadow roots → image-waiting walks them explicitly.

GVodyanov:
Images are blocked via `TransformImageSrc.php` and `TransformStyleURLs.php`

GVodyanov:
Currently the PR builds everything inside an iframe and then the contents of that iframe are extracted and printed

GVodyanov:
Should be ready @kesselb @odzhychko

kesselb:
Thanks, works well!

I've tested a couple of emails and compared the results with Roundcube and Thunderbird. They all look pretty similar (which is good). 

I found it weird to print one message with the menu item, but a whole thread with ctrl + p :shrug: :confused: 

--- 

Claude Feedback:

1)

Add a "printable" method to ThreadEnvelope.vue and use that in Thread.vue instead of the loadedThread counter because loadedThreads counts @loaded events, but @loaded is not one-per-message. 

     printable() {
         return this.loading === Loading.Done && this.message !== undefined
     },

     while (!this.$refs.envelopeRefs?.every((c) => c.printable) && Date.now() < deadline) {
         await wait(THREAD_RENDER_POLL_INTERVAL)
     }


2) 

waitForImages is unbounded and runs before cleanup is armed. One stuck image leaves isolatedPrint true and every later print returns silently at the guard. Arm the timer before awaiting.

3) 

Mailvelope's chrome-extension:// iframe has a null contentDocument, buildMessageContent throws, and a thread with one PGP message prints nothing. Guard iframe?.contentDocument.

4)

printMessage.ts:448 — `event.key !== 'p'` fails under CapsLock and non-Latin layouts. Use event.code === 'KeyP'.

GVodyanov:
> I found it weird to print one message with the menu item, but a whole thread with ctrl + p 🤷 😕

This is how we've always done it and how I comunicated it should work to customers, but we could possibly change it in the future

> Claude Feedback:

Addressed all points and added tests, please re-review @kesselb Thanks!

src/components/MessageHTMLBody.vue
```diff
@@ -132,8 +124,7 @@ export default {
 	},
 
 	beforeDestroy() {
-		scout.off('beforeprint', this.onBeforePrint)
-		scout.off('afterprint', this.onAfterPrint)
+		this.$refs.iframe?.contentDocument?.removeEventListener('keydown', this.onFrameKeyDown)
```
odzhychko: question(non-blocking): Why are we calling `removeEventListener` on something that is going to be detached from DOM and garbage collected anyway?

Might have a good reason, but it is not obviously apparent.
So might be worth adding a comment why this is done.

src/components/MessageHTMLBody.vue
```diff
@@ -132,8 +124,7 @@ export default {
...
 		this.$refs.iframe.iFrameResizer.close()
```
odzhychko: nit(non-blocking): One line uses `this.$refs.iframe?.` and the next one `this.$refs.iframe.`. 

Either inconsistent code or second line might fail in some edge case.

src/components/MessageHTMLBody.vue
odzhychko: question(non-blocking): Why is the event listener added on `iframe.contentDocument || iframe.contentWindow.document` but removed only on `iframe.contentDocument`.?

Do we not need to remove the event listener from `iframe.contentWindow.document` if was falsy `iframe.contentDocument` and we added it to `iframe.contentWindow.document`?

Relates to https://github.com/nextcloud/mail/pull/13210/changes#r3688681978

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
+		 * Only this one way of printing is turned away. Ctrl/Cmd+P never reaches
+		 * here — `handleKeyDown` takes it and prints the thread — so this fires for
+		 * a print the app was never asked to make. Such a print always renders the
+		 * top-level document, and no message ever goes in there: it could not be
+		 * printed correctly, because a message is rendered in a frame that the
+		 * printer cuts off at the end of a page instead of breaking it across
+		 * pages, and it would put the (untrusted) email markup in the app's own
+		 * document, outside the sandboxed frame `printMessages` confines it to.
+		 */
+		onBrowserBeforePrint() {
```
odzhychko: issue(non-blocking): Needlessly complicated code for showing the notice. The `buildBrowserPrintNotice` can just always be part or the HTML. No need to listen on `beforeprint` and `afterprint` and timout handling.

odzhychko: Not even `showError(${browserPrintTitle()}. ${browserPrintHint()})` is required in `onBrowserAfterPrint` because the message is shown to the user in the print dialog anyway.

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
+			if (this.isolatedPrint) {
+				return
+			}
+
+			this.teardownBrowserPrint()
+			document.body.appendChild(buildBrowserPrintNotice(document))
```
odzhychko: issue(non-blocking): Printing with background makes note not well legible. Any reason for the background at all?

Image

Image

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
+			// thread once more, and it keeps `onBrowserBeforePrint` from rendering
+			// a print document into the live page for the same print, should a
+			// browser propagate the frame's print events to the app window.
+			if (this.isolatedPrint) {
+				return
+			}
+			this.isolatedPrint = true
+
+			const frame = document.createElement('iframe')
+			frame.setAttribute('aria-hidden', 'true')
```
odzhychko, line 412: question(non-blocking): Why is `	frame.setAttribute('aria-hidden', 'true')` required?

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
+			// The messages are copied in from the sanitized message frames, but
+			// unlike those this frame is not protected by the backend's
+			// Content-Security-Policy. Without `allow-scripts` no script or inline
+			// event handler in the content can run; `allow-same-origin` lets us
```
odzhychko: question(non-blocking): Are you sure `frame.setAttribute('sandbox', 'allow-same-origin allow-modals')` this is required?

Printing works without that line as expected to. Comment doesn't make sense to me.

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
+			// populate it and `allow-modals` lets it open the print dialog.
+			frame.setAttribute('sandbox', 'allow-same-origin allow-modals')
+			frame.style.cssText = `${PRINT_STAGING_STYLE} height: ${PRINT_CONTENT_HEIGHT}; border: 0;`
```
odzhychko, line 421: issue(non-blocking): Seems like overly complicated code.

Instead of:

```
frame.setAttribute('aria-hidden', 'true')
frame.style.cssText = `${PRINT_STAGING_STYLE} height: ${PRINT_CONTENT_HEIGHT}; border: 0;
```

just

```
frame.style.display = 'none'
```

seems to work the same.

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
+					return
+				}
 
-			virtualIframeDocument.body.appendChild(hr)
-			virtualIframeDocument.body.appendChild(subjectSpan)
-			virtualIframeDocument.body.appendChild(senderSpan)
-			virtualIframeDocument.body.appendChild(dateSpan)
-			virtualIframeDocument.body.appendChild(recipientSpan)
+				frame.contentWindow.addEventListener('afterprint', cleanup, { once: true })
+				frame.contentWindow.focus()
```
odzhychko, line 462: issue(non-blocking): Seems like overcomplicated code.

1. `frame.contentWindow.focus()` seems not to be required.
2. `print()` is blocking, so no need for event listeners.
    ```
    frame.contentWindow.print()
    cleanup()
     ```

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
-			dateSpan.textContent = t('mail', 'Date') + ': ' + formatDateTimeFromUnix(this.thread[index].dateInt)
+				await waitForImages(doc)
 
-			const recipientSpan = virtualIframeDocument.createElement('p')
-			recipientSpan.style.fontWeight = 'bold'
-			recipientSpan.textContent = t('mail', 'To') + ': ' + this.thread[index].to[0].label + this.thread[index].to[0].email
+				// The frame is gone once the cleanup has run, and there is
+				// nothing left to print from.
+				if (cleanedUp) {
+					return
```
odzhychko: issue(non-blocking): I guess, this case should show a user warning.

e.g., "Failed to load images for printing."

src/components/Thread.vue
```diff
@@ -277,224 +309,261 @@ export default {
...
-			dateSpan.style.fontWeight = 'bold'
-			dateSpan.textContent = t('mail', 'Date') + ': ' + formatDateTimeFromUnix(this.thread[index].dateInt)
+				await waitForImages(doc)
 
-			const recipientSpan = virtualIframeDocument.createElement('p')
-			recipientSpan.style.fontWeight = 'bold'
-			recipientSpan.textContent = t('mail', 'To') + ': ' + this.thread[index].to[0].label + this.thread[index].to[0].email
+				// The frame is gone once the cleanup has run, and there is
+				// nothing left to print from.
+				if (cleanedUp) {
```
odzhychko: issue(non-blocking): Seems like a case never that never can happen. `waitForImages` has a `timeout` of `10000`. `PRINT_CLEANUP_TIMEOUT` is  `60000`. Seems no way that `cleanup` can run before `waitForImages` can finish.

src/util/printMessage.ts
```diff
@@ -0,0 +1,481 @@
...
+
+/** Height the print document is staged at: one A4 page. */
+export const PRINT_CONTENT_HEIGHT = '297mm'
+
+/**
+ * Off-screen staging for the print frame. It is parked out of sight rather than
+ * hidden with `display: none`, because only a rendered message loads and
+ * decodes its images before the browser takes its print snapshot.
+ */
+export const PRINT_STAGING_STYLE = `position: fixed; top: 0; inset-inline-start: -9999px; width: ${PRINT_CONTENT_WIDTH};`
```
odzhychko, line 58: > It is parked out of sight rather than hidden with `display: none`, because only a rendered message loads and decodes its images before the browser takes its print snapshot.

question(non-blocking): You sure? 

Relates to https://github.com/nextcloud/mail/pull/13210/changes#r3689023226

src/util/printMessage.ts
```diff
@@ -0,0 +1,481 @@
...
+ * A header is rendered next to its message rather than inside its shadow root,
+ * so that the (untrusted) email CSS cannot restyle it, that is what produced
+ * the oversized "big letters" header in the old implementation. For a native
+ * browser print that puts it in the live page, where the app's own stylesheet
+ * would style it instead. Inline declarations outrank every declaration a
+ * stylesheet can make, `!important` ones included, but only for the element
+ * they sit on, so every part of the header carries its own.
+ */
+const PRINT_HEADER_STYLES = {
+	header: `${PRINT_HEADER_BASE_STYLE} margin: 0 0 16px 0 !important;`,
```
odzhychko: nit(non-blocking): All this `!important` stuff seems not to be needed not that each mail is in its shadow DOM.

src/util/printMessage.ts
```diff
@@ -0,0 +1,481 @@
...
+ * A message is rendered from an endpoint of its own, so its relative URLs
+ * resolve against that endpoint rather than against the app. A frame could be
+ * told so with a `<base>`; a shadow root cannot — it has no base of its own and
+ * inherits the one of the document it sits in — so the URLs are resolved here
+ * while the copy still belongs to the message's document.
+ *
+ * @param root the copied message to rewrite in place
+ * @param baseUrl the message document's base URL
+ */
+function absolutizeUrls(root: Element, baseUrl: string): void {
```
odzhychko, line 317: question(non-blocking): When do we even have relative URLs in our HTML bodies?

I tried:

1. External image in HTML mail
    Seems we use an absolute URL to our `/proxy` endpoint there.
    e.g., `Image`
2. Image that is an attachment of the HTML mail
    Seems we already use absolute URLs there.
    e.g., `cid:Djx8QHiwMB-iP4pEPKEIHpC@c989464695ca`

src/util/printMessage.ts
```diff
@@ -0,0 +1,481 @@
...
+	root.querySelectorAll('[src], [href], [poster], [background], [srcset]').forEach((node) => {
```
odzhychko: issue(non-blocking): First part text is duplicate of `URL_ATTRIBUTES`. Easy to modify `URL_ATTRIBUTES` but for get to change it here.

odzhychko (approved):
Works.

Some code that could be simplified.
Some code where I'm not sure why it's needed.
Some small possible improvements.

But nothing blocking from my side.

src/components/Thread.vue
GVodyanov: Yeah did that

src/components/Thread.vue
```diff
@@ -277,224 +312,229 @@ export default {
...
-			const participantsLine = document.createElement('p')
-			participantsLine.textContent = this.threadParticipants
-				.map((participant) => `${participant.label} <${participant.email}>`)
-				.join(', ')
-			threadInfo.appendChild(participantsLine)
+			const iframe = messageEl.querySelector('iframe')
+			if (iframe) {
+				renderHtmlMessage(message, iframe.contentDocument)
+				return
+			}
```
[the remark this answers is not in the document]
GVodyanov: Handled at Thread.vue:505. Not with the suggested contentWindow.document fallback though: the only real case is a Mailvelope frame from the extension origin, where contentWindow.document throws too. That message prints as its header alone and the rest of the thread still prints.

src/components/Thread.vue
GVodyanov: Because the frame is parked off-screen, not display: none, off-screen content is still in the accessibility tree and would be announced

src/components/Thread.vue
GVodyanov: Printing works without it, yes; the sandbox isn't there to make printing work. The live message frames are served by the backend under a restrictive CSP; this frame isn't, and we copy the email markup into it. sandbox without allow-scripts is what replaces that CSP. allow-same-origin and allow-modals are the two capabilities we then have to grant back (populate the doc, open the dialog)

src/components/Thread.vue
GVodyanov: Both are needed. focus() is a Firefox requirement: it prints the focused window, so without it you get the app page instead of the frame.

src/util/printMessage.ts
GVodyanov: What absolutizeUrls covers is a URL the sender wrote that HTMLPurifier passed through untouched, <a href="/x">, , which used to resolve against the message endpoint in the frame but would resolve against the app once copied into a shadow root, silently pointing at Mail's own routes. It's cheap and the tests pin the behaviour, so I kept it.

GVodyanov:
Fixed (claude description):

absolutizeUrls selector duplicates URL_ATTRIBUTES: agreed, it would drift. The selector is now derived: URL_ATTRIBUTES + a named SRCSET_ATTRIBUTE build URL_SELECTOR, and the srcset branch uses the constant too.

!important no longer needed now each mail is in shadow DOM — right for message headers, which only ever land in the standalone print document. PRINT_HEADER_STYLES is now plain inline styles. The notice is a different case — it lives in the app document where the app stylesheet does reach it — so I split it into PRINT_NOTICE_STYLES, which keeps !important.

"Printing with background makes the note not well legible" — no reason for it; that's the app's dark theme body background. BROWSER_PRINT_STYLE's print block now sets html, body { background: #fff !important; }.

"if (cleanedUp) can never happen" — correct, 10s < 60s. The real thing the backstop guards is a missing afterprint, so it's now armed right before print() instead of before waitForImages, and cleanup() clears it. The dead branch is gone.

"This case should show a user warning" — that branch no longer exists, but the underlying concern (images missing from the print) was real and unreported. waitForImages now returns whether everything settled and printMessages logs a warning when it gave up. I did not add a toast: it would fire exactly as the print dialog opens, so the reader would never see it.

"Needlessly complicated code for showing the notice" — agreed, and done. The notice is appended once in mounted and removed in beforeDestroy; onBrowserBeforePrint, onBrowserAfterPrint, teardownBrowserPrint, browserPrintTimeout and both window listeners are gone (−52 lines in Thread.vue). The showError on afterprint is gone too — you're right that the print dialog already shows it. This also covers a print the browser never announces.

One addition it forced: with the notice permanent, body > *:not(#notice) would hide our own staging frame if a browser ever formats the app document for print media during an iframe print. The rule now excludes it via a new PRINT_FRAME_ID (mail-print-frame), with tests.

MessageHTMLBody removeEventListener (three threads) — you're right that it's not strictly needed; the frame's document dies with the frame. But it's a listener on a document this component doesn't own, and nothing guarantees immediate teardown, so I kept it and added a comment saying exactly that. The contentDocument vs contentWindow.document inconsistency is gone — both add and remove now go through getIframeDoc(), which I made null-safe. And this.$refs.iframe.iFrameResizer.close() is now optional-chained to match.

PR #13212, opened 2026-07-04
fix: more stringent checks on valid email addresses while autocompleting recipients

### LINKED ISSUE #10432: Contact search with @ is added to recipients, blocking outbound
### Steps to reproduce

1. Forward an email
2. Type in a filter for recipient containing an @, e.g. `myemail@e`
3. Select an item from the results, e.g. myemail@example.org
4. Click Send

### Expected behavior

When you click send, no contact is added, and the mail is sent

### Actual behavior

When you click send, a contact with your search text (here `myemail@m`) is added as a recipient, blocking the mail outbound.

### Mail app version

4.0.6

### Nextcloud version

30.0.0

### Mailserver or service

any

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

madbob:
To avoid duplications I've reused the `isValidEmail()` function took from `AccountForm.vue` and moved in the global utilities file `emailAddress.js`

Fixes #10432


ChristophWurst:
Thanks for the PR

ChristophWurst:
I tried the STR from https://github.com/nextcloud/mail/issues/10432 but can not reproduce on main.

How can we test this?

madbob:
- have a Contact with an email (e.g. foobar@example.com)
- in Mail, start a new message
- in the To field, start writing "foobar@ex". The fact to add @ and one or more chars for the domain is relevant
- select the contact from the autocomplete list: this will add "foobar@example.com" to the recipients list
- blur the To field
- at this point, to the To field is appended another recipient having as email address "foobar@ex" (the partial address you were writing)

https://github.com/user-attachments/assets/73ccfe09-65c9-4cd6-a32e-4f55e2b99f7c

PR #13213, opened 2026-07-04
fix: moved "Delete thread" at the end of the contextual actions menu

### LINKED ISSUE #10071: Destructive action "Delete thread" is not last item in action list
### Steps to reproduce

1. Open the action menu of a thread you want to delete
2. See it’s not at the end

### Expected behavior

Destructive actions are always at the end of an action menu (most destructive last).

### Actual behavior

image

### Mail app version

4.0.0-beta.2

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

madbob:
The "Delete" option has been moved at the end of the actions' menu, under the "More actions" submenu.

Unsure if this was the actually desired behavior, as "Delete" is also one of the primary action buttons (so: anyway, it is well visible and clickable by the distracted user...).

Fixes #10071

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -328,20 +328,6 @@
 					</template>
 				</ActionButton>
 				<ActionButton
-					v-if="hasDeleteAcl"
-					:close-after-click="true"
-					@click.prevent="onDelete">
-					<template #icon>
-						<DeleteIcon :size="20" />
-					</template>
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Thanks! Looks good 😎 :shipit:

PR #13214, opened 2026-07-04
fix: hide "create task" option when no task calendar is available

### LINKED ISSUE #9884: "create task from email" option should not be there when Tasks app is not installed
### Steps to reproduce

1. have mail app installed but not tasks
2. in a received email, click the ... menu > more actions > create task
3. modal shows, you enter details
4. error: could not create task

### Expected behavior

the option "create task" should only be there when Tasks app is installed

### Actual behavior

4. error: could not create task

### Mail app version

3.7.5

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

NC 29.0.1

madbob:
The "Create task" modal ask for the calendar to which append the new task, but if none is available it is not possible to select anything and the whole operation cannot be concluded.

Fixes #9884

### DIFF
--- a/src/components/Envelope.vue
+++ b/src/components/Envelope.vue
@@ -422,6 +422,7 @@
 					{{ t('mail', 'Reply with meeting') }}
 				</ActionButton>
 				<ActionButton
+					v-if="tasksEnabled"
 					:close-after-click="true"
 					@click.prevent="showTaskModal = true">
 					<template #icon>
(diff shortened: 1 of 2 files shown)


ChristophWurst (commented):
Thanks!

When are the XHRs for the collection listing happening and does this cause them to happen earlier?
In other words: any (performance) downsides?

madbob:
User's calendars are loaded from the server as soon as the main Vue `App` [is mounted](https://github.com/nextcloud/mail/blob/main/src/App.vue#L56), so there is not any extra DAV call.

ChristophWurst (approved):
Let's do this

PR #13255, opened 2026-07-06
feat: expand groups to individual addresses

### LINKED ISSUE #2723: sending to groups > insert individual addresses
### Feature Request

**Actual behaviour:**
I type a group name (group in contacts app), the group is suggested as "groupname(Contacts)", I click on it, it stays like that in the To: field.

**Expected behaviour:**
Individual Email addresses are inserted when clicking on the group name.

**Comments**
This was already discussed in https://github.com/nextcloud/mail/issues/41 but not fixed.
Also, the "...(Contacts)" is not helpful to understand what is going on.

### LINKED ISSUE #3132: Avoid BCC fail by warning when nextcloud groups are adressed
It's very useful to be able to put nextcloud groups in the recipients fields.

**It would be great to have kind of a warning when nextcloud groups are put in the 'To:' or 'CC' field.**

It might seem logical but it's not visible that all mails are listed.

We just had an incident where a big group of people received an e-mail with visible adresses.

Thanks for the great work!!

### LINKED ISSUE #9075: Writing am mail to a contact group should only use the preferred Mail-Address

Writing a mail to a contact group (using the NC mail client): all Mailaddresses of the contacts belonging to a contact group are taken.

Only the preferred mail address of the contacts should be used. (like a distribution list)

_No response_

### Additional context

_No response_

madbob:
Considers this a proof-of-concept, as some more work is required including:

- handle contacts' avatars
- suppress the no-longer required `GroupsIntegration::expand()` function (and adjust `TransmissionService`)

Fixes #2723 
Fixes #9075 
Fixes #3132

### DIFF
--- a/lib/Service/Group/ContactsGroupService.php
+++ b/lib/Service/Group/ContactsGroupService.php
@@ -97,7 +97,9 @@ class ContactsGroupService implements IGroupService {
 			}
 			foreach ($emails as $email) {
 				$receivers[] = [
-					'email' => $email
+					'id' => $r['UID'],
+					'name' => $r['FN'] ?: $email,
+					'email' => $email,
 				];
 			}
 		}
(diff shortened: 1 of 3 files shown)


madbob:
Any feedback about this approach?

The action to replace a Group selection with the exploded list of recipients involved in that same Group is a viable new behavior for Mail?

PR #13256, opened 2026-07-06
feat: save email as .eml file in Files

### LINKED ISSUE #675: Save mail to files
It would be nice if I could save a copy of an e-mail (and its attachments) in my Nextcloud files. Is this an enhancement the mail team can get behind?

### LINKED ISSUE #9240: Save Email to Files

How do I save and email to a Files folder? The current option is to download it to a device. Then upload it to the folder using DAV folder on device. 

Have a Save to Files feature.

The current option is to download it to a device. Then upload it to the folder using DAV folder on device.

### Additional context

For those of us who want to have a solid records management system, having the emails stored in the related folder is important.
Consider a lawsuit where the disclosure is requested, emails are an important part of that disclosure. The easiest solution is to have the user save the emails to that topic in files.

madbob:
Fixes #675 
Fixes #9240


kesselb:
Thanks for the PR :+1:

lib/Controller/MessagesController.php
```diff
@@ -585,6 +587,72 @@ public function export(int $id): Response {
...
+		$fileExtension = 'eml';
+		$fullPath = "$targetPath/$fileName.$fileExtension";
+		$counter = 2;
+		while ($this->userFolder->nodeExists($fullPath)) {
+			$fullPath = "$targetPath/$fileName ($counter).$fileExtension";
+			$counter++;
+		}
+
+		$newFile = $this->userFolder->newFile($fullPath);
+		$newFile->putContent($source ?? '');
```
ChristophWurst: if source is null we should abort earlier

ChristophWurst (commented):
Tested & works

Thanks!

src/components/Envelope.vue
```diff
@@ -437,6 +445,17 @@
...
 					{{ t('mail', 'Download message') }}
 				</ActionLink>
+				<ActionButton
+					class="message-save-to-cloud"
+					:disabled="savingToCloud"
+					:close-after-click="true"
+					@click="() => isFilePickerOpen = true">
+					<template #icon>
+						<IconSave v-if="!savingToCloud" :size="20" />
+						<IconLoading v-else-if="savingToCloud" :size="20" />
```
ChristophWurst: The loading state is never visible because of the close-after-click=true

lib/Controller/MessagesController.php
madbob: Done.

src/components/Envelope.vue
madbob: I've preferred to remove IconLoading, as the menu immediately closes also when not `close-after-click`

ChristophWurst (approved):

ChristophWurst:
Conflicts :grimacing:

madbob:
Indeed: conflicts by myself!
Merged it.

PR #13257, opened 2026-07-06
fix: check tag while renaming

### LINKED ISSUE #9058: Renaming tags allows saving a tag with an existing name.
### Steps to reproduce

1. Open "edit tag" from action menu on a envelope/thread
2. Edit a tag's name and write same name as an existing one
3. See the tag being saved

### Expected behavior

The tag is not saved and the toats that it exists is shown

### Actual behavior

The tag is saved

### Mail app version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

Screenshot from 2023-11-09 13-15-42

madbob:
I've replicated the same checks done while creating a new tag also when renaming it (not a duplicate, not empty, not an hidden tag).

Please note that here I've also removed some unused code left in `TagModal` when `TagItem` was introduced (db38c4c).

Fixes #9058

### DIFF
--- a/src/components/TagItem.vue
+++ b/src/components/TagItem.vue
@@ -30,7 +30,9 @@
 			</NcColorPicker>
 			<ActionInput
 				v-if="renameTagInput"
-				:value="tag.displayName"
+				v-model="currentTagName"
+				:error="hasError()"
+				:helper-text="errorMessage"
 				@submit="renameTag(tag, $event)" />
 			<ActionText v-if="showSaving">
 				<template #icon>
@@ -70,7 +72,7 @@ import IconEdit from 'vue-material-design-icons/PencilOutline.vue'
 import DeleteIcon from 'vue-material-design-icons/TrashCanOutline.vue'
 import logger from '../logger.js'
 import useMainStore from '../store/mainStore.js'
-import { translateTagDisplayName } from '../util/tag.js'
+import { translateTagDisplayName, validateTag } from '../util/tag.js'
 
 export default {
 	name: 'TagItem',
@@ -106,13 +108,21 @@ export default {
 			tagLabel: true,
 			tagInput: false,
 			showSaving: false,
+			currentTagName: '',
 			renameTagLabel: true,
 			renameTagInput: false,
+			errorMessage: '',
 		}
 	},
 
 	computed: {
 		...mapStores(useMainStore),
+	},
+
+	watch: {
+		currentTagName() {
+			this.errorMessage = ''
+		},
 	},
 
 	methods: {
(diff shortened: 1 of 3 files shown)


kesselb:
Thanks for the PR :+1:

src/components/TagItem.vue
```diff
@@ -151,19 +152,29 @@ export default {
 		},
 
 		async renameTag(tag, event) {
-			this.renameTagInput = false
-			this.showSaving = false
 			const displayName = event.target.querySelector('input[type=text]').value
+			if (displayName.toLowerCase() in hiddenTags) {
+				showError(this.t('mail', 'Tag name is a hidden system tag'))
+				return
```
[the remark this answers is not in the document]
madbob: As validation logic is the same for `TagModal` and `TagItem` I've just created a new `validateTag()` shared function to avoid duplication, and apply the same improvements suggested by Copilot also for tags creation.

madbob:
By the way: is there any reason why `hiddenTags` is in `src/components/tags.js` and not in `src/util/tag.js`?

ChristophWurst:
> is there any reason why `hiddenTags` is in `src/components/tags.js` and not in `src/util/tag.js`?

Not that I'm aware of

src/components/TagItem.vue
```diff
@@ -151,19 +151,22 @@ export default {
 		},
 
 		async renameTag(tag, event) {
-			this.renameTagInput = false
+			const displayName = event.target.querySelector('input[type=text]').value.trim()
+			const valid = validateTag(tag.id, displayName)
+			if (valid !== true) {
+				showError(valid)
```
ChristophWurst: If I may be nitpicky, the error feedback as toast is not ideal. It comes only when you save and it's disconnected from the element you are interacting with.

How about leveraging the error state and helper text of the input? The components supports it - tested on the netlify docs page:

Image

src/util/tag.js
```diff
@@ -19,3 +21,22 @@ export function translateTagDisplayName(tag) {
...
+
+	if (testableDisplayName === '') {
+		return t('mail', 'Tag name cannot be empty')
+	}
+
+	if (Object.keys(hiddenTags).some((tag) => tag.toLowerCase() === testableDisplayName)) {
+		return t('mail', 'Tag name is a hidden system tag')
+	}
+
+	const mainStore = useMainStore()
```
ChristophWurst: I personally do not like to see the store being accessed here. The functions in this file are generally pure, free of side effects. This breaks the paradigm.
How about passing the tag list into `validateTag` so the caller handles the store access?

This would also make unit testing pleasant.

ChristophWurst (approved):
Tested and works!

Commented UX and architectural suggestions

ChristophWurst (approved):
Tested & works

Image

PR #13266, opened 2026-07-08
feat: message/reply templates

### LINKED ISSUE #6360: Message/reply templates

As a user of the app I might find myself replying with similar messages often. Therefore I would like to be able to define canned responses that prefill the editor.

1) Add a setting where each user can define their response templates.
2) Add a picker or shortcut to the editor that loads one of the templates.

N/a

### Additional context

We could take Github's responses as example.

If we go the extra mile there could be system-wide templates for any sort of message templates the whole organization needs. Admins would have to configure them.

steven-mpawulo:
Starts work on saving reusable message or response snippets for reuse in the composer.
Created MessageTemplate Entity, Mapper, Service and Controller.

fixes #6360


kesselb:
Thanks for your PR :+1:

PR #13274, opened 2026-07-09
Animate composer flying up on send

jancborchardt:
Part of the general animation initiative https://github.com/nextcloud/server/issues/47739

Nice animation for the composer modal flying out to the top on sending, like Apple Mail does. (Also then coming back from the top on clicking "Undo" to make it symmetrical.)

The animation starts instantly so the perceived performance is good and not blocked by other actions that can be done in the background, like the draft handling, moving to Outbox, etc.

If `prefers-reduced-motion` is set, of course there is no animation.

## Before

[Sending now.webm](https://github.com/user-attachments/assets/412f47c6-2706-406c-a4f7-3e2dbdbf00d4)

## After

[Send animation.webm](https://github.com/user-attachments/assets/b162020b-495c-4b5d-93c8-851c69f93c7b)

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/NewMessageModal.vue
+++ b/src/components/NewMessageModal.vue
@@ -5,9 +5,11 @@
 <template>
 	<Modal
 		v-if="showMessageComposer"
+		:class="{ 'composer-fly-up': flying, 'composer-fly-in': composerFlyIn }"
 		:size="modalSize"
 		:name="modalTitle"
 		:additional-trap-elements="additionalTrapElements"
+		@animationend.native="onComposerAnimationEnd"
 		@close="$event.type === 'click' ? onClose() : onMinimize()">
 		<div class="modal-content">
 			<div class="left-pane">
@@ -146,6 +148,9 @@ import useOutboxStore from '../store/outboxStore.js'
 import { messageBodyToTextInstance } from '../util/message.js'
 import { toPlain } from '../util/text.js'
 
+// Duration of the "fly up" send animation, matches var(--animation-slow) (300ms)
+const SEND_ANIMATION_DURATION = 300
+
 export default {
 	name: 'NewMessageModal',
 	components: {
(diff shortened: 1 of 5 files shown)


kra-mo (changes requested):
Hmm, looks pretty good but

1. The code looks way too complex for this. Not really my territory, but I'm sure it could be simplified.
2. It doesn't look nice that the dark fade behind the modal still disappears instantly. It should probably crossfade out while the modal is animating.

kra-mo:
Also, I wonder if we should change the send icon to ↑ like Apple Mail so the animation makes a bit more sense:

image

jancborchardt:
Updated, yeah looks better with the opacity fade:
(And code also simplified a bit)

[Screencast From 2026-07-09 16-49-09.webm](https://github.com/user-attachments/assets/2278e283-6788-4412-ba16-dcbd36fdf540)

jancborchardt:
> Also, I wonder if we should change the send icon to ↑ like Apple Mail so the animation makes a bit more sense:
> image

Not sure about this one. The "Sent" folder uses the "paper plane" icon as well, and it would be nice to use the same for that. And an "up arrow" would look quite odd there?

kra-mo:
> Not sure about this one. The "Sent" folder uses the "paper plane" icon as well, and it would be nice to use the same for that. And an "up arrow" would look quite odd there?

Apple Mail has the same split. Paper plane for folder and arrow for the action.

nimishavijay:
We don't have many sliding-type animations do we? So while I quite like the animation itself I'm worried it's a bit out of place with our design language. Also, the opacity fade could be a component-level thing too, it's nice! 

And FYI the floating composer is planned for the next release, so this won't be very visible in the first place? cc @ChristophWurst @GretaD

PR #13276, opened 2026-07-09
feat: unified attachments (as file and as link) from Files

### LINKED ISSUE #6080: Better UI for attaching items from File

There are 3 ways to attach files right now:
1. Upload an attachment from your local system
2. Attach a file from NC
3. Add a share link to a file on NC 

image

As @jancborchardt mentioned in https://github.com/nextcloud/mail/issues/5884#issuecomment-1048665891 it is confusing to have 2) and 3) as separate entries when they are both sharing files from NC, and other email clients that allow sharing from cloud (Gmail and Outlook) have just combined the 2 options.

We could have just one option in the action menu "Add attachment from Files" that opens the file picker.
In the file picker there could be 2 buttons 
- "Add share link"  
- "Add attachment"

I'd think that adding share link is a more common action in sharing cloud files, so that could be the primary button and add attachment could be the secondary button, but this is also kinda subjective.

This would probably streamline adding an attachment from Files.
What do you think?

_No response_

### Additional context

_No response_

madbob:
As `FilePickerVue` permits to customize action buttons, it is easy to pack multiple attachment stategies in a single modal.

Fixes #6080

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -350,14 +350,6 @@
 							t('mail', 'Add attachment from Files')
 						}}
 					</ActionButton>
-					<ActionButton :close-after-click="true" :disabled="encrypt" @click="onAddCloudAttachmentLink">
-						<template #icon>
-							<IconPublic :size="20" />
-						</template>
-						{{
-							t('mail', 'Add share link from Files')
-						}}
-					</ActionButton>
 				</Actions>
 
 				<Actions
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Thanks! That's a smooth improvement :sunglasses:

GretaD:
I think both buttons should be the same type. So both primary.

Screenshot from 2026-07-13 17-30-16

GretaD (approved):
approving so it doesnt get blocked, but please change the button type

madbob:
Done: both buttons are "primary"

PR #13282, opened 2026-07-10
feat: floating composer editor

### LINKED ISSUE #12011: Floating editor

As a Nextcloud Mail user I want to be able to composer an email while I navigate between other emails. The current editor requires me to close and re-open the modal all the time.

Have an editor that floats over the content. This could look similar to Gmail.

https://github.com/nextcloud/mail/issues/2831

### Additional context

https://github.com/nextcloud/mail/issues/2831

GretaD:
fixes: #12011
needs: https://github.com/nextcloud/contacts/pull/5583

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1890,7 +1890,6 @@ export default {
 			margin-bottom: 0;
 			background-color: transparent;
 			border: none;
-			opacity: 0.5;
 			padding: calc(var(--default-grid-baseline) * 2) calc(var(--default-grid-baseline) * 4);
 		}
 
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1849,8 +1849,7 @@ export default {
 	z-index: 100;
 	display: flex;
 	flex-direction: column;
-	height: 100%;
-	max-height: 100%;
+	min-height: 100%;
 }
 
 .composer-actions {
@@ -1935,7 +1934,7 @@ export default {
 
 // Make composer editor expand
 .message-editor {
-	flex: 1 1 100%;
+	flex: 1 1 auto;
 	min-height: 200px;
 	border-top: 1px solid var(--color-border);
 }
@@ -1964,6 +1963,17 @@ export default {
 .to-select :deep(.vs__selected-options) {
 	padding-inline-end: calc(var(--default-grid-baseline) * 10);
 }
+
+// NcSelect caps the toggle at 100px with overflow-y:auto, preventing expanded
+// chips from pushing CC/BCC down. The no-wrap/select--no-wrap CSS already
+// controls whether chips wrap (flex-wrap:nowrap on vs__selected-options), so
+// removing the height cap here is safe for both collapsed and expanded states.
+:deep(.v-select.select .vs__dropdown-toggle) {
+	max-height: none;
+	overflow-y: visible;
+}
+
+
 
 .recipient-overflow {
 	display: inline-flex;
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -43,20 +43,15 @@
 			</div>
 		</div>
 		<div class="composer-fields">
-			<div class="composer-fields__label">
-				<label class="to-label" for="to">
-					{{ t('mail', 'To') }}
-				</label>
-				<ButtonVue size="small" type="tertiary-no-background" @click.prevent="toggleViewMode">
-					{{ t('mail', 'Cc/Bcc') }}
-				</ButtonVue>
-			</div>
+			<label class="to-label" for="to">
+				{{ t('mail', 'To') }}
+			</label>
 			<div class="composer-fields--custom">
 				<NcSelect
 					id="to"
 					ref="toLabel"
 					:model-value="selectTo"
-					class="select"
+					class="select to-select"
 					:options="selectableRecipients.filter(recipient => !selectTo.some(to => to.email === recipient.email))"
 					:get-option-key="(option) => option.email"
 					:taggable="true"
@@ -68,7 +63,7 @@
 					:loading="loadingIndicatorTo"
 					:reducible="true"
 					:clearable="true"
-					:no-wrap="false"
+					:no-wrap="!toExpanded"
 					:append-to-body="false"
 					:create-option="createRecipientOption"
 					:clear-search-on-blur="() => clearOnBlur('to')"
@@ -83,12 +78,30 @@
 							class="vs__search"
 							v-bind="attributes"
 							v-on="events">
+						<ButtonVue
+							size="small"
+							type="tertiary-no-background"
+							class="copy-toggle"
+							@mousedown.stop
+							@click.prevent="toggleViewMode">
+							{{ t('mail', 'Cc/Bcc') }}
+						</ButtonVue>
 					</template>
 					<template #selected-option-container="{ option }">
 						<RecipientListItem
+							v-if="toExpanded || getRecipientIndex(selectTo, option) === 0"
 							:option="option"
 							class="vs__selected selected"
 							@remove-recipient="onRemoveRecipient(option, 'to')" />
+						<ButtonVue
+							v-else-if="getRecipientIndex(selectTo, option) === 1"
+							:key="option.email"
+							variant="tertiary"
+							class="vs__selected recipient-overflow"
+							@click.prevent.stop="toExpanded = true">
+							+{{ selectTo.length - 1 }}
+						</ButtonVue>
+						<span v-else />
 					</template>
 					<template #option="option">
 						<div>
@@ -116,7 +129,7 @@
 					:class="{ opened: !autoLimit }"
 					:options="selectableRecipients.filter(recipient => !selectCc.some(cc => cc.email === recipient.email))"
 					:get-option-key="(option) => option.email"
-					:no-wrap="false"
+					:no-wrap="!ccExpanded"
 					:filter-by="(option, label, search) => filterOption(option, label, search, 'cc')"
 					:dropdown-should-open="shouldOpenRecipientDropdown"
 					:taggable="true"
@@ -144,9 +157,19 @@
 					</template>
 					<template #selected-option-container="{ option }">
 						<RecipientListItem
+							v-if="ccExpanded || getRecipientIndex(selectCc, option) === 0"
 							:option="option"
 							class="vs__selected"
 							@remove-recipient="onRemoveRecipient(option, 'cc')" />
+						<ButtonVue
+							v-else-if="getRecipientIndex(selectCc, option) === 1"
+							:key="option.email"
+							variant="tertiary"
+							class="vs__selected recipient-overflow"
+							@click.prevent.stop="ccExpanded = true">
+							+{{ selectCc.length - 1 }}
+						</ButtonVue>
+						<span v-else />
 					</template>
 					<template #option="option">
 						<div>
@@ -172,7 +195,7 @@
 					:model-value="selectBcc"
 					class="select"
 					:class="{ opened: !autoLimit }"
-					:no-wrap="false"
+					:no-wrap="!bccExpanded"
 					:filter-by="(option, label, search) => filterOption(option, label, search, 'bcc')"
 					:options="selectableRecipients.filter(recipient => !selectBcc.some(bcc => bcc.email === recipient.email))"
 					:get-option-key="(option) => option.email"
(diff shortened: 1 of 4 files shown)


nimishavijay:
Really cool! Ik this is still a draft but getting some early feedback in:

- [ ] Position and resizing: 
  - [x] There is now an extra 56px gap below the bottom of the editor, it should not be there, and should just be `calc(var(--body-container-margin, 0px) + var(--default-grid-baseline))`
  - [x] The editor should automatically expand when text is added, right now it only expands when you click on the fullscreen button. We can always have `height: auto`. The original idea was to use the big editor modal when you click the fullscreen button (like in Gmail) but if time does not permit it can be a follow up. 
- [ ] Recipients:
  - [x] Adding more than 1 recipient directly shows a `+1` chip in the select component. Now that we have more space we can resize it when focused 
  - [ ] clicking on the `+1` expands the recipients but there is a bug where it overlaps the next field
  image
  image
  image


  - [x] There is no way to "unexpand". The ideal flow is: when the select field is focused, all recipients in that field are shown, when it is unfocused, the `+1` is shown
  - [x] There is not enough contrast between the `+1` and the background. I would remove all opacity properties and use only `--color-text-maxcontrast`
- [ ] Cc/Bcc button
  - [x] The size of the button ends up being a bit bigger than we need it to be. We can do in the `button` class: `padding: calc(var(--default-grid-baseline) * 1) calc(var(--default-grid-baseline) * 2);`
  - [x] The position of the button is always situated in the middle of the select component, so as the select component changes size it will move around. I find that removing in `copy-toggle` `top: 50%;` and `transform: translateY(-50%);` fixes it. 
  - [x] The button is also too light, we can remove the opacity property in `copy-toggle`

The rest are small alignment fixes that can be done later on also. Looking really cool! :)

GretaD:
Screenshot from 2026-07-31 14-32-11

jancborchardt (dismissed):
Very nice! :) Some details and thoughts on first look:
- [ ] Would the "New message" header change to whatever subject is put in? That would be nice, especially needed if minimized
- [ ] The "From", "To", "Cc" and "Bcc" labels are not vertically aligned with the text in the inputs, but too high up. (For multiline inputs, should be aligned with the first line of the input)
- [ ] The "Cc/Bcc" toggle would be nicer in the bottom right corner of the "To" field, so bottom-aligned from the current state. Then it potentially overlaps less content.
- [ ] All text on the left, from "New message" through To/Cc etc, Subject, Write message to "Draft saved" on the bottom should be left-aligned on one line.
- [ ] On the bottom, there is more bottom margin than right margin for the "Send" button. This should be even.

GretaD:
> * [ ]  The "From", "To", "Cc" and "Bcc" labels are not vertically aligned with the text in the inputs, but too high up. (For multiline inputs, should be aligned with the first line of the input)

thank you Jan, the first point is already done. The rest i will fix asap.

SebastianKrupinski:
Hey @GretaD 

Great work so far. But from a usability concept, here are my thoughts of what is missing:

1. ability to resize the editor (imagine having a ultra wide screen and being forced to edit on this tiny window )

image

2. ability to move the editor to another part of the screen. (Have a look at the image below, and now imaging how many times you would need to minimize and maximize the editor if you where referencing and email, but you can't see half of it)

image

GretaD:
@SebastianKrupinski  very good points, but i think those are for designers to decide, I only delivered what it was requested. Since this feature needs to be in soon, we can propose these points as follow-ups.

jancborchardt:
Considering @SebastianKrupinski's points – but @GretaD your call on whether it’s a follow-up or in here:

>     1. ability to resize the editor (imagine having a ultra wide screen and being forced to edit on this tiny window )

This makes sense, but not as manual resizing but better to use some percentage of the viewport. Of course with some max width for desktop and min-width for mobile.

>     2. ability to move the editor to another part of the screen. (Have a look at the image below, and now imaging how many times you would need to minimize and maximize the editor if you where referencing and email, but you can't see half of it)

This is for sure something for a follow-up, or rather to be discussed if and how we would want it.

GretaD:
The feature freeze is today, so i would like to have this one merged, and i will follow up soon after with everything else.  Reviewing-wise this is also better.

src/components/RecipientInfo.vue
```diff
@@ -91,8 +90,16 @@ export default {
...
-				this.expandedRecipients = this.recipients.map(() => false)
+			handler(newRecipients) {
+				const next = newRecipients.length
+				const current = this.expandedRecipients.length
+				if (next > current) {
+					for (let i = current; i < next; i++) {
+						this.$set(this.expandedRecipients, i, false)
+					}
+				} else if (next < current) {
+					this.expandedRecipients = this.expandedRecipients.slice(0, next)
```
odzhychko, line 114: issue: This logic expends/unexpends details unexpectedly when removed from the middle.

Example:

- add recipient A
- add recipient B
- add recipient C
- open expend details for recipient B
- remove remove recipient B
- details for recipient C expand (because they now have the same index as recipient B before)

src/components/RecipientInfo.vue
```diff
@@ -6,50 +6,49 @@
...
-							<span>{{ t('mail', 'Show more') }}</span>
-						</template>
+					<div class="recipient-info__name">
+						<strong>{{ recipient.label }}</strong>
+						<span v-if="recipient.email !== recipient.label" class="recipient-info__contact-email">
+							{{ recipient.email }}
+						</span>
 					</div>
+					<button class="recipient-info__expand-toggle" @click="toggleExpand(index)">
+						<IconArrowUp v-if="isExpanded(index)" :size="20" />
```
odzhychko: issue: I guess, now that the text is removed, the button the need a `title` or/and `aria-label` for a11y

src/components/NewMessageModal.vue
```diff
@@ -32,14 +36,40 @@
...
+					<template #icon>
+						<CloseIcon :size="20" />
+					</template>
+				</NcButton>
+			</div>
+		</div>
+		<div class="floating-composer__content">
+			<div v-if="showRecipientPane && !error && !warning && recipientPaneOpen" class="recipient-pane">
+				<NcButton
+					class="recipient-pane__close"
```
odzhychko, line 53: Has `:aria-label` but no `:title`. That's different to the 'Close composer' button. Not sure what is correct.

src/components/Composer.vue
```diff
@@ -83,12 +78,30 @@
...
+							{{ t('mail', 'Cc/Bcc') }}
+						</ButtonVue>
 					</template>
 					<template #selected-option-container="{ option }">
 						<RecipientListItem
+							v-if="toExpanded || getRecipientIndex(selectTo, option) === 0"
 							:option="option"
 							class="vs__selected selected"
 							@remove-recipient="onRemoveRecipient(option, 'to')" />
+						<button
```
hamza221: are you using a button here on purpose
```suggestion
						<ButtonVue
```

src/components/Composer.vue
```diff
@@ -144,9 +157,19 @@
 					</template>
 					<template #selected-option-container="{ option }">
 						<RecipientListItem
+							v-if="ccExpanded || getRecipientIndex(selectCc, option) === 0"
 							:option="option"
 							class="vs__selected"
 							@remove-recipient="onRemoveRecipient(option, 'cc')" />
+						<button
```
hamza221: same here

src/components/NewMessageModal.vue
```diff
@@ -188,16 +214,24 @@ export default {
...
 				email: '',
 			},
+
+			recipientPaneOpen: true,
 		}
 	},
 
+	watch: {
+		'composerData.to': function(newTo, oldTo) {
+			if (newTo?.length > 0 && !oldTo?.length) {
```
odzhychko, line 229: Not sure, what the intention of the code is.
Seems to be "(Re)open the recipient pane if the going from 0/none to one or more recipients."

Weirdly enough it doesn't work. Not sure why.

- Add recipient A
- Add recipient B
- Close pane with button
- Remove recipient A
- Remove recipient B
- Add recipient C
- Expected: Pane reopens
- Actual: Pane does not reopen

src/components/NewMessageModal.vue
```diff
@@ -648,70 +685,124 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-@use '../../css/variables.scss';
+$composer-width: 600px;
+$recipient-pane-width: 400px;
+$composer-height: 500px;
+$header-height: calc(var(--default-clickable-area) + calc(var(--default-grid-baseline) * 2));
+$bottom-offset: calc(var(--body-container-margin, 0px) + var(--default-grid-baseline));
+$panel-max-height: calc(100vh - (var(--body-container-margin, 0px) + var(--default-grid-baseline) * 3) - 40px);
```
hamza221, line 674: Not against it and not blocking. 
But afai see these are only use once, why not just use the value ?

src/components/NewMessageModal.vue
```diff
@@ -3,27 +3,31 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<Modal
+	<div
```
odzhychko, line 6: visual issue: Heading looks odd in mobile view. Overlays with Nextcloud menu...

Image

src/components/NewMessageModal.vue
```diff
@@ -290,8 +312,24 @@ export default {
...
+				return
+			}
+			if (this.$refs.floatingPanel.contains(event.target)) {
+				return
+			}
+			// Don't minimize when the user is interacting with a modal, popover, or CKEditor toolbar
+			if (event.target.closest('.modal-wrapper, .ck-body-wrapper, [data-popper-placement], .v-popper__popper')) {
+				return
+			}
+			this.onMinimize()
```
odzhychko: issue: Clicking outside the editor closes the editor. Should stay open.

> As a Nextcloud Mail user I want to be able to composer an email while I navigate between other > emails. The current editor requires me to close and re-open the modal all the time.
>
>#12809 

But maybe I missed some change in plans ^^'

odzhychko (commented):
@GretaD Found some things that might be logical issues/inconsistencies.

I think the biggest would be https://github.com/nextcloud/mail/pull/13282/changes#r3758848779

---

> The feature freeze is today, so i would like to have this one merged, and i will follow up soon after with everything else. Reviewing-wise this is also better.

I see nothing breaking that could not be followed up.

src/components/Composer.vue
```diff
@@ -43,20 +43,15 @@
...
 		<div class="composer-fields">
-			<div class="composer-fields__label">
-				<label class="to-label" for="to">
-					{{ t('mail', 'To') }}
-				</label>
-				<ButtonVue size="small" type="tertiary-no-background" @click.prevent="toggleViewMode">
-					{{ t('mail', 'Cc/Bcc') }}
-				</ButtonVue>
-			</div>
+			<label class="to-label" for="to">
```
odzhychko, line 46: visual issue (not sure if already mentioned)

- CC/BCC not aligned inside select
  - looks odd if hover over
-  Select text "Contact or email..." can cut into CC/BCC

Image

src/components/RecipientInfo.vue
```diff
@@ -6,50 +6,49 @@
...
-						:email="recipients[0].email"
-						:size="55"
-						:disable-tooltip="true"
-						:disable-menu="true"
-						:avatar="getAvatarForRecipient(recipients[0])" />
-				</div>
-				<div class="recipient-info__details">
-					<DisplayContactDetails :email="recipients[0].email" />
-				</div>
+			<div class="recipient-info__avatar">
```
odzhychko: visual issues (not sure if already mentioned by someone)

- close button and expand button do not align
- user name is shown twice

Image

odzhychko (commented):
The visual issue I noticed.

odzhychko (approved):

src/components/NewMessageModal.vue
hamza221: Should be awaited

src/components/NewMessageModal.vue
```diff
@@ -648,70 +685,124 @@ export default {
...
+$composer-height: 500px;
+$header-height: calc(var(--default-clickable-area) + calc(var(--default-grid-baseline) * 2));
+$bottom-offset: calc(var(--body-container-margin, 0px) + var(--default-grid-baseline));
+$panel-max-height: calc(100vh - (var(--body-container-margin, 0px) + var(--default-grid-baseline) * 3) - 40px);
+
+.floating-composer {
+	position: fixed;
+	bottom: $bottom-offset;
+	inset-inline-end: calc(var(--body-container-margin, 0px) + var(--default-grid-baseline));
+	z-index: 9999999;
```
hamza221, line 680: Would there be a way to assure it's on top ?

hamza221 (approved):
Nothing blocking for the release.
Should be easy fixes

src/components/NewMessageModal.vue
GretaD: done

src/components/Composer.vue
GretaD: done

src/components/Composer.vue
GretaD: done

src/components/NewMessageModal.vue
GretaD: fixed

src/components/RecipientInfo.vue
GretaD: fixed

src/components/RecipientInfo.vue
GretaD: fixed

src/components/NewMessageModal.vue
GretaD: fixed

PR #13285, opened 2026-07-13
feat: add link to file location

### LINKED ISSUE #13166: Add link to file location in confirmation dialog
### Feature category

Accessibility / workflow improvement

Image

On successful saving message attachment to files, include a link to the file location in the confirmation dialog. 

Clicking dialog should follow behavior of app or open a new tab with files.

_No response_

### Additional context

_No response_

steven-mpawulo:
Adds a clickable link to the saved file's location in the confirmation shown after saving a message attachment to Files.
Fixes #13166

### DIFF
--- a/src/components/MessageAttachment.vue
+++ b/src/components/MessageAttachment.vue
@@ -240,11 +240,10 @@ export default {
 					dir: saved?.path,
 				})
 
-				showSuccess(t('mail', '<div><p>Attachment saved to files</p><p><strong><u>Go to Files</u></strong></p></div>'), {
+				showSuccess(t('mail', '<div><p>Attachment saved to Files</p><p><a href="{fileLocationUrl}" target="_blank" rel="noopener noreferrer"><strong><u>Go to Files</u></strong></a></p></div>', {
+					fileLocationUrl,
+				}), {
 					isHTML: true,
-					onClick: () => {
-						window.open(fileLocationUrl, '_blank')
-					},
 				})
 			} catch (e) {
 				Logger.error('not saved', { error: e })
--- a/src/components/MessageAttachment.vue
+++ b/src/components/MessageAttachment.vue
@@ -240,8 +240,8 @@ export default {
 					dir: saved?.path,
 				})
 
-				showSuccess(t('mail', `Attachment saved to Files
- 					${fileLocationUrl}`), {
+				showSuccess(t('mail', '<div><p>Attachment saved to files</p><p><strong><u>Go to Files</u></strong></p></div>'), {
+					isHTML: true,
 					onClick: () => {
 						window.open(fileLocationUrl, '_blank')
 					},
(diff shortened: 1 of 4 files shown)


GretaD:
it works well, but i think it should not be a link that is shown on the popup, but maybe something like - Attachement saved to Files - Go to Files (and this has the link)

Screenshot from 2026-07-14 12-53-18


Gmail has this option:

Screenshot from 2026-07-14 12-51-01

steven-mpawulo:
Hello @GretaD 
Great suggestion.
Let me update.

src/components/MessageAttachment.vue
```diff
@@ -231,9 +232,20 @@ export default {
...
+					fileId: saved?.fileId,
+					dir: saved?.path,
+				})
+
+				showSuccess(t('mail', '<div><p>Attachment saved to files</p><p><strong><u>Go to Files</u></strong></p></div>'), {
+					isHTML: true,
+					onClick: () => {
+						window.open(fileLocationUrl, '_blank')
+					},
+				})
```
ChristophWurst: I have the impression that this is problematic for accessibility. Is there any way to display a real anchor element that appears clickable (also for screen readers) and has the native _blank target without a JavaScript handler?

src/components/MessageAttachment.vue
```diff
@@ -240,8 +240,8 @@ export default {
 					dir: saved?.path,
 				})
 
-				showSuccess(t('mail', `Attachment saved to Files
- 					${fileLocationUrl}`), {
+				showSuccess(t('mail', '<div><p>Attachment saved to files</p><p><strong><u>Go to Files</u></strong></p></div>'), {
```
GretaD: "Files" has to be capitalized to indicate the app.

src/components/MessageAttachment.vue
steven-mpawulo: You're absolutely right the onClick/window.open approach won't be announced as a link by screen readers. It'd be best to use a real anchor tag instead. Let me update.

src/components/MessageAttachment.vue
steven-mpawulo: Ohh, I hadn't seen that. 
I'm modifying.

src/components/MessageAttachment.vue
```diff
@@ -231,9 +232,19 @@ export default {
...
+				const saved = await saveAttachmentToFiles(id, this.id, path)
 				Logger.info('saved')
-				showSuccess(t('mail', 'Attachment saved to Files'))
+
+				const fileLocationUrl = generateUrl('/apps/files/files/{fileId}?dir={dir}&openfile=true', {
+					fileId: saved?.fileId,
+					dir: saved?.path,
+				})
+
+				showSuccess(t('mail', '<div><p>Attachment saved to Files</p><p><a href="{fileLocationUrl}" target="_blank" rel="noopener noreferrer"><strong><u>Go to Files</u></strong></a></p></div>', {
```
ChristophWurst: This is a tricky string for translators. I suggest to translate *Attachment saved to Files* and *Go to Files* without any HTML and build the HTML with these two translated strings. Don't forget to run them through HTML encoding so any special characters are handled.

Issue #13163, opened 2026-06-25
Opening messages is slow: Metadata endpoint


kesselb:
### Is your feature request related to a problem? Please describe.

Follow up for https://github.com/nextcloud/mail/issues/10384 and https://github.com/nextcloud/mail/issues/13162

### Is your feature request related to a problem? Please describe.

Opening a message currently fires separate backend requests for DKIM, itineraries, and (once deferred, #13162) the phishing check. Each request opens its own IMAP connection.

### Describe the solution you'd like

Merge DKIM, itineraries, and phishing into a single metadata request. The implementation must use one IMAP connection and reuse the message already fetched/cached for the body instead of re-fetching.

madbob:
I'm not really convinced that a single metadata request may solve the issue. Running multiple calls from client natively permit to execute them in parallel, and the bottleneck often is not the access to the IMAP mailbox (for DKIM, is DNS querying).

The body of the message may be temporarily cached for a short interval of time (using native Nextcloud's caching abilities), and multiple requests for the same message ID may be locked to avoid concurrency (while keeping parallel execution), but my question is: why not compute those stuff once at sync time on the database? At least DKIM validation is a boolean flag, and phishing detection is a bit more complex but still a JSON object.

madbob:
By the way: I've loosely profiled what happens when a message is selected and I've found a great amount of avoidable IMAP commands (which, of course, require time to be performed and slow down the whole).

Even before apply more radical approaches, some spare optimization may help...

## HordeImapClient::login()

#### Problem

Generates tens of `ID` commands, even when unnecessary. Internally, Horde calls `login()` very often (almost before any other operation) and everytime a new `ID` command is sent to the IMAP server.

#### Solution (?)

Evaluate `$this->_isAuthenticated` and proceed with the extra `ID` command only when initially `false`.

#### Starting from

https://github.com/nextcloud/mail/pull/12259

## ImapMessageFetcher

#### Problem

As different nuances of IMAP contents are required for each operation (especially in nested operations), `$this->client->fetch()` is called multiple times (sending multiple `UID FETCH` commands) with slightly different queries.

#### Solution (?)

Simplify `ImapMessageFetcher` to the base minimum: it is used only by `MessageMapper::findByIds()` and covers a very specific use case, which can be isolated and optimized.

kesselb:
> Evaluate $this->_isAuthenticated and proceed with the extra ID command only when initially false.

[...]

PR #13287, opened 2026-07-13
perf: reduce multiple ID IMAP commands by Horde client

madbob:
Fix to avoid multiple (and avoidable) ID commands to the IMAP server, as discussed in #13163


madbob:
I think the failed test is not related to actual submitted code...

kesselb (approved):

Issue #12096, opened 2025-11-28
Email signature – “Show more items” action is hidden


rabab81:
### Steps to reproduce

1.Integrating an email account
2.Email account settings
3.Signature menu
4.Add a rich text signature
5.Browser Firefox & edge & vivaldi

### Expected behavior

Being able to add a signature in rich text with access to all formatting parameters (bold text, color, etc.) so that I can modify the signature according to my needs.

<img width="624" height="355" alt="Image" src="https://github.com/user-attachments/assets/b292ab68-8fa5-457e-8e75-967a970315c3" />

### Actual behavior

All elements should be displayed and not hidden by the left-hand menu.
The workaround is to increase the browser's visibility to 170%.

<img width="1064" height="834" alt="Image" src="https://github.com/user-attachments/assets/3de4f671-452c-47c5-9701-da972f4217f9" />

### Mail app version

5.6.1

### Nextcloud version

32

### PHP engine version

None

### Web server

None

### Database

None

madbob:
Fixed in 0eaad6f

PR #13301, opened 2026-07-15
fix: signature editor on mobile, enlarged menu for formatting options

madbob:
In the Signature editor, the width of formatting options menu has been (correctly) limited to avoid overlaps with the configurations' modal sidebar (in 0eaad6f).

But on mobile it is squeezed to barely two columns of icons overflowing the menu itself. This PR defines a larger width for small screens.

Ref. #12096

### DIFF
--- a/src/components/SignatureSettings.vue
+++ b/src/components/SignatureSettings.vue
@@ -257,9 +257,15 @@ export default {
 .warning-large-signature {
 	color: darkorange;
 }
+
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Thanks! I'll check it out and test soon.

For visual changes we appreciate when PR authors can provide before/after screenshots. It's helpful for the reviewers to see the critical changes but also if the code changes later on raise any questions and we can go back to the fix PR and check why it was added.

madbob:
You are totally right...

sample

kesselb (approved):

PR #13129, opened 2026-06-19
perf(frontend): prevent overlapping background inbox syncs

hamza221:
The impact is probably tiny, just to avoid a few api calls that we now will respond with `MailboxLockedError` 

Also fix the cleanup hooks in Mailbox.vue and Outbox.vue, which used the Vue 3 "unmounted" name and therefore never ran under Vue 2.7; renamed to "destroyed" so the interval and event-bus teardown actually execute. 
->https://v2.vuejs.org/v2/guide/instance#Lifecycle-Diagram 

Assisted-by: ClaudeCode:claude-opus-4-8[1m]

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbox synchronization so only one sync run can happen at a time, reducing duplicate work and race conditions.
  * Fixed background cleanup behavior when leaving mailbox and outbox screens, helping keep polling and event handling from lingering unexpectedly.

PR #13303, opened 2026-07-15
fix(ui): rename beforeUnmount hooks back to beforeDestroy

### LINKED ISSUE #13031: Mail app shows a confirmation popup before leaving - beforeunload listener not removed when closing composer

ChristophWurst:
Vue 2.7 only invokes the Options API hook beforeDestroy; the Vue 3 name beforeUnmount is silently ignored. The premature rename (https://github.com/nextcloud/mail/pull/12590) meant cleanup code in these components never ran, leaking window listeners and event bus subscriptions. Most visibly, the composer's beforeunload handler survived after sending or discarding a message and blocked navigation with a bogus confirmation popup.

The hooks must be renamed again when the app actually switches to Vue 3.

Fixes #13031

Assisted-by: Claude:claude-fable-5

## How to test

1. Open the app
2. Send an email
3. Press F5

main: unload handler prevents reload
here: reload possible

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/Composer.vue
+++ b/src/components/Composer.vue
@@ -1129,7 +1129,7 @@ export default {
 		}
 	},
 
-	beforeUnmount() {
+	beforeDestroy() {
 		window.removeEventListener('mailvelope', this.onMailvelopeLoaded)
 	},
 
--- a/src/components/DisplayContactDetails.vue
+++ b/src/components/DisplayContactDetails.vue
@@ -35,7 +35,7 @@ export default {
 		}
 	},
 
-	async beforeUnmount() {
+	async beforeDestroy() {
 		if (this.vm) {
 			this.vm.$destroy()
 		}
--- a/src/components/EnvelopeList.vue
+++ b/src/components/EnvelopeList.vue
@@ -377,7 +377,7 @@ export default {
 		dragEventBus.on('envelopes-dropped', this.unselectAll)
 	},
 
-	beforeUnmount() {
+	beforeDestroy() {
 		dragEventBus.off('envelopes-dropped', this.unselectAll)
 	},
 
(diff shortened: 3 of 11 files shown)


madbob (approved):
Note for future context: the functions renamed from `destroyed` to `unmounted` in #12590 have been already renamed back in #13129

ChristophWurst:
Yep it was a bit out of sync, hence the "manual" revert and not a plain git revert.

PR #13304, opened 2026-07-16
fix(ui): archive shortcut in unified mailboxes

mrx8086:
I noticed that in the unified inbox ("All inboxes"), when I wanted to archive a message with the A key, I got an error, because the unified account itself has no archive folder, so I couldn't archive these mails from the unified inbox. Archiving the same message via the three-dot menu works, it is only the keyboard shortcut that fails.

This fix does the following: instead of looking at the unified account's archive folder, it looks at which account the email belongs to and uses that account's archive folder. That way the mail ends up in the correct archive from now on.

(I'm not a native English speaker — I wrote this in German and used AI only to translate it.)

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/Mailbox.vue
+++ b/src/components/Mailbox.vue
@@ -434,10 +434,14 @@ export default {
 					}
 
 					break
-				case 'arch':
+				case 'arch': {
 					logger.debug('archiving via shortcut')
 
-					if (this.account.archiveMailboxId === null) {
+					// In unified mailboxes this.account is the unified account which
+					// has no archive mailbox, so resolve the envelope's actual account
+					const account = this.mainStore.getAccount(env.accountId)
+
+					if (account.archiveMailboxId === null) {
 						showWarning(t('mail', 'To archive a message please configure an archive folder in account settings'))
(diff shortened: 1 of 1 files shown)


src/components/Mailbox.vue
```diff
@@ -434,10 +434,14 @@ export default {
 					}
 
 					break
-				case 'arch':
+				case 'arch': {
```
ChristophWurst, line 437: I don't think this block is necessary for the switch

ChristophWurst (commented):
Nice catch. Thank you!

ChristophWurst:
I have restored the pull request template https://github.com/nextcloud/.github/blob/master/.github/pull_request_template.md?plain=1
Please open PRs manually and not via the agent and use the template.

Thanks a lot

src/components/Mailbox.vue
ChristophWurst: nvm, needed for the `const`

ChristophWurst (approved):
Tested & works

Thank you

Issue #12491, opened 2026-02-26
OAuth2 support for mail app


jothoma1:
### SSO for mail app with Oauth2/OpenId compatible servers (like Stalwart)

Cannot find a way to work with Oauth2/OpenID SSO. 
OAuth2 against a third-party IdP should be possible
I have seen this https://docs.nextcloud.com/server/latest/admin_manual/groupware/mail.html#xoauth2-authentication-with-microsoft-azure-ad 
but cannot find in my nextcloud (version mail app 5.7.0 and with groupware) a way to work with something else than Microsoft or Gmail

### Describe the solution you'd like

What i want to do is using my mail server (Stalwart) which is OpenID compatible with SSO connection for my users:

    IdP: Keycloak (with a client id and client secret)
    Email Server: Stalwart (via its OAuth2)
    Email client: mail app from nextcloud 

Mail app redirects user to IdP for authentication, receives code, redeens into an access token, and uses access token.
My nextcloud already use OpenID for connection

JJJayKob:
I'm in the same boat as op just with a different mail server and this would be extremely useful! 

I hope this can be implemented fairly easily since its already possible to use google or microsoft.

jothoma1:
For information i saw this new nextcloud application : https://github.com/NK-IT-CLOUD/x2mail 
dont know if it works with stalwart but it seems to be exactly what i want

PR #13311, opened 2026-07-17
feat(provisioning): authenticate provisioned accounts with the user's OIDC token (XOAUTH2)

LeahAuroraV:
# OIDC Token Yoink

This PR extends the provisioning functionality of nextcloud mail to work with access and refresh tokens obtained on login through [user_oidc](https://github.com/nextcloud/user_oidc). 

There's a new "use the user's OIDC token" toggle on a provisioning config (in the groupware section under admin settings). accounts then authenticate via XOAUTH2 with no stored password. This token is mirrored from the user_oidc (first provider) session on interactive requests, and refreshed via the stored refresh token for background/cron sync (so it keeps working after the access token expires). It's kinda similar to how the existing Google/Microsoft integrations are already working for this app. The toggle is disabled and validation rejects it when user_oidc isn't installed, and the integration will stay disabled when user_oidc is not present.

It was developed with authentik as central identity provider to which both nextcloud and dovecot/postfix are connected. Dovecot is configured in such a way that it will accept XOAUTH2 as authentication. It has been verified against the local dev setup mentioned below, as well as a more extensive testing environment that adds ldap as user directory for the mail servers to the mix.

Because this feature requires quite a big dev setup I've added documentation to get a test environment up and running: [doc/oidc-xoauth2.md](https://github.com/LeahAuroraV/nc-mail/blob/feat/oidc-xoauth2/doc/oidc-xoauth2.md)

## Changes

- New `OidcIntegration` service that mirrors the user_oidc session token onto the account and refreshes it via the stored refresh token for background sync (`lib/Integration/OidcIntegration.php`, with Psalm stubs for the optional `user_oidc` classes so there's no hard dependency).
- Migration adding a nullable `oidc_enabled` column to `mail_provisionings`, plus the entity/mapper field and validation (existing configs are untouched).
- Provisioning wiring that provisions matching accounts as xoauth2 with no password and seeds/refreshes the token per request (`Service/Provisioning/Manager.php`, `Http/Middleware/ProvisioningMiddleware.php`).
- Token refresh before send: a new `BeforeSmtpClientCreated` event (IMAP already had something similar), dispatched by the SMTP factory and handled by the existing OAuth refresh listener alongside Google/Microsoft.
- Admin UI toggle in the provisioning settings, disabled with a hint when `user_oidc` isn't installed (`Settings/AdminSettings.php` + the two Vue components), plus unit tests for the new paths.

## 🤖 AI (if applicable)


ChristophWurst:
Thanks for the PR

lib/Integration/OidcIntegration.php
```diff
@@ -0,0 +1,270 @@
...
+ *  - Interactive requests: the current (auto-refreshed) login token is read from the
+ *    user_oidc session via ExternalTokenRequestedEvent and mirrored into the account row
+ *    (ProvisioningMiddleware and refresh()).
+ *  - Background jobs: refresh() renews the mirrored token at the IdP's token endpoint
+ *    using the stored refresh token, exactly like the Google/Microsoft integrations.
+ *
+ * The IdP token endpoint and client credentials are read from the user_oidc provider
+ * configuration; this integration is inert when user_oidc is not installed.
+ */
+class OidcIntegration {
```
ChristophWurst, line 37: To manage the complexity of this change I'd favor getting individual OIDC support in first (and close https://github.com/nextcloud/mail/issues/12491), and then build OIDC provisioning on top.

lib/Integration/OidcIntegration.php
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	/**
+	 * Whether accounts can be provisioned with OIDC token authentication, i.e. the
+	 * user_oidc app is installed, enabled and exposes the token API this integration
+	 * builds on.
+	 */
+	public function isAvailable(): bool {
+		return $this->appManager->isEnabledForAnyone('user_oidc')
+			&& class_exists(\OCA\UserOIDC\Event\ExternalTokenRequestedEvent::class);
```
ChristophWurst, line 59: *\OCA\** is the private API of another app. This is not supposed to be used by other apps to avoid inter app dependencies. The clean way is going through a OCP public API that lives in Nextcloud server and has a strong stability guarantee.

tests/stubs/oca_useroidc.php
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+// Psalm stubs for the optional user_oidc app (https://github.com/nextcloud/user_oidc).
+// Only the surface used by OCA\Mail\Integration\OidcIntegration is declared.
+
+namespace OCA\UserOIDC\Model {
```
ChristophWurst, line 13: Without a mechanism to update these stups they give a false sense of safety. Psalm will continue to pass in our repository when the user_oidc APIs change (break).

ChristophWurst (commented):
Quick architectural review

Thank you

lib/Integration/OidcIntegration.php
LeahAuroraV: Thanks for the quick review!

So basically create an oauth client within the mail app in which users can choose OpenID as option when adding a mailbox, with a well-known discovery url (or the individual endpoints that would contain), client id and client secret, and then have the open id authentication open in a new tab?

Then the provisioning on top would be defining those urls, client id and secret in the admin settings, rather than using the the user_oidc app.

lib/Integration/OidcIntegration.php
LeahAuroraV: https://github.com/nextcloud/user_oidc/blob/main/lib/Event/ExternalTokenRequestedEvent.php does specifically mention in its docblock that it is meant for external apps.

lib/Integration/OidcIntegration.php
LeahAuroraV: Or would the clients always be global settings?

ChristophWurst: We already have XOAUTH2 support for Google and Microsoft. Admins set the client id and secret. When a user sets up an account we detect if oauth should be used based on the IMAP server.

For OIDC I could imagine the same. Admin sets client id, secret and server hostname. When a user then creates a new account for that hostname it will open the auth popup for that server.

ChristophWurst:
Putting this into draft state until https://github.com/nextcloud/mail/pull/13317 is done and merged

PR #13314, opened 2026-07-18
fix: avatars in dashboard widgets

### LINKED ISSUE #11634: Avatars not shown in Dashboard widget and message detail
### Steps to reproduce

1. Get an email from a contact which has a photo set in contacts
2. Check that mail in the Dashboard and in Mail

### Expected behavior

The avatar shows in all and any places.

### Actual behavior

The proper picture only shows in the message list, as well as in the message header (where the participant chips are shown).

The placeholder shows in the Dashboard widgets for both Important mail and Unread mail:

Image

Image

And neither in the detail list of open messages:
Image

### Mail app version

5.3.1

### Nextcloud version

31.0.8

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

madbob:
Fixes #11634 

Warning: those changes do not address apparent issues caused by the avatars' cache: just changing an image may not provide immediate results until the cache is expired and recalculated.

## Summary by CodeRabbit

* **New Features**
  * Mail dashboard items now display sender avatars when available.
  * External avatars are supported with appropriate URL handling.
  * A fallback guest avatar is shown when no sender avatar is available.

* **Tests**
  * Added coverage for avatar retrieval and display in important mail items.

### DIFF
--- a/lib/Dashboard/MailWidget.php
+++ b/lib/Dashboard/MailWidget.php
@@ -9,7 +9,9 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Dashboard;
 
+use OCA\Mail\Address;
 use OCA\Mail\AppInfo\Application;
+use OCA\Mail\Contracts\IAvatarService;
 use OCA\Mail\Contracts\IMailSearch;
 use OCA\Mail\Db\Message;
 use OCA\Mail\Exception\ClientException;
@@ -40,6 +42,7 @@ abstract class MailWidget implements IAPIWidget, IAPIWidgetV2, IIconWidget, IOpt
 		IL10N $l10n,
 		IURLGenerator $urlGenerator,
 		IUserManager $userManager,
+		private IAvatarService $avatarService,
 		protected AccountService $accountService,
 		protected IMailSearch $mailSearch,
 		IInitialState $initialState,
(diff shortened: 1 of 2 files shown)


lib/Dashboard/MailWidget.php
```diff
@@ -128,6 +130,42 @@ protected function getMailboxIdsToExclude(string $userId): array {
 		return array_values(array_filter($mailboxIdsToExclude));
 	}
 
+	protected function getAvatar($userId, $from) {
```
ChristophWurst:
```suggestion
	protected function getAvatar(string $userId, ?string $from): ?string {
```

lib/Dashboard/MailWidget.php
```diff
@@ -128,6 +130,42 @@ protected function getMailboxIdsToExclude(string $userId): array {
...
+		$url = null;
+
+		if ($from) {
+			$email = $from->getEmail();
+
+			$avatar = $this->avatarService->getAvatar($email, $userId);
+			if ($avatar) {
+				if ($avatar->isExternal()) {
+					$url = $this->urlGenerator->getAbsoluteURL(
```
ChristophWurst: return the result right away

lib/Dashboard/MailWidget.php
```diff
@@ -128,6 +130,42 @@ protected function getMailboxIdsToExclude(string $userId): array {
...
+						$this->urlGenerator->linkToRoute('mail.avatars.image', [
+							'email' => $email,
+						])
+					);
+				} else {
+					$url = $avatar->getUrl();
```
ChristophWurst: same, just return. the else block around will no longer be necessary

lib/Dashboard/MailWidget.php
```diff
@@ -128,6 +130,42 @@ protected function getMailboxIdsToExclude(string $userId): array {
...
+				}
+			}
+		}
+
+		if ($url == null) {
```
ChristophWurst: impossible with the instant returns

lib/Dashboard/MailWidget.php
```diff
@@ -128,6 +130,42 @@ protected function getMailboxIdsToExclude(string $userId): array {
...
+			$url = $this->urlGenerator->getAbsoluteURL(
```
ChristophWurst: return the result right away

ChristophWurst (changes requested):
Thanks!

lib/Dashboard/MailWidget.php
madbob: `$from` is not a string, is a `OCA\Mail\Address`!
Anyway: fixed with typed parameters and return type (`string`, not `?string`, as none of the invoked functions is expected to return `null`).

lib/Dashboard/MailWidget.php
madbob: Done

lib/Dashboard/MailWidget.php
madbob: Done

lib/Dashboard/MailWidget.php
madbob: Done

ChristophWurst:
@coderabbitai review

PR #13315, opened 2026-07-18
fix: visualization of multiple attachments

### LINKED ISSUE #10408: Mail with lots of attachments not readable, overlays text
### Steps to reproduce

I received a mail with 27 attached files. They completely overlay the message.

Image

### Expected behavior

The text should be readable first, then maybe have a drop-down below/above the first 2 attachments, which says "so many more files"

### Actual behavior

The text of the mail is not readable, I had to click 'reply' to read the message.

### Mail app version

4.0.3

### Nextcloud version

30.0.1

### Mailserver or service

strato

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

madbob:
The original issue occurs on mobile, when expanding the "more attachments" area. I've introduced an explicit scrollable area that permits to navigate them.

The most invasive modification here is the change of the background color, which was a linear gradient and now is a static color. A linear gradient is proportional, and the height of the transparent portion is unpredictable as may vary accordingly to the number of attachments. With many attachments, the transparent area growns and interact badly with the overlapped contents of the message.

I've also introduced some minor adjustments to bottom buttons ("View more attachments", "Show all on Files", "Download zip") alignment on mobile.

(Don't care broken attachment previews in my screencasts: I'm breaking stuff around :stuck_out_tongue: )

Original state:

https://github.com/user-attachments/assets/33291535-fe66-47f5-9985-326ec104fb82

New state:

https://github.com/user-attachments/assets/de0b6742-10e1-4648-9703-29a3d965b9fa

Fixes #10408

### DIFF
--- a/src/components/MessageAttachments.vue
+++ b/src/components/MessageAttachments.vue
@@ -23,19 +23,26 @@
 					@open="showViewer(fileInfos[idx])" />
 			</div>
 		</div>
-		<div
+
+		<NcButton
 			v-if="hasNextLine"
-			class="show-more-attachments"
+			variant="tertiary"
+			size="small"
+			:disabled="savingToCloud"
+			class="attachment-button"
 			@click="isToggled = !isToggled">
-			<ChevronDown v-if="isToggled" :size="20" />
-			<ChevronUp v-if="!isToggled" :size="20" />
+			<template #icon>
+				<ChevronDown v-if="isToggled" />
+				<ChevronUp v-if="!isToggled" />
+			</template>
 			<span v-if="isToggled">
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Tested with responsive mode of FF for Linux & works well

The only nitpick is consideration of using multiple of `var(--default-grid-baseline)` instead of magic numbers

madbob:
Indeed, you are so right that I've just removed the existing clickable elements and have replaced them with proper `NcButton` for improved themability and accessibility. So there is no need at all to hard-wire spacing and margins for them.

ChristophWurst (approved):
Tested & works still :)

PR #13305, opened 2026-07-16
fix: subfolder selection for quick actions

madbob:
Fixes #11790

PR #13318, opened 2026-07-18
fix: improved widget for mailbox selection

### LINKED ISSUE #8726: Default folder selection jumps
### Steps to reproduce

1. Open the app
2. Add an account
3. Open the account settings
4. Scroll to *Default folders*
5. Assign/unassign mailboxes

### Expected behavior

Steady view

### Actual behavior

https://github.com/nextcloud/mail/assets/1374172/c4e9b3a4-c0ee-411e-8518-4da397a55a01

It's possibly the disabled state that is slighly smaller and therefore makes the view jump/move at the bottom

### Mail app version

3

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

cc @GretaD

madbob:
Diving into #8726 I stumbled upon `MailboxInlinePicker`, the only user of [@riophae/vue-treeselect](https://www.npmjs.com/package/@riophae/vue-treeselect), a not-so-lightweight dependency which did not received any update in the last 7 years.

So I rewrote it using the more familiar `NcSelect`, which do not permit nested navigation and selection of subfolders. Let you evaluate if this is an acceptable loss.

This PR includes:
- major rewrite of `MailboxInlinePicker`, as specified above
- major rewrite of `AccountDefaultsSettings`, to adapt to the new `MailboxInlinePicker`, remove most of the duplicated code around, and separate the flags to disable the single inputs while saving (which was the original request of #8726)
- minor review of `ActionFileinto`, also to be adapted to the new `MailboxInlinePicker`
- remove of all specific CSS style targeting `vue-treeselect`

My only concern is about the filter on `mailboxHasRights`, which has been here removed. It is not entirely clear to me what kind of check was here performed, but if it was required perhaps the same elaboration should be added also to Quick Folders selections (cfr. #13305) and similar folder/subfolder selections.

Fixes #8726

Note: if this PR is merged, I recommend to remove then the dependency from `@riophae/vue-treeselect` in `package.json`: I avoided to change that myself in order to not mess up with `package-lock.json`

### DIFF
--- a/src/components/AccountDefaultsSettings.vue
+++ b/src/components/AccountDefaultsSettings.vue
@@ -4,42 +4,52 @@
 -->
 
 <template>
-	<div>
-		<p>
-			{{ t('mail', 'Drafts are saved in:') }}
-		</p>
-		<MailboxInlinePicker v-model="draftsMailbox" :account="account" :disabled="saving" />
+	<div class="default-folders">
+		<MailboxInlinePicker
+			:label="t('mail', 'Drafts are saved in:')"
+			:value="account.draftsMailboxId"
+			:account="account"
+			:disabled="saving.draftsMailboxId"
+			@update="(folderId) => updateFolder('draftsMailboxId', folderId)" />
 
-		<p>
-			{{ t('mail', 'Sent messages are saved in:') }}
-		</p>
+		<MailboxInlinePicker
+			:label="t('mail', 'Sent messages are saved in:')"
+			:value="account.sentMailboxId"
+			:account="account"
+			:disabled="saving.sentMailboxId"
+			@update="(folderId) => updateFolder('sentMailboxId', folderId)" />
 
-		<MailboxInlinePicker v-model="sentMailbox" :account="account" :disabled="saving" />
-		<p>
-			{{ t('mail', 'Deleted messages are moved in:') }}
-		</p>
+		<MailboxInlinePicker
+			:label="t('mail', 'Deleted messages are moved in:')"
+			:value="account.trashMailboxId"
+			:account="account"
+			:disabled="saving.trashMailboxId"
+			@update="(folderId) => updateFolder('trashMailboxId', folderId)" />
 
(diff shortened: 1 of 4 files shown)


ChristophWurst:
> which did not received any update in the last 7 years

:face_with_peeking_eye: good catch

kesselb:
> My only concern is about the filter on mailboxHasRights, which has been here removed. It is not entirely clear to me what kind of check was here performed, but if it was required perhaps the same elaboration should be added also to Quick Folders selections (cfr. https://github.com/nextcloud/mail/pull/13305) and similar folder/subfolder selections.

Context: IMAP ACL https://datatracker.ietf.org/doc/html/rfc4314.html#section-2.1

A mailbox can have a list of acls. If that's the case (often for shared mailboxes), then we need to check if the mailbox has "i" otherwise we cannot use it as target.

src/components/AccountDefaultsSettings.vue
```diff
@@ -4,42 +4,52 @@
...
-		<p>
-			{{ t('mail', 'Junk messages are saved in:') }}
-		</p>
-		<MailboxInlinePicker v-model="junkMailbox" :account="account" :disabled="saving" />
+	<div class="default-folders">
+		<MailboxInlinePicker
+			:label="t('mail', 'Drafts are saved in:')"
+			:value="account.draftsMailboxId"
+			:account="account"
+			:disabled="saving['draftsMailboxId']"
```
kesselb:
```suggestion
			:disabled="saving.draftsMailboxId"
```

kesselb (commented):
Thanks for taking care 👍 

Please add the mailboxHasRights check back.

src/components/AccountDefaultsSettings.vue
madbob: Fixed

madbob:
> Please add the mailboxHasRights check back.

Added.

And this also replies to my question about Quick Actions: the check about permissions have to be applied also in that context. Or `MailboxInlinePicker` can be directly used also there. I will submit a PR in the next days.

PR #13342, opened 2026-07-22
refactor(config): add config lexicon for app config keys

ChristophWurst:
Declare the Mail app's appconfig keys in a ConfigLexicon with their type, default and definition, and route call sites through key constants and the typed IAppConfig API instead of scattered string literals and mixed accessors.

Scope is appconfig values only. User preferences (stored via IUserPreferences with hyphenated keys) remain out of scope and unchanged.

Ref https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/config/lexicon.html

## 🤖 AI (if applicable)

### DIFF
--- a/tests/Unit/Service/ContextChat/ContextChatSettingsServiceTest.php
+++ b/tests/Unit/Service/ContextChat/ContextChatSettingsServiceTest.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Tests\Unit\Service\ContextChat;
+
+use ChristophWurst\Nextcloud\Testing\TestCase;
+use OCA\Mail\AppInfo\Application;
+use OCA\Mail\ConfigLexicon;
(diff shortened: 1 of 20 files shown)


tests/Unit/Service/ContextChat/ContextChatSettingsServiceTest.php
```diff
@@ -0,0 +1,78 @@
...
+		$this->service = new ContextChatSettingsService(
+			$this->preferences,
+			$this->appConfig,
+		);
+	}
+
+	public function testIsIndexingEnabledByDefault(): void {
+		$this->appConfig->expects(self::once())
+			->method('getValueBool')
+			->with(Application::APP_ID, 'index_context_chat_default', false)
```
kesselb: Why would you not use the const also in the tests?

kesselb (approved):

SebastianKrupinski:
Lol. I started a PR for the DAV app to do the exact same thing yesterday 🤣

tests/Unit/Service/ContextChat/ContextChatSettingsServiceTest.php
SebastianKrupinski: Good point

DerDreschner: Agree, makes it easier to understand the whole impact of a config key.

ChristophWurst:
Tests are now using the constants too

PR #13346, opened 2026-07-23
fix: Correct hasLocks() expiry check

kesselb:
Extracted from https://github.com/nextcloud/mail/pull/13208

hasLocks() used `||` where it should be `&&` between "lock is set" and "lock is still within LOCK_TIMEOUT", so any non-null lock was treated as active forever instead of expiring after 5 minutes. Whenever a lock field is set, the first term alone already makes the condition true, so the timeout comparison can never actually flip the result.

Assisted-by: Claude:claude-sonnet-5

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/Mailbox.php
+++ b/lib/Db/Mailbox.php
@@ -131,13 +131,13 @@ class Mailbox extends Entity implements JsonSerializable {
 	}
 
 	public function hasLocks(int $now): bool {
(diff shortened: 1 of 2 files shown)


madbob (approved):

ChristophWurst (approved):
🙈

PR #13356, opened 2026-07-25
feat: text blocks empty state

madbob:
When no Text Blocks are defined, the picker modal accesible from Composer is a bit confusing as requires to choose something that do not exists.

This PR includes a basic message, aimed to increase discoverability of the Text Blocks feature.

preview

Plus: I've moved the buttons of `TextBlockModal` to be handled as native `NcDialog` buttons. I must confess it is not entirely clear to me if there is some policy about this functionality of `NcDialog`: sometime buttons are managed with dialog's proper property, sometime those are just hard-coded into the modal's body (often with custom CSS style to force alignment). In my opinion usage of native `buttons` property improves themeability, but guidance and suggestions are appreciated.

### DIFF
--- a/src/components/textBlocks/TextBlockModal.vue
+++ b/src/components/textBlocks/TextBlockModal.vue
@@ -3,7 +3,12 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<NcDialog id="text-block-picker" :name="t('mail', 'Text blocks')" @closing="handleClose">
+	<NcDialog
+		v-if="hasTextBlockes"
+		id="text-block-picker"
+		:name="t('mail', 'Text blocks')"
+		:buttons="saveButtons"
+		@closing="handleClose">
 		<p>{{ t('mail', 'Choose a text block to insert at the cursor') }}</p>
 		<ListItem
 			v-for="textBlock in getMyTextBlocks()"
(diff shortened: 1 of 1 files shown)


src/components/textBlocks/TextBlockModal.vue
```diff
@@ -20,22 +25,20 @@
...
-			<template #icon>
-				<IconCheck :size="20" :name="t('mail', 'Insert text block')" />
-			</template>
-		</NcButton>
+	</NcDialog>
+	<NcDialog
+		v-else
+		id="text-block-picker"
+		:name="t('mail', 'Text blocks')"
+		:message="t('mail', 'Text blocks are reusable pieces of text that can be inserted in messages. Visit the Settings panel to create your owns.')"
```
kesselb:
```suggestion
		:message="t('mail', 'Text blocks are reusable pieces of text that can be inserted in messages. Visit the Settings panel to create your own.')"
```

kesselb:
Thanks a lot, good initiative to migrate to NcDialog.

kesselb (commented):
Thx 👍 

Is the loading state still working?

madbob:
> Is the loading state still working?

Which loading state? I don't see that in the original `TextBlockModal`...

Fixed the typo in the text.

src/components/textBlocks/TextBlockModal.vue
madbob: Fixed.

kesselb (approved):

PR #13358, opened 2026-07-25
fix(imap): surface auth failures to the client and make the rate limiter self-healing

### LINKED ISSUE #7279: Invalid passwords or IMAP login errors do not trigger a comprehensive error message for existing accounts
### Steps to reproduce

1. Set up an account in mail
2. log in
3. change password or shut down your IMAP server
4. try to open mail app

### Expected behavior

The mail app tells you what's wrong

### Actual behavior

No new emails and erros are shown

### Mail app version

1.13.5

### Mailserver or service

Zimbar

### Operating system

unknown

### PHP engine version

_No response_

### Web server

_No response_

### Database

_No response_

### Additional info

_No response_

### LINKED ISSUE #12136: Ask user for new password if connection to mailbox failes because of wrong password

Connected to #7279
When adding additional mailboxes into Nextcloud Mail the password of the mailbox gets stored. If this password is changed outside of Nextcloud Mail (for example a password change is requested by the mail provider) Nextcloud Mail at the moment shows the message `Connection failed. Please verify your information and try again`.

There should exist some easy way to insert the new password - at the moment the user has to open the account settings, goto "Mail server" and insert the new password.
From a user point of view when trying to open the mailbox (maybe when clicking on the title or the error message) the user should directly be asked to insert the new password.

_No response_

### Additional context

This feature would also give some options to administrators when deploying new mailboxes to the users: at the moment only one mailbox can be automatically created by the provisioning but companies often have shared mailboxes that more than one user has to use. When the user gets asked for his password the admin could deploy the mailboxes using `occ mail:account:create`.

alexdimarco:
Fixes #7279
Fixes #12136

### Problem

When IMAP rejects an account's credentials mid-session (changed/expired password, or a transient failure of the mail server's auth backend), the resulting `Horde_Imap_Client_Exception` with `LOGIN_AUTHENTICATIONFAILED` bubbles up to `ErrorMiddleware` as a generic HTTP 500 "Server error". The web client cannot distinguish this from any other failure, so:

- the background sync loop in `App.vue` keeps polling `mailboxes#sync` every 30 seconds forever,
- the user gets no hint that their credentials are wrong (bodies and attachments just stop loading),
- the server log fills with error-level entries — we measured **149 entries/hour** from a single affected account, and
- the client-side auth rate limiter compounds it: after 3 failures it hard-blocks the account for the rest of a 3-hour window, and the account-settings **Save** cannot recover with an unchanged password because its own connectivity test is rate-limited by the same cache key (the lockout concern described in #7279).

We hit exactly this in production on 2026-07-25: a mail-server backend blip rejected *valid* credentials a handful of times within an hour, and the account stayed dead for hours with no actionable feedback.

### What this PR does

**Commit 1 — surface the failure and stop the doomed polling:**
- New `ImapAuthenticationFailedException` (a `ClientException`, HTTP **424 Failed Dependency**) carrying a `reason` of `AUTHENTICATION_FAILED` or `RATE_LIMITED`. 424 was chosen over 401 deliberately: the Nextcloud session is fine, only the IMAP dependency failed, and 401 would trigger the session-expiry machinery.
- `ErrorMiddleware` walks the exception chain of any uncaught exception from a `#[TrapError]` method; a Horde `LOGIN_AUTHENTICATIONFAILED` anywhere in the chain becomes the typed jsend `fail` payload, logged **once at info level** instead of error. Covers sync, body and attachment endpoints with one hunk. `getDkim` gets the previously missing `#[TrapError]`.
- `ImapToDbSynchronizer::sync()`'s unguarded `$client->login()` is wrapped in a `ServiceException`; `SyncJob` detects auth failures anywhere in the chain so cron logs them at info too.
- Frontend: the new type maps in `convert.js`; on auth failure the account is flagged via the existing `account.error` banner in `Navigation.vue`, a persistent toast names the account, background sync skips flagged accounts, and IMAP-backed side lookups (DKIM/itineraries/smart replies) are skipped while flagged. Initial syncs fail loudly (individual and unified/priority fan-out) so `initializeCache` cannot loop requesting envelopes that can never be cached; `initializeCache()` now returns its promise chain so failures reach the error UI, which shows an actionable message instead of "Not found".

**Commit 2 — make the rate limiter self-healing (leaky bucket):**
Instead of hard-blocking for the rest of a 3-hour window, the limiter keeps allowing real attempts on a schedule: 3 unthrottled attempts, one immediate retry, then one per 30 seconds (×3), then one per 5 minutes. A successful login clears all throttle state. The frontend probe scheduler mirrors the cadence, so a transient server-side failure now heals within seconds-to-minutes of the server recovering — no manual cache surgery, no page reload — and the settings Save always recovers within one interval.

**Trade-off, made deliberately:** against a server that keeps rejecting (genuinely wrong credentials) the attempt budget rises from 3 per 3h to roughly one gentle attempt per 5 minutes with background sync paused — still far below abusive rates, while cutting self-recovery after transient failures from up to 3 hours to at most 5 minutes. Happy to discuss a slower tail if preferred.

### Test plan

- New unit tests: `ErrorMiddleware` translation (raw/wrapped/doubly-wrapped/rate-limited), and `HordeImapClient` schedule pinning (threshold arithmetic, 30s/5min anchors, success reset, both counted rejection messages incl. `Mail server denied authentication.`); the integration rate-limit test is updated for the new semantics.
- Full unit suite green against a Nextcloud 34 server checkout.
- Validated on a real production outage (Nextcloud 34 / backported to 5.10.9): banner + toast appear, polling stops (network tab quiet apart from probes), a single info line per request in the log, and automatic recovery once the server accepted logins again.

### Notes for reviewers

- The `RATE_LIMITED` reason relies on the limiter's exact `'Too many auth attempts'` message (same coupling as `CouldNotConnectException::getReason()`); happy to move both onto a shared constant.
- A throttled settings-Save currently surfaces as generic `CONNECTION_ERROR`; a rate-limit-specific hint would be a nice follow-up.

### DIFF
--- a/lib/BackgroundJob/SyncJob.php
+++ b/lib/BackgroundJob/SyncJob.php
@@ -8,10 +8,9 @@ declare(strict_types=1);
 
 namespace OCA\Mail\BackgroundJob;
 
-use Horde_Imap_Client_Exception;
 use OCA\Mail\AppInfo\Application;
+use OCA\Mail\Exception\ImapAuthenticationFailedException;
 use OCA\Mail\Exception\IncompleteSyncException;
-use OCA\Mail\Exception\ServiceException;
 use OCA\Mail\IMAP\MailboxSync;
 use OCA\Mail\Service\AccountService;
 use OCA\Mail\Service\Sync\ImapToDbSynchronizer;
@@ -123,9 +122,7 @@ class SyncJob extends TimedJob {
 				'exception' => $e,
 			]);
 		} catch (Throwable $e) {
-			if ($e instanceof ServiceException
-				&& $e->getPrevious() instanceof Horde_Imap_Client_Exception
-				&& $e->getPrevious()->getCode() === Horde_Imap_Client_Exception::LOGIN_AUTHENTICATIONFAILED) {
+			if (ImapAuthenticationFailedException::findHordeAuthFailure($e) !== null) {
 				$this->logger->info('Cron mail sync authentication failed for account {accountId}', [
 					'accountId' => $accountId,
 					'exception' => $e,
--- a/lib/Controller/MessagesController.php
+++ b/lib/Controller/MessagesController.php
@@ -297,6 +297,7 @@ class MessagesController extends Controller {
 	 * @param int $id
 	 * @return JSONResponse
 	 */
+	#[TrapError]
 	public function getDkim(int $id): JSONResponse {
 		if ($this->userId === null) {
 			return new JSONResponse([], Http::STATUS_UNAUTHORIZED);
--- a/lib/Exception/ImapAuthenticationFailedException.php
+++ b/lib/Exception/ImapAuthenticationFailedException.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Exception;
+
+use Horde_Imap_Client_Exception;
+use OCP\AppFramework\Http;
+use Throwable;
+
+/**
+ * IMAP rejected the account's credentials (or the local auth rate limiter is
+ * engaged). Unlike a generic ServiceException this is sent to the client as a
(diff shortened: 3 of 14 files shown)


kesselb:
Thanks for your pr.

Issue #10977, opened 2025-04-07
Cannot enter a link in the signature editor — input field unresponsive


piotrsz0772:
### Steps to reproduce

1. Go to Settings > Signature in the Mail app.
2. Click on the hyperlink icon to insert a link.
3. Try to type a URL into the input field.

### Expected behavior

You should be able to type or paste a URL into the field and confirm it. The link should be inserted into the signature.

### Actual behavior

Clicking the hyperlink icon opens the link input popup, but the input field is completely unresponsive — no typing is possible, and the field doesn’t react at all. Nothing can be entered.

![Image](https://github.com/user-attachments/assets/31792834-d487-43e6-8857-780f5022b7f1)

### Mail app version

4.3.5

### Nextcloud version

31.0.2

### PHP engine version

None

### Web server

None

### Database

None

chung03:
This is because the input fields created by the Link plugin for ckeditor is not added to the focus trap. I actually fixed this bug in my own version of the mail app:

------------------------ src/components/TextEditor.vue ------------------------
@@ -399,8 +399,24 @@ export default {
 					this.$emit('mention', { email: item.email[0], label: item.label })
 				}
 			}, { priority: 'high' })

+			// Try to add all elements created by the Link plugin to the focus trap
+			// Add link form listener
+			this.linkUIPluginInstance = editor.plugins.get('LinkUI')
+			this.isLinkUIFormViewInFocusTrap = false;
+
+			editor.model.document.on( 'change', (eventInfo , batch) => {
+				// Only add the Link UI Form View to the focus trap if it is not already added
+				if(this.linkUIPluginInstance.formView != null && !this.isLinkUIFormViewInFocusTrap)
+				{
+					this.$emit('show-toolbar', this.linkUIPluginInstance.formView.urlInputView.fieldView.element);
+					this.isLinkUIFormViewInFocusTrap = true;
+				}
+				else if(this.linkUIPluginInstance.formView == null && this.isLinkUIFormViewInFocusTrap)
+				{
+					this.isLinkUIFormViewInFocusTrap = false;
+				}
+			} );


Basically, this change adds the input element created by Link UI to the focus trap. This allows the browser to focus on the input, and then the user can interact with the input field. This is a rough solution and needs some cleanup, though

ChristophWurst:
Thank you for your report @piotrsz0772 and kudos on the debugging @chung03 🙏 

@st3iny what do you think about the suggestions?

[...]

Issue #9605, opened 2024-04-26
Cannot create a link in the CKEditor


handkerchief333:
### Steps to reproduce

1. Create a message
2. Select a text and try to create a link

### Expected behavior

Create a link

### Actual behavior

Nothing can be inserted into the "Link-Url" field. It seems that the focus of the cursor is still on the tool menu and not in the link URL field, which is why nothing can be inserted.

![link-not-possible](https://github.com/nextcloud/mail/assets/41755733/12b222c7-6b91-4109-b84c-23b63f39b24a)

Some error messages (perhaps nothing at all or directly related to this issue)

In logs:
```
[PHP] Error: Optional parameter $bcc declared before required parameter $body is implicitly treated as a required parameter at //apps/mail/lib/Model/NewMessageData.php#120
	POST /index.php/apps/mail/api/drafts/move/1179
	von xxx von example um 26.04.2024, 17:47:18
```

Error in Javascript Console when the new message modal opens:
```
mail: could not find CKEditor translations for "de" 
{
  "app": "mail",
  "uid": "example",
  "level": 2,
error: Error: Cannot find module 'undefined'
code: "MODULE_NOT_FOUND"
columnNumber: 22569
fileName: "/apps/mail/js/mail.549.0be06e3ccd0f2ffcaa22.js"
lineNumber: 2
message: "Cannot find module 'undefined'"
}

```

### Mail app version

3.6.0
Nextcloud: 28.0.4

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

Tested with browser:
- Firefox 125.0.2
- Chromium 124.0.6367.60

madbob:
In my NextCloud instance I've selected a language with no CKEditor translation (Sardinian), and it correctly fallbacks in English.
I've not been able to reproduce this issue about the links in Mail 5.10, and my guess is that it is not related to the selected language. Or, at least: it may have been already fixed.

#10977 includes more details about the same issue.

PR #13364, opened 2026-07-27
fix: disabling shortcut for links when not editing html

### LINKED ISSUE #7641: Adding a link to message or signature doesn't work
### Steps to reproduce

1. Compose a new email.
2. Write the name of the link.
3. Add link by shortcut ctrls+k or from the UI.
4. Type the address.

### Expected behavior

One is able to write the address.

### Actual behavior

Instead of writing the address in the address field, one writes the address as text in the message and the link is consequently not created.

### Mail app version

2.1.1

### Mailserver or service

SMTP

### Operating system

Ubuntu 18.04

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

The same thing happens when trying to add a link to a signature.

madbob:
Even when formatting options of `TextEditor` are disabled, the keyboard shortcut to add links (Ctrl+K) is enabled. But, for some reason I've not entirely understood, the floating modal to insert the link is totally unusable and cannot be edited.

This PR [disables](https://ckeditor.com/docs/ckeditor5/latest/api/module_core_command-Command.html#function-forceDisabled) the "link" command and, consequently, [the opening of the sub-modal](https://github.com/ckeditor/ckeditor5/blob/070381b0f2800ac5881cd5a36c0c5982b8d19a67/packages/ckeditor5-link/src/linkui.ts#L815). The user who wants to add a link have to explicitely enable rich formatting (or "accidentally" enable it, for example using a mention, which is unaffected by this change).

Fixes #7641 

Please note that this do not explicitely targets similar issue related to the sometimes breaking link sub-modal, as #9605 and #10977.

### DIFF
--- a/src/components/TextEditor.vue
+++ b/src/components/TextEditor.vue
@@ -619,6 +619,11 @@ export default {
 					editor.execute('heading', { value: 'heading3' })
 					cancel()
 				})
+			} else {
+				const linkCommand = editor.commands.get('link')
+				if (linkCommand) {
+					linkCommand.forceDisabled('linking')
+				}
 			}
 
 			this.bus.on('append-to-body-at-cursor', this.appendToBodyAtCursor)


madbob:
Superseeded by #13510

PR #13376, opened 2026-07-28
[stable5.10] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (3 unfixable, 28 only fixable manually using --force)


ChristophWurst:
will trigger a rerun

PR #13387, opened 2026-07-30
feat: sync imap and smtp credentials while filling new account form

### LINKED ISSUE #2732: UX improvement: assume SMTP / IMAP servers + users + passwords to be same on setting new email account
### Feature Request

UX improvement: assume SMTP / IMAP servers to be same, if not requested otherwise

#### Summary

IMAP / SMTP servers are often simplified under one domain (URL) and service depends on an accessing port. Same goes for an user account + a password. Simplified settings are OK, but it takes too long to setup for multiple email accounts.

Advanced settings currently require user to enter separate mail servers for SMTP/IMAP as well as separate user + password. These are most of the time the same, so there could be a checkbox to check to allow to enter different details for both connections (IMAP / SMTP) instead of that being a default screen.

This would reduce number of text inputs present at first and only further advanced config would display more text inputs.

madbob:
Fixes #2732

### DIFF
--- a/src/components/AccountForm.vue
+++ b/src/components/AccountForm.vue
@@ -74,7 +74,7 @@
 					:placeholder="t('mail', 'IMAP Host')"
 					:disabled="loading"
 					required
-					@change="clearFeedback" />
+					@change="syncCredentials" />
 				<h4 class="account-form__heading--required">
 					{{ t('mail', 'IMAP Security') }}
 				</h4>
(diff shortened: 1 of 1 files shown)


kesselb:
Thanks for your PR :+1: 

What about the hostname? 

Works well. I'd find the ux a bit weird and would prefer a checkbox instead of magic, but we can always do that later.

kesselb (approved):

madbob:
> What about the hostname?

Personal taste: in my experience (which is a very limited statistical sample...) I've never found email providers using the same hostname for both SMTP and IMAP. But I can sync also that parameter: users can always override that on the go.

madbob:
Added syncronization also for hostname.

Issue #10617, opened 2025-01-28
Confusing User Experience: Writing Mode "Plain Text" vs. HTML Signatures with Images


kesselb:
### Steps to reproduce

1. Add a mail account in Nextcloud Mail
2. The default "Writing mode" is "Plain text"
3. Add an image to the signature
4. Save
5. Start typing a new message
6. Notice the image is missing in your signature 😕 

**Settings**
![Image](https://github.com/user-attachments/assets/9e75c337-d1d2-49c6-860e-2cb89c2bc354)

**Composer**

![Image](https://github.com/user-attachments/assets/5f4d0067-f71e-418a-9aa8-6bfb961966c7)

**Composer (after enabling formatting and switching the alias once)**

![Image](https://github.com/user-attachments/assets/468c17ca-c48b-405a-9cad-73025bfdc7ee)

### Expected behavior

It should be more intuitive. 

### Actual behavior

The signature is converted automatically to plain text. 

### Additional info

Technically, plain and rich are both HTML messages, but we load a different set of plugins for CKEditor. 

Related: https://github.com/nextcloud/mail/issues/7142

kesselb:
I believe we previously discussed the issues around email formats, HTML, and the inconsistencies that arise, and decided to move forward with the current compose behavior as a starting point.

Roundcube supports multiple identities, with each identity having its own signature. However, the behavior when using an image in a signature while composing in plain text mode is the same as in Nextcloud Mail.

Thunderbird allows HTML signatures but restricts the use of images within them. Instead, images can only be used as the sole content of the signature, without any accompanying text. Additionally, the image is only visible in the HTML version of the email, not in the plain text version.

cc @nimishavijay @ChristophWurst

christiansass:
I agree that the default should allow the most options and should be switched to Rich text

kesselb:
Fixed by https://github.com/nextcloud/mail/pull/13388

PR #13388, opened 2026-07-30
fix(composer): keep signature images by switching to rich text

kesselb:
## Summary

GeneralHtmlSupport was loaded in both editor modes, so plain text mode kept  elements that toPlain() then dropped without notice on submit. Load it only in html mode and switch the editor mode when a signature contains an image.

We documented the weird ux about signature (with html elements) and using plain text at https://github.com/nextcloud/mail/issues/10617, but never improved the current state.

Adding the GeneralHtmlSupport plugin made it worse, because before the image (was unsupported) and removed by ckeditor (=> not visible in the editor). With GeneralHtmlSupport it's supported and the image is not removed. However the actual sending process will still drop it.

That PR's brings back the old behavior, by loading html only for rich text, with the small addition that we'd automatically switch to rich text. Also the signature editor is now using the editor mode from the account and shows a warning if you signature enforces html. 

image

## Steps to reproduce

- Pick editor mode plain text
- Add an image to the signature
- Open composer
- See the image in the signature
- Send message 
- Image missing

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/SignatureSettings.vue
+++ b/src/components/SignatureSettings.vue
@@ -27,16 +27,23 @@
 			@option:selected="changeIdentity" />
 		<!-- Added wrapper to give the signature editor a clear input-style border -->
 		<div class="signature-editor-wrapper">
+			<!-- Plugins are picked when the editor is created, so re-create it whenever html changes -->
 			<TextEditor
+				:key="allowHtml"
 				v-model="signature"
-				:html="true"
+				:html="allowHtml"
 				:placeholder="t('mail', 'Signature …')"
 				:bus="bus"
 				class="signature-editor-wrapper__editor" />
 		</div>
-		<p v-if="isLargeSignature" class="warning-large-signature">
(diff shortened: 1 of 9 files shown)


src/components/SignatureSettings.vue
```diff
@@ -27,16 +27,23 @@
...
-				:html="true"
+				:html="allowHtml"
 				:placeholder="t('mail', 'Signature …')"
 				:bus="bus"
 				class="signature-editor-wrapper__editor" />
 		</div>
-		<p v-if="isLargeSignature" class="warning-large-signature">
-			{{ t('mail', 'Your signature is larger than 2 MB. This may affect the performance of your editor.') }}
-		</p>
+		<!-- TRANSLATORS: "It is added to every message" refers to the signature, not to the image -->
```
odzhychko: Will this translation be picked up? Seems like it has to be one line above.

https://docs.nextcloud.com/server/stable/developer_manual/basics/translations.html#vue

odzhychko (approved):

PR #13392, opened 2026-07-31
refactor(store): replace Vue.set/Vue.delete with direct assignment

GretaD:
part of: https://github.com/nextcloud/groupware/issues/64

Vue.set()	No longer needed	
Vue.delete()		No longer needed	
## 🤖 AI (if applicable)


GVodyanov (commented):
I looked into this, and it looks good but it can't be merged into main directly because not all of these updates work on vue 2, so this will be moved to wave B

Claude did an analysis and said that around half of the instances need to be on vue3 for the reactivity to work

PR #13416, opened 2026-08-03
fix: show the "Set as default mail app" button only if possible to register the protocol handler

madbob:
Minor fix: in the Settings panel, display the "Set as default mail app" button only if `registerProtocolHandler` is available.

[Not all browsers support this](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler#browser_compatibility), and it is required a secure context (HTTPS) to actually register a protocol handler.

### DIFF
--- a/src/components/AppSettingsMenu.vue
+++ b/src/components/AppSettingsMenu.vue
@@ -13,6 +13,7 @@
 			:open.sync="showSettings">
 			<NcAppSettingsSection id="general" :name="t('mail', 'General')">
 				<NcButton
+					v-if="canRegisterProtocolHandler"
 					variant="secondary"
 					:aria-label="t('mail', 'Set as default mail app')"
 					wide
@@ -409,6 +410,10 @@ export default {
 			'getSharedTextBlocks',
 		]),
 
+		canRegisterProtocolHandler() {
+			return window.navigator.registerProtocolHandler !== undefined
+		},
+
 		useBottomReplies() {
 			return this.mainStore.getPreference('reply-mode', 'top') === 'bottom'
 		},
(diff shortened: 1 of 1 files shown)


src/components/AppSettingsMenu.vue
```diff
@@ -13,6 +13,7 @@
 			:open.sync="showSettings">
 			<NcAppSettingsSection id="general" :name="t('mail', 'General')">
 				<NcButton
+					v-if="canRegisterProtocolHandler"
```
kesselb: I'd say we should also have a v-else block explaining why it's not possible to register the app as handler.

kesselb (commented):
Thanks for the pr

src/components/AppSettingsMenu.vue
madbob: I've added an extra check about the protocol, which is the only case that can be tested.
If `window.navigator.registerProtocolHandler` is not defined it may also depend by the browser not supporting the registration of apps, ungranted permissions, blocking extensions, or any other untestable condition.

Please check my copy about the warning notices: you know I'm very bad writer!

src/components/AppSettingsMenu.vue
```diff
@@ -416,6 +414,18 @@ export default {
 			'getSharedTextBlocks',
 		]),
 
+		canRegisterProtocolHandler() {
+			if (window.navigator.registerProtocolHandler !== undefined) {
+				return true
+			} else {
+				if (window.location.protocol !== 'https:') {
+					return t('mail', 'To use Nextcloud as your default mail application, your instance have to be secured with HTTPS')
```
kevgathuku: nit: grammar
```suggestion
					return t('mail', 'To use Nextcloud as your default mail application, your instance has to be secured with HTTPS')
```

src/components/AppSettingsMenu.vue
```diff
@@ -805,14 +789,13 @@ export default {
...
-			if (window.navigator.registerProtocolHandler) {
-				const url
-					= window.location.protocol + '//' + window.location.host + generateUrl('apps/mail/compose?uri=%s')
-				try {
-					window.navigator.registerProtocolHandler('mailto', url, OC.theme.name + ' Mail')
-				} catch (err) {
-					Logger.error('could not register protocol handler', { err })
-				}
+			const url
+				= window.location.protocol + '//' + window.location.host + generateUrl('apps/mail/compose?uri=%s')
```
kevgathuku: nit: Definitely not required since it was this way before but what do you think about uring the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL_API)

```suggestion
			const url = new URL(generateUrl('apps/mail/compose?uri=%s'), window.location.origin).href
```

Not much of a difference in the end result just pointing it out as a potential improvement to avoid building the URL manually 🙂

Issue #9913, opened 2024-07-25
Some mails are displaying block code for  lists <![if !supportLists]>1.    <![endif]>


Aveyron-RetD:
### Steps to reproduce

Some mails are received with list numbers in block code

### Expected behavior

Received as list with normal formatting

### Actual behavior

<![if !supportLists]>1.    <![endif]>
see screenshot
![image](https://github.com/user-attachments/assets/e7728b1c-1b03-4e91-9f39-876be4a95fef)

### Mail app version

3.7.5

### Operating system

Nextcloud AIO on docker ubuntu 20.04

### PHP engine version

None

### Web server

None

### Database

None

ChristophWurst:
Export the email as .eml, rename to .txt and upload it here for an engineer to try and reproduce.

Aveyron-RetD:
[gestion documentaire.eml.txt](https://github.com/user-attachments/files/16377917/gestion.documentaire.eml.txt)

Aveyron-RetD:
[gestion documentaire.eml.txt](https://github.com/user-attachments/files/16378057/gestion.documentaire.eml.txt)

Aveyron-RetD:
[
[gestion.documentaire.eml(1).txt](https://github.com/user-attachments/files/16378074/gestion.documentaire.eml.1.txt)
](url)

Aveyron-RetD:
Could it be linked to the lack of MS fonts ? https://github.com/nextcloud/richdocuments/issues/3297

Aveyron-RetD:
in a diffusion list they are correctly displayed.

Original mail
![image](https://github.com/user-attachments/assets/d092aa56-10f2-4633-b39d-b367b302d88d)

in a diffusion list
![image](https://github.com/user-attachments/assets/fb69606f-2705-4251-bf42-d6c5661ee155)

Aveyron-RetD:
Ok mail received today with following behaviour
Mail received today with the cryptic message 
![image](https://github.com/user-attachments/assets/c5aa895b-470e-4791-a5b6-e294ec77fe6f)

I think this is linked to #9913 same person/same client 

HR persons sends me some fancy emails jobs with lots of styling in the mail from MS 
![image](https://github.com/user-attachments/assets/4c4a2aaf-d392-48e0-a049-d934149c7ce5)

I can see the mail in the source

Downloading it and displaying it in thunderbird works


After righting a bug report the mail displayed again with those code blocks so adding to this bug

[...]

PR #13425, opened 2026-08-04
fix(html): strip conditional markup from html mail bodies

### LINKED ISSUE #13383: Outlook conditional HTML comments are rendered as visible text
### Steps to reproduce

1. Receive an HTML email generated by Microsoft Outlook that contains Outlook conditional HTML markup.
2. Open the message in Nextcloud Mail.
3. Observe the rendered message body.

[INCIDENCIA_HTML_minimo.eml](https://github.com/user-attachments/files/30537719/INCIDENCIA_HTML_minimo.eml)

Image

### Expected behavior

Outlook conditional HTML markup should be processed correctly or ignored during rendering.

The message should be displayed without exposing Outlook-specific conditional markup such as:

<![if !mso]>
<![endif]>

The same email is displayed correctly in Outlook, Thunderbird and other mail clients.

### Actual behavior

Parts of the Outlook conditional HTML markup are rendered as visible text inside the message body.

For example:

<![if !mso]>

...

<![endif]>

This breaks the formatting of the email and exposes HTML implementation details that should remain hidden.

### Mail app version

5.5.17

### Nextcloud version

30.0.17.2

### Mailserver or service

Dovecot IMAP

### Operating system

Ubuntu 24.04

### PHP engine version

PHP 8.3

### Nextcloud memory caching

memcache.local

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

_No response_

kesselb:
Fix https://github.com/nextcloud/mail/issues/13383 https://github.com/nextcloud/mail/issues/9913

Senders wrap content in downlevel-revealed conditionals to hide it from Outlook:

`<![if !mso]>Shown everywhere except Outlook<![endif]>`

Those markers are not comments. Every non-Outlook renderer is supposed to ignore them and show the content between them, but the parser only drops them on older libxml: 2.9.10 removes them, 2.9.14 keeps them as text, so they show up in the rendered message as literal "<![if !mso]>" markers.

Strip them before purifying so a message renders the same everywhere.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/Html.php
+++ b/lib/Service/Html.php
@@ -193,6 +193,11 @@ class Html {
 
 		$purifier = new HTMLPurifier($config);
 
+		// Downlevel-revealed conditionals are not comments, so no HTMLPurifier comment
+		// setting covers them and some libxml versions leave them as visible text.
+		// Keep the original on a PCRE error; purify(null) would silently return an empty body.
+		$mailBody = preg_replace('/<!\[\s*(?:end)?if\b[^\]]*\]\s*>/i', '', $mailBody) ?? $mailBody;
+
 		$result = $purifier->purify($mailBody);
 		// eat xml parse errors within HTMLPurifier
 		libxml_clear_errors();
--- a/tests/Unit/Service/HtmlTest.php
+++ b/tests/Unit/Service/HtmlTest.php
@@ -158,6 +158,70 @@ class HtmlTest extends TestCase {
 		$this->assertStringContainsString('data-cid="image001@example.com"', $result);
 	}
 
+	/**
+	 * Non-Outlook renderers ignore the <![if !mso]> markers and show the content
+	 * between them. Some libxml versions leave the markers in as visible text.
+	 */
+	public function testSanitizeHtmlMailBodyStripsDownlevelRevealedConditionals(): void {
+		$urlGenerator = $this->createStub(IURLGenerator::class);
+		$request = $this->createStub(IRequest::class);
+		$hmacGenerator = $this->createStub(ProxyHmacGenerator::class);
+		$mailBody = <<<'HTML'
+			<!DOCTYPE html>
+			<html>
+			<head><meta charset="utf-8"><title>MSO conditional rendering test</title></head>
+			<body style="font-family: Arial, sans-serif;">
+			<p>Hello,</p>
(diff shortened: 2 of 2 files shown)


odzhychko (approved):
Fix looks like the best solution for now.

---

Some more info, I got while digging into the issue:

- the issues affects HTML content of the pattern `<![...]>`
  - another prominent such content is `<![CDATA[...]]>`
- In HTML5 `<![...]>` should be handled as comment and never shown (except in foreign content like SVGs)
  - if we would send `<![if !mso]>` to the browser it would be no issue
- it becomes an issue because `HTMLPurifier` converts `<![if !mso]>` to `&lt;![if !mso]&gt;`
  - another parser (or a change in the underlying parser of `HTMLPurifier`) might change this in the future

lib/Service/Html.php
```diff
@@ -193,6 +193,10 @@ public function sanitizeHtmlMailBody(int $messageId, string $mailBody, array $in
 
 		$purifier = new HTMLPurifier($config);
 
+		// Downlevel-revealed conditionals are not comments, so no HTMLPurifier comment
+		// setting covers them and some libxml versions leave them as visible text.
+		$mailBody = preg_replace('/<!\[\s*(?:end)?if\b[^\]]*\]\s*>/i', '', $mailBody);
```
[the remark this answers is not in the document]
kesselb: Fallback added

lib/Service/Html.php
```diff
@@ -193,6 +193,11 @@ public function sanitizeHtmlMailBody(int $messageId, string $mailBody, array $in
 
 		$purifier = new HTMLPurifier($config);
 
+		// Downlevel-revealed conditionals are not comments, so no HTMLPurifier comment
+		// setting covers them and some libxml versions leave them as visible text.
+		// Keep the original on a PCRE error; purify(null) would silently return an empty body.
+		$mailBody = preg_replace('/<!\[\s*(?:end)?if\b[^\]]*\]\s*>/i', '', $mailBody) ?? $mailBody;
```
odzhychko, line 199: > preg_last_error()

Does this need tot be called?
Warning would be nice.
But also cleaning up for next callers I guess?

SebastianKrupinski: use preg_last_error_msg() instead php 8.0+

PR #13428, opened 2026-08-05
fix: configurable timeout for connectivity tester

### LINKED ISSUE #7872: imap server not reachable
### Steps to reproduce

1. update mail app from the former(2.2.1) version to the latest(2.2.2) on nextcloud version 25.0.2
2. add a new mail account.
3. using port 143 and 25 no certificates
4. When  I want to save the settings it says  imap server not reachable.
5. from the docker image I can ping and access port 143 directly. using the exact same server name.
6. The imap server is not on the same machine.
7. with version 2.2.1 or 2.2.0 I was still able to add accounts. But now I can not because of the failing imap server test.
8. also checked with tcpdump on the Nextcloud machine and the imap server but there were no packets exchanged between these two when I pushed the save button for the account.

### Expected behavior

I expected the mail app to reach the imap server since it is reachable there and online and manully testing proves it is from the docker container.

### Actual behavior

It says imap server unreachable, and it is not sending packets to the imap server to actually test.

### Mail app version

2.2.2

### Mailserver or service

dovecot

### Operating system

docker image Linux 3.10.0-1160.59.1.el7.x86_64

### PHP engine version

PHP 7.4

### Web server

Nginx

### Database

MariaDB

### Additional info

_No response_

madbob:
Small change in `ConnectivityTester` to use a user-configurable value as timeout interval, instead of an hard-wired constant.

I've arbitrarily choose `app.mail.imap.timeout` instead of other existing configuration: it is not meant to be semantically correct, but to give a chance to user intervention.

Fixes #7872 (in particular [this comment](https://github.com/nextcloud/mail/issues/7872#issuecomment-1404924657)).

### DIFF
--- a/lib/Service/AutoConfig/ConnectivityTester.php
+++ b/lib/Service/AutoConfig/ConnectivityTester.php
@@ -10,16 +10,14 @@ declare(strict_types=1);
 
 namespace OCA\Mail\Service\AutoConfig;
 
+use OCP\IConfig;
 use Psr\Log\LoggerInterface;
 
 class ConnectivityTester {
-	public const CONNECTION_TIMEOUT = 5;
-
-	/** @var LoggerInterface */
-	protected $logger;
-
-	public function __construct(LoggerInterface $logger) {
-		$this->logger = $logger;
+	public function __construct(
+		protected IConfig $config,
+		protected LoggerInterface $logger,
+	) {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/Service/AutoConfig/ConnectivityTester.php
```diff
@@ -30,7 +28,8 @@ public function __construct(LoggerInterface $logger) {
 	 */
 	public function canConnect(string $url, int $port): bool {
 		$this->logger->debug("attempting to connect to <$url> on port <$port>");
-		$fp = @fsockopen($url, $port, $error, $errorstr, self::CONNECTION_TIMEOUT);
+		$timeout = (float)$this->config->getSystemValue('app.mail.imap.timeout', 5);
```
kesselb, line 31: Please make it an appconfig

kesselb (commented):
Thanks a lot

lib/Service/AutoConfig/ConnectivityTester.php
madbob: I'm not sure about your request...
Do you want I move the existing `app.mail.imap.timeout` configuration (used before to init the IMAP client) to the graphical UI interface? What about similar configs as `app.mail.smtp.timeout` or `app.mail.sieve.timeout`?
Or do you want I create a completely new setting, to be exposed through AppConfig? To be used just in this condition (which is: first-time configuration of a new account)?

kesselb (approved):

PR #13431, opened 2026-08-06
fix: minor adjustment in moving messages, threads and folders

madbob:
Follow-up https://github.com/nextcloud/mail/pull/13357

Implementing suggestions by @kesselb to wire up `MoveModal` with other components.


kesselb (approved):
Thanks a lot

PR #13443, opened 2026-08-07
fix(contextchat): use a consistent item id for indexed messages

kesselb:
Content was indexed as "{mailboxId}:{messageId}" but deleted by bare IMAP uid, so deleting a message never removed it from the knowledge base. Key both sides off account, mailbox and uid via a shared helper.

- [ ] Todo: Check with integration how to clear the existing data

## 🤖 AI (if applicable)

### DIFF
--- a/lib/ContextChat/ContextChatProvider.php
+++ b/lib/ContextChat/ContextChatProvider.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace OCA\Mail\ContextChat;
 
 use OCA\Mail\AppInfo\Application;
+use OCA\Mail\Db\MailboxMapper;
 use OCA\Mail\Db\Message;
 use OCA\Mail\Db\MessageMapper;
 use OCA\Mail\Events\MessageDeletedEvent;
@@ -34,12 +35,12 @@ class ContextChatProvider implements IContentProvider, IEventListener {
 	public const CONTEXT_CHAT_MESSAGE_MAX_AGE = 31557600; // 60 * 60 * 24 * 365.25 (1 year)
 	public const CONTEXT_CHAT_IMPORT_MAX_ITEMS = 1000;
 	public const CONTEXT_CHAT_JOB_INTERVAL = 300; // 60 * 5 (5 minutes)
-
 	public function __construct(
 		private TaskService $taskService,
 		private AccountService $accountService,
 		private MailManager $mailManager,
 		private MessageMapper $messageMapper,
+		private MailboxMapper $mailboxMapper,
 		private IURLGenerator $urlGenerator,
 		private IUserManager $userManager,
 		private IContentManager $contentManager,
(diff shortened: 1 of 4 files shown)


edward-ly (commented):
Also roping in @marcelklehr for additional insight. I'll test this later if I have time, but it looks good to me so far. Do we need to do some sort of migration as well (e.g. updating the IDs of existing items, or deleting and re-importing all messages again)?

kesselb:
> Do we need to do some sort of migration as well (e.g. updating the IDs of existing items, or deleting and re-importing all messages again)?

@kyteinsky and I did some brainstorming about it (in helpdesk). It would be indeed good to reset the index. The best option for now seems to change the provider id. My first idea was to trigger deleteProvider with a migration, but Anupam explain it runs async and hence we cannot really know when it's done. By using a different provider id deleting and re-importing can run in parallel.

kesselb:
>  Todo: Check with integration how to clear the existing data

Follow up

lib/ContextChat/ContextChatProvider.php
```diff
@@ -105,11 +111,29 @@ public function getAppId(): string {
...
+			return $this->urlGenerator->linkToRouteAbsolute('mail.page.index', []);
+		}
+
+		// Context chat calls this when it renders the sources of an answer, so the
+		// message id is looked up on demand rather than baked into the item id.
+		try {
+			$mailbox = $this->mailboxMapper->findById((int)$mailboxId);
+			$messageId = $this->messageMapper->getIdForUid($mailbox, (int)$uid);
+		} catch (\Throwable) {
+			// Context chat fails the whole answer if this throws
```
kyteinsky, line 125: thanks for pointing this out, fixed in https://github.com/nextcloud/context_chat/pull/266

lib/ContextChat/ContextChatProvider.php
```diff
@@ -105,11 +111,29 @@ public function getAppId(): string {
...
+		[, $mailboxId, $uid] = array_pad(explode(':', $id), 3, null);
+		if (!$mailboxId || !$uid) {
+			return $this->urlGenerator->linkToRouteAbsolute('mail.page.index', []);
+		}
+
+		// Context chat calls this when it renders the sources of an answer, so the
+		// message id is looked up on demand rather than baked into the item id.
+		try {
+			$mailbox = $this->mailboxMapper->findById((int)$mailboxId);
+			$messageId = $this->messageMapper->getIdForUid($mailbox, (int)$uid);
```
kyteinsky, line 123: `getItemUrl` is called for every item in the sources list when the output is rendered, true
but without this change, it was still not looked up until the link was clicked by the user. Here, every link would be checked even when not clicked.

not sure what's the difference between `uid` and `id` but the with `id` approach, the message was only checked when the user clicked on it and the webpage from `'mail.page.thread'` opened.

I'm not familiar with the mail's code so sorry if the point does not make much sense.

kesselb: Thanks a lot, that's a valid concern. The previous version was a cheap call. 

id = oc_mail_messages.id 
uid = identifier on the imap server, unique per mailbox, also stored as oc_mail_messages.uid

Better would be the id, but some of the events (e.g. MessageDeletedEvent) does not carry it. 

But that might change very soon, and hence I will move this PR back to draft for now and continue once the **JMAP** changes are in. At best we have id everywhere, otherwise I will add a new entrypoint that accepts mailboxId/uid doing the lookup internally.

PR #13445, opened 2026-08-09
feat: passing full reference to mailbox in BeforeMessageDeletedEvent

madbob:
Right now the `BeforeMessageDeletedEvent` delivers only the name of the parent mailbox, which is a bit limiting when handling the event.

This change attachs the whole `Mailbox`, as already happens in `MessageDeletedEvent`.


kesselb:
Thanks. 

Test failure look related.

kesselb (approved):

PR #12930, opened 2026-05-12
chore: implementrspack

SebastianKrupinski:
### Summary
- converted to rspack instead of webpack
- added css extraction

#### Rspack
Time: 5s
Size: 1.57 (mail.js)

image

#### Webpack
Time: 26s
Size: 1.69 (mail.js)

image

Unfortunately due to the  webpack specific configuration in the entry points the webpack dependency can not be fully removed yet.

PR #13449, opened 2026-08-10
refactor: move scss files out of css folder

SebastianKrupinski:
### Summary
- move all scss files out of css/ folder to src/styles
- modified all imports
- followup to https://github.com/nextcloud/mail/pull/12930


ChristophWurst (approved):
LGTM

Could have been commited as `refactor:` because that is what is done here - moving code around without changes for the end user.

SebastianKrupinski:
> Could have been commited as `refactor:` because that is what is done here - moving code around without changes for the end user.

Done

PR #13459, opened 2026-08-11
fix(migration): support the new OCP DB schema API

ChristophWurst:
OCP's ISchemaWrapper::getTable() now returns OCP\DB\Schema\ITable instead of Doctrine\DBAL\Schema\Table on newer server versions. Widen the addMailboxKey() parameter types to Table|ITable so the migration works on both, and suppress the not-yet-vendored ITable stub in psalm.

Ref https://github.com/nextcloud/server/pull/63013

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Migration/Version1140Date20220628174152.php
+++ b/lib/Migration/Version1140Date20220628174152.php
@@ -61,24 +61,20 @@ class Version1140Date20220628174152 extends SimpleMigrationStep {
 			$alterQuery = 'ALTER TABLE `%s` MODIFY `%s` longtext null;';
 
 			$accountsTable = $schema->getTable('mail_accounts');
-			$accountsSignatureColumn = $accountsTable->getColumn('signature');
 
 			$this->connection->executeStatement(
-				sprintf($alterQuery, $accountsTable->getName(), $accountsSignatureColumn->getName())
+				sprintf($alterQuery, $accountsTable->getName(), 'signature')
 			);
 
 			$aliasesTable = $schema->getTable('mail_aliases');
-			$aliasesSignatureColumn = $accountsTable->getColumn('signature');
 
 			$this->connection->executeStatement(
-				sprintf($alterQuery, $aliasesTable->getName(), $aliasesSignatureColumn->getName())
+				sprintf($alterQuery, $aliasesTable->getName(), 'signature')
 			);
 
 			unset(
 				$accountsTable,
-				$accountsSignatureColumn,
-				$aliasesTable,
-				$aliasesSignatureColumn
+				$aliasesTable
 			);
 		}
 	}
--- a/lib/Db/MessageMapper.php
+++ b/lib/Db/MessageMapper.php
@@ -1262,13 +1262,14 @@ class MessageMapper extends QBMapper {
 		if ($ids === []) {
 			return [];
 		}
+		$direction = strtoupper($sortOrder) === 'DESC' ? 'DESC' : 'ASC';
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('*')
 			->from($this->getTableName())
 			->where(
 				$qb->expr()->in('id', $qb->createParameter('ids'))
 			)
-			->orderBy($orderBy, $sortOrder);
+			->orderBy($orderBy, $direction);
 
 		$results = [];
 		foreach (array_chunk($ids, 1000) as $chunk) {
--- a/lib/Migration/Version2300Date20230120085320.php
+++ b/lib/Migration/Version2300Date20230120085320.php
@@ -30,7 +30,7 @@ class Version2300Date20230120085320 extends SimpleMigrationStep {
 		$schema = $schemaClosure();
 
 		$table = $schema->getTable('mail_smime_certificates');
-		if (!$table->hasPrimaryKey()) {
+		if ($table->getPrimaryKey() === null) {
 			$table->setPrimaryKey(['id'], 'mail_smime_certs_id_idx');
 		}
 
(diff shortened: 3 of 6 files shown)


CarlSchwan (approved):

ChristophWurst:
There are more changes necessary to make Psalm happy, but there are also legit gaps.

1. There is no replacement to enumerate all foreign keys: https://github.com/nextcloud/mail/blob/1968041ebcbd120bfce97be65d40fe636428252b/lib/Migration/Version1130Date20220412111833.php#L116
2. ModifyColumn changed in a breaking way. It's not possible to be compatible with 34 and 35 at the same time: https://github.com/nextcloud/mail/blob/1968041ebcbd120bfce97be65d40fe636428252b/lib/Migration/Version1130Date20220412111833.php#L129-L131

susnux (approved):

CarlSchwan:
> There are more changes necessary to make Psalm happy, but there are also legit gaps.
> 
>     1. There is no replacement to enumerate all foreign keys: https://github.com/nextcloud/mail/blob/1968041ebcbd120bfce97be65d40fe636428252b/lib/Migration/Version1130Date20220412111833.php#L116

This should still work, since getForeignKeys is not implemented in Table, it's forwarding the method call to the DBAL Table via the `__call` in `OC\DB\Schema\Table` 

>     2. ModifyColumn changed in a breaking way. It's not possible to be compatible with 34 and 35 at the same time: https://github.com/nextcloud/mail/blob/1968041ebcbd120bfce97be65d40fe636428252b/lib/Migration/Version1130Date20220412111833.php#L129-L131

Same, modifyColumn in OCP\DB\Schema, has some more strict type hints but at runtime this should still work as it is just forwarding the arguments.

CarlSchwan:
Looking at the latest psalm issues, I probably went to far for requiring non empty lowercase string for index. I will change that

CarlSchwan:
Last psalm issue seems to be a false positive from psalm

lib/Migration/Version1040Date20200506111214.php
```diff
@@ -59,19 +59,13 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
 			'notnull' => true,
-			'precision' => 10,
-			'scale' => 5,
-		]);
+		])->setPrecision(10)->setScale(5);
 		$table->addColumn('f1_score_important', Types::DECIMAL, [
 			'notnull' => true,
-			'precision' => 10,
-			'scale' => 5,
-		]);
```
CarlSchwan, line 74: this can be reverted, last version support precision and scale

lib/Migration/Version1140Date20220628174152.php
```diff
@@ -61,24 +61,20 @@ public function preSchemaChange(IOutput $output, Closure $schemaClosure, array $
...
-				sprintf($alterQuery, $aliasesTable->getName(), $aliasesSignatureColumn->getName())
+				sprintf($alterQuery, $aliasesTable->getName(), 'signature')
 			);
 
 			unset(
 				$accountsTable,
-				$accountsSignatureColumn,
-				$aliasesTable,
-				$aliasesSignatureColumn
+				$aliasesTable
```
CarlSchwan, line 77: same here getName return a non-empty-string

lib/Migration/Version5006Date20251015082003.php
```diff
@@ -43,12 +43,12 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
...
 		 */
 		$mailboxesTable = $schema->getTable('mail_mailboxes');
-		if (!$mailboxesTable->hasIndex('UNIQ_45754FF89B6B5FBA5E237E06')) {
+		if (!$mailboxesTable->hasIndex('uniq_45754ff89b6b5fba5e237e06')) {
 			// Nothing to do
 			return null;
 		}
 
-		$mailboxesTable->dropIndex('UNIQ_45754FF89B6B5FBA5E237E06');
+		$mailboxesTable->dropIndex('uniq_45754ff89b6b5fba5e237e06');
```
CarlSchwan, line 52: can also be reverted

lib/Migration/Version3500Date20231115184458.php
```diff
@@ -46,7 +46,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		 * @see \OCA\Mail\Migration\Version0161Date20190902103701::changeSchema
 		 */
 		foreach ($mailboxesTable->getIndexes() as $index) {
-			if ($index->isUnique() && $index->spansColumns(['account_id', 'name'])) {
+			if ($index->isUnique() && $index->getColumns() === ['account_id', 'name']) {
```
kesselb: Would be nice to also expose spansColumns because it ignores the order.

As hardening we should array diff expected and actual columns and test for an empty array.

ChristophWurst: But the other matters. An index on [a,b] is very different from [b,a].

For this specific case we only need to catch the condition where the ordering is [account_id, name], right?

kesselb: spansColumns does not ignore the order. 
The change can be reverted, spansColumns is part of IIndex now.

lib/Migration/Version1140Date20220628174152.php
ChristophWurst: The issue is that `IColumn::getName` does not exist. It *works* because of `__call` but Psalm won't like it.

CarlSchwan: https://github.com/nextcloud/server/pull/63318

PR #13474, opened 2026-08-12
fix(dav): use the default webdav client instead of patching axios

kesselb:
# Summary

Add attachment from files and Import to calendar are broken on main. 

webdav 5 is built on fetch, but our client patched webdav's request hook with axios, which stems from webdav 4. Every operation that reads a response body then fails with "response.text is not a function", e.g. adding a file from Files as an attachment or listing the user's calendars.

Use the client from @nextcloud/files/dav, which patches webdav's fetch hook and keeps the request token up to date.

## 🤖 AI (if applicable)

### DIFF
--- a/src/dav/client.js
+++ b/src/dav/client.js
@@ -4,18 +4,8 @@
  */
 
 import { getCurrentUser } from '@nextcloud/auth'
-import axios from '@nextcloud/axios'
+import { getClient as getDavClient } from '@nextcloud/files/dav'
 import { generateRemoteUrl } from '@nextcloud/router'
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Regression of https://github.com/nextcloud/mail/pull/13260 then. FYI @GretaD.

We had a previous downgrade in https://github.com/nextcloud/mail/pull/9587. The blocked upgrade at https://github.com/nextcloud/mail/pull/12743 was autoclosed.

ChristophWurst (approved):

PR #13475, opened 2026-08-12
fix(navigation): show the unified inbox icon

kesselb:
Oversight from https://github.com/nextcloud/mail/pull/13468

UNIFIED_INBOX_ID was imported but never exposed to the template, so the comparison always failed and the unified inbox rendered a generic folder icon.

## 🤖 AI (if applicable)


madbob (approved):
You are right, sorry :-\

By the way, I also missed to remove now unused reference to `UNIFIED_INBOX_ID` in `NavigationMailbox`.

PR #13477, opened 2026-08-12
feat(composer): insert images from Files

kesselb:
Register an integration in CKEditor's insert image dropdown that opens the Nextcloud file picker.

## 🤖 AI (if applicable)

### DIFF
--- a/src/ckeditor/image/FilesImagePlugin.ts
+++ b/src/ckeditor/image/FilesImagePlugin.ts
@@ -1,0 +1,96 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { getFilePickerBuilder, showError } from '@nextcloud/dialogs'
+import { translate as t } from '@nextcloud/l10n'
+import { ButtonView, IconImageAssetManager, ImageInsertUI, MenuBarMenuListItemButtonView, Plugin } from 'ckeditor5'
+import { getClient } from '../../dav/client.js'
+import logger from '../../logger.js'
+
+const MIME_TYPES = ['image/png', 'image/jpeg', 'image/gif', 'image/bmp', 'image/webp']
+
+const SIZE_LIMIT = 10 * 1024 * 1024
+
+/**
+ * Nextcloud file picker for CKEditor.
+ */
+export default class FilesImagePlugin extends Plugin {
+	static get requires() {
+		return [ImageInsertUI] as const
+	}
+
+	static get pluginName() {
+		return 'FilesImage' as const
+	}
+
+	init(): void {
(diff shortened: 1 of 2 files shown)


kesselb:
- [x] Bug: Picker allows to select multiple files, only the first is processed.

madbob:
Very cool!

I'm a little hesitant about introducing TypeScript parts just out of the blue, and mix different styles in the codebase.
Do it exists some directive/guideline/reccomendation about TypeScript adoption in Nextcloud?

Side note not meant to be targetted in this PR: the list of mime types is OK, but different from the one [in MessageController](https://github.com/nextcloud/mail/blob/a1df525b9db7f8fa74902f965f020fa9c24c6d9c/lib/Controller/MessagesController.php#L1171), which is different from the one [in AvatarService](https://github.com/nextcloud/mail/blob/a1df525b9db7f8fa74902f965f020fa9c24c6d9c/lib/Service/AvatarService.php#L51). Sooner or later, we should have a single list of "mimetypes that can be considered an image and put into an img tag"...

hamza221 (approved):
Looks good. 
+1 for the centralized Mime types allowlist especially with `lib/Controller/MessagesController.php` but not a blocker.
Using ts is the right call for new files as we aim to fully switch to typescript at some point  ✅

hamza221:
> * [ ]  Bug: Picker allows to select multiple files, only the first is processed.

I can't reproduce 🤔
Edit: 
| current (expected) | multiple (wrong in this case)  |
|--------|--------|
|image |image |

src/ckeditor/image/FilesImagePlugin.ts
```diff
@@ -0,0 +1,96 @@
...
+
+		return button
+	}
+
+	async _pickAndInsert(): Promise<void> {
+		let node
+		try {
+			const nodes = await getFilePickerBuilder(t('mail', 'Choose an image'))
+				.setMimeTypeFilter(MIME_TYPES)
+				.setMultiSelect(false)
```
hamza221, line 59: Set to false here, which is correct if you're expecting one file

kesselb:
> I can't reproduce 🤔

Thanks for checking. It's actually something different. In single mode it's currently not possible to select a different node when one is already selected. I'd overlooked that it's necessary to unselect the last node first. I filed an issue upstream because it looks unintended to me: https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/2567.

Similar to moveOrCopyAction.ts in files the picker now disabled the button if no node is selected and also inserts the filename to be added. 

Screencast showing the old state:


https://github.com/user-attachments/assets/a8a09046-47dd-44a4-afe2-f1eb744f6a3a

kesselb:
> Similar to moveOrCopyAction.ts in files the picker now disabled the button if no node is selected and also inserts the filename to be added.

https://github.com/nextcloud/mail/pull/13484

PR #13481, opened 2026-08-13
fix: enabling html ids inside email body to permit css targetting

### LINKED ISSUE #9976: Problematic email rendering (font and background color)
### Steps to reproduce

1. Using Nextcloud mail.
2. This email looks like this, has been an issue for years and not only from the sender.
image

### Expected behavior

In Google mail it looks like this.
image

### Actual behavior

It just not right, even Snappy mail it looks right.
image

### Mail app version

3.7.6

### Mailserver or service

Gmail

### Operating system

Unraid Docker.

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database

MariaDB

### Additional info

_No response_

madbob:
Examining the samples provided in #9976, and in particular [this one](https://github.com/nextcloud/mail/issues/9976#issuecomment-3650745616), I've found that the original message includes CSS style targetting IDs inside the HTML body itself.

By default HTMLPurifier strips IDs, as it is [considered](https://github.com/ezyang/htmlpurifier/blob/master/library/HTMLPurifier/ConfigSchema/schema/Attr.EnableID.txt) unsafe to keep them and risk conflicts with IDs already existing in the web application embedding the manipulated HTML. But in Mail the full email body is confined into an iframe, so no clashing is expected to happen.

Fixes #9976 

## Summary by CodeRabbit

* **Bug Fixes**
  * HTML email content now preserves permitted ID attributes during sanitization.


ChristophWurst:
Thanks!

ChristophWurst:
@coderabbit review

ChristophWurst (approved):
Makes sense and can't think of a way that these IDs break anything outside the iframe.

@kesselb would appreciate a second review from you nonetheless

I'd also suggest a specific test for this in the style of

```php
  public function testSanitizeHtmlMailBodyPreservesId(): void {
      $urlGenerator = $this->createStub(IURLGenerator::class);
      $request = $this->createStub(IRequest::class);
      $hmacGenerator = $this->createStub(ProxyHmacGenerator::class);

      $html = new Html($urlGenerator, $request, $hmacGenerator);

      $result = $html->sanitizeHtmlMailBody(42, '<p id="target">hello</p>', []);

      $this->assertStringContainsString('id="target"', $result);
  }
```

madbob:
Pushed the extra test (thanks!)

PR #13482, opened 2026-08-13
fix(composer): keep resized image dimensions in sent messages

### LINKED ISSUE #9290: Incorrect display of the signature image (dimension)
### Steps to reproduce

1. Save an imange under Signature-Settings and change the size to 50%
2. Send the email with the signature to multiple mail-clients, some of them display the signature to big (100%)
3. I don't know exactly which email clients are displaying the wrong size, but I have seen the wrong image size in reply messages. One of my clients uses outlook.com.

### Expected behavior

- In order for signature images to be displayed sharply on Retina displays, the images must be available in twice the resolution in which they are actually to be displayed. For example, the end result in the email = `300px x 80px`, which means that the image must have dimensions of `600px x 160px`.
- The correct DPI should be `72 DPI`.
- To prevent the image from being displayed in the email clients at `600px x 160px`, the following attributes and styles must be added to the image: `style="width: 300px; max-width: 300px;" width="300"`

### Actual behavior

- Under Mailbox -> Account settings -> Signature, you can simply reduce the size of an image by dragging it smaller. The following attributes are then assigned in the HTML: `class="image_resized" style="width:50%;"`
- Unfortunately, this is not enough, and in certain e-mail clients, the image is displayed at `600px x 160px` instead of `300px x 80px`, which is very unattractive.

### Mail app version

3.5.4

### Mailserver or service

Outlook

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

kesselb:
## Summary

Fix https://github.com/nextcloud/mail/issues/9290

**STR:**

- New message
- Enable formatting
- Upload/attach image
- Resize the image 
- Sent

**Main:**

Mail (and also other clients) show the image with the original dimension

**Here:**

The image is shown with the resized dimension
 

## 🤖 AI (if applicable)

## Summary by CodeRabbit

* **New Features**
  * Improved HTML image output by preserving resized pixel widths.
  * Automatically removes fixed height values so images retain their natural aspect ratio.
  * Supports resized and linked images while preserving percentage-based sizing behavior.

* **Tests**
  * Added coverage for image resizing, serialization, reopening, and HTML editor configuration.

### DIFF
--- a/src/ckeditor/image/ImageDowncastPlugin.ts
+++ b/src/ckeditor/image/ImageDowncastPlugin.ts
@@ -1,0 +1,84 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { ViewDocumentFragment, ViewElement } from 'ckeditor5'
+
+import { ImageUtils, Plugin, UpcastWriter } from 'ckeditor5'
+
+/**
+ * Parse a CSS length into whole pixels. Anything but an absolute pixel value
+ * yields null.
+ *
+ * @param value the CSS length to parse
+ */
+function toPixels(value: string | undefined): number | null {
+	if (value === undefined) {
+		return null
+	}
+
+	const pixels = /^\s*([\d.]+)\s*px\s*$/.exec(value)
+	if (pixels === null) {
+		return null
+	}
+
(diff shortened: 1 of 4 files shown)


kesselb:
cc @CreatorOfBiontium @joeldj-nl

kesselb:
@coderabbitai please review :see_no_evil:

odzhychko (approved):
Looks good and works

PR #13486, opened 2026-08-13
feat: mailbox icon in thread view

### LINKED ISSUE #9535: Show mailbox in thread view

The threading works across mailboxes, with trash and junk excluded. 

I would like to know in the thread view where a message is actually stored. 

Screenshot from 2024-04-08 14-35-20

Above is an example. One message in my inbox, the other one in sent. 
I would like to move the inbox message to a different folder. 

_No response_

_No response_

### Additional context

_No response_

madbob:
A proposal for #9535: in the thread view, I've added the icon of the parent mailbox for each message (and a tooltip with the full name of the mailbox itself).

Screenshot From 2026-08-13 19-37-10

Evaluating this feature, I've been concerned of adding the full name of the mailbox as it would eventually squeezes even more the envelope's header (especially on mobile). An icon is mostly decorative, slightly informative, and the tooltip provides the full information in unobtrusive way (in particular for non-special mailboxes not having their own recognizable icon).

Here I've also fixed a minor alignment issue with the sender line and message's preview line, and I've removed the `mailboxId` parameter for `ThreadEnvelope` (I've not really understood his utility, as the same information is carried by `envelope` and indeed the `mailboxId` binded to the local route in `Thread` impacts in the functionalities enabled for every child envelope).

Fixes #9535

### DIFF
--- a/src/components/Thread.vue
+++ b/src/components/Thread.vue
@@ -24,7 +24,6 @@
 				ref="envelopeRefs"
 				:key="env.databaseId"
 				:envelope="env"
-				:mailbox-id="$route.params.mailboxId"
 				:thread-subject="threadSubject"
 				:expanded="expandedThreads.includes(env.databaseId)"
 				:full-height="thread.length === 1"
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Thanks for looking into this!

Would appreciate input from https://nextcloud-vue-components.netlify.app/#/Components/NcChip.

Maybe a NcChip? https://nextcloud-vue-components.netlify.app/#/Components/NcChip

madbob:
Added a wrapping `NcChip`.

I would still avoid to include the textual name of the mailbox, both to save screen space and to avoid misalignment with different length names.

Screenshot From 2026-08-14 12-06-10

PR #13487, opened 2026-08-13
fix(composer): only warn about missing 'To' when cc/bcc are set

ChristophWurst:
The no-'To'-recipient hint fired whenever the auto-focused To field was blurred empty, annoying users who write the body before the recipients. Restrict it to the case it exists for: cc or bcc filled but To empty.

## 🤖 AI (if applicable)


steven-mpawulo (approved):
This is much better.

kesselb (approved):

PR #13497, opened 2026-08-16
fix: review of OutOfOffice settings form

### LINKED ISSUE #9975: Account settings autoresponder options text alignment
### Steps to reproduce

1. Add an account
2. Enable Sieve
3. Go to the autoresponder settings

### Expected behavior

Good design

### Actual behavior

image

### Mail app version

4.0.0

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Web server

None

### Database

None

### Additional info

_No response_

madbob:
Totally unrequested review of the OOO settings form, migrated to `Nc*` components.

Boldest changes:
- replacement of internal `TextEditor` with generic `NcTextArea`, applied because - after all - HTML editing was already disabled and it is not that different
- settings are hidden when "autorespond" is "off". The previous conditions to disable edit fields were, anyway, broken (`this.enabled` was tested as a boolean, while it is always a string), and my personal (and questionable!) opinion is that hiding unrequired components from the crowded settings panel helps in making it more readable

sample

Fixes #9975


kesselb:
Thanks, could you take a look at the conflicts?

> replacement of internal TextEditor with generic NcTextArea, applied because - after all - HTML editing was already disabled and it is not that different

Newlines are not working anymore

madbob:
I've preferred to entirely drop plaintext/HTML/plaintext conversion, and just use the bare version of the string.

PR #13500, opened 2026-08-17
feat(provisioning): Add possibility to use LDAP attributes within provisioning templates

DerDreschner:
At the moment, only the user id and the mail address attached to the account can be used within provisioning templates. This isn't enough for enterprise use-cases where the LDAP may contain attributes that need to be used within the provisioning template. 

To enable that use-case, this PR adds a general `%LDAP:...%` placeholder that allows reading any LDAP attribute associated with the user account. Most often, this may be used for accessing the `sAMAccountName` via `%LDAP:sAMAccountName%`, but I've built it without limitations to that attribute for flexibility. In case the read-request fails (e.g., because the attribute doesn't exist), the placeholder isn't being replaced within the template (same behavior like `%EMAIL%`).

## TO-DO

- [ ] Add documentation about accessing LDAP attributes within provisioning templates

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/Provisioning.php
+++ b/lib/Db/Provisioning.php
@@ -60,6 +60,24 @@ class Provisioning extends Entity implements JsonSerializable {
 	public const WILDCARD = '*';
 	public const MASTER_PASSWORD_PLACEHOLDER = '********';
 
+	/**
+	 * Captured names reach the directory as the requested attribute of an LDAP read,
+	 * so the class stays too narrow for any LDAP special character to pass.
+	 */
+	private const LDAP_PLACEHOLDER_PATTERN = '%LDAP:([A-Za-z][A-Za-z0-9-]*)%';
+	private const LDAP_PLACEHOLDER_REGEX = '/' . self::LDAP_PLACEHOLDER_PATTERN . '/';
+	private const LDAP_PLACEHOLDER_ANCHORED_REGEX = '/^' . self::LDAP_PLACEHOLDER_PATTERN . '$/D';
+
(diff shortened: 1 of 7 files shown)


ChristophWurst (approved):

lib/Db/Provisioning.php
```diff
@@ -60,6 +60,24 @@ class Provisioning extends Entity implements JsonSerializable {
 	public const WILDCARD = '*';
 	public const MASTER_PASSWORD_PLACEHOLDER = '********';
 
+	/**
+	 * Captured names reach the directory as the requested attribute of an LDAP read,
+	 * so the class stays too narrow for any LDAP special character to pass.
+	 */
+	private const LDAP_PLACEHOLDER_PATTERN = '%LDAP:([A-Za-z][A-Za-z0-9-]*)%';
+	private const LDAP_PLACEHOLDER_REGEX = '/' . self::LDAP_PLACEHOLDER_PATTERN . '/';
+	private const LDAP_PLACEHOLDER_ANCHORED_REGEX = '/^' . self::LDAP_PLACEHOLDER_PATTERN . '$/';
```
kesselb: Claude flagged, that the current regex also matches a trailing newline:

```bash
php -r 'var_dump(preg_match("/^[A-Za-z0-9._+@-]{1,64}\$/", "BWAYNE\n"));'
Command line code:1:
int(1)
```

Add D modifier to prevent.

DerDreschner: Good point! It's fixed now

kesselb (approved):

PR #13501, opened 2026-08-17
Replace LLM-based translation hints with local language detection

### LINKED ISSUE #12688: Use cheaper algorithm for translation need detection

As a Nextcloud admin, I want more efficient detection of the language to keep costs low. Right now the detection of the language can eat lots of tokens quickly.

Try to use a cheap model for this task, if possible. Lower quality is absolutely fine for this job.

_No response_

### Additional context

_No response_

GVodyanov:
Closes #12688

### What

Deciding whether a message needs translating used to run a text2text task per message. That burns tokens for what is only a hint, so detection now happens in the browser and the server endpoint (`messages#needsTranslation`), its prompt and its service method are gone.

### Why franc (library for local language detection)

It's the middle option of the three franc packages, as agreed in the issue. `franc-min` misses Danish, Finnish, Norwegian, Slovak, Slovenian and others; `franc-all` covers nearly everything but is ~4× the size. `franc` covers 178 languages, and it's loaded as a lazy webpack chunk (`language-detection`), so it costs nothing on first paint, `mail.js` is unchanged.

### Why cldr-core

franc reports ISO 639-3, the UI speaks 639-1, so the two need bridging. A plain 639-3 table isn't enough because franc prefers individual codes over macrolanguages, it returns `cmn`, `arb`, `pes`, which have no direct 639-1 equivalent, so Chinese, Arabic and Persian would never match. CLDR's `languageAlias` resolves both the macrolanguage and the overlong cases in one table, and it's actively maintained by the Unicode CLDR-JSON project. It doubles as a filter: a code absent from the table has no 639-1 equivalent, which also suppresses franc's habit of reading short English text as Scots.

### Detected language reaches the translation modal as the input language

`detectForeignLanguage()` returns the 639-1 code instead of a boolean. The body components emit it with the `translate` event, `Message` forwards it, and `ThreadEnvelope` passes it to `TranslationModal` as `detectedLanguage`. The modal uses it to preselect the source language **only when the provider offers no "Detect language" option**, provider-side detection is the better guess when available.

### Intentionally not detectable

franc has no data for 10 of our UI languages: Asturian, Breton, Welsh, Irish, Interlingua, Icelandic, Kabyle, Luxembourgish, Occitan and Sardinian. If your interface is in one of those, the hint is disabled for you rather than showing a permanently wrong banner. Messages *in* those languages are still detected as something close (Occitan reads as Catalan), which is fine since the banner only names your own language.

Also: the 60-character floor is counted in characters, so very short CJK messages get no hint even though the script alone would identify them. Lowering it brings back false positives on Latin text.

### How to test

1. Open a message of 60+ characters written in a language other than your UI language → "Translate this message to X" appears shortly after the body renders (the chunk loads async).
2. Open one in your own language → no banner.

The banner is hidden when no translation provider is configured. To test without setting up AI, either enable the `testing` app (its `FakeTranslateProvider` registers a translate task type), or temporarily drop this line from `src/components/NeedsTranslationInfo.vue`:

```
v-if="availableOutputLanguages.length > 0"
```

## 🤖 AI (if applicable)


package.json
```diff
@@ -45,11 +45,13 @@
     "@riophae/vue-treeselect": "^0.4.0",
     "address-rfc2822": "^2.2.4",
     "ckeditor5": "^47.6.2-alpha.0",
+    "cldr-core": "^48.2.0",
```
odzhychko: tmp: why `cldr-core`?

src/components/MessageHTMLBody.vue
```diff
@@ -118,9 +116,7 @@ export default {
 			scrolling: true,
 		}, this.$refs.iframe)
 
-		if (this.enabledFreePrompt && this.message) {
-			this.needsTranslation = await needsTranslation(this.message.databaseId)
-		}
+		this.detectedLanguage = await detectForeignLanguage(this.message.body ?? '')
```
odzhychko: tmp: detectedLanguage vs detectedForeignLanguage

src/components/TranslationModal.vue
```diff
@@ -154,6 +159,8 @@ export default {
 	async mounted() {
 		this.selectedTo = this.availableOutputLanguages.find((language) => language.value === this.userLanguage) || null
 		this.selectedFrom = this.availableInputLanguages.find((language) => language.value === 'detect_language')
```
odzhychko, line 161: question(non-blocking): Is the value `"detect_language"` and unused value (aka. dead code)?

src/components/MessageHTMLBody.vue
odzhychko: nit(non-blocking): lets call it `detectedForeignLanguage`

I was confused because I though for a moment, that `detectedLanguage` might contain the users language.

src/util/languageDetection.ts
```diff
@@ -0,0 +1,45 @@
...
+const MIN_LENGTH = 60
+
+/**
+ * Detect the language of a message body, if it differs from the user's own.
+ *
+ * @param body message body as rendered, i.e. HTML in both body components
+ * @return ISO 639-1 code of the detected language, or null when the message is
+ *         in the user's language, too short, or not detectable
+ */
+export async function detectForeignLanguage(body: string): Promise<string | null> {
```
odzhychko: nit(non-blocking): might be worth some unit test

odzhychko: > franc reports ISO 639-3, the UI speaks 639-1, so the two need bridging. A plain 639-3 table isn't enough because franc prefers individual codes over macrolanguages, it returns cmn, arb, pes, which have no direct 639-1 equivalent, so Chinese, Arabic and Persian would never match. 

That sounds easy to break during refactoring ^^

src/util/languageDetection.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	// franc speaks ISO 639-3 and prefers individual codes over macrolanguages
+	// (cmn, not zho), so the aliases double as a filter: a code without one has
+	// no 639-1 equivalent and therefore no counterpart in the UI languages.
+	const languageAlias = aliases.supplemental.metadata.alias.languageAlias
+	const toIso6391 = (code: string) => languageAlias[code]?._replacement.split(/[\s_-]/)[0]
+
+	const userLanguage = getLanguage().split(/[-_]/)[0]
+	const detectable = [...Object.keys(expressions), ...Object.values(data).flatMap((script) => Object.keys(script))]
+	if (!detectable.some((code) => toIso6391(code) === userLanguage)) {
```
odzhychko: nit(non-blocking): The JSDoc lies. Also `null` if user language is not detectable and we therefore do not know if the message is in the users language or not.

>  @return ISO 639-1 code of the detected language, or null when the message is
>          in the user's language, too short, or not detectable
 
I'm not bothered with the JSDoc. Rather a test would be nice that pins down that behavior.

odzhychko (commented):
Code look good.

A bit hard to understand for me if it works as expect for the conversion of iso-639-3 to iso-639-1 with the macrolanguages etc. I'll need to poke it a bit to make sure I understood it correctly.

Will test also try to execute the translation-flow locally to see how it works.

src/components/TranslationModal.vue
odzhychko: Never mind. "detect_language" is something by `ISynchronousProvider` (e.g. FakeTranslateProvider)

odzhychko: But now the modal uses `detect_language` by default instead of the `detectedLanguage`. Is this an issue?

src/components/TranslationModal.vue
```diff
@@ -154,6 +159,8 @@ export default {
 	async mounted() {
 		this.selectedTo = this.availableOutputLanguages.find((language) => language.value === this.userLanguage) || null
```
odzhychko: note(non-blocking): The detected language is usually only used to decide whether to offer (or not to offer) a translation modal.

If the translation service offers `"detect_language"` as available input, we preselect "Detect language" and do not automatically choose the language detected in the frontend.

This makes sense and works as before.

Image

src/components/TranslationModal.vue
odzhychko: Ah, should have read the PR description better. This explains it:

> The modal uses it to preselect the source language only when the provider offers no "Detect language" option, provider-side detection is the better guess when available.

src/components/TranslationModal.vue
```diff
@@ -154,6 +159,8 @@ export default {
...
 		this.selectedFrom = this.availableInputLanguages.find((language) => language.value === 'detect_language')
+			?? this.availableInputLanguages.find((language) => language.value === this.detectedLanguage)
```
odzhychko: note(non-blocking): This would break if `detectedLanguage` has a region/country code e.g. `pt_BR`.

`detectForeignLanguage` as it is now, can only returns primary languages cods (e.g `pt`)

src/components/TranslationModal.vue
odzhychko, line 160: issue(non-blocking, out-of-scope): the following could be considered a bug

- `userLanguage` being `pt_BR`
- `availableOutputLanguages` only including `pt`.
- `selectedTo` actually becomes `null`
- One could expect  `selectedTo` becoming `pt`

Basically same issue as https://github.com/nextcloud/mail/pull/13501/changes#r3801120524 but this one could really happen ^^

src/components/Message.vue
```diff
@@ -36,7 +36,7 @@
 			:full-height="fullHeight"
 			@load="$emit('load', $event)"
 			@print-shortcut="$emit('print-shortcut')"
-			@translate="$emit('translate')" />
+			@translate="$emit('translate', $event)" />
```
odzhychko, line 39: nit(non-blocking) many `$emit('translate', ...)` calls. Might simplify code pulling language detection up into `ThreadEnvelope`.

src/util/languageDetection.ts
```diff
@@ -0,0 +1,45 @@
...
+		import(/* webpackChunkName: "language-detection" */ 'franc'),
+		import(/* webpackChunkName: "language-detection" */ 'franc/data.js'),
+		import(/* webpackChunkName: "language-detection" */ 'franc/expressions.js'),
+		import(/* webpackChunkName: "language-detection" */ 'cldr-core/supplemental/aliases.json'),
+	])
+
+	// franc speaks ISO 639-3 and prefers individual codes over macrolanguages
+	// (cmn, not zho), so the aliases double as a filter: a code without one has
+	// no 639-1 equivalent and therefore no counterpart in the UI languages.
+	const languageAlias = aliases.supplemental.metadata.alias.languageAlias
```
odzhychko, line 25: nit(non-blocking): After loading the modules once `detectable` will never change. Might be worth caching it to not construct in from scratch every time. Also could be a `Set`. Currently `detectable` holds some duplicates.

odzhychko (approved):
Works as expected. Some pedantic nits, but none of them blocking.

---

In case someone else also uses a rather minimal instance for testing:

- Enable `testing`
  - `php occ app:enable testing`
- Enable `llm_processing`
  - `php occ config:app:set mail llm_processing --value=true --type=boolean`
- Run a worker for background task
  - `php occ background-job:worker 'OC\TaskProcessing\SynchronousBackgroundJob'`

src/components/MessageHTMLBody.vue
GVodyanov: Done, agreed the old name suggested it might hold the user's own language.

src/util/languageDetection.ts
GVodyanov: Tests added. The macrolanguage case is pinned explicitly, so a refactor that drops the CLDR bridge fails the suite instead of silently disabling Chinese, Arabic and Persian.

src/util/languageDetection.ts
GVodyanov: Both done, the doc no longer claims the null cases are only about the message, and there is a test pinning that an undetectable interface language means "no opinion" rather than "same language".

src/components/TranslationModal.vue
GVodyanov: Yeah made it strip it, good catch

src/components/TranslationModal.vue
GVodyanov: Yeah I agree, I'll do it in a subsequent PR

src/util/languageDetection.ts
GVodyanov: Sounds good, done

src/components/Message.vue
GVodyanov: The banner renders inside the body components, so they need the detected value either way tho.

odzhychko: Yeah, comment wasn't clear. I thought about eliminating passing the detected language as `$event`.

odzhychko (approved):
LGTM

src/components/TranslationModal.vue
GVodyanov: Subsequent PR made: https://github.com/nextcloud/mail/pull/13521

PR #13507, opened 2026-08-18
fix(imap): skip fetch results without a UID during sync

ChristophWurst:
Some IMAP servers/proxies return a FETCH response with an envelope but no UID, violating RFC 3501. getUid() then returns null and ImapMessageFetcherFactory::build() throws a TypeError, aborting the whole mailbox sync. The crash surfaces only after the bulk header fetch returns, so on large mailboxes the sync appears to hang before dying.

Require a non-null UID in the filter so such results are dropped instead of crashing the sync. A result without a UID cannot be persisted anyway, and findAll still advances highestKnownUid from the messages that do persist, so the sync keeps progressing.

## 🤖 AI (if applicable)


kesselb (approved):
Nice finding 👍

PR #13133, opened 2026-06-21
feat(composer): improve image handling

joeldj-nl:
Fixes https://github.com/nextcloud/mail/issues/9960
Fixes https://github.com/nextcloud/mail/issues/12097
Fixes https://github.com/nextcloud/mail/issues/11261
Fixes https://github.com/nextcloud/mail/issues/11131
Fixes https://github.com/nextcloud/mail/issues/9350
Fixes https://github.com/nextcloud/mail/issues/9290
Fixes https://github.com/nextcloud/mail/issues/10617
Fixes https://github.com/nextcloud/mail/issues/13109
Fixes https://github.com/nextcloud/mail/issues/10977
Fixes https://github.com/nextcloud/mail/issues/7641
Fixes https://github.com/nextcloud/mail/issues/7142

### Changes

This PR improves image handling in the rich text editor, adds several new ways to insert images, and fixes a number of related editor issues. I have invested quite some time into getting this working, and I am really looking forward to using these new features and fixes in the official app. I hope this can be reviewed and considered for inclusion soon.

* Added support for inserting images from a URL
* Added support for inserting images via the file manager
* Added support for SVG images
* Added support for copying and pasting images
* Added an image dropdown menu with image-related actions
* Added support for adding and editing image alt text
* Added support for adding links to images
* Added preconfigured image size options, similar to Gmail
* Added support for changing image alignment
* Fixed image alignment issues
* Fixed image resizing by using pixel-based sizes instead of percentages
* Fixed the popup for adding links (in the signature editor) to selected text not appearing
* Fixed the signature text editor not respecting the configured editor mode and always using rich text mode
* Fixed rendering external SVG images
* Fixed images appearing side-by-side instead of stacked

### Testing

Tested successfully in a fresh Nextcloud development installation using PHP 8.3.

Manual tests performed:

* Tested the new image dropdown menu
* Tested inserting an external image
* Tested inserting an image via the file manager
* Tested inserting SVG images using all available methods
* Tested the default image alignment and changing the alignment
* Tested the new pixel-based image sizing
* Tested the preconfigured image size options
* Tested how different inserted images look when sent and received in Thunderbird mobile, Thunderbird desktop, Gmail mobile, Gmail desktop, and Nextcloud Mail
* Tested adding and editing alt text for images
* Tested adding links to images
* Tested using images in signatures
* Tested sending emails with signatures that include images
* Tested that the configured writing mode now also applies to the signature text editor

### Discussion

[...]

PR #13510, opened 2026-08-18
fix(settings): mount account-settings dialog at app root

### LINKED ISSUE #9605: Cannot create a link in the CKEditor

### LINKED ISSUE #10977: Cannot enter a link in the signature editor — input field unresponsive

### LINKED ISSUE #13109: Links dont work well on ckeditor for signatures
### Steps to reproduce

1. Open account settings
2. Write a signature and try to add a link by clicking the link icon on ckeditr styling bar
3. See nothing popsup

This doesn't happen on the normal composer modal.

### Expected behavior

I can add a link on my signature

### Actual behavior

I cannot add it

### Mail app version

_No response_

### Nextcloud version

_No response_

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

_No response_

kesselb:
## Summary

Fix https://github.com/nextcloud/mail/issues/13109
Fix https://github.com/nextcloud/mail/issues/10977
Fix https://github.com/nextcloud/mail/issues/9605

Pulled from https://github.com/nextcloud/mail/pull/13133

Inside the navigation, vue-frag pulled the dialog back into the overflow-clipped sidebar after NcModal had relocated it to <body>. CKEditor's Rect.getVisible() then returned null, so the find-and-replace dialog and every balloon ended up off-screen.

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/AccountSettings.vue
+++ b/src/components/AccountSettings.vue
@@ -10,7 +10,7 @@
 		:show-navigation="true"
 		:additional-trap-elements="trapElements"
 		:name="t('mail', 'Account settings')"
-		@update:open="updateOpen">
+		@update:open="onClose">
 		<NcAppSettingsSection
 			id="alias-settings"
 			:name="t('mail', 'Aliases')">
@@ -206,7 +206,6 @@ export default {
 	data() {
 		return {
 			trapElements: [],
-			fetchActiveSieveScript: this.account.sieveEnabled,
 			loadingClassificationToggle: false,
 			systemVersion: parseInt(OC.config.version.split('.')[0], 10),
 		}
@@ -224,24 +223,26 @@ export default {
 	},
 
 	watch: {
-		open(newState, oldState) {
-			if (newState === true && this.fetchActiveSieveScript === true) {
-				logger.debug(`Load active sieve script for account ${this.account.accountId}`)
-				this.fetchActiveSieveScript = false
-				this.mainStore.fetchActiveSieveScript({
-					accountId: this.account.id,
+		scrollToSection: {
+			immediate: true,
+			handler(newState) {
+				this.$nextTick(() => {
+					this.$refs[newState]?.$el?.scrollIntoView({
+						behavior: 'smooth',
+						block: 'start',
+					})
 				})
-			}
+			},
 		},
+	},
 
-		scrollToSection(newState) {
-			this.$nextTick(() => {
-				this.$refs[newState]?.$el?.scrollIntoView({
-					behavior: 'smooth',
-					block: 'start',
-				})
+	mounted() {
+		if (this.account.sieveEnabled) {
+			logger.debug(`Load active sieve script for account ${this.account.accountId}`)
+			this.mainStore.fetchActiveSieveScript({
+				accountId: this.account.id,
 			})
-		},
+		}
 	},
 
 	methods: {
@@ -251,8 +252,8 @@ export default {
 			})
 		},
 
-		updateOpen() {
-			this.$emit('update:open')
+		onClose() {
+			this.$emit('close')
 		},
 
 		async onToggleClassification(classificationEnabled) {
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -82,6 +82,16 @@
 			</div>
 		</template>
 		<AppSettingsMenu :open.sync="showSettings" />
+
+		<!-- Must stay outside the #list slot: within NavigationAccount's vue-frag
+		     fragment the dialog gets pulled back into the clipped sidebar after
+		     NcModal relocated it to <body>. -->
+		<AccountSettings
+			v-if="settingsAccount"
+			:open="true"
+			:account="settingsAccount"
+			:scroll-to-section="settingsSection"
+			@close="onCloseAccountSettings" />
 	</NcAppNavigation>
 </template>
 
@@ -104,6 +114,7 @@ export default {
 	name: 'Navigation',
 	components: {
 		NcAppNavigation,
+		AccountSettings: () => import(/* webpackChunkName: "account-settings" */ './AccountSettings.vue'),
 		AppSettingsMenu,
 		NavigationAccount,
 		NavigationAccountExpandCollapse,
@@ -120,11 +131,17 @@ export default {
 		return {
 			refreshing: false,
 			showSettings: false,
+			settingsAccountId: null,
+			settingsSection: undefined,
 		}
 	},
 
 	computed: {
 		...mapStores(useOutboxStore, useMainStore),
+		settingsAccount() {
+			return this.settingsAccountId ? this.mainStore.getAccount(this.settingsAccountId) : null
+		},
+
 		menu() {
 			return this.mainStore.getAccounts
 				.filter((account) => account.id !== UNIFIED_ACCOUNT_ID)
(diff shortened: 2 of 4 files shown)


src/views/Home.vue
```diff
@@ -82,6 +97,23 @@ export default {
...
+					// Mount first (settingsAccount), then open on the next tick so the
+					// dialog's `open` watcher fires and runs its on-open side effects.
+					this.$nextTick(() => {
+						this.settingsOpen = true
+					})
+				} else {
+					this.settingsOpen = false
+				}
+			},
+		},
```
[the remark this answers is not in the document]
madbob: I was just wondering why to use an "immediate" watch?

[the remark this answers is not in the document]
kesselb: The block reads a bit different after my last push.

Main: The component always exist inside the navigation.

Here: The component is created on demand.

When creating the component for the first time, showAccountSettings has true. Without immediate the watcher would not fire.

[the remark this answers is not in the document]
madbob: Why should `showAccountSettings` be `true`? It is [inited to `null`](https://github.com/nextcloud/mail/blob/ab08c6af9d14252ab2cbfbd8b975390b790376a1/src/store/mainStore.js#L105).
I've removed `immediate: true` and tested all possible conditions to open the account's settings (three-dots menu, [from "Mail Settings"](https://github.com/nextcloud/mail/blob/ab08c6af9d14252ab2cbfbd8b975390b790376a1/src/components/AppSettingsMenu.vue#L597) and [on the error notice](https://github.com/nextcloud/mail/blob/ab08c6af9d14252ab2cbfbd8b975390b790376a1/src/components/Navigation.vue#L173)) always starting from a freshly reloaded page: it is always opened at first click.

Perhaps I'm missing something: my Vue is quite modest... My concern is just avoid `immediate: true` when not strictly required: it is rarely used across the application, and may lead to some confusion while reading at the code. If kept, I would recommend at least one line of comment to explain the motivation.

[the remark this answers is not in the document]
kesselb: Good spot, immediate is gone now.

src/components/AccountSettings.vue
```diff
@@ -10,7 +10,7 @@
 		:show-navigation="true"
 		:additional-trap-elements="trapElements"
 		:name="t('mail', 'Account settings')"
-		@update:open="updateOpen">
+		@update:open="onClose">
```
madbob, line 13:
```suggestion
		@close="onClose">
```

As pointed out by CoPilot and already fixed [elsewhere](https://github.com/nextcloud/mail/pull/13510#discussion_r3806331300).

src/views/Home.vue
```diff
@@ -82,6 +97,20 @@ export default {
 
 			this.hasComposerSession = true
 		},
+
+		showAccountSettings: {
+			handler(settings) {
```
madbob:
``` suggestion
		showAccountSettings(settings) {
```

Keeping coherent style.

src/components/AccountSettings.vue
```diff
@@ -224,24 +223,26 @@ export default {
...
 
 	watch: {
-		open(newState, oldState) {
-			if (newState === true && this.fetchActiveSieveScript === true) {
-				logger.debug(`Load active sieve script for account ${this.account.accountId}`)
-				this.fetchActiveSieveScript = false
-				this.mainStore.fetchActiveSieveScript({
-					accountId: this.account.id,
+		scrollToSection: {
+			immediate: true,
```
madbob, line 227: For the sake of completeness: I've tested also removing this immediate, and here makes total sense as - without this immediate + nextTick - the panel do not scrolls to the required section.

madbob:
Wait: I've found an issue...

Having at least one account configured:
- go to "Mail Settings"
- click "Add Account"
- you are routed to "/apps/mail/setup". Which is a different view than Home.vue
- from here, existing accounts' settings modal is no longer accessible (from the three-dots menu nor from Mail Settings)

src/components/AccountSettings.vue
kesselb: https://github.com/nextcloud-libraries/nextcloud-vue/blob/adff7200b708295a835438a4e93e749445c66484/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue#L72-L74

The name of the event is `update:open`.

src/views/Home.vue
```diff
@@ -15,6 +15,16 @@
...
+
+		<!-- Must not live in the navigation: vue-frag pulls the dialog back
+		     into the clipped sidebar after NcModal relocated it to <body>. -->
+		<AccountSettings
+			v-if="settingsAccount"
+			:key="settingsAccount.id"
+			:open="settingsOpen"
+			:account="settingsAccount"
+			:scroll-to-section="settingsSection"
+			@update:open="onCloseAccountSettings" />
```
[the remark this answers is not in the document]
kesselb: I think the finding is invalid. The event handler in AccountSettings early exists when isOpen is true and hence, does not emit the update:open event for open modals: https://github.com/nextcloud-libraries/nextcloud-vue/blob/adff7200b708295a835438a4e93e749445c66484/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue#L141-L145

kesselb:
> from here, existing accounts' settings modal is no longer accessible (from the three-dots menu nor from Mail Settings)

Good finding. Resolved by mounting AccountSettings in Navigation (instead Home).

src/components/AccountSettings.vue
madbob: Totally true.
Yet: the issue initially reported by CoPilot (cit: "@update:open passes the new boolean open state, but the handler ignores it and unconditionally clears the store") is still true. `onClose` always emits `close`, even on open.
After all, this doesn't really matters in this conditions: as the dialog is created by a flag, a proper "on open" event never occours.

madbob (approved):

src/components/AccountSettings.vue
kesselb: I think the finding is invalid. The event handler in AccountSettings early exists when isOpen is true and hence, does not emit the update:open event for open modals: https://github.com/nextcloud-libraries/nextcloud-vue/blob/adff7200b708295a835438a4e93e749445c66484/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue#L141-L145

PR #13521, opened 2026-08-19
fix(TranslationModal): language code detection

GVodyanov:
Fix https://github.com/nextcloud/mail/pull/13501#discussion_r3801141549

Basically the issue was:
- Have a language with a region selected, like `en_UK`
- Open email translation modal
- English would not be automatically selected as the output language because `en != en_UK`, this makes the comparison a little smarter

## 🤖 AI (if applicable)

### DIFF
--- a/src/tests/unit/components/TranslationModal.vue.spec.js
+++ b/src/tests/unit/components/TranslationModal.vue.spec.js
@@ -1,0 +1,113 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { createLocalVue, shallowMount } from '@vue/test-utils'
+import { createPinia, setActivePinia } from 'pinia'
+import TranslationModal from '../../../components/TranslationModal.vue'
+import Nextcloud from '../../../mixins/Nextcloud.js'
+import useMainStore from '../../../store/mainStore.js'
+
+const { getLanguage } = vi.hoisted(() => ({ getLanguage: vi.fn() }))
+
+vi.mock('@nextcloud/l10n', async () => ({
+	...await vi.importActual('@nextcloud/l10n'),
+	getLanguage,
+}))
+
+const localVue = createLocalVue()
+
+localVue.mixin(Nextcloud)
+
+describe('TranslationModal', () => {
+	const language = (value) => ({ id: value, value, name: value })
+
+	const mountModal = ({ input = [], output = [], detectedForeignLanguage = null } = {}) => {
+		const store = useMainStore()
+		store.translationInputLanguages = input.map(language)
+		store.translationOutputLanguages = output.map(language)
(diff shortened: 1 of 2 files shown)


src/tests/unit/components/TranslationModal.vue.spec.js
```diff
@@ -0,0 +1,113 @@
...
+		getLanguage.mockReturnValue('en')
+	})
+
+	describe('target language', () => {
+		it('picks the interface language', () => {
+			getLanguage.mockReturnValue('de')
+
+			const view = mountModal({ output: ['en', 'de'] })
+
+			expect(view.vm.selectedTo?.value).toBe('de')
```
odzhychko, line 52: thought(non-blocking): I would have gone with asserting the UI state.
E.g. assert the select input having the language"English" selected etc.

It would hopefully be more stable and test closer to what really happens for a user.
But just asserting `view.vm.selectedTo` and `view.vm.selectedFrom` seems fine.
It's short and concise.
In case it breaks due to refactoring one can always rewrite them to assert the UI state 🤷 

If you want to change it, you can get the language names in `const language = (value) => ({ id: value, value, name: value })` through `new Intl.DisplayNames(['en'], { type: 'language' })` to get the actual name for test data.

src/tests/unit/components/TranslationModal.vue.spec.js
```diff
@@ -0,0 +1,113 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { createLocalVue, shallowMount } from '@vue/test-utils'
+import { createPinia, setActivePinia } from 'pinia'
+import TranslationModal from '../../../components/TranslationModal.vue'
+import Nextcloud from '../../../mixins/Nextcloud.js'
+import useMainStore from '../../../store/mainStore.js'
+
+const { getLanguage } = vi.hoisted(() => ({ getLanguage: vi.fn() }))
```
odzhychko, line 12: note(non-blocking): No need to mock. Can use `setLanguage` from '@nextcloud/l10n'

odzhychko: Also, if want to keep mock, `vi.mock('@nextcloud/l10n', () => ({ getLanguage: vi.fn() }))` would be enough?

odzhychko (approved):
LGTM
Feel free to apply or ignore the non-blocking comments 😆

kesselb:
Is it intentional to not re-use the detected language when requesting the translation (e.g. because the Translation Model might be better in detecting the language as the franc heuristic)?

image

GVodyanov:
> Is it intentional to not re-use the detected language when requesting the translation (e.g. because the Translation Model might be better in detecting the language as the franc heuristic)?

Yes exactly, franc doesn't really support that many languages and does occasionally confuse them, especially with few words to work off of. It's good enough for saying whether an email is the same language as the UI or not but using it to exactly determine what the other language is isn't too accurate.

PR #13525, opened 2026-08-19
chore: Update psalm to v6.16.0

kesselb:
## Summary

Refined version of https://github.com/nextcloud/mail/pull/11224

Psalm works on CI, but is broken locally with newer PHP versions. 

Main different to the previous attempt: psalm/phar is moved to vendor-bin and merged with psalm-stubs. composer.json forces php to 8.1 (for nextcloud 32). For psalm itself that is not necessary (and covered by the php version in psalm.xml). Moving the vendor-bin allows to use the latest version. 

Using vimeo/psalm (instead of the phar) failed with `Fatal error: Cannot redeclare function Amp\delay() (previously declared in vendor/amphp/amp/lib/functions.php:131) in vendor-bin/psalm/vendor/amphp/amp/src/functions.php on line 64`, hence we stick with phar for now. 

## 🤖 AI (if applicable)


kesselb:
```
Problem 1
    - psalm/phar is locked to version 6.16.1 and an update of this package was not requested.
    - psalm/phar 6.16.1 requires php ^8.2 -> your php version (8.1.34) does not satisfy that requirement.
```

I'd forgot :see_no_evil:

PR #13526, opened 2026-08-20
feat: expose message snoozing via the web API

### LINKED ISSUE #12672: Expose Snooze functionality via a Web API

Nextcloud Mail's Snooze feature is great, it's a much-needed feature that IMAP unfortunately doesn't provide out of the box.

Currently, it is only available via Nextcloud Mail's own UI, and there is no compatibility with other software providing similar functionality.

It would be awesome if other email clients (such as Thunderbird, via addons) could profit from Nextcloud Mail's server-side Snooze functionality and provide their own UI to use Nextcloud Mail as a Snooze backend.

So Nextcloud Mail would have to expose a Web API that allows clients to snooze and unsnooze messages and to query messages for whether they are snoozed and until when.

I'm currently using a Thunderbird addon but that only works on Desktop, for mobile I have to use Nextcloud Mail's UI. Both can share the same Snoozed folder, but they are not compatible in the sense that a message snoozed with one client will not be shown as snoozed by the other.

### Additional context

_No response_

steven-mpawulo:
Exposes snoozing capability on MessageApiController

Fixes #12672

### DIFF
--- a/tests/Unit/Controller/MessageApiControllerTest.php
+++ b/tests/Unit/Controller/MessageApiControllerTest.php
@@ -32,6 +32,7 @@ use OCA\Mail\Service\DkimService;
 use OCA\Mail\Service\ItineraryService;
 use OCA\Mail\Service\MailManager;
 use OCA\Mail\Service\OutboxService;
+use OCA\Mail\Service\SnoozeService;
 use OCA\Mail\Service\TrustedSenderService;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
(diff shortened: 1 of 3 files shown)


tests/Unit/Controller/MessageApiControllerTest.php
```diff
@@ -760,4 +767,48 @@ public function testNoAccount(): void {
 
 		$this->assertEquals($expected, $actual);
 	}
+
+	public function testSnooze(): void {
```
kevgathuku, line 771: What do you think about adding the test cases where the Message, Account or Mailbox is not found?

PR #13577, opened 2026-08-26
ci(test): skip PHP unit/integration tests on non-backend PRs

ChristophWurst:
Gate the PHP unit and integration matrices behind a dorny/paths-filter "backend" check so frontend-only, CSS-only, and doc-only PRs no longer trigger the ~55-minute integration matrix. Frontend unit and e2e tests still run on every PR since either stack can break them. Also add a concurrency group so superseded runs are cancelled.

The test-summary status check already tolerates skipped jobs, so the required branch-protection context is unaffected.

## 🤖 AI (if applicable)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,13 +6,43 @@ on: pull_request
 permissions:
   contents: read
 
+concurrency:
+  group: test-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 env:
   E2E_NODE_VERSION: "20"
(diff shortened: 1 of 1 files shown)


odzhychko (approved):

odzhychko:
>  Frontend unit and e2e tests still run on every PR since either stack can break them.

Frontend unit tests depend on backend code?

ChristophWurst:
No, you are right. Pure `frontend-unit-test` can also have a path filter. I'll send a another PR for that. But the e2e should always run.

PR #13579, opened 2026-08-26
test: setup component tests

odzhychko:
Adapts https://playwright.dev/docs/test-components for Vue 2.

But put stories next to tests and not component sources.
This follows https://github.com/nextcloud-libraries/nextcloud-vue.

## 🤖 AI (if applicable)


src/tests/component/snapshots/components/NewMessageModal.spec.ts-snapshots/shows-the-only-account-as-sender-1-linux.png
odzhychko, line 1: @GretaD @GVodyanov Maybe this visual regression testing setup will be useful for you during migration to Vue 3.

I suspect many components will break visually when going from `@nextcloud/vue@8` (Vue 2) to `@nextcloud/vue@9` (Vue3) because we often use `:deep` customize standard components.

odzhychko, line 1: task for me: Find out if, git lfs should be enabled.

.github/workflows/server-styling-update.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          ref: ${{ matrix.branch }}
+
+      - name: Download css
+        run: |
+          cd src/tests/component/gallery/assets/server
+          curl https://nextcloud.github.io/server/apps/theming/css/default.css > apps/theming/css/default.css
```
odzhychko, line 30: My preference is to commit the server style like https://github.com/nextcloud-libraries/nextcloud-vue/blob/main/.github/workflows/server-styling-update.yml

##### Pros

- It works offline
- It's clear which styles is tested with
- Can run `npm run test:component` without needing the checkout of sever
  - Simplifies CI running
  - Gives more flexibility in local development (e.g. fixing visual issue or creating mockups without server running)

###### Alternatives

- Directly reference https://nextcloud.github.io like https://github.com/nextcloud-libraries/nextcloud-upload
-  Assume Repo is checked out in servers app directory and use the Css from that server

Not sure how you develop with this repo, let me know if you want to keep it as proposed or do it differently.

odzhychko: @susnux I want to apply component testing including visual regression testing to a groupware apps. Do you have thoughts of a good long term solution for making server styles available for that?

I thought of creating an NPM package `nextcloud/testing` (repo `nextcloud-libraries/nextcloud-testing` that bundles styling.  NPM package would automatically be published for server versions and updated through renovate in using repos.

```html
<head>
	<link href="node_modules/nextcloud/testing/styles/server/stable35/apps/theming/css/default.css" rel="stylesheet" />
	<link href="node_modules/nextcloud/testing/styles/server/stable35/core/css/server.css" rel="stylesheet" />
</head>
```

It would have `/server/stable35/`, `/server/stable34`, `/server/master` etc. for matrix testing.

This would remove CI complexity and maintance for this from individual apps (mail, calendar, contacts) and libraries (nextcloud-vue).

src/tests/component/snapshots/components/NewMessageModal.spec.ts-snapshots/shows-the-only-account-as-sender-1-linux.png
odzhychko: Not for now. Didn't find usage of Git LFS in org:nextcloud or org:nextcloud-libraries. Also hard to change back to not having LFS. Will consider enabling it if repo size growth actually to big because of snapshots.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,123 @@
+# This workflow is provided via the organization template repository
+#
+# https://github.com/nextcloud-libraries/.github
+# https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization
+#
+# SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+
+name: Playwright Tests
```
odzhychko, line 9: Gate run only for fronted changes.
See https://github.com/nextcloud/mail/pull/13586

odzhychko: Maybe even wrap in to the workflow that runs the node tests. Also simplify, because we do not need parllel running.

.github/workflows/server-styling-update.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_dispatch:
+  schedule:
+    - cron: "0 5 * * *"
+
+jobs:
+  server-styling-update:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        branch: ['main', 'stable8']
```
odzhychko, line 19: Use correct branches

odzhychko:
The draft is outdated. Will recreate it once it is setup in Calendar https://github.com/nextcloud/calendar/pull/8798 and probably only after Vue 3 migration https://github.com/nextcloud/mail/issues/13622.

PR #13584, opened 2026-08-27
ci(test): skip frontend unit tests on non-frontend PRs

ChristophWurst:
Follow-up to #13577. Extend the changes job with a frontend filter and gate frontend-unit-test on it, so PRs that touch only PHP, docs, or CI config no longer run the vitest suite. The filter covers src, the package manifests, and the frontend tool configs that affect the run (tsconfig, vitest, babel).

frontend-e2e-tests stays ungated: both the frontend and backend can break it. The test-summary check already tolerates skipped jobs.

## 🤖 AI (if applicable)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,7 @@ jobs:
       pull-requests: read
     outputs:
       backend: ${{ steps.changes.outputs.backend }}
+      frontend: ${{ steps.changes.outputs.frontend }}
     steps:
       - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
         id: changes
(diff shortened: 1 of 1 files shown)


ChristophWurst:
:bulb: I'll combine backend and frontend path filters to only run e2e when either of them hits -> another PR :grimacing:

.github/workflows/test.yml
```diff
@@ -38,6 +39,14 @@ jobs:
               - 'vendor-bin/**'
               - 'composer.json'
               - 'composer.lock'
+            frontend:
+              - '.github/workflows/**'
```
odzhychko, line 43: suggestion(non-blocking): Maybe add the `webpack.**` files. The technically are not relevant for unit test but are "frontend".

odzhychko (approved):

odzhychko (approved):

PR #13586, opened 2026-08-27
ci(test): run e2e tests only when back- or frontend changes

ChristophWurst:
Follow-up to #13584. The e2e suite builds the frontend and exercises it against the PHP backend, so gate it on the union of the backend and frontend path filters: it runs whenever either stack changes and is skipped only for PRs that touch neither (docs, l10n, images). The matrix helper job, which exists solely to feed e2e, is gated on the same condition.

## 🤖 AI (if applicable)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,8 +37,14 @@ jobs:
               - 'tests/**'
               - 'vendor/**'
               - 'vendor-bin/**'
+              - 'composer/**'
               - 'composer.json'
               - 'composer.lock'
+              - '.patches/**'
+              - 'patches.json'
+              - 'patches.lock.json'
+              - 'resources/**'
+              - 'img/**'
             frontend:
               - '.github/workflows/**'
               - 'src/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -316,7 +318,8 @@ jobs:
   frontend-e2e-tests:
     runs-on: ubuntu-latest
     name: Front-end E2E tests
-    needs: matrix
+    needs: [changes, matrix]
+    if: needs.changes.outputs.backend != 'false' || needs.changes.outputs.frontend != 'false'
```
odzhychko, line 330: note(non-blocking): Can be simplified. `frontend-e2e-tests` already depends on `matrix` and `matrix` now already is conditional.

odzhychko, line 330: note(non-blocking,out-of-scope): We might be missing the following files in our filter.
- `.patches` -> backend
- `composer/` -> backend
- `css/` -> frontend
- `img/` -> frontend and backend
	- backend only for `$content = file_get_contents(__DIR__ . '/../../img/blocked-image.png');`
- `resources/` -> backend
- `patches.json` -> backend
- `patches.lock.json` -> backend

odzhychko (approved):

.github/workflows/test.yml
ChristophWurst: I get that it works without but I prefer to have it explicit

PR #13595, opened 2026-08-31
ci(codecov): carry forward coverage for skipped test flags

ChristophWurst:
Since the test jobs are gated on path filters, PRs that touch only the frontend (or neither stack) skip the backend unit/integration jobs, so no unittests/integrationtests coverage is uploaded. Codecov then never posts the required project/unit and project/integration statuses and the PR is stuck waiting for them (e.g. #13492).

Enable carryforward for both flags so Codecov reuses the parent commit's coverage when a flag is not uploaded, letting the statuses resolve instead of hanging.

Assisted-by: Claude:claude-opus-4-8

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

PR #13492, opened 2026-08-14
chore:change translate as, into t and n

GretaD:
@nextcloud/l10n version 3+ requires { t } { n }

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

PR #13596, opened 2026-08-31
ci(codecov): pass coverage checks when tests are skipped

ChristophWurst:
Path-filtered test gating skips the backend unit/integration jobs on PRs that do not touch PHP, so no coverage is uploaded and Codecov's required `project/unit`/`project/integration` statuses never post — the PR gets stuck (e.g. #13492).

This runs Codecov's `empty-upload` from a dedicated job when the backend jobs are skipped, so Codecov processes the commit and marks the coverage statuses as passing. `empty-upload` only passes when every changed file is Codecov-ignored, so the non-PHP source paths (`src`, `css`, `img`) are added to the ignore list.

Replaces #13595 (carryforward doesn't help the zero-upload case).

_Drafted with AI assistance (Claude Code, Opus 4.8) — maintainer to review/reword._

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -15,3 +15,6 @@ coverage:
 ignore:
   - "tests"
   - "lib/Vendor"
+  - "src"
+  - "css"
+  - "img"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -391,6 +391,20 @@ jobs:
     env:
       CI: true
 
+  codecov-empty:
+    runs-on: ubuntu-latest-low
+    needs: changes
+    if: needs.changes.outputs.backend == 'false'
+    name: codecov-empty-upload
+    steps:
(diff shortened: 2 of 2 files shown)


ChristophWurst:
Force merging because this PR touched workflows, which triggers real test runs and therefore CI doesn't show us if this really works. Rebasing https://github.com/nextcloud/mail/pull/13492 now to see it live :crossed_fingers:

codecov.yml
```diff
@@ -15,3 +15,6 @@ coverage:
 ignore:
   - "tests"
   - "lib/Vendor"
+  - "src"
```
odzhychko, line 18: I guess, this works only as long as we don't start adding coverage for frontend tests.

Wondering changes to `ignore` can be avoided when using empty uploads and carry forward together somehow...

But if it works for now, that fine :D

PR #13604, opened 2026-09-01
perf(internal-address): prefix userId in the idx

hamza221:
Prefix uid from mail_internal_address_uniq to avoid full table scans from 
```
SELECT
  *
FROM
  `oc_mail_internal_address`
WHERE
  `user_id` = ?
```

## 🤖 AI (if applicable)


ChristophWurst (approved):
Every method in https://github.com/nextcloud/mail/blob/main/lib/Db/InternalAddressMapper.php filters at least by the user id. New ordering makes more sense 👍

Issue #13391, opened 2026-07-31
PreviewEnhancementProcessingJob loads the whole 14-day backlog into memory at once and crash-loops


giggio:
### Steps to reproduce

1. Configure a mail account whose recent history is large — e.g. an account that has just been added, or one where a bulk import/migration gave several thousand messages a sent_at within the last 14 days, so they are all sitting with structure_analyzed = false.
2. Let background jobs run normally (system cron, every 5 minutes).
3. Wait for OCA\Mail\BackgroundJob\PreviewEnhancementProcessingJob to be picked up, and watch the PHP error log / journal.

### Expected behavior

The job processes unanalysed messages in bounded batches, so its peak memory is a function of the batch size and not of how many messages are waiting. Progress is persisted per batch, so a run that is interrupted (OOM, deploy, SIGTERM) still leaves the messages it did analyse marked as analysed, and the next run resumes rather than starting over.

### Actual behavior

The job dies with a PHP fatal and makes no progress at all, then repeats the same failure on every subsequent run — it never converges, because nothing is persisted until the very end of each mailbox.

Jul 30 19:32:27 server php[736579]: PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 692224 bytes) in /nix/store/…-nextcloud-34.0.2/3rdparty/sabre/vobject/lib/Property/Binary.php on line 61
Jul 30 19:32:27 server systemd[1]: nextcloud-cron.service: Main process exited, code=exited, status=255/EXCEPTION
Jul 30 19:32:27 server systemd[1]: nextcloud-cron.service: Consumed 42.002s CPU time over 1min 10.305s wall clock time, 585.4M memory peak

PID 736579 maps to this job (occ background-job:running, timestamps in UTC, journal in -03):

| 112416118501888000 | OCA\Mail\BackgroundJob\PreviewEnhancementProcessingJob | 2026-07-30 22:31:57 | 170 | 736579 | 21 minutes |

The failing allocation is only 676 KB, so the vobject frame is where it happened to land, not the cause — the process was already at ~511 MB of the 512 MB limit. For comparison, every other background job on this instance peaks at 27–41 MB (occ background-job:history).

The code path, on main as of today:

- PreviewEnhancementProcessingJob::run() calls PreprocessingService::process() with $limitTimestamp = now - 14 days.
[...]

PR #13619, opened 2026-09-02
perf(jobs): run preview pre-warm daily, time-insensitive

ChristophWurst:
PreviewEnhancementProcessingJob fired hourly per account. It can't be scaled per activity — run()-time setInterval() isn't persisted, so the constructor interval governs pickup — and previews are already computed on demand when a mailbox is opened, so the background job only backfills. Run it once a day and mark it time-insensitive so it can be deferred under load.

Assisted-by: ClaudeCode:claude-opus-4-8

**Critical implication**: this means that imip invitations might be processed delayed by up to a date. The mechanic is that the preview processing job find out if an email is an invitation. It will flag it. The imip job that runs every 5min will then pick it up. If the flagging is delayed so is the actual processing.

## 🤖 AI (if applicable)

### DIFF
--- a/tests/Unit/BackgroundJob/PreviewEnhancementProcessingJobTest.php
+++ b/tests/Unit/BackgroundJob/PreviewEnhancementProcessingJobTest.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Mail\Tests\Unit\BackgroundJob;
+
+use ChristophWurst\Nextcloud\Testing\TestCase;
+use OCA\Mail\BackgroundJob\PreviewEnhancementProcessingJob;
+use OCA\Mail\Service\AccountService;
+use OCA\Mail\Service\PreprocessingService;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\IJobList;
+use OCP\IUserManager;
+use PHPUnit\Framework\MockObject\MockObject;
+use Psr\Log\LoggerInterface;
+
+class PreviewEnhancementProcessingJobTest extends TestCase {
+
+	private ITimeFactory&MockObject $time;
+	private IUserManager&MockObject $userManager;
+	private AccountService&MockObject $accountService;
+	private PreprocessingService&MockObject $preprocessingService;
+	private LoggerInterface&MockObject $logger;
+	private IJobList&MockObject $jobList;
+	private PreviewEnhancementProcessingJob $job;
+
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):

AndyScherzinger (approved):

tests/Unit/BackgroundJob/PreviewEnhancementProcessingJobTest.php
```diff
@@ -0,0 +1,56 @@
...
+			$this->accountService,
+			$this->preprocessingService,
+			$this->logger,
+			$this->jobList,
+		);
+	}
+
+	public function testRunsDailyAndDeferrableUnderLoad(): void {
+		self::assertSame(24 * 3600, $this->job->getInterval());
+		self::assertFalse($this->job->isTimeSensitive());
```
provokateurin, line 54: Pretty useless test IMO. You're literally just checking what it says in the code, while there is no logic in between that would be worth testing.

provokateurin (dismissed):

tests/Unit/BackgroundJob/PreviewEnhancementProcessingJobTest.php
ChristophWurst: Fair but this is a simple unit test coverage. I don't see this as strictly bad. It's asserting expectations about the job. Doesn't matter if it involves logic or not IMO.

The test would probably look a bit less out of place if the `run` method were covered already.

provokateurin: You're just asserting that a constant value matches the expected constant value.

ChristophWurst: That is true.

This repo strives for a solid (unit) test coverage and I still see value in this specific test because it is asserting the specific aspect "does the job run only once a day and at off-peak times?".

I'll keep it.

kesselb:
Tricky. 

Why would it be an issue to run this job every 5 minutes? The delay for imip is indeed critical and something a lot of user's flagged even in current version with rather short delay (at best 10min). You see the email in the inbox, but no event in calendar. 

We discussed to trigger the processing of an imip message when a message is opened in the web client, but that is still todo. 

As far as I recall is imip is only set through the PreviewEnhancementProcessingJob, and hence running it once a day is an issue for the imip processing. 

More important for PreviewEnhancementProcessingJob is to actually add a limit (c.f.  https://github.com/nextcloud/mail/issues/13391) to avoid that adding a new mailbox causes a lot load.

Do you have any data, why it was recommended to run this job only once per day?

ChristophWurst:
The implication with the delayed imip flagging is true. Also noted that in the PR description. It's a trade-off.

> Do you have any data, why it was recommended to run this job only once per day?

Suggested internally because this was the most executed background job during a 1h30m high server load period.

PR #13621, opened 2026-09-02
perf(db): load local-message children via PK join

ChristophWurst:
findByLocalMessageIds filtered mail_recipients / mail_attachments by local_message_id IN (...). That column is NULL for nearly all rows, so its index statistics estimate ~1099 rows per value; a large IN list (a user's whole outbox+drafts, loaded on every Mail page open) pushes the estimate past the full-scan threshold and the optimizer scans the whole ~2M-row recipients table. It was the top query by rows examined on a 903-account instance, 85% full scans.

Join mail_local_messages and put the IN on its primary key instead, so the child table is reached by a ref lookup on the existing FK index at any list size. Same result set (ids originate from mail_local_messages; the FK cascades on delete), so no behavior change.

EXPLAIN of the new query:

```
+------+-------------+-------+------+---------------------------+---------------------+---------+---------+------+-------------------------------------------------+
| id | select_type | table | type | possible_keys | key | key_len | ref | rows | Extra |
+------+-------------+-------+------+---------------------------+---------------------+---------+---------+------+-------------------------------------------------+
| 1 | SIMPLE | a | ref | PRIMARY,mail_userid_index | mail_userid_index | 258 | const | 1 | Using where; Using index |
| 1 | SIMPLE | m | ALL | PRIMARY | NULL | NULL | NULL | 1017 | Using where; Using join buffer (flat, BNL join) |
| 1 | SIMPLE | r | ref | IDX_715DB7E31594979 | IDX_715DB7E31594979 | 5 | oc.m.id | 1099 | |
+------+-------------+-------+------+---------------------------+---------------------+---------+---------+------+-------------------------------------------------+
3 rows in set (0.000 sec
```

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/LocalAttachmentMapper.php
+++ b/lib/Db/LocalAttachmentMapper.php
@@ -48,10 +48,11 @@ class LocalAttachmentMapper extends QBMapper {
 			return [];
 		}
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('*')
-			->from($this->getTableName())
+		$qb->select('a.*')
+			->from($this->getTableName(), 'a')
+			->innerJoin('a', 'mail_local_messages', 'lm', $qb->expr()->eq('lm.id', 'a.local_message_id'))
 			->where(
-				$qb->expr()->in('local_message_id', $qb->createNamedParameter($localMessageIds, IQueryBuilder::PARAM_INT_ARRAY), IQueryBuilder::PARAM_INT_ARRAY)
+				$qb->expr()->in('lm.id', $qb->createNamedParameter($localMessageIds, IQueryBuilder::PARAM_INT_ARRAY), IQueryBuilder::PARAM_INT_ARRAY)
 			);
 		return $this->findEntities($qb);
 	}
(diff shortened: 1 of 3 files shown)


lib/Db/LocalAttachmentMapper.php
```diff
@@ -48,10 +48,11 @@ public function findByLocalMessageIds(array $localMessageIds): array {
 			return [];
 		}
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('*')
-			->from($this->getTableName())
+		$qb->select('a.*')
+			->from($this->getTableName(), 'a')
+			->innerJoin('a', 'mail_local_messages', 'm', $qb->expr()->eq('m.id', 'a.local_message_id'))
```
kesselb: The m aliases is used for mail_messages in many places, would prefer using a different letter here.

kesselb (approved):

lib/Db/LocalAttachmentMapper.php
ChristophWurst: llm for **l**oca**l** **m** /s

ChristophWurst:
Effect confirmed by applied patch on prod. App page no longer cause a full recipient table scan while looking for outbox email recipients

image

I just don't get why the new queries doesn't show up.

PR #13633, opened 2026-09-03
Account menu tweaks

kesselb:
## Summary

Change 1: Move folder search below calendar and classification settings. I think it fit's there much better.

Change 2: Move sieve script editor below sieve server. It takes a lot of space in the current version and is usually not required for your day work.

Change 3: Show the filter's component even with sieve disabled. Like autoresponder it gets a small hint to enable sieve. Now using a NcNoteCard.

Change 4: Add the key for each component. That resolves an issue that was introduced by https://github.com/nextcloud/mail/pull/13510 because the account component is not destroyed anymore.

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/AccountSettings.vue
+++ b/src/components/AccountSettings.vue
@@ -99,17 +99,6 @@
 			<Settings :key="account.accountId" ref="quickActions" :account="account" />
 		</NcAppSettingsSection>
 		<NcAppSettingsSection
-			v-if="account && account.sieveEnabled"
-			id="sieve-filter"
-			:name="t('mail', 'Sieve script editor')">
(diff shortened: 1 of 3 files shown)


madbob (approved):

madbob:
Tested, working and approved.

A side consideration I made some time ago (non blocking for this PR): the Sieve editor is a bit too... visible. Most of users are not able to manually write Sieve rules, yet putting an (apparently harmless) text editor in front of them may lead to unintented edits (and breakings). Maybe "hiding" it in a modal, accessible through an "Advanced settings" button (where "Advanced" is used as a deterrent), may keep the casual user a step further away from disaster.

PR #13641, opened 2026-09-05
Fix case-insensitive charset detection for ISO-2022-JP

### LINKED ISSUE #13472: ISO-2022-JP emails are garbled in Mail 5.10.12 [Minimal fix available]
### Steps to reproduce

1. Set up Nextcloud Server 34.0.0 with Nextcloud Mail 5.10.12.

2. Use a PHP environment with `mbstring` and `iconv` enabled.

3. Receive an email encoded as ISO-2022-JP, for example:

   ```text
   Content-Type: text/plain; charset=ISO-2022-JP
   Content-Transfer-Encoding: 7bit
   ```

4. Open the email using Nextcloud Mail.

5. Observe that the Japanese text is garbled or replaced with `?`.

The issue can be reproduced with emails sent by Cybozu MailWise 5.4.7.

### Expected behavior

The ISO-2022-JP encoded email should be correctly decoded and displayed as UTF-8 Japanese text.

### Actual behavior

Japanese text in ISO-2022-JP emails is displayed incorrectly, with characters being garbled or replaced with `?`.

The issue does not occur when the same message body is manually converted using PHP's `mb_convert_encoding()`.

### Mail app version

5.10.12

### Nextcloud version

docker nextcloud:34.0.0-fpm-alpine 

### Mailserver or service

Cybozu MailWise 5.4.7

### Operating system

uname -a Linux iris 7.0.0-28-generic #28~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jul  1 15:50:57 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### PHP engine version

PHP 8.3

### Nextcloud memory caching

_No response_

### Web server

Nginx

### Database

MariaDB

### Environment

* Nextcloud Server: 34.0.0
* Nextcloud Mail: 5.10.12
* PHP: Docker image based on Alpine Linux
* PHP `mbstring`: enabled
* PHP `iconv`: enabled
* iconv implementation: `libiconv`
* libiconv version: 1.18

### Investigation

The issue was traced to:

`lib/IMAP/Charset/Converter.php`

The relevant code is:

```php
if (in_array($charset, mb_list_encodings(), true)) {
    $converted = mb_convert_encoding($data, 'UTF-8', $charset);
} else {
    $converted = iconv($charset, 'UTF-8', $data);
}
```

For the affected messages, Horde provides the charset as:

```text
iso-2022-jp
```

However, in this PHP environment:

```text
in_array("iso-2022-jp", mb_list_encodings(), true)
→ false

in_array("ISO-2022-JP", mb_list_encodings(), true)
→ true
```

Because the comparison is case-sensitive, `iso-2022-jp` is not recognized as an encoding supported by mbstring, and the code falls back to `iconv()`.

### Conversion results

The original ISO-2022-JP message body was tested directly.

Using `mb_convert_encoding()`:

```php
$r = mb_convert_encoding($s, "UTF-8", "ISO-2022-JP");
```

produces the correct Japanese text.

Using `iconv()`:

```php
$r = iconv("ISO-2022-JP", "UTF-8", $s);
```

produces:

```text
Notice: iconv(): Detected an illegal character in input string
```

and returns `false`.

The data passed to `Converter::convert()` was also inspected. The affected plain-text part contained the expected ISO-2022-JP escape sequence:

```text
1b 24 42
```

(`ESC $ B`), confirming that the message body itself is valid ISO-2022-JP data at that point.

### Root cause

The issue appears to be caused by the case-sensitive comparison of the MIME charset name with the encoding names returned by `mb_list_encodings()`.

The processing path is:

```text
MIME charset: iso-2022-jp
        ↓
mb_list_encodings() comparison
        ↓
case-sensitive mismatch
        ↓
fallback to iconv()
        ↓
libiconv: "Detected an illegal character in input string"
        ↓
conversion failure / garbled text
```

### Workaround / minimal patch

The following minimal change fixes the issue in the affected environment:

```php
if (in_array(strtoupper($charset), array_map('strtoupper', mb_list_encodings()), true)) {
    $converted = mb_convert_encoding($data, 'UTF-8', $charset);
} else {
    $converted = iconv($charset, 'UTF-8', $data);
}
```

After applying this change, the affected ISO-2022-JP emails are displayed correctly.

This also confirms that the problem is not caused by corrupted email data or an inability of PHP's mbstring extension to decode ISO-2022-JP.

### Request

Please consider making charset matching case-insensitive so that MIME charset names such as `iso-2022-jp` are correctly recognized as encodings supported by mbstring.

If appropriate, please also consider backporting the fix to supported maintenance versions.

neconohitomi:
## 🤖 AI (if applicable)

* [x] The content of this PR was partly or fully generated using AI

## Summary

Fixes ISO-2022-JP email conversion when the charset is provided in lowercase.

## Problem

Some emails declare their charset as `ISO-2022-JP`, but Horde can return the charset as lowercase (`iso-2022-jp`).

`Converter` previously checked the charset against `mb_list_encodings()` using a case-sensitive comparison. As a result, `iso-2022-jp` was not recognized as an mbstring encoding and the conversion fell back to `iconv()`.

On the affected Alpine/libiconv environment, `iconv()` fails to convert these ISO-2022-JP messages, resulting in garbled email content.

## Fix

Make the charset comparison case-insensitive when checking the available mbstring encodings.

The existing charset normalization and conversion logic are otherwise unchanged.

## Tests

Added a regression test covering lowercase `iso-2022-jp`.

Tested with:

* `ConverterTest`: **15 tests, 30 assertions — OK**
* Full Mail unit test suite: **1518 tests, 4378 assertions — OK**

Fixes #13472

### DIFF
--- a/lib/IMAP/Charset/Converter.php
+++ b/lib/IMAP/Charset/Converter.php
@@ -35,7 +35,7 @@ class Converter {
 		$charset = trim($charset);
 		$lowerCharset = strtolower($charset);
 
-		return self::CHARSET_MAP[$lowerCharset] ?? $charset;
+		return self::CHARSET_MAP[$lowerCharset] ?? strtoupper($charset);
 	}
 
 	/**
@@ -45,7 +45,7 @@ class Converter {
 		/** @var list<string>|null $encodings */
 		static $encodings = null;
 		if ($encodings === null) {
-			$encodings = mb_list_encodings();
+			$encodings = array_map('strtoupper', mb_list_encodings());
 		}
 
 		return $encodings;
--- a/tests/Unit/IMAP/Charset/ConverterTest.php
+++ b/tests/Unit/IMAP/Charset/ConverterTest.php
@@ -56,10 +56,18 @@ class ConverterTest extends TestCase {
 		$iso88591MimePart_noCharset = new Horde_Mime_Part();
 		$iso88591MimePart_noCharset->setContents('בה בדף לחבר ממונרכיה, בקר בגרסה ואמנות דת');
 		// Japanese
+		$iso2022jpText = '外せ園査リツハワ題';
+		$iso2022jpBytes = mb_convert_encoding($iso2022jpText, 'ISO-2022-JP', 'UTF-8');
+
 		$iso2022jpMimePart = new Horde_Mime_Part();
 		$iso2022jpMimePart->setType('text/plain');
 		$iso2022jpMimePart->setCharset('ISO-2022-JP');
-		$iso2022jpMimePart->setContents(mb_convert_encoding('外せ園査リツハワ題', 'ISO-2022-JP', 'UTF-8'));
+		$iso2022jpMimePart->setContents($iso2022jpBytes);
+
+		$iso2022jpLowerMimePart = new Horde_Mime_Part();
+		$iso2022jpLowerMimePart->setType('text/plain');
+		$iso2022jpLowerMimePart->setCharset('iso-2022-jp');
+		$iso2022jpLowerMimePart->setContents($iso2022jpBytes);
 		// Korean (Outlook) - all ks_c_5601 spellings map to UHC (CP949). Encode
 		// with iconv to avoid depending on mbstring's UHC support, and cover the
 		// case-insensitive charset spellings.
@@ -84,7 +92,8 @@ class ConverterTest extends TestCase {
 			[$utfMimePart, '😊'],
 			[$utfMimeStreamPart, '💦'],
 			[$iso88591MimePart, 'Ümlaut'],
-			[$iso2022jpMimePart, '外せ園査リツハワ題'],
+			[$iso2022jpMimePart, $iso2022jpText],
+			[$iso2022jpLowerMimePart, $iso2022jpText],
 			[$iso88591MimePart_noCharset, 'בה בדף לחבר ממונרכיה, בקר בגרסה ואמנות דת'],
 		], $koreanCases, [
 			[$windowsMimePart, 'قام زهاء أوراقهم ما,'],


lib/IMAP/Charset/Converter.php
```diff
@@ -74,7 +74,7 @@ public function convert(Horde_Mime_Part $p): string {
 		if ($charset !== null) {
 			$normalizedCharset = $this->normalizeCharset($charset);
 			try {
-				if (in_array($normalizedCharset, $this->mbEncodings(), true)) {
+				if (in_array(strtoupper($charset), array_map('strtoupper', $this->mbEncodings()), true)) {
```
kesselb, line 77: Please move array_map(strtoupper) into the mbEncodings function.

kesselb, line 77: strtoupper($charset) should be done in normalizedCharset. At best by altering the map to also use uppercase and flip lowerCharset to upperCharset and return that version as fallback.

kesselb:
Good finding, Thanks for your pr :+1:

kesselb:
The commit message needs a DCO signoff and the Assisted-By trailer.

It's often enough to point your Agent https://github.com/nextcloud/mail/blob/main/AGENTS.md#commit-message-format and https://github.com/nextcloud/mail/pull/13641/checks?check_run_id=101341532634 to resolve it.

neconohitomi:
> The commit message needs a DCO signoff and the Assisted-By trailer.
> 
> It's often enough to point your Agent https://github.com/nextcloud/mail/blob/main/AGENTS.md#commit-message-format and https://github.com/nextcloud/mail/pull/13641/checks?check_run_id=101341532634 to resolve it.

Thanks for the review and for pointing out that the current commit is missing the required DCO `Signed-off-by` and `Assisted-by` trailers.

I used an AI agent (ChatGPT) while preparing this PR, and this makes me wonder about the following situation. I would appreciate your advice.

According to `AGENTS.md`, review changes should be submitted as fixup commits without force-pushing. However, the DCO check reports that the existing commit itself is missing the sign-off.

What would be the preferred way to fix the DCO/`Assisted-by` trailers in this PR while following the review workflow?

I can prepare the required fixup commit if that is the preferred approach.

kesselb:
> What would be the preferred way to fix the DCO/Assisted-by trailers in this PR while following the review workflow?

Force push

PR #13663, opened 2026-09-09
fix(outbox): send even when no sent mailbox is configured

### LINKED ISSUE #4153: "No sent mailbox configured" Error when sending a message

### LINKED ISSUE #10546: Can't send mail using OVH SMTP

tareko:
**Note:** Assisted by GLM 5.3.

## Problem

When an account has no sent mailbox configured (`sent_mailbox_id` is `NULL` — e.g. the IMAP server does not advertise special-use folders and the folder was never mapped), `SentMailboxHandler` aborts the send chain **before** the SMTP submission. The message strands in the outbox indefinitely with `STATUS_NO_SENT_MAILBOX`, while the UI first shows "message sent" and then an error a few seconds later.

## Decision
There are two paths to addressing this: The first is to automatically generate some default folders like "Sent" - I felt this required a wider discussion. Instead, I chose to just allow the send and log a warning.

## Changes

- `SentMailboxHandler` no longer blocks the chain: it logs a warning and continues, so the message is actually sent.
- `CopySentMessageHandler` already handles the missing-sent-mailbox case *after* the SMTP send, marking the message `STATUS_IMAP_SENT_MAILBOX_FAIL`, which the API surfaces as 202 "sent, copying will be retried".
- `STATUS_NO_SENT_MAILBOX` stays defined and mapped in `MessageApiController` for rows written by older versions.
- Updated `SentMailboxHandlerTest` to assert the chain continues and no status is set pre-send.

## Testing

- `tests/Unit/Send/` — 21 tests, 82 assertions ✅
- `tests/Unit/Service/OutboxServiceTest.php` — 14 tests ✅
- `tests/Unit/Controller/MessageApiControllerTest.php` — 19 tests ✅
- `composer cs:check` ✅
- Psalm not runnable locally (PHP 8.5 vs `^7.4 || ~8.0.0 || ~8.1.0 || ~8.2.0 || ~8.3.0`); left to CI

Fixes #10546
Fixes #4153


kesselb:
Why would you keep `SentMailboxHandler` if it's basically a no-op?

tareko:
Good point. Thank you and removed. `SentMailboxHandler` is deleted and the chain now starts at `AntiAbuseHandler`; `CopySentMessageHandler` owns the missing-sent-mailbox semantics after the send. Its existing warning now includes the account id. Pushed as a fixup commit.

PR #13668, opened 2026-09-10
fix(smtp): respect configured SMTP user for XOAUTH2 authentication

### LINKED ISSUE #13667: SMTP XOAUTH2 authentication ignores the configured SMTP user

When an account uses OAuth (`xoauth2`), the SASL XOAUTH2 identity for SMTP is hardcoded
to the account's email address:

https://github.com/nextcloud/mail/blob/main/lib/SMTP/SmtpClientFactory.php#L78-L81

```php
$params['xoauth2_token'] = new Horde_Smtp_Password_Xoauth2(
    $account->getEmail(),
    $decryptedAccessToken,
);
```

The configured SMTP user (`smtp_user`) is passed to Horde as `username`, but for XOAUTH2
the identity that actually gets used is the one inside the token object, so the setting
has no effect. For password auth the SMTP user *is* honoured, and provisioning can
already derive a distinct SMTP user via `buildSmtpUser()`, so the xoauth2 path is the
only place where this configuration is silently ignored.

**Practical impact: sending from a Microsoft 365 shared mailbox is impossible.**

Exchange Online treats the two protocols differently:

| Protocol | XOAUTH2 identity sent | Result |
|---|---|---|
| IMAP | `shared@example.org` | Works — this is what Microsoft documents for shared mailboxes |
| SMTP | `shared@example.org` | `535 5.7.3 Authentication unsuccessful` |

A shared mailbox has no sign-in credentials of its own, so it cannot be an SMTP AUTH
identity. Microsoft [documents the userName substitution for shared mailboxes](https://learn.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth#sasl-xoauth2-authentication-for-shared-mailboxes-in-office-365)
without noting that it does not apply to SMTP, and a Microsoft engineer confirms the
limitation in [this Q&A thread](https://learn.microsoft.com/en-us/answers/questions/162233/shared-mailbox-oauth-access)
describing exactly this symptom (IMAP fine, SMTP 535 5.7.3):

> Yes, we cannot use shared mailbox for SMTP AUTH client submission.

The supported approach is to authenticate as the delegate's *own* mailbox and let
Exchange Send As handle the sender address. Thunderbird does exactly this, which is why
users report that Thunderbird can send from a shared mailbox while Nextcloud Mail cannot.
Mail cannot do it today because the two identities cannot be separated.

Use the configured SMTP user as the XOAUTH2 identity when it is set, falling back to the
account's email address otherwise:

```php
$xoauth2User = $mailAccount->getOutboundUser();
if (empty($xoauth2User)) {
    $xoauth2User = $account->getEmail();
}
```

This is backwards compatible by construction: the account form already requires
`smtpUser` in the OAuth flow and pre-fills it with the email address, so for existing
accounts both values are identical and nothing changes. Only users who deliberately
configure a different SMTP user see different behaviour.

`IMAPClientFactory` should stay as it is — there the shared mailbox address is the
correct identity, and that path works today.

- **Adding the shared mailbox as an alias on the personal account.** This works and is
  what we currently advise, but the sent message is stored in the *personal* Sent folder,
  because the sent mailbox is bound to the account rather than the alias. Colleagues
  sharing the functional mailbox do not see the replies, which defeats the purpose of a
  shared mailbox.
- **Fixing it in Exchange.** Not possible; there is no tenant or app registration setting
  that makes a shared mailbox a valid SMTP AUTH identity.

### Additional context

This surfaced in the SURF Works pilot, where Nextcloud is offered to Dutch research and
education institutions. Functional/shared mailboxes are widely used there, so this
affects more organisations than just the one that reported it.

Related but distinct: #12223 (admin-managed shared mailbox access) would still need this
fix underneath it to work with Microsoft 365. #6524 and #7723 cover the IMAP and
in-Nextcloud delegation sides respectively.

There is precedent for decoupling the auth identity from the account address in this
codebase: #12442 (Dovecot master user support) does the same thing across the IMAP, SMTP
and Sieve client factories.

I have verified this end to end against a real Exchange Online shared mailbox (Full
Access + Send As granted to the personal account): IMAP works, SMTP is refused while the
shared mailbox is the XOAUTH2 identity, and sending succeeds as soon as the personal
mailbox is used instead — with the sent message landing in the shared mailbox's own Sent
Items folder.

I have a patch with unit tests ready and will open a PR.

Rikdekker:
Fixes #13667

## Summary

For `xoauth2` accounts the SASL XOAUTH2 identity used for SMTP was hardcoded to
`$account->getEmail()`, so the configured SMTP user had no effect. This makes it
impossible to authenticate as one mailbox and send as another — which is exactly what
Microsoft 365 requires for shared mailboxes.

This PR uses `getOutboundUser()` when set and falls back to the account's email address
otherwise.

## Why

Exchange Online accepts a shared mailbox as the XOAUTH2 identity for IMAP but rejects it
for SMTP AUTH with `535 5.7.3 Authentication unsuccessful`, because a shared mailbox has
no sign-in credentials of its own. A Microsoft engineer confirms this in
[this Q&A thread](https://learn.microsoft.com/en-us/answers/questions/162233/shared-mailbox-oauth-access):
"we cannot use shared mailbox for SMTP AUTH client submission."

The supported pattern is to authenticate as the delegate's own mailbox and let Exchange
Send As handle the sender address. Thunderbird works this way, which is why users report
that Thunderbird can send from a shared mailbox while Mail cannot.

## Why only SMTP

`IMAPClientFactory` deliberately keeps using the account email: for IMAP the shared
mailbox address *is* the correct identity, and Microsoft documents that substitution.
That path works today and is untouched.

## Backwards compatibility

No behaviour change for existing accounts:

- `AccountForm.vue` requires `smtpUser` in the OAuth flow as well (only the password
  fields are gated behind `!useOauth`) and pre-fills it with the email address.
- Autoconfig fills it from `smtpConfig.username`; provisioning computes it via
  `buildSmtpUser()`.
- The submit payload deliberately keeps `smtpUser` for OAuth accounts.

So for accounts created through the UI both values are already identical, and the
fallback covers CLI/provisioning cases where the column may be empty.

## Precedent

#12442 (Dovecot master user) introduced the same concept — an SMTP auth identity distinct
from the account address — across the IMAP, SMTP and Sieve factories.

## Tests

`tests/Unit/SMTP/SmtpClientFactoryTest.php` had no xoauth2 coverage at all. This PR adds
two cases:

- SMTP user set and different from the email address → used as XOAUTH2 identity
- SMTP user empty → falls back to the email address

Verified locally against a Nextcloud instance (PHP 8.5):

```
OK (3 tests, 11 assertions)
```

Reverting the production change makes the new test fail with exactly the reported
symptom, confirming the tests are meaningful:

```
-    'username' => 'user@domain.tld'
+    'username' => 'shared@domain.tld'
```

## End-to-end verification against a real Microsoft 365 tenant

Verified with an actual Exchange Online shared mailbox (`shared@example.org`), with Full Access and
Send As granted to the personal account (`user@example.org`), configured exactly as users do today:
the shared mailbox added as its own account, authorised with the personal user's OAuth
token.

| Step | XOAUTH2 identity | Result |
|---|---|---|
| IMAP (read) | `shared@example.org` | OK — 11 mailboxes listed |
| SMTP, before fix (`smtp_user` = account address) | `shared@example.org` | **Server denied authentication** |
| SMTP, after fix (`smtp_user` = `user@example.org`) | `user@example.org` | **Login OK** |
| Actually sending `From: shared@example.org` | `user@example.org` | **Message delivered** |
| Counter-test: revert `smtp_user` to `shared@example.org` | `shared@example.org` | Denied again |

So the same account, token and mailbox only differ in the SMTP auth identity, and that is
what decides between rejection and delivery. Sending as the shared mailbox works through
Exchange Send As once the identity is decoupled — which is what this PR enables.

The sent message is stored in the shared mailbox's own Sent Items folder (verified over
IMAP), so it is visible to everyone with access to that mailbox. That is a meaningful
difference from the alias-based workaround, where the copy ends up in the sender's
personal Sent folder and colleagues sharing the mailbox never see the replies.

## Context

Reported during the SURF Works pilot, which offers Nextcloud to Dutch research and
education institutions where functional/shared mailboxes are common.

## Checklist

- [x] Code is properly formatted (php-cs-fixer clean)
- [x] Unit tests added and passing
- [x] Backwards compatible
- [x] Conventional commit message with DCO sign-off

### DIFF
--- a/lib/SMTP/SmtpClientFactory.php
+++ b/lib/SMTP/SmtpClientFactory.php
@@ -75,8 +75,18 @@ class SmtpClientFactory {
 			}
 
 			$params['password'] = $decryptedAccessToken; // Not used, but Horde wants this
+			// The SASL XOAUTH2 user is the identity we authenticate as. It usually
+			// matches the account's email address, but some providers require a
+			// different one. Microsoft 365, for example, rejects shared mailboxes as
+			// SMTP AUTH identity while accepting them for IMAP. Respecting the
+			// configured SMTP user allows authenticating as the personal mailbox and
+			// sending on behalf of the shared one.
+			$xoauth2User = $mailAccount->getOutboundUser();
+			if (empty($xoauth2User)) {
+				$xoauth2User = $account->getEmail();
+			}
 			$params['xoauth2_token'] = new Horde_Smtp_Password_Xoauth2(
-				$account->getEmail(),
+				$xoauth2User,
 				$decryptedAccessToken,
 			);
 		}
--- a/tests/Unit/SMTP/SmtpClientFactoryTest.php
+++ b/tests/Unit/SMTP/SmtpClientFactoryTest.php
@@ -11,6 +11,7 @@ namespace OCA\Mail\Tests\Unit\Smtp;
 
 use ChristophWurst\Nextcloud\Testing\TestCase;
 use Horde_Mail_Transport_Smtphorde;
+use Horde_Smtp_Password_Xoauth2;
 use OCA\Mail\Account;
 use OCA\Mail\Db\MailAccount;
 use OCA\Mail\SMTP\SmtpClientFactory;
@@ -18,6 +19,7 @@ use OCA\Mail\Support\HostNameFactory;
 use OCP\IConfig;
 use OCP\Security\ICrypto;
 use PHPUnit\Framework\MockObject\MockObject;
+use ReflectionProperty;
 
 class SmtpClientFactoryTest extends TestCase {
 	/** @var IConfig|MockObject */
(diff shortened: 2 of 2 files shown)


ppardi:
Confirming this fix works in production, from an independent diagnosis of the same problem.

**Setup:** Nextcloud 34.0.4, Mail 5.12.0, Microsoft 365 with a shared mailbox (`service@example.com`) accessed by a delegate (`user@example.net`) holding Full Access and Send As.

**What we measured before finding this PR:**

- **IMAP** XOAUTH2 with the shared mailbox as the identity and the delegate's token: `a1 OK AUTHENTICATE completed` — delegation works.
- **SMTP** XOAUTH2 with the shared mailbox as the identity: `535 5.7.3 Authentication unsuccessful`, even after `Set-CASMailbox -SmtpClientAuthenticationDisabled $false` on that mailbox. Before that setting it failed with `535 5.7.139 ... SmtpClientAuthentication is disabled for the Tenant`, so the two failures are distinct — enabling SMTP AUTH for the mailbox is necessary but not sufficient.
- **SMTP** XOAUTH2 as the delegate, then `MAIL FROM:<service@example.com>`: `235` auth, `250 2.1.0 Sender OK`, message accepted. Exchange applies the Send As permission at submission time.

**Why this is confusing for users:** `SmtpClientFactory` already passes `'username' => $mailAccount->getOutboundUser()` a few lines above, and the account settings UI shows and stores an SMTP User field. So the setting looks like it is honoured for OAuth accounts when it silently isn't.

We applied the same one-line change locally and the account now sends in place from the shared mailbox, authenticating as the delegate. Happy to test a build if that helps.

PR #13683, opened 2026-09-15
fix(sync): apply the search filter as a restriction

### LINKED ISSUE #13673: Search in a large mailbox freezes the browser: sync ignores the filter when the query has a subject term
### Steps to reproduce

1. Have a mailbox with many messages spread over several years (mine: 14,911 messages, ~10,500 threads, 2015 to today).
2. In that mailbox, type a term in the search box above the envelope list. The frontend sends `to:X from:X subject:X mentions:false match:anyof`. Pick a term where at least one match is old (for example a message from 2016).
3. The search returns the few matching envelopes. Right after that (and again on every 30 s background sync while the search is active) the frontend calls `POST /apps/mail/api/mailboxes/{id}/sync` with the matching ids as `ids` and the filter as `query`.

### Expected behavior

The sync response only contains messages that match the search filter, so it stays small (a few envelopes at most).

### Actual behavior

The sync response contains every thread in the mailbox that is newer than the oldest known match, as `newMessages`. In my case one sync response was ~997 KB gzipped (~7-9 MB JSON, ~8,400 envelopes). Firefox shows "This page is slowing down Firefox" and the Mail UI freezes while it adds all those envelopes to the store. Because the sync repeats every 30 s while the search is active, it keeps happening.

I reproduced it server-side by calling `SyncService::syncMailbox()` directly with the same known ids, once with and once without the subject term:

| query | newMessages | JSON size |
|---|---|---|
| `to:X from:X subject:X mentions:false match:anyof` (what the search box sends) | 8,411 | 7.1 MB |
| `to:X from:X mentions:false match:anyof` (no subject term) | 0 | 0 |

### Cause

`SyncService::getDatabaseSyncChanges()` narrows the candidate "new" messages to the ones that match the filter by passing their UIDs to `findIdsByQuery()`:

https://github.com/nextcloud/mail/blob/f0460858291c984aa6defa9c7532ef3f1ce900db/lib/Service/Sync/SyncService.php#L141-L145

`findIdsByQuery()` treats `$uids` as the result of an IMAP body search. When the query has a subject term, it ORs the UIDs into the text conditions instead of ANDing them:

https://github.com/nextcloud/mail/blob/f0460858291c984aa6defa9c7532ef3f1ce900db/lib/Db/MessageMapper.php#L924-L937

So the WHERE clause becomes `(from matches OR to matches OR subject matches OR uid IN candidates)`, and every candidate passes. The OR makes sense for body+subject search in `MailSearch::getIdsLocally()`, but in the sync path the UIDs are meant as a restriction. Since the search box always adds a `subject:` term, the sync filter is effectively disabled for every search.

The same applies to the `changedIds` call a few lines below. There the result also picks up text matches from the whole mailbox that the client never had.

`findNewIds()` returns every thread newer than the oldest known id (by design), so the size of the leak grows with the age of the oldest match.

### Possible fix

Keep the OR semantics for the body-search case, but apply the sync's candidate set as a hard restriction. For example, a separate parameter in `findIdsByQuery()` that is always ANDed, or in `SyncService` intersect the filter result with the candidates:

```php
$matching = $this->messageMapper->findIdsByQuery($mailbox, $query, $order, null);
$newIds = array_values(array_intersect($newIds, $matching));
```

### Mail app version

5.10.10 (code on main at f046085 is the same)

### Nextcloud version

34.0.1

### Mailserver or service

Hosted IMAP (Soverin)

### Operating system

Debian 13 (official nextcloud:apache Docker image)

### PHP engine version

Other (8.5.8)

### Nextcloud memory caching

APCu (local), Redis (locking)

### Web server

Apache (supported)

### Database

PostgreSQL

### Additional info

Found while tracing an intermittent UI freeze, with help from Claude Code. The access log showed one 997 KB `mailboxes/{id}/sync` response right after a folder search, where normal syncs are ~4 KB.

kesselb:
Fix #13673

SyncService filtered its candidates with a parameter that widens the search instead of narrowing it. With a subject term in the query the filter stopped restricting, so a search returned every thread newer than the oldest match. Add a separate parameter that always narrows.

Assisted-by: ClaudeCode:claude-opus-5

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI (N/A)


lib/Db/MessageMapper.php
```diff
@@ -791,14 +799,12 @@ public function findByMessageId(Account $account, string $messageId): array {
 	}
 
 	/**
-	 * @param Mailbox $mailbox
-	 * @param SearchQuery $query
-	 * @param int|null $limit
-	 * @param int[]|null $uids
+	 * @param int[]|null $uids IMAP body search matches, combined with the text conditions via OR
```
madbob, line 802: Here you wiped docstrings about $mailbox, $query, $sortOrder, and $limit parameters.
Also: a text about those parameters (just mentioned but never explained) should be really appreciated by future contributors.

madbob (approved):

odzhychko:
> 2. developing

@kesselb Is this ready for review?

lib/Service/Sync/SyncService.php
```diff
@@ -139,16 +139,13 @@ private function getDatabaseSyncChanges(Account $account,
 		}
 		$order = $sortOrder === 'oldest' ? IMailSearch::ORDER_OLDEST_FIRST : IMailSearch::ORDER_NEWEST_FIRST;
 		if ($query !== null) {
-			// Filter new messages to those that also match the current filter
-			$newUids = $this->messageMapper->findUidsForIds($mailbox, $newIds);
-			$newIds = $this->messageMapper->findIdsByQuery($mailbox, $query, $order, null, $newUids);
+			$newIds = $this->messageMapper->findIdsByQuery($mailbox, $query, $order, null, null, $newIds);
```
odzhychko: wip: why refilter

```
$newIds = $this->messageMapper->findIdsByQuery($mailbox, $query, $order, null, null, $newIds);
```

odzhychko (approved):

PR #13708, opened 2026-09-20
fix(ui): load messages when list cannot scroll

### LINKED ISSUE #13642: List Layout only shows top 20 messages if there's no vertical scrollbar (tall screen, tiny text)
### Steps to reproduce

1. Have more than 20 messages in your E-mail Inbox
2. Open your Inbox -- for me it's `https://cloud.SOMEDOMAIN.de/index.php/apps/mail/box/MYnumericalID#`
3. In the Mail settings, set "Layout" to "list" (hope that's the name, I'm translating from German "Liste")
4. Make sure you see the list of messages in your Inbox, and make sure your window height can show more than 20 liste rows at a time (!!). (I'm using my 1920x1200 px screen in "portrait" orientation. You could even zoom out...)
5. If needed, reload the page to see that...
6. ...only the first 20 messages are shown (as list rows) in your Inbox! (That's the actual bug.)
7. Since there's no scroll bar (Window is taller than required to show those 20 list rows), there's no obvious way to load more messages, or even just to tell that there are more messages in that folder (Inbox).
8. Reduce the Window height (or zoom in) to make the vertical scroll bar appear
9. Scroll down -- additional messages (beyond the top 20) appear.

### Expected behavior

If possible (enough messages in the folder), the message list should always show so many rows that a vertical scrollbar is shown and I can use the "infinite scroll" to load more messages.

### Actual behavior

If the first 20 messages fit into the Window without a scroll bar, I'm not given any hint to tell that the folder contains more messages, nor is there a way (e.g. by scrolling) to display and open them.

### Mail app version

5.11.5

### Nextcloud version

33.0.3.2

### Mailserver or service

_No response_

### Operating system

_No response_

### PHP engine version

None

### Nextcloud memory caching

_No response_

### Web server

None

### Database

None

### Additional info

I'm not an admin of this Nextcloud instance, therefore cannot provide the technical details I've left blank.

abhinavohri:
Fixes #13642

In list layout, the first 20 messages can fit within a tall viewport. No scroll event fires, so Mail never requests the remaining messages.

To fix this, we check the infinite-scroll threshold after the message list updates. This allows the next page to load even when the list cannot yet scroll. The load-more handler also prevents overlapping requests and stops after the final page.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI.


GretaD (approved):
tested and works, thanks a lot

PR #13711, opened 2026-09-21
fix(ui): skip creating empty drafts

### LINKED ISSUE #13291: Do not save empty mails as draft

Currently, empty drafts get saved littering the draft folder.

Reproduce:

- Start writing a new mail (click "+ New message")
- Do nothing (no changes, no modifications)
- Close editor
- Empty draft is saved

If the Mail is empty, do not save it as draft.
A mail is empty, if:

- has no subject
- has no recipient
- has no attachments
- body is empty or only has the automatic signature

_No response_

### Additional context

_No response_

abhinavohri:
## Summary

Empty drafts could be created in two cases:

1. The existing guard checked `body.value`, but composer data uses `bodyHtml` or `bodyPlain`, so an empty body was not detected.
2. An automatically inserted signature made the body appear non-empty even when the user had not added any content.

This change checks the actual editor content, ignores the automatic signature when determining whether the body is empty, and applies the guard through the shared draft save path used by autosave and closing.

Fixes #13291


abhinavohri:
One related case remains unchanged:

If a user enters some content and waits for autosave, a draft is created. If they then remove all content and close the composer, the existing draft remains because it has already been saved.

Should an existing draft be deleted when the composer becomes completely empty?

